Issue #314, opened 2020-09-03
context::ContextValue cannot be compiled with Visual Studio 2015


maxgolov:
`ContextValue` has been recently introduced in SDK. Unfortunately the two `nostd::shared_ptr<...>` variant members are failing to compile with Visual Studio 2015, due to issue in `nostd::variant` backport.

This code fails to compile:

```cpp
using ContextValue = nostd::variant<bool,
                                    int64_t,
                                    uint64_t,
                                    double,
                                    nostd::shared_ptr<trace::Span>,
                                    nostd::shared_ptr<trace::SpanContext>>;
```

with the following error:

```
1>c:\work\opentelemetry-cpp\api\include\opentelemetry\nostd\variant.h(168): error C2672: 'opentelemetry::v0::nostd::detail::visitation::base::make_farray': no matching overloaded function found
1>c:\work\opentelemetry-cpp\api\include\opentelemetry\nostd\variant.h(173): note: see reference to class template instantiation 'opentelemetry::v0::nostd::detail::visitation::base::make_fmatrix_impl<F,opentelemetry::v0::nostd::detail::base<opentelemetry::v0::nostd::detail::Trait::Available,bool,int64_t,uint64_t,double,opentelemetry::v0::nostd::shared_ptr<opentelemetry::v0::trace::Span>,opentelemetry::v0::nostd::shared_ptr<opentelemetry::v0::trace::SpanContext>> &>::impl<opentelemetry::v0::nostd::integer_sequence<std::size_t>,opentelemetry::v0::nostd::integer_sequence<std::size_t>>' being compiled
```

whereas a simpler definition of the variant that does not use `nostd::shared_ptr` - compiles successfully:
```cpp
using ContextValue = nostd::variant < bool, int64_t, uint64_t, double>;
```
The issue is in a way how `nostd::variant` backport handles complex objects. Primitive types are OK.

There could be a few alternative solutions to this issue:
- declare that we cannot support Visual Studio 2015
- declare that RuntimeContext API does not work with Visual Studio 2015
- simplify the implementation to avoid positioning complex objects on variant, possibly using `raw pointers` instead
- fix an issue in implementation of `nostd::detail::visitation::base`
- explore ability to substitute the `nostd::variant` with `absl::variant` ( somewhat related to #275 - ability to substitute `nostd` with alternate implementation )

maxgolov:
[...]

Issue #215, opened 2020-07-27
Add Visual Studio 2015 + Windows 7.1 / 8.1 SDK build


maxgolov:
Visual Studio 2015 provides fairly 'complete' support of C++11. Currently we are already adding extra code to maintain older gcc-4.x for Linux; that one even predates vs2015. Some customers on Windows may still be operating on code built using Visual Studio 2015 (vc140) compiler, and may be targeting Windows 7.x and above. Our current CI loops are only testing Visual Studio 2017 (C++14) and Visual Studio 2019 (C++19), both by default targeting Windows 10.x and above. And we do not really cover Visual Studio 2015 (C++11) yet.

Recommendation is, as a minimum, add Visual Studio 2015 + CMake build using `choco install microsoft-build-tools --version=14.0.25420.1` , to make sure that the API and SDK all compiles with an older compiler. And that our backports of `nostd::` classes are all operational. There could be issues discovered that'd need fixing.

This is a blocker for adoption of OpenTelemetry SDK by a few larger codebase customers.

maxgolov:
As a minimum, our `notsd::string_view` is having this issue with Visual Studio 2015 (C++11) , esp. due to extra SCL / secure C++ coding practices (which may be a false-positive, but requires a tweak) :

```
1>c:\program files (x86)\microsoft visual studio 14.0\vc\include\xutility(2983): error C4996: 'std::equal::_Unchecked_iterators::_Deprecate': Call to 'std::equal' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'
1>c:\program files (x86)\microsoft visual studio 14.0\vc\include\xutility(2983): note: see declaration of 'std::equal::_Unchecked_iterators::_Deprecate'
1>c:\program files (x86)\microsoft visual studio 14.0\vc\include\xutility(3008): note: see reference to function template instantiation 'bool std::equal<_InIt1,_InIt2,std::equal_to<void>>(_InIt1,_InIt1,_InIt2,_Pr)' being compiled
1>        with
1>        [
1>            _InIt1=const char *,
1>            _InIt2=const char *,
1>            _Pr=std::equal_to<void>
1>        ]
[...]

PR #374, opened 2020-10-26
Using STD library for API surface

maxgolov:
I created a separate design doc to explain the logic and what's done in `docs/building-with-stdlib.md`

@jsuereth


api/include/opentelemetry/nostd/mpark/variant.h
```diff
@@ -0,0 +1,1281 @@
+// MPark.Variant
```
maxgolov, line 1: Not new, just moved here from `nostd/variant.h` to eventually eliminate - replace by `absl::variant`.

api/include/opentelemetry/nostd/variant.h
```diff
@@ -1,1281 +1,33 @@
-// MPark.Variant
+// Copyright 2020, OpenTelemetry Authors
```
maxgolov, line 1: Moved contents of MPark.Variant to `mpark/variant.h` and replaced with shim on top of Standard Library and Abseil.

api/include/opentelemetry/std/span.h
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020, OpenTelemetry Authors
```
maxgolov, line 1: Abseil's span does not strive to be fully `std::span` compatible, whereas `gsl::span` is making that commitment. This is becoming irrelevant with C++20, where `std::span` is available natively.

api/include/opentelemetry/std/variant.h
```diff
@@ -0,0 +1,232 @@
...
+// Standard Type aliases in nostd namespace
+namespace nostd
+{
+
+// nostd::variant<...>
+template <class... _Types>
+using variant = std::variant<_Types...>;
+
+#if defined(__APPLE__) && defined(_LIBCPP_USE_AVAILABILITY_APPLE)
+// Apple Platforms provide std::bad_variant_access only in newer versions of OS.
```
maxgolov, line 34: Additional complexity here is that `std::variant` in Apple LLVM is only available with certain version of Mac OS X, and the standard version detection (C++17 and above) - won't work.. We have to stub exception handling for Mac, in order to benefit from the library implementation + methods for parts that are missing.

sdk/src/common/core.cc
```diff
@@ -0,0 +1,12 @@
+#include "opentelemetry/nostd/variant.h"
```
maxgolov, line 1: `core.cc` contains the only method `absl::variant_internal::ThrowBadVariantAccess` that would otherwise be necessary for inclusion via `${CORE_RUNTIME_LIBS}` which is defined as `set(CORE_RUNTIME_LIBS absl::any absl::base absl::bits absl::city)` ... We could avoid this by borrowing `absl::variant` implementation and plugging it in instead of MPark.Variant. That is the reason why I moved the MPark.Variant out into separate folder, to replace it with variant from Abseil. This will be needed to solve these issues: #215 #314

sdk/src/trace/tracer_provider.cc
```diff
@@ -14,7 +14,8 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer> TracerProvider::G
     nostd::string_view library_name,
     nostd::string_view library_version) noexcept
 {
-  return opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer>(tracer_);
+  // TODO: do we have to transform this at all, if we can keep it nostd::shared_ptr at source?
+  return tracer_;
```
maxgolov, line 18: Basically I have a question why did we have to change the `tracer_` type to `std::shared_ptr` , that we eventually end up converting to `nostd::shared_ptr` .. Why don't we maintain this as original type, and then if we use STD library - there is no unnecessary type conversion back and forth.

tools/setup-buildtools.sh
```diff
@@ -22,13 +22,6 @@ yum -y install devtoolset-7-valgrind
 
 yum-config-manager --enable rhel-server-rhscl-7-rpms
 
-if [ `gcc --version | grep 7` == "" ]; then
```
maxgolov, line 25: This code for RPM/RedHat/CentOS should check for min-version, not equality.

jsuereth (commented):
This looks pretty good.  I'm taking a crack at seeing if I can expose the same flexibility in the bazel-build without too much effort.

api/include/opentelemetry/std/span.h
maxgolov: GSL can be used on any platform. But only if necessary - only if the runtime does not natively provide `std::span`, e.g. it is not needed at all on latest Ubuntu 20.04 (because it already has C++20 compiler with `std::span` ). Also it is not needed on _latest_ Mac OS X. In fact, it is also not needed in Visual Studio 2019 either. I'm thinking here how to add flexibility, as we could have used either `gsl::span` or `absl::span`. I picked GSL because it sets the goal to be fully compatible with the latest C++20 standard.

sdk/include/opentelemetry/sdk/common/empty_attributes.h
```diff
@@ -23,5 +25,11 @@ static const opentelemetry::common::KeyValueIterableView<std::array<std::pair<st
 
   return kEmptyAttributes;
 }
+#else
+static const opentelemetry::common::NullKeyValueIterable GetEmptyAttributes() noexcept
```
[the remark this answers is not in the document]
maxgolov: Good catch. I'll check on this. I'll consider reverting this, as it should not matter to the primary scope of this PR.

sdk/src/common/core.cc
maxgolov: Since `std::variant` is available since  C++17, we only need non-standard variant for older compilers, such as Visual Studio 2015 (where MPark variant does not compile), OR for older gcc-4.8 (Ubuntu-14.04).. Whereas any other more recent LTS version of Linux, e.g. Ubuntu-16.04 comes with gcc-5, where we won't be needing the Abseil at all.

sdk/src/trace/tracer_provider.cc
```diff
@@ -14,7 +14,8 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer> TracerProvider::G
...
 }
 
 void TracerProvider::SetProcessor(std::shared_ptr<SpanProcessor> processor) noexcept
```
maxgolov, line 25: STOP HERE and IGNORE THE REST OF PR, as it is covered in a separate PR #377

.gitmodules
```diff
@@ -1,3 +1,22 @@
 [submodule "third_party/prometheus-cpp"]
 	path = third_party/prometheus-cpp
 	url = https://github.com/jupp0r/prometheus-cpp.git
+
+[submodule "tools/vcpkg"]
+	path = tools/vcpkg
+	url = https://github.com/Microsoft/vcpkg
+	branch = master
```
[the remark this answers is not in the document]
maxgolov: We are good here. We basically pin submodule to commit id. Every time when we need to upgrade it, it's gonna be zero-diff PR that shows to what new commit ID we are moving on to... This will only be used for CMake, as Bazel manages all deps separately. The only use-case where we need this, is Prometheus library for the Metrics exporter --- right now Bazel build expects it to be there, checked out in source tree. I think we'll eventually need to fix this for Bazel, to fetch from git instead... Then for Bazel builds there would be no need to even clone the submodules: it could be 'shallow' clone without submodules.

docs/building-with-stdlib.md
```diff
@@ -0,0 +1,102 @@
...
+that the incoming types does not need to be 'transferred' across ABI boundary.
+This is the case for statically linked executables and when SDK is implemented
+as 'header-only' library.
+
+### Avoiding unnecessary extra memcpy (perf improvements)
+
+No need to transform from 'native' standard library types, e.g. `std::map` via
+`KeyValueIterable` means we can bypass that transformation process, if and when
+we know that the ABI compatibility is not a requirement in certain environment.
+ETA perf improvement is 1.5%-3% better perf since an extra transform, memcpy,
```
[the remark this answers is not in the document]
maxgolov: I'm adding a few benchmarks, yes.

docs/building-with-stdlib.md
```diff
@@ -0,0 +1,102 @@
...
+Feature allows to alias from `nostd::` to `std::` classes for C++17 and above.
+
+Consistent handling of `std::variant` across various OS:
+
+- backport of a few missing variant features, e.g. `std::get` and `std::visit`
+  for older version of Mac OS X. Patches that enable proper
+  handling of `std::visit` and `std::variant` irrespective of OS version
+  to resolve [this quirk](https://stackoverflow.com/questions/52310835/xcode-10-call-to-unavailable-function-stdvisit).
+
+- ability to optionally borrow implementation of C++20 `gsl::span` from
```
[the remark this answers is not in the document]
maxgolov: Makes sense. I'll add this to the write-up.

Appears like we'd only be needing this for:
- Visual Studio 2015 (C++11/14)
- gcc-4.8 (C++14)

Anything gcc-5+ or Visual Studio 2017+ -- won't need this, as it _should_ have `std::span`. I'm doing a runtime feature check though, so if we don't provide `span` implementation, and if the system doesn't have the `<span>` header - build would fail.

alolita:
Triaged Dec 7 2020: @maxgolov will rebase and update.

docs/building-with-stdlib.md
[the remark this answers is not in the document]
maxgolov: I added a few benchmarks that show better perf. But the full potential here is when we actually test that the incoming type is `std::map` or -alike container, then we have a template method that would no longer need to transform it. This is an extra piece of code that I'd like to add in all the places where we perform the transformation via ABI-safe `KeyValueIterable`

docs/building-with-stdlib.md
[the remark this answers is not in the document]
maxgolov: I'll submit another iteration with detailed table.

sdk/include/opentelemetry/sdk/common/empty_attributes.h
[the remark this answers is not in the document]
maxgolov: I reverted this.

.gitmodules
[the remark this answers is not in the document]
maxgolov: We will handle it nicely: if our dependency publishes their release on the master, we refer to their master Commit ID that is on of their release Tags. If they don't publish on the master, we will use their release branch. In most cases we use the master. GSL, for example, publishes their releases with one branch - master only.

api/include/opentelemetry/std/span.h
maxgolov: I added a table that lists where `gsl::span` is necessary. Also added paragraph from Abseil `absl::Span` that explains that Abseil, for example, is not targeting to be the substitute for `std::span`, whereas `gsl::span` is a 100% compatible substitute for `std::span` ... Thus, in environments where it's missing - we are using that.

api/include/opentelemetry/std/variant.h
maxgolov: I added a separate note on this in the markdown doc.

maxgolov:
@jsuereth @lalitb @ThomsonTan - I think I'm kinda done with this, with the following follow-ups that I will send in a separate PR:
- better document that explains how to cook the two flavors with CMake in Visual Studio 2019
- add CI loop for `stdlib` flavor

The only part that I feel a bit uneasy about is `CMakeList.txt` - adding `${CORE_RUNTIME_LIBS}` variable in a few spots. And it is only set only if compiling with `WITH_ABSEIL` for variant implementation. Basically it's an implementation of an exception handler in case if variant throws `nostd::bad_variant_access`. I tried to explain it in a comment here - https://github.com/open-telemetry/opentelemetry-cpp/pull/374/files#r512342232 - and this will be needed to resolve this issue later on: #215 #314 .. Right now we cannot compile with Visual Studio 2015 at all. Only Abseil variant (`absl::variant`) would work. I can take care of it in a fix for #215 ... if you feel like we'd rather remove these changes in `CMakeList.txt` - I can remove and send in a separate PR.. but it's kinda benign, because right now we don't set that variable at all yet. So it has no effect. Maybe we have to include this functionality in some sort of `opentelemetry-common.a` library, and require that instead of requiring `${CORE_RUNTIME_LIBS}`.

api/include/opentelemetry/common/attribute_value.h
```diff
@@ -10,20 +10,30 @@
...
-                                      nostd::span<const nostd::string_view>>;
+using AttributeValue =
+    nostd::variant<bool,
+                   int32_t,
+                   int64_t,
+                   uint32_t,
+                   uint64_t,
+                   double,
+                   nostd::string_view,
+#ifdef HAVE_CSTRING_TYPE
```
jsuereth, line 21: QQ For my sanity:  Do you know if adding types like this breaks ABI compatibility?

IIRC:  std::variant assigns a unique integer to each type, and this integer is defined by its memory location at load (at least in ABSL this was true).   Theoretically, this means we can add a type without breaking source compatibility but AFAIK this would definitely break ABI compatibility?

Given new ABI requirements from the spec, we may need to comment on that in the README.

api/test/nostd/BUILD
```diff
@@ -16,6 +16,7 @@ cc_test(
     ],
     deps = [
         "//api",
+        "@com_github_google_benchmark//:benchmark",
```
jsuereth, line 19: For here and other benchmarks, can we keep the benchmarks in separate CC files, for the bazel build?

api/test/nostd/shared_ptr_test.cc
```diff
@@ -179,3 +180,26 @@ TEST(SharedPtrTest, Sort)
...
+}
+
+static void SharedPtrTestSort(benchmark::State &state)
+{
+  for (auto _ : state)
+  {
+    SharedPtrTest_Sort(10000);
+  }
+}
+BENCHMARK(SharedPtrTestSort);
```
jsuereth, line 196: Can you move this into its own benchamrk CC file?


In bazel there is a "smoketest" for benchmarks which will run them as a test, but if we can keep them separated via the benchmark bazel rule, we'll be able to automatically generate performance reports on every CL going forward (something our team is working on).

api/test/nostd/span_test.cc
```diff
@@ -1,5 +1,8 @@
 #include "opentelemetry/nostd/span.h"
 
+#include <benchmark/benchmark.h>
```
jsuereth, line 3: some request for benchmark code.

api/test/nostd/string_view_test.cc
```diff
@@ -2,7 +2,10 @@
 
 #include <gtest/gtest.h>
 
-#include "map"
+#include <benchmark/benchmark.h>
```
jsuereth, line 5: and again here.

docs/building-with-stdlib.md
```diff
@@ -0,0 +1,192 @@
...
+
+Standard Library build flavor works best for statically linking the SDK in a
+process (environment where ABI compat is not a requirement), or for
+"header-only" implementation of SDK.
+
+Proposed approach cannot be employed for shared libs in environments where
+ABI compatibility is required:
+
+- OpenTelemetry SDK binary compiled with compiler A + STL B
+- may not be ABI compatible with the main executable compiled with compiler
```
jsuereth, line 11: I think we have to say "will not" here vs. "may not".

docs/building-with-stdlib.md
```diff
@@ -0,0 +1,192 @@
...
+  C++17 works but with additional dependencies, e.g. either MS-GSL or Abseil
+  for `std::span` implementation (`gsl::span` or `absl::Span`).
+- `absl`  - TODO: this should allow using Abseil C++ library only (no MS-GSL).
+
+Currently only `nostd` and `stdlib` configurations are implemented.
+`absl` is reserved for future use.
+
+### Build matrix
+
+List of compilers that support building with standard library classes:
```
jsuereth, line 150: We may also want to note that (for now) this is cmake-only.

jsuereth (changes requested):
Would like to cleanup all the benchmark code, but otherwise looking forward to pushing this to the next level!

api/include/opentelemetry/nostd/span.h
```diff
@@ -1,14 +1,31 @@
...
 
-#include <array>
-#include <cassert>
-#include <cstddef>
-#include <exception>
-#include <iterator>
-#include <type_traits>
+#ifdef HAVE_CPP_STDLIB
+#  include "opentelemetry/std/span.h"
+#else
```
lalitb, line 19: Should there be check for`HAVE_ABSEIL` macro too, and include `absl/span.h` if set.

api/include/opentelemetry/nostd/string_view.h
```diff
@@ -1,13 +1,30 @@
...
 
-#include <algorithm>
-#include <cstddef>
-#include <cstring>
-#include <ostream>
-#include <stdexcept>
-#include <string>
+#ifdef HAVE_CPP_STDLIB
+#  include "opentelemetry/std/string_view.h"
+#else
```
lalitb, line 19: Check for HAVE_ABSEIL ?

api/test/nostd/shared_ptr_test.cc
maxgolov: Sure. Will do.

api/test/nostd/BUILD
maxgolov: Yes, I will do it.

docs/abi-compatibility.md
```diff
@@ -0,0 +1,35 @@
+# Windows
+
+## Visual Studio 2015, 2017, 2019
+
+The Microsoft C++ (MSVC) compiler toolsets in Visual Studio 2013 and earlier don't guarantee binary compatibility across versions. You can't link object files, static libraries, dynamic libraries, and executables built by different versions. The ABIs, object formats, and runtime libraries are incompatible.
+
+We've changed this behavior in Visual Studio 2015, 2017, and 2019. The runtime libraries and apps compiled by any of these versions of the compiler are binary-compatible. It's reflected in the C++ toolset major number, which is 14 for all three versions. (The toolset version is v140 for Visual Studio 2015, v141 for 2017, and v142 for 2019). Say you have third-party libraries built by Visual Studio 2015. You can still use them in an application built by Visual Studio 2017 or 2019. There's no need to recompile with a matching toolset. The latest version of the Microsoft Visual C++ Redistributable package (the Redistributable) works for all of them.
```
lalitb, line 7:
```suggestion
This behavior is changed in Visual Studio 2015, 2017, and 2019. The runtime libraries and apps compiled by any of these versions of the compiler are binary-compatible. It's reflected in the C++ toolset major number, which is 14 for all three versions. (The toolset version is v140 for Visual Studio 2015, v141 for 2017, and v142 for 2019). Say you have third-party libraries built by Visual Studio 2015. You can still use them in an application built by Visual Studio 2017 or 2019. There's no need to recompile with a matching toolset. The latest version of the Microsoft Visual C++ Redistributable package (the Redistributable) works for all of them.
```

api/include/opentelemetry/common/attribute_value.h
maxgolov: It's a good question. My hope is that a change like this does not create ABI compat issue. I can verify this.

For two types I'm adding:

- adding `const char *` support is a non-breaking change that follows spec, that we should add before `v1.0.0`. Even protobuf / gRPC code has construct that respects string input from either `string` or `string_view` or `const char *`. I hope we can just remove the "feature gate" from that soon, in a follow-up PR.

- adding byte buffers (not spec'd, but I filed a request for spec) - I anticipate that byte buffers would eventually be added to spec. And, for sure, we will needed this for our customer our flows in Azure.

I'd like to keep some flexibility here, for the case where we statically compile the SDK into product that would need both `const char *` and `uint8_t[]` support. Feature flags are off by default, and are not affecting the mainline OpenTelemetry builds. Unfortunately there is no better way to add types to SDK at this point.

docs/building-with-stdlib.md
maxgolov: I'll change to `will not` .. It's more stable now for vs2015-vs2017-vs2019, as at least those 5 compilers have a promise of ABI compatibility. So on Windows it's more `may not` and on Linux between gcc-4.x and gcc-9x+ it's `will not`.

docs/building-with-stdlib.md
maxgolov: Abstracting away from concrete build system (CMake or Bazel), it's more like this feature is behind a `feature gate` or `feature flag` controlled by `#define WITH_STL` that is off by default. I can add that to the notes.

lalitb (approved):
LGTM overall. Somewhat large PR, but thanks for documenting the changes, and clarifying ABI related issues we will have with stl.

api/include/opentelemetry/common/attribute_value.h
maxgolov: I'll keep it under `#ifdef` - NOT enabled by default, and log a separate issue to investigate if adding new types to variant would break ABI compatibility.

maxgolov: https://github.com/open-telemetry/opentelemetry-cpp/issues/474

api/include/opentelemetry/common/attribute_value.h
```diff
@@ -10,20 +10,30 @@
...
+                   // TODO: add C-string as possible value on API surface
+                   const char *,
+#endif
+#ifdef HAVE_SPAN_BYTE
+                   // TODO: 8-bit byte arrays / binary blobs are not part of OT spec yet!
+                   // Ref: https://github.com/open-telemetry/opentelemetry-specification/issues/780
+                   nostd::span<const uint8_t>,
```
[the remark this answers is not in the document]
maxgolov: https://github.com/open-telemetry/opentelemetry-cpp/issues/474

api/include/opentelemetry/nostd/span.h
maxgolov: I was unconditionally using either `std::span` or `gsl::span` since GSL span implementation is guarantee to be 100% conformant, whereas `absl::Span` is not guarnateed to be conformant. I added a brief description to the markdown document.... Basically for now I'd use variant from `absl::variant` , because there is no other option that would compile with Visual Studio 2015. But not using absl for Span. I can see if we can use that in a separate PR.

api/include/opentelemetry/nostd/string_view.h
maxgolov: Not yet. I'm not adding full `HAVE_ABSEIL` support, planning to only use that for `absl::variant` with Visual Studio 2015. 

In all other build systems we prefer the following - in the following order:
- standard classes in C++17
- `gsl::span` for C++11, C++14, C++17 (as GSL is the only fully conforming implementation)
- `absl::variant` for C++11 (vs2015, because Linux and Mac are fine without it already with MPark variant).. and for C++14. Then C++17 already has `std::variant`.

api/test/nostd/BUILD
maxgolov: Marking as Resolved. Moving Benchmarks into separate PR.

docs/abi-compatibility.md
maxgolov: I'll keep the original cause I stole it verbatim letter-by-letter from Visual Studio documentation 😄

jsuereth (approved):

api/include/opentelemetry/nostd/span.h
maxgolov: @lalitb - this is described in the markdown doc here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/maxgolov/nostd_stl/docs/building-with-stdlib.md#note-on-gslspan-vs-abslspan

Issue #326, opened 2020-09-15
ETW exporter


maxgolov:
This project is created in response to Community Bridge Q3-Q4. We are looking for interns to work on this project.

https://github.com/open-telemetry/community/issues/468

**Project**

Add exporter for **ETW (Event Tracing for Windows)** as one of the optional recommended exporters that languages may decide to implement as part of SDK.

**Logical Reasoning**

Application logs, service logs, even Docker-on-Windows logs can be directed to ETW (ref. https://docs.docker.com/config/containers/logging/etwlogs/ ). Many products running on Windows (most Microsoft products, Intel products, Docker, etc.) - rely on ETW channel.

Adding support for ETW for Windows platforms in SDK would allow any **cross-platform codebase** to be instrumented with OpenTelemetry instead of using Win32 API directly, e.g. replacing Docker etwlogs driver with structured ETW traces emitted using OpenTelemetry C++ SDK. Final product demo could be the Docker using OpenTelemetry C++ SDK for traces, and Docker on Windows - using OpenTelemetry C++ SDK with ETW exporter.

**Extra bonus points**

Two 'extra' pieces that may be covered in addition to the main exporter piece:

- writing a projection layer for other languages (e.g. Java or node on Windows), that allows to dispatch traces from OT Java to OT C++ SDK with ETW exporter. That flow would allow Java and node services / servers running on Windows hosts to route their telemetry to ETW listener agent.

- separate effort that implements ETW Provider listener (Agent), that subsequently uses OpenTelemetry .NET SDK to dispatch all telemetry collected to OTLP exporter (or any other exporter supported by .NET SDK). Very nice and simple example how to implement such a listener: https://github.com/fireeye/SilkETW

**Two alternate implementations of the ETW message packing**

- First approach is to utilize MsgPack-encoded event payload. This approach allows to send structured objects / nested objects over ETW to the Agent. This approach has no external dependencies, but is incompatible with some existing ETW tracing tools. Work on this track can be started right away.

- Second approach is to use standard ETW "dynamic manifest" events. This approach allows to express an event as key-value property bag, flat list. I am working with MSFT CELA to contribute Microsoft TraceLoggingDynamic C++ implementation in OSS. Main benefit of this approach is it will be compatible with already existing ETW tracing tools (such as Windows Performance Analyzer, xperf, etc.). ETA for OSS is by the end of this year.

shivam-51:
[...]

PR #376, opened 2020-10-27
ETW Tracer exporter

mishal23:
The PR is the initial commit for the tracer exporter for ETW from the example written by @maxgolov. Issue #326 

This PR adds:
- [x] Headers for the ETW Tracer exporter
- [x] Unit Tests for the classes

Edit 1 (25/11/2020):

code has been updated as per the specification.

-- cc @maxgolov


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "third_party/prometheus-cpp"]
 	path = third_party/prometheus-cpp
 	url = https://github.com/jupp0r/prometheus-cpp.git
+[submodule "third_party/benchmark"]
```
maxgolov, line 4: You should be able to remove that, unless you are adding Benchmark tests here.

api/include/opentelemetry/common/key_value_iterable.h
```diff
@@ -19,16 +19,31 @@ class KeyValueIterable
    * Iterate over key-value pairs
    * @param callback a callback to invoke for each key-value. If the callback returns false,
    * the iteration is aborted.
-   * @return true if every key-value pair was iterated over
+   * @return true if every key-value pair was iterated ove
```
maxgolov, line 22: nit: over

api/include/opentelemetry/common/key_value_iterable.h
```diff
@@ -19,16 +19,31 @@ class KeyValueIterable
...
    */
-  virtual bool ForEachKeyValue(
-      nostd::function_ref<bool(nostd::string_view, common::AttributeValue)> callback) const
-      noexcept = 0;
+  virtual bool ForEachKeyValue(nostd::function_ref<bool(nostd::string_view, common::AttributeValue)>
```
maxgolov, line 24: Is there a reason this is re-formatted? You should be able to avoid this change.

api/include/opentelemetry/common/key_value_iterable.h
```diff
@@ -19,16 +19,31 @@ class KeyValueIterable
...
+                                   callback) const noexcept = 0;
 
   /**
    * @return the number of key-value pairs
    */
   virtual size_t size() const noexcept = 0;
 };
+
```
maxgolov, line 32: There should be no need to add that. I'll add in a separate commit. Not needed for ETW exporter.

api/include/opentelemetry/trace/span.h
```diff
@@ -167,5 +169,14 @@ class Span
   // AddEvent).
   virtual bool IsRecording() const noexcept = 0;
 };
+
```
maxgolov, line 172: Same here: this is internal detail. You probably don't need to introduce it in your commit.

api/include/opentelemetry/trace/span_context.h
```diff
@@ -33,6 +33,10 @@ namespace trace_api = opentelemetry::trace;
 class SpanContext final
 {
 public:
+  // An invalid SpanContext.
```
maxgolov, line 36: Where do we require this?

api/include/opentelemetry/trace/tracer.h
```diff
@@ -43,11 +43,13 @@ class Tracer
 
   template <class T,
             nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
+  // template <class T, nostd::enable_if_t<detail::is_key_value_iterable<T>::value> * = nullptr>
```
maxgolov, line 46: delete this comment

api/include/opentelemetry/trace/tracer.h
```diff
@@ -43,11 +43,13 @@ class Tracer
...
   nostd::shared_ptr<Span> StartSpan(nostd::string_view name,
                                     const T &attributes,
                                     const StartSpanOptions &options = {}) noexcept
   {
     return this->StartSpan(name, common::KeyValueIterableView<T>(attributes), options);
+    // return this->StartSpan(name, KeyValueIterableView<T>(attributes), options);
```
maxgolov, line 52:
```suggestion
```

api/include/opentelemetry/trace/tracer.h
maxgolov, line 46:
```suggestion
```

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,644 @@
+// Copyright 2020, OpenTelemetry Authors
```
maxgolov, line 1: nit: naming. Since it's in the `exporters` directory, it implies all code under is related to exporter.. So should we name it `etw_provider_exporter.h` or simply `etw_provider.h` ? I think a shorter name would be nicer.

exporters/etw/include/opentelemetry/exporters/etw/utils.h
```diff
@@ -0,0 +1,227 @@
...
+#  pragma comment(lib, "Advapi32.lib")
+#  pragma comment(lib, "Rpcrt4.lib")
+#endif
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+namespace utils
+{
+
+#if 0
```
maxgolov, line 41: You may remove this. It's not used.

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
```diff
@@ -0,0 +1,409 @@
+#pragma once
```
maxgolov, line 1: Add copyright here.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
mishal23: I added the `exporter` suffix because it was there for other exporter files. Can change it to `etw_provider.h`

.gitmodules
mishal23: removed

api/include/opentelemetry/common/key_value_iterable.h
mishal23: fixed

api/include/opentelemetry/common/key_value_iterable.h
mishal23: I used the `./tools/format.sh` with the latest version (clang-10) and that caused this formatting -- that's the reason the CI for formatter is still failing. I've fixed this one. I'll see what could be done to fix the formatting issues at other places.

api/include/opentelemetry/trace/span_context.h
mishal23: we need for the `StartSpan` function in `etw_tracer_exporter.h`, line 115
[link](https://github.com/open-telemetry/opentelemetry-cpp/blob/4df4ce60ff3e274e9959b1ae118ca999f3b8a25b/exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h#L115)

api/include/opentelemetry/trace/tracer.h
mishal23: deleted

api/include/opentelemetry/trace/tracer.h
mishal23: fixed

api/include/opentelemetry/trace/tracer.h
mishal23: fixed

exporters/etw/include/opentelemetry/exporters/etw/utils.h
mishal23: removed

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
mishal23: added

api/include/opentelemetry/common/key_value_iterable.h
mishal23: I got access to a Ubuntu machine, and have fixed the formatter at all places -- the formatter for CI is passing now

api/include/opentelemetry/trace/span_context.h
maxgolov: Please double-check how it's done in the `ostream` exporter. Why didn't they need it?

api/include/opentelemetry/trace/tracer.h
```diff
@@ -114,6 +114,12 @@ class Tracer
         static_cast<uint64_t>(std::chrono::duration_cast<std::chrono::microseconds>(timeout)));
   }
 
+  void Close() noexcept
```
maxgolov, line 117: Let's remove it for now.

exporters/etw/BUILD
```diff
@@ -0,0 +1,52 @@
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
```
maxgolov, line 3: Modify the Bazel build to include this only on Windows.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,664 @@
...
+      default:
+        // This is user error. Invalid event name!
+        // We supply default 'NoName' event name in this case.
+        break;
+    }
+
+    /* clang-format off */
+    nlohmann::json jObj =
+    {
+      { "env_name", "Span" },
```
maxgolov, line 256: Let's discuss what attribute names we use here, because `env_name` and `env_ver` were originally from another implementation that may use some Microsoft-specific schema.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,664 @@
...
+      { "env_ver", "4.0" },
+
+      // TODO: remove these fields. Provided for illustrative purposes only.
```
maxgolov, line 259: Remove these fields.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,664 @@
...
+      // TODO: should we consider uint64_t format with Unix timestamps for ELK stack?
+      { "env_time",
+        {
+          { "TypeCode", 255 },
+          { "Body","0xFFFFFC60000000005F752C2C" }
+        }
+      },
+      //
+
+      // TODO: follow JSON implementation of OTLP or place protobuf for non-Microsoft flows
```
maxgolov, line 275: Here we need to understand what are the attribute names for `common` fields:
- name
- time (as in when `event` got logged)
- traceId
- spanId
- (span?) start time
- (span?) end time

Introduce a set of macros that would define the attribute names. Use these macros, then we can rename these. The other alternative is to use `const char *` - set of pooled constants in some namespace, e.g.
```cpp
    nlohmann::json jObj =
    {
      { OPENTELEMETRY_SPAN_NAME , "Span" },
...
```

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,664 @@
...
+          jObj[name] = temp;
+          break;
+        }
+        case common::AttributeType::TYPE_STRING:
+        {
+          auto temp  = nostd::get<nostd::string_view>(value);
+          jObj[name] = temp;
+          break;
+        }
+        case common::AttributeType::TYPE_CSTRING:
```
maxgolov, line 343: Watch out for this one - you may need to comment that line, OR wait for my other PR. Currently the `CSTRING` (`const char *`) is NOT exactly supported by other exporters.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,406 @@
...
+  /// <summary>
+  /// Add event data to span associated with tracer
+  /// </summary>
+  /// <param name="span"></param>
+  /// <param name="name"></param>
+  /// <param name="timestamp"></param>
+  /// <returns></returns>
+  void AddEvent(Span &span, nostd::string_view name, core::SystemTimestamp timestamp) noexcept
+  {
+    AddEvent(span, name, timestamp, common::NullKeyValueIterable());
```
maxgolov, line 194: Check how other exporters are handling it without `NullKeyValueIterable` (something like `GetEmptyAttributes` ?)

exporters/etw/test/etw_tracer_test.cc
```diff
@@ -0,0 +1,42 @@
...
+  auto tracer = tp.GetTracer(providerName);
+  auto span   = tracer->StartSpan("MySpan");
+
+  ETWEvent event = {
+      {"uint32Key", (uint32_t)123456}, {"uint64Key", (uint64_t)123456}, {"strKey", "someValue"}};
+
+  EXPECT_NO_THROW(span->AddEvent(eventName, event));
+  EXPECT_NO_THROW(span->End());
+  EXPECT_NO_THROW(tracer->Close());
+}
```
maxgolov, line 42: nit: add EOL here

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: No, that's ok. We can change later.

ci/do_ci.sh
```diff
@@ -112,30 +112,30 @@ EOF
   examples/plugin/load/load_plugin_example ${PLUGIN_DIR}/libexample_plugin.so /dev/null
   exit 0
 elif [[ "$1" == "bazel.test" ]]; then
-  bazel build $BAZEL_OPTIONS //...
-  bazel test $BAZEL_TEST_OPTIONS //...
+  bazel build $BAZEL_OPTIONS -- //... -//exporters/etw/...
```
maxgolov, line 115: @jsuereth - Josh, we need your help on this one. While manually excluding works, I don't think this is the proper way of doing it. Is there a better option to exclude some piece of functionality from Bazel build for POSIX, or - only include if building on Windows? something like `#if WIN32` or `if WIN32` ?

exporters/etw/BUILD
maxgolov: @mishal23 - something along the lines described here:
https://blog.bazel.build/2019/02/11/configurable-builds-part-1.html

exporters/etw/src/etw_provider_exporter.cc
```diff
@@ -0,0 +1,3 @@
+#define HAVE_NO_TLD
```
maxgolov, line 1: You can also pull these files into Bazel build, but have a check on very top and bottom as

```cpp
#ifdef _WIN32
...
#endif
```

Of course this is not as good as configuring it at Bazel-build system level, but then you'd have one configuration across all OS: you pull in the file at all times, but it compiles into an empty object.

exporters/etw/src/etw_tracer_exporter.cc
```diff
@@ -0,0 +1,3 @@
+#define HAVE_NO_TLD
```
maxgolov, line 1: Same `#ifdef _WIN32` here around the entire file.

exporters/etw/test/etw_provider_test.cc
```diff
@@ -0,0 +1,73 @@
+/*
```
maxgolov, line 1: And here - same `#ifdef _WIN32` here around the entire file.

Then your code does not have to depend on whether the build system pulls it in or not. The build system can always add it, but it'll be ignored - compiled into an empty object, not linked into final set of tests.

exporters/etw/test/etw_tracer_test.cc
maxgolov, line 42: For prometheus below (I can't add comment on submodule) - I am not sure you have to update the submodule in your commit. It might be safe, but submodule update is unrelated to your change.

maxgolov:
@mishal23 - check the `Export` method in stdout exporter, see how this code needs to be adjusted to benefit from that. Otherwise it seems we are still re-implementing the entire SDK? It's probably good - we do not have unnecessary global lock in the exporter. But it doesn't fit in the definition of `exporter` per se, as it does not follow the existing established exporter API.

maxgolov:
I'm suggesting to address the Bazel build question, possibly having `#ifdef _WIN32` within the files themselves, that way the Bazel build configs don't have to change. We may integrate this. But it's not the final shape / form of it, as it's more like a total reimplementation of the Trace Provider rather than exporter (still, should work though...).

mishal23:
> @mishal23 - check the `Export` method in stdout exporter, see how this code needs to be adjusted to benefit from that. Otherwise it seems we are still re-implementing the entire SDK? It's probably good - we do not have unnecessary global lock in the exporter. But it doesn't fit in the definition of `exporter` per se, as it does not follow the existing established exporter API.

yes @maxgolov , I'm taking a look at it, will update as soon as possible

ci/do_ci.sh
mishal23: I found this way to skip incompatible targets: [https://docs.bazel.build/versions/master/platforms.html#skipping-incompatible-targets](link)

But the CI gives an error saying ` no such attribute 'target_compatible_with' in 'cc_library' rule`, need to investigate properly.

mishal23: seems like, it's not yet part of the official release -- hence the error

exporters/etw/BUILD
maxgolov: We briefly discussed this with @jsuereth that you can for now keep this excluded the way you are excluding it with "-"

jsuereth: Please open a bug/ticket to clean up the Bazel build to remove the need for this though.

api/include/opentelemetry/common/key_value_iterable.h
mishal23: removed

api/include/opentelemetry/trace/tracer.h
mishal23: removed

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
mishal23: done

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
mishal23: Yes, replaced with GetEmptyAttributes 👍

exporters/etw/test/etw_tracer_test.cc
mishal23: done

mishal23: yes, removed

api/include/opentelemetry/trace/span_context.h
mishal23: Have replaced with `trace::SpanContext::GetInvalid()`, an existing method.

maxgolov:
@mishal23 - please reformat with latest cmake formatter:
```console
    pip3 install cmake_format==0.6.13
    ./ci/do_ci.sh format
```
Then push reformatted files.

maxgolov (approved):

mishal23:
> @mishal23 - please reformat with latest cmake formatter:
> 
> ```
>     pip3 install cmake_format==0.6.13
>     ./ci/do_ci.sh format
> ```
> 
> Then push reformatted files.

Done!

maxgolov:
@mishal23 - please update the branch.

alolita:
Triaged Dec 7 2020: @jsuereth @ThomsonTan can you please review? @maxgolov's intern has contributed this PR.

mishal23:
> @mishal23 - please update the branch.

done

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+      { "env_time",
+        {
+          { "TypeCode", 255 },
+          { "Body","0xFFFFFC60000000005F752C2C" }
+        }
+      },
+      //
+
+      // TODO: follow JSON implementation of OTLP or place protobuf for non-Microsoft flows
+      { "env_dt_traceId", "6dcdae7b9b0c7643967d74ee54056178" },
```
maxgolov, line 270: @mishal23 - please log a separate issue to properly populate this. Currently it's hardcoded at some bogus value, but it has to be captured from the originating span data.

sdk/include/opentelemetry/sdk/trace/etw_data.h
```diff
@@ -0,0 +1,530 @@
...
+#pragma warning(disable : 4307) /* Integral constant overflow - OK while computing hash */
+    auto h = utils::hashCode(args.data());
+    switch (h)
+    {
+      case CONST_HASHCODE(MSGPACK):
+        // nobrk
+      case CONST_HASHCODE(MsgPack):
+        // nobrk
+      case CONST_HASHCODE(MessagePack):
+        evtFmt = ETWProvider::EventFormat::ETW_MSGPACK;
```
lalitb, line 372: how does a pseudo-random hash-value determines which event format to use ? Perhaps I am missing something ?

sdk/include/opentelemetry/sdk/trace/etw_data.h
```diff
@@ -0,0 +1,530 @@
...
+  }
+};
+}  // namespace ETW
+
+namespace sdk
+{
+namespace trace
+{
+
+class ETWSpanData final : public Recordable
```
lalitb, line 402: Good to move this in separate class under `exporters/etw/` directory.

sdk/include/opentelemetry/sdk/trace/etw_data.h
```diff
@@ -0,0 +1,530 @@
+#pragma once
```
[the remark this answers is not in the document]
maxgolov: Agree with moving out of trace _going forward in a separate PR_ . But there is a reason why it is in there: this PR was opened before Logging API was added. Eventually we need to refactor it into common.

Disagree with `-contrib` : this is official Linux Foundation -sponsored activity. `ETW` on Windows is akin to `syslog` on Linux : `standard platform facility`. It would be odd if OpenTelemetry does not support the `standard platform facility` as its output. As a reference, I'd like to mention the Docker project, which supports the following destinations as 1st class citizen destinations --- and ETW is in there as well :)

Driver | Description
-- | --
none | No logs are available for the container and docker logs does not return any output.
local | Logs are stored in a custom format designed for minimal overhead.
json-file | The logs are formatted as JSON. The default logging driver for Docker.
syslog | Writes logging messages to the syslog facility. The syslog daemon must be running on the host machine.
journald | Writes log messages to journald. The journald daemon must be running on the host machine.
gelf | Writes log messages to a Graylog Extended Log Format (GELF) endpoint such as Graylog or Logstash.
fluentd | Writes log messages to fluentd (forward input). The fluentd daemon must be running on the host machine.
awslogs | Writes log messages to Amazon CloudWatch Logs.
splunk | Writes log messages to splunk using the HTTP Event Collector.
etwlogs | Writes log messages as Event Tracing for Windows (ETW) events. Only available on Windows platforms.
gcplogs | Writes log messages to Google Cloud Platform (GCP) Logging.
logentries | Writes log messages to Rapid7 Logentries.



Right now we binary-pack the output in MsgPack format. But eventually we will add an additional option to allow for the output of "dynamic manifest" ETW, that is fully compatible with all existing ETW trace tooling, such as Windows Performance Analyzer / Recorder, xperf. etc.

The listener for the MsgPack format will be added to `contrib`, as the listener is just a reference example how to monitor for incoming ETW events.

sdk/include/opentelemetry/sdk/trace/etw_data.h
maxgolov: It's not random.

sdk/include/opentelemetry/sdk/trace/etw_data.h
maxgolov: Agreed.

ci/do_ci.sh
ThomsonTan: Is it possible to exclude running on non-pox environment in the bazel build file under //exporters/etw/...? In case the bazel build is invoked via the `do_ci.sh` script.

maxgolov:
@mishal23 - there are two ways to fix the build failure you are hitting:
- define `NOMINMAX` in the build system
- add parentheses around the `max` method

Both solutions described here: https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max

mishal23:
> @mishal23 - there are two ways to fix the build failure you are hitting:
> 
> * define `NOMINMAX` in the build system
> * add parentheses around the `max` method
> 
> Both solutions described here: https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max

done, thanks @maxgolov for the input!

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: Tracking it https://github.com/open-telemetry/opentelemetry-cpp/issues/466 - which **I think we should solve in a separate PR**.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: https://github.com/open-telemetry/opentelemetry-cpp/issues/466

ci/do_ci.sh
maxgolov: @ThomsonTan - I think we discussed this issue, how to exclude - and appears like we converge on using "-" for now. It's been discussed in the comment above: https://github.com/open-telemetry/opentelemetry-cpp/pull/376#discussion_r523107518

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+
+#ifndef WIN32_LEAN_AND_MEAN
+#  define WIN32_LEAN_AND_MEAN
+#endif
+
+#ifdef _MSC_VER
+// evntprov.h(838) : warning C4459 : declaration of 'Version' hides global declaration
+#  pragma warning(disable : 4459)
+// needed for Unit Testing with krabs.hpp
+#  pragma warning(disable : 4018)
```
ThomsonTan, line 25: Moving this `pragma` to Unit Testing header file before including this file?

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+    }
+    return false;
+  }
+
+  /// <summary>
+  /// Get Provider by Name or string representation of GUID
+  /// </summary>
+  /// <param name="providerId"></param>
+  /// <returns></returns>
+  Handle &open(const std::string &providerId, EventFormat format = EventFormat::ETW_MANIFEST)
```
ThomsonTan, line 110: Moving these non-trivial functions to .cc file?

sdk/include/opentelemetry/sdk/trace/processor.h
```diff
@@ -47,7 +47,7 @@ class SpanProcessor
    * timeout is applied.
    */
   virtual bool ForceFlush(
-      std::chrono::microseconds timeout = std::chrono::microseconds::max()) noexcept = 0;
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept = 0;
```
ThomsonTan, line 50: What is this change for?

maxgolov: This is to resolve the `NOMINMAX` compilation issue. Visual Studio and Windows SDK is notoriously bad for defining a macro for `max` like `#define max` - which breaks any C++ Standard Library methods that use `max` method 😄

maxgolov: https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max

Cit. `If you're really desperate, put parentheses around the function names...`, - we were desperate 😁

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: I'd really like to keep it header-only. Our (known) customer for this prefers the entire thing header-only.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: We can remove line 25, but we still need the pragma at 23. Best to handle this with pragma and pragma push-pop here rather than outside.

ci/do_ci.sh
jsuereth: Still waiting for 3.8 release of BAZEL before we do anything fancy here.  The pending release has a decent experience to limit building non-compatible things (the `target_compatible_with` feature), otherwise we'll have wonky build flags all over.

TL;DR: for Bazel
- If you're adding new TARGETs (e.g. an exporter `cc_library`), the next major bazel release makes this easy.
- If you're adding different SOURCE on existing target (`cc_library` that has different `-D` or source files per platofrm) we can do this (easily) today.


I suggest we do one of the following:

1. We don't write build rules for ETW in bazel right now (unless you need them).   Open a ticket to add BAZEL support for the ETW exporter and we can block it on the 3.8 release of bazel.  This has the *least* impact for folks like myself who use bazel for development and want `bazel test //...` to work.
2. We use bazel's switch statement to make the ETW export compile a no-op plugin on non-windows platforms.  (i.e. it selects a source file that has nothing in it.  We can open a bug to rework this when bazel 3.8 hits.
3. You do what you've done here.  I think this impacts bazel users dramatically because `test //...` is broken on non-windows environments.  I'd prefer either (1) or (2).

exporters/etw/BUILD
jsuereth: Sorry, I changed my mind here given the impact of day-to-day development using Bazel on non-windows.  See comment above.

jsuereth (changes requested):
Thinking about this more, I think we can just drop the BAZEL build for now.

1. ETW would be buildable in cmake and *does* have test coverage.
2. We can wait for the "nice" BAZEL features around platforms rather than do some of the wonky workarounds available today.

mishal23:
Hi all, thanks for reviewing the PR, I'll update the PR with necessary changes by tomorrow.

ci/do_ci.sh
maxgolov: Can we just do `#ifdef _WIN32` , in all relevant ETW files. And unconditionally build. Then the `.cc` files are gonna produce 0-byte objects, that could safely be ignored?

maxgolov:
Please consider `#ifdef _WIN32 ... #endif` in all ETW modules. This way you may unconditionally build it (or unconditionally NOT build it, in fact 😄 ) across all possible platforms. We can also consider adding a feature flag for that. But I'd like to add the feature gate / flat in a separate PR..... maybe `WITH_ETW` ?

mishal23:
> Please consider `#ifdef _WIN32 ... #endif` in all ETW modules. This way you may conditionally build it (or, rather, unconditionally NOT build it, in fact 😄 ) across all possible platforms: it's not Bazel, but compiler / preprocessor gonna exclude it from the build. So you would not need `"-"` in Bazel to exclude it. We can also consider adding a feature flag for that. But I'd like to add the feature gate / flat in a separate PR..... maybe `WITH_ETW` ?

For CMake, we already have the feature flag: `WITH_ETW_EXPORTER`, hence we are not needing to do anything more.
```cmake
if(WIN32)
  option(WITH_ETW_EXPORTER "Whether to include the ETW Exporter in the SDK" ON)
endif(WIN32)
```
 we need to look for Bazel, I can do the `#ifdef _WIN32 ... #endif` wrapping around all modules for now.

sdk/include/opentelemetry/sdk/trace/etw_data.h
mishal23: moved!

ci/do_ci.sh
mishal23: wrapped with `ifdef _WIN32` and removed all the wonky workarounds in `ci/do_ci.sh`, it's all clean now. thanks, @jsuereth @maxgolov for your inputs on this!

maxgolov:
@jsuereth - Mishal added `#ifdef _WIN32` around ETW, so there is no need to exclude it on other platforms. It simply won't produce any objects out of any `.cc` . Let us know if it is acceptable. We are discussing if we should migrate to a different place under `contrib`. Somewhat related to both Elastic, ETW, and other exporters - need to sort out how Vendors would include their additional exporters in Bazel. The story is okay with CMake build options. But not clear how this should be handled for Bazel.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: Let's solve this in https://github.com/open-telemetry/opentelemetry-cpp/issues/466

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+    v.push_back(0);
+    v.push_back(0);
+    if (v.size() % 2)
+    {
+      v.push_back(0);
+    };
+    void *buff = v.data();
+
+    UCHAR level        = 0;  // LogAlways
+    auto writeResponse = EventWriteString(providerData.providerHandle, level, 0, (PCWSTR)buff);
```
pyohannes, line 415: Are you sure we can use `EventWriteString` here? MsgPack encoded data might well contain sequences of zeros, which might be interpreted as end of the string. Maybe `EventWrite` is a safer choice here?

ci/do_ci.sh
jsuereth: Yeah, would be good to open a ticket to clean this up in a follow on PR.

jsuereth (approved):
Approved for workaround on bazel.  Thanks!

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: It's a good question and a valid concern. We may hit this with a sequence of two zeroes: since the originating string is wide string.. I'd like to merge this, but keep that question of 0 0 as an open issue - to see what are the use-cases where MsgPack would result in a payload with two consecutive zeroes. As an alternative, the code may also use Trace Logging Dynamic logic (that would be safe). But we have not finished the open sourcing of TLD header.

maxgolov: ..basically in case if the payload ended up having two consecutive zeroes, it appears that resulting event may get truncated by receiver. I need to see if we can reuse `EventWrite` (and how). I'll handle it in a separate PR.

pyohannes: Thanks!

It's not too hard to come up with scenarios with two consecutive zero bytes (the [msgpack site](https://msgpack.org) has a nice converter to verify those). `65536` encodes to `ce 00 01 00 00`, `[0,0]` encodes to `92 00 00`, and `40000000000` encodes to `cf 00 00 00 09 50 2f 90 00`.

ThomsonTan (approved):

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: I logged a separate issue for this. #483 - side-effect is that the receiving end in this case will get a smaller buffer and would fail to decode it. `EventWrite` is the way to go. But it'd need a bit extra prep-work - to generate the proper struct that describes the event. It'd need about 20-30 lines of code patch. I'm afraid this is too big ask for Mishal, and I'll cover it in a subsequent PR.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+          auto temp = nostd::get<GUID>(value);
+          // TODO: add transform from GUID type to string?
+          jObj[name] = temp;
+          break;
+        }
+#  endif
+        // TODO: arrays are not supported yet
+#  if 0
+        // TODO: array of uint8_t is not supported by OT spec
+        case common::AttributeType::TYPE_SPAN_BYTE:
```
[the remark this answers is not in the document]
maxgolov: I'll fix it myself :)

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+          jObj[name] = temp;
+          break;
+        }
+        case common::AttributeType::TYPE_STRING:
+        {
+          auto temp  = nostd::get<nostd::string_view>(value);
+          jObj[name] = temp;
+          break;
+        }
+        case common::AttributeType::TYPE_CSTRING:
```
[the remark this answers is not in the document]
maxgolov: I'll fix it myself :)

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,657 @@
...
+    // Layer 2
+    nlohmann::json l2 = nlohmann::json::array();
+    // Layer 3
+    nlohmann::json l3 = nlohmann::json::array();
+
+    l1.push_back("Span");
+
+    {
+      // TODO: clarify why this is needed
+      // TODO: fix time here
```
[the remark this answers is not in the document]
maxgolov: For the `TypeCode` and `Body` containing some hardcoded values - it appears like a copy-paste from another exporter example (Elastic?). We'd need to remove it. I'll handle it in a separate PR.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,110 @@
...
+  {
+    for (auto &recordable : recordables)
+    {
+      auto span = std::unique_ptr<sdk::trace::ETWSpanData>(
+          dynamic_cast<sdk::trace::ETWSpanData *>(recordable.release()));
+      if (span != nullptr)
+      {
+        std::cout << span->GetName() << std::endl;
+      }
+    }
```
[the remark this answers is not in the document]
maxgolov: I'll take it over.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+    }
+#endif
+
+#ifdef HAVE_MSGPACK
+    if (encoding == ETWProvider::ETW_MSGPACK)
+    {
+      etwProvider().writeMsgPack(provHandle, m);
+      return;
+    }
+#endif
```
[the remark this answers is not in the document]
maxgolov: You are right. We should. Because `write` abstracts away either / or. I think Mishal only tested the MsgPack path because for the other implementation we did not share the `TraceLoggingDynamic.h` in OSS yet.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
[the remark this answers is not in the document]
maxgolov: Right now the `write` is done straight on `AddEvent`, and I agree we should move this into `Export`.. One issue we need to agree on if we batch the trace info and flush once as a series of attributes on a single ETW event, or if we react on `AddEvent` - pass to out-of-proc listener, which itself then may perform aggregation of all data emitted for the span.. I'd like to merge this, then explore better options.

maxgolov:
@lalitb - addressed your comment about `#ifdef` for the missing types.

For the tracer logic and `export` function not called --- let's assume that for each `AddEvent` right now the code logic simply pass-thru everything as ETW events to the out-of-proc listener "agent". Then that out-of-proc "agent" may react and invoke OpenTelemetry C# SDK, e.g. every OT `AddEvent` is sent to ETW, and may be translated into `AddEvent` in C# in the out-of-proc agent. We consolidate all span ops and data outside of C++ SDK. And we would  be still retaining the actual OpenTelemetry semantics: as if we are remoting / using C# API. Then if we don't split those API calls into individual ETW events, we need to agree on format how we package multiple `AddEvent` calls into a single span.. Also I have a minor concern that if we don't pass-thru as individual ETW in realtime, then we are losing data from unconcluded / unfinished spans.

Since it affects mostly how we batch (individual operations or pass-thru in realtime), but not the actual data being passed from one process to another, I'd like to discuss it offline before we make a change.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: This will be covered in a follow-up PR.

maxgolov:
@pyohannes - I resolved your concern re. `EventWriteString` by using `EventWrite`. I verified that `EventWrite` works and respects the zero words in the middle of string.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: Will be addressed in #466

maxgolov (approved):

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+#include "opentelemetry/exporters/etw/etw_provider_exporter.h"
+
+#include "opentelemetry/exporters/etw/utils.h"
+
+namespace core  = opentelemetry::core;
+namespace trace = opentelemetry::trace;
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+namespace ETW
```
pyohannes, line 40: Usually we have exporters under `opentelemetry::exporter` namespaces.

exporters/etw/src/etw_tracer_exporter.cc
```diff
@@ -0,0 +1,7 @@
+#ifdef _WIN32
+
+#  define HAVE_NO_TLD
+
+#  include "opentelemetry/exporters/etw/etw_tracer_exporter.h"
```
pyohannes, line 8: Is it necessary to have this in the compilation unit? Or could we go with a header-only exporter here?

Otherwise, could we just have class definitions in the headers and move the implementation here?

exporters/etw/src/etw_tracer_exporter.cc
```diff
@@ -0,0 +1,7 @@
+#ifdef _WIN32
+
+#  define HAVE_NO_TLD
```
pyohannes: Should this be defined here, when we also have it defined in a `CMakeLists.txt` file? Otherwise this will always override what comes from CMake.

pyohannes (commented):
I'm giving this a first review (sorry for chiming in so late in the game). I put some thoughts and questions inline.

Ideally we'd get rid of the (30) TODOs, but I think it's planned to do that after merging this PR, to not prolong the process further.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+  /// <returns>Span</returns>
+  virtual nostd::shared_ptr<trace::Span> StartSpan(
+      nostd::string_view name,
+      const common::KeyValueIterable &attributes,
+      const trace::SpanContextKeyValueIterable &links,
+      const trace::StartSpanOptions &options = {}) noexcept override
+  {
+    // TODO: support attributes
+    UNREFERENCED_PARAMETER(attributes);
+    UNREFERENCED_PARAMETER(links);
```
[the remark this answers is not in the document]
maxgolov: This will be a problem for doxygen: https://stackoverflow.com/questions/55421894/how-do-i-doxygen-document-unnamed-parameters-of-a-function ... Since this module is Windows-specific, I believe `UNREFERENCED_PARAMETER` would be appropriate, esp. if we want to run doxygen on that code. Support for both parameters (attributes, links) will be added in a separate PR.

exporters/etw/src/etw_tracer_exporter.cc
maxgolov: This was originally made header-only. I think the reason why it got included in source to satisfy the unit testing requirements. I don't have an answer. @mishal23 may clarify why he needed this.

exporters/etw/src/etw_tracer_exporter.cc
maxgolov: This is kept for now to keep Bazel build happy. This will be removed once we OSS the `TraceLoggingDynamic.h` header. Equivalent TLD functionality is available in Go, for example, here: https://godoc.org/github.com/microsoft/go-winio/pkg/etw .. so we will democratize the proper full support of ETW events and this define will go away.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+#  ifndef min
+#    define min(a, b) (((a) < (b)) ? (a) : (b))
+#  endif
+#endif
+
+#define MICROSOFT_EVENTTAG_NORMAL_PERSISTENCE 0x01000000
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+class ETWProvider
```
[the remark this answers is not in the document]
maxgolov: zurd

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
maxgolov: Yes.

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
```diff
@@ -0,0 +1,423 @@
...
+#include <string>
+#include <vector>
+
+#ifdef _WIN32
+#  include "Windows.h"
+#endif
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+namespace event
```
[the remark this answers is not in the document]
maxgolov: Ideally need to be refactored into utils.hpp

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,110 @@
...
+
+#include "opentelemetry/sdk/trace/exporter.h"
+#include "opentelemetry/sdk/trace/recordable.h"
+
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <unordered_map>
```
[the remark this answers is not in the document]
maxgolov: @mishal23

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+      case CONST_HASHCODE(tld):
+        // nobrk
+        evtFmt = ETWProvider::EventFormat::ETW_MANIFEST;
+        break;
+
+      default:
+        break;
+    }
+#pragma warning(pop)
+    return nostd::shared_ptr<trace::Tracer>{new (std::nothrow) Tracer(*this, name, evtFmt)};
```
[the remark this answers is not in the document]
maxgolov: `version` argument as a concept is dubious for statically linked SDK, esp. given in the other thread we are discussing that most likely we are at `v1.0` for very long time; and it will be problematic to provide ABI-stable SDK for `v1.1` , if we need to start supporting new types (e.g. byte arrays). I'd like to keep this as an open item, and actually repurpose the version - making it a mechanism to pass other config params to tracer. Then we do not have to force the tracer provider to support only one format or encoding. It may support different formats / encodings.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+  }
+
+  void SetDuration(std::chrono::nanoseconds duration) noexcept override { duration_ = duration; }
+
+  void InitTracerProvider(std::string providerName)
+  {
+    ETW::TracerProvider tracer_provider_;
+
+    tracer_ = tracer_provider_.GetTracer(providerName);
+    span_   = tracer_->StartSpan(name_);
```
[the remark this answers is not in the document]
maxgolov: I agree.

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
```diff
@@ -0,0 +1,423 @@
+// Copyright 2020, OpenTelemetry Authors
```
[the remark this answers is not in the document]
maxgolov: Have to be somewhat careful to avoid the clash with `uuid/uuid.h` and `include/linux/uuid.h` in Linux kernel though.. Okay.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,536 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+namespace ETW
+{
+class Span;
+
+/// <summary>
+/// stream::Tracer class that allows to send spans to ETW
+/// </summary>
```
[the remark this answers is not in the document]
maxgolov: By default, the stub generation in Visual Studio for C++ is set to XML Doc Comments (`///`). I agree with your suggestion, but I'd like to clean it up in a separate PR, which will have no code changes other than cleaning up the comments from `///` to `/**`.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
maxgolov: > We should probably avoid creating a new tracer for each GetTracer call, as this could get expensive...

This is actually optimized as follows:
- Tracer object holds only `encoding` and `provider handle`
- subsequently the `ETWProvider::open` method already maintains the map of registered providers and their handles. And it won't invoke the registration at Win32 API layer, but rather return a wrapper object (`Tracer`) that refers to already registered open handle.

If we are to cache the tracers additionally, not sure we'd save significant CPU time.. Maybe some memory in concurrent use scenario? But I'd argue if users allocate a shared_ptr and forget about it, they should optimize this in their code if that becomes a significant issue. I doubt it's significant, since we already optimize the opening in the `ETWProvider` class.

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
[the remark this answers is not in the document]
maxgolov: I'll fix this.

exporters/etw/include/opentelemetry/exporters/etw/uuid.hpp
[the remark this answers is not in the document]
maxgolov: This code was borrowed from my proposed implementation of `Event` with additional overloads, e.g. higher-level support for GUID, and other complex types not directly supported by OpenTelemetry. We'll refactor this into utility class.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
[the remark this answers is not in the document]
maxgolov: Cleaned it up.

exporters/etw/src/etw_tracer_exporter.cc
mishal23: yeah, not needed. I kept it with the aim to move the implementations from the header class to here, but not needed right now, since header only is expected.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,110 @@
...
+   * note: passing custom timeout values is not currently supported for this exporter
+   */
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override
+  {
+    return true;
+  };
+
+private:
+  std::string providerName_;
+  std::shared_ptr<sdk::trace::ETWSpanData> data_;
```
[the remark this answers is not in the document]
maxgolov: Will remove it.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
pyohannes: I get your point.

We follow the pattern with unnamed parameters in other places, because it's true to [ISO CPP standards](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused). I also think that in most cases implementations of virtual methods in derived classes don't necessarily need documentation.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
pyohannes: Remaining span data (trace id, span id, links, start time, and duration) must not be dropped, but need to be sent too.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
pyohannes: The `version` attribute here doesn't refer to the version of the SDK, but to the version of the instrumentation library (this means, the library that is using OpenTelemetry to add instrumentation), this is the [part of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer) describing this.

So, it would look like this:
```c++
auto tracer = provider->GetTracer("io.opentelemetry.contrib.mongodb", "4.4");
```

This would, ideally, be invoked inside MongoDB to obtain a tracer for adding instrumentation.

This would, one day, allow the SDK to enable/disable instrumentation for certain libraries and library versions based on some setting or configuration. If we abuse the `name` or `version` parameter, we'll break this.

Also, your approach will not work. Imagine a third-party library that is instrumented with OpenTelemetry and doesn't know anything about ETW (for example, imagine a `GetTracer` call in nginx). There's no way you can provide the encoding information via the `version` parameter in those instances. The current design here is broken and needs to be fixed.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
pyohannes: This needs to be refactored, at the latest when we send complete spans.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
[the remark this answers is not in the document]
pyohannes: Namespaces have to be used consistently with other parts of the SDK and with the style guide.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+{
+
+public:
+  const unsigned long STATUS_OK    = 0;
+  const unsigned long STATUS_ERROR = ULONG_MAX;
+  const unsigned long STATUS_EFBIG = ULONG_MAX - 1;
+
+  enum EventFormat
+  {
+    ETW_MANIFEST = 0,
```
pyohannes, line 71: Use enumerator names according to [the style guide](https://google.github.io/styleguide/cppguide.html#Enumerator_Names), here and in all other places.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+
+class ETWProvider
+{
+
+public:
+  const unsigned long STATUS_OK    = 0;
+  const unsigned long STATUS_ERROR = ULONG_MAX;
+  const unsigned long STATUS_EFBIG = ULONG_MAX - 1;
```
pyohannes, line 67: Use constant names according to [the style guide](https://google.github.io/styleguide/cppguide.html#Constant_Names), here and in all other places.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+    std::vector<BYTE> providerMetaVector;
+    GUID providerGuid;
+  };
+
+  /// <summary>
+  /// Check if given provider is registered.
+  /// </summary>
+  /// <param name="providerId"></param>
+  /// <returns></returns>
+  bool is_registered(const std::string &providerId)
```
pyohannes, line 91: Name functions and methods according to [they style guide](https://google.github.io/styleguide/cppguide.html#Function_Names), here and in all other places.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
```diff
@@ -0,0 +1,534 @@
...
+  virtual void ForceFlushWithMicroseconds(uint64_t) noexcept override{};
+
+  /// <summary>
+  /// Close tracer, spending up to given amount of microseconds to flush and close.
+  /// </summary>
+  /// <param name="timeout">Allow Tracer to drop data if timeout is reached</param>
+  /// <returns></returns>
+  virtual void CloseWithMicroseconds(uint64_t) noexcept override
+  {
+    etwProvider().close(provHandle);
```
pyohannes, line 131: Could you have, in one process, multiple tracers sharing the same provider handle? If so, would then closing one tracer also close the provider handles of the other tracers?

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+
+    if (!builder.End())  // Returns false if the metadata is too large.
+    {
+      return STATUS_EFBIG;  // if event is too big for UTC to handle
+    }
+
+    tld::EventDescriptor eventDescriptor;
+    // eventDescriptor.Keyword = MICROSOFT_KEYWORD_CRITICAL_DATA;
+    // eventDescriptor.Keyword = MICROSOFT_KEYWORD_TELEMETRY;
+    // eventDescriptor.Keyword = MICROSOFT_KEYWORD_MEASURES;
```
pyohannes, line 588: Is this needed?

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+
+    EVENT_DATA_DESCRIPTOR pDataDescriptors[3];
+
+    EventDataDescCreate(&pDataDescriptors[2], byteDataVector.data(),
+                        static_cast<ULONG>(byteDataVector.size()));
+
+    // Event size detection is needed
+    int64_t eventByteSize = byteDataVector.size() + byteVector.size();
+    int64_t eventKBSize   = (eventByteSize + 1024 - 1) / 1024;
+    // bool isLargeEvent     = eventKBSize >= LargeEventSizeKB;
```
pyohannes, line 598: Is this needed?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,103 @@
...
+#include "opentelemetry/trace/span_context_kv_iterable_view.h"
+#include "opentelemetry/trace/span_id.h"
+#include "opentelemetry/trace/trace_id.h"
+#include "opentelemetry/trace/tracer_provider.h"
+
+#include "opentelemetry/sdk/trace/exporter.h"
+#include "opentelemetry/sdk/trace/recordable.h"
+
+#include "opentelemetry/exporters/etw/etw_data.h"
+#include "opentelemetry/exporters/etw/etw_provider_exporter.h"
```
pyohannes, line 37: Unused headers need to be cleaned up.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer_exporter.h
```diff
@@ -0,0 +1,103 @@
...
+   */
+  sdk::trace::ExportResult Export(
+      const nostd::span<std::unique_ptr<sdk::trace::Recordable>> &recordables) noexcept override
+  {
+    for (auto &recordable : recordables)
+    {
+      auto span = std::unique_ptr<ETWSpanData>(dynamic_cast<ETWSpanData *>(recordable.release()));
+      if (span != nullptr)
+      {
+        std::cout << span->GetName() << std::endl;
```
pyohannes, line 81: We mustn't clutter `stdout` of processes.

exporters/etw/src/etw_tracer_exporter.cc
pyohannes: Compiling header-only code in compilation units should be avoided.

exporters/etw/test/etw_provider_test.cc
```diff
@@ -0,0 +1,78 @@
...
+
+#  include "opentelemetry/exporters/etw/etw_provider_exporter.h"
+
+using namespace OPENTELEMETRY_NAMESPACE;
+
+TEST(ETWProvider, ProviderIsRegisteredSuccessfully)
+{
+  std::string providerName = "OpenTelemetry";
+  static ETWProvider etw;
+  auto handle = etw.open(providerName.c_str());
```
pyohannes, line 33: I think you have to call `etw.close` for the handles you registered, to keep the state clean after tests? Otherwise those stay stored in the singleton `map` and my obfuscate results of other tests.

exporters/etw/include/opentelemetry/exporters/etw/etw_data.h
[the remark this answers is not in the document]
pyohannes: I also just realized that the tracer name is used as ETW provider name. This is broken, for the same reasons given above.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+        break;
+#  endif
+      default:
+        // This is user error. Invalid event name!
+        // We supply default 'NoName' event name in this case.
+        break;
+    }
+
+    /* clang-format off */
+    nlohmann::json jObj =
```
pyohannes, line 256: The logic for creating the JSON representation of a span should be refactored into its own method and unit tested.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
```diff
@@ -0,0 +1,658 @@
...
+
+    EventDataDescCreate(&pDataDescriptors[2], byteDataVector.data(),
+                        static_cast<ULONG>(byteDataVector.size()));
+
+    // Event size detection is needed
+    int64_t eventByteSize = byteDataVector.size() + byteVector.size();
+    int64_t eventKBSize   = (eventByteSize + 1024 - 1) / 1024;
+    // bool isLargeEvent     = eventKBSize >= LargeEventSizeKB;
+
+    // TODO: extract
```
pyohannes, line 600: All TODO items need to be fixed.

exporters/etw/test/etw_tracer_test.cc
```diff
@@ -0,0 +1,95 @@
...
+  exporter::ETW::TracerProvider tp;
+  auto tracer = tp.GetTracer(providerName);
+  auto span   = tracer->StartSpan("MySpan");
+
+  ETWEvent event = {
+      {"uint32Key", (uint32_t)123456}, {"uint64Key", (uint64_t)123456}, {"strKey", "someValue"}};
+
+  EXPECT_NO_THROW(span->AddEvent(eventName, event));
+  EXPECT_NO_THROW(span->End());
+  EXPECT_NO_THROW(tracer->CloseWithMicroseconds(0));
```
pyohannes, line 46: It would be great to have an example application illustrating the usage of this exporter, ideally with some instructions on how to receive ETW events. This would help people to use, test and verify the functionality.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
pyohannes: I think independent from the possibility for vendor specific solutions, one should also implement and provide a standard encoding. This should be documented, ideally in a README provided with the exporter.

pyohannes (approved):
Approved, as @maxgolov prefers to apply changes in different PRs. I tried to summarize open issues below in separate comments.

lalitb:
> Approved, as @maxgolov prefers to apply changes in different PRs. I tried to summarize open issues below in separate comments.

Thanks for reviewing it @pyohannes - I am merging this PR and @maxgolov agreed to fix them in separate incremental PRs

exporters/etw/test/etw_tracer_test.cc
maxgolov: I will commit the ETW listener (written in C#) to contrib repo. Maybe with example how to use this exporter as forward-channel to C#, that may itself then use OpenTelemetry C# to upload the incoming events.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider_exporter.h
maxgolov: Agreed.

Issue #375, opened 2020-10-26
CI cmake-format tools 0.6.5 vs 0.6.13 produce different output


maxgolov:
CI uses older versions of the tools from Ubuntu 18.04. However, the more recent versions of the tools in Ubuntu 20.xx produce different formatting output, thus failing the CI.

CI:

```console
# Ubuntu 18.04
clang-format version 8.0.0-3~ubuntu18.04.2 (tags/RELEASE_800/final)
Running cmake-format 0.6.5.
```

Latest Ubuntu:

```console
# Ubuntu 20.04
clang-format version 8.0.1-9 (tags/RELEASE_801/final)
Running cmake-format 0.6.13.
```

This is a bit interesting, cause one using more recent version of the distro would unlikely be willing to downgrade their version, given that the same `install_format_tools.sh` is gonna deploy the more recent versions.

Morale: `tools/format.sh` can't be trusted if you are using latest Ubuntu.

maxgolov:
I don't know how to solve this. To me the output from `cmake-format 0.6.13` looks more neat / clean, but the CI would fail any `CMakeLists.txt` formatted with latest version of the tool. Maybe we should use more recent version of the tool, and auto-commit re-formatted code back automatically without failing CI (at least for the cmake format, not C++ code format).

maxgolov:
I added the fix in my build tools PR: #377  - to upgrade to 0.6.13, plus manually reformatted all CMakeLists. They look prettier when formatted with recent CMake tooling.

maxgolov:
I'm closing this - since we updated to latest `v0.6.13` , there is a lesser chance of running into this issue. Recent Ubuntu 18.xx and 20.xx both have `v0.6.13` available, with `v0.6.13` being the current latest.

PR #377, opened 2020-10-27
Build tools: build configurations, vcpkg support, vcpkg port file for OpenTelemetry, use submodules for CMake deps

maxgolov:
# List of tools

- `build.cmd` - builds OpenTelemetry using selected Visual Studio version (2015, 2017, 2019). Could be done side-by-side on one machine, as their output directories would differ. This allows building and comparing the output across all compilers at once. I also added an option to pass custom build flags for 2017 and 2019 (that would be needed for my other PR). Section related to `-DWITH_STL`. This has no effect if CMakeLists.txt does not have a corresponding build option.

- `build-benchmark.cmd` - builds Google Test & Google Benchmark with vs2019 from pinned submodule commit id

- `build-gtest.sh` - builds Google Test without samples and tests for POSIX OS (Linux or Mac)

- `build-vs2015.cmd` `build-vs2017.cmd` `build-vs2019.cmd` - are shortcuts to setup corresponding build tools first, then build using CMake, targeting selected version. Build scripts use the default toolchain for each version:

`2015 - vc140 - C++11/14`
`2017 - vc141 - C++17`
`2019 - vc142 - C++20`

Assuming that `CMakeLists.txt` of the product contains corresponding `CMAKE_CXX_STANDARD` to decide what language standard to use. Lowest, 11, is applied if the code is compiled with vs2015 (vc140). Highest, 20, is applied if the code is compiled with vs2019 (vc142).

- plus a few other accessory scripts.

# Developer experience

When dev has only one compiler installed on their machine, the scripts would sort out what compiler to use automagically.

For example:

- `tools\build.cmd` - builds for Windows using whatever installed compiler. Preference is given to latest.

- `tools/build.sh` - builds for either Linux or Mac using whatever installed default `CC` compiler.

# Additional setup scripts

- `setup-buildtools-mac.sh` - standalone script to setup CMake + dependencies using `brew`

- `setup-buildtools.sh` - standalone script to setup Debian, Ubuntu or RedHat packages needed to build with CMake

- `setup-buildtools.cmd` - standalone script to setup all deps + vcpkg using chocolatey package manager (available by default on GitHub Actions runner). This script also integrates all deps into current Visual Studio installation.

# Submodules and Reasons why adding submodules is beneficial for CMake build system

Unfortunately package managers across different distros do not typically allow to easily pin to given dependency version. Recent example: our CI code formatting loop behaves differently on Ubuntu 18.x vs 20.xx; same comment generally applies to `vcpkg` - it is a bit harder to pin to given dependency version (need to locally cache given version of CONTROL file).

When we add submodule - we can control what upstream git repo we track, what branch, and can pin submodule to specific Commit ID. That way we shield ourselves from spontaneous changes in any of our dependencies. This would be the most predictable setup (somewhat more predictable than `vcpkg`) for our official releases.

Whereas when we do CI - it is usually easier to test with dependency version coming either from OS package manager:
- Linux distros: `rpm` `yum` `apt` `deb` 
- Cross-plat: `vcpkg`
- Mac: `brew`

Thus, I'm including both - examples how to install 'quickly' using OS-provided packager manager; versus how to build needed test deps from source, originating from git submodule.

# Note on GSL: Guidelines Support Library

The Guidelines Support Library (GSL) contains functions and types that are suggested for use by the C++ Core Guidelines maintained by the Standard C++ Foundation... The entire implementation is provided *inline in the headers* under the gsl directory. The implementation generally assumes a platform that implements C++14 support.

This could be used as *optional* dependency for header-only build of SDK using `C++14` compatible compiler, such as Visual Studio 2015. This library is MIT-licensed, not required by default by OpenTelemetry C++ SDK, and not to be used by prebuilt binary distribution of OpenTelemetry C++ SDK.  Main logical reasoning here - GSL could be used as the lightest header only way to compile the code with *STANDARD* Library, without using non-standard C++ Library provided by OpenTelemetry C++ SDK. Thus, resulting min-size build for header-only SDK; OR for SDK embedded / statically linked in products running on Windows OS.

Other OS, esp. any system with C++17 compiler - would not benefit from MS-GSL library.


ci/do_ci.sh
```diff
@@ -156,6 +156,7 @@ elif [[ "$1" == "format" ]]; then
   if [[ ! -z "$CHANGED" ]]; then
     echo "The following files have changes:"
     echo "$CHANGED"
+    git diff
```
maxgolov, line 159: It's a pain when formatting tools fail. At least this should show what changed exactly.

tools/ports/benchmark/portfile.cmake
```diff
@@ -0,0 +1,49 @@
...
+if (VCPKG_PLATFORM_TOOLSET STREQUAL "v140")
+  # set(CMAKE_C_COMPILER_WORKS 1)
+  # set(CMAKE_CXX_COMPILER_WORKS 1)
+  set(CMAKE_C_COMPILER cl.exe)
+  set(CMAKE_CXX_COMPILER cl.exe)
+  set(MSVC_TOOLSET_VERSION 140)
+  # set(VCPKG_VISUAL_STUDIO_PATH "C:\\Program Files (x86)\\Microsoft Visual Studio 14.0")
+  # set(VCPKG_PLATFORM_TOOLSET v140)
+else()
+  # Make sure vs2019 compiled binaries are compat with vs2017
```
maxgolov, line 14: This was needed to pin to specific commit ID. Also the issue in the mainline `vcpkg` that by default, if built with vs2019 - it installs the library globally and that library cannot be reused by vs2017 installation on the same machine. That is because ABI compatibility is INSUFFICIENT on Windows: the libraries must all be linked by the lowest version of the toolchain. The flag `/d2FH4-` allows to turn off vs2019-specific exception semantics and make the library link successfully with vs2017-compiled binaries, even if Google Benchmark itself is compiled with vs2019.

tools/setup-buildtools-mac.sh
```diff
@@ -24,8 +24,11 @@ sudo chown -R $(whoami) /usr/local/var/homebrew
 sudo chown -R $(whoami) /usr/local/etc/bash_completion.d /usr/local/include /usr/local/lib/pkgconfig /usr/local/share/aclocal /usr/local/share/locale /usr/local/share/zsh /usr/local/share/zsh/site-functions /usr/local/var/homebrew/locks
 
 brew install cmake
+brew install coreutils
 brew install wget
 brew install clang-format
 brew install google-benchmark
 brew tap nlohmann/json
 brew install nlohmann-json
+brew install abseil
```
maxgolov, line 33: Technically we do not need `abseil` on Mac (only for old Mac, and/or only for old build with an old C++14 gcc on Mac). Recent Apple clang included has (almost complete) support for `std::variant` .

tools/setup-buildtools.cmd
```diff
@@ -0,0 +1,50 @@
...
+if ERRORLEVEL 1 (
+  REM Build our own vcpkg from source
+  pushd .\vcpkg
+  call bootstrap-vcpkg.bat
+  popd
+)
+
+REM Install it
+vcpkg install gtest:x64-windows
+vcpkg install --head --overlay-ports=%~dp0\ports benchmark:x64-windows
```
maxgolov, line 44: This is a trick that allows to pin to specific older version of a package with `vcpkg` package manager, but then we'd have to clone / cache the port files from the mainline `vcpkg` repo. We were already broken once by `vcpkg` update before.

tools/setup-buildtools.cmd
```diff
@@ -0,0 +1,50 @@
...
+vcpkg install ms-gsl:x64-windows
+vcpkg install nlohmann-json:x64-windows
+vcpkg install abseil:x64-windows
```
maxgolov, line 47: Abseil is only needed for `std::variant` on Visual Studio 2015, and not necessary otherwise.

tools/vcvars.cmd
```diff
@@ -0,0 +1,86 @@
+@echo off
```
maxgolov, line 1: This is Visual Studio version detection and environment setup script.

maxgolov:
Bazel on Mac is broken (unrelated to my commit):

```console
ERROR: /Users/runner/work/opentelemetry-cpp/opentelemetry-cpp/examples/zpages/BUILD:17:10: Target '//examples/zpages:zpages_example' depends on toolchain '@local_config_cc//:cc-compiler-darwin', which cannot be found: error loading package '@local_config_cc//': cannot load '@local_config_cc_toolchains//:osx_archs.bzl': no such file'
ERROR: Analysis of target '//examples/zpages:zpages_example' failed; build aborted: Analysis failed
INFO: Elapsed time: 232.660s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (54 packages loaded, 103 targets configured)
FAILED: Build did NOT complete successfully (54 packages loaded, 103 targets configured)
Error: Process completed with exit code 1.
```

pyohannes:
What's the exact intention of those scripts? Are they meant to replace the setup scripts we currently use in our CI workflow?

pyohannes:
> Unfortunately package managers across different distros do not typically allow to easily pin to given dependency version. Recent example: our CI code formatting loop behaves differently on Ubuntu 18.x vs 20.xx

I'm afraid this still doesn't solve the code formatting issues, as for that we're relying on `clang-fomat`, which has to be installed on the system.

ci/do_ci.sh
jsuereth: +1 to this!!!!

maxgolov:
> What's the exact intention of those scripts? Are they meant to replace the setup scripts we currently use in our CI workflow?

These are meant to be used for local dev builds. Some of it could be used by CI. Current CI coverage for Windows / msvc is incomplete. These scripts add the necessary setup for vs2015/vs2017/vs2019.

> I'm afraid this still doesn't solve the code formatting issues

There was no intent to cover this part.

maxgolov:
There is another practical reason why I'd like to integrate this change first: my other change (STL) may depend on some of it. Integrating this `no-code change, tools only` will dramatically shrink the other PR I have open ( #374 ).

maxgolov:
I can provide a separate write-up on how to use this and Visual Studio IDE to streamline the work with any compiler (vs2015, vs2017, vs2019) and `vcpkg` on Windows. We may eventually add the scripts as well, to illustrate the prototyping of `vcpkg install opentelemetry` - since we'd have all the tools setup infra as well. From installing CMake, Visual Studio, `vcpkg` - to actually using `vcpkg' to pull the repo and validate that the compilation of the recipe succeeds. Then we can send the OpenTelemetry port file for integration in the mainline `vcpkg`. All of this needs validation with various versions of Visual Studio, for which I'm integrating these scripts. Existing contents of `ci` does not have nearly the same level of streamlined tooling setup.

tools/build-vs2015.cmd
```diff
@@ -0,0 +1,6 @@
+set "VS_TOOLS_VERSION=vs2015"
+REM Ref. https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2014%202015.html
+set "CMAKE_GEN=Visual Studio 14 2015 Win64"
```
ThomsonTan, line 3: Are we going to support 32-bit Windows?

maxgolov: Yes. I can add a parameter for that, with default being Win64.

ci/do_ci.sh
maxgolov: @jsuereth - I decided to add this small patch in another simpler PR, so that it gets in sooner.

maxgolov:
I'm not sure why some Bazel runs are failing. They are all passing on my machine. Both `./ci/do_ci.sh bazel.test` and `CC=/usr/bin/gcc-4.8 ./ci/do_ci.sh bazel.legacy.test`

docker/ubuntu14.04/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:14.04
```
lalitb, line 1: Thanks for adding these `Dockerfile`. This would help us to test changes locally before raising PR. Just thinking, if we can use these Dockerfiles in the CI environment - i.e.. running our builds within these container instead of directly on github virtual machines.  Github action does support it : https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/creating-a-docker-container-action.

jsuereth:
@maxgolov Looks like the bazel build is failing due to not finding resource.proto.  Did something change in this CL with how the proto files are imported?

docker/ubuntu14.04/Dockerfile
maxgolov: Yes, possibly. I was thinking about this too. For now my focus was on improving / speeding up the dev experience. I did this for Windows-Docker, but I will add corresponding script for Mac and Linux docker as well. Right now the set of files allows us to capture wide spectrum of execution environments:
- Ubuntu 14.04 - gcc-4.8
- Ubuntu 14.04 - gcc-5
- Ubuntu 18.04 - gcc-7
- Ubuntu 20.04 - gcc-8

Incremental builds, the way how the docker images are now set up  - with pre-installed latest CMake, are blazing fast! The goal is to build, including ability to incrementally build, and run all the tests in one run.

ci/do_ci.sh
```diff
@@ -112,33 +112,40 @@ EOF
   examples/plugin/load/load_plugin_example ${PLUGIN_DIR}/libexample_plugin.so /dev/null
   exit 0
 elif [[ "$1" == "bazel.test" ]]; then
+  bazel clean --expunge
```
maxgolov, line 115: I am struggling with Bazel build failing on GitHub action. But working totally fine on my local setup. It fails around `opentelemetry-protobuf` repo, not finding file..

tools/build-benchmark.cmd
```diff
@@ -0,0 +1,35 @@
+@echo off
```
maxgolov, line 1: This may no longer be needed per se. At least with latest VS2019 and `vcpkg` installed - it is no longer used.

maxgolov:
> @maxgolov Looks like the bazel build is failing due to not finding resource.proto. Did something change in this CL with how the proto files are imported?

@jsuereth - I replaced the location from a local snapshot copy of `opentelemetry-proto` repo to git submodule instead. Let me check if the code is cloning it recursively. Since my local build passes - it could be a parameter in the checkout action. I'm looking at this.

maxgolov:
DocFX check is miserably failing on ALL submodules not following the rules 😁  ... I'll remove submodules from the DocFX build.

maxgolov:
@jsuereth  - yay! I fixed the Bazel build. It was that I started using submodule for tracking `opentelemetry-proto` from its origin via commit id, but did not add recursive cloning in the GitHub aciton. Thus, my local build was fine and the cloud build was not fine.

I think there has to be some way to tell Bazel build to just use github URL directly, so then there would be no need to clone the code recursively. Unfortunately I don't know how to do it for Bazel.

maxgolov:
(fixing the the remaining issue with prometheus exporter)

ci/do_ci.sh
maxgolov: @pyohannes - I updated the cmake formatting tools to latest, which matches latest as of now in Ubuntu 20.04. Test it all works in both 18.04 and 20.04, have not tested in other distros. Fixes #375

maxgolov:
Right now we do not recursively clone all submodules in CI. Since only `prometheus-cpp` (prometheus exporter) needs this - I kept back the build script in CI to do this recursive cloning.. However, users that would like to build with our `third_party` deps folder, need to perform recursive clone of the repo:

```console
git clone --recurse-submodules https://github.com/open-telemetry/opentelemetry-cpp
```

Or, for existing devs who already have the repo snapshot:

```console
git submodule update --recursive --init
```

There could be other alternatives, i.e. no recursive cloning required if all deps get installed either by Bazel; or by `brew` or by `vcpkg` (for CMake). Only `opentelemetry-proto` presently is not accessible via `vcpkg` or `brew`, but we can contribute a recipe for that. Then the entire build system can be made flexible, allowing to rebuild the SDK from given tag - branched off or cooked with custom dependency version, i.e. in case if security vulnerability is reported in one of our dependencies. Support for package versions in `vcpkg` is coming soon.

maxgolov:
@pyohannes @jsuereth @ThomsonTan @lalitb @reyang

I would like to send a follow-up documentation PR. All tests are passing. To recap what's done:
- use git submodules with precise pinning to repo branch and commit id
- add build scripts that enable multiple build configurations for CMake on Windows, Linux and Mac
- add docker build scripts for rapid dev builds
- add `vcpkg` support - that works for building SDK from source (this is very early draft and will be tuned separately)
- add docs that document how `vcpkg` installation works
- on Windows only: the `build.cmd` allows to spin two build configurations at once - `nostd` and `std`
- on Linux: default build configuration is `nostd`, but alternate config could be supplied as command-line argument
- add support for Visual Studio 2019 `CMake` - no need to use any command-line tooling (the build also self-deploys the missing packages via `vcpkg`, if packages were not found elsewhere)
- prefer local GTest and Benchmark over one with distro, as in case if you use different compilers (gcc-4.x and gcc-7) on one machine, you cannot deploy these to `/usr/local` - you must use the build targeting specific compiler+configuration, otherwise your `std::` classes are gonna fail on you miserably. This is if you work with dramatically old gcc on latest machine, that builds the rest of the world with latest gcc.
- minor moves: moving `bazel` stuff to `/bazel`, and moving `.cmake` stuff to `/cmake`, as it was originally intended to be. Excluding submodules from Bazel build: these are used for CMake only, except that `opentelemetry-proto` module right now used in both builds. Ideally Bazel should fetch `opentelemetry-proto` from git commit id, not using checked out tree, but all works either way.
- necessary scripts to update all deps, tooling, building (if necessary) latest CMake for systems that do not have it, e.g. `Ubuntu 14.04` - which now allows us to relax our min CMake version, bumping it to 3.18 .. The installation script allows to upgrade and install CMake to `/usr/local/bin` only if the min version found on a system is not good enough.
- using `ninja` (dramatically faster builds) if available. Which is the case if you build with `vcpkg` or if `ninja` is installed somewhere else on a system.

I'll send in a follow-up "just documentation" PR to explain the scripts and parameters. Usually it's just:
- `tools/build.sh` or
- `tools/build.cmd` or
- `vcpkg install opentelemetry`

The rest is magic done by helper scripts, that should not necessarily be used directly. Min required shell is `bash` or `cmd.exe`/`PowerShell` on Windows, no need for Python or alike. I'll describe env vars / dependencies in a separate PR.

Please review. I am sorry for 85 files.

CMakeLists.txt
```diff
@@ -15,44 +15,91 @@ option(WITH_OTLP "Whether to include the OpenTelemetry Protocol in the SDK" OFF)
...
 endif()
 
+if(WITH_OTLP)
+  set(WITH_PROTOBUF ON)
+endif()
+
 if(WITH_PROTOBUF)
   set(protobuf_MODULE_COMPATIBLE ON)
-  find_package(Protobuf CONFIG NAMES protobuf)
-  # Older versions of protobuf don't use cmake config files.
```
maxgolov, line 42: We can't use old versions anyways, since the `opentelemetry-proto` REQUIRES at least proto3...

CMakeLists.txt
```diff
@@ -15,44 +15,91 @@ option(WITH_OTLP "Whether to include the OpenTelemetry Protocol in the SDK" OFF)
...
+      endif()
+    endif()
+    # Latest Protobuf uses mixed case instead of uppercase
+    set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
   endif()
+  message("PROTOBUF_PROTOC_EXECUTABLE=${PROTOBUF_PROTOC_EXECUTABLE}")
 endif()
 
 if(WITH_OTLP)
-  include(third_party/opentelemetry-proto/Protobuf.cmake)
```
maxgolov, line 49: This cmake file could be contributed back to the mainline `opentelemetry-proto` repo. We can't add it in here since we now clone the origin, and moved the cmake under the `/cmake` directory (the rightful place for the rest of cmake scripts per se).

CMakeLists.txt
```diff
@@ -15,44 +15,91 @@ option(WITH_OTLP "Whether to include the OpenTelemetry Protocol in the SDK" OFF)
...
+  include(cmake/opentelemetry-proto.cmake)
 endif()
 
+list(APPEND CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}")
+
 if(BUILD_TESTING)
-  find_package(GTest REQUIRED)
-  find_package(benchmark REQUIRED)
+  include(CTest)
+  if(EXISTS ${CMAKE_BINARY_DIR}/lib/libgtest.a)
```
maxgolov, line 76: This is needed to ensure that our GTest is compiled with matching compiler / STL. Otherwise if we try to use the sytem-lib GTest, we may experience a break if compiling with `gcc-4.8`, but the system GTest is compiled with `gcc-10` or alike.

CMakeSettings.json
```diff
@@ -0,0 +1,51 @@
+﻿{
+    "configurations": [
```
maxgolov, line 2: This is configuration file generated for Visual Studio - CMake GUI support. Dev does not need to use command line. I will describe this process in a separate doc. You can build from GUI and you can run all Google Test from GUI, visually showing the results with Visual Studio Test Explorer - Google test plugin for that.

WORKSPACE
```diff
@@ -82,7 +82,7 @@ http_archive(
 
 http_archive(
     name = "github_nlohmann_json",
-    build_file = "//third_party/json:nlohmann_json.BUILD",
+    build_file = "//bazel:nlohmann_json.BUILD",
```
maxgolov, line 85: I'd like to reserve `third_party/json` for submodule, moving the bazel files into `/bazel` build and ignoring all of `third_party` directory, which is now used for predictable recursive cloning when using CMake.

ci/do_ci.sh
```diff
@@ -56,14 +56,11 @@ elif [[ "$1" == "cmake.exporter.prometheus.test" ]]; then
...
 #  apt-get install zlib1g-dev
 #  apt-get -y install libcurl4-openssl-dev
-  cd third_party
-  git clone https://github.com/jupp0r/prometheus-cpp
-  cd prometheus-cpp
-  git checkout v0.9.0
-  git submodule init
-  git submodule update
+  cd third_party/prometheus-cpp
+  git submodule update --recursive --init
```
maxgolov, line 60: Right now we are tracking to latest release of `prometheus-cpp` as a submodule, and all tests pass. Recursive update needed because at top-level config we do shallow clonining, not recursive. We can change this later to deep recursive cloning.

ci/do_ci.sh
```diff
@@ -56,14 +56,11 @@ elif [[ "$1" == "cmake.exporter.prometheus.test" ]]; then
...
+  [[ -d _build ]] && rm -rf ./_build
   mkdir _build && cd _build
-  cmake .. -DBUILD_SHARED_LIBS=ON
+  cmake .. -DBUILD_SHARED_LIBS=ON -DUSE_THIRDPARTY_LIBRARIES=ON
```
maxgolov, line 63: Use deps from prometheus build tree, to ensure we build all with the matching compiler.

ci/install_format_tools.sh
```diff
@@ -3,6 +3,6 @@
 set -e
 
 apt install -y clang-format-8 python3-pip git curl
-pip3 install cmake_format==0.6.5
+pip3 install cmake_format==0.6.13
```
maxgolov, line 6: Latest official formatter produces much cleaner result.

cmake/opentelemetry-proto.cmake
```diff
@@ -37,7 +37,7 @@ add_custom_command(
     ${TRACE_PROTO}
 )
 
-include_directories(SYSTEM "${CMAKE_BINARY_DIR}/generated/third_party/opentelemetry-proto")
+include_directories("${GENERATED_PROTOBUF_PATH}")
```
maxgolov, line 40: It's not a system and there's already variable defined for that same path above.

docker/ubuntu14.04/Dockerfile
maxgolov: I logged it in #393

docker/ubuntu14.04/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+	tar			\
+	zip			\
+	unzip			\
+	wget			\
+	zlib1g-dev
+
+RUN mkdir -p /usr/local/bin
+
+## Install cmake since it's an expensive operation and best be done once
+COPY ./setup-cmake.sh /usr/local/bin/setup-cmake.sh
```
maxgolov, line 32: This dramatically speeds up incremental build on old Ubuntu, as old distro only has CMake2 and the build of CMake-3.18 itself is very expensive operation. Ideally this could have been handled by providing CMake PPA for all distros. But we build from source (once only, per cached docker image).

tools/ports/opentelemetry/portfile.cmake
```diff
@@ -0,0 +1,76 @@
+include(vcpkg_common_functions)
```
maxgolov, line 1: This is very early draft. It is functional. We need to tune it later on by adding the proper support for `feature packages` : build with Abseil, build with Standard Library, and other possible build flavors, like build with prometheus, build with this or with that.

tools/setup-buildtools.sh
```diff
@@ -22,30 +25,12 @@ yum -y install devtoolset-7-valgrind
 
 yum-config-manager --enable rhel-server-rhscl-7-rpms
 
-if [ `gcc --version | grep 7` == "" ]; then
```
maxgolov, line 25: Check for 7 was bad, and cmake installation got moved out into its own separate much smarter script right now.

tools/setup-cmake.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
maxgolov, line 1: This works across all distros to install more recent CMake, if build system needs 3.18, for example. Technically we can decide on what's min required whenever we need to ask for a new CMake feature, like `ninja` support - we can make sure our build scripts deploy the needed CMake version first before triggering the build.

tools/setup-protobuf.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
maxgolov, line 1: Running OTLP exporter on old distros requires new `protoc3`, i.e. I'm fixing here something that never worked before on Ubuntu 14.04. Now it works great.

meastp:
@maxgolov I really like this as an opentelemetry consumer on Windows (using Visual Studio with MSBuild):

You only mention `vcpkg install opentelemetry` but when using the manifests feature it is only necessary to specify opentelemetry in `vcpkg.json`. Even though vcpkg is currently missing versioning (adding versions to vcpkg.json) and repositories (makes it possible to retrieve libraries from sources other than the official vcpkg git repo), perhaps it is worth mentioning? 👍 

I guess, for development builds to be able to use vcpkg, versioning support is required...

maxgolov:
@meastp - thank you for your feedback! The `vcpkg` part is very early draft at this point. I used the "response file" approach for now, but I think what you are suggesting is gonna be better. Practical reasons why (for now) I had to use submodules pinned at certain commit ID for dependencies -- is to ensure that we can rebuild with the exact set of supported dependencies.. However, going forward - once we get `vcpkg` to support versioning, we can fully rely on `vcpkg` for _reproducible_ builds that result in the same functionality.

maxgolov:
@meastp - I will also provide a separate write-up how to consume this with Visual Studio 2019 community edition. It does have an awesome built-in CMake and Google Test support. One of the files in there is specifically for defining multiple build configurations in a json file.. We can probably have certain build configurations that "cook" OpenTelemetry C++ SDK with certain slimmed down set of exporters needed for particular use-case.

CMakeSettings.json
pyohannes: For what use cases does this file make sense? Do those default settings make sense for most VS users?

CMakeSettings.json
```diff
@@ -0,0 +1,51 @@
...
+        {
+            "name": "nostd-x64-Debug",
+            "generator": "Ninja",
+            "configurationType": "Debug",
+            "inheritEnvironments": [ "msvc_x64_x64" ],
+            "buildRoot": "${projectDir}\\out\\vs2019\\${name}",
+            "installRoot": "${projectDir}\\out\\vs2019\\${name}\\install",
```
pyohannes, line 9: Does this only make sense for VS2019, or for previous versions too?

docker/ubuntu14.04/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+
+## Update cache and upgrade image
+RUN apt-get -y update && apt-get -y upgrade && apt-get -y dist-upgrade
+
+## Build environment packages
+RUN apt-get install -qq -y --ignore-missing \
+	apt-utils		\
+	automake		\
+        bc                      \
```
pyohannes, line 11: Small indentation nit.
```suggestion
      bc                      \
```

docker/ubuntu16.04/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+## Update cache and upgrade image
+RUN apt-get -y update && apt-get -y upgrade && apt-get -y dist-upgrade
+
+## Build environment packages
+RUN apt-get install -qq -y --ignore-missing \
+	apt-utils		\
+	automake		\
+	build-essential		\
+        bc                      \
```
pyohannes, line 12:
```suggestion
      bc                      \
```

docker/ubuntu18.04/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+
+## Update cache and upgrade image
+RUN apt-get -y update && apt-get -y upgrade && apt-get -y dist-upgrade
+
+## Build environment packages
+RUN apt-get install -qq -y --ignore-missing \
+	apt-utils		\
+	automake		\
+        bc                      \
```
pyohannes, line 11:
```suggestion
      bc                      \
```

docker/ubuntu20.04/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+
+## Update cache and upgrade image
+RUN apt-get -y update && apt-get -y upgrade && apt-get -y dist-upgrade
+
+## Build environment packages
+RUN apt-get install -qq -y --ignore-missing \
+	apt-utils		\
+	automake		\
+        bc                      \
```
pyohannes, line 11:
```suggestion
      bc                      \
```

pyohannes (approved):
That's a big PR, thanks for getting this up. I gave it a basic review, and there's nothing that should block a merge.

Maybe it's time to introduce platform specific folders in `tools` and `ci`, like `tools/windows`, `tools/posix`, `ci/windows`, and `ci/posix`. This might make it easier to keep good oversight.

.gitmodules
```diff
@@ -1,3 +1,29 @@
 [submodule "third_party/prometheus-cpp"]
 	path = third_party/prometheus-cpp
-	url = https://github.com/jupp0r/prometheus-cpp.git
+	url = https://github.com/jupp0r/prometheus-cpp
+	branch = master
```
ThomsonTan, line 4: Do we need create submodule based on stable branch instead of master?

tools/build-docker.cmd
```diff
@@ -0,0 +1,69 @@
+@echo off
```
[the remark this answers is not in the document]
lalitb: Nit: would be good to have a ticket for tracking purpose if this is going to be done outside this PR.

lalitb (approved):
Thanks for the PR. Appreciate for coming up with submodules and DockerFiles. It's somewhat big PR to review thoroughly, but looks good in general : )

docker/ubuntu14.04/Dockerfile
ThomsonTan: As these scripts only require to be run for one time, why copying them into /usr/local/bin, instead of running it in place?

docker/ubuntu20.04/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM ubuntu:20.04
+ENV DEBIAN_FRONTEND noninteractive
+
+## Update cache and upgrade image
+RUN apt-get -y update && apt-get -y upgrade && apt-get -y dist-upgrade
```
ThomsonTan, line 5: Seems the setup process for all distros are almost the same, perhaps it could be extracted into a script, then all Dockerfiles could just run the shared script?

tools/build-benchmark.cmd
```diff
@@ -0,0 +1,35 @@
...
+set "GOOGLETEST_PATH=%ROOT%\third_party\googletest"
+if not exist "build" (
+  mkdir build
+)
+cd build
+
+REM By default we generate the project for the older Visual Studio 2017 even if we have newer version installed
+cmake ../ -G %CMAKE_GEN% -Ax64 -DBENCHMARK_ENABLE_TESTING=OFF
+set SOLUTION=%ROOT%\third_party\benchmark\build\benchmark.sln
+msbuild %SOLUTION% /maxcpucount:%MAXCPUCOUNT% /p:Configuration=Debug /p:Platform=x64
```
ThomsonTan, line 31: Use `cmake --build .` instead of calling `msbuild` to be consistent?

tools/download.cmd
```diff
@@ -0,0 +1,3 @@
+@REM This script allows to download a file to local machine. First argument is URL
+set "PATH=C:\Windows;C:\Windows\System32;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files\Git\bin"
```
ThomsonTan, line 2: Reference system drive via `%SystemRoot%` instead of `C:`?

tools/install-vs-addons.cmd
```diff
@@ -0,0 +1,39 @@
+set "PATH=C:\Windows;C:\Windows\System32;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files\Git\bin"
```
ThomsonTan, line 1: Prefer `%SystemRoot%` instead of `C:`?

tools/install-vs-addons.cmd
```diff
@@ -0,0 +1,39 @@
...
+cd %~dp0
+call powershell -File .\install_llvm-win64.ps1
+
+REM Download Visual Studio LLVM extension required for clang build to succeed
+call download.cmd https://llvmextensions.gallerycdn.vsassets.io/extensions/llvmextensions/llvm-toolchain/1.0.363769/1560930595399/llvm.vsix
+
+REM Install optional components required for ARM build - vs2017-BuildTools
+IF EXIST "%ProgramFiles(x86)%\Microsoft Visual Studio\2017\BuildTools" (
+"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vs_installer.exe" ^
+	-- modify --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2017\BuildTools" -q ^
```
ThomsonTan, line 11: Curious whether the `-- ` before `modify` is necessary?

tools/install.sh
```diff
@@ -0,0 +1,13 @@
...
+WORKSPACE_ROOT=$DIR/..
+pushd $WORKSPACE_ROOT
+OPENTELEMETRY_INSTALL_DIR=${1-/usr/local}
+echo "Install SDK to $OPENTELEMETRY_INSTALL_DIR"
+mkdir -p $OPENTELEMETRY_INSTALL_DIR/lib
+# TODO: install libraries?
+cp -R api/include $OPENTELEMETRY_INSTALL_DIR
+cp -R sdk/include/opentelemetry/sdk $OPENTELEMETRY_INSTALL_DIR/include/opentelemetry
+mkdir -p $OPENTELEMETRY_INSTALL_DIR/lib
+echo dummy > $OPENTELEMETRY_INSTALL_DIR/lib/libopentelemetry.stub
```
ThomsonTan, line 12: What is this `.stub` file for?

tools/ports/benchmark/portfile.cmake
ThomsonTan: Is sharing between VS 2019 and 2017 the reason to add this vcpkg port? I think vcpkg includes port of `benchmark`.

tools/ports/benchmark/portfile.cmake
```diff
@@ -0,0 +1,49 @@
+if(VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
```
ThomsonTan, line 1: Do we plan to upstream the port to official vcpkg? That will make opentelemetry-cpp easier to be consumed.

maxgolov:
I'll fix the formatting. It's because I upgraded the formatting tools - it got broken after update 😄

tools/ports/benchmark/portfile.cmake
maxgolov: Yes, absolutely.

tools/install.sh
maxgolov: The issue is that `vcpkg` _requires_ us to have _something_ in its own workspace lib dir. I added some temporary stub there, but I think we need to eventually post the actual `libopentelemetry_common.a`, etc. - all other libs. This is draft at this point, just to make sure that `vcpkg` compiles, builds and installs successfully. I will incrementally improve it later on, removing the `*.stub` file.

tools/install-vs-addons.cmd
maxgolov: Good question. Let me follow-up on this.

tools/build-benchmark.cmd
maxgolov: We may be able to remove this script, as right now we have a few ways to inject the benchmark via `vcpkg`.

docker/ubuntu20.04/Dockerfile
maxgolov: There could be subtle differences, esp. since I will add rpm and yum (CentOS, RHEL) later on. I think it also helps visually to see the sequence of commands in a central spot. One exception is the CMake-installation script, the rest of deps are all described in `Dockerfile` ... They are usually almost the same.. until they are not 😄

.gitmodules
maxgolov: They are publishing tags on `master` , so we do use their stable commit associated with latest tag. When submodule checkout is happening, it is fetching from `master` + commit id. So we do pin to stable.

.github/workflows/ci.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+      with:
+        submodules: 'true'
```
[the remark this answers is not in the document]
maxgolov: Indeed, I handled this manually in the `do_ci.sh` script to perform shallow clone rather than deep recursive. Maybe..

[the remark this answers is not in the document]
maxgolov: OK, I'll change this to deep clone, then we do not have to perform cloning of prometheus in `do_ci.sh` script anymore.

CMakeSettings.json
maxgolov: Unfortunately the proper CMake support was only added to recent update of Visual Studio 2019. This file is unused by older versions of Visual Studio. I can check if it works for Visual Studio code.. Most likely it should also work for Visual Studio Code with extension.

CMakeSettings.json
maxgolov: These configurations are identical to default CMake configuration, building SDK with OTLP protocol and Examples. We can add more configurations here. I was also planning to add (in a separate commit) - configuration that builds with Standard Library instead of nostd.. since there is a large of possible configuration, maybe we can discuss what are the configurations we'd like to show here.

How it works - you select one of the ready-to-use configurations to build. And build it. You can also add more configurations, which get added to this file.

We can review this list and tune in a follow-up PR.

ci/do_ci.sh
maxgolov: We should be fine now even without it - because I changed the YAML file for the GitHub Action to do deep recursive clone. But even with this - we are fine, since we'd just run thru the update without updating anything. It'll be the same as what's already fetched, exiting quite quickly from this line.

CMakeLists.txt
```diff
@@ -15,44 +15,91 @@ option(WITH_OTLP "Whether to include the OpenTelemetry Protocol in the SDK" OFF)
...
+      # Always use x64 protoc.exe
+      if(NOT EXISTS "${Protobuf_PROTOC_EXECUTABLE}")
+        set(Protobuf_PROTOC_EXECUTABLE
+            ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/packages/protobuf_x64-windows/tools/protobuf/protoc.exe
+        )
+      endif()
+    endif()
+    # Latest Protobuf uses mixed case instead of uppercase
+    set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
   endif()
```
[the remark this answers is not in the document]
maxgolov: I remember about it. I'll organize this better in a separate PR, where I'll add nlohmann JSON dependency.

ThomsonTan (approved):

docker/ubuntu14.04/Dockerfile
maxgolov: These still exist in prebuilt docker image. So, say, if you want to re-install or upgrade - you can reuse the script to pass a newer version of CMake to that and it'll upgrade the CMake version just for your current Docker session.

tools/build-benchmark.cmd
maxgolov: I'll keep it as-is for now. But it's not used right now by the outside `build.cmd` script. I'll probably remove this `build-benchmark.cmd` in a follow-up PR.

tools/install-vs-addons.cmd
maxgolov: I think it wasn't necessary. Not sure where it came from. It was benign and I removed the empty param.

tools/install.sh
maxgolov: Will address in more proper vcpkg support / follow-up PR that will publish the `opentelemetry*.a` files to install dir

tools/ports/benchmark/portfile.cmake
maxgolov: This one was a "hack" to make sure we fix the compat issue in Google Benhmark, i.e. allow the benchmark build with vs2019 to be used by vs2017 compile. I'll see what we can do, if I can contribute it as an issue / recommending the patch we applied here..

Issue #337, opened 2020-09-24
Logging prototype


reyang:
The goal:

* Prototype the logging API that can be used by both the SDK consumer and SDK author.
  * If we could find an existing C++ logging API that is widely adopted + cross platform, we would want to leverage it instead of reinventing wheels.
* Implement the Processor and Exporter logic similar like SpanProcessor and trace exporter.
* Implement ElasticSearch Log Exporter which sends batches of logs from the SDK directly to ElasticSearch via HTTP REST API.

The stretch goal:

* Provide a console log exporter for local debugging purpose.
* Provide a way to associate trace id / span id and other context/baggage to log entries.
* Implement OTLP Exporter https://github.com/open-telemetry/opentelemetry-proto/tree/master/opentelemetry/proto/logs/v1.

Non-goal:

* Provide log sampler interface and implementation.
* Implement Fluentd/Fluentbit exporter.
* Implement local file exporter.

Initial list of work items:

* [x] API to fetch Logger (e.g. LoggerProvider)
* [x] Logging API
* [x] Log Processor interface
* [x] Log Exporter interface
* [x] SimpleExportProcessor similar like this https://github.com/open-telemetry/opentelemetry-cpp/blob/2973baaf476a92e9d11d55e82f1dad78fefe6de1/sdk/include/opentelemetry/sdk/trace/simple_processor.h.
* [x] BatchExportProcessor similar like this https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/batch_span_processor.cc.
* [x] Console Exporter
* [x] ElasticSearch Exporter

reyang:
@alolita @MarkSeufert @xukaren

reyang:
@maxgolov @lalitb @ThomsonTan FYI

xukaren:
> Seems I cannot assign the issue to @MarkSeufert and @xukaren at this moment. Please reply back once you are member of the OpenTelemetry community and I'll assign this to you. Thanks!
> 
> @alolita @MarkSeufert @xukaren

done!

xukaren:
Both of us got approved :)

MarkSeufert:
> Seems I cannot assign the issue to @MarkSeufert and @xukaren at this moment. Please reply back once you are member of the OpenTelemetry community and I'll assign this to you. Thanks!
> 
> @alolita @MarkSeufert @xukaren

@reyang I also just got my membership!
I'm excited to get started on this issue :)

[...]

PR #378, opened 2020-10-27
Initial Commit of Logging API 

xukaren:
This PR is an initial commit of the Logging API for C++ #337. The goal of the Logging API is to only implement the minimum set of features that are required to capture (not send/configure) data from the user.  This API closely follows the [Log Data Model](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md) and [Logging specification overview](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/overview.md). 

This PR adds:
- Classes and basic methods for the Logging API, following the conventions of `metrics` and `traces`
- Unit tests for all added classes (except `LogRecord`- to be finalized)

Future enhancements include:
-  Overloaded functions for each severity level (e.g. `void debug(...)`, `void trace(...)`, `void info(...)`, ..., or `LogDebug(...)`, `LogTrace(...)`, `LogInfo(...)`,...) 
- Macros for each severity level (e.g. `LOG_ERROR`, `LOG_E`, `LOG_E_IF`, ...) 
- Functions/easier ways to create a `LogRecord` (e.g. overloaded constructors)

The design doc will be published as part of a future PR. 

Suggestions and feedback are appreciated. Further comments are also being collected on RFC #356. 

-- cc @MarkSeufert  @alolita @reyang @ThomsonTan @maxgolov


api/include/opentelemetry/logs/log_record.h
```diff
@@ -0,0 +1,124 @@
...
+  common::KeyValueIterable &resource;    // key/value pair list
+  common::KeyValueIterable &attributes;  // key/value pair list
+
+  /* Default log record if user does not overwrite this.
+   *  TODO: find better data type for "body" field
+   *  Future enhancement: Potentially add other constructors to take default arguments from the user
+   **/
+  LogRecord() : resource(_nullKV), attributes(_nullKV)
+  {
+    timestamp = core::SystemTimestamp(std::chrono::system_clock::now());
```
reyang, line 101: I think this needs to be moved to the SDK.
Consider adding a TODO comment here.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Handles log record creation.
+ **/
+class Logger
+{
+public:
+  virtual ~Logger() = default;
+
+  /* Returns the name of the logger */
+  virtual nostd::string_view getName() = 0;
```
reyang, line 42: Is this needed?

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Creates a log message with the passed characteristics.
```
reyang, line 45: Curious - what is "passed characteristics"?

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param name the name of the log event.
+   * @param sev the severity level of the log event.
```
reyang, line 48: I wish to get more opinions on the names, do we use `sev` / `msg` or the full name `severity` / `message`?
I would vote for using the full name.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+   */
+  virtual void log(const LogRecord &record) noexcept = 0;
+
+  // First, finalize core,
+  // Afterwards, come back and add overloads
+
+  /** Overloaded methods for unstructured logging **/
+  void log(nostd::string_view msg) noexcept
+  {
+    log(Severity::kNone, msg);  // Set severity to NONE as default then call the log method below
```
reyang, line 70: Not sure if `Severity::kNone` is the right level or not.
Java log4j defaults to `DEBUG` and .NET ILogger defaults to `Information`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  void log(Severity sev, nostd::string_view msg) noexcept
+  {
+    // Set timestamp to now as default then call the log method below
+    log(sev, msg, core::SystemTimestamp(std::chrono::system_clock::now()));
```
reyang, line 76: Same here, I think we need to avoid getting timestamp from the API - the API could be called when there is no actual SDK and calling timestamp would be a perf penalty (in case no SDK is used, or the sampler decided to ignore the log).

api/include/opentelemetry/logs/noop.h
```diff
@@ -0,0 +1,77 @@
...
+// used directly. Please refer to logger.h for documentation on these interfaces.
+
+#include "opentelemetry/context/runtime_context.h"
+#include "opentelemetry/logs/logger.h"
+#include "opentelemetry/logs/logger_provider.h"
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/nostd/unique_ptr.h"
+#include "opentelemetry/version.h"
+
+#include <memory>
```
reyang, line 30: Please move this above the quoted-form includes.

api/include/opentelemetry/logs/noop.h
```diff
@@ -0,0 +1,77 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace logs
+{
+/**
+ * No-op implementation of Logger. This class should not be used directly. It should only be
+ * instantiated using a LoggerProvider's GetLogger() call.
+ */
+class NoopLogger final : public Logger
```
reyang, line 39: Given we have no-op logger, I wonder if `Logger` should be an abstract class.

api/include/opentelemetry/logs/noop.h
```diff
@@ -0,0 +1,77 @@
...
+{
+public:
+  NoopLogger() = default;
+
+  // returns the name of the logger
+  nostd::string_view getName() noexcept override { return "NOOP Logger"; }
```
reyang, line 45: Do we need this?

api/include/opentelemetry/logs/logger.h
MarkSeufert: Since the LoggerProvider has a virtual GetLogger(name) method, we were thinking that a corresponding virtual getName() method would also belong in the API.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+   * @param name the name of the log event.
+   * @param sev the severity level of the log event.
+   * @param msg the string message of the log (perhaps support std::fmt or fmt-lib format).
+   * @param record the log record (object type LogRecord) that is logged.
+   * @param attributes the attributes, stored as a 2D list of key/value pairs, that are associated
+   * with this log.
+   * @throws No exceptions under any circumstances.
+   */
+
+  // Log issue about macros (disabling log events below certain level)
```
ThomsonTan, line 56: Is this comment still needed?

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  void log(Severity sev, nostd::string_view msg, core::SystemTimestamp time) noexcept
+  {
+    LogRecord r;
+    r.body      = msg;
+    r.severity  = sev;
+    r.timestamp = time;
+
+    log(r);  // converts to a LogRecord object, then calls log(LogRecord)
```
ThomsonTan, line 86: This comment looks a little confuse. It seems imply convert `r` to `LogRecord`, but `r` is a `LogRecord`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /** Overloaded methods for structured logging**/
+  template <class T,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
+  void log(nostd::string_view name = "",
+           Severity sev            = Severity::kNone,
+           const T &attributes     = {}) noexcept
+  {
```
ThomsonTan, line 95: `log()` could be called without any input parameters, is this needed and useful?

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,116 @@
...
+  // Afterwards, come back and add overloads
+
+  /** Overloaded methods for unstructured logging **/
+  void log(nostd::string_view msg) noexcept
+  {
+    log(Severity::kNone, msg);  // Set severity to NONE as default then call the log method below
+  }
+
+  void log(Severity sev, nostd::string_view msg) noexcept
+  {
```
ThomsonTan, line 74: Add severity filter before getting timestamp and constructing `LogRecord`?

api/include/opentelemetry/logs/logger.h
ThomsonTan: Mark `getName()` as `const` if it is needed?

api/include/opentelemetry/logs/logger.h
```diff
@@ -39,75 +39,95 @@ class Logger
...
    * which should at minimum assign the trace_id, span_id, and timestamp
    */
   virtual void log(const LogRecord &record) noexcept = 0;
 
-  // First, finalize core,
-  // Afterwards, come back and add overloads
 
   /** Overloaded methods for unstructured logging **/
-  void log(nostd::string_view msg) noexcept
+  void log(nostd::string_view message) noexcept
```
reyang, line 66: inline? @maxgolov @ThomsonTan what do you think?

api/include/opentelemetry/logs/logger.h
reyang: Good catch, I think we should prevent folks from calling `log()` without any argument.

api/test/logs/logger_provider_test.cc
```diff
@@ -4,6 +4,7 @@
 #include "opentelemetry/nostd/string_view.h"
 
 #include <gtest/gtest.h>
+#include <array>
```
reyang, line 7: We would want to sort things alphabetically unless there is a reason not to do so.

reyang (approved):
LGTM, there are things not hammered out but no need to block this PR.

api/test/logs/logger_provider_test.cc
xukaren: I just tried moving it but the CI format test moves it back

api/include/opentelemetry/logs/logger.h
maxgolov: Yes.

api/include/opentelemetry/logs/logger.h
maxgolov: For this one - I like the way how this is done. I think we need to follow the same order for the structured logging. Maybe some way to figure out if you got two arguments, you get unstructured event. If you pass the required additional property bag - it means you are emitting `named` -  `structured` event.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,138 @@
...
+  template <class T,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
+  void log(Severity severity       = Severity::kDefault,
+           nostd::string_view name = "",
+           const T &attributes     = {}) noexcept
+  {
+    log(severity, name, common::KeyValueIterableView<T>(attributes));
+  }
+
+  void log(Severity severity,
```
maxgolov, line 117: I swear I saw a different order of arguments in your screenshot during our offline discussion? 😄 

I like this order: `severity, name, attributes`. This is also in alignment with unstructured message order `severity, message`. This is also what you'd typically see in Android OS logging macros, for example. 👍 

You mentioned some other complexity related to too many methods needed. And I'd like to understand where that other concern originated from. If this has been solved since then, please disregard my comment.

api/include/opentelemetry/logs/logger.h
```diff
@@ -0,0 +1,138 @@
...
+    r.timestamp = time;
+
+    log(r);
+  }
+
+  /** Overloaded methods for structured logging**/
+  // TODO: separate this method into separate methods since it is not useful for user to create
+  // empty logs
+  template <class T,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
```
maxgolov, line 109: I think you can also safely add a template for `log(severity, name, T)`, where T is NOT `KeyValueIterable`. This would live in API header, not implemented, but allows a customer to provide their own implementation. That way if I do `log(severity, name,  nlohmann::json obj)` - I can plug in my own `flattener` implementation outside of SDK to existing class. Not in this PR, but consider this. I can help to provide this as a `contrib` example to show how to use this feature.. We can theoretically support wider set of types, and have that outside-of-SDK implementation to flatten into OT typing system. This will certainly be needed to migrate customers that might have been using wider typing system in their other logging framework.

api/include/opentelemetry/logs/logger.h
xukaren: (Yup we just updated it after the meeting! 😃)

So the other question was how to support optional fields, like: 
 `log(severity, name, attributes)` 
 `log(severity, attributes)` 
 `log(name, attributes)` 
 `log(attributes)`, etc., 

It seems like we would need to write a separate function for each combination of optional parameters. (Since we won't be allowed to set default values of parameters earlier in the list without setting default values for ones after it too). 

In this example, we just have 3 fields, but we're not sure how it would be done to support all 10?

api/include/opentelemetry/logs/logger.h
MarkSeufert: Yea I totally agree with this, we'll submit this addition in a later commit! We'll leave it as a non-pure virtual function so that a definition isn't required to compile.

api/test/logs/logger_provider_test.cc
reyang: In this way we can ignore it for now.

ThomsonTan (approved):

Issue #214, opened 2020-07-27
Update Recordable implementations to use SpanContext


nadiaciobanu:
**Problem**
In the `Recordable` interface, links can be added by passing a `SpanContext` (containing `trace_id`, `span_id` and `trace_state`) into `AddLinks()`. Currently, `SpanContext` is not implemented, so the `SpanContext`-related fields in the links cannot be populated.

**Solution**
Update `AddLinks()` in implementations of the `Recordable` interface, including the [`OTLP Recordable`](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/exporters/otlp/src/recordable.cc) and [`SpanData`]( https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/span_data.h) to set `trace_id`, `span_id` and `trace_state` using the `SpanContext` parameter.

lalitb:
#380 fixes it.

PR #380, opened 2020-10-29
Populate trace_id and spand_id to Recordable in Recordable::AddLink

ThomsonTan:
This covers https://github.com/open-telemetry/opentelemetry-cpp/issues/214.

The `SpanContext` parameter was changed to const reference because `SpanContext` have quite a few Id fields which are heavy for passing by value.

### DIFF
--- a/exporters/otlp/src/recordable.cc
+++ b/exporters/otlp/src/recordable.cc
@@ -132,7 +132,7 @@ void Recordable::AddEvent(nostd::string_view name,
   });
 }
 
-void Recordable::AddLink(opentelemetry::trace::SpanContext span_context,
+void Recordable::AddLink(const opentelemetry::trace::SpanContext &span_context,
                          const common::KeyValueIterable &attributes) noexcept
 {
   auto *link = span_.add_links();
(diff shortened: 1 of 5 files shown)


exporters/otlp/src/recordable.cc
```diff
@@ -132,15 +132,20 @@ void Recordable::AddEvent(nostd::string_view name,
...
                          const common::KeyValueIterable &attributes) noexcept
 {
   auto *link = span_.add_links();
   attributes.ForEachKeyValue([&](nostd::string_view key, common::AttributeValue value) noexcept {
     PopulateAttribute(link->add_attributes(), key, value);
     return true;
   });
-  // TODO: Populate trace_id, span_id and trace_state when these are supported by SpanContext
+
+  span_.set_trace_id(reinterpret_cast<const char *>(span_context.trace_id().Id().data()),
```
reyang, line 144: I might be missing something - why would we put the link's trace id as the span's trace id?

ThomsonTan: That's a good question. I thought Recordable is meant to capture the link `SpanContext`. Seems I am wrong and we don't need capture the link `SpanContext` here?

reyang: I think each link has to hold the spancontext data plus a set of attributes.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#specifying-links

pyohannes: This seems wrong. Adding a link should never overwrite the span's trace id.

Can you please add unit tests for this, so we can more better verify those conditions?

reyang (approved):
LGTM.

PR #381, opened 2020-10-30
Conditional compile generic lambda

ThomsonTan:
Conditional compile generic lambda which is available in C++14 and above. With this conditional compilation, we could remove the fallback when upgrading our language standard.


reyang (approved):
LGTM.

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
```diff
@@ -89,6 +89,9 @@ class OStreamSpanExporter final : public sdktrace::SpanExporter
     sout_ << ']';
   }
 
+// Prior to C++14, generic lambda is not available so fallback to functor.
+#if __cplusplus < 201402L
```
lalitb, line 93: LGTM in general. Although I had concern with this macro whether all the compilers define it properly ( I know there were issues earlier, probably not any more ).
jfyi - there are also feature specific recommendation - eg  `__cpp_generic_lambdas` for compilers supporting generic lambda.  Ref: https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#__cpp_generic_lambdas

lalitb (approved):

PR #383, opened 2020-10-30
Change strategy for detaching contexts

### LINKED ISSUE #219: Memory management and asynchronous span handling for Context API
We've decided to move forward with the context API PR #209 in order to unblock other work stream.
However there is memory ownership issue as we've discussed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/209#discussion_r461084358).
Use this to track the issue until we have a code fix.

pyohannes:
With the existing solution, users had to make sure that they always detached contexts in exactly the reverse orders they were attached. So, for attaching contexts in the order `A - B - C`, they have to be detached in the order `C - B - A`. If a user would try to detach `B` before `C`, this would lead to an error (which the user doesn't notice if they don't explicitly check the return code of `Detach`). This can lead problems if a user forgets to detach a context (in this case, this context would never be detached).

With this PR the strategy is changed so that if a context is detached, all it's active child contexts are detached too. So if a user attaches contexts in the order `A - B - C`, and then detached `B`, both `B` and `C` will be detached. This is a much more robust solution that will avoid dangling contexts.

Fixes #219.

### DIFF
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -175,8 +175,9 @@ public:
   // Return the current context.
   Context GetCurrent() noexcept override { return GetStack().Top(); }
 
-  // Resets the context to a previous value stored in the
-  // passed in token. Returns true if successful, false otherwise
+  // Resets the context to the value previous to the passed in token. This will
+  // also detach all child contexts of the passed in token.
+  // Returns true if successful, false otherwise.
   bool Detach(Token &token) noexcept override
   {
     if (!GetStack().Contains(token))
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -179,11 +179,18 @@ public:
   // passed in token. Returns true if successful, false otherwise
   bool Detach(Token &token) noexcept override
   {
-    if (!(token == GetStack().Top()))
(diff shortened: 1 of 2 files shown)


api/include/opentelemetry/context/runtime_context.h
```diff
@@ -175,15 +175,23 @@ class ThreadLocalContextStorage : public RuntimeContextStorage
...
 
-  // Resets the context to a previous value stored in the
-  // passed in token. Returns true if successful, false otherwise
+  // Resets the context to the value previous to the passed in token. This will
+  // also detach all child contexts of the passed in token.
+  // Returns true if successful, false otherwise.
   bool Detach(Token &token) noexcept override
   {
-    if (!(token == GetStack().Top()))
+    if (!GetStack().Contains(token))
```
reyang, line 183: I guess in 99% of the chance `GetStack().Top()` would equal to the token, so it might worth a special case optimization (might save few CPU cycles on L188?):
```c
if (token == GetStack().Top())
{
  GetStack().Pop();
  return true;
}

// check if token exists on the stack and pop items above it
```

reyang (approved):
LGTM.

api/include/opentelemetry/context/runtime_context.h
pyohannes: Good point. Added.

PR #386, opened 2020-11-06
Add Logging SDK implementation for Logging API

MarkSeufert:
This PR contains an initial implementation of the Logging SDK, which follows from the [initial API commit](https://github.com/open-telemetry/opentelemetry-cpp/pull/378#partial-pull-merging) and the [logging prototype issue](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/overview.md). 

This PR adds:
* SDK LoggerProvider class - returns instances of the SDK Logger and stores a common processor
* SDK Logger class - filters logs based off severity and sends to a processor
* Processor Interface - interface that defines what a processor implementation will require. A future PR will contain concrete processor classes derived from this
* Unit tests for the LoggerProvider and Logger classes

Any suggestions or feedback would be greatly appreciated! The design doc for the Logging SDK will be released in a future PR.

cc @xukaren @alolita @reyang @ThomsonTan @maxgolov @tigrannajaryan

### DIFF
--- a/sdk/include/opentelemetry/sdk/logs/logger.h
+++ b/sdk/include/opentelemetry/sdk/logs/logger.h
@@ -1,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "opentelemetry/logs/log_record.h"
+#include "opentelemetry/logs/logger.h"
+#include "opentelemetry/nostd/shared_ptr.h"
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/sdk/common/atomic_shared_ptr.h"
+#include "opentelemetry/sdk/logs/logger_provider.h"
+#include "opentelemetry/sdk/logs/processor.h"
+
+#include <unordered_map>
+#include <vector>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class LoggerProvider;
+
+class Logger final : public opentelemetry::logs::Logger
+{
+public:
+  /**
+   * Initialize a new logger.
+   * @param logger_provider The logger provider that owns this logger.
+   */
+  explicit Logger(std::shared_ptr<LoggerProvider> logger_provider) noexcept;
+
+  /**
+   * Writes a log record into the processor.
+   * @param record The record to write into the processor.
+   */
+  void log(const opentelemetry::logs::LogRecord &record) noexcept override;
+
+private:
+  // The logger provider of this Logger. Uses a weak_ptr to avoid cyclic dependancy issues the with
+  // logger provider
+  std::weak_ptr<LoggerProvider> logger_provider_;
+};
+
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/include/opentelemetry/sdk/logs/logger_provider.h
+++ b/sdk/include/opentelemetry/sdk/logs/logger_provider.h
@@ -1,0 +1,101 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <memory>
+#include <mutex>
+#include <string>
+#include <vector>
+
+#include "opentelemetry/logs/logger_provider.h"
+#include "opentelemetry/logs/noop.h"
+#include "opentelemetry/nostd/shared_ptr.h"
+#include "opentelemetry/sdk/common/atomic_shared_ptr.h"
+#include "opentelemetry/sdk/logs/logger.h"
+#include "opentelemetry/sdk/logs/processor.h"
+
+// Define the maximum number of loggers that are allowed to be registered to the loggerprovider.
+// TODO: Add link to logging spec once this is added to it
+#define MAX_LOGGER_COUNT 100
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class Logger;
+
+class LoggerProvider final : public opentelemetry::logs::LoggerProvider,
+                             public std::enable_shared_from_this<LoggerProvider>
+{
+public:
+  /**
+   * Initialize a new logger provider. A processor must later be assigned
+   * to this logger provider via the SetProcessor() method.
+   */
+  explicit LoggerProvider() noexcept;
+
+  /**
+   * Creates a logger with the given name, and returns a shared pointer to it.
+   * If a logger with that name already exists, return a shared pointer to it
+   * @param name The name of the logger to be created.
+   * @param options (OPTIONAL) The options for the logger. TODO: Once the logging spec defines it,
+   * give a list of options that the logger supports.
+   */
+  opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> GetLogger(
+      opentelemetry::nostd::string_view name,
+      opentelemetry::nostd::string_view options = "") noexcept override;
+
+  /**
+   * Creates a logger with the given name, and returns a shared pointer to it.
+   * If a logger with that name already exists, return a shared pointer to it
+   * @param name The name of the logger to be created.
+   * @param args (OPTIONAL) The arguments for the logger. TODO: Once the logging spec defines it,
+   * give a list of arguments that the logger supports.
+   */
+  opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> GetLogger(
+      opentelemetry::nostd::string_view name,
+      nostd::span<nostd::string_view> args) noexcept override;
+
+  /**
+   * Returns a shared pointer to the processor currently stored in the
+   * logger provider. If no processor exists, returns a nullptr
+   */
+  std::shared_ptr<LogProcessor> GetProcessor() noexcept;
+
+  // Sets the common processor that all the Logger instances will use
+  /**
+   * Sets the processor that is stored internally in the logger provider.
+   * @param processor The processor to be stored inside the logger provider.
+   * This must not be a nullptr.
+   */
+  void SetProcessor(std::shared_ptr<LogProcessor> processor) noexcept;
+
+private:
+  // A pointer to the processor stored by this logger provider
+  opentelemetry::sdk::AtomicSharedPtr<LogProcessor> processor_;
+
+  // A vector of pointers to all the loggers that have been created
+  std::unordered_map<std::string, opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger>>
+      loggers_;
+
+  // A mutex that ensures only one thread is using the map of loggers
+  std::mutex mu_;
+};
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/include/opentelemetry/sdk/logs/processor.h
+++ b/sdk/include/opentelemetry/sdk/logs/processor.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <chrono>
+#include <memory>
+#include "opentelemetry/logs/log_record.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+/**
+ * This Log Processor is responsible for conversion of logs to exportable
+ * representation and passing them to exporters.
+ */
+class LogProcessor
+{
+public:
+  virtual ~LogProcessor() = default;
+
+  virtual void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept = 0;
+
+  virtual void ForceFlush(
+      std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept = 0;
+
+  virtual void Shutdown(
+      std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept = 0;
+};
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/src/logs/logger.cc
+++ b/sdk/src/logs/logger.cc
@@ -1,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "opentelemetry/sdk/logs/logger.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+Logger::Logger(std::shared_ptr<LoggerProvider> logger_provider) noexcept
+    : logger_provider_(logger_provider)
+{}
+
+void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
+{
+  // If this logger does not have a processor, no need to create a log record
+  auto processor = logger_provider_.lock()->GetProcessor();
+  if (processor == nullptr)
+  {
+    return;
+  }
+
+  // TODO: Sampler logic (should include check for minSeverity)
+
+  /**
+   * Convert the LogRecord to the heap first before sending to processor.
+   * TODO: Change the API log(LogRecord) function to log(*LogRecord) so the following line
+   * converting record a heap variable can be removed
+   */
+  auto record_pointer =
+      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+
+  // TODO: Do not want to overwrite user-set timestamp if there already is one -
+  // add a flag in the API to check if timestamp is set by user already before setting timestamp
+
+  // Inject timestamp if none is set
+  record_pointer->timestamp = core::SystemTimestamp(std::chrono::system_clock::now());
+  // TODO: inject traceid/spanid later
+
+  // Send the log record to the processor
+  processor->OnReceive(std::move(record_pointer));
+}
+
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/src/logs/logger_provider.cc
+++ b/sdk/src/logs/logger_provider.cc
@@ -1,0 +1,81 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "opentelemetry/sdk/logs/logger_provider.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+
+LoggerProvider::LoggerProvider() noexcept : processor_{nullptr} {}
+
+opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> LoggerProvider::GetLogger(
+    opentelemetry::nostd::string_view name,
+    opentelemetry::nostd::string_view options) noexcept
+{
+  // Ensure only one thread can read/write from the map of loggers
+  std::lock_guard<std::mutex> lock_guard{mu_};
+
+  // If a logger with a name "name" already exists, return it
+  auto loggerkv = loggers_.find(name.data());
+  if (loggerkv != loggers_.end())
+  {
+    return opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger>(loggerkv->second);
+  }
+
+  // Check if creating a new logger would exceed the max number of loggers
+  // TODO: Remove the noexcept from the API's and SDK's GetLogger(~)
+  /*
+  if (loggers_.size() > MAX_LOGGER_COUNT)
+  {
+#if __EXCEPTIONS
+    throw std::length_error("Number of loggers exceeds max count");
+#else
+    std::terminate();
+#endif
+  }
+  */
+
+  // If no logger with that name exists yet, create it and add it to the map of loggers
+
+  opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> logger(
+      new Logger(this->shared_from_this()));
+  loggers_[name.data()] = logger;
+  return logger;
+}
+
+opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> LoggerProvider::GetLogger(
+    opentelemetry::nostd::string_view name,
+    nostd::span<nostd::string_view> args) noexcept
+{
+  // Currently, no args support
+  return GetLogger(name);
+}
+
+std::shared_ptr<LogProcessor> LoggerProvider::GetProcessor() noexcept
+{
+  return processor_.load();
+}
+
+void LoggerProvider::SetProcessor(std::shared_ptr<LogProcessor> processor) noexcept
+{
+  processor_.store(processor);
+}
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/test/logs/logger_provider_sdk_test.cc
+++ b/sdk/test/logs/logger_provider_sdk_test.cc
@@ -1,0 +1,87 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "opentelemetry/logs/provider.h"
+#include "opentelemetry/nostd/shared_ptr.h"
+#include "opentelemetry/sdk/logs/logger.h"
+#include "opentelemetry/sdk/logs/logger_provider.h"
+
+#include <gtest/gtest.h>
+
+using namespace opentelemetry::sdk::logs;
+
+TEST(LoggerProviderSDK, PushToAPI)
+{
+  auto lp = opentelemetry::nostd::shared_ptr<opentelemetry::logs::LoggerProvider>(
+      new opentelemetry::sdk::logs::LoggerProvider());
+  opentelemetry::logs::Provider::SetLoggerProvider(lp);
+
+  // Check that the loggerprovider was correctly pushed into the API
+  ASSERT_EQ(lp, opentelemetry::logs::Provider::GetLoggerProvider());
+}
+
+TEST(LoggerProviderSDK, LoggerProviderGetLoggerSimple)
+{
+  auto lp = std::shared_ptr<opentelemetry::logs::LoggerProvider>(new LoggerProvider());
+
+  auto logger1 = lp->GetLogger("logger1");
+  auto logger2 = lp->GetLogger("logger2");
+
+  // Check that the logger is not nullptr
+  ASSERT_NE(logger1, nullptr);
+  ASSERT_NE(logger2, nullptr);
+
+  // Check that two loggers with different names aren't the same instance
+  ASSERT_NE(logger1, logger2);
+
+  // Check that two loggers with the same name are the same instance
+  auto logger3 = lp->GetLogger("logger1");
+  ASSERT_EQ(logger1, logger3);
+}
+
+TEST(LoggerProviderSDK, LoggerProviderLoggerArguments)
+{
+  // Currently, arguments are not supported by the loggers.
+  // TODO: Once the logging spec defines what arguments are allowed, add more
+  // detail to this test
+  auto lp = std::shared_ptr<opentelemetry::logs::LoggerProvider>(new LoggerProvider());
+
+  auto logger1 = lp->GetLogger("logger1", "");
+
+  // Check GetLogger(logger_name, args)
+  std::array<opentelemetry::nostd::string_view, 1> sv{"string"};
+  opentelemetry::nostd::span<opentelemetry::nostd::string_view> args{sv};
+  auto logger2 = lp->GetLogger("logger2", args);
+}
+
+class DummyProcessor : public LogProcessor
+{
+  void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept {}
+  void ForceFlush(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+  void Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+};
+
+TEST(LoggerProviderSDK, GetAndSetProcessor)
+{
+  // Create a LoggerProvider without a processor
+  LoggerProvider lp;
+  ASSERT_EQ(lp.GetProcessor(), nullptr);
+
+  // Create a new processor and check if it is pushed correctly
+  std::shared_ptr<LogProcessor> proc2 = std::shared_ptr<LogProcessor>(new DummyProcessor());
+  lp.SetProcessor(proc2);
+  ASSERT_EQ(proc2, lp.GetProcessor());
+}
--- a/sdk/test/logs/logger_sdk_test.cc
+++ b/sdk/test/logs/logger_sdk_test.cc
@@ -1,0 +1,71 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "opentelemetry/sdk/logs/logger.h"
+
+#include <gtest/gtest.h>
+
+using namespace opentelemetry::sdk::logs;
+
+TEST(LoggerSDK, LogToNullProcessor)
+{
+  // Confirm Logger::log() does not have undefined behavior
+  // even when there is no processor set
+  // since it calls Processor::OnReceive()
+
+  auto lp     = std::shared_ptr<opentelemetry::logs::LoggerProvider>(new LoggerProvider());
+  auto logger = lp->GetLogger("logger");
+
+  // Log a sample log record to a nullptr processor
+  opentelemetry::logs::LogRecord r;
+  r.name = "Test log";
+  logger->log(r);
+}
+
+class DummyProcessor : public LogProcessor
+{
+  void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept {}
+  void ForceFlush(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+  void Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+};
+
+TEST(LoggerSDK, LogToAProcessor)
+{
+  // Create an API LoggerProvider and logger
+  auto api_lp = std::shared_ptr<opentelemetry::logs::LoggerProvider>(new LoggerProvider());
+  auto logger = api_lp->GetLogger("logger");
+
+  // Cast the API LoggerProvider to an SDK Logger Provider and assert that it is still the same
+  // LoggerProvider by checking that getting a logger with the same name as the previously defined
+  // logger is the same instance
+  auto lp      = static_cast<LoggerProvider *>(api_lp.get());
+  auto logger2 = lp->GetLogger("logger");
+  ASSERT_EQ(logger, logger2);
+
+  // Set a processor for the LoggerProvider
+  std::shared_ptr<LogProcessor> processor = std::shared_ptr<LogProcessor>(new DummyProcessor());
+  lp->SetProcessor(processor);
+  ASSERT_EQ(processor, lp->GetProcessor());
+
+  // Should later introduce a way to assert that
+  // the logger's processor is the same as "proc"
+  // and that the logger's processor is the same as lp's processor
+
+  // Log a sample log record to the processor
+  opentelemetry::logs::LogRecord r;
+  r.name = "Test log";
+  logger->log(r);
+}
--- a/sdk/src/CMakeLists.txt
+++ b/sdk/src/CMakeLists.txt
@@ -1,3 +1,4 @@
 add_subdirectory(common)
 add_subdirectory(trace)
 add_subdirectory(metrics)
+add_subdirectory(logs)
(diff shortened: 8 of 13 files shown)


sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,94 @@
...
+namespace sdk
+{
+namespace logs
+{
+class Logger final : public opentelemetry::logs::Logger, public std::enable_shared_from_this<Logger>
+{
+public:
+  /**
+   * Initialize a new logger with a specified name.
+   * @param name The name of this logger.
```
reyang, line 39: Do we have this param?

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,94 @@
...
+   */
+  explicit Logger() noexcept;
+
+  /**
+   * Initialize a new logger with a specified name and processor.
+   * @param name The name of this logger.
+   * @param processor The name of the processor.
```
reyang, line 46: This is the processor instead of _the name of processor_?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,101 @@
...
+  opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> GetLogger(
+      opentelemetry::nostd::string_view name,
+      nostd::span<nostd::string_view> args) noexcept override;
+
+  /**
+   * Removes a logger with the given name. If no logger exists with that name,
+   * return false.
+   * @param name The name of the logger to be removed.
+   */
+  bool RemoveLogger(std::string name) noexcept;
```
reyang, line 76: Curious - why do we need this?

sdk/src/logs/logger_provider.cc
```diff
@@ -0,0 +1,93 @@
...
+  if (loggerkv != loggers_.end())
+  {
+    return opentelemetry::nostd::shared_ptr<Logger>(loggerkv->second);
+  }
+
+  // If no logger with that name exists yet,
+  // Create it and add it to the map of loggers
+  auto loggerInstance = new Logger(GetProcessor());
+  std::shared_ptr<Logger> loggerPtr{loggerInstance};
+  loggers_[name.data()] = loggerPtr;
```
reyang, line 46: Is this thread safe?

sdk/include/opentelemetry/sdk/logs/logger.h
maxgolov: I'm also curious what happened to the name parameter. In log4cxx - you'd have this:

```cpp
    static LoggerPtr getLogger(const std::string& name)
```

If we have this, we'd provide 1-1 mapping, easy-to-implement projection from log4cxx to OT logger. You can tell that the logger name is an identifier: either class name, or module name, or module / component tag.. It could be up to exporter to decide what to do with this name, but having it, in my opinion, is a good thing.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -0,0 +1,47 @@
...
+ * representation and passing them to exporters.
+ */
+class LogProcessor
+{
+public:
+  virtual ~LogProcessor() = default;
+
+  virtual void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept = 0;
+
+  virtual void ForceFlush(
```
maxgolov, line 39: My general comment, while we are at it: what is wrong with just `Flush`, and why name `ForceFlush` is better than `Flush`? There is no `ComfortablyOpen` on file descriptors, neither `QuicklyClose` or alike... in that sense I personally feel that word **Force** in a `Flush` - is a misnomer. And simply `Flush` would be much cooler. [fflush](https://man7.org/linux/man-pages/man3/fflush.3.html) - in Linux API is not `f` for force, but rather `f` for file.

sdk/src/logs/logger_provider.cc
```diff
@@ -0,0 +1,93 @@
...
+{
+namespace logs
+{
+LoggerProvider::LoggerProvider(std::shared_ptr<LogProcessor> processor) noexcept
+    : processor_{processor}
+{}
+
+LoggerProvider::LoggerProvider() noexcept : processor_{nullptr} {}
+
+opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> LoggerProvider::GetLogger(
```
maxgolov, line 30: If I read it correctly: we can allocate a named logger, but the logger object itself does not remember its name?

sdk/include/opentelemetry/sdk/logs/processor.h
xukaren: I can change it to just Flush() here. 
ForceFlush() was how `traces` did it as well - should I change [processor.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/2973baaf476a92e9d11d55e82f1dad78fefe6de1/sdk/include/opentelemetry/sdk/trace/processor.h) for traces as well?

reyang: The current one looks good as it aligns with the spec.

sdk/src/logs/logger_provider.cc
reyang: Yes, this is following the spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#get-a-tracer.

sdk/include/opentelemetry/sdk/logs/logger.h
MarkSeufert: Originally the logger constructor took in the name of the logger instance and it would be stored internally inside the logger. The `@param name` comment was left in by accident so we can remove that. 

The reason we removed the name parameter is that we chose to use a map inside the LoggerProvider to store the list of Loggers, with the key being the name. We used the logger name simply as a way to register the logger to the logger provider. If the logger also contained its name then it would waste memory, since it would be stored twice. Yea I agree with you Max that it could be useful for the logger to inject its name into the logs for contextual information. I'm thinking for now we leave it as is, and in the future re-add it if required?

sdk/include/opentelemetry/sdk/logs/logger.h
MarkSeufert: Yup good catch, I'll change that comment!

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: I was thinking that if a user creates a logger instance in a scope, then after user exits the scope they would lose the logger reference. The LoggerProvider would still have the logger registered though, so the user would call the RemoveLogger(name) method to deregister it. Hmm now that I type that it sounds like it would belong in the logger destructor such that it's automatically done. Would this work?

sdk/src/logs/logger_provider.cc
MarkSeufert: Yup the logger's name is stored in the logger provider's [map of loggers](https://github.com/open-telemetry/opentelemetry-cpp/blob/774ec7f5a970df18cfcba450730716a7a2f5c5f2/sdk/include/opentelemetry/sdk/logs/logger_provider.h#L97)

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include <unordered_map>
+#include <vector>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class Logger final : public opentelemetry::logs::Logger, public std::enable_shared_from_this<Logger>
```
ThomsonTan, line 34: nit: run clang-format?

MarkSeufert: I based this line from the [equivalent traces line](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/tracer.h#L17)

sdk/src/logs/logger_provider.cc
MarkSeufert: Right... if two threads simultaneously call GetLogger(name) then the map can be written to while the other thread is reading/writing from it. I'll fix this by using a spin lock, like what we did for the [API's provider](https://github.com/open-o11y/opentelemetry-cpp/blob/a2d883471ba422d7a3ddfcb435e794eb7958a5f3/api/include/opentelemetry/logs/provider.h#L52).

sdk/include/opentelemetry/sdk/logs/logger.h
maxgolov: I'm OK to keep as-is, just clean the comment. We can discuss later. This is not a show stopper.
You can mark this conversation as resolved.

sdk/include/opentelemetry/sdk/logs/processor.h
maxgolov: @reyang - can we change the spec? There is never a POSIX or STL library function or any other API for that matter that'd use an adjective before verb. Example of functions with similar semantics:
- fsync
- fflush
- `ostream::flush`
- (Java) `sync`
- (C#) `NetworkStream.Flush`

maxgolov: https://github.com/open-telemetry/opentelemetry-specification/issues/1206

sdk/src/logs/logger_provider.cc
maxgolov: @reyang - this doesn't appear to be a logging spec?

maxgolov: @reyang @MarkSeufert - can the logging provider still decorate events originating from logger with the logger name? Does it mean that you have to perform a search of logger in the map - what is the cost of that for each event emitted?

maxgolov: To further elaborate on this in pseudocode:
```cpp
auto logger1 = lp->GetLogger("Storage");
auto logger2 = lp->GetLogger("Network");
auto logger3 = lp->GetLogger("Incoming");
auto logger4 = lp->GetLogger("Outgoing");
```

in most logging systems I can associate the logger with some system, submodule, class or component, means eventually I need to carry that information to display to developer. So developer can search / grep by logger name.

How is it done in our case if the logger itself does not know its name?

sdk/include/opentelemetry/sdk/logs/processor.h
maxgolov: Not saying to fix in this PR. But while we are at it - designing / adding a new API, I think it's a good opportunity to revisit the naming to make it more consistent with practically any other existing framework.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
lalitb: Incorrect call to this function can lead to existence of dangling logger(s) - i.e., loggers removed from `LoggerProvider` map, but still used by application for logging.

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,75 @@
...
+  auto record_pointer =
+      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+
+  // Inject timestamp
+  record_pointer->timestamp = core::SystemTimestamp(std::chrono::system_clock::now());
+
+  // TODO: inject traceid/spanid later
+
+  // Send the log record to the processor
+  GetProcessor()->OnReceive(std::move(record_pointer));
```
lalitb, line 49: Can this lead to crash/seg-fault if Processor was not initialized while creating `LoggerProvider` instance ? Should we check for nullptr before using here or at `LoggerProvider` level.

sdk/src/logs/logger_provider.cc
```diff
@@ -0,0 +1,93 @@
...
+    opentelemetry::nostd::string_view name,
+    nostd::span<nostd::string_view> args) noexcept
+{
+  // Currently, no args support
+  return GetLogger(name);
+}
+
+bool LoggerProvider::RemoveLogger(std::string name) noexcept
+{
```
lalitb, line 59: If we are going to keep this function, please ensure this is thread safe too.

sdk/test/logs/logger_provider_sdk_test.cc
```diff
@@ -0,0 +1,73 @@
...
+TEST(LoggerProviderSDK, LoggerProviderGetLogger)
+{
+  LoggerProvider lp;
+
+  auto logger1 = lp.GetLogger("logger1");
+  auto logger2 = lp.GetLogger("logger2");
+
+  // Check that the logger is not nullptr
+  ASSERT_NE(logger1, nullptr);
+
```
lalitb, line 44: Please add test case what happens if we try calling Logger::log() here ( i.e, no processor initialized ).

sdk/src/logs/logger.cc
xukaren: Yes, will add a nullptr check

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: Inside the `RemoveLogger` method we first convert the Logger reference from an SDK Logger into a no-op logger before removing it from the `LoggerProvider` map. So in this exact case the user could still have a pointer to the logger, but calling its `log` method would no longer do anything.

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#include "opentelemetry/sdk/logs/logger.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+Logger::Logger() noexcept : processor_{nullptr} {}
```
xukaren, line 24: Are there any use cases for this constructor?
Since a Logger is always created by` LoggerProvider::GetLogger()`, which always calls the `Logger(processor)` constructor, should this constructor be removed, along with `Logger::GetProcessor()` and `Logger::SetProcessor()`?

sdk/src/logs/logger_provider.cc
reyang: > @reyang - this doesn't appear to be a logging spec?

Correct. The link above is for tracer.
The logger spec is still under construction, and based on the [data model](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md#field-name), there is no such thing called `logger name` from consumption perspective.
If there is a valid scenario, we could consider to add it later but not now, and we will need to work on the spec first.

sdk/include/opentelemetry/sdk/logs/processor.h
maxgolov: I'm resolving this. Let's merge your PR. We can discuss my concern in spec.

sdk/src/logs/logger_provider.cc
maxgolov: @reyang - how is it done in Java, while mapping log4j API to OpenTelemetry? log4j is the most prominent Java Logging API (as discussed in several community meetings). log4j follows the model having named loggers:
https://logging.apache.org/log4j/1.2/manual.html

In my opinion, it is absolutely necessary to have ability to name a logger.

maxgolov: Python cookbook - named loggers:
https://docs.python.org/3/howto/logging-cookbook.html

maxgolov: https://github.com/open-telemetry/opentelemetry-specification/issues/1215

sdk/include/opentelemetry/sdk/logs/logger_provider.h
ThomsonTan: Then the function doesn't sound like remove logger, probably `DisableLogger` is an better alternative?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,101 @@
...
+   */
+  std::shared_ptr<LogProcessor> GetProcessor() noexcept;
+
+  // Sets the common processor that all the Logger instances will use
+  /**
+   * Sets the processor that is stored internally in the logger provider.
+   * @param processor The processor to be stored inside the logger provider.
+   * This must not be a nullptr.
+   */
+  void SetProcessor(std::shared_ptr<LogProcessor> processor) noexcept;
```
ThomsonTan, line 90: Seems we cannot add multiple processors to one `LogProvider`?

MarkSeufert: Yea I based this off what was done in the .NET repo, where the LoggerProvider only has one processor but the processor can potentially be a [composite processor](https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry/CompositeProcessor.cs), which would contain any number of processors.

sdk/src/logs/logger_provider.cc
MarkSeufert: Yea I agree this is important functionality for a logging library. It's not included in the spec right now, but once it is included we can go back and add it to the SDK.

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,92 @@
...
+   * processor.
+   * @param sev The minimum severity
+   */
+  void SetMinSeverity(opentelemetry::logs::Severity sev) noexcept;
+
+  /**
+   * Returns whether a severity level is enabled or not.
+   * @param sev The severity level to be checked
+   */
+  bool IsEnabled(opentelemetry::logs::Severity sev);
```
pyohannes, line 67:
```suggestion
  bool IsEnabled(opentelemetry::logs::Severity sev) const noexcept;
```

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Returns a shared pointer to the processor currently stored in the
+   * logger provider. If no processor exists, returns a nullptr
+   */
+  std::shared_ptr<LogProcessor> GetProcessor() noexcept;
```
pyohannes, line 73:
```suggestion
  std::shared_ptr<LogProcessor> GetProcessor() const noexcept;
```

sdk/include/opentelemetry/sdk/logs/logger_provider.h
pyohannes: Not sure about this, but I think loggers once created (maybe based on some configuration) should stay available during the existence of the logger provider. `GetLogger` can called in different program parts (libraries), and removing the logger in one place might cause negative effects in other places.

I'd vote to remove this method, until we have a compelling use case to add it.

sdk/src/logs/logger.cc
pyohannes: I don't see a need for the default constructor. We should always initialize a logger and a logger provider with a processor.

The `SetProcessor` method should stay though, as one might possibly want to change the processor.

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,86 @@
...
+  auto processor = GetProcessor();
+  if (processor == nullptr)
+  {
+    return;
+  }
+
+  /**
+   * Convert the LogRecord to the heap first before sending to processor.
+   * TODO: Change the API log(LogRecord) function to log(*LogRecord) so the following line
+   * converting record a heap variable can be removed
```
pyohannes, line 47: We definitely need to do that. For any production usage, we should avoid this unnecessary copying of the complete log record.

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,86 @@
...
+   */
+  auto record_pointer =
+      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+
+  // Inject timestamp
+  record_pointer->timestamp = core::SystemTimestamp(std::chrono::system_clock::now());
```
pyohannes, line 53: I think we shouldn't generally overwrite that here, as the user (according to the API) has the ability to specify a custom timestamp.

sdk/src/logs/logger_provider.cc
pyohannes: Is it already specified what the exact behavior of this method should be, in relation to the passed logger name?

If not, maybe for now you can just do what the `TracerProvider` does and return a new `Logger`, regardless of the name?

With the current solution, I'm not sure about the benefit how having differently named loggers that all share the very same processor.

pyohannes (commented):
Thanks a lot for all your great work on this. I have a few question/suggestions/opinions below.

sdk/src/logs/logger.cc
MarkSeufert: Yes 100% agreed. We're thinking of adding a timestamp flag in the LogRecord class that says whether the timestamp was provided by the user or not. Or if adding a timestamp flag makes the LogRecord too messy we will give the timestamp a default value (like 0) and the SDK would check if it still has that default, then it would overwrite. We'll make these API changes in a later PR, but for now we are assuming the user doesn't provide a timestamp.

sdk/src/logs/logger_provider.cc
MarkSeufert: Yea right now having all the loggers reference the same processor is pretty much the same as having a single logger instance (like in traces). The logging spec isn't well defined right now so we took a hybrid approach of basing our implementation off traces while also trying to offer functionality that common logging libraries have, such as creating unique logger instance. When the logging spec gets more well defined we can change our implementation to match with it, for example having a unique processor for each logger.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,102 @@
...
+   * Initialize a new logger provider. A processor must later be assigned
+   * to this logger provider via the SetProcessor() method.
+   */
+  explicit LoggerProvider() noexcept;
+
+  /**
+   * Creates a logger with the given name, and returns a shared pointer to it.
+   * If a logger with that name already exists, return a shared pointer to it
+   * @param name The name of the logger to be created.
+   * @param options (OPTIONAL) The options for the logger. TODO: Give list of options
```
jsuereth, line 60: Nit: is it possible to tie this TODO to an issue open against the project?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,102 @@
...
+      opentelemetry::nostd::string_view name,
+      nostd::span<nostd::string_view> args) noexcept override;
+
+  /**
+   * Returns a shared pointer to the processor currently stored in the
+   * logger provider. If no processor exists, returns a nullptr
+   */
+  std::shared_ptr<LogProcessor> GetProcessor() noexcept;
+
+  // Sets the common processor that all the Logger instances will use
```
jsuereth, line 82: Question on these docs:

Does setting a new processor update ALL Logger instances that have already been created, or does it only change Logger instances that were created by this instance of LoggerProvider.

Additionally, is it worth denoting concurrency behavior here?

(i.e. this will lock acquiring new Loggers, but won't stop logs from being written due to the atomic-swap on the pointer).

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Convert the LogRecord to the heap first before sending to processor.
+   * TODO: Change the API log(LogRecord) function to log(*LogRecord) so the following line
+   * converting record a heap variable can be removed
+   */
+  auto record_pointer =
+      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+
+  // Inject timestamp
+  // TODO: add a flag in the API to check if timestamp is set by user already or not, then inject
```
jsuereth, line 55: Are these TODOs tracked anywhere for follow up outside of comments?

jsuereth (approved):
This looks like a great first PR for logging!

My high level comments:
- I saw TODOs about pulling LogRecord in via Heap.  I'd also love to see support for pulling from a ring/mempool or some other mechanism to limit overall memory consumption from log-processing.  We can discuss that in any follow-on PRs / SiG.  (This is a nice-to-have)
- I have some reservation around the shared map + mutex currently.  I saw the discussion and I think you're aware of the nuance you're in, I expect we'll want to throw some benchmarks around that code and tune it over time.   If we find `GetLogger` is frequently used in user-code we'll want to make sure that's as efficient as it can be (e.g. components which instantiate on every incoming HTTP Request means we're calling it often, and on many threads).    I think you should open a ticket around bench-marking and determining if we should try to migrate to RW lock.
- Tactically, how are you tracking TODOs and follow on work?  Is it feasible to link TODOs to issues on the project?   Two reason for this:
   1. It'd be nice to understand what you plan to tackle yourself in follow-on PRs
   2. It'd be nice for others in the community to see whether or not a TODO is actively worked on or if they can pick it up.

sdk/src/logs/logger.cc
xukaren: Most of the TODOs here, except the sampler, will be addressed by us (we hope) in a future PR. Since some of the PRs require changes to the Logging API, we were planning to PR a Logging API Update first then address some of the TODOs here

maxgolov:
We discussed the issue of propagating the `logger name` (or `log name`) from API to SDK to the protocol layer. @MarkSeufert @xukaren , correct me if I'm wrong on this one, it seems like Log SIG has generally been supportive of idea of NOT dropping the logger name, as the information flows from `API->SDK-Exporter`. I don't consider this blocking for your PR, but it would be _nice-to-have_ the ability to retain the name on logger object. That way you do not need to do any additional lookup when Log Record gets to exporter. There is no precise place decided, but my goal is - if I plug in a custom protocol exporter (not necessarily OTLP exporter), I should be able to easily fetch the `logger` or `log` name and populate it somewhere in my own exporter. [Log SIG notes](https://docs.google.com/document/d/1cX5fWXyWqVVzYHSFUymYUfWxUK5hT97gc23w595LmdM/edit#heading=h.lro8kgalys10). Recording of what's been discussed should be available on YouTube later today.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: Hey Josh, that's a good question. Currently when the SetProcessor() method is called, it will change all logger instances that were created by this LoggerProvider to use this new processor. However, since the LoggerProvider is expected to be a singleton (like what was done in [metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#obtaining-a-meter)) we don't have to worry about having different LoggerProviders making logger instances. 

This `SetProcessor(p)` functionality isn't solidified at all, and it will most likely be removed in a future PR. For example, if we'd like each logger instance to have a different processor, having a common processor defined in the LoggerProvider doesn't make much sense! Because of this future removal, we didn't focus on the concurrency issues that arise from this method.

Hope that answers your questions!

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: Right now the spec doesn't say anything about the different options/arguments that can be passed to the logger, so if we made an issue we wouldn't have enough detail for it. We'll update the comment to reflect that the TODO is waiting for the spec to give more detail.

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,87 @@
...
+   * @param record The record to write into the processor.
+   */
+  void log(const opentelemetry::logs::LogRecord &record) noexcept override;
+
+  /**
+   * Sets the minimum severity of a log in order to be sent to the
+   * processor.
+   * @param sev The minimum severity
+   */
+  void SetMinSeverity(opentelemetry::logs::Severity sev) noexcept;
```
reyang, line 56: This is not in the API, consider remove it since it might introduce problem with sampler.

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Returns whether a severity level is enabled or not.
+   * @param sev The severity level to be checked
+   */
+  bool IsEnabled(opentelemetry::logs::Severity sev) const noexcept;
```
reyang, line 62: This is not in the API, consider remove it since it might introduce problem with sampler.

xukaren: Sure, should we consider adding them back in after it's added to the API then?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
reyang: I think we probably would want this:
1. Logger will have a pointer to the LoggerProvider. Logger should not own processor.
2. LoggerProvider will own the processor(s).
3. Logger should always respect the current processor(s) on the LoggerProvider.
4. LoggerProvider can allow AddProcessor at any time, and it should take effect on all the existing loggers PLUS future loggers created from this provider.

reyang (changes requested):
1. There are several unnecessary methods from the Logger class, need to remove them (we can add them later if there is a need).
2. The processor ownership seems incorrect, need to move it to the provider.

sdk/include/opentelemetry/sdk/logs/logger.h
reyang: We could consider it in a separate PR, although I think we want to gather more perspectives - e.g. how would this interact with samplers.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: Sure I can make those changes, but I have a slight modification that could be pretty useful. 

The LoggerProvider would store all the processors that the loggers would use, but it would be stored inside a map with the name of the processor as the key. So for example, if there were two processors that export to different destinations, it could look like:
```
{
  {"ConsoleProcessor", shared_ptr<Processor>}
  {"ElasticSearchProcessor", shared_ptr<Processor>}
}
```

Then, when a logger is created via the GetLogger(), the user could specify which processor it will use by passing a second argument. So instead of `GetLogger(name)`, they would call `GetLogger(name, processor_name)`. By doing it this way, different loggers can have different processors.

If this way just complicates things then I can do it the way Reiley mentioned above.

reyang: I am very confused. There is no console processor or ElasticSearch processor. These should be exporters.
Processors have strict ordering.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,102 @@
...
+   */
+  explicit LoggerProvider() noexcept;
+
+  /**
+   * Creates a logger with the given name, and returns a shared pointer to it.
+   * If a logger with that name already exists, return a shared pointer to it
+   * @param name The name of the logger to be created.
+   * @param options (OPTIONAL) The options for the logger. TODO: Give list of options
+   */
+  opentelemetry::nostd::shared_ptr<opentelemetry::logs::Logger> GetLogger(
```
ThomsonTan, line 62: Could `GetOrCreateLogger` be a better name? Get implies it is very lightweight but actually it could create if the given named logger doesn't exist. As this is just `override`, we may need to change the API if take this.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: What I meant was that each processor has exactly one exporter attached to it, so the processor will indirectly have an export destination. So there could be, for example, a batch processor with a console exporter and another batch exporter with an ElasticSearch exporter.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
MarkSeufert: Yea GetLogger() is a bit misleading since doesn't only get loggers but can also create new loggers. Log4cxx and Log4j do it this way though so I figure we should stay as close to their implementation as possible.

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -0,0 +1,62 @@
...
+   * Writes a log record into the processor. The log record will only be
+   * sent if the severity level is greater than the minimum severity set by
+   * SetMinSeverity(s).
+   * @param record The record to write into the processor.
+   */
+  void log(const opentelemetry::logs::LogRecord &record) noexcept override;
+
+private:
+  // The minimum severity of this Logger
+  opentelemetry::logs::Severity minSeverity_;
```
reyang, line 54: Not needed?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,103 @@
...
+class LoggerProvider final : public opentelemetry::logs::LoggerProvider,
+                             public std::enable_shared_from_this<LoggerProvider>
+{
+public:
+  /**
+   * Initialize a new logger provider with a specified processor
+   * @param processor The processor for this logger provider. This must
+   * not be a nullptr.
+   */
+  explicit LoggerProvider(std::shared_ptr<LogProcessor> processor) noexcept;
```
reyang, line 49: Not needed?

sdk/include/opentelemetry/sdk/logs/logger.h
xukaren: No? The logger constructor currently sets the minSeverity and Logger::log() checks if the logrecord passed in meets the minSeverit or not. Should this be part of a future PR?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
xukaren: Is this not the expected way to initialize a logger provider (similar to tracer provider), or do you mean this should be added as part of simple/batch processor PR instead?

sdk/include/opentelemetry/sdk/logs/logger_provider.h
maxgolov: I think GetLogger is a good name, as it is in alignment with all popular log4* logging frameworks. I like this name.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
maxgolov: I think in future we can use some outside configurator class to map named logger to exporter, how this is done in log4j... I think I have seen a notion of configurator in some other OpenTelemetry SDK (python?)

sdk/include/opentelemetry/sdk/logs/logger_provider.h
reyang: I think this API should not exist, since the same functionality can be achieved by the following code. I think we want to keep the API orthogonal (e.g. if in the future we have sampler, we don't want to introduce a new `LoggerProvider.ctor(sampler)`).
```cpp
var provider = new LoggerProvider();
provider.AddProcessor(processor);
```

sdk/include/opentelemetry/sdk/logs/logger.h
reyang: Yes, I think this should be part of the sampler design/PR.

sdk/test/logs/logger_sdk_test.cc
```diff
@@ -0,0 +1,45 @@
...
+
+using namespace opentelemetry::sdk::logs;
+
+class DummyProcessor : public LogProcessor
+{
+  void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept {}
+  void ForceFlush(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+  void Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept {}
+};
+
```
lalitb, line 29: Would there be logger test cases with `DummyProcessor` ( in current or as part of separate PR). 
Would be good to have below tests:
```
logger.log(r)
logger.log(severity, message, time)
std::map attr= {{"k1", "v1"}, {"k2", "v2"}};
logger.log(severity, name, attr);
logger.log(severity, name, {{"k1", "v1"}, {"k2", "v2"}}};
```
The last one mayn't work, as it may need logger api change ( to support initialization list), but worth the change maybe in another PR.
```

File: sdk/include/opentelemetry/sdk/logs/logger_provider.h (hunk 24 above)
Comment (lalitb) [reply sdk/include/opentelemetry/sdk/logs/logger_provider.h:49]: Just ensure to test all corner cases ( eg trying to log when processor is not initialized , although I see one tests for it ).

File: sdk/test/logs/logger_sdk_test.cc (hunk 25 above)
Comment (xukaren) [reply sdk/test/logs/logger_sdk_test.cc:29]: Yes good catch - in my opinion these would be better as logging API tests? To follow [traces](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/test/trace/noop_test.cc) which does these in the API for example.

Comment (lalitb) [reply sdk/test/logs/logger_sdk_test.cc:29]: Right, these would make sense to have them in logging API.

File: sdk/include/opentelemetry/sdk/logs/logger_provider.h (hunk 18 above)
Comment (jsuereth) [reply sdk/include/opentelemetry/sdk/logs/logger_provider.h:76]: Yeah, my expectation (from other Logging API designs) is that the `Logger` class itself is just there to attach some semantics (the name of it) and fire log-events into a buffer for processing (the log-processor).  All routing of logs is done centrally with config.

We *don't* have to do that in OpenTelemetry, but if we don't I'd love to better understand why and what trade-off you're looking for.

Regarding singleton vs. separate instances of LoggerProvider, I think we should try to honor the tracing spec here w/ [TracerProvider](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider).  The TL;DR; 

- The LogProcessor "singleton" is used by default.  This allows APIs to add telemetry information that can be routed *later* when an Application-Developer wires + configured OT.    The reason this is needed is because the API-developer doesn't know where the telemetry (logs) SHOULD go, they just want to issue them.
- A second uses case of LogsProcessor is for applications where there is a strict need for separation of Telemetry (i.e. mutli-tenant).  In this case the application is expected to instantiate and share their own LogProcessors to ensure strict separation.

File: sdk/include/opentelemetry/sdk/logs/logger_provider.h (hunk 22 above)
Comment (jsuereth) [reply sdk/include/opentelemetry/sdk/logs/logger_provider.h:56]: Also matches slf4j.    I agree it's a misleading BUT industry standard name.

File: sdk/include/opentelemetry/sdk/logs/logger.h (hunk 23 above)
Comment (jsuereth) [reply sdk/include/opentelemetry/sdk/logs/logger.h]: Two things:

1. I agree this portion of the API should be deferred and discussed.
2. There (can be) a significant performance boost to filtering out logs at the Logger level.   Once we have a sampler design/PR *and* benchmark tests, we should re-evaluate minSeverity-in-logger.

File: sdk/src/logs/logger.cc (hunk 19 above)
Comment (MarkSeufert) [reply sdk/src/logs/logger.cc]: Yea most of these TODOs are from the original [Logging prototype issue](https://github.com/open-telemetry/opentelemetry-cpp/issues/337).

File: sdk/include/opentelemetry/sdk/logs/logger_provider.h (hunk 24 above)
Comment (MarkSeufert) [reply sdk/include/opentelemetry/sdk/logs/logger_provider.h]: We could do [what traces does](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L28) and, when a sampler is added to logging, also have it in the constructor but have a default value so it's an optional parameter?

File: sdk/include/opentelemetry/sdk/logs/logger.h [hunk 26]
```diff
@@ -0,0 +1,60 @@
...
+#include <vector>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class LoggerProvider;
+
+class Logger final : public opentelemetry::logs::Logger
```
ThomsonTan, line 37: Why is `public std::enable_shared_from_this<Logger>` removed?

xukaren: Is it needed? Since logger is not using `shared_from_this()` anywhere.

sdk/src/logs/logger.cc
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include "opentelemetry/sdk/logs/logger.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class LoggerProvider;
```
ThomsonTan, line 24: This `class LoggerProvider;` declaration is not needed because it is declared in `opentelemetry/sdk/logs/logger.h` which is included by this file.

xukaren: This is used so that the Logger(LoggerProvider) constructor is able to find LoggerProvider before it uses it. Is there another way around this?

MarkSeufert: I tried removing it but it doesn't compile without it. The reason its there is because the logger.h and logger_provider.h both include each other, so we need a forward declaration to tell the compiler that the class exists and will be defined later (or something like that :)).

ThomsonTan: Emm, are removing this line or the same declaration in `logger.h`? I just removed this line and produced build successfully. I meant this forward declaration is unnecessary because it was already forwarded declared [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/386/files#diff-66a32cfcf1db2f0ca92857788c2c37ad1b1ae6c4913e8b560e1e63f939b10b21R35)

xukaren: Thats right, thanks for the catch!

sdk/src/logs/logger_provider.cc
```diff
@@ -0,0 +1,80 @@
...
+    throw std::length_error("Number of loggers exceeds max count");
+#else
+    std::terminate();
+#endif
+  }
+  */
+
+  // If no logger with that name exists yet, create it and add it to the map of loggers
+  auto loggerInstance = new Logger(this->shared_from_this());
+  opentelemetry::nostd::shared_ptr<Logger> loggerPtr{loggerInstance};
```
ThomsonTan: Combine these 2 lines as `std::make_shared`?

sdk/test/logs/logger_provider_sdk_test.cc
```diff
@@ -0,0 +1,86 @@
...
+TEST(LoggerProviderSDK, LoggerProviderGetLoggerSimple)
+{
+  auto lp = std::shared_ptr<opentelemetry::logs::LoggerProvider>(new LoggerProvider());
+
+  auto logger1 = lp->GetLogger("logger1");
+  auto logger2 = lp->GetLogger("logger2");
+
+  // Check that the logger is not nullptr
+  ASSERT_NE(logger1, nullptr);
+
```
ThomsonTan, line 45: nit: add `ASSERT_NE(logger2, nullptr);`

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -0,0 +1,100 @@
...
+#define MAX_LOGGER_COUNT 100
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+class Logger;
+
+class LoggerProvider final : public opentelemetry::logs::LoggerProvider,
```
ThomsonTan, line 42: I remember we talked about make `LoggerProvider` a singleton, or do we need multiple log provider support?

xukaren: I believe a singleton, like traces

xukaren: I believe it was a singleton, like traces - but correct me if I'm wrong.

reyang: I don't think trace provider should be a singleton.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider

_Notwithstanding any global TracerProvider, some applications may want to or have to use multiple TracerProvider instances, e.g. to have different configuration (like SpanProcessors) for each (and consequently for the Tracers obtained from them), or because its easier with dependency injection frameworks. Thus, implementations of TracerProvider SHOULD allow creating an arbitrary number of TracerProvider instances._

xukaren: Oh I see- since the spec requires support for multiple TracerProviders, but traces doesn't have support this either yet, should we flag it an issue to update the implementation for traces? So that the specification and implementation is aligned for GA? Or should we implement it for logging first? If so would this PR be a good place to do that, or should we leave it as a later enhancement?

reyang: I think if you leave the PR as it is, it already has support for multiple providers.

xukaren: Oh I thought you meant to implement GetLoggerProvider() similar to GetLogger() (e.g. using a map to store them all). 

Is there anything else that is needed for this PR? Or is this ready to be merged?

reyang (approved):
LGTM.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
reyang: Looks good for now and we're good to merge. Would you rebase to the lastest master branch? Looks like I cannot do it due to some issue with GitHub:

image

PR #387, opened 2020-11-07
Update shared_ptr_test.cc

maxgolov:
`std::sort` and `std::reverse` used in this test both require `#include <algorithm>` [ref](https://en.cppreference.com/w/cpp/algorithm/sort)

However, this include was missing. How found it: I was experimenting with different version of Google Test and gcc-9.x with CMake build and hit the compilation issue. It appears that the header was previously included by the `gtest` header. But this is no longer the case if we upgrade to recent / latest version of Google Test.

Thus, fixing this by adding the proper missing include.

### DIFF
--- a/api/test/nostd/shared_ptr_test.cc
+++ b/api/test/nostd/shared_ptr_test.cc
@@ -1,5 +1,6 @@
 #include "opentelemetry/nostd/shared_ptr.h"
 
+#include <algorithm>
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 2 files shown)


maxgolov:
> EasyCLA Expected — Waiting for status to be reported

Stuck for some reason.

lalitb (approved):
LGTM

ThomsonTan (approved):

PR #388, opened 2020-11-11
Switch strerror to strerror_s

ThomsonTan:
`strerror` triggers  below warning when building by MSVC. Follow the
suggestion and switch to `strerror_s`.

warning C4996: 'strerror': This function or variable may be unsafe.
Consider using strerror_s instead. To disable deprecation, use
_CRT_SECURE_NO_WARNINGS. See online help for details.

### DIFF
--- a/examples/plugin/load/main.cc
+++ b/examples/plugin/load/main.cc
@@ -24,7 +24,9 @@ int main(int argc, char *argv[])
   std::ifstream config_in{argv[2]};
   if (!config_in.good())
   {
-    std::cerr << "Failed to open config file: " << std::strerror(errno) << "\n";
+    char err_msg[256] = {'\0',};
+    strerror_s(err_msg, sizeof(err_msg), errno);
+    std::cerr << "Failed to open config file: " << err_msg << "\n";
     return -1;
   }
   std::string config{std::istreambuf_iterator<char>{config_in}, std::istreambuf_iterator<char>{}};
--- a/examples/plugin/load/main.cc
+++ b/examples/plugin/load/main.cc
@@ -24,11 +24,7 @@ int main(int argc, char *argv[])
   std::ifstream config_in{argv[2]};
   if (!config_in.good())
   {
-    char err_msg[256] = {
-        '\0',
-    };
-    strerror_s(err_msg, sizeof(err_msg), errno);
-    std::cerr << "Failed to open config file: " << err_msg << "\n";
+    std::perror("Failed to open config file");
     return -1;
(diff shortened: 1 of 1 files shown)


maxgolov:
Unfortunately it won't work on POSIX, because no reasonable non-Windows compiler (clang, gcc) is implementing this Annex K. It is not treated favorably in POSIX world. Feedback from Principal Compiler engineer at RedHat on this:
http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1967.htm

I would personally suggest not doing it. We may introduce a helper function with `#ifdef _WIN32` that implements this, or see how we can handle it with this codebase (but it is not Apache License 2.0 , AFAIK):
https://github.com/microsoft/ChakraCore/tree/master/pal/src/safecrt

Basically, `Annex K` is not widely supported nor endorsed outside of msvc.

maxgolov:
Another great article from my ex-colleague written 2 days ago. This engineer is presently working on QNX kernel:
https://membarrier.wordpress.com/2020/11/08/stop-checking-for-null-pointers/

It explains why the `NULL` checking - implemented in many of the bounds-checking functions - is a waste. If `NULL` refers to protected region of memory, we need to let the apps immediately crash - let the OS handle it with `Access Violation` exception this rather than trying to make code `safe`. In general, `Annex K` is only giving an illusion of safety without actually providing one for the program flow end-to-end.

Not saying we should disrespect the warning. Probably `#ifdef _WIN32 /* use Annex K * #else /* use std lib */ #endif` is the best way to fix it.

examples/plugin/load/main.cc
```diff
@@ -24,7 +24,11 @@ int main(int argc, char *argv[])
   std::ifstream config_in{argv[2]};
   if (!config_in.good())
   {
-    std::cerr << "Failed to open config file: " << std::strerror(errno) << "\n";
+    char err_msg[256] = {
+        '\0',
+    };
+    strerror_s(err_msg, sizeof(err_msg), errno);
+    std::cerr << "Failed to open config file: " << err_msg << "\n";
```
pyohannes, line 31:
```suggestion
    std::perror("Failed to open config file");
```

This should be safe across platforms.

ThomsonTan:
@maxgolov thanks for raising the issue that `Annex K` is not supported in posix. `std::perror` suggested by @pyohannes seems working here.

pyohannes (approved):

lalitb (approved):
LGTM perror is the right approach.

PR #392, opened 2020-11-16
Documentation ( Installation, Quick Start), Changelog template , Release process

lalitb:
-  Updated README.md with notes on Installation and Quick Start.
-  Added  Changelog.md template file.
-  Added Release process. ( Similar to how it is done across other open-telemetry projects )

To summarize the release process as part of this PR:

1. CHANGELOG.md is managed manually, and as part of raising any PR, dev needs to add the summary of change under `Unreleased` section.
2. On day of release, Release manager decides on a tag say ( v0.0.1-beta).
3. He/She runs pre_release.sh which
     - creates a branch ( pre_release_v0.0.1-beta) 
     - modifies CHANGELOG.md ( Search for `##Unreleased` section and adds a new section with `<tag> - <date>`  header beneath it. That way `##Unreleased` section remains intact for new changes, and existing change-notes go within newly created section. I will reword the documentation.) 
4. He/She creates a PR for these changelog.md changes, and waits for it to be merged to master.
5.  Once PR is merged, he creates the above decided tag on the new merged commit hash.
6. He/She then goes to the github release UI page, and select "Create New Release". Specifies the tag created in step 5, and takes the curated contents as release notes, and creates the release.


pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+        *)
+            usage
+            ;;
+    esac
+done
+if [ ! -z "${t}" ]; then
+    usage
+fi
+
+#validate tag
```
ThomsonTan, line 19: nit: give some example on semver in comment?

pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+    exit -1
+fi
+
+#ensure tag doesn't exits
+if [[ $(git tag --list ${tag}) ]]; then
+   echo "Error: Tag ${tag} already exists. Exiting"
+   exit -1
+fi
+
+if ! git diff --quiet; then \
```
ThomsonTan, line 34: Does this mean `git diff --quiet` exists with `0` if there is any local change?

pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+    exit 1
+fi
+
+git checkout -b pre_release_${tag} master
+if [ $? -ne 0 ]; then
+    echo "Error: Cannot create release branch. Ensure you have sufficient permissions to repo and try again."
+    exit 1
+fi
+
+#update CHANGELOG.md
```
reyang, line 59:
```suggestion
# update CHANGELOG.md
```

reyang (approved):
LGTM with some minor formatting suggestions.

pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+#ensure tag doesn't exits
+if [[ $(git tag --list ${tag}) ]]; then
+   echo "Error: Tag ${tag} already exists. Exiting"
+   exit -1
+fi
+
+if ! git diff --quiet; then \
+	printf "Working tree is not clean, can't proceed with the release process\n"
```
ThomsonTan, line 35: nit: echo "Error: ..."?

pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+	git status
+	git diff
+	exit 1
+fi
+
+# check changelog exists
+changelog_file="./CHANGELOG.md"
```
maxgolov, line 42: How does manually managing `CHANGELOG.md` interact with built-in GitHub release management tab? Is intention to populate the markdown, then copy individual sections into Release tab?

README.md
```diff
@@ -20,12 +20,36 @@ Any exceptions to this are noted in the individual `README.md` files.
 
 ## Installation
 
-TBD
+The repository layout structure is as defined by [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/library-layout.md). The `api`
```
maxgolov, line 23: I think we need to decide on a few practical things. When you create a package, what is the actual installation / deployment layout:

- /usr/local/lib
- /usr/local/include/opentelemetry - for API?
- /usr/lcoal/include/opentelemetry/sdk - fo SDK developers?

# Why this needs to be described

Many packaging tools, e.g. `vcpkg` or `deb` would prefer you to have a clear-cut definition of what are the ship'able artifacts, even if include headers. Usually the layout for C++ projects requires a single collated directory. In our case it is two directories: one for customers `api` and another for devs, `sdk`... Are we shipping two separate packages, such as `opentelemetry` and `libopentelemetry-dev` - for developers?

README.md
```diff
@@ -20,12 +20,36 @@ Any exceptions to this are noted in the individual `README.md` files.
...
+The library would be distributed in source-only format. No shared/static library or distribution package (rpm, deb etc) would be made available as part of Release. See [RELEASING.md](./RELEASING.md).
+
+This project supports `Bazel` and `CMake` builds.
+
+### Using CMake
+On most platforms with all dependencies installed, the following command will compile the source files, and install the header files:
 
-## Quick Start
+```sh
+cmake -Bcmake-out
```
maxgolov, line 37: This may be a bit too generic. We need more detailed step-by-step instructions here, such as min cmake version, min compiler version, min required versions of Google Test. Just having generic `cmake` instruction is a good start. But may not necessarily be sufficient. For `ctest` - we can run this if built with tests as part of a single `cmake build` command.

RELEASING.md
```diff
@@ -0,0 +1,39 @@
+# Release Process
+
+## Pre Release
+
+1. Make sure all relevant changes for this release are included under `Unreleased` section in `CHANGELOG.md` and are in language that non-contributors to the project can understand.
+
+2. Run the pre-release script. It creates a branch `pre_release_<new-tag>` and updates `CHANGELOG.md` with the `<new-tag>`:
```
maxgolov, line 7: My biggest concern here, is right now you can easily author a release tag thru GitHub UI. For that - there is also new GH console utility, which can be used to automate all GH UI abilities, including creating a new tag + adding release notes. It may be unnecessary to have a separate script for that, if we say that we use the GitHub release process, documented here:

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository

That way even a person without console, e.g. Program Manager or Project Manager, can efficiently work on creating a new release quickly, without running any special tooling. I think it's a big advantage.

pre_release.sh
maxgolov: I think we're a little bit reinventing the wheel here, since most of this - including `semver` convention is documented here: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository - why not adopt this process?

maxgolov:
For managing release process - I'd suggest we use:
- GitHub UX for Release management: it covers creating tags in semver and creating release notes.
- for automated (nightly) - use the latest technology available: https://cli.github.com/ - that is `gh release create`: https://cli.github.com/manual/gh_release_create - which is gonna be harder for us to create something better than this in a limited time given.

Both options (GitHub Release tab and `gh` CLI) are industry standard. It is easy for any person, not necessarily dev, to use these tools.

README.md
```diff
@@ -20,12 +20,36 @@ Any exceptions to this are noted in the individual `README.md` files.
...
+cmake --build cmake-out
+sudo cmake --build cmake-out --target install
 
+#To run the tests
+cd cmake-out && ctest
```
maxgolov, line 42: To elaborate a little bit on this, ideally - the local dev build process could be as easy as:

- `tools/build.cmd` - on Windows, which will auto-detect the compiler AND setup all the build tools needed (unless you want to explicitly specify the compiler, for that we'll have `tools/build-vs201x.cmd` )

- `tools/build.sh` - for anything POSIX that only has `/bin/bash` : Linux, Mac, Android.

And

- crossplat `vcpkg install opentelemetry[:feature_pack]`

Maybe we can remove the cmake section here? And I can elaborate on a separate doc with the tooling I'm adding, as t's near-zero effort to onboard. I already added some drafts for `vcpkg` build process in the other PR.

lalitb: sure, I can remove the steps from here, and instead add one-liner saying Refer to (say) INSTALL.md for details.

RELEASING.md
lalitb: The  github UI is supposed to be used to publish a release as mentioned in final section of this document  (https://github.com/open-telemetry/opentelemetry-cpp/blob/c0ea1fd587e1be0d99a97d639a1f82994515000c/RELEASING.md#release). I can make it explicit if this is not clear in document. The idea behind the first two steps are to manage the changelog.md. I am fine removing the second step ( Tagging ), and let the tag be created automatically by github UI if that's the ask ?

lalitb:
> For managing release process - I'd suggest we use:
> 
> * [GitHub UX for Release management](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/managing-releases-in-a-repository) - it covers creating tags in semver and creating release notes.
> * for automated (nightly) - use the latest technology available: https://cli.github.com/ - that is `gh release create`: https://cli.github.com/manual/gh_release_create - which is great. It is gonna be harder for us to create something better than this in a limited time given.
> 
> Both options (GitHub Release tab and `gh` CLI) are industry standard. These are designed to be easy-to-use by any person, not necessarily dev. Plus, we'd have less of our own tooling to maintain.

The github UI is supposed to be used to publish a release as mentioned in final section of this document (https://github.com/open-telemetry/opentelemetry-cpp/blob/c0ea1fd587e1be0d99a97d639a1f82994515000c/RELEASING.md#release). I can make it explicit if this is not clear in document. The idea behind the first two steps are to manage the changelog.md. I am fine removing the second step ( Tagging ), and let the tag be created automatically by github UI if that's the ask ?

README.md
lalitb: Sure, we can reference to "INSTALL.md" document, as you plan to document it separately ?

pre_release.sh
lalitb: Please refer to my comment on managing changelog.md.

pre_release.sh
lalitb: This is checking the (standard ) output from the command, and not the return code (`$?`) of the command  So if there is any local changes, the command would print those changes. The return code would remain 0 irrespective of any local changes or not.

pre_release.sh
lalitb: Correct

README.md
lalitb: I am now confused here. As per my understanding, and also explicitly specified in this PR, we are only going to have source-only releases in same layout as the repo ( same as provided by github release ui by default). Which would mean, it would be responsibility of developers to decide how they would like to package this and distribute further - which should ideally be the structure as you mentioned above. Do you mean, even though we provide source-only releases, there should be recommendation for developer how to package them ?

maxgolov: In order to consume the package with tooling, like `vcpkg` (which builds the release for you from source and installs it for you), we need to identify the layout of how what we expect to get as a result of the build. In other words, if you build from source and invoke `make install` - what is the expected outcome?

README.md
maxgolov: I can cover this part. You can remove the build instructions for now.

RELEASING.md
maxgolov: We can probably skip the auto-tagging, and suggest that Release Manager uses their best judgement to tag the needed release sequentially, using semver 2.0, on that exact branch they wanna tag. We can keep the auto-tagging part if we want to cover tagging nightlies (but I am not sure we want this, as it'll create a ton of tags). One imporant moment where we may need to get the tag of the repo is to generate the `version.h` from the tag, not auto-increment the tag itself.

README.md
maxgolov: I can post the recommendations in `INSTALL.md`, to cover various  ways to build (or build from source and install) instructions. Separate sections for common build options:
- CMake
- Bazel
- vcpkg
- Visual Studio 2019 with CMake support (GUI build, no command-line tooling)
- command line wrappers to facilitate the build
- docker cross-plat builds

I'll send a separate PR with the doc. Most of it is already implemented in #377

maxgolov:
> let the tag be created automatically by github UI if that's the ask ?

Yeah, we can assume that Release Manager can use their best judgement and UI (no scripts) to generate the next matching tag, given UI totally allows to do this without running any command-line tooling ( or `gh` utility can be used by those who prefer command line tooling).

I think it's more valuable to write the opposite script: generate `version.h` from git tag, so that when the build is built - we properly report the actual SDK version.

README.md
lalitb: Ok, thanks for clarification. I will add expected deployment layout structure from packaging tools propspective in the document. 
As of now, we are installing api headers at `/usr/local/include/opentelemetry` for cmake builds. We can extended that as suggested above:

```
/usr/local/lib  -> For libraries
/usr/local/include/opentelemetry -> for api headers
/usr/local/include/opentelemetry/sdk -> for sdk headers
```

--
Are we shipping two separate packages, such as opentelemetry and libopentelemetry-dev - for developers?
 --
We are not shipping anything :) Although same recommendation on the final library/header deployment structure would apply for source only package like `libopentelemetry-dev` once it is `installed` and build from `/usr/local/src`.

README.md
lalitb: I have added reference to INSTALL.md under this section. I still feel it makes sense to keep generic installation steps as it given some idea to devs/users on how to build the library. I have already mentioned "with all dependencies specified". Regarding ctest, the default build using `cmake build` does build the tests , so these command should work out of box to run tests too.

Also have added the recommended deployment directory structure to be used by package manager. Let me know if this is ok.

pyohannes: We should add a _Requirements_ section, either here on in the `INSTALL.md`. I think the LightStep OpenTracing C++ project does a good job in describing requirements as well as detailed build instructions: https://github.com/lightstep/lightstep-tracer-cpp#requirements

README.md
pyohannes: It's just important that we have detailed CMake documentation somewhere, be it here on in the `INSTALL.md`. It should also cover the most useful build flags, like `WITH_OTLP`, `WITH_TESTS` and `WITH_EXAMPLES`.

RELEASING.md
lalitb: The steps to be followed as per this PR is
```
1. CHANGELOG.md is managed manually, and as part of raising any PR, the dev needs to add the summary of change under "Unreleased" section.
2. On day of release, Release manager decides on a tag say ( v0.0.1-beta).
3. He creates a branch ( pre_release_v0.0.1-beta) - this is done by running bash script.
4. He modified the CHANGELOG.md ( replacing Unreleased section with "<tag>-<date>", and adding empty "Unreleased" section on top for next release.) - Again, this is done by bash script
5. He creates a PR for these change log changes, and waits for it to be merged to master.
6.  Once PR is merged, he creates the above decided tag on the new merged commit hash.
7. He then goes to the github release UI page, and select "Create New Release". Specifies the already created tag in step 6, and takes the curated contents as release notes, and creates the release.
```
Please let me know what change we are expecting in these steps ? We can remove step 6 and let github release UI create the tag automatically, but wonder what will we achieve with that :)

README.md
```diff
@@ -20,12 +20,49 @@ Any exceptions to this are noted in the individual `README.md` files.
...
+```
 
-## Quick Start
+In case the library is distributed through package manager, below is the recommended deployment directory structure for header and librarie(s) on Linux and MacOS platforms.
 
+```
+/usr/local/lib:  For shared or static opentelemetry-cpp library
+/usr/local/include/opentelemetry: For api headers
+/usr/local/include/opentelemetry/sdk: For sdk headers
+/usr/local/src: For installing source package ( say libopentelemetry-dev )
```
pyohannes, line 51: I think this related to a previous discussion. However, I don't think we should put it here with this wording. We might need to document where `make install` puts our libraries, but we shouldn't recommend deployment structures for package managers.

Any proper DEB or RPM package will not install in `/usr/local`, but in `/usr/include` and `/usr/lib`. We should leave it to package maintainers to figure those things out (as I think we can't give valid general advice here).

pre_release.sh
lalitb: Thanks, Done :)

RELEASING.md
lalitb: > One imporant moment where we may need to get the tag of the repo is to generate the version.h from the tag, not auto-increment the tag itself.

I was thinking about this. Should we be having the release version in `version.h`. We do have `OPENTELEMETRY_VERSION` in `version.h` which seems to be used for dynamic loading of plugin libraries.

lalitb:
> > let the tag be created automatically by github UI if that's the ask ?
> 
> Yeah, we can assume that Release Manager can use their best judgement and UI (no scripts) to generate the next tag, given UI totally allows to do this without running any command-line tooling ( or `gh` utility can be used by those who prefer command line tooling ).
> 
> I think it's more valuable to do the following: generate `version.h` from git tag. So that when the build is built - we properly report the actual SDK version. We can probably re-generate `version.h` in git commit hook for incremental builds that includes `commit id` (or manage this in CI). But if we build from source based on tag - the `version.h` should reflect that tag information. That is the build tooling or CMake or bazel should re-generate the local `version.h` file (or mod it, if stored in source).. SDK must be able to know (and even in most cases, report) its version at runtime.

OK Just saw this comment. Let me go through this now :)

README.md
lalitb: Thanks Johannes. I do somewhat agree with your comment.  @maxgolov  your thought : )

README.md
```diff
@@ -42,7 +42,18 @@ sudo cmake --build cmake-out --target install
 cd cmake-out && ctest
 ```
 
-For more details on configuration and installation steps, please refer to [INSTALL.md](./INSTALL.md).
+In case the library is distributed through package manager, below is the recommended deployment directory structure for header and librarie(s) on Linux and MacOS platforms.
+
+```
+/usr/local/lib:  For shared or static opentelemetry-cpp library
```
maxgolov, line 48: I like it that way. We can discuss more and tune.

maxgolov (approved):

maxgolov:
>  Probably we can discuss this further in community meeting. I feel there are some gaps here.

We can describe the steps needed to create release / create tag with `gh`, and then publish the artifacts.... Maybe - as we publish the release using web-based UI, we can attach some form of a hook to automate the re-generation of the version header. Then even if Release Manager uses the release management tab / web-based UI to publish their notes, we can still properly patch the header.

In another project we, first:
- manually update the version header.
- check it in (thru PR review process).
- maybe at this point of PR review we should also review the release notes.
- then once the updated header is checked in -- we use web-based UX to create the tag + release notes under the release tab.

README.md
maxgolov: Ideally it should say `$INSTALL_PREFIX` instead of `/usr` - as this is either `/usr` or `/usr/local` . Maybe we should remove this and handle in a separate installation document, where we can describe the layout more precisely, depending on target OS and package manager: for rpm, deb, brew, vcpkg and others most common package managers. For CMake - we should also list what variables are populated by `find_package`, such as - include and library dirs of the opentelemetry package.

RELEASING.md
maxgolov: I think we absolutely must include the runtime detail (as in numeric and semver string version of the library) in the SDK header. This will be needed for exporters that must populate SDK version in the attributes, for analytics and for crash analysis / stability reasons.

maxgolov: There has to be two version concepts:
- version for API stability `VER_MAJOR.VER_MINOR`
- rest of semver 2.0 version for analytics and tracking; as there could be bugfix in SDK without a need to bump-up ver major and ver minor. 3rd tuple is the build increment, whereas the first two should be used to determine if the library has ABI-breaking changes.

maxgolov: e.g. no need to change the namespace version if no ABI breaking changes, but there is still a need to regenerate the build id information in the header, for runtime use.

README.md
pyohannes: > For CMake - we should also list what variables are populated by find_package, such as - include and library dirs of the opentelemetry package(s).

Do we intend to provide a CMake integration so the library can be discovered via `find_package`? We don't have this yet.

I'm fine with providing this directory list under the label of _default_ (CMake) locations, I'm a bit confused by labeling this as _recommended deployment directory structure_, as recommendations might vary considerable depending on the use case.

pre_release.sh
```diff
@@ -0,0 +1,71 @@
...
+    exit 1
+fi
+
+# update CHANGELOG.md
+date=$(date '+%Y-%m-%d')
+sed  -i "/\#\# \[Unreleased\]/a\\ \n\#\# \[${tag}\] ${date}"  $changelog_file
+if [$? -ne 0]; then
+    echo "Error: Cannot update CHANGELOG.md file. Update it manually, create the ${tag} and push changes to upstream"
+    exit 1
+fi
```
pyohannes, line 65: As far as I can see, this  doesn't add a new _Unreleased_ section, as claimed in the documentation.

README.md
```diff
@@ -20,12 +20,49 @@ Any exceptions to this are noted in the individual `README.md` files.
...
+For more details on configuration and installation, please refer to [INSTALL.md](./INSTALL.md).
+
+
+### Using Bazel
 TBD
 
+
+## QuickStart
+The `examples/simple` directory contains a minimal program demonstrating
+how to instrument a small library using a simple `processor` and console `exporter`, along with build files for CMake and Bazel.
```
pyohannes, line 64: Ideally, we'd have something like [this](https://github.com/open-telemetry/opentelemetry-java/blob/master/QUICKSTART.md) one day. But for now I think the reference to the example is fine.

README.md
```diff
@@ -20,12 +20,49 @@ Any exceptions to this are noted in the individual `README.md` files.
...
+/usr/local/include/opentelemetry/sdk: For sdk headers
+/usr/local/src: For installing source package ( say libopentelemetry-dev )
+
+```
+
+For more details on configuration and installation, please refer to [INSTALL.md](./INSTALL.md).
+
+
+### Using Bazel
 TBD
```
pyohannes, line 59: Instead of TBD, can we leave this section out and create an issue to add it?

RELEASING.md
ThomsonTan: Could we also append the latest commit hash to the version number and embed it into the build binary? This will help troubleshooting for developers.

maxgolov: I think we can have both - short semver 2.0 in there with three tuples, but also some sort of `OsBuild` or `BUILD_ID`, which will be the full semver 2.0 --- with commit hash after dash. In many ways it's easier / nicer to work with just X.Y.Z, but agree for remote crash analysis - in prebuilt SDK shipped in products, we definitely need to capture and track the full build id including hash.

README.md
maxgolov: @pyohannes - I can work on this for CMake.

README.md
maxgolov: I think we should just skip the install entirely and handle it outside of README, splitting it maybe in separate docs:
- build with bazel
- build with cmake
- build with vcpkg (in essence, package manager helper around cmake)
- build with Visual Studio 2019 (CMake, GUI)
- easy-to-use one-liner build scripts in tools ( `build.sh` and `builld.cmd` for those who don't want to read the docs)

maxgolov: I'm actually working right now to do these write-ups..

pre_release.sh
lalitb: This [line](https://github.com/lalitb/opentelemetry-cpp/blob/2fcd4b3068a5374f6958842c11878106429abb10/pre_release.sh#L61) actually does something similar:
```
sed  -i "/\#\# \[Unreleased\]/a\\ \n\#\# \[${tag}\] ${date}"  $changelog_file
``` 
It searches for `##Unreleased` section and adds a new section with `<tag> - <date>`  header beneath it. That way `##Unreleased` section remains intact for new changes, and existing change-notes go within newly created section. I will reword the documentation.

lalitb: reworded the documentation,

README.md
lalitb: Have removed the TBD.

README.md
lalitb: Have removed all the install information, and just kept a reference to ./INSTALL.md. Hope that is fine.

alolita:
Some items to consider for releasing artifacts:
* Add guidelines on commit comments
* Automate changelog generation
* Determine what features consititute major, minor, patch builds in line with semver.org guidelines
* Add PR builds, nightly builds and release builds
* Is there a need to have a compatibility matrix (e.g. ABI version compatibility)
Also to improve customer adoption:
* Make a sample app w the C++ library artifact in a Docker image available

RELEASING.md
jsuereth: QQ:  Does the tagged version ever make its way into a header file such that, e.g. we can use it in UserAgent strings when we write metrics?

jsuereth (approved):
under the assumption this is the *start* of the release process, looks great.

- I assume we'll be upadting the prepare_release script with more things we'd like.  Specifically I'd like runtime visibility into the release version.
- I like Max's suggestion of gh utilities for doing release.  The more automated the process can be the better.  However, I'd always prefer HAVING a process over not, so let's not let perfect be the enemy of good.

reyang:
There are still many open comments/discussion here, @lalitb please follow up in separate PRs. Thanks!

RELEASING.md
maxgolov: > Does the tagged version ever make its way into a header file such that, e.g. we can use it in UserAgent strings when we write metrics?

@jsuereth - Absolutely! My opinion we must have it and it could be used as such. In some future OpenTelemetry exporters, that my team would be working on, it is absolutely necessary to identify the version of the instrumentation library in some dedicated field. For both - user engagement analytics and crash analytics. Generated version info as both string and numeric constant, is a good starting point for enabling that. We also can reuse that for DLLs, like - there is a way to include a C header in a resource file, to be embedded into DLL on Windows. So when you right-click on a DLL - you see precisely what version it is. Usually just `X.Y.Z` but some projects include the full SemVer 2.0 with commit ID.

lalitb: Yes. I have a git issue to add version information into header file.

lalitb:
> Some items to consider for releasing artifacts:
> 
> * Add guidelines on commit comments
> * Automate changelog generation
> * Determine what features consititute major, minor, patch builds in line with semver.org guidelines
> * Add PR builds, nightly builds and release builds
> * Is there a need to have a compatibility matrix (e.g. ABI version compatibility)
>   Also to improve customer adoption:
> * Make a sample app w the C++ library artifact in a Docker image available

Really good points
 - This is incremental process, and most of these points are planned.

lalitb:
> under the assumption this is the _start_ of the release process, looks great.
> 
> * I assume we'll be upadting the prepare_release script with more things we'd like.  Specifically I'd like runtime visibility into the release version.
> * I like Max's suggestion of gh utilities for doing release.  The more automated the process can be the better.  However, I'd always prefer HAVING a process over not, so let's not let perfect be the enemy of good.

Runtime visibility of release version is planned, we have an open ticket for this.

PR #394, opened 2020-11-18
Add Shutdown method to TracerProvider

ThomsonTan:
Add `Shutdown` to `TracerProvider` but only for SDK, also add `nullptr` check against exporter member in `SpanProcessor` before calling `Shutdown`, because the constructor of `SpanProcessor` accepts null `SpanExporter`.


lalitb (approved):
LGTM once formatting is fixed.

sdk/include/opentelemetry/sdk/trace/simple_processor.h
```diff
@@ -57,12 +57,14 @@ class SimpleSpanProcessor : public SpanProcessor
   void Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override
   {
     // We only call shutdown ONCE.
-    if (!shutdown_latch_.test_and_set(std::memory_order_acquire))
+    if (exporter_ != nullptr && !shutdown_latch_.test_and_set(std::memory_order_acquire))
```
pyohannes, line 60: Currently we implicitly assume that `exporter_` is not `nullptr`. Maybe we should make that explicit in the doc string, instead of adding `nullptr` checks all around?

Because, to be thorough, we'd also to need all `nullptr` checks on the `Export` call in `OnEnd`, and around the `MakeRecordable` call.

ThomsonTan: Looked at this a little more and I think it is fine to only check `exporter_ != nullptr` in `Shutdown` call, because `Shutdown` is supposed to be called in destructor, and the span processor could be created like `new SimpleSpanProcessor(nullptr)`, we'd better destruct this object gracefully. For other function like `Export`, `OnEnd`, missing `exporter_` might mean an issue in the user code, so we could trigger and throw nullptr access exception.

Besides, `Shutdown` is declared as `noexcept` but `Export()` is not so it looks fine to throw exception there.

ThomsonTan:
From the above [chart](https://github.com/open-telemetry/opentelemetry-cpp/pull/394#issuecomment-729553738), the coverage for changed files in the PR is increased. The reduced coverage was caused by curl files. So it code coverage is not a blocker.

reyang:
The spec requires `TracerProvider.Shutdown` to provide a way to indicate whether the shutdown succeeded, failed or timed out.
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shutdown

Can be addressed in a follow up PR.

sdk/src/trace/batch_span_processor.cc
```diff
@@ -175,12 +175,14 @@ void BatchSpanProcessor::DrainQueue()
 
 void BatchSpanProcessor::Shutdown(std::chrono::microseconds timeout) noexcept
 {
-  is_shutdown_ = true;
+  is_shutdown_.store(true);
```
maxgolov, line 178: @ThomsonTan - I think we may want to add a safety check here, such as:

```cpp
if (!is_shutdown.exchange(true))
{
  cv_.notify_one();
  worker_thread_.join();
  if (exporter_ != nullptr)
  {
    exporter_->Shutdown();
  }
}
```

Otherwise we may run into C++ exception at `join()` with `system_error` in case if it's been already joined. Do we need to make it more forgiving?

PR #399, opened 2020-11-20
script to add versioning info

### LINKED ISSUE #397: Make version available for runtime query
Track follow-up for PR https://github.com/open-telemetry/opentelemetry-cpp/pull/392#issuecomment-730570978. As we discussed in the meeting, we need make the version available at runtime like for exporters to query SDK version.

lalitb:
This fixes #397 

This PR brings script for adding versioning information in `version.cc` in sdk. It can be used in following ways:

 * Executing as  pre-commit/pre-merge-commit git hooks - It won't be feasible to enforce this for every developer, unless we checkin it at `.git\hooks\pre-commit`. Not sure if this is good practice to enforce developers to run scripts locally. As of now developers need to create this link manually first time after fork.

* This can be added as github server-side [webooks](https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/about-webhooks), which would need more management as separate server need to be created to listen to github events. Ignoring this option for now.

* It is definitively required to run this script as part of release process, and this has been documented in RELEASING.md.

* If it is not important to checkin the modified `version.cc` file in repo, and only add it as part of release asset, the process can be automated through separate Github Release workflow which first runs this script, create release and subsequently upload release asset (https://github.com/actions/upload-release-asset). This way, only created releases will have the correct build information, and not the github repo, and we can automate release creation through github action. Earlier this year, github announced new feature to manually trigger github workflow from action tab in github UI (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) , which can be used to manually trigger Release creation from UI.


RELEASING.md
```diff
@@ -35,5 +35,19 @@ Failure to do so will leave things in a broken state.
     git push upstream
     ```
 
+## Versioning:
+
+Once tag is created, it's time to use that tag for Runtime Versioning
+
+1. Create a new brach for updating version information in `sdk/src/version.cc`
```
pyohannes, line 42: At what point do we update the version in [api/include/opentelemetry/version.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/version.h#L6)? Is this automated?

lalitb: version.h has extern declarations, so its not needed to modify it. Only version.cc is recreated.

pyohannes: I mean the `version.h` file in the API directory (not the SDK one). See the link above.

lalitb: Ok got it. Didn't check the link initially. Agree that needs to be changed too. I will do required changes.

lalitb: Have updated the script to modify api version.h.

RELEASING.md
```diff
@@ -35,5 +35,19 @@ Failure to do so will leave things in a broken state.
     git push upstream
     ```
 
+## Versioning:
+
+Once tag is created, it's time to use that tag for Runtime Versioning
```
jsuereth, line 40: I'm not sure I followed the previous discussion well, but one subtle question:

Is this bumping the submitted version for the NEXT round of iterative development, or is this updating the existing version fields for the CURRENT release?  If the later, shouldn't that be done prior to creating the tag?

jsuereth (approved):

RELEASING.md
pyohannes: I think @jsuereth is right, the version info should either be updated before creating the tag, or the tag should be updated after updating the version info.

RELEASING.md
```diff
@@ -5,8 +5,8 @@
 1. Make sure all relevant changes for this release are included under `Unreleased` section in `CHANGELOG.md` and are in language that non-contributors to the project can understand.
```
pyohannes, line 6: Can we add one more point to the checklist: we should also check if any of the changes made since the last release broke ABI compatibility. If so, [this number](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/version.h#L5) should be updated.

lalitb: Good point. Have updated this information. Also we should be updating major version ( in release tag) for any ABI/API change. I have added guidelines for that too in `CHANGELOG.md` ( slightly modified from semver guidelines to indicate ABI change)

RELEASING.md
lalitb: I have added the last step to move the tag to newly created commit-hash. Thanks for pointing out.

sdk/src/version/version.cc
```diff
@@ -0,0 +1,24 @@
+// Please DONOT touch this file.
+// Any changes done here would be overwritten by pre-commit git hook
```
ThomsonTan, line 2: Updating version via githook for each commit/merge might break forked repo? Is it better to just generated the version info and build it into the library in our build script, but don't commit it?

lalitb: Yes valid point. Things would be much simpler if we decide not to commit the version information. We can just call the `buildscripts/pre-commit` script created above as part of GitHub release action as I mentioned in the PR description. The problem with adding it as part of build script ( or cmake / bazel ) is that we won't be doing build as part of release  ( only source is released ). So we can create wrap release process through  github action(s) which would actually run the `buildscripts/pre-commit` first to create version.cc file and then release the source. 

Also agree on issues of adding it as pre-commit hook. It shouldn't be enforced. And the release manager should be doing this step ( i.e. copying `buildscripts/pre-commit` to `.git/hooks/pre-commit`) before doing release process. Anyway not something recommended. I will remove the comment about `pre-commit` hook from above with warning message that this file is updated automatically through script.

lalitb: Have updated the comment.

RELEASING.md
```diff
@@ -35,5 +35,30 @@ Failure to do so will leave things in a broken state.
...
+1. Create a new brach for updating version information in `./sdk/src/version.cc`.
+    ```
+        git checkout -b update_version_${tag} master
+    ```
+2. Run the pre-commit script to update the version:
+    ```console
+        ./buildscripts/pre-commit
+    ```
+
+3. Check if any changes masde since last release broke ABI compatibility. If yes, update `OPENTELEMETRY_ABI_VERSION_NO` in [version.h](./api/include/opentelemetry/version.h).
```
ThomsonTan, line 51:
```suggestion
3. Check if any changes made since last release broke ABI compatibility. If yes, update `OPENTELEMETRY_ABI_VERSION_NO` in [version.h](./api/include/opentelemetry/version.h).
```

ThomsonTan (approved):

pyohannes (approved):

PR #402, opened 2020-11-24
CURL support for Bazel

### LINKED ISSUE #390: bazel build for curl based http client implementation
Bazel build is missing for curl based http client implementation ( #385 ). 
From CMake prospective, the changes are minimal. As it's pure header based implementation for curl, we check for the presence of curl library at unit-test level, and compile and run these tests only if library is present.
Refer: https://github.com/lalitb/opentelemetry-cpp/blob/http_client_curl/ext/test/http/CMakeLists.txt 
Probably something similar needs to be done for Bazel.

jsuereth:
This wires CURL into the BAZEL build, and enables the CURL http tests to also be run in the BAZEL build.

Fixes #390 

Note:  This does NOT provide CURL build in a consumable way for release, that will be done as part of a larger PR for enabling OT CPP releases to be consumed by downstream bazel users.


third_party/curl.BUILD
```diff
@@ -0,0 +1,37 @@
...
+config_setting(
+    name = "osx",
+    values = {"cpu": "darwin"},
+    visibility = ["//visibility:private"],
+)
+
+cc_library(
+    name = "curl",
+    srcs = glob([
+        "lib/**/*.c",
```
ThomsonTan, line 22: `*.cc`?

jsuereth: CURL seems to use all `.c` files, see: https://github.com/curl/curl/tree/master/src

jsuereth:
@lalitb I got the build to work in bazel.  It looks like there's some kind of runtime error in the CURL test for "POST" for windows/osx.  I'm getting a non-zero exit code running the tests after I build locally.  I'm going to take a deeper look when I can, but if you have any insight, it'd be appreciated.

lalitb:
> @lalitb I got the build to work in bazel. It looks like there's some kind of runtime error in the CURL test for "POST" for windows/osx. I'm getting a non-zero exit code running the tests after I build locally. I'm going to take a deeper look when I can, but if you have any insight, it'd be appreciated.

@jsuereth  - I would instead suggest to remove the curl build/test for Windows, as we would like to have WinInet as the default http client for Windows. Either me/Max would work on this client. I will look into the osx failure, as it should work fine there.

jsuereth:
Disabling the unit test on windows is out-of-scope for this PR (I'm still working on that entire mechanism, trying to make sure we don't go down a path of madness).

@lalitb I did get the CURL test working on windows CI, just not locally.  I'll disable it in the future.  For the Darwin build, it *may* be a similar issue to windows where the CURL `-D` definitions (in bazel) are NOT saved in a `config.h` for later usage, so you need to define a subset of them to compile against it.

lalitb:
> Add comment on why this fixes the timeout test?

Sorry for not being clear enough in my earlier comment. The fix is to change the non-existent IP address value from "127.0.0.X" to some random address "222.222.X.X". The reason was that loopback address subnet range is different in Linux and MacOS and so timeout test works in one, and fails in another. The new IP address would not reachable in both the platforms, and so test will pass.

ThomsonTan (approved):
LGTM once the format issue is fixed.

lalitb (approved):
LGTM

Issue #2, opened 2019-08-15
Streaming SDK proposal


rnburn:
Hello Everyone,

I work on the OpenTracing C++ specification and the instrumentations for NGINX and Envoy. Recently, I've been working with LightStep to improve the performance of their C++ tracer. Our efforts to reduce instrumentation cost and improve collection throughput led us to a more efficient and streamlined design that I propose we adopt as one of the SDKs for OpenTelemetry. The key components of the design are 1) we remove intermediate storage on span objects and instead serialize eagerly as methods are called and 2) we use a domain-specific load balancing algorithm built upon non-blocking sockets, vectored-io, and io multiplexing.

Note: This only discusses the design as it releates to tracing. I plan to updated it to include metrics as well as the proposal progresses.

### Design
At a high level, the design looks like this
1. To each span we associate a chain of buffers that represent the span's protobuf serialization and we build up the serialization as methods on the span are called.
2. When a span is finished, its serialization chain gets moved into a lock-free multiple-producer, single-consumer circular buffer (or discarded if the buffer is full).
3. A separate recorder thread regularly flushes the serializations in the circular buffer and streams them to multiple endpoints using the HTTP/1.1 chunked encoding (one chunk per span) with domain-specific load balancing that works as follows:
    1. When the recorder flushes, it picks an available endpoint connection and makes a non-blocking [vectored-io write](https://linux.die.net/man/2/writev) system call to the connection's socket with the entire contents of the circular buffer.
    2. If the amount of data written is less than the size of the circular buffer, the recorder binds any remaining partially written span to the connection (to be written later when its socket is available for writing again) and repeates (i) with one of the other available connections.
    3. If no connections are available for writing, it blocks with [epoll](https://linux.die.net/man/2/writev) (or the target platform's equivalent) until one of the connections is available for writing again.

[Here's](https://github.com/lightstep/lightstep-tracer-cpp/blob/master/doc/diagram.png) a diagram of the architecture.

    
LightStep's C++ tracer provides an example implementation of this design. These are the main components
1. The [serialization chain](https://github.com/lightstep/lightstep-tracer-cpp/blob/9f690475084e1aa76ef477674d7a2d0102919d6c/src/common/serialization_chain.h) for a span's protobuf serialization.
[...]

PR #403, opened 2020-11-24
Add simple log processor and log exporter interface

xukaren:
This PR adds the simple log processor implementation and unit tests, as well as a log exporter interface for [#337](https://github.com/open-telemetry/opentelemetry-cpp/issues/337). The simple processor sends Log Records from the SDK directly to an exporter in a batch of one, and its design currently follows the [trace specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor).  The log exporter interface is used for the simple processor unit tests.  

Currently, the processor's Shutdown() function does not forcibly abort if the processor shutdown exceeds the max timeout. This functionality should be added in the future, and similarly for [traces' simple span processor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/simple_processor.h#L57) as well. 

--cc @alolita @MarkSeufert


sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+ * LogExporter defines the interface that log exporters must implement.
+ */
+class LogExporter
+{
+public:
+  virtual ~LogExporter() = default;
+
+  /**
+   * Exporters that implement this should typically format each LogRecord into the format
+   * required by the exporter destination (e.g. JSON), then send the LogRecord to the exporter.
```
reyang, line 50: Minor: the second part could be confusing, it sounded like "exporters send the LogRecord to the exporter".

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+   * The exporter may retry logs a maximum of 3 times before dropping and returning kFailure.
```
reyang, line 51: Probably remove `3 times`? It is the individual exporters' decision how they want to retry.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+   * TODO: This method should not block indefinitely. Should abort within timeout.
+   */
+  virtual ExportResult Export(
+      const std::vector<std::unique_ptr<opentelemetry::logs::LogRecord>> &records) noexcept = 0;
+
+  /**
+   * Marks the exporter as ShutDown and cleans up any resources as required.
+   * Shutdown should be called only once for each Exporter instance.
+   * @param timeout this method should not block indefinitely; should abort within timeout.
+   * @return a ShutDownResult code (if it succeeded, failed or timed out)
```
reyang, line 65: How about returning boolean value (true/false), and use false to represent failure/timeout?

This is what almost all other languages are doing:
* Java https://github.com/open-telemetry/opentelemetry-java/blob/69117e6dea5449077ff6c121fda1d60988d21819/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java#L63
* .NET https://github.com/open-telemetry/opentelemetry-dotnet/blob/b710c314ae041ad5db655cce4547f921b88e7495/src/OpenTelemetry/.publicApi/net461/PublicAPI.Unshipped.txt#L8

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -25,21 +25,44 @@ namespace sdk
 {
 namespace logs
 {
+enum class ShutdownResult
+{
+  kSuccess = 0,
```
ThomsonTan, line 30: Make the make more specific like `kShutdownSuccess`? `enum class ShutdownResult` and `enum class ExportResult` live in the same namespace, and define the same enum field like `kSuccess` may cause confuse.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -25,21 +25,44 @@ namespace sdk
...
+ * This Log Processor is responsible for the batching of log records
+ * and passing them to exporters.
  */
 class LogProcessor
 {
 public:
   virtual ~LogProcessor() = default;
 
+  /**
+   * OnReceive is responsible for batching every log record that is created by the SDK
```
reyang, line 44: Processor doesn't imply batching, please update the description.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -25,21 +25,44 @@ namespace sdk
...
+   * @param record a log record that has all the user data and injected data
```
reyang, line 45: It could be confusing. How about "the log record".

sdk/include/opentelemetry/sdk/logs/exporter.h
jsuereth: +1, I'd just say "the exporter may attempt to retry sending logs, but should drop and return `kFailure` after a timeout.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+   * required by the exporter destination (e.g. JSON), then send the LogRecord to the exporter.
+   * The exporter may retry logs a maximum of 3 times before dropping and returning kFailure.
+   * If this exporter is already shut down, should return kFailure.
+   * @param records: a vector of unique pointers to log records
+   * @returns an ExportResult code (whether export was success or failure)
+   *
+   * TODO: This method should not block indefinitely. Should abort within timeout.
+   */
+  virtual ExportResult Export(
+      const std::vector<std::unique_ptr<opentelemetry::logs::LogRecord>> &records) noexcept = 0;
```
jsuereth, line 59: Do we really want to pass-by-reference here?

1. Is it possible we want `LogRecord`s to go to multiple exporters?
2. Shouldn't the lifecycle of `LogRecord` destruction be fully owned by the `Exporter` at this point?

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Exporters that implement this should typically format each LogRecord into the format
+   * required by the exporter destination (e.g. JSON), then send the LogRecord to the exporter.
+   * The exporter may retry logs a maximum of 3 times before dropping and returning kFailure.
+   * If this exporter is already shut down, should return kFailure.
+   * @param records: a vector of unique pointers to log records
+   * @returns an ExportResult code (whether export was success or failure)
+   *
+   * TODO: This method should not block indefinitely. Should abort within timeout.
```
jsuereth, line 56: Can you comment on concurrency expectations of this?  Is it expected to block and only be called on a single thread?   Is it expected to be fully concurrent and thread-safe?

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,72 @@
...
+   */
+  virtual ExportResult Export(
+      const std::vector<std::unique_ptr<opentelemetry::logs::LogRecord>> &records) noexcept = 0;
+
+  /**
+   * Marks the exporter as ShutDown and cleans up any resources as required.
+   * Shutdown should be called only once for each Exporter instance.
+   * @param timeout this method should not block indefinitely; should abort within timeout.
```
jsuereth, line 64: nit: for description, you should describe the timeout itself.   E.g. "minimum amount of microseconds to wait for shutdown before giving up and returning failure."

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,83 @@
...
+{
+  std::vector<std::unique_ptr<opentelemetry::logs::LogRecord>> batch;
+  batch.emplace_back(std::move(record));
+
+  // Get lock to ensure Export() is never called concurrently
+  std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
+
+  if (exporter_->Export(batch) == ExportResult::kFailure)
+  {
+    /* TODO: alert user of the failed or timedout export result */
```
jsuereth, line 50: Should failure of the exporter be reported *by the exporter* or by the log processor?

I think the log processor should report its own failure and let the Exporter report its own failure as well.

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,83 @@
...
+void SimpleLogProcessor::OnReceive(
+    std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept
+{
+  std::vector<std::unique_ptr<opentelemetry::logs::LogRecord>> batch;
+  batch.emplace_back(std::move(record));
+
+  // Get lock to ensure Export() is never called concurrently
+  std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
+
+  if (exporter_->Export(batch) == ExportResult::kFailure)
```
ThomsonTan, line 48: Check the return value is not success instead of explicit failure, like `exporter_->Export(batch) != ExportResult::kSuccess)`, in case the more failure cases are added. We could do the same even boolean is returned.

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: Yes, it should only be called on a single thread, based on this [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#exportbatch) - "the `Export` method should only be called again after the current call returns". Aborting the method within a certain timeout should ensure it doesn't block indefinitely

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: I think either the processor or exporter could be responsible for destroying the log records, since there is only 1 configured exporter per processor. If the processor were to destroy it,  I think either having a pass by value or pass by reference here would be fine, since neither the processor nor exporter will be making changes to the log record. I'm leaning towards a pass by reference here to prevent unnecessary copying?

sdk/src/logs/simple_log_processor.cc
xukaren: I'm not sure about this, but these language repos seem to report the exporter's shutdown as part of the processor's shutdown result as well: 
* Dotnet: https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry/BatchExportProcessor.cs#L184
* Java: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L123

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: This [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#exportbatch) states that exporter should only be called on a single thread. , and the `Export` method should only be called again after the current call returns". So aborting the method within a certain timeout should ensure it doesn't block indefinitely. I don't think there should be any concurrency issues here.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -25,21 +25,44 @@ namespace sdk
...
+   */
   virtual void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept = 0;
 
-  virtual void ForceFlush(
+  /**
+   * Exports all log records that have not yet been exported to the configured Exporter.
+   * @param timeout that the forceflush is required to finish within.
+   * A default timeout of 0 mean no timeout is applied.
```
reyang, line 52: I don't have strong opinion here but would like to get more perspectives.

In .NET the idiomatic way of specifying **infinite** is to use `Timeout.Infinite` which is `-1`. Here goes what OpenTelemetry .NET did https://github.com/open-telemetry/opentelemetry-dotnet/blob/2785ffbbaa5487d54e4f3cefe1821fc18af86550/src/OpenTelemetry/BaseExporter.cs#L65.

I wonder what's the idiomatic way for C++. Should we treat 0 as "trigger the flush and return as quick as we can, don't wait at all", or "block indefinitely until the operation succeeded/failed"?

sdk/include/opentelemetry/sdk/logs/processor.h
reyang: Probably a boolean value would make it simpler?

sdk/include/opentelemetry/sdk/logs/processor.h
xukaren: I would agree with option # 2.

The current implementation currently follows [traces](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/simple_processor.h#L57), but I think we should do what is done in the Dotnet above, is actually better. We could use `std::chrono::microseconds::max()`, instead of `std::chrono::microseconds(0)` to indicate an "infinite" timeout. This returns the [largest duration possible](http://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/chrono/duration/max.html), so would not actually be "infinite" amount of time, so it should align with the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shutdown-2)'s requirement for `Shutdown` not to block indefinitely.

xukaren: I agree the default of 0 doesn't make a lot of sense, but I think neither does not enforcing a max timeout at all.

I am leaning towards a middle ground- I think we could use `std::chrono::microseconds::max()`, instead of `std::chrono::microseconds(0)` for the max timeout. This would return the [largest duration possible](http://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/chrono/duration/max.html), so would not actually be "infinite" amount of time so it would not ever block indefinitely as per the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shutdown-2). 

If that is in line with what you're thinking, I can make that change.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * ExportResult is returned as result of exporting a batch of Log Records.
+ */
+enum class ExportResult
+{
+  // The batch was exported successfully
+  kSuccess = 0,
+  // The batch was exported unsuccessfully and was dropped
+  kFailure
+};
```
lalitb, line 38: Nit: It there are multiple such enum definitions duplicated across trace and api sdk/api, better to have common definitions to be shared across. Not important to change as part of this PR, and also not important if there are just very  few such definitions.

sdk/include/opentelemetry/sdk/logs/processor.h
lalitb: No strong opinion either, but I think it is good to have a well defined default value how large it may be to ensure that we are not blocking indefinitely.

sdk/include/opentelemetry/sdk/logs/exporter.h
jsuereth: Right, I don't think you should make the users read the spec to know this class should only be used on a single thread.

I'm asking for you to add that to the documentation of the class so someone implementing it sees what they need to know to follow the spec.

sdk/include/opentelemetry/sdk/logs/exporter.h
jsuereth: I was suggesting we use a move here (`auto&&` vs `const auto&`).

Given this is expected to be used from a single thread, not important, was just curious what the lifecycle thought process was.

sdk/src/logs/simple_log_processor.cc
jsuereth: Gotcha.  The real issue here is that *theoretically* you are the logging framework, so how do you alert the user of failure if you're already failing to send out logs.

Note:  My meta-comment is I'm trying to understand if your TODO has a solution or not.  I think maybe it'd be useful to open a bug/ticket discussing the need to report Failures in the SDK in a consistent way that can work around issues in  Exporting Telemetry.   E.g. we don't want to *log* a failure if we are the logging framework and that failure would disappear because we are the ones failing.

I don't remember the specification having any specifics on how to handle failures of this type in a way that users can find + diagnose the problem.   I'm just trying to figure out how big this TODO is, and I think I'd love to have a bug open about it so we can start a (broader) discussion.

reyang: FYI - self diagnostics is briefly covered by the spec https://github.com/open-telemetry/opentelemetry-specification/blob/609a63784b2da7d59efe2a39822a23b09ae5b637/specification/error-handling.md#self-diagnostics.

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: Ah gotcha, will add that in!

sdk/src/logs/simple_log_processor.cc
reyang: In OpenTelemetry .NET, folks are using a simple custom logging API (which is not part of ILogger) and route data to a memory mapped file https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry/Internal/SelfDiagnostics.cs.

Another way of doing this is via ZPages, which is an experimental project though.

lalitb: Good point @jsuereth . This is something we need to define for our project, and needs broader discussion - either using callback mechanism or providing a ErrorHandler which can be used throughout the project.  I will create a ticket for this.

lalitb: Created #408. Let's discuss there. Thought this would be out of scope for this PR.

sdk/include/opentelemetry/sdk/logs/exporter.h
pyohannes: Any reason not to use a `nostd::span` instead of the `std::vector`? That's how the [exporter interface for the tracing SDK](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/exporter.h#L53) does it. I think we should aim for a consistent interface among the SDK parts.

I'm not sure about the `std::unique_ptr` though. @jsuereth rightly mentions that there might be several exporters (users can have several processors per logger), I think this will be a popular use case. I see a few possible solutions here:
1. Making copies of log records for each processor/exporter, and passing those as `std::unique_ptr`. This is possibly expensive and not ideal.
2. Keep one instance of each log record and pass it to each processor/exporter as a `std::shared_ptr`. Probably the most pragmatic solution in terms of implementation effort.
3. Following a similar approach to the tracing SDK, where a `Recordable` is used (see here for [the idea](https://github.com/open-telemetry/opentelemetry-cpp/issues/2)). Probably the most performant, but also the most implementation effort.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Marks the exporter as ShutDown and cleans up any resources as required.
+   * Shutdown should be called only once for each Exporter instance.
+   * @param timeout minimum amount of microseconds to wait for shutdown before giving up and
+   * returning failure.
+   * @return true if the exporter shutdown succeeded, false otherwise
+   */
+  virtual bool Shutdown(
+      std::chrono::microseconds timeout = std::chrono::microseconds::max()) noexcept = 0;
```
pyohannes, line 67: In the tracing SDK, we use `0` as default timeout, here we use `std::chrono::microseconds::max()`. What is your reasoning behind that?

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,82 @@
...
+{
+  return true;
+}
+
+/**
+ * TODO: This method should not block indefinitely. Should abort within timeout.
+ */
+bool SimpleLogProcessor::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  if (timeout < std::chrono::microseconds(0))
```
pyohannes, line 66: I think we can make this a pass-through parameter and let the exporter handle this case. No need to handle it in two places.

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,82 @@
...
+/**
+ *  The simple processor does not have any log records to flush so this method is not used
+ */
+bool SimpleLogProcessor::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  return true;
+}
+
+/**
+ * TODO: This method should not block indefinitely. Should abort within timeout.
```
pyohannes, line 62: You can remove this TODO. Timeouts should be handled by the exporter.

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: I wanted to set a defined timeout (and chose the max possible value for it) instead of allowing the shutdown to hang forever, as the spec prohibits `Forceflush` and `Shutdown` from blocking indefinitely

sdk/src/logs/simple_log_processor.cc
xukaren: Here in .NET, they have this similar check: https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry/BaseProcessor.cs#L121

xukaren: I have no strong opinion however, so I am happy to remove it here

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: But open to discussion on this, also related to this [comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/403#discussion_r529887038)

sdk/src/logs/simple_log_processor.cc
xukaren: My original concern was whether it was safe to make a user who writes their own exporter be responsible for this. I think if the processor `Shutdown` function didn't abort as a safety check, there _could_ be hanging - let me know whether you think this makes sense or not.

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: No reason, so I will change `vector` to `span`. Thanks for the catch.

The `unique_ptr` is used here to keep consistency with the fact that the current logging SDK uses `std::move` on the `LogRecord` from the Logger to the processor. The `std::move` was the original idea since the SDK initially only has support for [1 processor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/logs/logger_provider.cc#L70) so far.

In the next PR, we will be changing the Logger API/SDK to create a `shared_ptr LogRecord` instead of a plain `LogRecord` in the Logger API, which will then be passed to the Logger SDK, which can then be shared with all the processors/exporters. I think this would match suggested option # 2. 

I looked into the Recordable for traces, and based on what I've learned, I'm not sure it could be completed before our last day. Would it be possible to flag this as an issue to tackle in the future?

jsuereth: Let's discuss more on `shared_ptr` in the follow on PR (as you're not migrating it to shared_ptr here).  It'd be nice if we can have consistent memory-management performance here, and `shared_ptr` can create inconsistent performance.   There's a LOT of interactions and flow of data that we expect in a logging API and it'd be nice if we have a benchmark to evaluate realistic loads + performance characteristics to better understand the tradeoffs we're making here.   It sounds simple, but the runtime characteristics can change drastically in large systems.

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Batches the log record it receives in a batch of 1 and immediately sends it
+ * to the configured exporter
+ */
+void SimpleLogProcessor::OnReceive(
+    std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept
+{
+  nostd::span<std::unique_ptr<opentelemetry::logs::LogRecord>> batch(&record, 1);
+  // Get lock to ensure Export() is never called concurrently
+  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
```
jsuereth, line 44: Non Actionable Comment:

If you run into any deadlocks here on a single-core machine, it's because I gimped the SpinLock implementation.   I'll prioritize making that single-core friendly.

sdk/include/opentelemetry/sdk/logs/exporter.h
jsuereth: Also, for the record @pyohannes #3 is what I think we'll need to implement, and it is pretty sizeable.

I think you should flag it as an issue.  What you have is fine for prototyping and setting up benchmarks and filling out more of the project, but will need to be revisited.

You should open an issue and we should avoid spinning up too many exporters prior to resolving it, though, as that would increase the amount of code we'd need to refactor.

jsuereth (approved):
The changes as-is look like a great start with two caveats:

1. We should open an issue around how Processor => Exporter SDK interface, as we'll want to make sure we *can* implement something highly efficient with predicable performance implications.
2. We aren't planning to implement any complicated exporters prior to the first issue getting resolved

lalitb (approved):
LGTM in current state, assuming we have a follow up ticket to discuss 

 - Whether there are use-case to support multiple exporters through single processor

If answer is no, we probably need a PR to transfer the ownership of logrecord to (single) exporter. If yes, we would definitely need a PR to either share ownership through std::shared_ptr or  implement recordable interface as done for trace api.

xukaren:
> LGTM in current state, assuming we have a follow up ticket to discuss
> 
> * Whether there are use-case to support multiple exporters through single processor
> 
> If answer is no, we probably need a PR to transfer the ownership of logrecord to (single) exporter. If yes, we would definitely need a PR to either share ownership through std::shared_ptr or implement recordable interface as done for trace api.

I believe in logging, it is common to append multiple export destinations to a single Logger, or also called logging "sinks" or "appenders". For example [Log4cxx](http://svn.apache.org/repos/asf/incubator/log4cxx/tags/v0_9_7/xdocs/www/manual/Introduction.html#AppendersAndLayouts), [Plog](https://github.com/SergiusTheBest/plog#multiple-appenders), [Log4J](https://logging.apache.org/log4j/2.x/manual/appenders.html), amongst many.

xukaren:
> The changes as-is look like a great start with two caveats:
> 
> 1. We should open an issue around how Processor => Exporter SDK interface, as we'll want to make sure we _can_ implement something highly efficient with predicable performance implications.
> 2. We aren't planning to implement any complicated exporters prior to the first issue getting resolved

Flagged in #412

lalitb:
> I believe in logging, it is common to append multiple export destinations to a single Logger (AKA logging "sinks" or "appenders"). For example in these logging libraries, there is the ability to add multiple Appenders: [Log4cxx](http://svn.apache.org/repos/asf/incubator/log4cxx/tags/v0_9_7/xdocs/www/manual/Introduction.html#AppendersAndLayouts), [Plog](https://github.com/SergiusTheBest/plog#multiple-appenders), and [Log4J](https://logging.apache.org/log4j/2.x/manual/appenders.html). Would you say an issue should still be opened up for this?

I totally agree on supporting multiple destinations. Just thinking what should be right way to support them:
```
logger -> processor1 -> memoryExporter
       -> processor2 -> consoleExporter
       -> processor3 -> fileExpoter
```
or

```
logger -> processor   -> memoryExporter
                      -> consoleExporter
                      -> fileExporter
```
or 

mix of both.

If we follow the trace [SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.m), my interpretation is it talks about first approach ( eg, it mentions about `processors` configured).

Our implementation will change based on which approach we use. Again this is nothing to block current PR, but needs more discussion.

xukaren:
> I totally agree on supporting multiple destinations. Just thinking what should be right way to support them:
> 
> ```
> logger -> processor1 -> memoryExporter
>        -> processor2 -> consoleExporter
>        -> processor3 -> fileExpoter
> ```
> 
> or
> 
> ```
> logger -> processor   -> memoryExporter
>                       -> consoleExporter
>                       -> fileExporter
> ```
> 
> or
> 
> mix of both.
> 
> If we follow the trace [SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.m), my interpretation is it talks about first approach ( eg, it mentions about `processors` configured).

Yes, the trace spec refers to the first option, and is the one we followed. 

I can file an issue for this discussion soon after though :) 

> Our implementation will change based on which approach we use. Again this is nothing to block current PR, but needs more discussion.

Great, thank you! Should this PR be ready to merge then?

lalitb:
> Great, thank you! Should this PR be ready to merge then?

Yes, there are some comments by @pyohannes regarding handling timeout, we can briefly discuss with him in today's community meeting before merging.

sdk/include/opentelemetry/sdk/logs/simple_log_processor.h
```diff
@@ -0,0 +1,63 @@
...
+public:
+  explicit SimpleLogProcessor(std::unique_ptr<LogExporter> &&exporter);
+  virtual ~SimpleLogProcessor() = default;
+
+  void OnReceive(std::unique_ptr<opentelemetry::logs::LogRecord> &&record) noexcept override;
+
+  bool ForceFlush(
+      std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
+
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
```
cbandy, line 51: Should this be `std::chrono::microseconds::max()` to match `LogProcessor`?

ThomsonTan: Good question. This will be included in my coming PR which adds timeout support to `Shutdown`.

xukaren: I will make the change here for consistency too

xukaren: To avoid confusion after this PR is merged, also updated to `max()` here for consistency with `LogProcessor`

alolita:
@pyohannes can you please review? We're blocked on getting this merged. ty!

sdk/include/opentelemetry/sdk/logs/exporter.h
pyohannes: Then let's submit an issue so we can make this consistent in the different exporter interfaces. It's not intuitive to have `0` as default in the trace exporter and `max()` as default in the log exporter. We should settle on one default across exporters.

sdk/src/logs/simple_log_processor.cc
pyohannes: > But similar to below, would it be reliable to assume the exporter will be implementing this?

Yes, I would see this as the responsibility of the exporter. A negative might convey a special meaning for the exporter, so I think we should just pass it through as is.

pyohannes (commented):
Looks good so far. I saw you created an issue for the revisiting the implementation with a `Recordable` approach in mind - thanks.

Two things, in addition to my comments below:
* Please fix the formatting so the CI format job passes. Otherwise this will break other CI jobs after merging.
* Please submit issues for all TODOs and remove the TODOs.

sdk/src/logs/simple_log_processor.cc
```diff
@@ -0,0 +1,80 @@
...
+
+  if (exporter_->Export(batch) != ExportResult::kSuccess)
+  {
+    /* TODO: alert user of the failed or timedout export result */
```
xukaren, line 48: Should alert user of failed export result here - flagged in https://github.com/open-telemetry/opentelemetry-cpp/issues/408

sdk/src/logs/simple_log_processor.cc
xukaren: Done

sdk/src/logs/simple_log_processor.cc
xukaren: Done

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: Flagged in https://github.com/open-telemetry/opentelemetry-cpp/issues/418

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: Flagged in #412

xukaren:
> Looks good so far. I saw you created an issue for the revisiting the implementation with a `Recordable` approach in mind - thanks.
> 
> Two things, in addition to my comments below:
> 
> * Please fix the formatting so the CI format job passes. Otherwise this will break other CI jobs after merging.
> * Please submit issues for all TODOs and remove the TODOs.

@pyohannes thank you, removed TODOs and fixed formatting. 

Related issues: #408, #412, #417, and #418

sdk/include/opentelemetry/sdk/logs/processor.h
xukaren: Flagged in #418

sdk/include/opentelemetry/sdk/logs/exporter.h
xukaren: Followed up in #420

lalitb:
@xukaren  - Can you update this branch with base, so that it should be ready for merge.

PR #415, opened 2020-12-02
SpinLock Improvements

jsuereth:
Level-up the spinlock implementation.

- Use `std::atomic` instead of `std::atomic_flag` for the core
- Use an incremental back-off strategy instead of just a busy loop
   - First we busy-loop
   - Next we use `std::current_thread::yield()` in a loop (single-core cpu friendly)
   - Next we legitimately sleep.
- Adds a `try_lock` method (as previously requested) that does NOT do any spinning.

A few caveats in the current implementation:
- The constants for # of iterations or sleep ms are NOT optimised, and likely could use tuning.
- We do NOT issue a PAUSE/YIELD instruction during the busy loop, as we should.  I stuck to as many `std::` functions as I could to keep this simple to evaluate
- I'm sure someone better in C++ arcana could do loop unrolling or something to make this even better.  Please do!  This is just an incremental improvement over the initial version.


api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -32,22 +55,42 @@ class SpinLockMutex
...
+    }
+    return;
   }
   /** Releases the lock held by the execution agent. Throws no exceptions. */
-  void unlock() noexcept { flag_.clear(std::memory_order_release); }
+  void unlock() noexcept { flag_.store(false, std::memory_order_release); }
 
 private:
-  std::atomic_flag flag_{ATOMIC_FLAG_INIT};
+  std::atomic<bool> flag_{0};
```
ThomsonTan, line 93: `flag_{false}`?

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -23,6 +37,15 @@ class SpinLockMutex
   SpinLockMutex &operator=(const SpinLockMutex &) = delete;
   SpinLockMutex &operator=(const SpinLockMutex &) volatile = delete;
 
+  /**
+   * Attempts to lock the mutex.  Return immediately with `true` (success) or `false` (failure).
+   */
+  bool try_lock() noexcept
+  {
+    return !flag_.load(std::memory_order_relaxed) &&
```
ThomsonTan, line 45: Is this `.load` needed? Seems just on call to `.exchange` is sufficient for `try_lock`?

jsuereth: It's to help with cache miss issues.

See this blog which changed my mind on it: 
https://rigtorp.se/spinlock/

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -32,22 +55,42 @@ class SpinLockMutex
...
+      // Spin-Yield
+      for (std::size_t i = 0; i < SPINLOCK_YIELD_ITERATIONS; ++i)
+      {
+        if (try_lock())
+        {
+          return;
+        }
+        std::this_thread::yield();
+      }
+      // Sleep and then start the whole process again.
```
ThomsonTan, line 84: As there is an unconditional loop anyway, wondering why not do that in the last Spin-Yield instead of retrying Spin-Fast? Is there some benefit to do Spin-Fast after Spin-Yield timeout?

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -32,22 +55,42 @@ class SpinLockMutex
...
+        // e.g. __builtin_ia32_pause() / YieldProcessor() / _mm_pause();
+      }
+      // Spin-Yield
+      for (std::size_t i = 0; i < SPINLOCK_YIELD_ITERATIONS; ++i)
+      {
+        if (try_lock())
+        {
+          return;
+        }
+        std::this_thread::yield();
```
ThomsonTan, line 82: Would exponential back-off help here to avoid synchronized lock cadence?

lalitb (approved):
LGTM. 
Thanks for coming up with these changes in improving coherent cache misses, cpu contention and better performance on single core.  Would probably need benchmarking to come up with right values for the loop constants, and/or backoff logic ( exponential or incremental ) in subsequent iteration.

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -32,22 +55,42 @@ class SpinLockMutex
...
+      }
+      // Spin-Fast
+      for (std::size_t i = 0; i < SPINLOCK_FAST_ITERATIONS; ++i)
+      {
+        if (try_lock())
+        {
+          return;
+        }
+        // TODO: Issue PAUSE/YIELD instruction to reduce contention.
+        // e.g. __builtin_ia32_pause() / YieldProcessor() / _mm_pause();
```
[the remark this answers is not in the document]
jsuereth: Agreed on all points!

jsuereth:
> LGTM.
> Thanks for coming up with these changes in improving coherent cache misses, cpu contention and better performance on single core. We may need benchmarking to come up with right values for the loop constants, and/or backoff logic ( exponential or incremental ) in subsequent iteration.

Yes, I'd like to get a set of benchmarks going.  Have you seen the recent update to the specification around benchmarks?

api/include/opentelemetry/common/spin_lock_mutex.h
jsuereth: The idea is that (theoretically) spin fast is on the order of 10s of ns, while yield is 100s of NS and the full sleep is 1ms, i.e. each is an order of magnitude slower.  The goal is to "catch" the lock when you're scheduled and no need to give up to another thread if you only need to wait a few ns, not a big deal.

This theory could be wrong and we should validate with a benchmark the behavior we want.

api/include/opentelemetry/common/spin_lock_mutex.h
jsuereth: We could just do a single yield before sleep.   This lock assumes short contention times.


 Any exponential back off would be done via sleep timeout, rather than yield (which is meant to be between the spin-fast delay and an actual sleep delay, in wall clock time)

ThomsonTan: A single yield LGTM.

ThomsonTan (approved):

api/include/opentelemetry/common/spin_lock_mutex.h
jsuereth: done.

lalitb:
> Yes, I'd like to get a set of benchmarks going. Have you seen the recent update to the specification around benchmarks?

Not really, the only spec document I know around benchmark is the one written by @ThomsonTan 
 (https://github.com/open-telemetry/opentelemetry-specification/blob/0a2c727e2e6c7e20aecd107db42e60bd9fae79b1/specification/performance-benchmark.md), but not sure if any SIG team has come up with actual performance data around it.

PR #416, opened 2020-12-02
Bazel Build alterations

jsuereth:
This PR attempts to do the following:

- Abide by nested-workspace conventions in BAZEL community (see https://github.com/bazelbuild/bazel/issues/1943)
  _Note: This follows OpenCensusCpp implementation pretty closely_
- Update BAZEL to *also* use git submodules, keeping the build as close to CMAKE as possible.
  _Note: These submodules are NOT locked to version/tags, but really should be.  Also, lack of SHA verification has me concerned.  I'm not sure what the git submodule equivalent of this is._
- Update documentation to make sure developers know to init the submodules to build.

PR #421, opened 2020-12-04
Add -recursive to recommended git clone command line

ThomsonTan:
`-resursive` is available an option for git since 1.6.5 which clones all submodules. The current version of git is 2.28 or above so I think it is fine to expect `-recursive` for git. This would be very helpful as we added many dependencies as submodule to this repo.

Also add line-ending setting for `.cc` file which is missing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ requests (PRs).
 To create a new PR, fork the project in GitHub and clone the upstream repo:
 
 ```sh
-git clone https://github.com/open-telemetry/opentelemetry-cpp.git
+git clone --recursive https://github.com/open-telemetry/opentelemetry-cpp.git
 ```
 
 Add your fork as a remote:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ requests (PRs).
 To create a new PR, fork the project in GitHub and clone the upstream repo:
 
 ```sh
-git clone https://github.com/open-telemetry/opentelemetry-cpp.git
+git clone --recursive https://github.com/open-telemetry/opentelemetry-cpp.git
```
lalitb, line 67: nit: Just want to be sure that the changes done in COTRIBUTING.md as part of #416 are still relevant after this ?

ThomsonTan: https://github.com/open-telemetry/opentelemetry-cpp/pull/416 says "if you haven't, make sure loading submodules` so the commands are not required if submodules are available. I am fine with either keeping it or removing the submodule commands in #416.

lalitb (approved):

alolita:
Triaged Dec 7 2020: @jsuereth can you please review so that this can be merged?

pyohannes (approved):

jsuereth (approved):

PR #422, opened 2020-12-05
Supporting Logging API .Log() Overloads and GetName() method

MarkSeufert:
This PR adds `.Log()` overloads to the [Logging API](https://github.com/open-telemetry/opentelemetry-cpp/pull/378#partial-pull-merging), as well as a `GetName()` method. More specifically for the overloads, the following was added:
* There are now 6 Log overloads: `Log(Severity, message)`, `Log(Severity, name, message)`, `Log(Severity, KeyValue object)`, `Log(Severity, name, KeyValue object)`, `Log(Severity, initializer list)` and `Log(Severity, name, initializer list)`.
* Each of the above overloads has 6 additional methods that automatically set the severity for it: `.Trace()`, `.Info()`, `.Debug()`, `.Warn()`, `.Error()`, and `.Fatal()`. 

These methods make it easier for users to write simple logs using the API, since they won't have to populate the [9 argument logging statement](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/logs/logger.h#L71).

cc - @alolita @xukaren


api/include/opentelemetry/logs/log_record.h
```diff
@@ -82,25 +77,28 @@ struct LogRecord
...
+  trace::TraceFlags trace_flags;    // byte
+  Severity severity;                // Severity enum that combines severity_text and severity_number
 
   // other fields that will not be set by default
   nostd::string_view name;  // string
   nostd::string_view body;  // currently a simple string, but should be changed "Any" type
-  common::KeyValueIterable &resource;    // key/value pair list
-  common::KeyValueIterable &attributes;  // key/value pair list
+  nostd::shared_ptr<common::KeyValueIterable> resource;    // key/value pair list
+  nostd::shared_ptr<common::KeyValueIterable> attributes;  // key/value pair list
```
lalitb, line 87: Can't we avoid this shared ownership by overloaded Log constructor to accept reference to kv-iterable. 

btw, not related to this PR but this is where we would have benefited from `Recordable` interface - Instead of storing these attributes/resources here, we could have directly propagated them to exporters.

api/include/opentelemetry/logs/log_record.h
```diff
@@ -82,25 +77,28 @@ struct LogRecord
...
+  LogRecord()
   {
-    // TODO: in SDK, assign a default timestamp if not specified
-    name = "";
+    // Assign default values
+    timestamp   = core::SystemTimestamp(std::chrono::seconds(0));
+    severity    = Severity::kDefault;
+    trace_id    = trace::TraceId();
+    span_id     = trace::SpanId();
+    trace_flags = trace::TraceFlags();
```
lalitb, line 101: nit: these can also be initialized through initializer list.

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
...
     LogRecord r;
     r.severity   = severity;
     r.name       = name;
-    r.attributes = attributes;
+    r.attributes = nostd::shared_ptr<common::KeyValueIterable>(
+        new common::KeyValueIterableView<T>{attributes});
+
+    // Call the main Log(LogRecord) method
+    Log(r);
+  }
```
lalitb, line 138: It would be important to support the `log` function to take the attributes as part of function call.
Eg, 
```
logger.log(severtity.kTrace,  "test", {{"k1", "v1"}, {"k2","v2"}}
```
Refer to trace api ([Tracer::StartSpan](https://github.com/open-telemetry/opentelemetry-cpp/blob/11cc6cd8b18932c92e19bc92eb87aff1c7c5baad/api/include/opentelemetry/trace/tracer.h#L70) function ) on how this can be achieved.

lalitb, line 138: It would be important to support the `log` function to take the attributes as part of function call.
Eg, 
```
logger.log(severtity.kTrace,  "test", {{"k1", "v1"}, {"k2","v2"}}
```
Refer to trace api ([Tracer::StartSpan](https://github.com/open-telemetry/opentelemetry-cpp/blob/11cc6cd8b18932c92e19bc92eb87aff1c7c5baad/api/include/opentelemetry/trace/tracer.h#L70) function ) on how this can be achieved. 
This is also valid for the 6 severity-based functions.

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
...
-  inline void log(Severity severity       = Severity::kDefault,
-                  nostd::string_view name = "",
-                  const T &attributes     = {}) noexcept
+  inline void Log(Severity severity, const T &attributes) noexcept
   {
-    log(severity, name, common::KeyValueIterableView<T>(attributes));
+    // Create a LogRecord to hold this information
+    LogRecord r;
+    r.severity   = severity;
+    r.attributes = nostd::shared_ptr<common::KeyValueIterable>(
```
lalitb, line 112: Can we use `r.SetAttributes(attributes)` ?

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
...
+  inline void Log(Severity severity, nostd::string_view name, const T &attributes) noexcept
   {
-    // creates a LogRecord object with given parameters, then calls log(LogRecord)
+    // Create a LogRecord to hold this information
     LogRecord r;
     r.severity   = severity;
     r.name       = name;
-    r.attributes = attributes;
+    r.attributes = nostd::shared_ptr<common::KeyValueIterable>(
+        new common::KeyValueIterableView<T>{attributes});
```
lalitb, line 134: Can we use `r.SetAttributes(attributes)` ?

lalitb (changes requested):

api/include/opentelemetry/logs/log_record.h
MarkSeufert: I was thinking that since `LogRecord` is what is storing the information of the log from the API all the way to the Exporters, then having the log record have shared ownership over the resource and attributes would make sense here. If it's just a reference, and the initial attributes/resources goes out of scope and gets deallocated while the record is still in the processor/exporter, then that information would be lost which would be bad. Yea when the `Recordable` interface is implemented it would definitely simplify things.

api/include/opentelemetry/logs/logger.h
MarkSeufert: Yup that looks pretty handy for the user, I'll add that in!

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
    * @throws No exceptions under any circumstances.
    */
 
-  /* The below method is a logging statement that takes in a LogRecord.
-   * A default LogRecord that will be assigned if no parameters are passed to Logger's .log() method
-   * which should at minimum assign the trace_id, span_id, and timestamp
+  /**
+   * Logs a LogRecord, which contains all the fields of the Log Data Model. Normally called
+   * indirectly from other Log() Methods, but can be called directly for high detail.
```
ThomsonTan, line 59:
```suggestion
   * indirectly from other Log() methods, but can be called directly for high detail.
```

ThomsonTan: More elaboration on below comment?

> but can be called directly for high detail

api/include/opentelemetry/logs/log_record.h
jsuereth: One reason for the recordable interface is to ensure the Exporter can allocate the record.  This allows, e.g. an Exporter to pre-allocate a chunk of memory and ensure *all* logs fit within that limited amount of memory rather than contesting with the HEAP of the main application.

Two questions:

1. Should "resource" be mutable at all, or can we ensure there is one when creating the LogRecord?
  - If `LogRecord` has a resource attached later in the API than when it's created, I understand it would have to be a some kind of pointer.
  - Are Log API users able to modify the "Resource" object itself?  What's the use case for that?
2. Shouldn't the LogRecord own the "attributes"? (i.e. not be a shared pointer at all)?   When do you expect these to be shared?

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
...
-  inline void log(Severity severity,
-                  nostd::string_view name,
-                  const common::KeyValueIterable &attributes) noexcept
+  /**
+   * Writes a log.
+   * @param severity The severity of the log, from 1 to 24
+   * @param name The name of the log
+   * @param attributes A key/value object
+   */
+  template <class T,
```
ThomsonTan, line 125: nit: for all template class, prefer `template <typename T,`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,322 @@ class Logger
...
+  inline void Info(nostd::string_view name, const T &attributes) noexcept
+  {
+    Log(Severity::kInfo, name, attributes);
+  }
+
+  /**
+   * Writes a log with a severity of warn.
+   * @param message The message to log
+   */
+  inline void Warn(nostd::string_view message) noexcept { Log(Severity::kWarn, message); }
```
ThomsonTan, line 261: Seems name `Warning` is more practical than `Warn`?

api/include/opentelemetry/logs/log_record.h
ThomsonTan: Just initialize these fields along with its definition line.

api/include/opentelemetry/logs/noop.h
```diff
@@ -41,7 +41,8 @@ class NoopLogger final : public Logger
 public:
   NoopLogger() = default;
 
-  void log(const LogRecord &record) noexcept override {}
+  void Log(const LogRecord &record) noexcept override {}
+  const nostd::string_view GetName() noexcept override { return "noop logger"; };
```
ThomsonTan, line 45:
```suggestion
  const nostd::string_view GetName() noexcept override { return "noop logger"; }
```

api/include/opentelemetry/logs/logger.h
xukaren: `Warn` seems to be used by the [Log Data Model](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md#field-severitynumber), and looks like [Log4cxx](http://www.yolinux.com/TUTORIALS/Log4cxx.html) and [Log4J](https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm) also do it this way

api/test/logs/logger_test.cc
```diff
@@ -68,19 +66,68 @@ class TestProvider : public LoggerProvider
...
+  // "test logger" as defined in the custom implementation
   auto logger = lp->GetLogger("TestLogger");
+  ASSERT_EQ("test logger", logger->GetName());
+}
 
-  // GetLogger(name, args) function
-  std::array<string_view, 1> sv{"string"};
-  span<string_view> args{sv};
-  auto logger2 = lp->GetLogger("TestLogger2", args);
+TEST(Logger, LogMethodOverloads)
```
ThomsonTan, line 83: nit: inconsistent naming for `Logger`, should be `LoggerTest`?

api/include/opentelemetry/logs/logger.h
MarkSeufert: For this we were following what was done in tracer.h and its [StartSpan template classes](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/trace/tracer.h#L46).

api/include/opentelemetry/logs/log_record.h
xukaren: Hmm could we remove the constructor as a whole? Default values are set as follows by constructors in timestamp.h, trace_id.h, span_id.h, trace_flags.h, and string_view.h respectively. I believe shared_ptrs are set to nullptr by default (tested this locally). 

So we could have: 

```
struct LogRecord
{
  core::SystemTimestamp timestamp;    // default is 0
  trace::TraceId trace_id;            // default is 00000000000000000000000000000000
  trace::SpanId span_id;              // default is 0000000000000000
  trace::TraceFlags trace_flags;      // default is 00
  Severity severity;                  // default is 0 for enum initialization
  nostd::string_view name;            // default is length of 0 and data of nullptr
  nostd::string_view body;            // default is length of 0 and data of nullptr
  nostd::shared_ptr<common::KeyValueIterable> resource;    // default is nullptr
  nostd::shared_ptr<common::KeyValueIterable> attributes;  // default is nullptr 
}
```

Also if possible, we can also initialize severity to `kInfo` as default here, instead of having it injected in the SDK.

api/include/opentelemetry/logs/log_record.h
MarkSeufert: For the resources and attributes, it probably doesn't need to be mutable. When creating a LogRecord, I made it a shared_ptr such that the user wouldn't need to pass it into the constructor, but could set it afterwards like the other LogRecord fields (for consistency). 

And for your second question, the LogRecord should definitely own the attributes. Instead of changing the LogRecord/making a recordable interface, for now I could simply duplicate the memory from the attributes into the SDK heap, and the LogRecord's shared_ptr would point to that chunk. Since there would be no other shared_ptr to that object, it would technically "own" it as it would get deallocated when the LogRecord gets destructed. This isn't the most optimal approach, and in the future the LogRecord could directly contain the attributes and resources, but I think using shared_ptr would work to get the basic prototype out and running.

lalitb: > but I think using shared_ptr would work to get the basic prototype out and running.

I think it would be important to do it right at first go. Based on the answers of the questions put up by @jsuereth , it doesn't look that the shared_ptr is the right approach of doing this. We should either 
  - transfer the ownership of attributes/resources to LogRecord.
  -  Keep using references to these attributes/reference. What challenge you see here ?

api/include/opentelemetry/logs/log_record.h
lalitb: Yeah, that would be fine too.

ThomsonTan: If using the default generated ctor, `Severity severity` will be initialized to 0, which is actually not an valid value in `enum class Severity`. Perhaps we could add 0 to the enum list as `kInvalid`?

MarkSeufert: For sure, and later the SDK could inject the severity to something else like `kInfo` if it's `kInvalid`.

api/include/opentelemetry/logs/log_record.h
MarkSeufert: Yea sure, I'll convert it back to using references since its less overhead than using shared/unique ptr.

MarkSeufert: One issue I'm thinking of is if we do it this way without transferring the ownership, then the LogRecord is going to be dependant on the attributes/resource object within the user's code. If the user's object gets changed or deleted while the LogRecord is inside the processor or exporter, then this information is going to be lost.

We can solve this by doing a memcpy in the LogRecord so that it owns the memory of the attributes/resources, but it would also decrease performance... I'll leave it as reference for now but if anyone has any strong opinions about this I can take a look at changing it.

lalitb: > If the user's object gets changed or deleted while the LogRecord is inside the processor or exporter, then this information is going to be lost.

Correct, that can happen if User creates `LogRecord` object instance, deletes the attribute (or it goes out of scope), and then calls actual logger method.

lalitb: And if that is the case, would prefer to have the overhead of unique/shared ptr , than to see the actual application crashing. Would wait for others to comment cc @jsuereth @ThomsonTan @pyohannes

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,65 +54,321 @@ class Logger
...
   }
 
-  inline void log(Severity severity, nostd::string_view message) noexcept
+  /**
+   * Writes a log.
+   * @param severity The log's severity
+   * @param name The name of the log
+   * @param message The message to log
+   */
+  inline void Log(Severity severity, nostd::string_view name, nostd::string_view message) noexcept
```
pyohannes, line 89: Are there any plans to support specifying a format for the log message? Like this:
```c++
logger->Log(Severity::kError, "Log Name") << "Found " << num_cookies << " cookies";
```

Without a feature like this, the API might be pretty cumbersome to use.

api/include/opentelemetry/logs/logger.h
lalitb: Look the way we do in Trace::StartSpan overload. `initializer_list` overloded function calls `nostd::span` overloaded function which calls `const T*` overloaded function

MarkSeufert: Yea I took at look at that, but with traces the `initializer_list` immediately gets stored in memory in a recordable inside the SDK (after being converted to `nostd::span` -> `const T*`). In our case though, only the KeyValueIterable gets stored in the LogRecord, but the KeyValueIterable doesn't own the attributes/resources but just points to it, so the `initializer_list` has no memory to live at.

api/include/opentelemetry/logs/logger.h
MarkSeufert: Yea, there definitely are plans for this ([based from this comment](https://github.com/open-telemetry/opentelemetry-cpp/issues/356#issuecomment-707128159)), but due to time constraints we only have time to PR an MVP logging API. I'll open up an issue to support formatting log messages this after this PR gets merged.

xukaren:
The Recordable implementation for logs should simplify the overloads here. This PR will be coming in the next few days. Would be good to update/review again this PR after the Recordable is merged.

jsuereth (approved):
Fixes with Recordable look great!

lalitb (approved):
LGTM

api/include/opentelemetry/logs/log_record.h
```diff
@@ -107,7 +107,8 @@ struct LogRecord
             nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
   inline void SetResource(const T &_resource)
   {
-    resource = common::KeyValueIterableView<T>(_resource);
+    resource = nostd::shared_ptr<common::KeyValueIterable>(
+        new common::KeyValueIterableView<T>{_resource});
```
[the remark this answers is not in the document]
MarkSeufert: hmm I'm pretty sure the code you're looking at is outdated, since there's no longer a log record in the API haha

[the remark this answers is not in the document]
MarkSeufert: What's being done in the logger.h overloads now is it converts the attributes and resources into a `const KeyValueIterable &`, and it passes that straight to the SDK. So there's no longer any saving of memory inside the API and its all being done in the SDK's recordable.

lalitb:
@MarkSeufert Can you please rebase this branch with base, so that it is good to merge.

maxgolov (approved):

PR #423, opened 2020-12-07
Propagate span kind to exporter

lalitb:
Currently we initialize span-kind through [StartSpanOptions ](https://github.com/open-telemetry/opentelemetry-cpp/blob/11cc6cd8b18932c92e19bc92eb87aff1c7c5baad/api/include/opentelemetry/trace/span.h#L32 ) as part of span [creation](https://github.com/open-telemetry/opentelemetry-cpp/blob/11cc6cd8b18932c92e19bc92eb87aff1c7c5baad/api/include/opentelemetry/trace/tracer.h#L35), but this is not propagated to exporter through [Recordable](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/recordable.h) interface.  Adding the support.


exporters/otlp/src/recordable.cc
```diff
@@ -159,6 +159,39 @@ void Recordable::SetName(nostd::string_view name) noexcept
   span_.set_name(name.data(), name.size());
 }
 
+void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept
+{
+  opentelemetry::proto::trace::v1::Span_SpanKind proto_span_kind =
+      opentelemetry::proto::trace::v1::Span_SpanKind::Span_SpanKind_SPAN_KIND_UNSPECIFIED;
+
+  if (span_kind == opentelemetry::trace::SpanKind::kInternal)
```
jsuereth, line 167: Nit: Is there a reason to use if/else over switch here?

jsuereth (approved):

exporters/ostream/src/span_exporter.cc
```diff
@@ -10,6 +10,25 @@ namespace exporter
...
+    case opentelemetry::trace::SpanKind::kInternal:
+      return os << "Internal";
+    case opentelemetry::trace::SpanKind::kServer:
+      return os << "Server";
+    case opentelemetry::trace::SpanKind::kProducer:
+      return os << "Producer";
+    case opentelemetry::trace::SpanKind::kConsumer:
+      return os << "Consumer";
+  };
+  return os << "Internal";
```
ThomsonTan, line 29: nit: perhaps `assert(false)` to ensure `SpanKind` is exhausted by above switch?

ThomsonTan (approved):

alolita:
Triaged Dec 7 2020: Ready for merging.

pyohannes (approved):

exporters/otlp/src/recordable.cc
lalitb: No particular reason. Have switched to `switch` now.

exporters/ostream/src/span_exporter.cc
lalitb: Aborting mayn't be right option for library. I have changed it to return empty string instead. In-fact control should never reach to this line, as we are checking for all possible enum values above : )

Issue #408, opened 2020-11-26
Self-Diagnostics ( Error handling ) support


lalitb:
As of now, the project is missing a consistent way of  exposing errors / diagnostic information at sdk/exporter/processor level. We can use this Issue to discuss on whether/how to provide this. 

Specification briefly talks about it:
https://github.com/open-telemetry/opentelemetry-specification/blob/609a63784b2da7d59efe2a39822a23b09ae5b637/specification/error-handling.md#self-diagnostics

How different projects are handling it: 
.Net: https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry/Internal/SelfDiagnostics.cs
Go: https://github.com/open-telemetry/opentelemetry-go/blob/5a728db2e9713b66615943129568834c5af86ff8/handler.go

PR #430, opened 2020-12-08
Add Ostream log exporter and tests

xukaren:
This PR adds:
* A OStream Log Exporter 
* Unit tests and a full logging pipeline test  

cc @MarkSeufert @alolita


exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
```diff
@@ -0,0 +1,151 @@
...
+#include "opentelemetry/logs/log_record.h"
+#include "opentelemetry/nostd/type_traits.h"
+#include "opentelemetry/sdk/logs/exporter.h"
+#include "opentelemetry/version.h"
+
+#include <iostream>
+#include <map>
+#include <sstream>
+
+namespace nostd   = opentelemetry::nostd;
```
g-easy, line 12: I think we should avoid defining namespaces like this in a header.

exporters/ostream/src/log_exporter.cc
```diff
@@ -0,0 +1,92 @@
...
+    constexpr int trace_flags_bytes         = 1;
+    char trace_flags[trace_flags_bytes * 2] = {0};
+    record->trace_flags.ToLowerBase16(trace_flags);
+
+    /*** Print out each field of the log record ***/
+
+    // Print fields most useful to user first
+    sout_ << "{\n"
+          << "    timestamp   : " << record->timestamp.time_since_epoch().count() << "\n"
+          << "    severity    : " << severityNumToText[static_cast<int>(record->severity)] << "\n"
```
g-easy, line 43: What guarantees that record->severity won't be out of bounds?

MarkSeufert: I'm thinking that the only way this would be out of bounds is if the user manually sets the severity to a value larger than 24, which would be difficult for them to do since a [class enum](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/logs/log_record.h#L39) doesn't let the user implicitly cast from int.

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,69 +51,92 @@ class Logger
    * @throws No exceptions under any circumstances.
    */
 
-  /* The below method is a logging statement that takes in a LogRecord.
-   * A default LogRecord that will be assigned if no parameters are passed to Logger's .log() method
-   * which should at minimum assign the trace_id, span_id, and timestamp
+  /**
+   * Logs a LogRecord, which contains all the fields of the Log Data Model. Normally called
+   * indirectly from other log() Methods, but can be called directly for high detail.
```
ThomsonTan, line 56: I think we are going to rename the method name from `log()` to `Log()`. So make sure the new comment matches the new name.

xukaren: Once #422 is merged, it should ideally overwrite all the changes in this PR from `logger.h` and `log_record.h`. I included a subset of the changes from the other PR here so the CI would run properly. Currently since the log() method is named with a lowercase, I think it would be good to keep the comment consistent here.

xukaren: Currently since the log() method is named with a lowercase, I think it would be good to keep the comment consistent with the code. This change could be part of #422?

exporters/ostream/test/ostream_log_test.cc
```diff
@@ -0,0 +1,263 @@
+#include "opentelemetry/exporters/ostream/log_exporter.h"
+#include "opentelemetry/logs/provider.h"
+#include "opentelemetry/sdk/logs/logger_provider.h"
+#include "opentelemetry/sdk/logs/simple_log_processor.h"
+
+#include <gtest/gtest.h>
+#include <iostream>
+
+#include <gtest/gtest.h>
```
MarkSeufert, line 9: Remove the duplicated includes?

exporters/ostream/test/ostream_log_test.cc
```diff
@@ -0,0 +1,263 @@
...
+  ASSERT_EQ(stdcoutOutput.str(), expectedOutput);
+}
+
+// Print log to std::cerr
+TEST(OStreamLogExporter, PrintLogWithResourceToCerr)
+{
+  // Initialize an Ostream exporter to cerr
+  auto exporter = std::unique_ptr<sdklogs::LogExporter>(
+      new opentelemetry::exporter::logs::OStreamLogExporter(std::cerr));
+  auto processor =
```
MarkSeufert, line 106: nit: call the exporter's Export() directly to avoid having to create a processor instance?

xukaren: Thanks, great suggestion! I'll leave testing with a processor for the Integration Test below

xukaren:
@lalitb @g-easy @ThomsonTan Would love another review on this!

exporters/ostream/src/log_exporter.cc
```diff
@@ -0,0 +1,105 @@
...
+namespace sdklogs = opentelemetry::sdk::logs;
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace logs
+{
+OStreamLogExporter::OStreamLogExporter(std::ostream &sout) noexcept : sout_(sout) {}
+
+std::unique_ptr<sdklogs::Recordable> OStreamLogExporter::MakeRecordable() noexcept
```
xukaren, line 31: question: wondering if here it would make sense for `MakeRecordable()` should return a `nullptr` if the exporter is already shutdown?

lalitb: No strong preference, though it will allow us to fail early.

xukaren: Will make the change then if there's no other reason against it :)

lalitb (approved):
LGTM

exporters/ostream/src/log_exporter.cc
xukaren: Hmm on second thought, might be better to leave it as is actually, in case the caller is unaware of the reason `nullptr` is being thrown (e.g. exception or shutdown) -  might be later to be introduced consistently across traces/metrics/logs later - so I've left it unchanged :)

exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
```diff
@@ -0,0 +1,144 @@
...
+    nostd::visit(OwnedAttributeValueVisitor(*this), value);
+#else
+    nostd::visit([this](auto &&arg) { print_value(arg); }, value);
+#endif
+  }
+
+  void printMap(std::unordered_map<std::string, sdk::common::OwnedAttributeValue> map)
+  {
+    size_t size = map.size();
+    size_t i    = 1;
```
ThomsonTan, line 106: Do we need output `{}` to wrap the whole map, like `{{key: value}, {key1: value1}}`, or just as `{key: value, key1: value1}`?

exporters/ostream/src/log_exporter.cc
```diff
@@ -0,0 +1,106 @@
...
+
+  for (auto &record : records)
+  {
+    // Convert recordable to a LogRecord so that the getters of the LogRecord can be used
+    auto log_record =
+        std::unique_ptr<sdklogs::LogRecord>(static_cast<sdklogs::LogRecord *>(record.release()));
+
+    if (log_record == nullptr)
+    {
+      // Error "recordable data was lost"
```
ThomsonTan, line 52: Does this mean log an SDK internal error here? Perhaps add a TODO if so.

exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
xukaren: the outer `{}` around the map is done in the main `Export()` function [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/430/files#diff-5bab08f8a94686dae0568791dbdee4c097ead2922ae0c855436326dc52ccf145R79)

xukaren: This is the [expected output](https://github.com/open-telemetry/opentelemetry-cpp/pull/430/files#diff-30ea5c00a60dd17b4b14adefe2b4bee8d41888ef8afc74703862e9b3514d19f6R220-R221) I imagined

exporters/ostream/src/log_exporter.cc
xukaren: Yes, same as #408. Okay wrap with a `TODO`

exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
ThomsonTan: Ok, the caller outputs `{}` and assumed its content is map. Better to output `{}` in `printMap` in case it is called somewhere else.

xukaren: Good call, will make that change

exporters/ostream/src/log_exporter.cc
```diff
@@ -0,0 +1,106 @@
...
+      continue;
+    }
+
+    // Convert trace, spanid, traceflags into exportable representation
+    constexpr int trace_id_len    = 32;
+    constexpr int span_id__len    = 16;
+    constexpr int trace_flags_len = 2;
+
+    char trace_id[trace_id_len]       = {0};
```
ThomsonTan, line 61: Perhaps allocates 1 more char for tailing `\0`, so no need to conver to `std::string` for output.

exporters/ostream/src/log_exporter.cc
```diff
@@ -0,0 +1,106 @@
...
+
+    // Print out each field of the log record, noting that severity is separated
+    // into severity_num and severity_text
+    sout_ << "{\n"
+          << "  timestamp     : " << log_record->GetTimestamp().time_since_epoch().count() << "\n"
+          << "  severity_num  : " << static_cast<int>(log_record->GetSeverity()) << "\n"
+          << "  severity_text : "
+          << opentelemetry::logs::SeverityNumToText[static_cast<int>(log_record->GetSeverity())]
+          << "\n"
+          << "  name          : " << log_record->GetName() << "\n"
```
ThomsonTan, line 77: nit: `<< std::endl` instead of `<< "\n"`.

xukaren: Related- should traces' [span exporter](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/exporters/ostream/src/span_exporter.cc#L64) be updated as well? Or should we leave it here as `\n` to keept it consistent?

ThomsonTan (approved):

exporters/ostream/src/log_exporter.cc
ThomsonTan: Thanks for pointing it out, I think we could leave this to keep it consistent.

exporters/ostream/src/log_exporter.cc
xukaren: Haha sorry again, but maybe this optimization is also needed for [traces](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/exporters/ostream/src/span_exporter.cc#L65)?

exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
```diff
@@ -0,0 +1,144 @@
...
+  /*********************************** Helper functions ************************/
+
+  /*
+    print_value is used to print out the value of an attribute within a vector.
+    These values are held in a variant which makes the process of printing them much more
+    complicated.
+  */
+
+  template <typename T>
+  void print_value(const T &item)
```
jsuereth, line 53: if "print_value" and "OwndedValueVisitor" are helper functions, why not just put them into the `.cc` file?

Yes, you have to additionally pass sout_, but you'd still get your template-value-static-do-the-right-thing-with-my-type behavior you want, and we won't have a bunch of weak symbols to resolve later when linking *or* public methods that shouldn't be public.


Something like:

```
template <typename T>
void print_value(const T& item, std::ostream& sout) { ... }
template <typename T>
void print_value(const std::vector<T>& item, std::ostream& sout) { ...}
template <typename T>
void print_value(sdk::common::OwnedAttributeVisitor& item, std::ostream& sout) { ...}
```

jsuereth (changes requested):
Looks good, would just like to move helper methods into CC file or (at a minimum) make them private.

exporters/ostream/include/opentelemetry/exporters/ostream/log_exporter.h
xukaren: Mmk yeah, sorry, they were supposed to be private - they're now moved to the `.cc` file

jsuereth (approved):

PR #431, opened 2020-12-09
Add CMake gRPC support when OTLP is enabled

ThomsonTan:
Added gRPC to CMake build when OTLP is enabled (`WIHT_OTLP=ON`), also enabled building example_otlp when OTLP is on.

gRPC is enabled when OTLP is enabled without introducing a new flag `WITH_GRPC`.


cmake/opentelemetry-proto.cmake
```diff
@@ -1,48 +1,177 @@
...
-set(RESOURCE_PB_CPP_FILE "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/resource/v1/resource.pb.cc")
-set(RESOURCE_PB_H_FILE "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/resource/v1/resource.pb.h")
-set(TRACE_PB_CPP_FILE "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/trace/v1/trace.pb.cc")
-set(TRACE_PB_H_FILE "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/trace/v1/trace.pb.h")
+set(COMMON_PB_CPP_FILE
+    "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/common/v1/common.pb.cc")
+set(COMMON_PB_H_FILE
+    "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/common/v1/common.pb.h")
+set(COMMON_GRPC_PB_CPP_FILE
+    "${GENERATED_PROTOBUF_PATH}/opentelemetry/proto/common/v1/common.grpc.pb.cc"
```
lalitb, line 28: Just curious if `*.grpc.pb.[cc|h]` would be generated for common/resource/trace/logs as there is no `service` definition in their proto files.

lalitb (approved):
Thanks for the fix. LGTM :)

cmake/opentelemetry-proto.cmake
ThomsonTan: Good catch. `*.grpc.pb.[cc|h]` for non service are generated but almost empty and not used. Will remove them for GRPC code generation.

lalitb:
@jsuereth  pls review this file. As per the review policy, we need at-least one approver from different company. Thanks

jsuereth (approved):

Issue #412, opened 2020-12-01
Implementing a Recordable interface for Logs 


xukaren:
**Problem**
An important use case of logging is to allow log records to be sent to multiple export destinations, as seen in these logging libraries:
* [Log4cxx](http://svn.apache.org/repos/asf/incubator/log4cxx/tags/v0_9_7/xdocs/www/manual/Introduction.html#AppendersAndLayouts)
* [Plog](https://github.com/SergiusTheBest/plog#multiple-appenders)
* [Log4J](https://logging.apache.org/log4j/2.x/manual/appenders.html)

In the current Logging SDK, a LogRecord is created as a `unique_pointer` by the SDK, then moved to a processor, then passed by `const reference` to an exporter. This implementation does not allow more than one processor per Logger.

**Solution**
One solution could be to implement a `Recordable` interface, similar to what is currently used for [traces](https://github.com/open-telemetry/opentelemetry-cpp/blob/984e6ca3c311b00397559d8486972a8b23efd067/sdk/include/opentelemetry/sdk/trace/recordable.h). The recordable idea outlined [here](https://github.com/open-telemetry/opentelemetry-cpp/issues/2).

**Alternatives Considered**
Another possible (simpler but less performant) solution is to pass a `LogRecord` as a `shared_ptr` to be shared between the multiple processors/exporters. However, this approach could offer inconsistent performance related to memory-management which is a goal. 

**Additional Context**
Further discussion about the Recordable interface can be found in https://github.com/open-telemetry/opentelemetry-cpp/pull/44 and https://github.com/open-telemetry/opentelemetry-cpp/pull/49. 

cc @alolita @MarkSeufert

meastp:
@xukaren I don't know if you think it's relevant, as you already have a multiple examples, but Boost.Log is also designed around multiple sinks / export destinations as seen [in the Boost.Log design overview documentation](https://www.boost.org/doc/libs/1_74_0/libs/log/doc/html/log/design.html).

xukaren:
Resolved by #438

PR #434, opened 2020-12-09
Add batch log processor implementation with test coverage

xukaren:
This PR adds a batch log processor for the [Logging Prototype](#337), based off the [batch span processor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/batch_span_processor.cc), and following the [simple log processor](https://github.com/open-telemetry/opentelemetry-cpp/pull/403). The batch log processor has a main thread that is responsible for batching the `LogRecord`s, and a worker thread that sends these records to an exporter. Note: `std::cerr` messages are included here as a placeholder before [error diagnostics](https://github.com/open-telemetry/opentelemetry-cpp/issues/408) is available.

cc @MarkSeufert  @alolita


sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,225 @@
...
+          ptr.Swap(swap_ptr);
+          records_arr.push_back(std::unique_ptr<LogRecord>(swap_ptr.release()));
+          return true;
+        });
+      });
+
+  ExportResult export_status = exporter_->Export(records_arr);
+  if (export_status != ExportResult::kSuccess)
+  {
+    std::cerr << "[Batch Log Processor] Export failed \n";
```
lalitb, line 171: We don't have any error handling mechanism yet. Better to remove this error message, and update the existing github ticket #408 for handling this error,

sdk/test/logs/batch_log_processor_test.cc
```diff
@@ -0,0 +1,273 @@
...
+#include <thread>
+
+using namespace opentelemetry::sdk::logs;
+using opentelemetry::logs::LogRecord;
+
+/**
+ * A sample log exporter
+ * for testing the batch log processor
+ */
+class TestLogExporter final : public LogExporter
```
lalitb, line 31: nit: how about naming it as `MockLogExporter` ?

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,225 @@
...
+    Export(was_force_flush_called);
+
+    auto end      = std::chrono::steady_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+
+    // Subtract the duration of this export call from the next `timeout`.
+    timeout = schedule_delay_millis_ - duration;
+  }
+}
+
```
lalitb, line 141: We need to add `MakeRecordable()` function impl in this file once design is changed.

sdk/test/logs/batch_log_processor_test.cc
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * A fixture class for testing the BatchLogProcessor class that uses the TestExporter defined above.
+ */
+class BatchLogProcessorTest : public testing::Test  // ::testing::Test
+{
+public:
+  // returns a batch log processor that received a batch of log records, a shared pointer to a
+  // is_shutdown flag, and the processor configuration options (default if unspecified)
+  std::shared_ptr<LogProcessor> GetTestProcessor(
```
lalitb, line 83: nit: `GetMockProcessor`

lalitb (commented):
LGTM in general.  Few minor changes would be needed in this PR ( as mentioned in comments) once `Recordable` interface PR is reviewed and merged. As I understand, @xukaren  is working on that PR. We can hold this PR for merging till then.

sdk/include/opentelemetry/sdk/logs/batch_log_processor.h
```diff
@@ -0,0 +1,136 @@
...
+      std::unique_ptr<LogExporter> &&exporter,
+      const size_t max_queue_size                           = 2048,
+      const std::chrono::milliseconds schedule_delay_millis = std::chrono::milliseconds(5000),
+      const size_t max_export_batch_size                    = 512);
+
+  /**
+   * Called when the Logger's log method creates a log record
+   * @param record the log record
+   */
+
```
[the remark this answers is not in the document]
xukaren: Absolutely, thanks for flagging all the areas to change after implementing #412. Working towards a PR for this

sdk/src/logs/batch_log_processor.cc
xukaren: Sure - will make that change

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,224 @@
...
+  // sent to the worker thread to start a new export cycle.
+  if (buffer_.size() >= max_queue_size_ / 2)
+  {
+    // signal the worker thread
+    cv_.notify_one();
+  }
+}
+
+bool BatchLogProcessor::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
```
maxgolov, line 66: With 2 vars and atomic, and two while loops -- this code is overly complex and not easy to read. It appears you can rewrite the while loops with condvar `cv.wait_for` instead.

Also, in general, across all of this code - be careful about something like:
```cpp
             auto val = x.load();
             if (val)
```

Because this `load` then `compare` is not atomic. The state of what you read might have been changed by another thread. When you assign with `load` - you no longer compare the actual state, but some prior state you fetched. It's safer to use atomic compare-and-exchange or [atomic exchange](https://en.cppreference.com/w/cpp/atomic/atomic/exchange) where applicable.

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,224 @@
...
+  while (buffer_.empty() == false)
+  {
+    Export(false);
+  }
+}
+
+// Note: Timeout functionality is currently not implemented
+bool BatchLogProcessor::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  if (is_shutdown_.load() == true)
```
maxgolov, line 195: This is not atomic :) it would break. You just can't do it like this, no :)

You have to use [atomic exchange](https://en.cppreference.com/w/cpp/atomic/atomic/exchange).

Let me elaborate, as this is fundamental issue that might bite us hard.

- line 195 you do non-atomic compare
- line 200 you do non-atomic assignment
- imagine for a second that two threads A and B both called into `Shutdown` at the same time
- on multicore system you have zero guarantees about how the two threads are getting scheduled: you may race A and B executing A:195, B:195 - _at nearly the same time_ seeing `false` ; then threads A and B both assigning `is_shudown_` true, and both stuck.

You have to use atomic exchange to address this.

xukaren: Hmm I wasn't sure how the atomic swap-and-exchange would work here, since the `is_shutdown` variable would never need to be set to `false` again if it were already set to `true`. 

I added a lock instead that persists for the scope of this function. This will ensure the function doesn't get executed by more than 1 thread at once. 

Also, line 200, that used the assignment `operator =` originally, was in fact atomic (see[cpp ref](https://en.cppreference.com/w/cpp/atomic/atomic/operator=)), but I changed it to its equivalent `.store(true)` for clarity as well :)

xukaren: Also related is - I think this is a problem for traces as well. I can either file an issue for it or use the same implementation for traces as well in a later PR

sdk/src/logs/batch_log_processor.cc
xukaren: Sorry for the delay, I had to take a while to think this through - so

1. Yup, makes sense to change the `while` loops to  `cv.wait_for` with a predicate, for readability (done)

2. Another change I have made is changing all instances of `operator=` assignment for `atomic<bool>` variables to `.store(...)` for clarity (though they are actually equivalent and both atomic - see comment below). 

2. The three instances I found of non-atomic load/assign-then-compare (in the example that you listed) are in `batch_log_processor.cc`, and here's why I think they are working correctly in our case, although they're not atomic:
* Lines 110-113 in in `DoBackgroundWork()`: Although it is not atomic, there is only 1 thread calling the `DoBackgroundWork()` function at once, which is the worker thread. There won't be any more threads created, as the worker thread is only created/assigned once to `DoBackgroundWork()` in the Processor's constructor. (I have added a comment above the function to clarify this) 
* Line 169 in `Export()`: There is no other thread calling `Exporter::Export()` so the condition here should never change. 
* Lines 177-179 of `Export()`: I removed while loop around the cv notify (`force_flush_cv_.notify_one()`) altogether, because `is_force_flush_notified_` is only being set to either true/false in two places. First it's set to true by lines 177-188 in `Export()`, then the only other place where its value is being modified to `false` is in line 88, where it is _after_ being woken up by the cv's condition on it (in line 85). 

Let me know if I'm thinking about these correctly!

sdk/src/logs/batch_log_processor.cc
xukaren: 1. Maybe I'm misunderstanding the usage of the exchange operation- but I am not sure how the atomic swap-and-exchange/exchange would work here, since we would never want to mistakenly set the `is_shutdown` variable to `false` again if it were already set to `true`? 
* Instead, I added a lock that persists for the scope of this function, along with a global mutex. This will ensure the function doesn't get executed by more than 1 thread at once. Let me know if this works!

2. Also, in line 200 (now line 202): According to [cpp ref](https://en.cppreference.com/w/cpp/atomic/atomic/operator=), the assignment `operator =` is "Equivalent to `.store(...)`"  (but nonetheless, I've updated all occurences of it for clarity)

xukaren: Since this implementation is based off traces, let me know whether I should file an issue for it as well (or address it in a later PR if the changes are the same as the ones here)

sdk/src/logs/batch_log_processor.cc
xukaren: Sorry for the delay, I had to take a while to think this through - so

1. Yup, makes sense to change the `while` loops to  `cv.wait_for` with a predicate, for readability (done)

2. Another change I have made is changing all instances of `operator=` assignment for `atomic<bool>` variables to `.store(...)` for clarity (though they are actually equivalent and both atomic - see comment below). 

2. The three instances I found of non-atomic load/assign-then-compare (in the example that you listed) are in `batch_log_processor.cc`, and here's why I think they are working correctly in our case, although they're not atomic:
* Lines 110-113 in in `DoBackgroundWork()`: Although it is not atomic, there is only 1 thread calling the `DoBackgroundWork()` function at once, which is the worker thread. There won't be any more threads created, as the worker thread is only created/assigned _once_ to `DoBackgroundWork()` in the Processor's constructor. (I have added a comment above the function to clarify this) 
* Line 169 in `Export()`: There is no other thread calling `Exporter::Export()` so the condition here should never change. 
* Lines 177-179 of `Export()`: I removed while loop around the cv notify (`force_flush_cv_.notify_one()`) altogether, because `is_force_flush_notified_` is only being set to either true/false in two places. First it's set to true by lines 177-188 in `Export()`, then the only other place where its value is being modified to `false` is in line 88, where it is _after_ being woken up by the cv's condition on it (in line 85). 

Let me know if I'm thinking about these correctly!

sdk/src/logs/batch_log_processor.cc
maxgolov: You'd do this:

```cpp
if (!is_shutdown._exchange(true))
{
...
}
```

How this code works:
- it atomically reads and changes the value, guaranteed once, safe if multiple threads are racing thru the same code concurrently
- if your thread is lucky to toggle the state to `true`, it'll be the only one executing the shutdown code
- next thread would try to assign `true` again, but the value is already true - so assignment works, however the function returns that the previous value has already been `true`, so the code inside the scope won't get executed.

It is not about `.load()` and `.store()` being non-atomic: both of them are atomic. But they are separate operations. You can't use them separately, two of them are already non-atomic :) ... You'd use `exchange()` instead.

xukaren: Ohh makes sense, makes sense - I was looking at compare and swap instead of this get and set exchange, so got super confused. Thanks for explaining!

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,222 @@
...
+      return exporter_->Shutdown();
+    }
+    return true;
+  }
+  return false;
+}
+
+BatchLogProcessor::~BatchLogProcessor()
+{
+  if (is_shutdown_.load() == false)
```
jsuereth, line 214: Nit: Just call shutdown here.  You're guarding calling shutdown twice anyway.

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,222 @@
...
+    // wait until worker thread completes current export
+    worker_thread_.join();
+    // calls the exporter to shutdown
+    if (exporter_ != nullptr)
+    {
+      return exporter_->Shutdown();
+    }
+    return true;
+  }
+  return false;
```
jsuereth, line 209: You have a race condition on the "else" here for your destructor.  Scenario:

1. Someone manuals calls `Shutdown` on a thread.
2. The thread owning the BatchLogProcessor call the destructor prior to the `Shutdown` call being complete.
3. Your memory is deleted WHILE you're trying to shutdown.

What I'd recommend is to do the following:

1. Create a "is_shutdown_" that is *only* true IFF shutdown logic is complete.
2. Create a new "latch" "is_shutting_down_" which is true IFF shutdown logic is IN PROGRESS.
3. Create a new condition where you can park `Shutdown()` calls to wait for shutdown.

You'd have logic something like:

```
while (!is_shutdown.load()) {
   if (is_shutting_down.exchange(true)) {
      doTheShutdownThing();
      is_shutdown.store(true);
      waiting_on_shutdown.notify_all();
      is_shutting_down.store(false);
   } else {
     waiting_on_shutdown.wait(timeout);
   }
}
```

Note: I likely didn't get all my atomic arcana correct here, just nudging in the right direction.  I think if you'd rather use a big ole' mutex + wait for the whole shutdown method, that'd be fair for an initial impl and someone can optimise later.   Better to be correct than "Segmentation Fault"

sdk/test/logs/batch_log_processor_test.cc
```diff
@@ -0,0 +1,286 @@
...
+
+  for (int i = 0; i < num_logs; ++i)
+  {
+    auto log = batch_processor->MakeRecordable();
+    log->SetName("Log" + std::to_string(i));
+    batch_processor->OnReceive(std::move(log));
+  }
+
+  // Give some time to export the logs
+  std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
jsuereth, line 239: IIUC:  You should NOT need any sleep if you `ForceFlush` and your multithreaded code is correct.    I'd remove all the "sleep" code in your unit tests.  If they aren't passing, it's likely a synchronizations/concurrency issue that needs to be fixed.

jsuereth (changes requested):
I still think the concurrency portions of this may need some help.

Given the use of locks + condition for the core of this class, may make sense to just use a real lock/mutex instead of Atomics for now.

sdk/src/logs/batch_log_processor.cc
xukaren: Yeah for sure, will make the change!

sdk/src/logs/batch_log_processor.cc
xukaren: Yeah, I agree, with the initial implementation, I think it would be simpler to use the lock, so people who are looking at this later won't have bugs they need to fix. I can definitely look into optimizing this further in a future PR

sdk/include/opentelemetry/sdk/logs/batch_log_processor.h
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * NOTE: Timeout functionality not supported yet.
+   */
+  bool Shutdown(
+      std::chrono::microseconds timeout = std::chrono::microseconds::max()) noexcept override;
+
+  /**
+   * Class destructor which invokes the Shutdown() method.
+   */
+  ~BatchLogProcessor();
```
ThomsonTan, line 89: Add `override` as `LogProcessor` has virtual destructor?

sdk/test/logs/batch_log_processor_test.cc
xukaren: Right, I based it off the way te batch span processor tests were written, but I just removed them and looks like nothing is breaking. I'll definitely look more into optimizing this processor in a separate PR and after testing it more

sdk/include/opentelemetry/sdk/logs/batch_log_processor.h
xukaren: Right, will do

sdk/test/logs/batch_log_processor_test.cc
xukaren: Yup, you're right, the `ForceFlush()` calls should not require this extra sleep time, since the `ForceFlush()` method needs to wait for the worker thread to signal back that its flushing has completed before returning `true`. Removed these from the tests. 

In the last test,`TestScheduleDelayMillis` - I added a comment why it needed that small delay is necessary, since there is no _mandatory_  requirement for  the worker thread (which is performing asynchronously) to complete the current export, before proceeding to the next line of the tests

sdk/src/logs/batch_log_processor.cc
xukaren: Right, I see that race condition is definitely a problem. With this initial implementation, I agree it would be simpler to use the lock. I've changed that in the latest commit

xukaren:
@jsuereth, please let me know what you think of these changes! :) 
@maxgolov would love a final stamp as well before merging!

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,238 @@
...
+
+  // In case this export was the result of a force flush, notify main thread that force flush is
+  // complete
+  if (was_force_flush_called == true)
+  {
+    is_force_flush_notified_.store(true);
+
+    while (is_force_flush_notified_.load() == true)
+    {
+      force_flush_cv_.notify_one();
```
ThomsonTan, line 90: Why notify main thread by `force_flush_cv_.notify_one()` for multiple times, should just one notification work?

xukaren: I had the same question originally, but the `while` around it is to prevents the notification call being missed, in case the `notify` runs before the `wait`

ThomsonTan: I think it is fine for the `notify_one()` be missed, because `force_flush_cv_.wait()` has checks `is_force_lufsh_notified` is its predicate, which is already set before `notify_one()`. So if `notify_one()` happened before the `force_flush_cv_wait()`, the latter wait will return immediately without entering wait state.

xukaren: I removed it in [this commit](https://github.com/open-telemetry/opentelemetry-cpp/pull/434/commits/14033b514096017b57d9c8cd00217e383b2798e4#diff-f7c189661e15834743b64c453ad9e3548337c19515074da4bc43bb0205c94ffaR85) and it seemed to be working. But following the [batch span processor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/batch_span_processor.cc#L163), which wraps all the `cv.notify()` with `while` loops`, I thought it should be kept consistent

xukaren:
OK I just reverted this log processor to the [batch span processor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/batch_span_processor.cc#L163) implementation - this leaves both consistent and ensures there aren't any concurrency issues that are newly introduced. 

Instead of addding any new logic that could break something here, both traces/logs can have the same implementation that can possibly be optimized at a later time.

jsuereth (approved):
I think keeping this the same as the Trace batch processor (for now) is definitely the right call.  Threading logic looks good.

sdk/test/logs/batch_log_processor_test.cc
```diff
@@ -0,0 +1,275 @@
...
+        is_shutdown_(is_shutdown),
+        is_export_completed_(is_export_completed),
+        export_delay_(export_delay)
+  {}
+
+  std::unique_ptr<Recordable> MakeRecordable() noexcept
+  {
+    return std::unique_ptr<Recordable>(new LogRecord());
+  }
+  // Export method stores the logs received into a shared list of record names
```
ThomsonTan, line 48: nit: add an empty line before the `Export(` function definition.

ThomsonTan (approved):

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,213 @@
...
+      });
+
+  exporter_->Export(
+      nostd::span<std::unique_ptr<Recordable>>(records_arr.data(), records_arr.size()));
+
+  // Notify the main thread in case this export was the result of a force flush.
+  if (was_force_flush_called == true)
+  {
+    is_force_flush_notified_ = true;
+    while (is_force_flush_notified_.load() == true)
```
maxgolov, line 174: Something isn't good about this while loop..

sdk/src/logs/batch_log_processor.cc
```diff
@@ -0,0 +1,213 @@
...
+bool BatchLogProcessor::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  if (is_shutdown_.load() == true)
+  {
+    return false;
+  }
+
+  is_force_flush_ = true;
+
+  // Keep attempting to wake up the worker thread
```
maxgolov, line 75: Could you elaborate on why the worker may not immediately change the state. We are effectively burning CPU cycles here..?

xukaren: Yeah I think we're in agreement - Tom also mentioned this in an earlier comment.

I had removed it in an earlier commit [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/434/commits/14033b514096017b57d9c8cd00217e383b2798e4#diff-f7c189661e15834743b64c453ad9e3548337c19515074da4bc43bb0205c94ffaR179) and explained why [in this comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/434#discussion_r543525388) but didn’t receive any replies so i wasnt sure if it was good or not

sdk/src/logs/batch_log_processor.cc
xukaren: Yeah I'm not sure whether the batch span processor for implementing it this way. i looked through the comments and couldn't find any discussion about this. 

I had removed it in an earlier commit, and Tom also agreed this probably isn't necessary. I also checked that this passes all the tests with/without it.

xukaren: Would it be easier to fix in a followup issue simultaneously for the batch log/span processors?

maxgolov: Ok.

maxgolov (approved):

sdk/src/logs/batch_log_processor.cc
maxgolov: https://github.com/open-telemetry/opentelemetry-cpp/issues/477

sdk/src/logs/batch_log_processor.cc
maxgolov: https://github.com/open-telemetry/opentelemetry-cpp/issues/477

maxgolov:
Logged a separate issue to make this code readable, and/or avoiding the while looks on waiting for variable update:
https://github.com/open-telemetry/opentelemetry-cpp/issues/477

PR #435, opened 2020-12-09
Added Logging SDK Context Injection and Logger Limit

MarkSeufert:
**Note:** This PR is going to be updated once the [recordable PR](https://github.com/open-telemetry/opentelemetry-cpp/pull/438) gets merged.

This PR makes the following additions to the [logging SDK](https://github.com/open-telemetry/opentelemetry-cpp/pull/386):
* The SDK’s Logger injects the relevant fields of the LogRecord if the user didn’t specify them. The fields that get injected are timestamp, severity, traceid, spanid, and traceflags.
* Added a logger limit to the SDK LoggerProvider ([relevant spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1259)), where the `GetLogger(name)` method returns a noop logger if the limit has been reached.
* Unit tests for the above functionality.

The CI tests will fail until the recent [API PR](https://github.com/open-telemetry/opentelemetry-cpp/pull/422) gets merged since it uses its `trace_flag` -> `trace_flags` change and `Severity:kInvalid = 0` change.

cc - @xukaren @alolita  @ThomsonTan @lalitb

PR #438, opened 2020-12-10
Add a Recordable interface, implementation, and tests for logs

xukaren:
This PR implements a recordable for logs, as part of the [logging prototype](https://github.com/open-telemetry/opentelemetry-cpp/issues/337), and also resolving issue #412. 

This PR:

* Adds a Recordable interface for logs
* Add a default `LogRecord` Recordable  implementation, with unit tests 
* Changes the methods and signatures to take in a `Recordable` instead of the previous `LogRecord` struct, across the processor and exporter files
* Refactors all affected files (e.g. tests for the logging API, SDK, and processors) 

API/SDK changes: 
* Gets the TraceId, SpanId, TraceFlags, and current Timestamp in the SDK and injects into the LogRecord if it is not user-specified. 
* Removed the `Severity::kDefault` enum value (originally set as kInfo), and added a `Severity::kInvalid`= 0. (Setting the default Severity to kInfo has been moved to the SDK instead of the API)

Other minor changes: 
* Removes the `LogRecord` struct from the API 
* Stylistic change: Renames the `log()` methods to `Log()`, and adds `this->` before calling Log() methods
 
This PR introduces changes that blocks a few other open PRs: #422, #430, #434, and #435.

cc @MarkSeufert @alolita


api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+           nostd::string_view body,
+           std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>> resource,
+           std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>> attributes,
+           trace::TraceId trace_id,
+           trace::SpanId span_id,
+           trace::TraceFlags trace_flags) noexcept
   {
-    // TODO: set default timestamp later (not in API)
-    log(severity, message, core::SystemTimestamp(std::chrono::system_clock::now()));
+    return this->Log(timestamp, severity, name, body,
```
maxgolov, line 130: Is there a practical reason why you need `this->` here ?

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+   */
+  template <class T,
+            class U,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<U>::value> * = nullptr>
+  void Log(core::SystemTimestamp timestamp,
+           Severity severity,
+           nostd::string_view name,
+           nostd::string_view body,
+           const T &resources,
```
maxgolov, line 110: Should this be `resource` ?
```suggestion
           const T &resource,
```

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+           const U &attributes,
+           trace::TraceId trace_id,
+           trace::SpanId span_id,
+           trace::TraceFlags trace_flags) noexcept
   {
-    // Set severity to the default then call log(Severity, String message) method
-    log(Severity::kDefault, message);
+    Log(timestamp, severity, name, body, common::KeyValueIterableView<T>(resources),
```
maxgolov, line 116:
```suggestion
    Log(timestamp, severity, name, body, common::KeyValueIterableView<T>(resource),
```

api/include/opentelemetry/logs/logger.h
xukaren: Not really, but I thought it would be better to follow the convention of [StartSpan()](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/trace/tracer.h#L52)

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+        common::KeyValueIterableView<U>(attributes), trace_id, span_id, trace_flags);
   }
 
-  inline void log(Severity severity, nostd::string_view message) noexcept
+  void Log(core::SystemTimestamp timestamp,
+           Severity severity,
+           nostd::string_view name,
+           nostd::string_view body,
+           std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>> resource,
```
maxgolov, line 124: Basically either `resources` here or `resource` above for the sake of consistency.

xukaren: Typo, my mistake - it should be `resource` to match the [Log Data Model field](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/logs/data-model.md#log-and-event-record-definition)

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
-  template <class T,
-            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
-  inline void log(Severity severity       = Severity::kDefault,
-                  nostd::string_view name = "",
-                  const T &attributes     = {}) noexcept
+  // Set default time, and call base Log(severity, message, time) method
+  inline void Log(Severity severity, nostd::string_view message) noexcept
   {
-    log(severity, name, common::KeyValueIterableView<T>(attributes));
+    this->Log(severity, message, default_timestamp_);
```
maxgolov, line 152: I do not have a strong opinion on this. But ideally the `default_timestamp_` should be something like `std::chrono::system_clock::now()`? Maybe we can make an assumption that duration of time `inline` call to `Log` and to the moment when we set it to that actual value (in Exporter?) is rather miniscule.. Basically we log `current name` by default. Maybe an implementation of overridable getter that allows to populate this in concrete SDK logger implementation.

api/include/opentelemetry/logs/severity.h
```diff
@@ -0,0 +1,57 @@
...
+namespace logs
+{
+
+/**
+ * Severity Levels assigned to log events, based on Log Data Model
+ */
+enum class Severity : uint8_t
+{
+  kInvalid = 0,
+  kTrace   = 1,
```
maxgolov, line 30: This looks cool and quite readable. But `If the first enumerator does not have an initializer, the associated value is zero. For any other enumerator whose definition does not have an initializer, the associated value is the value of the previous enumerator plus one.` ... basically you do not have to write `= 0`, `= 1`, etc.. Is there a coding style standard that forces us to write more? Since we can safely write less.

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
```diff
@@ -0,0 +1,123 @@
...
+namespace sdk
+{
+namespace logs
+{
+/**
+ * A counterpart to AttributeValue that makes sure a value is owned. This
+ * replaces all non-owning references with owned copies.
+ */
+using LogRecordAttributeValue = nostd::variant<bool,
+                                               int32_t,
```
maxgolov, line 18: ❤️ the fact that you use standard types here, i.e. `int32_t` - looks much cleaner than `int` we had in another class for traces. Good job!

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  template <typename T, typename U = T>
+  LogRecordAttributeValue convertSpan(nostd::span<const U> vals)
+  {
+    std::vector<T> copy;
+    for (auto &val : vals)
+    {
+      copy.push_back(T(val));
+    }
```
maxgolov, line 81:
```suggestion
    const std::vector<T> copy(vals.begin(), vals.end());
```

sdk/src/logs/logger.cc
```diff
@@ -36,23 +51,76 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
...
   // TODO: Sampler logic (should include check for minSeverity)
 
-  /**
-   * Convert the LogRecord to the heap first before sending to processor.
-   * TODO: Change the API log(LogRecord) function to log(*LogRecord) so the following line
-   * converting record a heap variable can be removed
-   */
-  auto record_pointer =
-      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+  auto recordable = processor->MakeRecordable();
```
maxgolov, line 54: I do realize that we allow the SDK built with or without exceptions. However, if we build it with exceptions, is there a provision for the `processor->MakeRecordable()` to throw `std::bad_alloc`? What happens here, is in environment where exceptions are enabled - we'd be silently dropping telemetry. This may not be good in some cases, i.e. if it's a security audit log - we need to potentially bubble up the error condition to the caller, for the caller to decide what to do with that.

I do not think that this is blocking. But if you can log an issue about this, that'd work fine.

sdk/src/logs/logger.cc
```diff
@@ -36,23 +51,76 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
 
   // TODO: Sampler logic (should include check for minSeverity)
 
-  /**
-   * Convert the LogRecord to the heap first before sending to processor.
-   * TODO: Change the API log(LogRecord) function to log(*LogRecord) so the following line
-   * converting record a heap variable can be removed
-   */
-  auto record_pointer =
```
maxgolov, line 44: So this code was throwing `std::bad_alloc` before your change, right?

sdk/src/logs/logger.cc
```diff
@@ -36,23 +51,76 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
...
-      std::unique_ptr<opentelemetry::logs::LogRecord>(new opentelemetry::logs::LogRecord(record));
+  auto recordable = processor->MakeRecordable();
+  if (recordable == nullptr)
+  {
+    return;
+  }
+
+  // Populate recordable fields
+  if (timestamp == this->default_timestamp_)
```
maxgolov, line 61: Related to my other comment, that inline timestamp when we logged - would be different from when we populate it here, since this assignment actually has to wait for the lock on `logger_provider_.lock()` and a memory allocation. In case of a congested system it might be a few milliseconds difference.

sdk/src/logs/logger.cc
maxgolov: Okay, I see that this should be handled by new in `MakeRecordable()` method potentially throwing. Please ignore my above comment.

maxgolov (approved):
Good, with minor suggestions.

sdk/src/logs/logger.cc
xukaren: That's a good point, I hadn't thought of that. Would getting the actual timestamp as early as the API be a time-consuming task though? If the LogRecord were filtered out by the Sampler in the SDK for example, would the previous timestamp call have been wasted/unecessary?

sdk/src/logs/logger.cc
xukaren: If recordable is null, maybe an error could be invoked using #408?

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
xukaren: Can't take credit for this - these variants are already defined in `AttributeValue` :)

api/include/opentelemetry/logs/severity.h
xukaren: That makes sense, here since we don't have any non-continuous severity number mappings and the first one starts at 0, we could simplify by removing all the initializers. Will change this in the next commit!

sdk/src/logs/logger.cc
maxgolov: It depends. In some scenarios you'd rather prefer the log time as close to original as possible, as you need the actual time when something happened - not when you got thru the lock to assign some timestamp in the SDK, esp. given it's under some lock. This is not a blocker for your PR. I'm thinking maybe we should have a method added in API later on that allows to customize this behavior: whether it's the API stamps the `now` time _right away_ , or if API stamps the indicator `please stamp the time for me` (like you are doing it here), and then later on the next stage stamps it.

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+           trace::TraceFlags trace_flags) noexcept
   {
-    // TODO: set default timestamp later (not in API)
-    log(severity, message, core::SystemTimestamp(std::chrono::system_clock::now()));
+    return this->Log(timestamp, severity, name, body,
+                     nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+                         resource.begin(), resource.end()},
+                     nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+                         attributes.begin(), attributes.end()},
+                     trace_id, span_id, trace_flags);
```
lalitb, line 135: Nice to see this `initializer_list` for `resource` and `attributes`.

api/include/opentelemetry/logs/logger.h
lalitb: +1

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
```diff
@@ -0,0 +1,118 @@
...
+    attributes_[std::string(key)] = nostd::visit(converter_, value);
+  }
+
+private:
+  std::unordered_map<std::string, LogRecordAttributeValue> attributes_;
+  AttributeConverter converter_;
+};
+}  // namespace logs
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
```
lalitb, line 118: Is it possible we can share this class across trace and logs, and move it to `sdk/common`. Probably not as part of this PR if it's huge task, we can have a issue open for it.

lalitb (approved):
Thanks for working on this PR. We are now using the right approach to propagate logs to exporter : )

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
xukaren: I can fix this in a future PR! (If I don't have time I will create an issue for it)

sdk/src/logs/logger.cc
xukaren: If the processor returns a `nullptr`, could we note an error like:
 ```
if (recordable == nullptr) { 
     error "recordable creation failed";
     return;
}
```` 
where the error is handled by error diagnostics in #408? We may not know the reason for it if the processor/exporter handled it, but at least the user will know something went wrong

File: api/include/opentelemetry/logs/logger.h [hunk 14]
```diff
@@ -35,6 +41,17 @@ namespace logs
...
 class Logger
 {
+protected:
+  // Default values to set for fields that the user doesn't specify
+  core::SystemTimestamp default_timestamp_;  // uint64 nanoseconds since Unix epoch, default 0
+  Severity default_severity_;                // default 0 (kInvalid)
+  nostd::string_view default_name_;          // default 0 length, nullptr data
+  nostd::string_view default_body_;  // default 0 length, nullptr data. TODO: currently a simple
+                                     // string, but should be changed "Any" type
+  trace::TraceId default_trace_id_;  // default 00000000000000000000000000000000
```
ThomsonTan, line 51: Are these numbers in in hex? Better to have prefix like `0x` to make the indication clear.

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
+                     nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+                         attributes.begin(), attributes.end()},
+                     trace_id, span_id, trace_flags);
   }
 
-  inline void log(Severity severity,
+  /** Wrapper methods that the user could call for convenience when logging **/
+
+  // Set default values for unspecified fields, then call the base Log() method
+  inline void Log(Severity severity,
```
ThomsonTan, line 141: The `inline` keywords in this header file seems unnecessary because implicit inline applied here?

https://stackoverflow.com/questions/28183729/why-i-can-implement-the-classs-member-function-in-header-file

api/include/opentelemetry/logs/logger.h
```diff
@@ -43,81 +60,105 @@ class Logger
...
                   core::SystemTimestamp time) noexcept
   {
-    // creates a LogRecord object with given parameters, then calls log(LogRecord)
-    LogRecord r;
-    r.severity  = severity;
-    r.body      = message;
-    r.timestamp = time;
-
-    log(r);
+    this->Log(time, severity, default_name_, message, {}, {}, default_trace_id_, default_span_id_,
```
ThomsonTan, line 145: Could the `nullptr` be passed into `Log` instead of `{}` which may involves object construction?

api/include/opentelemetry/logs/logger.h
xukaren: Yeah `0x00000000000000000000000000000000` would be the value of the TraceId if it were converted to hex. But internally Do we refer to TraceIds or SpanIds in Hex? (i.e. or just simple as 

Internally, it is a 32 `uint8_t` array all initialized to `0`s, but it would be converted to lowercase base 16 representation before being used. 

Would it be simpler to say "default value all initialized to 0s"?

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Creates an owned copy (LogRecordAttributeValue) of a non-owning AttributeValue.
+ */
+struct AttributeConverter
+{
+  LogRecordAttributeValue operator()(bool v) { return LogRecordAttributeValue(v); }
+  LogRecordAttributeValue operator()(int32_t v) { return LogRecordAttributeValue(v); }
+  LogRecordAttributeValue operator()(uint32_t v) { return LogRecordAttributeValue(v); }
+  /*LogRecordAttributeValue operator()(int v)
```
ThomsonTan, line 40: Could this comment about `operator()(int v)` just be removed?

sdk/src/logs/logger.cc
ThomsonTan: This could be a workaround. Could you add an `TODO` in the `if` check to add such error diagnostic?

sdk/src/logs/logger.cc
```diff
@@ -36,23 +51,76 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
...
+
+  attributes.ForEachKeyValue([&](nostd::string_view key,
+                                 opentelemetry::common::AttributeValue value) noexcept {
+    recordable->SetAttribute(key, value);
+    return true;
+  });
+
+  // Inject trace_id/span_id/trace_flags if none is set by user
+  auto provider     = opentelemetry::trace::Provider::GetTracerProvider();
+  auto tracer       = provider->GetTracer("foo_library");
```
ThomsonTan, line 89: What is "foo_library", default tracer name?

api/include/opentelemetry/logs/logger.h
xukaren: Does this follow the convention for Trace/SpanIds? Or would it simply be better to say `default 00000000000000000000000000000000, in base 16`

sdk/src/logs/logger.cc
xukaren: I didn't notice I left this in - from an earlier example written. Yes, it seems `GetTracer()` doesn't accept an empty param list? Would it be better to remove injection from this PR completely, as it seems to be a separate topic? Or is there a better default name I could use here (which I could also revisit the naming in a later PR)?

sdk/src/logs/logger.cc
xukaren: Not that I'm aware of, but updated this to use a recordable for faster/more reliable perfomance  (#403), and also allow custom recordable implementations by exporters

api/include/opentelemetry/logs/logger.h
xukaren: True, updated this in the latest commit.

api/include/opentelemetry/logs/logger.h
ThomsonTan: Yes, the suggestion looks good too.

xukaren:
@pyohannes would you be able to review this? @jsuereth I think this addresses some of the concerns you had about the performance and efficiency concerns of the previous approach for log exporting

api/include/opentelemetry/logs/logger.h
xukaren: Looked into this. I believe since the `KeyValueIterable` is a `const &` reference, and not a pointer itself, it cannot point to a `nullptr` and needs to point to an object with allocated memory (https://stackoverflow.com/questions/4364536/is-null-reference-possible). In this case I believe `{}` should be the best option here

ThomsonTan (approved):

api/include/opentelemetry/logs/logger.h
```diff
@@ -35,89 +41,118 @@ namespace logs
...
-    // creates a LogRecord object with given parameters, then calls log(LogRecord)
-    LogRecord r;
-    r.severity  = severity;
-    r.body      = message;
-    r.timestamp = time;
+  /** Wrapper methods that the user could call for convenience when logging **/
 
-    log(r);
+  // Set default values for unspecified fields, then call the base Log() method
+  void Log(Severity severity, nostd::string_view message, core::SystemTimestamp time) noexcept
```
pyohannes, line 137: Why do we have `time` first in the virtual `Log` call, but `severity` first here? I think we should keep the order of the arguments consistent.

api/include/opentelemetry/logs/logger.h
```diff
@@ -35,89 +41,118 @@ namespace logs
...
 {
+protected:
+  // Default values to set for fields that the user doesn't specify
+  Severity default_severity_;        // default 0 (severity of kInvalid)
+  nostd::string_view default_name_;  // default 0 length, nullptr data
+  nostd::string_view default_body_;  // default 0 length, nullptr data. TODO: currently a simple
+                                     // string, but should be changed "Any" type
+  trace::TraceId default_trace_id_;  // default 00000000000000000000000000000000, in base 16
+  trace::SpanId default_span_id_;    // default 0000000000000000, in base 16
+  trace::TraceFlags default_trace_flags_;  // default 00, in base 16
```
pyohannes, line 52: We should get rid of this. I think we can specify `""` whereever `default_name_` or `default_body_` is used, and we could could use `{}` instead of `default_trace_id_`, `default_span_id_` and `default_trace_flags_`.

pyohannes (commented):
Thanks for refactoring this! That's much closer to a design that is production-ready.

api/include/opentelemetry/logs/logger.h
xukaren: I think it would be most ergonomic to users to specify severity, then message. The timestamp overload at the end I think could be an optional argument (which I can now add if you like that idea)

xukaren: It seems strange to specify the timestamp first since it won't be a commonly used feature. Hmm since it seems most ergonomic to users to specify severity, then message, I can move timestamp to the end of the main Log() method for consistency here. Would that work?

api/include/opentelemetry/logs/logger.h
xukaren: Sure, that sounds pretty intuitive - didn't know `{}` could be a value for trace/spanid :') 

Would you say it's better a default Severity be left as is (to `kInvalid=0`) or directly to `kInfo` as default? Would there be any benefits of knowing when the user intentionally set the severity to `kInfo` and when it was overwritten by the API overloads / SDK?

sdk/src/logs/logger.cc
```diff
@@ -34,25 +49,70 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
...
-  // TODO: inject traceid/spanid later
+  // Inject trace_id/span_id/trace_flags if none is set by user
+  auto provider     = opentelemetry::trace::Provider::GetTracerProvider();
+  auto tracer       = provider->GetTracer("foo_library");
+  auto span_context = tracer->GetCurrentSpan()->GetContext();
+
+  // Traceid
+  if (!trace_id.IsValid())
+  {
+    recordable->SetTraceId(span_context.trace_id());
```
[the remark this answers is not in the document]
xukaren: Added this check

sdk/src/logs/logger.cc
xukaren: Fixed - added Logger name and updated the Tracer name above to be the same in latest commit

sdk/src/logs/logger.cc
xukaren: Sure, added `TODO` and just opened up https://github.com/open-telemetry/opentelemetry-cpp/issues/441 for further discussion

sdk/src/logs/logger.cc
```diff
@@ -34,25 +50,70 @@ void Logger::log(const opentelemetry::logs::LogRecord &record) noexcept
...
+  auto span_context = tracer->GetCurrentSpan()->GetContext();
+
+  // Traceid
+  if (!trace_id.IsValid() && span_context.trace_id().IsValid())
+  {
+    recordable->SetTraceId(span_context.trace_id());
+  }
+  else
+  {
+    recordable->SetTraceId(trace_id);
```
jsuereth, line 92: with the new check above, it's possible you're setting trace_id here if `span_context.trace_id().Isvalid()` returns false, i.e. it could be invalid.

Might be easier just do:

```
if trace_id valid => setTraceId
else if context trace id valid => setTraceId
else // nothing.
```

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -41,18 +40,40 @@ class Logger final : public opentelemetry::logs::Logger
...
+           trace::TraceFlags trace_flags,
+           core::SystemTimestamp timestamp) noexcept override;
 
 private:
   // The logger provider of this Logger. Uses a weak_ptr to avoid cyclic dependancy issues the with
   // logger provider
   std::weak_ptr<LoggerProvider> logger_provider_;
+
+  // The name of this logger
+  opentelemetry::nostd::string_view logger_name_;
```
jsuereth, line 76: Note: you may have to take ownership of this string. 

You need to at least do one of two things:

1. Document in the constructor that the instantiator of the Logger is responsible for keeping this string accessible for the lifetime of the logger.
2. Copy the String into memory you own and can control the lifecycle.

jsuereth (commented):
Nice work!  Just a few minor nits from me, but thanks very much for tackling this!!

sdk/src/logs/logger.cc
xukaren: This is much cleaner :) Thanks, updated as above (minus the empty `else` block)

jsuereth (approved):
Great work!

sdk/include/opentelemetry/sdk/logs/attribute_utils.h
xukaren: @lalitb @jsuereth Just an update on this-  I've moved the `attribute_utils.h` file from `logs` to `common` for this PR, and renamed `LogRecordAttributeValue` to `OwnedAttributeValue`. All the necessary naming changes for `LogRecordAttributeValue` have been done in this PR. 

Seeing as this PR seems to already be running on the long side, and also unrelated to traces/spans, I've filed the traces area needing to be refactored as an issue here: #442 , which I may take a look at when working on the OStream exporter for logs, or if not would probably be a good first issue for someone to tackle as well

xukaren:
@pyohannes @lalitb Should this be ready to merge?

xukaren:
@pyohannes @lalitb Would this be good to merge?

lalitb:
@xukaren  - Please rebase the branch with base, so we should be ready to merge it. thanks.

xukaren:
> @xukaren - Please rebase the branch with base, so we should be ready to merge it. thanks.

Done!

PR #444, opened 2020-12-13
Add Elasticsearch Log Exporter with tests

MarkSeufert:
This PR adds an Elasticsearch exporter for logs, which is a requirement from [issue #337](https://github.com/open-telemetry/opentelemetry-cpp/issues/337). The main features include:
- Specifying the host, port, and index/data stream of the Elasticsearch instance
- Sending an HTTP POST request to Elasticsearch using the [HTTP Client Interface](https://github.com/open-telemetry/opentelemetry-cpp/pull/385), and parsing the response message to ensure it was successfully received
- Sending batches of LogRecords using the [Bulk API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html)
- Timeout logic, such that the Export() method doesn't block indefinitely
- Unit tests for the above functionality

Before reviewing this PR, it should be noted that:
* This exporter current does not support CMake build as there is currently no CMake support for the Nlohmann/JSON library - will try to find a way around this using a backport for it (cc @maxgolov)
* Changes to logger.h and log_record.h can be ignore, as this PR will soon be changed to use a recordable instead of an API LogRecord once the [recordable PR](https://github.com/open-telemetry/opentelemetry-cpp/pull/438) gets merged. 

cc - @xukaren @alolita @lalitb @jsuereth @ThomsonTan


lalitb:
Though I don't have strong opinion, would be good to get other's opinion for the right repo to merge this exporter PR, and other similar PRs - `main` or `contrib`. As per spec guidelines 
 - https://github.com/open-telemetry/opentelemetry-specification/blob/0229140bb13e8473f7c51513afd12e2a60618488/specification/library-guidelines.md - only limited set of exporters can go in main repo:
```
The SDK implementation should include the following exporters:

Jaeger.
Zipkin.
Prometheus.
OpenTelemetry Protocol (when the protocol is specified and approved).
Standard output (or logging) to use for debugging and testing as well as an input for the various log proxy tools.
In-memory (mock) exporter that accumulates telemetry data in the local memory and allows to inspect it (useful for e.g. unit tests).

Note: some of these support multiple protocols (e.g. gRPC, Thrift, etc). The exact list of protocols to implement in the exporters is TBD.

Other vendor-specific exporters (exporters that implement vendor protocols) should not be included in language libraries and should be placed elsewhere (the exact approach for storing and maintaining vendor-specific exporters will be defined in the future).
```

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,254 @@
...
+  if (options_.console_debug_)
+    std::cout << "waiting for response from Elasticsearch (timeout = " << options_.response_timeout_
+              << " seconds)" << std::endl;
+  bool receivedResponse = handler->waitForResponse(options_.response_timeout_);
+
+  // End the session
+  session->FinishSession();
+  session_manager.CancelAllSessions();
+  session_manager.FinishAllSessions();
+
```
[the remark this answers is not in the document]
MarkSeufert: Sure thing, I'll move them there after I add the session manager as a private member.

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,254 @@
...
+  std::string GetResponseBody()
+  {
+    if (!response_received_)
+      return "No response";
+
+    return body_;
+  }
+
+  // Virtual method definition that isn't used
+  void OnEvent(http_client::SessionState state,
```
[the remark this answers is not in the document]
MarkSeufert: Oh cool, I wasn't sure what this callback was used for but that sounds very useful. I'll add some functionality for this!

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,254 @@
...
+
+  // Send the request
+  std::unique_ptr<ResponseHandler> handler(new ResponseHandler());
+  session->SendRequest(*handler);
+
+  // Wait for the response to be received
+  if (options_.console_debug_)
+    std::cout << "waiting for response from Elasticsearch (timeout = " << options_.response_timeout_
+              << " seconds)" << std::endl;
+  bool receivedResponse = handler->waitForResponse(options_.response_timeout_);
```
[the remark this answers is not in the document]
MarkSeufert: Yea, the only time we wouldn't want the `SendRequest()` method to be synchronous is if the main thread didn't care about the status of the response message. All the exporters have this blocking for response status requirement so unless the http client is going to be used outside of the exporters, changing the API would simplify this a bunch :)

exporters/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ endif()
 
 add_subdirectory(ostream)
 add_subdirectory(memory)
+# add_subdirectory(elasticsearch)
```
ThomsonTan, line 7: Off by default and add an flag like `WITH_ELASTICSEARCH` to guard it?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,254 @@
...
+  log["traceid"] = std::string(trace_buf, sizeof(trace_buf));
+
+  char span_buf[16];
+  record->span_id.ToLowerBase16(span_buf);
+  log["spanid"] = std::string(span_buf, sizeof(span_buf));
+
+  char flag_buf[2];
+  record->trace_flags.ToLowerBase16(flag_buf);
+  log["traceflags"] = std::string(flag_buf, sizeof(flag_buf));
+
```
lalitb, line 248: Is there any document defining the mapping between opentelementry log fields to elasic ?

lalitb:
Also, please create a github issue for having an example on how to run this against elastic server. As we have for otlp:
https://github.com/open-telemetry/opentelemetry-cpp/tree/master/examples/otlp

exporters/elasticsearch/BUILD
```diff
@@ -0,0 +1,40 @@
...
+        "-DCURL_STATICLIB",
+    ],
+    linkopts = select({
+        "//bazel:windows": [
+            "-DEFAULTLIB:advapi32.lib",
+            "-DEFAULTLIB:crypt32.lib",
+            "-DEFAULTLIB:Normaliz.lib",
+        ],
+        "//conditions:default": [],
+    }),
```
jsuereth, line 33: Since you're baking the 'curl' client into the exporter itself, you need the copts to be part of the library above.

exporters/elasticsearch/src/es_log_exporter.cc
MarkSeufert: No I haven't seen any specs for this, but the Elasticsearch data schema is very flexible so I'm so I was keeping the data as close to the Log Data Model as possible. There's a recent issue for the collector about [supporting logs into their Metrics/Traces ES exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/1800), so I'll ping them about it there.

exporters/elasticsearch/BUILD
MarkSeufert: Worked like a charm, thanks Josh :)

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,177 @@
...
+
+public:
+  /**
+   * Set the severity for this log.
+   * @param severity the severity of the event
+   */
+  void SetSeverity(opentelemetry::logs::Severity severity) noexcept override
+  {
+    // Convert the severity enum to a string
+    static opentelemetry::nostd::string_view severityStringMap_[25] = {
```
xukaren, line 86: nit: this map is already included in https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/logs/severity.h#L65

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,177 @@
...
+        "kWarn2",   "kWarn3",  "kWarn4",  "kError",  "kError2", "kError3", "kError4",
+        "kFatal",   "kFatal2", "kFatal3", "kFatal4"};
+    json_["severity"] = severityStringMap_[static_cast<int>(severity)];
+  }
+
+  /**
+   * Set name for this log
+   * @param name the name to set
+   */
+  void SetName(nostd::string_view name) noexcept override { json_["name"] = name.data(); }
```
xukaren, line 98: nit: not sure if this would be better stored as a string? i.e. `std::string(name)`? (Same for `body` field)

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Set a resource for this log.
+   * @param name the name of the resource
+   * @param value the resource value
+   */
+  void SetResource(nostd::string_view key,
+                   const opentelemetry::common::AttributeValue &value) noexcept override
+  {
+    json_["resource"][key.data()] = ValueToString(value);
```
xukaren, line 114: (same here, and in line 125, maybe a string for instead of `key.data()`?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Shutdown this exporter.
+   * @param timeout The maximum time to wait for the shutdown method to return
+   */
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
+
+private:
+  // Stores if this exporter had its Shutdown() method called
+  bool isShutdown_ = false;
```
xukaren, line 109: name variable `is_shutdown_` to follow Google naming convention?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,160 @@
...
+   * @param records A list of log records to send to Elasticsearch.
+   */
+  sdklogs::ExportResult Export(const nostd::span<std::unique_ptr<opentelemetry::logs::LogRecord>>
+                                   &records) noexcept override;
+
+  /**
+   * Shutdown this exporter.
+   * @param timeout The maximum time to wait for the shutdown method to return
+   */
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
```
xukaren, line 105: default timeout `std::chrono::microseconds::max()`?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,258 @@
...
+    cv_.wait_for(lk, std::chrono::milliseconds(1000 * timeoutSec));
+    return response_received_;
+  }
+
+  /**
+   * Returns the body of the response
+   */
+  std::string GetResponseBody()
+  {
+    if (!response_received_)
```
xukaren, line 65: nit: add `{}` around if block?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,258 @@
...
+    cv_.notify_all();
+  }
+
+  /**
+   * A method the user calls to block their thread until the response is received, or the timeout is
+   * exceeded.
+   */
+  bool waitForResponse(unsigned int timeoutSec = 1)
+  {
+    std::mutex mutex_;
```
xukaren, line 54: nit: is this mutex useful if it is only defined in scope of this function?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,258 @@
...
+  record->trace_id.ToLowerBase16(trace_buf);
+  log["traceid"] = std::string(trace_buf, sizeof(trace_buf));
+
+  char span_buf[16];
+  record->span_id.ToLowerBase16(span_buf);
+  log["spanid"] = std::string(span_buf, sizeof(span_buf));
+
+  char flag_buf[2];
+  record->trace_flags.ToLowerBase16(flag_buf);
+  log["traceflags"] = std::string(flag_buf, sizeof(flag_buf));
```
xukaren, line 251: nit: I think when data gets to elasticsearch, ideally the types should be preserved? ie. traceid/spanid/traceflags should be preserved as a hex value, not a string?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,160 @@
...
+  /**
+   * Converts a log record into a nlohmann::json object.
+   */
+  json RecordToJSON(std::unique_ptr<opentelemetry::logs::LogRecord> record);
+
+  /**
+   * Converts a common::AttributeValue into a string, which is used for parsing the attributes
+   * and resource KeyValueIterables
+   */
+  const std::string ValueToString(const common::AttributeValue &value)
```
xukaren, line 126: same here, not sure if every single type should be converted to a string, or if original data types should be preserved when converted to json

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
MarkSeufert: cool I didn't realize the recordable PR had this, definitely makes it simpler :)

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,202 @@
...
+    cv_.wait_for(lk, std::chrono::milliseconds(1000 * timeoutSec));
+    return response_received_;
+  }
+
+  /**
+   * Returns the body of the response
+   */
+  std::string GetResponseBody()
+  {
+    if (!response_received_)
```
xukaren, line 66: nit: add {} around if block?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Shutdown this exporter.
+   * @param timeout The maximum time to wait for the shutdown method to return
+   */
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
+
+private:
+  // Stores if this exporter had its Shutdown() method called
+  bool isShutdown_ = false;
```
xukaren, line 114: name variable `is_shutdown_` to follow Google naming convention?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,124 @@
...
+   * @param records A list of log records to send to Elasticsearch.
+   */
+  sdklogs::ExportResult Export(
+      const nostd::span<std::unique_ptr<sdk::logs::Recordable>> &records) noexcept override;
+
+  /**
+   * Shutdown this exporter.
+   * @param timeout The maximum time to wait for the shutdown method to return
+   */
+  bool Shutdown(std::chrono::microseconds timeout = std::chrono::microseconds(0)) noexcept override;
```
xukaren, line 110: default timeout `std::chrono::microseconds::max()`?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
xukaren, line 98: nit: not sure if this would be better stored as a string? i.e. std::string(name)? (Same for body field)

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -0,0 +1,202 @@
...
+    cv_.notify_all();
+  }
+
+  /**
+   * A method the user calls to block their thread until the response is received, or the timeout is
+   * exceeded.
+   */
+  bool waitForResponse(unsigned int timeoutSec = 1)
+  {
+    std::mutex mutex_;
```
xukaren, line 55: nit: is this mutex useful if it is only defined in scope of this function?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
MarkSeufert: the data() method returns a `char *`, which is what a string is so I don't think it matters.

exporters/elasticsearch/src/es_log_exporter.cc
MarkSeufert: hmm it's needed for the condition variable `wait_for` method to work, but looks like traces does it [more condensed](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/batch_span_processor.cc#L92) so good catch, I'll do it that way too

exporters/elasticsearch/src/es_log_exporter.cc
MarkSeufert: sure

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
MarkSeufert: yup agreed, I'll take a look at switching it over to use the correct data types

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,177 @@
...
+namespace exporter
+{
+namespace logs
+{
+
+/**
+ * An Elasticsearch Recordable implemenation that stores the 10 fields of the Log Data Model inside
+ * a JSON object,
+ */
+class ElasticSearchRecordable final : public sdk::logs::Recordable
```
xukaren, line 36: nit: maybe rename `JSONLogRecord`? Could be reused for a few other exporters as well

xukaren, line 36: nit: maybe rename `JSONLogRecord`? Could maybe be used for other exporters?

exporters/elasticsearch/src/es_log_exporter.cc
MarkSeufert: 100% Agree about the types being preserved. For trace/span/flags though, I think it's more human readable if it is left as a string representation of the hex values, then if it was a raw binary blob.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,177 @@
...
+   */
+  void SetSeverity(opentelemetry::logs::Severity severity) noexcept override
+  {
+    // Convert the severity enum to a string
+    static opentelemetry::nostd::string_view severityStringMap_[25] = {
+        "kInvalid", "kTrace",  "kTrace2", "kTrace3", "kTrace4", "kDebug",  "kDebug2",
+        "kDebug3",  "kDebug4", "kInfo",   "kInfo2",  "kInfo3",  "kInfo4",  "kWarn",
+        "kWarn2",   "kWarn3",  "kWarn4",  "kError",  "kError2", "kError3", "kError4",
+        "kFatal",   "kFatal2", "kFatal3", "kFatal4"};
+    json_["severity"] = severityStringMap_[static_cast<int>(severity)];
```
xukaren, line 91: nit: I think `severity_num` and `severity_text` should both separated into 2 different fields and both added to the json object - following the spec's rec, having both is ideal :)

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
lalitb: Don't think we can use it for others, as json format here would be specific to Elastic. Current name looks good in that case.

lalitb: ok pls ignore my comment, I think there is no format defined for elastic search, and we are kind of using somewhat generic format here. Still I feel we can keep current name : )

xukaren: haha yeah, i havent got any strong opinions on it - I agree it's a pretty generic format. so I'm wondering whether there's more use cases for it - maybe logging structured JSON data to a file would be useful? so those files could be used to be imported to databases or something? Im good with the name either way :)

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
MarkSeufert: hmmm it might be overkill for the user, since the severity probably isn't the main focus of the log statement. Having one field that says the severity is probably fine, but if anyone else feels strongly I can change it :)

xukaren: wouldn't elasticsearch need to filter/query by severity? maybe the severity number would be easier to compare than the text then?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -0,0 +1,125 @@
...
+#include "opentelemetry/nostd/type_traits.h"
+#include "opentelemetry/sdk/logs/exporter.h"
+#include "opentelemetry/sdk/logs/log_record.h"
+
+#include <time.h>
+#include <iostream>
+
+namespace nostd   = opentelemetry::nostd;
+namespace sdklogs = opentelemetry::sdk::logs;
+using json        = nlohmann::json;
```
ThomsonTan, line 30: `namespace json = nolhmann::json`?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,190 @@
...
+        break;
+      case common::AttributeType::TYPE_STRING:
+      case common::AttributeType::TYPE_CSTRING:
+        json_["attributes"][key.data()] =
+            opentelemetry::nostd::get<opentelemetry::nostd::string_view>(value).data();
+        break;
+      default:
+        break;
+    }
+  }
```
lalitb, line 139: Nit: It seems most of the code is repeated in these two functions `SetAttributes` and `SetResources`. Can we do instead with single function, either through some template function, or pass another argument depicting "attributes" or "resource".

lalitb (approved):
LGTM: If we can invest some time for example code as done for OTLP exporter, that would be great:
https://github.com/open-telemetry/opentelemetry-cpp/tree/master/examples/otlp

Or else please create a github issue for same.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
MarkSeufert: Yup good idea, I'll move it over into a function that takes in a string as the parameter.

MarkSeufert:
I have an Elasticsearch example written in my team repo right now, but I'm holding off for PRing it until this one gets merged. I think Its the same situation for the ostream exporter too, so 2 more future PRs!

exporters/elasticsearch/BUILD
```diff
@@ -0,0 +1,40 @@
...
+    copts = [
+        "-DCURL_STATICLIB",
+    ],
+    linkopts = select({
+        "//bazel:windows": [
+            "-DEFAULTLIB:advapi32.lib",
+            "-DEFAULTLIB:crypt32.lib",
+            "-DEFAULTLIB:Normaliz.lib",
+        ],
+        "//conditions:default": [],
```
jsuereth, line 21: note to self:  Really want to clean this up... glad it's working though!

jsuereth (commented):
AFAIk you only have two concurrency things happening here:

1. The HTTP libraries connection handling thread(s).
2. Your exporter, waiting to fire off a batch of data.

The threading code looks a little suspect (especially instantiating a local mutex).   Effectively you'll want some kind of memory barrier around access the shared variables between threads *or* you can use atomic variables to control that.

exporters/elasticsearch/src/es_log_exporter.cc
lalitb: jfyi - this code would become much simplified once we have synchronous request PR merged, as we won't need to wait for response anymore  : https://github.com/open-telemetry/opentelemetry-cpp/pull/448

This is how it would look like ( ref- zipkin exporter ) : https://github.com/open-telemetry/opentelemetry-cpp/blob/1039f4c2b947f3b3fd0880457ec77743fd7a80e7/exporters/zipkin/src/zipkin_exporter.cc#L62

MarkSeufert: Alright alright, I'll add a lock_guard around body_ and response_received_ so they can't be read and written at the same time. Yea, this is probably going to be replaced by Lalit's synchronous request, but for the sake of getting this merged ASAP we can always change it in a later issue/PR.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -0,0 +1,174 @@
...
+   * being "attributes" and "resources"
+   */
+  void WriteKeyValue(nostd::string_view key,
+                     const opentelemetry::common::AttributeValue &value,
+                     std::string name)
+  {
+    switch (value.index())
+    {
+      case common::AttributeType::TYPE_BOOL:
+        json_[name][key.data()] = opentelemetry::nostd::get<bool>(value) ? true : false;
```
xukaren, line 53: nit: does this work? or would `true` and `false` need to be wrapped with double quotes?

exporters/elasticsearch/test/es_log_exporter_test.cc
```diff
@@ -0,0 +1,73 @@
...
+  record->SetTimestamp(std::chrono::system_clock::now());
+  record->SetBody("Body of the log message");
+
+  // Attributes and resource support different types
+  record->SetAttribute("key0", false);
+  record->SetAttribute("key1", "1");
+  record->SetResource("key2", 2);
+  record->SetResource("key3", 3.142);
+
+  exporter->Export(nostd::span<std::unique_ptr<sdklogs::Recordable>>(&record, 1));
```
xukaren, line 72: Suggestion for this test: maybe cast `record` to an `ElasticSearchRecordable` object, so the `Get()` methods can be tested?

E.g:
```
auto es_record = unique_ptr<ElasticSearchRecordable >(static_cast<ElasticSearchRecordable *>(record.release()));

EXPECT_EQ(es_record["name"], "Timeout Log");
EXPECT_EQ(es_record["severity"], "Fatal");
etc.
```

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
MarkSeufert: yea, instead of storing a string representation fo the bool it is storing an actual bool to preserve the datatype

exporters/elasticsearch/test/es_log_exporter_test.cc
MarkSeufert: yea good idea! I was assuming the nlohmann JSON library would work but better safe than sorry

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
xukaren: oh cool cool, maybe `opentelemetry::nostd::get<bool>(value)` would work then? (nbd just wondering)

exporters/elasticsearch/test/es_log_exporter_test.cc
xukaren: yeah would be safer to test the `AttributeValue` types more thoroghly maybe?

jsuereth (approved):

PR #445, opened 2020-12-14
Http client factory

lalitb:
A http client factory interface to select the correct http client implementation based on availability.


ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,34 @@
...
+namespace client
+{
+
+class HttpClientFactory
+{
+public:
+#ifdef WITH_CURL
+  static std::shared_ptr<SessionManager> Create()
+  {
+
```
ThomsonTan, line 21: nit: remove the extra empty line?

ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,34 @@
...
+    return std::make_shared<ext::http::client::curl::SessionManager>();
```
ThomsonTan, line 22: Just wrap the `return` statement in `#if`?

ext/test/http/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ find_package(CURL)
 if(CURL_FOUND)
   set(CURL_LIBRARY "-lcurl")
   set(FILENAME curl_http_test)
+  add_compile_definitions(WITH_CURL)
```
ThomsonTan, line 5: Is this needed to apply to all header/cpp source files which include `http_client_factory.h`?

lalitb: Yes, needed for all `cpp` files which include it.  We can add it at root level `CMakeLists.txt`, but probably I felt it would be good to keep that clean. Or if there is better suggestion.

ThomsonTan: Probably we could generate a header file dynamically as `opentelemetry_build_flags`, and put `#define WITH_CURL 1` there when detected, and included the global header file in the above header file or other global place?

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: Yes can be done. This looks bit more cleaner in vscode, as it would show the specific `Create()` function highlighted based on available client library. But was my personal choice, and I can change if it looks ugly to read.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: done thanks.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
ThomsonTan: I have strong preference on this. I was thinking about that unifying the method signature indicates that it is the same function with different body instead of different functions for CURL and else.

ext/test/http/CMakeLists.txt
lalitb: Thanks for suggestion. If I understood correctly, we need to add the code for dynamic generation of header file within `CMakeFiles`  - Which would be similar to defining a macro ?

ThomsonTan: To define it as macro, I think we need make it global and propagate it to all the consumes which includes the above header file. For dynamic generated header file the option propagation is not an issue because the flag file will always be included.

I think this is not blocking and we could open an issue to track how to make it better.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: sure will change it if that's causing confusion.

lalitb: Done.

ext/test/http/CMakeLists.txt
lalitb: Sure - have created the issue here to trace:: https://github.com/open-telemetry/opentelemetry-cpp/issues/446

ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+#ifdef WITH_CURL
+  static std::shared_ptr<SessionManager> Create()
+  {
+    return std::make_shared<ext::http::client::curl::SessionManager>();
+  }
+#else
+  static std::shared_ptr<SessionManager> Create()
+  {
+    return std::shared_ptr<SessionManager>(nullptr);
```
jsuereth, line 26: is it better to fail compilation if no HTTP library is wired in?

lalitb: Correct. This should be the right behavior. Will change.

lalitb: This is done now. `HttpClientFactory::Create` method will now be missing if no HTTP library is plugged in.

ext/test/http/CMakeLists.txt
lalitb: I reverted it back to original ( and removed default method ) implementation based on comment from Josh : https://github.com/open-telemetry/opentelemetry-cpp/pull/445#discussion_r542423350. 
Let me know if this is not fine.

ThomsonTan (approved):

ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,27 @@
...
+namespace ext
+{
+namespace http
+{
+namespace client
+{
+class HttpClientFactory
+{
+public:
+#ifdef WITH_CURL
```
maxgolov, line 17: It looks a bit odd to have `#ifdef WITH_CURL` in a generic factor interface?

lalitb: Correct, as of now, there is only one `HttpClientFactory::Create()` implementation, and we would be adding more as we add support for further clients. And the right factory method should be picked during compile time based on where it is compiled. Do you see any issues here ?

lalitb:
@maxgolov @jsuereth Thanks for reviewing it. Please let me know if there are any further comments, or else it is good for merge ?

pyohannes (approved):

ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,27 @@
...
+  static std::shared_ptr<SessionManager> Create()
```
maxgolov, line 18: Sorry for belated response: my thinking here is that the `http_client_factory.h` itself may remain abstract. Any client that implements this interface must implement all generic methods. Or provide a no-op method in case if some method is not applicable to specific HTTP client.. Basically, what I think looks wrong is that we have `#ifdef` in the interface specification. It should not need any ifdefs.

What is the full set of methods you'd need for the `HttpClientFactory` interface? If `SessionManager` is unique to curl, should this method be moved to `CurlHttpClient` rather than adding ifdef here?

I'd like to refer to these interfaces as example:
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/http/HttpClientFactory.hpp
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/include/public/IHttpClient.hpp

This code is generic with 6 different implementations cross-plat, no ifdefs.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
```diff
@@ -0,0 +1,27 @@
...
+  {
+    return std::make_shared<ext::http::client::curl::SessionManager>();
```
maxgolov, line 20: What are the cases when we need a SessionManager, can it become a private detail of curl client, and be auto-refreshed as-needed? First time you trigger the request/response, you maintain a session. Maybe I'm missing some fundamental reason _why_ the session manager is unique to curl, and why is it required at all? Looking at `prior art`, we were not needing it elsewhere 😄 
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/http/HttpClient_Curl.cpp
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/http/HttpClient_Curl.hpp

ext/test/http/curl_http_test.cc
```diff
@@ -184,9 +185,10 @@ TEST_F(BasicCurlHttpTests, HttpResponse)
 TEST_F(BasicCurlHttpTests, SendGetRequest)
 {
   received_requests_.clear();
-  curl::SessionManager session_manager;
+  auto session_manager = http_client::HttpClientFactory::Create();
+  EXPECT_TRUE(session_manager != nullptr);
 
-  auto session = session_manager.CreateSession("127.0.0.1", HTTP_PORT);
+  auto session = session_manager->CreateSession("127.0.0.1", HTTP_PORT);
```
maxgolov, line 191: Is this to avoid passing the URL detail every time to `CreateRequest` ?

ext/test/http/curl_http_test.cc
```diff
@@ -184,9 +185,10 @@ TEST_F(BasicCurlHttpTests, HttpResponse)
...
   auto request = session->CreateRequest();
```
maxgolov, line 192: I was thinking more about the [Javascript model](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest) of it, where your request (handled by whatever client) is not bound to destination URL until you actually tell it to open / fire for a given URL, e.g.
```javascript
var oReq = new XMLHttpRequest();
oReq.addEventListener("load", reqListener);
oReq.open("GET", "http://www.example.org/example.txt");
oReq.send();
```

That way the HTTP client may provide concrete implementation, that is not session-aware. Then you pass the `HttpRequest` object to it, that object contains the destination URL detail. Then HTTP client executes your request with given params.. In that model you do not need a session manager.. Unless you want the session manager to handle some additional details, such as authentication?

ext/test/http/curl_http_test.cc
```diff
@@ -200,9 +202,10 @@ TEST_F(BasicCurlHttpTests, SendGetRequest)
 TEST_F(BasicCurlHttpTests, SendPostRequest)
 {
   received_requests_.clear();
-  curl::SessionManager session_manager;
+  auto session_manager = http_client::HttpClientFactory::Create();
+  EXPECT_TRUE(session_manager != nullptr);
 
-  auto session = session_manager.CreateSession("127.0.0.1", HTTP_PORT);
+  auto session = session_manager->CreateSession("127.0.0.1", HTTP_PORT);
   auto request = session->CreateRequest();
```
maxgolov, line 209: Shouldn't `request` be not even tied to particular implementation of a client? In a model where `request` is generic, you can create it, specify parameters, headers, destination URL host / port. Then fire it via client, which then will create  (or reuse) the connection, etc.. Basically Session Manager then is becoming totally unneeded. I'm trying to understand why we require it as a separate class. This could be detail hidden in the client itself, which should auto-manage the connection, SSL handshake, maintain the handshake for persistent connections (without giving much info to user about it), and should be able to handle incoming requests. The requests objects should not need to be client-specific: any HTTP client impl may be able to handle a generic request object that contains the destination, body, headers, etc.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: Nice to see reference of `cpp_client_telemetry` now that it is public  :)
So, session Manager is not unique to `curl` but an abstract class which is implemented for all client libraries. 
`opentelmetry::ext::http::client::SessionManager` is similar to `MAT_NS_BEGIN::IHttpClient`, and it manages multiple ongoing HTTP requests.
`opentelementry::ext::http::client::Session` logically represent a (possibly) ongoing http request.

http client factory implementation is similar to what is done in `cpp_client_telemetry` as here:
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/http/HttpClientFactory.cpp

In above, we have #ifdef for each client implementation.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: Nice to see reference of `cpp_client_telemetry` now that it is public  :)
So, session Manager is not unique to `curl` but an abstract class which should be implemented for all client libraries. `opentelmetry::ext::http::client::SessionManager` is similar to `MAT_NS_BEGIN::IHttpClient`, and it manages multiple ongoing HTTP requests.
`opentelementry::ext::http::client::Session` logically represent a (possibly) ongoing http request.

lalitb: Please see my response above.

ext/include/opentelemetry/ext/http/client/http_client_factory.h
lalitb: Please see the example of the http request flow here: https://github.com/open-telemetry/opentelemetry-cpp/blob/aa7b3c8f17918c71785ff9213d7e1f58c0de3867/ext/include/opentelemetry/ext/http/client/http_client.h#L11

```
  SessionManager sessionManager; // implementer can provide singleton implementation for it
  auto session = sessionManager.createSession("localhost", 8000);
  auto request = session->CreateRequest();
  request->AddHeader(..);
  SimpleResponseHandler res_handler;
  session->SendRequest(res_handler);
  session->FinishSession() // optionally in the end
  ...shutdown
  sessionManager.FinishAllSessions()
```

ext/test/http/curl_http_test.cc
lalitb: Yes that is also an option - We can get rid of `Session` object and have `Request` managed directly with `SessionManager`.  But I don't see that scope of this PR ? Although we already reviewed and approved http interface PR, I would be fine revisiting it again separately.

ext/test/http/curl_http_test.cc
lalitb: Correct.

ext/test/http/curl_http_test.cc
lalitb: `SessionManager` handles multiple `Session`. I understand we can get rid of `Session` object, and directly use `Request` object . But this PR changes still hold true as this return `SessionManager` object specific to client implementation.

maxgolov (approved):

PR #448, opened 2020-12-14
Sending sync request through http_client

lalitb:
Currently, http_client api supports sending async http requests, with error and response received through callback. For exporters, we need to have the status of http request returned from export function ( refer - https://github.com/open-telemetry/opentelemetry-cpp/pull/444#discussion_r542112168 ) . Adding a method to send sync request in client interface, along with it's curl implementation. 

Summary of changes:
2. Split `SessionManager` to `HttpClient` ( default async http client ) and `HttpClientSync` ( sync http client )
2. Sync Request Example:

GET:
```
  HttpClientSync httpClient;
  auto result = httpClient.Get(url);
  if (result){
    auto response = result.GetResponse();
  } else {
    std::cout << result.GetSessionState();
  }
```

POST:
```
  HttpClientSync httpClient;
  auto result = httpClient.Post(url, data, headers); // GET request
  if (result){
    auto response = result.GetResponse();
  } else {
    std::cout << result.GetSessionState();
  }
```

Pending Change:

2. Add test case for HTTP POST sync request.


ext/test/http/curl_http_test.cc
```diff
@@ -248,16 +248,46 @@ TEST_F(BasicCurlHttpTests, CurlHttpOperations)
   std::multimap<std::string, std::string, curl::curl_ci> m1 = {
       {"name1", "value1_1"}, {"name1", "value1_2"}, {"name2", "value3"}, {"name3", "value3"}};
   curl::Headers headers = m1;
-  curl::HttpOperation http_operations1(http_client::Method::Head, "/get", handler, headers, body,
-                                       true);
+  curl::HttpOperation http_operations1(http_client::Method::Head, "/get", handler,
```
ThomsonTan, line 251: Perhaps renaming this test as `CurlHttpAsyncOperations`.

lalitb: All  previous tests are async by default. Including `http::client::SendRequest()` method. Have suffixed `Sync` explicitly where needed, keeping defaults as async.

lalitb:
@maxgolov Need you review for this : ) Thanks.

ext/test/http/curl_http_test.cc
```diff
@@ -248,16 +248,46 @@ TEST_F(BasicCurlHttpTests, CurlHttpOperations)
...
+  EXPECT_EQ(response->GetStatusCode(), 200);
+  EXPECT_EQ(session_state, http_client::SessionState::Response);
+}
+
+TEST_F(BasicCurlHttpTests, SendGetRequestSyncTimeout)
+{
+  received_requests_.clear();
+  curl::SessionManager session_manager;
+
+  auto session =
```
maxgolov, line 284: If I need to send 2 synchronous requests, do I need to create two separate sessions?

Why do I have to specify session IP, port, and Uri separately - is there a reason to not having something like:

`oReq.open("GET", "http://www.example.org/example.txt")`

ext/include/opentelemetry/ext/http/client/http_client.h
```diff
@@ -37,6 +39,17 @@
...
   ...shutdown
   sessionManager.FinishAllSessions()
+
+Sync Request:
+
+  SessionMamager sessionManager;
+  auto session = sessionManager.createSession("localhost", 8000);
+  auto request = session->CreateRequest();
+  request->AddHeader(..);
+  SessionState session_state;
```
maxgolov, line 49: So this is not a `session` state - this is given `request` state?

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -153,6 +153,33 @@ class Session : public http_client::Session
     });
   }
 
+  virtual std::unique_ptr<http_client::Response> SendRequestSync(
```
maxgolov, line 156: I believe it would be much cleaner if you operate only on `Response` object rather than on `SessionState`. Let me elaborate - right now you have two possible states:
- success - in this case you return `Response`; or
- falure - in this case you force the user to test if `Response` ptr is nullptr, and then inspect the `SessionState` parameter passed by reference.. which is not really `Session` state (because `Session` may have multiple requests), but rather `Request` state.

So since in reality it is `Request` state, can you have this aggregated on `Response` object - making sure that you always return `Response`. It may be successful response or empty response, failed response, with an indication on `Response` object that it's a failed response. Then you avoid the nullptr entirely, you avoid checking on two separate objects.


Could you please clarify - if the original intent of `SessionManager` was to support more than one concurrent request at a time? If so, then `SessionState` is a misleading name for a synchronous request. Session may have more than one request, one failed, one successful, and `SessionState` name does not represent the actual `SessionManager` Session state.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -153,6 +153,33 @@ class Session : public http_client::Session
...
+    session_state = curl_operation_->GetSessionState();
+    if (curl_operation_->WasAborted())
+    {
+      session_state = http_client::SessionState::Cancelled;
+    }
+    if (curl_operation_->GetResponseCode() >= CURL_LAST)
+    {
+      // we have a http response
+      auto response      = std::unique_ptr<Response>(new Response());
+      response->headers_ = curl_operation_->GetResponseHeaders();
```
maxgolov, line 174: How do you propagate Response status code? If you are using int for that, then in Java it'd typically allocate `-1` for failed response status code, which can then be used to determine if request was actually processed, as in connected and got some response code from server.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -153,6 +153,33 @@ class Session : public http_client::Session
     });
   }
 
+  virtual std::unique_ptr<http_client::Response> SendRequestSync(
+      http_client::SessionState &session_state) noexcept override
+  {
+    is_session_active_ = true;
+    std::string url    = host_ + std::string(http_request_->uri_);
+    curl_operation_.reset(new HttpOperation(http_request_->method_, url, nullptr, RequestMode::Sync,
```
ThomsonTan, line 161: Could we set `RequestMode::Sync` in `SendSync()` or just deduce it from `SendSync()` and `SendAsync()`. Is the inconsistency between ctor and `SendSync()` handled?

ext/test/http/curl_http_test.cc
lalitb: > If I need to send 2 synchronous requests from two different threads, do I need to create two separate sessions?

`
Session` is a logical entity here for maintaining session id, ssl cert ( if needed ), callback object ( for async request) etc. HTTP is stateless protocol, so it doesn't makes sense to have ongoing session, and sending HTTP Requests over that. 

>Why do I have to specify session IP, port, and Uri separately - is there a reason to not having something like:
>oReq.open("GET", "http://www.example.org/example.txt")

Yeah, its doable, and but don't think it is related to current PR.

ext/include/opentelemetry/ext/http/client/http_client.h
lalitb: Correct.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
lalitb: `Response` object represents HTTP Response containing - body, header and code. Any error during http handshake are returned in `SessionState`. This is to keep it consistent with how we achieve it in async requests:
```
EventHandler::OnResponse(Response &);
EventHandler::OnError(SessionState);
Session::SendRequest(EventHandler &); // async
unique_ptr<Request> Session::SendRequestSync(SessionState &); // sync
```

`Session` is more of logical entity, and doesn't represent multiple ongoing http requests.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
lalitb: Currently we use `uint16_t;` for status_code

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
lalitb: We do need to know the actual sync mode in ctor. There are some initializations done in ctor, and the `sync mode` information is used to propagate the error/results to clients. 

>Is the inconsistency between ctor and SendSync() handled?

What kind of inconsistency we are talking about.

lalitb:
@maxgolov I have incorporated the review comments , and update the PR description with the changes and example, can you please review it once. Thanks :)

ext/test/http/curl_http_test.cc
lalitb: Ok, I have done the modified sync requests as below:

```
  HttpClient httpClient;
  auto result = httpClient.Get(url);
  if (result){
    auto response = result.GetResponse();
  } else {
    std::cout << result.GetSessionState();
  }
```

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
```diff
@@ -117,7 +117,7 @@ class ElasticsearchLogExporter final : public sdklogs::LogExporter
   ElasticsearchExporterOptions options_;
 
   // Object that stores the HTTP sessions that have been created
-  std::unique_ptr<ext::http::client::SessionManager> session_manager_;
+  std::unique_ptr<ext::http::client::HttpClient> session_manager_;
```
ThomsonTan, line 120: Rename the variable name to `http_client_` too?

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -179,20 +180,25 @@ class Session : public http_client::Session
...
+    // session_manager_.CleanupSession(session_id_);
+  }
+
 private:
   std::shared_ptr<Request> http_request_;
   std::string host_;
   std::unique_ptr<HttpOperation> curl_operation_;
   uint64_t session_id_;
-  SessionManager &session_manager_;
+  HttpClient &session_manager_;
```
ThomsonTan, line 193: Rename the variable to `http_client_` as well?

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -228,7 +234,60 @@ class SessionManager : public http_client::SessionManager
     sessions_.erase(session_id);
   }
 
-  ~SessionManager() { curl_global_cleanup(); }
+  http_client::Result Get(const nostd::string_view &url,
+                          http_client::Headers &headers) noexcept override
+  {
+    HttpOperation curl_operation(http_client::Method::Get, url.data(), nullptr, RequestMode::Sync,
```
ThomsonTan, line 240: Should `http_client::Method::Post` be used here?

ThomsonTan (approved):
LGTM with some minor comments.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
lalitb: Yes, copy-paste error. Need to add test for POST sync request too. Thanks for pointing out :)

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_exporter.h
lalitb: yeah, this needs to be done in both cc and h. I was thinking to change it once @maxgolov is fine with the current changes ( to avoid reverting back).  But thanks for noticing it. I will add this as pending change in PR description.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
lalitb: This is fixed now.

maxgolov (approved):

Issue #287, opened 2020-08-20
Simple Example Not Working


snehilchopra:
The [simple example](https://github.com/open-telemetry/opentelemetry-cpp/tree/master/examples/simple) does not seem to be working. 
I believe this is because the [span destructor](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/src/trace/span.cc#L72) is not being invoked, which in turn invokes the `span.End()` method.

As a workaround, one has to explicitly call the `span.End()` method.

However, this is causing a lot of memory leaks. I tried running the example with `bazel run --config=asan //example/simple:example_simple`, and the report is too long to be pasted here.

I believe this has something to do with the newly implemented Runtime context, as per the report.

Any insights on this would be really appreciated, thanks!

snehilchopra:
@pyohannes @reyang @maxgolov

pyohannes:
I think the reason is that our `Context` implementation is not memory safe. A gobal `context_handler_` is initialized, but never de-initialized.

I will look into sanitizing and stabilizing the `Context` part.

nadiaciobanu:
This affects all examples. As a workaround, I could add calls to `span.End()` in the examples. What are your thoughts?

nadiaciobanu:
I've created an OTLP exporter example using explicit calls to `span->End()` (https://github.com/open-telemetry/opentelemetry-cpp/pull/296). These calls should be removed once this issue is resolved.

jajanet:
BTW the zPages example is buggy too, which is related to running spans that don't end I believe

pyohannes:
`End()` calls are added for spans (this is required now). Also, #321 fixes the memory problems related to the global `context_handler_` variable.

PR #449, opened 2020-12-14
Remove unnecessary call on Span::End()

ThomsonTan:
`End()` calls are added to Span so explicit call is no longer necessary, see https://github.com/open-telemetry/opentelemetry-cpp/issues/287#issuecomment-699019923

### DIFF
--- a/examples/otlp/foo_library/foo_library.cc
+++ b/examples/otlp/foo_library/foo_library.cc
@@ -13,12 +13,12 @@ nostd::shared_ptr<trace::Tracer> get_tracer()
 
 void f1()
 {
-  auto span = get_tracer()->StartSpan("f1");
+  auto scoped_span = trace::Scope(get_tracer()->StartSpan("f1"));
 }
 
 void f2()
 {
-  auto span = get_tracer()->StartSpan("f2");
+  auto scoped_span = trace::Scope(get_tracer()->StartSpan("f2"));
 
   f1();
   f1();
@@ -27,7 +27,7 @@ void f2()
 
 void foo_library()
 {
-  auto span = get_tracer()->StartSpan("library");
+  auto scoped_span = trace::Scope(get_tracer()->StartSpan("library"));
 
   f2();
 }
--- a/examples/otlp/foo_library/foo_library.cc
+++ b/examples/otlp/foo_library/foo_library.cc
@@ -11,13 +11,9 @@ nostd::shared_ptr<trace::Tracer> get_tracer()
   return provider->GetTracer("foo_library");
 }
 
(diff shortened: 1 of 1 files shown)


jsuereth (commented):
Is there a test somewhere that verifies the behavior is still correct?

lalitb (commented):
Looks Good. As @jsuereth mentioned, would be good to have unit test for this which seems missing.

lalitb (approved):

examples/otlp/foo_library/foo_library.cc
```diff
@@ -11,13 +11,9 @@ nostd::shared_ptr<trace::Tracer> get_tracer()
   return provider->GetTracer("foo_library");
 }
 
-// TODO: Remove all calls to span->End() once context memory issue is fixed
-// (https://github.com/open-telemetry/opentelemetry-cpp/issues/287)
-
 void f1()
 {
   auto span = get_tracer()->StartSpan("f1");
-  span->End();
```
pyohannes, line 20: An explicit call to `End` is still necessary, not for the context memory issue, but because spans are not automatically ended anymore when the object goes out of scope (since `StartSpan` returns a `shared_ptr` instead of an `unique_ptr` now).

This should take care of ending the span:
```c++
 opentelemetry::trace::Scope scope(span);
```

ThomsonTan: Just for spans in this file, even they are `shared_ptr`, there should be only one owner which is the declared `auto span`, so destruction of `auto span` will destruct the `Span` object and `End()` it accordingly. So current code should work.

ThomsonTan:
> Is there a test somewhere that verifies the behavior is still correct?

@jsuereth @lalitb seems we don't have test to cover these examples while I verified it locally. If this is correct, then we'll need add some test to make sure the examples run correctly.

lalitb:
> @jsuereth @lalitb seems we don't have test to cover these examples while I verified it locally. If this is correct, then we'll need add some test to make sure the examples run correctly.

Correct, and I see you have created a issue for this.

examples/otlp/foo_library/foo_library.cc
pyohannes: You're right. I still think we should add the `Scope`, otherwise spans will not be parented correctly (no current span will ever be set, so no parent will be set on any of the spans).

This was not broken by your change, that was already broken before.

pyohannes (approved):

jsuereth (approved):
Manual testing for now is fine.  We should try to avoid adding more examples that require manual tests though :)

PR #450, opened 2020-12-14
Add explicit type conversion from std::chrono::milliseconds to long

ThomsonTan:
`http_conn_timeout_` is in type `std::chrono::milliseconds` which has more than 32-bit count as `Count()`. Add explicit conversion from this `std::chrono::milliseconds::Count()` to `long` as `long` could be in 32-bit.


lalitb (approved):

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -205,7 +205,7 @@ class HttpOperation
 
     /* wait for the socket to become ready for sending */
     sockfd_ = sockextr;
-    if (!WaitOnSocket(sockfd_, 0, http_conn_timeout_.count()) || is_aborted_)
+    if (!WaitOnSocket(sockfd_, 0, (long)http_conn_timeout_.count()) || is_aborted_)
```
jsuereth, line 208: Style question:

Do we prefer C++ style casts?  Should this be `static_cast<long>` vs. `(long)`

PR #451, opened 2020-12-15
Add changelog at api and sdk level, abi version for sdk

lalitb:
As per the Version stability oteps (open-telemetry/oteps#143):

- There need to be separate release package for api and sdk. So adding separate changelog for both. We may have to add changelog files for each of the exporters, which would be done afterwards.
- Need to take care for ABI stability guarantee for sdk ( without bumping up the major version). So added ABI version for sdk, along with its separate namespace. Separate PR would be needed to move all the sdk components within this namespace.
- Small fix in pre_release.sh script to make it work with older version of bash.


buildscripts/pre_release.sh
```diff
@@ -45,7 +45,7 @@ if [ ! -f $changelog_file ]; then
     exit 1
 fi
 
-if [ ! grep -q "## [Unreleased]" $changelog_file ]; then
+if [ ! grep -q "\#\# \[Unreleased\]" $changelog_file ]; then
```
ThomsonTan, line 48: nit: add `^` to be beginning of the patter to match from the beginning of the line?

lalitb: Fixed -added `^` in beginning and `$` in end to search for whole word.

lalitb:
@ThomsonTan @maxgolov @jsuereth  please review this, and if looks fine approve. We plan to use this for creating a first test release from master.

ThomsonTan (approved):

lalitb:
Merging these changes to create a test release. Please feel free to comment on this PR if there are some issues, I will fix them.

PR #457, opened 2020-12-16
Use opentelemetry-proto from released versions

bogdandrutu:

### DIFF
--- a/bazel/repository.bzl
+++ b/bazel/repository.bzl
@@ -45,10 +45,14 @@ def opentelemetry_cpp_deps():
 
     # OTLP Protocol definition
     maybe(
-        native.new_local_repository,
+        http_archive,
         name = "com_github_opentelemetry_proto",
         build_file = "//bazel:opentelemetry_proto.BUILD",
-        path = "third_party/opentelemetry-proto",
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -23,7 +23,3 @@
 	url = https://github.com/google/benchmark
 	branch = master
 
-[submodule "third_party/opentelemetry-proto"]
```
ThomsonTan, line 26: Is CMake build affected by this?

bogdandrutu: Will double check

bogdandrutu: Ok, because I don't know how to download an archive with cmake, I made bazel to use the released version, and point the submodule to the same tag.

lalitb: Can't we not just update the submodule version, and let cmake and bazel use that ( as they are doing right now ).

bogdandrutu: https://github.com/open-telemetry/opentelemetry-cpp/issues/461

maxgolov (approved):

.gitmodules
maxgolov: @lalitb - we have to update submodule version. Latest looks good to me.

bogdandrutu: the submodule was updated to point to the same tag release commit

jsuereth (approved):

PR #458, opened 2020-12-17
Use google benchmarks directly from released tags in bazel

bogdandrutu:
Looks like in tools there is a script that uses the submodule, for whatever reasons :)... leaving for the moment the submodule.

### DIFF
--- a/bazel/repository.bzl
+++ b/bazel/repository.bzl
@@ -7,9 +7,13 @@ def opentelemetry_cpp_deps():
     # Google Benchmark library.
     # Only needed for benchmarks, not to build the OpenTelemetry library.
     maybe(
-        native.local_repository,
+        http_archive,
         name = "com_github_google_benchmark",
-        path = "third_party/benchmark",
+        sha256 = "dccbdab796baa1043f04982147e67bb6e118fe610da2c65f88912d73987e700c",
+        strip_prefix = "benchmark-1.5.2",
+        urls = [
+            "https://github.com/google/benchmark/archive/v1.5.2.tar.gz",
+        ],
     )
 
     # GoogleTest framework.


bazel/repository.bzl
```diff
@@ -7,9 +7,13 @@ def opentelemetry_cpp_deps():
...
     maybe(
-        native.local_repository,
+        http_archive,
         name = "com_github_google_benchmark",
-        path = "third_party/benchmark",
+        sha256 = "dccbdab796baa1043f04982147e67bb6e118fe610da2c65f88912d73987e700c",
+        strip_prefix = "benchmark-1.5.2",
+        urls = [
+            "https://github.com/google/benchmark/archive/v1.5.2.tar.gz",
+        ],
```
lalitb, line 16: Thanks for the fix. Just wondering if it would be better to directly upgrade the submodule ?

bogdandrutu: The problem is that with submodule is hard to use with bazel, because I need to redefine the repository, with http archive this is no longer a problem, and the most annoying is the proto because it does not work without that.

bogdandrutu: When I say use is depending on this project from a third party

bogdandrutu: To clarify, right now I have to do this:

```
# Load OpentTelemetry-CPP dependency
http_archive(
    name = "com_github_opentelemetry_cpp",
    strip_prefix = "opentelemetry-cpp-7d7bd5247f0b546cb674afd82a6ef57be91a83d1",
    sha256 = "a2940df54fb0087715f96e4cb6c860987a3141100fb27fcf0871fe2638c82f7e",
    urls = [
        "https://github.com/open-telemetry/opentelemetry-cpp/archive/7d7bd5247f0b546cb674afd82a6ef57be91a83d1.tar.gz",
    ],
)

# Redefine the com_github_opentelemetry_proto because uses submodule
http_archive(
        name = "com_github_opentelemetry_proto",
        build_file = "@com_github_opentelemetry_cpp//bazel:opentelemetry_proto.BUILD",
        sha256 = "08f090570e0a112bfae276ba37e9c45bf724b64d902a7a001db33123b840ebd6",
        strip_prefix = "opentelemetry-proto-0.6.0",
        urls = [
            "https://github.com/open-telemetry/opentelemetry-proto/archive/v0.6.0.tar.gz",
        ],
)
```

So if we use always http_archive for all the bazel repos I don't have to do the redefinition.

bogdandrutu: https://github.com/open-telemetry/opentelemetry-cpp/issues/461

lalitb: Thanks for explaining, that makes sense to me now. This would be the issue for all the libraries bazel picks from submodule. Will look forward for comments from @jsuereth our bazel expert :)

jsuereth (changes requested):

maxgolov:
> Looks like in tools there is a script that uses the submodule, for whatever reasons :)

As Josh explained:
- `vcpkg` (which is a custom add-on package manager on top of CMake) is, unfortunately, does not allow to pin to specific release version or tag. It fetches what it considers `latest stable`, whatever that is included in vcpkg "ports" repo..
- whereas CMake itself - does not have a mechanism for downloading and installing packages.

Thus, submodules is the only reliable predictable mechanism to guarantee pinning to specific branch + commit id / or git tag of the mainline release of a dependency that works for CMake builds.

I do not have a strong opinion on Bazel builds, as long as we create some sort of mapping how Bazel artifact id corresponds to published git release / git tag, so that we can keep both build systems in-sync. Thus, producing predictable build artifacts irrespective of whether we build with bazel or CMake.

bazel/repository.bzl
bogdandrutu: Done.

jsuereth (approved):

lalitb (approved):

lalitb:
@bogdandrutu - sorry as this PR kept waiting for some time. Would it be possible for you resolve merge conflicts, so that we merge it immediately. Thanks.

PR #463, opened 2020-12-17
Change release version to 0.1.0

lalitb:
As per semver2 recommendation, first development release should start with 0.1.0:

https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

--
The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.
`--

So changing the initial version in CHANGELOG.md and version.h


bogdandrutu (approved):

ThomsonTan (approved):

maxgolov (approved):

maxgolov:
@lalitb - shouldn't we move this document to `/docs` folder, where the rest of markdown documents are located?

maxgolov:
Same comment applies to `RELEASING.md` ... otherwise our root folder becomes a collection of documents, with some in there, and some in the `docs`. Can we keep most of it under the `/docs`, with `README.md` in the root folder linking to a corresponding document in the `/docs`. Would that be more consistent?

lalitb:
> Same comment applies to `RELEASING.md` ... otherwise our root folder becomes a collection of documents, with some in there, and some in the `docs`. Can we keep most of it under the `/docs`, with `README.md` in the root folder linking to a corresponding document in the `/docs`. Would that be more consistent?

@maxgolov  - I was trying to follow the repo structure maintained in other opentelemetry projects - where CONTRIBUTING.md, RELEASING.md and CHANGELOG.md are at root level, and rest of the project docs in `docs` folder. But I agree with you, it would be much cleaner to move them under `doc` folder. Will check if there is any specific guideline for opentelemetry, and accordingly move them in `docs`. `api\CHANGELOG.md` and `sdk\CHANGELOG.md` would remain as it is. Also need to fix `buildscripts` folder. Will create a separate cleanup PR for all these.

PR #467, opened 2020-12-18
Include CMAKE_TOOLCHAIN_FILE explicitly to enable find_package in vcpkg

ThomsonTan:
`CMAKE_TOOLCHAIN_FILE` takes effect only when it is set before `project(...)` statement in `CMakeLists.txt`, explicit `include()` is needed if set after that, or else it will not help resolve `find_package(...)` like form vcpkg.

Also unified `WITH_PROTOBUF` into `WITH_OTLP` because OTLP requires both gRPC and protobuf.


CMakeLists.txt
```diff
@@ -40,19 +39,21 @@ if(MSVC)
 endif()
 
 if(WITH_OTLP)
-  set(WITH_PROTOBUF ON)
-endif()
-
-if(WITH_PROTOBUF)
   set(protobuf_MODULE_COMPATIBLE ON)
-  find_package(Protobuf)
+  find_package(protobuf)
```
maxgolov, line 43: I think latest actually requires `Protobuf` with capital P?

ThomsonTan: Restored the casing for now. Both should work, `protobuf` is probably better as `protobuf-config.cmake` comes from protobuf project while `FindProtobuf.cmake` is provided as built-in module by cmake.

maxgolov: My comment based on what's written in CMake built-in, and I hit this issue with casing elsewhere on older Ubuntu-14.xx .. thus, the caution. Maybe it doesn't matter.

maxgolov (approved):
I have not tested locally. But looks okay.

lalitb (approved):

maxgolov:
@jsuereth - Josh, if you have no objections, would you sign-off on this CMake PR?

jsuereth (approved):

PR #469, opened 2020-12-18
Correct the doc about that approver is not expected to merge PR

ThomsonTan:
According to our policy, only maintainers are permitted to merge PRs.


jsuereth (approved):

lalitb:
Ideally it would good to have both approver and maintainers permissions to merge once there are required number of approval(s). This will also ensure that PRs are not stranded for long. Right now this policy seems to be driven from open telemetry community:

https://github.com/open-telemetry/community/blob/master/docs/how-to-configure-new-repository.md#policies
>Repository MUST enforce Restrict who can push to matching branches setting to only allow the members of foo-maintainers to push to the main branch.

Thanks for the correcting the document.

PR #470, opened 2020-12-18
Remove extra spaces which renders the items as code block

ThomsonTan:
Remove extra spaces in naming items because the extra spaces (more than 4) make the items being rendered as code block instead of list items.


lalitb (approved):
Thanks for the fix.

jsuereth (approved):

PR #476, opened 2020-12-21
Correctly initialize the curl::Session when a schema is given

pyohannes:
The initialization of a `curl::Session` did not work for cases where a scheme was given with the host name.

This PR adds a fix and tests.


lalitb (approved):
Thanks for fixing it.

ThomsonTan (approved):

PR #481, opened 2020-12-22
Resolve NOMINMAX issues where possible

maxgolov:
The issue is that `Windows.h` redefines both `min` and `max` as macros.

There are several fixes possible:
- targeted fix: surround the usage of `std::min` and `std::max` with parenthesis; OR
- globally define `NOMINMAX` for the entire project in CMake; OR
- in those modules where it is known that `std::*` is going to be used, `#define NOMINMAX` right before the inclusion of Windows.h

Since there were two recent hits @mishal23 @jsuereth , I'm applying targeted fix to avoid this issue wherever possible.

Not defining `NOMINMAX` globally, as this may lead to other interesting side-effects inside the bowels of Windows SDK 😁

### DIFF
--- a/api/include/opentelemetry/common/spin_lock_mutex.h
+++ b/api/include/opentelemetry/common/spin_lock_mutex.h
@@ -6,6 +6,17 @@
 
 #include "opentelemetry/version.h"
 
+#if defined(_MSC_VER)
+#  ifndef NOMINMAX
+#    define NOMINMAX
+#  endif
+#  include <Windows.h>
+#elif defined(__i386__) || defined(__x86_64__)
+#  if defined(__clang__)
+#    include <emmintrin.h>
+#  endif
+#endif
+
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace common
 {
--- a/api/include/opentelemetry/nostd/string_view.h
+++ b/api/include/opentelemetry/nostd/string_view.h
@@ -87,7 +87,7 @@ public:
 
   int compare(string_view v) const noexcept
   {
-    size_type len = std::min(size(), v.size());
+    size_type len = (std::min)(size(), v.size());
     int result    = Traits::compare(data(), v.data(), len);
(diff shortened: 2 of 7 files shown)


api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -6,6 +6,17 @@
 
 #include "opentelemetry/version.h"
 
+#if defined(_MSC_VER)
+#  ifndef NOMINMAX
+#    define NOMINMAX
+#  endif
+#  include <Windows.h>
```
ThomsonTan, line 13: C:\GH\otcpp\ext\include\opentelemetry\ext\http\client\curl\http_operation_curl.h(208):

maxgolov: Sorry Tom, what do you mean by `C:\GH\otcpp\ext\include\opentelemetry\ext\http\client\curl\http_operation_curl.h(208)` ?

maxgolov: Reason why I'm adding this is that Josh would need this in his PR that is failing. Related PR: #443

maxgolov: > if the user could include windows.h before include our header file

Yeah, then it will break for those users. We most likely need to add some blurb in our documentation, that since our SDK uses range limits extensively in public API, that means we should tell users that they must revisit their usage of `min` and `max` macros, to make sure we are not breaking them on Windows.

maxgolov: I tried fixing this where possible. But there are some spots, where our SDK defines class methods called `min` and `max` - that will surely break if our Windows customers don't define `NOMINMAX` themselves for their project.

jsuereth (approved):

lalitb (approved):

Issue #442, opened 2020-12-13
Removing and refactoring `SpanDataAttributeValue`


xukaren:
Currently in the SDK, `SpanDataAttributeValue`, defined in [trace/attribute_utils.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/attribute_utils.h) has the same implementation as `OwnedAttributeValue` in [common/attribute_utils.h](https://github.com/open-telemetry/opentelemetry-cpp/pull/438/commits/b28a759d0d095d61712ecc9111daa835373fc091#diff-9e843c3a348e462e97fdd4ba3c99136694e8f9acc04c382b40b4e5d7d9758be0). 

The redefinition is redundant for traces, so it can be removed, and all usages of `SpanDataAttributeValue` can be refactored to use `OwnedAttributeValue`.

lalitb:
#482 fixes this.

PR #482, opened 2020-12-22
Use attribute_utils.h from common instead of trace

xukaren:
This PR:
* Moves contents of trace/attribute_utils.h to common/attribute_utils.h
* Moves trace/attribute_utils_test.cc to common/attribute_utils_test.cc
* Refactors SpanDataAttributeValue to OwnedAttributeValue
* Changes affected class/struct names from namespace  `trace::` to `common::`

Resolves issue #442

cc @MarkSeufert @alolita


sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -42,8 +65,12 @@ struct AttributeConverter
   OwnedAttributeValue operator()(double v) { return OwnedAttributeValue(v); }
   OwnedAttributeValue operator()(nostd::string_view v)
   {
-    return OwnedAttributeValue(std::string(v));
+    return OwnedAttributeValue(std::string(v.data(), v.size()));
```
ThomsonTan, line 68: Are `std::string(v)` and `std::string(v.data(), v.size())` just the same?

ThomsonTan (approved):

lalitb (approved):

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Made the change to follow Max's PR [here](https://github.com/open-telemetry/opentelemetry-cpp/commit/1e7b9d8a950649809d3957a00e230e43b59a1c86#diff-8ac8b11ea3bd653dfa6b11b6f68ae56fcc246a90e4df905268f73ef65a32e85cR52) (cc @maxgolov )

lalitb:
@xukaren  - Can you please rebase it, so that we can merge the PR. Thanks : )

xukaren:
Does the CI need to be rerun to fix the `trace.CircularBufferTest.Simulation` test?

lalitb:
> Does the CI need to be rerun to fix the `trace.CircularBufferTest.Simulation` test?

Yes, have done that now. thanks.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
pyohannes: Both should have the same effect. The first form should use [this operator](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/nostd/string_view.h#L58).

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -42,8 +65,12 @@ struct AttributeConverter
...
   }
+  OwnedAttributeValue operator()(const char *s) { return OwnedAttributeValue(std::string(s)); }
```
pyohannes, line 70: This is unnecessary, as it's covered by `OwnedAttributeValue operator()(nostd::string_view v)`.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -21,6 +38,12 @@ using OwnedAttributeValue = nostd::variant<bool,
                                            uint64_t,
                                            double,
                                            std::string,
+#ifdef HAVE_CSTRING_TYPE
+                                           const char *,
```
pyohannes, line 42: Why do we need this? I never see us storing a `const char*` in this variant (as far as I can see, we always convert to `std::string`).

pyohannes (commented):
One question, but otherwise this looks good. +1 for the name `OwnedAttributeValue`.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Yeah at the moment, I don't think it's explicitly being used by `log_record` or `span_data`. Maybe there would be a future use? Not sure what the macro `HAVE_CSTRING_TYPE` will be used for. This was added in PR #374. (--cc @maxgolov)

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Ah okay, I'll revert it to the old `OwnedAttributeValue(std::string(v));` implementation then

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: This would also need to be changed in [attribute_value.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/common/attribute_value.h#L21)

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Would it be better to keep this one, and remove the `OwnedAttributeValue operator()(nostd::string_view v)`? Seems this one would be more versatile (e.g. supporting the CSTRING TYPE as defined in the macro)

lalitb: I think `OwnedAttributeValue operator()(nostd::string_view v)` should accept `const char *s` as argument. We should be good removing `OwnedAttributeValue operator()(const char *s)`.

sdk/test/common/BUILD
```diff
@@ -93,3 +93,14 @@ cc_test(
...
     ],
 )
+
+cc_test(
+    name = "attribute_utils_test",
+    srcs = [
+        "attribute_utils_test.cc",
+    ],
+    deps = [
+        "//sdk/src/trace",
```
jsuereth, line 103: This likely only need to depend on `//sdk/src/common`, I don't see anything trace specific in the test.

jsuereth (changes requested):

sdk/include/opentelemetry/sdk/common/attribute_utils.h
pyohannes: Yes, it exists in `AttributeValue`. But `AttributeConverter` converts it to a `std::string` when an `AttributeValue` is converted to an `OwnedAttributeValue`, so the `const char*` slot in `OwnedAttributeValue` will never get used.

sdk/test/common/BUILD
xukaren: Yup good catch, fixed it to just use ` //sdk:headers` which follows what is done for `atomic_unique_ptr_test`

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Gotcha, removed this from `OwnedAttributeValue`

sdk/include/opentelemetry/sdk/common/attribute_utils.h
xukaren: Gotcha, removed this method

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
```diff
@@ -110,16 +111,16 @@ class OStreamSpanExporter final : public sdktrace::SpanExporter
 
 #endif
 
-  void print_value(sdktrace::SpanDataAttributeValue &value)
+  void print_value(sdkcommon::OwnedAttributeValue &value)
```
maxgolov, line 114: There's something wrong about this name. I liked the previous name better. If we want to explain the semantics of some class, it has to be written in the class docs, e.g. `Owned` is not a good prefix for a common class.

xukaren: Instead, what if we simply named it `AttributeValue`, which follows the API's version of `AttributeValue` in [attribute_values.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/api/include/opentelemetry/common/attribute_value.h)? It could be referenced with diferent namespaces: `sdk::common::AttributeValue` and `opentelemetry::common::AttributeValue` respectively, like what's done for logs/traces/metrics.

lalitb: @maxgolov  @pyohannes to comment, as I understand johannnes liked the idea of naming it as `OwnedAttributeValue `.

pyohannes: I have no strong preference on this. I like that it's named different than the API variety, but it can be something different than `Owned`. I find it slightly more confusing to have it distinguished just by the namespace name, but I don't have a strong opinion on that either. I think it's not very important, as it's mostly an SDK internal concept (most exporters using their own Recordables will most likely not use it).

pyohannes (approved):

lalitb:
@xukaren Can you please rebase this PR, and ensure all comments are resolved using "Resolve Conservation". We can then merge this PR. Thansk.

xukaren:
@lalitb - Done

PR #486, opened 2020-12-23
Remove unnecessary include_directories for project root

ThomsonTan:
Header files are grouped by API/SDK, and there is no reason to include a file based on project root path.


lalitb (approved):

maxgolov:
What is the main reason why this change is needed, i.e. what is broken exactly if this change is not applied?

ThomsonTan:
> What is the main reason why this change is needed, i.e. what is broken exactly if this change is not applied?

This doesn't break anything, but it is also not used as we moved all header file to `include` subfolders in varies places.

pyohannes (approved):

jsuereth (approved):

PR #490, opened 2020-12-24
Build opentelemetry-proto as static instead of object library

ThomsonTan:
Build protobuf generated files and skeleton as object library could requires adding it directly to `add_executable(...)` or `add_library(...)` instead of linking it via `target_link_libraries(...)`. Build these files as static library to better fit linking as library.

### DIFF
--- a/cmake/opentelemetry-proto.cmake
+++ b/cmake/opentelemetry-proto.cmake
@@ -119,7 +119,7 @@ add_custom_command(
 include_directories("${GENERATED_PROTOBUF_PATH}")
 
 add_library(
-  opentelemetry_proto OBJECT
+  opentelemetry_proto STATIC
   ${COMMON_PB_CPP_FILE}
   ${RESOURCE_PB_CPP_FILE}
   ${TRACE_PB_CPP_FILE}


maxgolov (approved):

cmake/opentelemetry-proto.cmake
```diff
@@ -119,7 +119,7 @@ add_custom_command(
 include_directories("${GENERATED_PROTOBUF_PATH}")
 
 add_library(
-  opentelemetry_proto OBJECT
+  opentelemetry_proto STATIC
```
lalitb, line 122: How will this work if `BUILD_SHARED_LIBS` flag is set while compilation. Will the last few lines in this file work fine:
```
if(BUILD_SHARED_LIBS)
  set_property(TARGET opentelemetry_proto PROPERTY POSITION_INDEPENDENT_CODE ON)
endif()
```
Also, it seems we can use this object library in `add_executable` as below:
```
add_executable( foo $<TARGET_OBJECTS:opentelemetry_proto>)
```

jsuereth (approved):

cmake/opentelemetry-proto.cmake
ThomsonTan: As per the doc of [add_library](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library), `BUILD_SHARED_LIBS` provides default library type (`ON` => `SHARED`, `OFF` => `STATIC`) if no explicit library type is given to `add_library`, but we set it to either `OBJECT` or `STATIC`, so it should not affect the behavior of `add_library`.

For the setting of `POSITION_INDEPENDENT_CODE`, it is on by default for `SHARED`, but it should be fine to apply this to `OBJECT` or `STATIC`, but not necessary (the quality of compiled code may be lower than removing this option. Also I think we don't need build `opentelemetry-proto` as `SHARED` as it is for OTLP, I'll open an issue for further discussion on this.

Yes, `add_executable( foo $<TARGET_OBJECTS:opentelemetry_proto>)` will also work, but that requires we change all the references to `opentelemetry-proto` which is also different than the usual pattern.

PR #492, opened 2020-12-30
Add json.hpp dependency as submodule

maxgolov:
Add [nlohmann-json](https://github.com/nlohmann/json/) v3.9.1 (latest release) as a submodule for CMake build system. Currently for CMake on Windows we install it using `vcpkg install nlohmann-json:x64-windows`. And it generally works okay for Windows.

But there are two issues with that:
- we need a nicer cross-plat way to fetch `nlohmann/json` that does not necessarily require vcpkg
- we can't specify exact version of `nlohmann/json` via vcpkg since it does not (currently yet) support versioning.

So in order to provide _predictable_ build environment pinned to specific version of json.hpp , we add it here.

I'll send a separate PR that allows to determine if we use custom user-provided, vcpkg-provided, OS-provided (e.g. Ubuntu `nlohmann-json3-dev` package), OR the built-in git submodule version. That is to allow for more build-time flexibility.

However, for the official build and for predictable CI, ideally we should rely on git submodule -provided version added here. This will be similar in spirit to how Bazel is managing its build deps by pinning to specific commit ID.


.gitmodules
```diff
@@ -27,3 +27,8 @@
 	path = third_party/opentelemetry-proto
 	url = https://github.com/open-telemetry/opentelemetry-proto
 	branch = master
+
+[submodule "third_party/nlohmann-json"]
+	path = third_party/nlohmann-json
+	url = https://github.com/nlohmann/json
+	branch = master
```
ThomsonTan, line 34: Do we need specify the branch with tag like 3.9.1?

ThomsonTan (approved):
LGTM

.gitmodules
ThomsonTan: Also do we need add this to include path to activate this submodule?

lalitb:
Thanks for bringing this dependency,  We need this now at multiple places : )
 Just wondering if there is some other way of bringing this dependency - as it is adding up to ~250 mb size:
```
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh third_party/
22M     third_party/
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh third_party/nlohmann-json/
15M     third_party/nlohmann-json/
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh .git/modules/third_party/
254M    .git/modules/third_party/
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh .git/modules/third_party/nlohmann-json/
239M    .git/modules/third_party/nlohmann-json/
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$
```

Not sure if it would be much effective to include shallow checkout to remove `.git` history, or using some release tracking repo like https://github.com/astoeckel/json or https://github.com/azadkuh/nlohmann_json_release

.gitmodules
maxgolov: > Also do we need add this to include path to activate this submodule?

Yes, and I will handle it separately.

maxgolov: > Do we need specify the branch with tag like 3.9.1?

No, because `nlohmann-json` repo has two branches - `master` and `develop`. We point to `master` branch and the submodule actually points to release tag `v3.9.1`

maxgolov:
> Just wondering if there is some other way of bringing this dependency - as it is adding up to ~250 mb size

Prometheus exporter adds much more diskspace usage with its recursive dependencies. And gRPC adds 15-20+ minutes of build time.. nevertheless, we consume them. It's about consistency. I don't think we need to do anything special for json.hpp.

maxgolov:
@lalitb - there is something wrong with your numbers. It's 13MB, not 239MB. Prometheus CPP with recursive deps is much larger.

```console
maxgolov@MaxOMEN:/mnt/c/work/opentelemetry-cpp.main/third_party$ du -hs * | sort -hr
57M     prometheus-cpp
13M     nlohmann-json
4.4M    googletest
928K    benchmark
588K    ms-gsl
120K    opentelemetry-proto
4.0K    boost
0       BUILD
```

lalitb (approved):

lalitb:
> @lalitb - there is something wrong with your numbers. It's 13MB, not 239MB.

The actual submodule repo size is ~15 mb:
```
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh third_party/nlohmann-json/
15M     third_party/nlohmann-json/
```

But it's .git history size is around 239 mb.
```
labhas@MININT-47CLAOB:/home/labhas/ot/opentelemetry-cpp$ du -sh .git/modules/third_party/nlohmann-json/
239M    .git/modules/third_party/nlohmann-json/

```
So was wondering if there is way to only add repo source, and not all the history. But not a blocker here, we can see it later if possible.

maxgolov:
@lalitb - I got what you are saying. I don't know if recursive cloning on GitHub would clone only single branch? Since we specify the branch. My point is that the effect of adding this submodule is miniscule when compared to build time of some other deps, or of their combined build artifacts (for vcpkg on Windows we add up to 8GB, most of it comes from gRPC, zlib, openssl, http client, tooling, etc.).

lalitb:
@jsuereth - I am taking liberty to merge couple of small cmake and documentation PRs . Please comment on them if you see find some issues on them, to be fixed as separate PR.

PR #498, opened 2021-01-04
Don't override CURL_LIBRARY found by find_package

ThomsonTan:
Cmake `find_package` will set `CURL_LIBRARY` to the correct libraries and options, no need to override it.


lalitb (approved):
Thanks for fixing it. This was my mistake :)

Issue #334, opened 2020-09-18
Adding resource to TracerProvider


ThomsonTan:
OpenTelemetry [Resource SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md) mentioned associating resource to `TraceProvider`, which seems missing the current [TracerProvider](https://github.com/open-telemetry/opentelemetry-cpp/blob/5d1898c01f779d0c75f432f23c4a651ad29f1431/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L18) implementation. Actually it seems [resource](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/resource/TBD) is not implemented yet.

> When used with distributed tracing, a resource can be associated with the TracerProvider when the TracerProvider is created

jsuereth:
A summary of `Resource` (and `InstrumentationLibrary`) design so far.

Tl;DR: Current proposal is to:

- [ ] Update `Tracer` so that it has an instance of `InstrumentationLibrary` (a name/version set of strings)
- [ ] Update `TracerProvider` to abide by spec regarding creation of Tracers.
- [ ] Update `SpanProcessor` and `Exporter` in trace SDK to leverage `ExportableSpan` (or some new type that holds both the `Recordable` and a reference to `Tracer`.   This may have larger implications design/ownership in `BatchSpanProcessor`.  The class is ALMOST:
  ```
  struct ExportableSpan {
       std::unique_ptr<Recordable> recordable;
       std::shared_ptr<Tracer> tracer;
  }
   ```
- [ ] Update OTLP exporter.
  - *ideally* we assume ONE resource for all spans seen due to new ownership rules.  If not we need to group spans by resource.
  - We can sort by `Tracer` in some way to be able to group spans by `InstrumentationLibrary` as per the protocol.
  - We can pull both `Resource` and `InstrumentationLibrary` from the newly attached `Tracer` on `ExportableSpan` (or whatever we call it).
  

A very messy solution can be seen in this PR: #590

For context/comparison the previous proposal from @lalitb:

- We add methods to `Recordable` which allow specifying resource (and InstrumentationLibrary, i.e. the name/version specified when creating a Tracer)
- When creating a span, immediately set the known `Resource` and `InstrumentationLibrary` attributes on `Recordable`
- non-OTLP implementations of `Recordable` implement these new methods as no-ops.
[...]

PR #502, opened 2021-01-05
Resource sdk Implementation

lalitb:
Initial release of  [Resource SDK ](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md).
 
Supported -
 - Create resource
 - Merge resource
 - Empty resource creation
 - Base resource detection package.
 - Specify resource information from env variable ( OTEL_RESOURCE_ATTRIBUTES  )
 - Integrating with `TracerProvider ` and `Traces`

To be added as separate PR:
 - Adding in `SpanData` and propagating to exporters.
 - Resource detection logic from multiple resources, and merging all.

#334


sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,70 @@
...
+  template <class T,
+            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
+  Resource(const T &attributes) noexcept : Resource(common::KeyValueIterableView<T>(attributes))
+  {}
+
+  Resource(const std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>>
+               attributes) noexcept
+      : Resource(nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+            attributes.begin(), attributes.end()})
+  {}
```
pyohannes, line 36: What's your opinion about replacing those with a single:
```c++
Resource(const std::map<std::string, opentelemetry::sdk::trace::SpanDataAttributeValue> attributes) noexcept;
```

As resources are an SDK only concept, we don't need to worry about ABI compatibility here.

Also, we should move the `attribute_utils.h` (containing the `AttributeMap`) from `trace` to `common`. Resources will be used by metrics as well.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,74 @@
...
+  std::istringstream iss(attributes);
+  std::string token;
+  while (std::getline(iss, token, ','))
+  {
+    size_t pos        = token.find('=');
+    std::string key   = token.substr(0, pos);
+    std::string value = token.substr(pos + 1);
+    resource_attributes.SetAttribute(key, value);
+  }
+  return std::unique_ptr<Resource>(new Resource(resource_attributes));
```
pyohannes, line 69: I think it would be better to have this logic in the `OTELResourceDetector`, as it's related to the format of the environment variable.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,74 @@
...
+  if (attributes.size() > 0)
+  {
+    Resource input_resource(attributes);
+    auto merged_resource = default_resource.Merge(input_resource);
+    return merged_resource->Merge(*(OTELResourceDetector().Detect()));
+  }
+  return default_resource.Merge(*(OTELResourceDetector().Detect()));
+}
+
+std::unique_ptr<Resource> Resource::CreateEmpty()
```
pyohannes, line 52: I think when we start to use resources, we have to make them shared pointers. Resources are supposed to be attached to _every_ span. We don't want to have unique pointers there, but share a common resource with a shared pointer.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,74 @@
...
+{
+  return Create({});
```
pyohannes, line 54: Maybe we can use a static singleton for the default resource, like Java does: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L53

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,70 @@
...
+
+  const opentelemetry::sdk::trace::AttributeMap &GetAttributes() const noexcept;
+
+  std::unique_ptr<Resource> Merge(const Resource &other);
+
+  static std::unique_ptr<Resource> Create(
```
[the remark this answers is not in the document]
pyohannes: I see, probably because of creating unique pointers. Should we make the constructors private then, to prevent the creation of unmanaged resources?

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,44 @@
...
+    std::unordered_map<std::string, opentelemetry::sdk::trace::SpanDataAttributeValue>;
+
+class Resource
+{
+public:
+  Resource(const ResourceAttributes &attributes = ResourceAttributes()) noexcept;
+
+  const ResourceAttributes &GetAttributes() const noexcept;
+
+  std::shared_ptr<Resource> Merge(const Resource &other);
```
jsuereth, line 30: Why not just return a stack-allocated resource here?  Wouldn't that be a simpler api?

Similar for `Create`.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
```diff
@@ -0,0 +1,27 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+
+class ResourceDetector
+{
+public:
+  virtual std::shared_ptr<opentelemetry::sdk::resource::Resource> Detect() = 0;
```
jsuereth, line 16: Same nit:  Is there a reason to return a shared_ptr here?  AFAIK Resource detection happens once and generally isn't passing pointers between different memory owners.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,67 @@
...
+
+std::shared_ptr<Resource> Resource::Create(const ResourceAttributes &attributes)
+{
+  auto default_resource = Resource::GetDefault();
+
+  if (attributes.size() > 0)
+  {
+    Resource tmp_resource(attributes);
+    auto merged_resource = tmp_resource.Merge(default_resource);
+    return merged_resource->Merge(*(OTELResourceDetector().Detect()));
```
jsuereth, line 41: Nit:

Would it make sense to have both the OTEL resource + default resource be statics within this function?

```
{
  static auto default_resource = Resource::GetDefault();
  static auto otel_env_resource = OTELResourceDetector().Detect();
  ...
}

This would avoid re-running the OTEL resource detection for every created resource.

File: sdk/include/opentelemetry/sdk/resource/resource.h [hunk 9]
```diff
@@ -0,0 +1,44 @@
...
+namespace resource
+{
+
+using ResourceAttributes =
+    std::unordered_map<std::string, opentelemetry::sdk::trace::SpanDataAttributeValue>;
+
+class Resource
+{
+public:
+  Resource(const ResourceAttributes &attributes = ResourceAttributes()) noexcept;
```
jsuereth, line 26: The specification says that *ANY* created resource should go through the environment variable additions + default logic you have in `Create`.   Should we hide this constructor to only the `OTELResourceDetector` + `getDefault` ?

jsuereth (changes requested):

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: There is challenge to do this in C++. Ideally, I would like to make constructor as private, and make `OTELResourceDetector` as friend class of `Resource`, but we actually need all derivations of `ResourceDetector` access to this constructor. 
(opentelementry-java)[https://github.com/open-telemetry/opentelemetry-java/blob/f9bf909372a63bd048cfa58a384eba92005b2277/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L30] does it using `AutoValue` (https://github.com/google/auto/tree/master/value) annotation, and (opentelemetry-dotnet)[https://github.com/open-telemetry/opentelemetry-dotnet/blob/7a91923fb7a1ee6fbc75b6fdc145137e93b38917/src/OpenTelemetry/Resources/Resource.cs#L36]  does it using `internal` keyword.

sdk/include/opentelemetry/sdk/resource/resource.h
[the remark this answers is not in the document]
lalitb: There is challenge to do this in C++. Ideally, I would like to make constructor as private, and make `OTELResourceDetector` as friend class of `Resource`, but we actually need all derivations of `ResourceDetector` have access to this constructor. 
[opentelementry-java](https://github.com/open-telemetry/opentelemetry-java/blob/f9bf909372a63bd048cfa58a384eba92005b2277/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L30) does it using `AutoValue` annotation, and [opentelemetry-dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/blob/7a91923fb7a1ee6fbc75b6fdc145137e93b38917/src/OpenTelemetry/Resources/Resource.cs#L36) does it using `internal` keyword.

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: Good point. I have made it directory using STL class unordered_map now.  And `AttributeMap` is moved to `common` as part of another PR.

sdk/src/resource/resource.cc
lalitb: Correct. Moved it to `OTELResourceDetector`

sdk/src/resource/resource.cc
lalitb: Good point. Have changed it to use `shared_ptr`,

sdk/src/resource/resource.cc
lalitb: Good point. Have changed it to use static singleton now.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
```diff
@@ -0,0 +1,27 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+
+class ResourceDetector
+{
+public:
+  virtual std::unique_ptr<opentelemetry::sdk::resource::Resource> Detect() = 0;
```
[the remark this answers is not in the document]
pyohannes: I think the motivation behind that is to avoid excessive copying of a resource. A resource might be appended to each span. Having a `shared_ptr` (I think a unique pointer will not do) enables us to avoid copying the resource and its attributes.

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: Good point, 
Using `std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue` directly, and getting rid of `AttributeMap`.

sdk/src/resource/resource.cc
lalitb: Correct. Have changed this accordingly now.

sdk/src/resource/resource.cc
lalitb: Right. Have changed `std::unique_ptr` to `std::shared_ptr_ptr` now.

sdk/src/resource/resource.cc
lalitb: Good point. Using static local for `GetDefault()` and `GetEmpty()` now.

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,70 @@
...
+
+  Resource(const std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>>
+               attributes) noexcept
+      : Resource(nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+            attributes.begin(), attributes.end()})
+  {}
+
+  const opentelemetry::sdk::trace::AttributeMap &GetAttributes() const noexcept;
+
+  std::unique_ptr<Resource> Merge(const Resource &other);
```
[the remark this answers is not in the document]
lalitb: Sure. Will add the method documentation.

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: As it is possible to share resources across multiple spans, it would be better to use `shared_ptr` and avoid un-necessary copying.

sdk/include/opentelemetry/sdk/resource/resource.h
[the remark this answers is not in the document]
lalitb: Ok. I have now made constructor as protected ( so that it can be used in test cases ), and make `OTELResourceDetector` as friend class. It seems we don't need to invoke `Resource` constructor in any other detector, so we should be fine.

sdk/include/opentelemetry/sdk/resource/resource.h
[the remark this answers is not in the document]
lalitb: I have added the documentation now.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
lalitb: As mentioned in earlier comment - As it is possible to share resources across multiple spans, it would be better to use shared_ptr and avoid un-necessary copying.

sdk/src/resource/resource.cc
lalitb: Good point, I have changed it accordingly.
```
  static auto otel_resource = OTELResourceDetector().Detect();
  auto default_resource     = Resource::GetDefault();
```
and
```
Resource &Resource::GetDefault()
{
  static Resource default_resource({{TELEMETRY_SDK_LANGUAGE, "cpp"},
                                    {TELEMETRY_SDK_NAME, "opentelemetry"},
                                    {TELEMETRY_SDK_VERSION, OPENTELEMETRY_SDK_VERSION}});
  return default_resource;
}
```

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: Ok. Please ignore above comment. I have now made constructor `protected` ( so that it can be used in test cases ), and make `OTELResourceDetector` as its friend class. It seems we don't need to invoke Resource constructor in any other detector, so we should be fine.

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include "opentelemetry/sdk/common/attribute_utils.h"
+#include "opentelemetry/sdk/resource/resource_detector.h"
+#include "opentelemetry/sdk/version/version.h"
+#include "opentelemetry/version.h"
+
+#include <cstdlib>
```
ThomsonTan, line 8: Is this include necessasry?

lalitb: No, we actually need it in `resource.cpp`. Will move it there. Thanks for noticing it.

sdk/include/opentelemetry/sdk/resource/resource.h
jsuereth: This method doesn't need to determine how resources are shared.

A few points:
- Resource-attachment to Trace is NOT specified in the API.  In fact, I don't think you *can* attach via the API in any other language.
- The resource will be attached to a metric provider or trace provider (see: https://github.com/open-telemetry/opentelemetry-java/blob/efd559f1e09e02f6f5a93ac91c272042849b3bd1/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/OpenTelemetrySdkAutoConfiguration.java#L35) for an example in Java.

I think it's reasonable for *this* method to return on the Stack.   We then have our MetricProvider + TraceProvider have a `Resource` attribute they own, and Trace/Metrics generated from those provides can use Pointers or references to the resource associated with a provider.  

Resources should really *only* be a startup concern and not something we expect users to pass around.  I'd rather see a `Resource&& ` argument on `*Provider` classes just to make ownership clear.   This is similar to other APIs.

References:
- [Trace API spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md) has no reference to `Resource`
- [Trace SDK spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md) denotes the resource is *implicitly* associated with a Span.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
jsuereth: I'd like to avoid the counting of shared_ptr references if we can just use the attached TraceProvider.

I.e. shared_ptr has a cost, and I think we don't need it in this scenario.  If we still want to use `shared_ptr` within the SDK to associate resource from `*Provider` to `Span`, that's fine but a different PR.

I think without seeing the other half of the change it's hard to fully comment on merits of shared_ptr.

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Returns a new, merged {@link Resource} by merging the current Resource
+   * with the other Resource. In case of a collision, current Resource takes
+   * precedence.
+   *
+   * @param other the Resource that will be merged with this.
+   * @returns the newly merged Resource.
+   */
+
+  std::shared_ptr<Resource> Merge(const Resource &other) noexcept;
```
maxgolov, line 36: Do we have to implement these (internal to SDK surface) methods.. can we make it a `map` - inherited from standard library map (either `map` or `unordered_map`), and follow semantics of existing APIs available, e.g. [map::merge](https://en.cppreference.com/w/cpp/container/map/merge) .. Then we are not inventing some new Visual Basic-style API, but follow existing Standard Library semantics everywhere. This will be easier to comprehend by experienced modern C++ developer... Plus - I think it would be consistent if our container classes are compatible with existing `algorithm` templates [described here](https://en.cppreference.com/w/cpp/algorithm/merge)

sdk/test/resource/resource_test.cc
```diff
@@ -0,0 +1,138 @@
...
+protected:
+  TestResource resource;
+
+};
+*/
+
+TEST(ResourceTest, create)
+{
+
+  std::map<std::string, std::string> expected_attributes = {
```
maxgolov, line 33: Can you supplement this with a test that handles strongly-typed attributes of numeric (integer) type?

sdk/include/opentelemetry/sdk/resource/resource.h
pyohannes: The spec defines [some strict rules](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md#merge) around merging, which I think are not satisfied by `map::merge`:

> * If the value on the primary resource is an empty string, the result has the value of the secondary resource.
> * Otherwise, the value of the primary resource is used.

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: I agree here. `Resource` need to be associated with TracerProvider level, and we can actually pass the reference of same to traces generated from the provider. And in that case, we don't need to manage smart pointers. Let me add other half in this PR, and it would be easier to comprehend.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
lalitb: Yes, I did revisit again, and probably we mayn't need smart pointer here ( including unique_ptr). Let me add other half of implementation in this PR, so that it would be easy to comprehend and review.

sdk/test/resource/resource_test.cc
lalitb: Yes, sure that would be good test to showcase. I will add it.

sdk/src/resource/resource.cc
lalitb: Ok as per review comments from @jsuereth , Merge() and Create() now returns the stack allocated resource by value, and they resource is owned by `sdk::tracer_provider`, and passed as reference across `tracer` and `span`

lalitb: Ok. based on comments from @jsuereth , `Resources::Merge()` and `Resources::Create()` are modified to return stack allocated `Resource` by value. And this resource is then owned by `sdk::tracer_provider` and passed as reference to span. Please refer to `examples/simple/main.cc` in this PR on how the usage would be, and let me know if any concerns.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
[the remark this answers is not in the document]
lalitb: Based on suggestion from @jsuereth ,  `Resources::Merge()` and `Resources::Create()` are modified to return stack allocated `Resource` by value. And this resource is then owned by `sdk::tracer_provider` and passed as reference to span. Please refer to `examples/simple/main.cc` in this PR on how the usage would be, and let me know if any concerns.

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: OK,  `Resources::Merge()` and `Resources::Create()` are modified to return stack allocated `Resource` by value. And this resource is then owned by `sdk::tracer_provider` and passed as reference to span. Please refer to `examples/simple/main.cc` in this PR on how the usage would be, and let me know if any issues now.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
lalitb: Changes are done as advised ( refer above comments ).

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: This is done now.

sdk/test/resource/resource_test.cc
lalitb: Have modified the test to include both integer and double resource attributes. Please see if this is fine.

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -23,7 +24,9 @@ class Tracer final : public trace_api::Tracer, public std::enable_shared_from_th
    * nullptr.
    */
   explicit Tracer(std::shared_ptr<SpanProcessor> processor,
-                  std::shared_ptr<Sampler> sampler = std::make_shared<AlwaysOnSampler>()) noexcept;
+                  std::shared_ptr<Sampler> sampler = std::make_shared<AlwaysOnSampler>(),
+                  const opentelemetry::sdk::resource::Resource &resource =
+                      opentelemetry::sdk::resource::Resource::Create({})) noexcept;
```
jsuereth, line 29: I think this one may leak a nullptr exception from the default value.

I.e. the default argument won't be retained.   AFAIK it should be ok to *not* give this argument a default value and require the TracerProvider to pass one into the tracer.

sdk/src/trace/span.cc
```diff
@@ -108,6 +109,7 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   recordable_->SetSpanKind(options.kind);
   recordable_->SetStartTime(NowOr(options.start_system_time));
   start_steady_time = NowOr(options.start_steady_time);
+  // recordable_->SetResource(resource_); TODO
```
jsuereth, line 112: As an alternative idea (that might be harder?) we could instead push the Resource into the Exporter via the Processor.  I think that's not really specified (and actually the lack of specification around HOW Resource propagates may give us room to do something different if you think it'd be cleaner).

This approach looks fine, just wanted to mention that you have options AFAICT from the spec.

jsuereth (approved):
This looks really good now!  Just a few minor nits.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,69 @@
...
+  for (auto &elem : other.attributes_)
+  {
+    if ((merged_resource_attributes.find(elem.first) == merged_resource_attributes.end()) ||
+        (nostd::holds_alternative<std::string>(attributes_[elem.first]) &&
+         nostd::get<std::string>(attributes_[elem.first]).size() == 0))
+    {
+      merged_resource_attributes[elem.first] = elem.second;
+    }
+  }
+  return Resource(merged_resource_attributes);
```
pyohannes, line 30: There were some recent spec changes that require this to be revisited: https://github.com/open-telemetry/opentelemetry-specification/pull/1345

sdk/src/trace/span.cc
pyohannes: Another possibility: one could just iterate over the attributes in the resource and add them to the recordable at this point. I think that's the least amount of work that would make this PR functionally complete, and we'd have the ability to elaborate on that later on.

I like the idea about propagating the resource into the exporter (that would make it possible to do some optimization, some vendors have a section of common attributes for batches), on the other hand it would make processor and exporter interfaces and the lives of exporter developers harder.

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -0,0 +1,75 @@
...
+
+  static Resource &GetEmpty();
+
+  /**
+   * Returns a Resource that indentifies the SDK in use.
+   */
+
+  static Resource &GetDefault();
+
+protected:
```
ThomsonTan, line 58: Is `protected` necessary or just private is fine?

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "opentelemetry/sdk/resource/resource_detector.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+
+const std::string TELEMETRY_SDK_LANGUAGE = "telemetry.sdk.language";
```
ThomsonTan, line 12: Name the constants like `kTelemetrySdkLanguage`?

lalitb: Thanks. fixed now :)

sdk/include/opentelemetry/sdk/resource/resource.h
lalitb: `protected` helps creating unit tests case - to subclass `Resource` and create instance of it for testing. I have modified the comment to indicate it's protected :)

sdk/src/trace/span.cc
lalitb: @jsuereth - Good point, this way we don't need to propagate resource every time as part of span, but there would be extra task at exporter side to integrate this resource with span data before exporting.
@pyohannes  - My initial plan was to add `Recordable::SetResource()`. But let me explore more and add propagation in next PR as planned if that's fine ?

examples/batch/CMakeLists.txt
```diff
@@ -4,4 +4,5 @@ add_executable(batch_span_processor_example main.cc)
 
 target_link_libraries(
   batch_span_processor_example ${CMAKE_THREAD_LIBS_INIT} ${CORE_RUNTIME_LIBS}
-  opentelemetry_exporter_ostream_span opentelemetry_trace)
+  opentelemetry_exporter_ostream_span opentelemetry_trace
+  opentelemetry_resources)
```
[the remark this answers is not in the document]
lalitb: Good point, done.

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -57,6 +60,7 @@ class Tracer final : public trace_api::Tracer, public std::enable_shared_from_th
 private:
   opentelemetry::sdk::AtomicSharedPtr<SpanProcessor> processor_;
   const std::shared_ptr<Sampler> sampler_;
+  const opentelemetry::sdk::resource::Resource &resource_;
```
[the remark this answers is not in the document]
lalitb: fixed.

sdk/src/resource/resource.cc
lalitb: Thanks for noticing it. It's kind of reverse of original logic. Have incorporated these changes.

sdk/src/resource/resource.cc
```diff
@@ -0,0 +1,66 @@
...
+Resource Resource::Merge(const Resource &other) noexcept
+{
+  ResourceAttributes merged_resource_attributes(other.attributes_);
+  for (auto &elem : attributes_)
+  {
+    if (merged_resource_attributes.find(elem.first) == merged_resource_attributes.end())
+    {
+      merged_resource_attributes[elem.first] = elem.second;
+    }
+  }
```
pyohannes, line 27: I wonder if that could be even more simplified now:
```suggestion
  ResourceAttributes merged_resource_attributes(other.attributes_);
  merged_resource_attributes.insert(attributes_.begin(), attributes_.end());
```

pyohannes (approved):
Looks good to me.

sdk/include/opentelemetry/sdk/trace/tracer.h
lalitb: Right, default argument for reference can leak. Have removed default now.

PR #508, opened 2021-01-06
Upgrade prometheus to v0.11.0, consume directly the repo in bazel

bogdandrutu:


bogdandrutu (commented):
I did not remove the submodule because it seems that it is used by cmake.

bazel/repository.bzl
```diff
@@ -60,9 +60,13 @@ def opentelemetry_cpp_deps():
...
     # C++ Prometheus Client library.
     maybe(
-        native.local_repository,
+        http_archive,
         name = "com_github_jupp0r_prometheus_cpp",
-        path = "third_party/prometheus-cpp",
+        sha256 = "aab4ef8342319f631969e01b8c41e355704847cbe76131cb1dd5ea1862000bda",
+        strip_prefix = "prometheus-cpp-0.11.0",
+        urls = [
+            "https://github.com/jupp0r/prometheus-cpp/archive/v0.11.0.tar.gz",
```
maxgolov, line 68: Not part of this PR. But we need to eventually create some markdown doc or something that lists the versions of deps used by Bazel and CMake git submodules, ideally ensuring they are identical or nearly-identical.

maxgolov (approved):

bazel/repository.bzl
ThomsonTan: > Not part of this PR. But we need to eventually create some markdown doc or something that lists the versions of deps used by Bazel and CMake git submodules, ideally ensuring they are identical or nearly-identical.

For CMake, we could switch to [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) and download the same archive to make sure the consistency between CMake and Bazel, even if FetchContent is available in CMake 3.11?

ThomsonTan (approved):

bazel/repository.bzl
maxgolov: > For CMake, we could switch to FetchContent

I have a script that builds the latest CMake for old Ubuntu. But that step (building it) may need about 10 mins. We need to see if we can pre-install that in Docker image, cache the image, then we can upgrade to 3.11

PR #516, opened 2021-01-15
Upgrade CMake to 3.15.2 for Linux

ThomsonTan:
The Windows build installs CMake version 3.15.2 (see below link), so upgrade Linux to the same version. This version upgrade for Linux is helpful because the CMake from `apt install` on Ubuntu 18.10 is 3.10, and this version of CMake doesn't support installation from submodule (this affects gRPC installation from source tree, see [here](https://github.com/grpc/grpc/blob/0539a63274cd73288ec13fea16fa855b01c73e61/CMakeLists.txt#L62)).

https://github.com/open-telemetry/opentelemetry-cpp/blob/master/ci/setup_windows_cmake.ps1#L4

### DIFF
--- a/ci/setup_cmake.sh
+++ b/ci/setup_cmake.sh
@@ -5,9 +5,16 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 
+export CMAKE_VERSION=3.15.2
+
+pushd /tmp
+wget https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.sh
+chmod +x cmake-${CMAKE_VERSION}-Linux-x86_64.sh
+./cmake-${CMAKE_VERSION}-Linux-x86_64.sh --prefix=/usr/local --skip-license
+popd
+
 set +e
 echo		\
-                cmake \
                 libbenchmark-dev \
                 libgtest-dev \
                 zlib1g-dev \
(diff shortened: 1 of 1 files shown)


lalitb (approved):

pyohannes (approved):

ci/setup_cmake.sh
```diff
@@ -5,9 +5,17 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 
+export CMAKE_VERSION=3.15.2
+
+pushd /tmp
+wget https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.sh
```
maxgolov, line 11: I have a very similar script here. Main benefit it also validates what version you get, so it doesn't download if you already have a good version: https://github.com/open-telemetry/opentelemetry-cpp/blob/master/tools/setup-cmake.sh

It is cross-platform, but it builds from source. Also it does build it without ninja because on some older platforms that I tested, e.g. Ubuntu 14.x - ninja build of CMake fails. My main experience with building new CMake from source is that this entire process is rather slow. If we download / build it for some platform, we should cache it in github action cache.

maxgolov: Basically this upgrade would add 9 minutes to `CI / CMake test (with otlp-exporter)` run.

ThomsonTan: The current script is mostly for CI and may also help repro CI related issue locally. I think it is unnecessary to rebuild CMake from source in this restricted environment, and do it for every CI build.

Adding a version check sounds like a good improvement we could do in later PR.

maxgolov (approved):

maxgolov:
It's a binary installer. Ok.

PR #517, opened 2021-01-15
Correct package name in the _FOUND var in CMake config

ThomsonTan:
Follow the convention to set CMake `_FOUND` variable with correct package name (`opentelemetry-cpp`) instead of `OPENTELEMETRY_CPP`. The latter name doesn't match the CMake config file name we have and also is not the name passed to `find_package()`. This is not blocking because CMake detects the unexpected `_FOUND` variable and corrects it, but it raises below warning for this.

>  The package name passed to `find_package_handle_standard_args`
  (OPENTELEMETRY_CPP) does not match the name of the calling package
  (opentelemetry-cpp).  This can lead to problems in calling code that
  expects `find_package` result variables (e.g., `_FOUND`) to follow a
  certain pattern.


lalitb (approved):

lalitb (approved):

lalitb:
@ThomsonTan  - changes look good. Once you resolve the conflict. I will merge the changes :)

ThomsonTan:
> @ThomsonTan - changes look good. Can be merged once conflict is resolved :)

Thanks. Resolved the conflict.

Issue #466, opened 2020-12-17
ETW exporter field names


maxgolov:
Opened to follow-up to action item in this PR https://github.com/open-telemetry/opentelemetry-cpp/pull/376

**Problem Statement**

ETW - Event Tracing for Windows - allows to express an event as structured key-value pairs. Field / property names for event - could be anything. The idea is to align ETW field names with OpenTelemetry protocol spec / OTLP spec as close as possible. However, some other non-OTLP collector flows may require different standard schema / field names.

**Proposed Solution**

ETW exporter may handle this as follows:
- extension point (pluggable method, extension, additional class) that allows to remap from OpenTelemetry fields, attributes, resource names to custom protocol
- compile-time header that defines the mapping, that defines - `customer event field X` has to be assigned to `ETW event field Y`, that may be implemented either as a set of C const char* constants, or as `std::map`, that header with definitions can be added to the build system via macro define - to override some default set of fields.
- ability for vendor to implement custom flavor of ETW exporter by inheriting from the default ETW exporter implementation. Different ETW listener `agents` may expect their own unique standard schemas, so some extensibility is nice-to-have. Unfortunately since C++ SDK does not provide runtime loadable extension points for SDK, this would have to manifest as yet another additional exporter, `exporter-schemaA` or `exporter-schemaB` added into vendor-compiled SDK.

Due to limited time given to @mishal23 (his community bridge internship ends end of December 2020) / extent of changes / scope of this "flexible fields naming" proposal, tracking this issue separately from his PR, to implement the feature in a separate PR.

maxgolov:
Implemented this feature as a compile-time customization header with a list of custom field names.

PR #519, opened 2021-01-15
Code improvements for ETW exporter

maxgolov:
# Clean-up of ETW exporter

The following feature may be reusable by other components:
- Owning `Properties` collection (container) with convenient initializer list operators and constructors to pass incoming event properties as `std::string`, vectors of integers/floats, string literals, etc. The way how it should be to be simple and look like JSON object initialization.

## Usage example

```cpp
// Span attributes
  Properties attribs =
  {
    {"attrib1", 1},
    {"attrib2", 2}
  };
  auto outerSpan = tracer->StartSpan("MySpanOuter", attribs);

  // Add first event
  std::string eventName1 = "MyEvent1";
  Properties event1 =
  {
    {"uint32Key", (uint32_t)1234},
    {"uint64Key", (uint64_t)1234567890},
    {"strKey", "someValue"}
  };
  outerSpan->AddEvent(eventName1, event1);

  // Add second event
  std::string eventName2 = "MyEvent2";
  Properties event2 =
  {
    {"uint32Key", (uint32_t)9876},
    {"uint64Key", (uint64_t)987654321},
    {"strKey", "anotherValue"}
  };
  outerSpan->AddEvent(eventName2, event2);

  // Create nested span. Note how we share the attributes here..
  // It is Okay to either share, OR have your own attributes map.
  auto innerSpan = tracer->StartSpan("MySpanInner", attribs);
  std::string eventName3= "MyEvent3";
  Properties event3 =
  {
    {"uint32Key", (uint32_t)9876},
    {"uint64Key", (uint64_t)987654321},
    // {"int32array", {{-1,0,1,2,3}} },
    // Since this collection is owning, we may store values returned by other utility functions
    {"tempString", getTemporaryValue() }
  };
  
  innerSpan->AddEvent(eventName3, event3);
  innerSpan->End();    // end innerSpan
  outerSpan->End();    // end outerSpan
  tracer->CloseWithMicroseconds(0);
```

Note that the new class is fully compatible with `KeyValueIterable`, i.e. provides ABI-safe transform - should there be the need for ABI-safe interface. Note that across Visual Studio 2015-2017-2019, the ABI-safety for STL types is readily guaranteed on Windows. Thus, strictly speaking, unnecessary requirement that can be heavily optimized by avoiding the unnecessary tranformation.

## ETW channel -specific improvements

List of features added:
- preliminary support for Span `attributes` and `links`
- adding `TraceId` + `SpanId` on individual events
- separate expression for Spans being started / stopped; support for nested Spans
- code comments clean-up
- code refactor

## Tests

ETW listener 'Test Framework' is written in C# - will be submited in contrib repo. This code also provides ability to Debug OpenTelemetry ETW exporter events in Visual Studio 2019 using built-in `Diagnostic Events` tab. Code is vendor-agnostic. I included examples of some other vendors using ETW. Those could be the candidates to adopt OpenTelemetry C++ SDK with ETW channel. Field names for ETW channel are also configurable, i.e. vendors may decide their own naming convention. It would make sense to normalize this with OTLP protocol names.

## Debugging Telemetry in Visual Studio 2019

Example screenshot that illustrates how test events show-up in Visual Studio 2019 with no additional tooling necessary:

image

This article explains more about Event Tracing for Windows and possible ways to export the data further:
https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-diagnostics-how-to-monitor-and-diagnose-services-locally

Possible flows will be explained further in contrib repo. Existing tools that allow to monitor and export ETW events to Elastic Search, for example: https://www.fireeye.com/blog/threat-research/2019/03/silketw-because-free-telemetry-is-free.html


maxgolov:
There are a few (small) consistency improvement patches that I had to apply - about 5-7 lines of code unrelated to ETW exporter. Please let me know if it'd make sense to include, or I can split them into separate 7 PRs.

api/include/opentelemetry/trace/span_context_kv_iterable.h
```diff
@@ -24,11 +24,14 @@ class SpanContextKeyValueIterable
    */
   virtual bool ForEachKeyValue(
       nostd::function_ref<bool(SpanContext, const opentelemetry::common::KeyValueIterable &)>
-          callback) const noexcept = 0;
+          callback) const noexcept
+  {
+    return true;
```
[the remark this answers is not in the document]
maxgolov: This method here accepts a reference to attributes:
https://github.com/open-telemetry/opentelemetry-cpp/blob/51eef56efa88da876859030767bf6a4febf4b2b8/api/include/opentelemetry/trace/tracer.h#L52

But if we are to pass a reference to `SpanContextKeyValueIterable` - abstract pure virtual, then the code in Visual Studio will not compile with:
`Error	C2259	'opentelemetry::v0::trace::SpanContextKeyValueIterable': cannot instantiate abstract class`

I added some test code that passes incoming attributes as `SpanContextKeyValueIterable&` here:

```cpp
  // Span attributes
  Properties attribs =
  {
    {"attrib1", 1},
    {"attrib2", 2}
  };

  auto outerSpan = tracer->StartSpan("MySpanOuter", attribs);
  ```
where `Properties` - is a universal shared common owning collection that can be used for any sort of key-value iterable operation in most scenarios... Which itself is derived from `SpanContextKeyValueIterable` / `KeyValueIterable`, and itself is implementing the iterator and size. Thus, does not need another cast. That way the passing of attributes by reference works well, as long as base class is not abstract.

I can't seem to find another viable example where we are currently passing attributes by reference to existing instance. The only other example is initializer list, which may not be viable in some cases.

The other second overload that takes two arguments - right now does the cast to `common::KeyValueIterableView<T>`.

I'm trying to understand why we impose this restriction, i.e. why not just allow passing `SpanContextKeyValueIterable` by reference. Pros and Cons of:
Option 1: No implementation: dev MUST provide implementation (code doesn't compile if they don't).
Option 2: NULL object implementation - no-op by default. dev MUST override in order to make it work.

I'd rather pick the NULL object implementation. Also it avoids the cast. What are we losing if we do it that way?

File: sdk/include/opentelemetry/sdk/common/attribute_utils.h [hunk 2]
```diff
@@ -33,22 +33,44 @@ namespace common
  */
 using OwnedAttributeValue = nostd::variant<bool,
                                            int32_t,
-                                           uint32_t,
                                            int64_t,
+                                           uint32_t,
                                            uint64_t,
                                            double,
                                            std::string,
+// TODO: should we have support for C-strings here or std::string is sufficient?
```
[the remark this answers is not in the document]
maxgolov: In this exact case - for the owning class, I agree that we probably don't want the C-strings.. I can remove this comment.

For the outside containers on API surface, I see some value to have both - `std::string` and `const char *` / string literal initialization. Also for the gRPC / OTLP, - protobuf itself accept both `std::string` and `const char *` as arguments during serialization. It would be reasonable to support both on API surface.

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
```diff
@@ -0,0 +1,476 @@
...
+   * @param source
+   * @return
+   */
+  std::vector<std::string> static to_vector(const nostd::span<const nostd::string_view> &source)
+  {
+    std::vector<std::string> result;
+    result.reserve(source.size());
+    for (const auto &item : source)
+    {
+      result.push_back(std::string(item.data()));
```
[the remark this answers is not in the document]
maxgolov: I sort of agree (partially). Let me digest this.. Somewhere I agree with the `.size()`, but not because of NULs in between. We want all incoming string properties be UTF-8, thus no nulls in the middle of a string.

For the byte arrays - it'd be best to have this spec issue resolved, to allow for byte arrays. I added feature flag for that, so that even before spec issue is approved - it's possible to compile the SDK with byte arrays support:
https://github.com/open-telemetry/opentelemetry-specification/issues/780

Also if we add bytes support now ahead of spec, we can:
- export byte arrays as int for OTLP. This does not break the spec.
- export byte arrays as byte arrays for ETW. It supports `win:Binary` byte array type [described here](https://docs.microsoft.com/en-us/windows/win32/wes/eventmanifestschema-inputtype-complextype)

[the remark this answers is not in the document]
pyohannes: > We want all incoming string properties be UTF-8, thus no nulls in the middle of a string.

Agreed. Then I'll narrow my concern to strings that are not null terminated.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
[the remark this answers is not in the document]
pyohannes: I think from the API point of view it shouldn't matter that much. `const char*` will be automatically converted to `std::string`:
```c++
std::variant<std::string> x;

x = "abc";
x = std::string("def");
```
I don't have too strong an opinion here. Mostly I'm in favor of seeing the types that the variants hold being kept to a mininum, as otherwise it makes it harder to consume a variant (checking all possible types), which is a pain anyway.

maxgolov:
@ThomsonTan @lalitb @jsuereth - could you guys please take a look? While I'm addressing a few comments from Johannes, please let me know your feedback, so that I can quickly address combined feedback in a few commits.

CMakeLists.txt
```diff
@@ -86,9 +88,6 @@ if(WIN32)
   option(WITH_ETW "Whether to include the ETW Exporter in the SDK" ON)
   if(WITH_ETW)
     add_definitions(-DHAVE_MSGPACK)
-    # Option below will be removed once we donate the TraceLoggingDynamic.h to
```
ThomsonTan, line 89: Where is TraceLoggingDynamic.h in OSS?

CMakeLists.txt
```diff
@@ -166,8 +165,8 @@ endif()
 
 list(APPEND CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}")
 
+include(CTest)
```
ThomsonTan, line 168: Why include `CTest` if `BUILD_TESTING` is off?

CMakeSettings.json
```diff
@@ -1,108 +1,126 @@
 ﻿{
-    "configurations": [
+  "configurations": [
```
ThomsonTan, line 2: Should this file be updated manually for any new option added to cmake?

exporters/etw/include/opentelemetry/exporters/etw/.gitignore
```diff
@@ -0,0 +1 @@
+TraceLoggingDynamic.h
```
ThomsonTan, line 1: Why need this?

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
```diff
@@ -0,0 +1,476 @@
...
+#include "opentelemetry/common/key_value_iterable_view.h"
+
+#include <opentelemetry/nostd/span.h>
+#include <map>
+#include <string>
+#include <vector>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace core
+{};
```
ThomsonTan, line 27: What are the empty namespaces for?

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
```diff
@@ -0,0 +1,476 @@
...
+{
+
+  /**
+   * @brief Convert span<T> to vector<T>
+   * @tparam T
+   * @param source
+   * @return
+   */
+  template <typename T>
+  static std::vector<T> to_vector(const nostd::span<const T, std::size_t(-1)> &source)
```
ThomsonTan, line 78: Prefer `std::numeric_limits<std::size_t>::max()` over `-1`?

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
```diff
@@ -0,0 +1,476 @@
...
+
+  /**
+   * @brief Convert span<string_view> to vector<string>
+   * @param source
+   * @return
+   */
+  std::vector<std::string> static to_vector(const nostd::span<const nostd::string_view> &source)
+  {
+    std::vector<std::string> result;
+    result.reserve(source.size());
```
ThomsonTan, line 91: Create the vector like `std::vector<std::string> result(source.size())`, so no need to create one in default size and increase it in the following line?

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
```diff
@@ -0,0 +1,476 @@
...
+  static nostd::span<const T> to_span(const std::vector<T> &vec)
+  {
+    nostd::span<const T> result(vec.data(), vec.size());
+    return result;
+  };
+
+  /**
+   * @brief Convert vector<FLOAT> to span<const FLOAT>.
+   * @tparam T
+   * @param std::vector<T>
```
ThomsonTan, line 115: Should the comments like `@param` be fixed with the actual parameter name?

CMakeLists.txt
maxgolov: We are working to OSS it under Apache License 2.0 this semester. For now this spot requires conditional compilation depending on whether this header exists or not. I will send you the link to private snapshot of it for now. If this header is unavail, then we send ETW events packed with MsgPack payload.

CMakeLists.txt
maxgolov: This is actually how it's recommended in the original `CTest` guidelines. The trick is that if you do that, then `BUILD_TESTING` by default is `ON`, not the other way around:
https://stackoverflow.com/questions/50468620/what-does-enable-testing-do-in-cmake

I looked at a few projects, most of them do about the same. I do not have a strong preference, but I think building tests by default is not a bad idea.

CMakeSettings.json
maxgolov: Yes, unfortunately this file is manually edited. It is only used for the Visual Studio 2019 IDE builds. I would like to add a few predefined configurations. But we need to do a small write-up how customers may edit this to their own liking.

maxgolov:
I'm convering it to `Draft` since I need to resolve a few merge conflicts on latest.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -43,8 +43,7 @@ class Tracer
     return this->StartSpan(name, {}, {}, options);
   }
 
-  template <class T,
-            nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
```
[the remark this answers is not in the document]
maxgolov: It is necessary. But this template condition itself is not working too well for ancestors of `key_value_iterable` , I hit some issue with that... `is_key_value_iterable<T>` implementation needs fixing. I can elaborate more on what happens if I keep it in there.

exporters/etw/include/opentelemetry/exporters/etw/.gitignore
maxgolov: We can't share it in OpenSource yet. This is the place where I temporarily inject it. Once it's OSS, then we can remove it from `.gitignore` here.

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
maxgolov: Yes

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
maxgolov: Yes

maxgolov:
Example decoder for OpenTelemetry C++ SDK ETW/MsgPack using Azure Diagnostics EventFlow:
https://github.com/maxgolov/diagnostics-eventflow/blob/maxgolov/etw_exporter_demo/src/Microsoft.Diagnostics.EventFlow.Inputs.Etw/MsgPackTraceEventSession.cs

My plan is to contribute the "Agent" (ETW Listener) to OpenTelemetry C++ SDK contrib repo. Subsequently, the out-of-proc listener might as well use OpenTelemetry .NET SDK to re-ingest events to whatever destinations presently supported by OpenTelemetry .NET (on Windows, since ETW is a Windows OS-only technology).

CMakeLists.txt
maxgolov: ETA for OSS of TraceLogging: April 2021 (we got prelim approval). I'll refresh the code later when the OSS is complete.

api/include/opentelemetry/trace/tracer.h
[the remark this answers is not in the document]
maxgolov: I'll try to resolve this with a smarter check instead of removing it.

api/include/opentelemetry/trace/span_context_kv_iterable.h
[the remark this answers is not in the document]
maxgolov: I will remove the default implementation from this interface, but I'd still need to add another alternative, e.g.

```cpp
/**
 * @brief Null Span context that does not carry any information.
 */
class NullSpanContext : public SpanContextKeyValueIterable
{
public:
  bool ForEachKeyValue(
      nostd::function_ref<bool(SpanContext, const opentelemetry::common::KeyValueIterable &)>
          callback) const noexcept override
  {
    return false;
  }

  size_t size() const noexcept override { return 0; };
};
```

api/include/opentelemetry/trace/tracer.h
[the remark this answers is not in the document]
maxgolov: Short answer: is that I'll keep the original SFINAE, but it's not working for subclasses of `KeyValueIterable`.

exporters/etw/include/opentelemetry/exporters/etw/.gitignore
maxgolov: I'll resolve this conversation conversation. We will share this header in OSS ETA: April 2021 (almost there)>

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
maxgolov: I honestly don't remember why I needed it here at all, removing in the next iteration :)

exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
maxgolov: I remembered one thing about `max` here : it's by default broken, since `NOMINMAX` macro isn't default.
The old issue: http://www.suodenjoki.dk/us/archive/2010/min-max.htm

The pattern here is identical to implementation of `dynamic extent` in GSL from here:
https://github.com/microsoft/GSL/blob/ef0ffefe525a6219ff245d19a832ce06f3fd3504/include/gsl/span#L64

I think it'd probably be more prudent to add a similar definition of `dynamic_extent` constant to `nostd::span` instead.

maxgolov: Basically Cpp reference suggests:
https://en.cppreference.com/w/cpp/container/span/dynamic_extent

```cpp
inline constexpr std::size_t dynamic_extent = -1;
```
which I think we should be adding to `nostd::` namespace.

maxgolov: Ok, fixed in the next iteration.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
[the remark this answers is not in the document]
maxgolov: `string_view` itself is more of a `const char *` rather than `std::string`.

The idea is to support `const char *` upfront for future-proofing and to remain compatible with C API, if one to be ever built on top of C++ API, without needing a memcpy.

In essence - you do not necessarily need to construct a copy in `std::string` object every time, if you know the buffer & length, and if you can process the original `Event` without copying it. This also helps to easily build a C API on top of C++ API --- because you can still express an array of attributes in C, for example, and a for each key-value iterator on top of that array. For example, using a structure like here:
https://github.com/microsoft/cpp_client_telemetry/blob/3d48e213cb8e4a5693042f06ae7676901d05f560/lib/include/public/mat.h#L167

```cpp
    typedef union
    {
        /* Basic types */
        uint64_t            as_uint64;
        const char*         as_string;
        int64_t             as_int64;
        double              as_double;
        bool                as_bool;
        evt_guid_t*         as_guid;
        uint64_t            as_time;
        /* Array types are nullptr-terminated array of pointers */
        char**              as_arr_string;
        int64_t**           as_arr_int64;
        bool**              as_arr_bool;
        double**            as_arr_double;
        evt_guid_t**        as_arr_guid;
        uint64_t**          as_arr_time;
    } evt_prop_v;
```

Demanding this owned value to be always `std::string` -- implies a `memcpy` , that could be otherwise avoided. I believe this is also the reason why gRPC itself accepts a string to be either `std::string` or `const char *`, and I added the necessary fix for that here before:
https://github.com/open-telemetry/opentelemetry-cpp/blob/4751ae5a29e5eeed1b775b7c337bdbe294165346/exporters/otlp/src/recordable.cc#L65

From spec perspective, both are `strings` (on-wire), but from API perspective - both could be either this or that string, that should be totally within the constraints of the API spec. There is also `std::wstring`  😄 , but I'd rather not go there -- instead I'll offer an option how to transform that to one of the supported types in the updated iteration.

maxgolov:
@ThomsonTan @lalitb @pyohannes - I think I addressed some of your comments, and to two questions - I do not fully agree, I responded with an explanation. Let me know if it is reasonable. If no, I can take another stab at this.

I have not yet added the benchmarks / use-cases for scenario where `Properties` passed to `AddEvent` by reference instead of being constructed from initializer list or by-copy. I can illustrate the usage examples, where it would be _much_ more convenient and more efficient to populate the collection using map-style initializer list, tweaking / decorating its contents initially in customer code, then passing (by reference) to the implementation of exporter. Note in case if batching exporter is used, we could still make this work great with the same type because it is `KeyValueIterable` but then a copy for each-key value is going to be involved. What is good for synchronous ETW exporter may not necessarily be good for other asynchronous batching exporter.

This PR is already too big, and I would like to add benchmarks in a separate PR.

maxgolov:
@pyohannes @lalitb - I added a simple benchmark (for of a unit test-benchmark, not E2E test yet), with the sole purpose of comparing how different containers perform for `span->AddEvent(...)` API call. Here is my observation and a quick explanation of the results, which we discussed in the `ETW README.md` PR:

image

1. `Properties` container performs additional work of potentially re-mapping some non-standard types to OpenTelemetry standard `common::AttributeValue` type. For example, if customers on Windows need to log a `GUID` type, this helper container class - that is itself implementing Key-Value Iterable, may transform incoming attribute to type supported by OpenTelemetry. It may, for example, convert `GUID` to `string` or to `array` (of bytes), when array-of-bytes is approved by the spec.

2. `static Properties` container test shows how a customer may prepopulate a static container with values, passing it by-reference, thus avoiding heap churn. That way the container is passed by-reference, and straight to ETW exporter. In case if other exporter is used (e.g. stdout or OTLP), the container is still iterated using Key-Value iterable approach. Basically, this container logic is optimized for concrete exporter now. `static` is only used to illustrate that the container is kept the same, it could also have been a `class member` instead, same object passed by-reference. This does not change the benchmark outcome

3. initializer list is transformed to Key-Value iterable, and is heavily optimized at compile time.

4. `unordered_map` container showcases how a customer may have all their event attributes populated outside of SDK API call, then pass them (as Key-Value Iterable) to the API. This is similar to a few other test cases.

**Interpretation of Benchmark results**

`>>` means `faster than` :

```console
(static, by-reference, reusable) Properties (2202ns) >>
initializer list (2541ns) >>
Properties (2585ns) >>
unordered_map (2856ns)
```

So, in general, any customer who would like to populate an event in some form of container, then modify that container - decorating it with additional attributes in their code depending on code flow, would prefer `Properties` over plain `unordered_map`. And any customer that has a `fairly static` event contents, that is generally not concurrently modified by different threads at the same time, - would prefer reusing the same object, passing it by reference. This is giving better flexibility/usability/debug'ability (one can set a breakpoint to inspect the container before OpenTelemetry API call), and is also `2541/2202~=15%` faster than just using initializer list, because initializer list forces the API surface to re-create the object every time. This should be much faster if we test this with a container of 50+ properties.

Average Microsoft Windows 10 OS "random event" (as observed by Diagnostic Data Viewer) - contains 50-100 attributes.
Example screenshot for such "interactivity" event:

image

That's where I would anticipate savings of about 30% performance, if we reuse a reference to the same container rather than recreating it every time. There are some applicability boundaries here. But having this option (and documenting when it's best to use it, e.g. best fit for single-threaded flows!) is beneficial. Note: `Properties` container itself does not break the ABI surface, because the class `Properties` itself implements the proper Key-Value Iterable transform. You can feed it to any other TracerProvider Tracer, not just to ETW. And, thus, compatible with exporters other than ETW. I can add an example of a meta-provider that streams events to two destinations at once (e.g. ETW and OTLP socket) in a follow-up PR to illustrate. Maybe to Apache Httpd in contrib, if that works.

Hope that answer satisfies your performance concerns raised in the ETW doc PR.

CMakeLists.txt
ThomsonTan: Ok. Seems line 87 which defines `BUILD_TESTING` option could be deleted then as it is defined in `include(CTest)`.

exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
```diff
@@ -563,28 +560,28 @@ class ETWProvider
...
+      writeResponse = tld::WriteEvent(providerData.providerHandle, eventDescriptor,
+                                      providerData.providerMetaVector.data(), byteVector.data(), 3,
+                                      pDataDescriptors, ActivityId, RelatedActivityId);
+    }
+    else
+    {
+      writeResponse = tld::WriteEvent(providerData.providerHandle, eventDescriptor,
+                                      providerData.providerMetaVector.data(), byteVector.data(), 3,
+                                      pDataDescriptors);
+    };
```
ThomsonTan, line 582:
```suggestion
    }
```

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -0,0 +1,905 @@
...
+  /// Get Owner tracer of this Span
+  /// </summary>
+  /// <returns></returns>
+  trace::Tracer &tracer() const noexcept { return this->owner_; };
+};
+
+/**
+ * @brief ETW TracerProvider
+ */
+class TracerProvider : public trace::TracerProvider
```
ThomsonTan, line 803: Perhaps name the class as `EtwTracerProvider` to make it differentiable from the abstract base class?

ThomsonTan (approved):

lalitb (approved):
LGTM with few minor concerns

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Maybe you are right. Maybe not. I had doubts about what is the best name to use. My logical reasoning was that these live in two different namespaces, implement the same behavior and interface. And practically can be used interchangeably. Either concrete class with impl, or base class that defines the interface for the rest of exporter. One lives `trace::` and another lives in `ETW::` namespace. My main doubt is that `ETW::EtwTracerProvider` is a bit repetitive, `ETW` in the namespace, then `Etw` in the actual class name again. From that perspective - I merely follow the logic of what Abseil or GSL is doing. They have `absl::variant` vs `std::variant`, and `gsl::span` instead of `std::span`. Same class name. Same facet. Same behavior. But different namespace. If you don't mind, I'd consider this change for a follow-up PR. We can advise our customers to use a `using` alias if this changes in a follow-up PR.

api/include/opentelemetry/common/key_value_iterable_view.h
```diff
@@ -30,10 +30,19 @@ struct is_key_value_iterable
...
+ * convert the value type to one listed under AtributeValue (bool, int32_t, int64_t, uint32_t,
+ * uint64_t, double, nostd::string_view, or arrays of primite types). For example, if UUID,
+ * GUID, or UTF-16 string type is passed as one of values stored inside this container, the
+ * container itself may provide a custom implementation of `ForEachKeyValue` to transform the
+ * 'non-standard' type to one of the standard types.
+ */
 template <class T>
 class KeyValueIterableView final : public KeyValueIterable
 {
-  static_assert(detail::is_key_value_iterable<T>::value, "Must be a key-value iterable");
```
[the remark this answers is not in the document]
maxgolov: I removed it, adding an explanation of why exactly. Short answer is that it's a transform into `KeyValueIterable` where `Value = common::AtributeValue`, but the original container may have wider typing system. That then this transform aligns to use only `common::AttributeValue`. This is the original comment on top of this template:

```cpp
/**
 * @brief Container for key-value pairs that can transform every value in it to one of types
 * listed in common::AttributeValue. It may contain value types that are not directly map'able
 * to primitive value types. In that case the `ForEachKeyValue` method acts as a transform to
 * convert the value type to one listed under AtributeValue (bool, int32_t, int64_t, uint32_t,
 * uint64_t, double, nostd::string_view, or arrays of primite types). For example, if UUID,
 * GUID, or UTF-16 string type is passed as one of values stored inside this container, the
 * container itself may provide a custom implementation of `ForEachKeyValue` to transform the
 * 'non-standard' type to one of the standard types.
 */
 ```

File: CMakeLists.txt [hunk 16]
```diff
@@ -6,6 +6,8 @@ cmake_policy(SET CMP0057 NEW)
 
 project(opentelemetry-cpp)
 
+add_definitions(-DHAVE_CSTRING_TYPE)
```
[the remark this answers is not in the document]
maxgolov: We were discussing the value of `CSTRING` with @pyohannes , and I think this is the best explanation I can come up with:
https://github.com/open-telemetry/opentelemetry-cpp/pull/519/files#r602669793

I do not have a very strong opinion on this, BUT if an outside container allows C-string type (for example, if original data structure is passed down either from C or from C# projection, or from any other language projection, it won't have `std::string` storage!!). Then copying it into `std::string` container implies a `memcpy`, that could have been otherwise avoided. I'd, again, refer to gRPC for example - which accepts strings as either `std::string` container or `const char *` pointer, both are strings from gRPC library perspective. I think we should respect similar contract in our library as well, as it gives us some future optimization opportunities where we can assume that possibly a `non-modifyable` for the duration of the `AddEvent` API call C-string has been passed to us. Any synchronous non-batching exporter may benefit from that by large, by avoiding the `memcpy` transform.

**All other exporters actually have been already modified to support it, I committed the relevant changes too.** Instead of removing this concept - we can remove the `#ifdef HAVE_CSTRING_TYPE` everywhere. I am proving by 100% successful CI in this PR here that the entire set of all existing tests, for all exporters, works great with this option defined.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
ThomsonTan: Sure, feel free to revisit this nitpick in future PR.

For the example with `absl::variant` and `std::variant`, I think they are not used in the same place, and there is no inheritance relationship between them, so adding namespace to them is a little redundant. The namespace prefix helps to identify the right class but it is not required as we have `using`.

BTW: very often my ctags is confused to find the right class definition from API or SDK with the same class name.

maxgolov: > BTW: very often my ctags is confused to find the right class definition from API or SDK with the same class name.

^ Yes, with this one I agree completely. It makes a tiny bit harder to search via grep. Smarter indexers in Visual Studio IDE and Eclipse are _usually_ okay. But I agree it's easier to grep for unique name.

maxgolov:
@lalitb - I merged your latest exporter, added the same support of bytes and string to all exporters incl. Zipkin, that are running under CI. All tests pass. If we want to remove it, I'd rather remove it in a separate PR. Or maybe we should just remove the feature gate, if spec committee is fine with our API having additional types on surface (which is a future-proofing for the sake of easier ABI compat with subsequent spec revisions).

maxgolov:
Fixes #466 - I added customization that allows to flexibly rename the fields at compile time. There's some default set of names that matches generic ETW naming convention. But other flows may need different names, so that can be done either at compile time, or  in the remote process, in the ETW listener.

CMakeLists.txt
```diff
@@ -181,21 +180,17 @@ if(BUILD_TESTING)
         ${CMAKE_BINARY_DIR}/lib/libgmock.a)
   elseif(WIN32)
     # Make sure we are always bootsrapped with vcpkg on Windows
-    find_package(GTest CONFIG)
-    if(NOT GTest_FOUND)
+    find_package(GTest)
+    if(NOT (GTEST_FOUND OR GTest_FOUND))
```
maxgolov, line 184: @ThomsonTan - this is where I got into strange state, that after your `CONFIG` change done 3 days ago - Visual Studio 2019 IDE just completely refuse running Google Tests. It does compile into .exe with GTest in it. But it ran ZERO tests. I have to under your change, it works well after, and appears like we still got all tests passing. We can take it offline, or discuss in a separate issue.

api/include/opentelemetry/trace/trace_id.h
```diff
@@ -42,6 +42,9 @@ class TraceId final
     memcpy(rep_, id.data(), kSize);
   }
 
+  // Creates a TraceId with the given ID.
+  explicit TraceId(const uint8_t *data) noexcept { memcpy(rep_, data, kSize); }
```
pyohannes, line 46: I'd prefer to stick with the safe `nostd::span`. It's not that much effort to convert a `const uint8_t *` to a `nostd::span<const uint8_t, kSize>`. Then we're on the safer side.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -0,0 +1,1077 @@
...
+  /**
+   * @brief Tracer destructor.
+   */
+  virtual ~Tracer() { CloseWithMicroseconds(0); };
+};
+
+/**
+ * @brief ETW::Span allows to send event data to ETW listener.
+ */
+class Span : public trace::Span
```
pyohannes, line 727: I still don't fully understand why a `Span` and a `Recordable` is needed, and why not the `Recordable` passes data directly to the `Tracer`.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -0,0 +1,1077 @@
...
+#include "opentelemetry/exporters/etw/etw_provider.h"
+#include "opentelemetry/exporters/etw/utils.h"
+
+namespace core  = opentelemetry::core;
+namespace trace = opentelemetry::trace;
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace ETW
```
pyohannes, line 60: Namespace names should be lower case.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -33,22 +33,44 @@ namespace common
...
                                            std::vector<bool>,
                                            std::vector<int32_t>,
-                                           std::vector<uint32_t>,
                                            std::vector<int64_t>,
+                                           std::vector<uint32_t>,
                                            std::vector<uint64_t>,
                                            std::vector<double>,
                                            std::vector<std::string>>;
 
+enum OwnedAttributeType
```
pyohannes, line 53: Let's follow the Google style guide for enumeration value names, to be consistent across the project.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -0,0 +1,1077 @@
...
+ *
+ * @tparam SpanType     Expected to be ETW::Span
+ * @tparam TracerType   expected to be ETW::Tracer
+ * @param objPtr        Pointer to parent
+ * @param name          Span Name
+ * @param options       Span Options
+ * @return              Span instance
+ */
+template <class SpanType, class TracerType>
+SpanType *new_span(TracerType *objPtr,
```
pyohannes, line 124: Probably this should be in an anonymous namespace? I guess it's not intended to be used anywhere outside this file.

pyohannes (commented):
Would it possible to split out the changes in the `api` folder into a different PR? I'd need some more time to think about those changes and the implications they might have.

CMakeLists.txt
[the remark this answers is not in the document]
maxgolov: Ok, I'm fixing this.

api/include/opentelemetry/trace/trace_id.h
maxgolov: Ok, it does not truly add much extra safety while initializing `TraceId` from random GUID fixed-size buffer of 16 bytes, for example. And adds an extra line of code to wrap the raw buffer into span to pass it. But I got  your point.

CMakeLists.txt
maxgolov: I'll keep my changes for now since these pass all the tests and work well with vs2019 IDE.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
maxgolov: Ok

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Because `Recordable` is in SDK, not in API. And is not needed to implement this exporter. It is not used at all.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: It's a big refactor at this point. But Okay.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Could be potentially used if someone would like to inherit from the `etw::Tracer` class. It is in `exporter::etw` namespace only.

maxgolov:
@pyohannes - I addressed all of your comments. Re. API changes in a separate PR - I minimized the changes and removed the `HAVE_C_STRING` you didn't like, as well as renamed the enums to match Google coding style. There are no API / contract breaking changes in my PR.

api/include/opentelemetry/trace/span_context_kv_iterable.h
```diff
@@ -30,5 +30,22 @@ class SpanContextKeyValueIterable
...
+ * @brief Null Span context that does not carry any information.
+ */
+class NullSpanContext : public SpanContextKeyValueIterable
+{
+public:
+  bool ForEachKeyValue(
+      nostd::function_ref<bool(SpanContext, const opentelemetry::common::KeyValueIterable &)>
+          callback) const noexcept override
+  {
+    return false;
```
pyohannes, line 44: `false` signals an error during iterating over elements. `true` should be returned instead.

api/include/opentelemetry/trace/span_context_kv_iterable.h
```diff
@@ -30,5 +30,22 @@ class SpanContextKeyValueIterable
    */
   virtual size_t size() const noexcept = 0;
 };
+
+/**
+ * @brief Null Span context that does not carry any information.
+ */
+class NullSpanContext : public SpanContextKeyValueIterable
```
pyohannes, line 37: It's a long name, but let's use `NullSpanContextKeyValueIterable`. To be consistent with `SpanContextKeyValueIterable`. `NullSpanContext` is ambiguous.

pyohannes: Anyway, to avoid an additional class, one could also do this:
```c++
using SpanContextKeyValueIterableInitializerList = SpanContextKeyValueIterableView<std::initializer_list<
          std::pair<SpanContext,
                    std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>>>>>;
```

And then define your additional `StartSpan` method like this:
```c++
 nostd::shared_ptr<Span> StartSpan(nostd::string_view name,
                                    const common::KeyValueIterable &attributes,
                                    const StartSpanOptions &options = {}) noexcept
  {
    return this->StartSpan(name, attributes, SpanContextKeyValueIterableInitializerList({}), options);
  }
```

Just to avoid a clutter of types that we also need to document (and that are error-prone, e. g. return `false` in case of success). We also have `NullKeyValueIterable`, however, it's not used anywhere ...

api/include/opentelemetry/common/key_value_iterable_view.h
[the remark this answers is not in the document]
pyohannes: This should be safe. The actual contract is captured in the signature of `ForEachKeyValue`.

pyohannes (commented):
Some additional comments for API related changes.

For lack of time I didn't give a deep review and test to the SDK related changes.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -52,6 +52,13 @@ class Tracer
     return this->StartSpan(name, attributes, {}, options);
   }
 
+  nostd::shared_ptr<Span> StartSpan(nostd::string_view name,
```
pyohannes, line 55: I think you're still missing methods that allows to pass custom non-standard `KeyValueIterable` objects together with links?

pyohannes (approved):
No problematic changes on API side. If not in this PR, we can remove `NullSpanContext` (and maybe even `NullKeyValuIterable`, which is unused) in a separate PR.

api/include/opentelemetry/trace/span_context_kv_iterable.h
maxgolov: I can refactor this in a separate (unrelated PR) by moving these into separate helper header or something. That would also enable `NullTracerProvider` cases, unrelated to ETW per se. Right now it is more or less "internal" detail that I think we can rename / adjust before GA.

api/include/opentelemetry/trace/span_context_kv_iterable.h
maxgolov: Got it. I'll fix it.

jsuereth (approved):
Approving changes to Attribute type enums + adding better HAVE_SPAN_BYTE support.  +1 for diving through all that :)

Are you looking for bazel support for the various configuration of builds or are we ok just using CMake for that for now?

maxgolov:
> Approving changes to Attribute type enums + adding better HAVE_SPAN_BYTE support. +1 for diving through all that :)
> Are you looking for bazel support for the various configuration of builds or are we ok just using CMake for that for now?

I'm good with CMake for now, but I'll take a look at Bazel right after. The way it stands - this part is kinda "self-sufficient" in a way that it's a header-only library, and the current customer who's gonna "preview" it - uses MSBuild / Visual Studio anyways (no CMake, no Bazel). So I'll get back when I get a confirmation from them it works the way they wanted. Then will add the "proper" full range support in terms of our official build systems.

I will add E2E usage sample that covers the path from `OT C++ - to - ETW - to - out-of-proc listener - to - vendor-agnostic flow` , maybe even using OpenTelemetry.NET SDK in a "sink" (listener) to re-ingest to contrib repo.

lalitb:
@maxgolov - While we merge this, we can also have a ticket to consolidate unresolved comments, to ensure they are not lost after merge ?

maxgolov:
Yes, I'll take care of it.

api/include/opentelemetry/trace/tracer.h
maxgolov: Logged an issue here: https://github.com/open-telemetry/opentelemetry-cpp/issues/649

PR #521, opened 2021-01-19
Add missing include directories for ETW and HTTP client for CMake

ThomsonTan:
Add 2 missing include directories and also add the explicit file extension name (.cc) for `add_library()` source file.


exporters/etw/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-include_directories(include)
+include_directories(include
```
maxgolov, line 1: I was thinking we'd better add this at top level since there is another module (zpages) that needs this.

maxgolov: Maybe
`
# Add nlohmann/json submodule to include directories
include_directories(third_party/nlohmann-json/single_include)
`
straight in top-dir `CMakeLists.txt`. That way it'd be available to both:
- ETW exporter
- `ext/include/opentelemetry/ext/zpages/tracez_http_server.h`

I'm doing it in my PR https://github.com/open-telemetry/opentelemetry-cpp/pull/519

lalitb: Agreed, we also need this for zipkin exporter.

maxgolov (approved):
Approved with suggestions. Use your best judgement if you add it at project level or pull it up to the top.

exporters/etw/CMakeLists.txt
ThomsonTan: I am slightly prefer keep it here. Moving it to top level will cause the include directory to be searched for all compilation. But it also looks good to me if some future PR moves this to the top level CMakeLists.txt.

PR #523, opened 2021-01-19
Adding B3 propagator

TomRoSystems:
Basing on https://github.com/open-telemetry/opentelemetry-cpp/blob/880e392e10b6d8e964ade2e22d7231950d6d2443/api/include/opentelemetry/trace/propagation/http_trace_context.h


api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,278 @@
...
+
+    // now convert hex to objects
+    TraceId trace_id_obj       = GenerateTraceIdFromString(trace_id);
+    SpanId span_id_obj         = GenerateSpanIdFromString(span_id);
+    TraceFlags trace_flags_obj = GenerateTraceFlagsFromString(trace_flags);
+    return SpanContext(trace_id_obj, span_id_obj, trace_flags_obj, true);
+  }
+};
+
+// The B3Propagator class provides an interface that enables extracting and injecting
```
lalitb, line 205: Probably header description can be like :
`B3Propogator` class provides interface that enables extracting and injecting context into single header of HTTP Request.
`B3PropogatorMultiHeader` class provides interface that enables extracting and injecting context into multiple headers of HTTP Request ?

lalitb (commented):
Thanks for working on this. Looks good to me in general. Would wait for other's feedback for proper place for it:

`repo_root/api/include/opentelemetry/trace/propagation/` or
`repo_root/propagators`

As per Opentelemetry specs (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#propagators-distribution ) - only `W3C Tracecontext` and `W3C Baggage` MAY be distributed as part of API package, and rest as extension packages.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,278 @@
...
+      return nostd::get<nostd::shared_ptr<Span>>(span).get()->GetContext();
+    }
+    return SpanContext::GetInvalid();
+  }
+
+  static TraceId GenerateTraceIdFromString(nostd::string_view trace_id)
+  {
+    uint8_t buf[kTraceIdHexStrLength / 2];
+    uint8_t *b_ptr = buf;
+    GenerateHexFromString(trace_id, kTraceIdHexStrLength, b_ptr);
```
lalitb, line 81: As B3 trace_id could also be 16 hex chars, and we are always trying to convert 32 hex chars. Should there be any check to avoid that, or else put some comment that it can fail.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,278 @@
...
+#include "opentelemetry/trace/span.h"
+#include "opentelemetry/trace/span_context.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace trace
+{
+namespace propagation
+{
+
+static const nostd::string_view B3CombinedHeader = "b3";
```
lalitb, line 24: nit: we follow google C++ style guide (https://google.github.io/styleguide/cppguide.html#Constant_Names) which recommends `const` to be prefixed with `k`.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,278 @@
...
+  }
+};
+
+// The B3Propagator class provides an interface that enables extracting and injecting
+// context into headers of HTTP requests. HTTP frameworks and clients
+// can integrate with B3Propagator by providing the object containing the
+// headers, and a getter and setter function for the extraction and
+// injection of values, respectively.
+template <typename T>
+class B3Propagator : public B3PropagatorExtractor<T>
```
lalitb, line 211: nit: though no strong preference, just wondering if `B3PropagatorSingleHeader` would be more readable as class name ?

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,278 @@
...
+    context::Context ctx(context);
+    context::ContextValue span = ctx.GetValue(kSpanKey);
+    if (nostd::holds_alternative<nostd::shared_ptr<Span>>(span))
+    {
+      return nostd::get<nostd::shared_ptr<Span>>(span).get()->GetContext();
+    }
+    return SpanContext::GetInvalid();
+  }
+
+  static TraceId GenerateTraceIdFromString(nostd::string_view trace_id)
```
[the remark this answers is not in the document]
bogdandrutu: >Either have trace::SpanId and trace::TraceId constructor take base16 string as argument ( we already have method to return base16 trace::SpanId::ToLowerBase16 ), or move this functionality to some common place ( HTTPTextFormat ? ) as separate PR.

I think this is the right thing to do to avoid duplicate code.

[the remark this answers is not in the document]
TomRoSystems: I agree that having this in constructor would be handy. But what if hex value is malformed? Should then entire object be "silently" set as invalid or throw an exception?
Or should I create corresponding method like "FromLowerBase16" which takes as an argument string_view?
I will create separate PR to remove duplication once this gets merged.

api/include/opentelemetry/trace/propagation/b3_propagator.h
TomRoSystems: Yes! Thanks for pointing this out. I'll create fix and test for such case.

api/include/opentelemetry/trace/propagation/b3_propagator.h
TomRoSystems: Just worried that this might be misleading as at the moment this class inherits from B3PropagatorExtractor which extracts both [single and multiline headers](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#b3-extract) and injects as in specification:  [MUST default to injecting B3 using the single-header format](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#b3-inject) so I considered this class as "main" one as it sticks to default behaviour and hence it's just B3Propagator. Whereas class B3PropagatorSingleHeader differs only in a way how it is extracting headers (with same injecting strategy) and it's not default one so that's why it has longer name. But if there will be more comments for changing name I'm happy to change.

TomRoSystems:
> Thanks for working on this. Looks good to me in general. Would wait for other's feedback for proper place for it:
> 
> `repo_root/api/include/opentelemetry/trace/propagation/` or
> `repo_root/propagators`
> 
> As per Opentelemetry specs (https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#propagators-distribution ) - only `W3C Tracecontext` and `W3C Baggage` MAY be distributed as part of API package, and rest as extension packages.

Thank you for review.

Any more opinions on this one? In go there's [propagation](https://github.com/open-telemetry/opentelemetry-go/tree/master/propagation) directory same in [dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/tree/master/src/OpenTelemetry.Api/Context/Propagation) whereas Java has [extensions directory on top level](https://github.com/open-telemetry/opentelemetry-java/tree/master/extensions/trace-propagators/src/main/java/io/opentelemetry/extension/trace/propagation).

lalitb (approved):

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+        }
+        val = hexDigit1 * 16 + hexDigit2;
+      }
+      buf[posOut] = val;
+    }
+  }
+
+  // Converts a single character to a corresponding integer (e.g. '1' to 1), return -1
+  // if the character is not a valid number in hex.
+  static uint8_t HexToInt(char c)
```
seemk, line 135: Perhaps we should move hex to int conversion to a shared part? Currently the method (+ comment) is duplicated by b3 and http propagators.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+    nostd::string_view trace_flags;
+
+    // first let's try a single-header variant
+    auto singleB3Header = getter(carrier, kB3CombinedHeader);
+    if (!singleB3Header.empty())
+    {
+      // From: https://github.com/openzipkin/b3-propagation/blob/master/RATIONALE.md
+      // trace_id can be 16 or 32 chars
+      auto firstSep = singleB3Header.find('-');
+      trace_id      = singleB3Header.substr(0, firstSep);
```
seemk, line 169: This will allow an invalid length trace id and span id to pass through, e.g. `a-b` will produce `000000000000000a` as the trace ID.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+      int val = 0;
+      if (posInp)
+      {
+        int hexDigit2 = HexToInt(hexStr[--posInp]);  // low nibble
+        int hexDigit1 = 0;
+        if (posInp)
+        {
+          hexDigit1 = HexToInt(hexStr[--posInp]);
+        }
+        if (hexDigit1 < 0 || hexDigit2 < 0)
```
seemk, line 119: This branch is never taken as `HexToInt` returns `uint8_t` (`uint8_t(-1) = int(255)`), thus a malformed hex sequence is generated.

seemk (changes requested):

api/include/opentelemetry/trace/propagation/b3_propagator.h
TomRoSystems: Yep :( copied from here... https://github.com/open-telemetry/opentelemetry-cpp/blob/ce89aa10941c14a6e52e66e68e5c1831c143eaf2/api/include/opentelemetry/trace/propagation/http_trace_context.h#L120-L123

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+    {
+      return (int)(c - 'A' + 10);
+    }
+    else
+    {
+      return -1;
+    }
+  }
+
+  static SpanContext ExtractImpl(Getter getter, const T &carrier)
```
seemk, line 155: Might be a nitpick, but I think we should validate the parsed trace_id, span_id lengths for validity. Say if a malformed header with too large of a trace_id gets sent, currently this logic will produce a valid `SpanContext`, but with some of the trace_id cut off - my opinion is we should just fail early in these cases.

api/include/opentelemetry/trace/propagation/b3_propagator.h
TomRoSystems: Fixed

api/include/opentelemetry/trace/propagation/b3_propagator.h
TomRoSystems: Thanks.
Would like to get more opinion on this topic [Robustness](https://en.wikipedia.org/wiki/Robustness_principle) vs [it's potential harmful consequences](https://tools.ietf.org/html/draft-iab-protocol-maintenance-01) from others as I have not found clear guidance on this one.

lalitb: In case of any error in parsing `trace_id` and `span_id`, the right approach should be to set them invalid ( i.e. all zeros ) and return the SpanContext. This will ensure that SpanContext will be invalid.

api/include/opentelemetry/nostd/string_view.h
```diff
@@ -120,6 +120,20 @@ class string_view
     return substr(pos1, count1).compare(string_view(s, count2));
   };
 
+  size_type find(char ch, size_type pos = 0) const noexcept
```
maxgolov, line 123: Looks good. Minor concern is - since we're adding it here, it would'be been ideal to add all overloads:
```cpp
constexpr size_type find(basic_string_view v, size_type pos = 0) const noexcept;
constexpr size_type find(CharT ch, size_type pos = 0) const noexcept;
constexpr size_type find(const CharT* s, size_type pos, size_type count) const;
constexpr size_type find(const CharT* s, size_type pos = 0) const;
```

TomRoSystems: I'd like to avoid [Scope creep] (https://en.wikipedia.org/wiki/Scope_creep) and complaints related to complexity for those operations.

pyohannes:
Sorry, I'm a bit late with my comment.

> As per [Opentelemetry specs](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#propagators-distribution ) - only W3C Tracecontext and W3C Baggage MAY be distributed as part of API package, and rest as extension packages.

Would it be better to put this under `ext`, instead of `api`? This would be more true to the spec, and help to keep the API part small. [Java](https://github.com/open-telemetry/opentelemetry-java/tree/main/extensions/trace-propagators) has those in an extension directory.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+                             X-B3-TraceId                X-B3-SpanId    X-B3-ParentSpanId (ignored)
+*/
+
+static const int kTraceIdHexStrLength   = 32;
+static const int kSpanIdHexStrLength    = 16;
+static const int kTraceFlagHexStrLength = 1;
+
+// The B3PropagatorExtractor class provides an interface that enables extracting context from
+// headers of HTTP requests. HTTP frameworks and clients can integrate with B3Propagator by
+// providing the object  containing the headers, and a getter function for the extraction.
```
ThomsonTan, line 46: nit: extra space after object.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+    {
+      return nostd::get<nostd::shared_ptr<Span>>(span).get()->GetContext();
+    }
+    return SpanContext::GetInvalid();
+  }
+
+  static TraceId GenerateTraceIdFromString(nostd::string_view trace_id)
+  {
+    uint8_t buf[kTraceIdHexStrLength / 2];
+    uint8_t *b_ptr = buf;
```
ThomsonTan, line 80: `b_ptr` is unused now?

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+    GenerateBuffFromHexStrPad0(trace_id, sizeof(buf), buf);
+    return TraceId(buf);
+  }
+
+  static SpanId GenerateSpanIdFromString(nostd::string_view span_id)
+  {
+    uint8_t buf[kSpanIdHexStrLength / 2];
+    uint8_t *b_ptr = buf;
```
ThomsonTan, line 88: Ditto.

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+    }
+  }
+
+  // Converts a single character to a corresponding integer (e.g. '1' to 1), return -1
+  // if the character is not a valid number in hex.
+  static int8_t HexToInt(char c)
+  {
+    if (c >= '0' && c <= '9')
+    {
+      return (int)(c - '0');
```
ThomsonTan, line 139: Convert return value to `int8_t` to be consistent with the declared return type?

api/include/opentelemetry/trace/propagation/b3_propagator.h
```diff
@@ -0,0 +1,276 @@
...
+                          nostd::string_view trace_description);
+  // Sets the context for a HTTP header carrier with self defined rules.
+  void Inject(Setter setter, T &carrier, const context::Context &context) noexcept override
+  {
+    SpanContext span_context = B3PropagatorExtractor<T>::GetCurrentSpan(context);
+    if (!span_context.IsValid())
+    {
+      return;
+    }
+    char trace_id[32];
```
ThomsonTan, line 224: Declare the length with the constants defined in the begining of this file?

ThomsonTan (approved):

TomRoSystems:
@lalitb @ThomsonTan 
Guys I've created https://github.com/open-telemetry/opentelemetry-cpp/issues/539 and will move rest of work over there.

Can this be merged now?

lalitb:
> @lalitb @ThomsonTan
> Guys I've created #539 and will move rest of work over there.
> 
> Can this be merged now?

Thanks @TomRoSystems . @pyohannes suggested on moving B3 propagator to `ext` directory. If you agree on that, we can either do it as part of #539, or have a separate issue to let someone pick it up.

TomRoSystems:
> > @lalitb @ThomsonTan
> > Guys I've created #539 and will move rest of work over there.
> > Can this be merged now?
> 
> Thanks @TomRoSystems . @pyohannes suggested on moving B3 propagator to `ext` directory. If you agree on that, we can either do it as part of #539, or have a separate issue to let someone pick it up.

I'd like to do it as part of #539

PR #525, opened 2021-01-20
CMake: fix gRPC_CPP_PLUGIN_EXECUTABLE import

seemk:
When building with `-DWITH_OTLP=ON` then `gRPC_CPP_PLUGIN_EXECUTABLE` can't be imported from a current gRPC build, changing it to `TARGET_FILE` fixes the issue. (gRPC examples use the same method: https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/CMakeLists.txt#L123)

### DIFF
--- a/cmake/opentelemetry-proto.cmake
+++ b/cmake/opentelemetry-proto.cmake
@@ -82,8 +82,7 @@ foreach(IMPORT_DIR ${PROTOBUF_IMPORT_DIRS})
   list(APPEND PROTOBUF_INCLUDE_FLAGS "-I${IMPORT_DIR}")
 endforeach()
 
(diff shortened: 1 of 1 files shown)


ThomsonTan (approved):
LGTM. Thanks for your contribution.

Just curious, are you referencing a debug flavor of GRPC build? If so, I overlooked it and assumed always linking to release flavor of GRPC.

seemk:
> LGTM. Thanks for your contribution.
> 
> Just curious, are you referencing a debug flavor of GRPC build? If so, I overlooked it and assumed always linking to release flavor of GRPC.

Afaik I did not specify a build type when building gRPC, so if the default is debug then possibly :thinking:

PR #528, opened 2021-01-21
Fix install location of CMake config script

ThomsonTan:
The search path for CMake `find_package` which works for both Linux and Windows is `lib/cmake/${PROJECT_NAME}`. We are missing `${PROJECT_NAME}` here so `find_package()` cannot find the installation without specify the package root `<PackageRoot>_DIR`. Check below reference link for more details.

https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure


ThomsonTan:
@open-telemetry/cpp-approvers, this is supporting our repo to be installed as CMake package, PTAL.

lalitb (approved):

PR #534, opened 2021-01-26
Add Jaeger exporter

ThomsonTan:
Implementation of Jaeger exporter based on spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk_exporters/jaeger.md.

### DIFF
--- a/exporters/jaeger/CMakeLists.txt
+++ b/exporters/jaeger/CMakeLists.txt
@@ -14,12 +14,33 @@
 
 include_directories(include)
 
-add_library(jae_trace_exporter)
+find_package(thrift REQUIRED)
 
-if(BUILD_TESTING)
-  add_executable(jaeger_exporter_test )
-  target_link_libraries(jaeger_exporter_test ${GTEST_BOTH_LIBRARIES})
-                        ${CMAKE_THREAD_LIBS_INIT} jaeger_trace_exporter)
+set(GENERATED_THRIFT_PATH
+    "${CMAKE_BINARY_DIR}/generated/thrift-gen/cpp")
 
-  gtest_add_tests()
-endif() # BUILD_TESTING
+file(MAKE_DIRECTORY "${GENERATED_THRIFT_PATH}")
+
+set(jaeger_thrift_gencpp_SOURCES
+    ${GENERATED_THRIFT_PATH}/Agent.cpp
+    ${GENERATED_THRIFT_PATH}/jaeger_types.cpp
+    ${GENERATED_THRIFT_PATH}/Collector.cpp
+)
+
+add_custom_command(OUTPUT ${GENERATED_THRIFT_PATH}/Agent.cpp
+    COMMAND ${THRIFT_COMPILER} --gen cpp -out ${GENERATED_THRIFT_PATH} -r ${CMAKE_CURRENT_SOURCE_DIR}/thrift/agent.thrift
+)
+
+add_custom_command(OUTPUT ${GENERATED_THRIFT_PATH}/Collector.cpp ${GENERATED_THRIFT_PATH}/jaeger_types.cpp
+    COMMAND ${THRIFT_COMPILER} --gen cpp -out ${GENERATED_THRIFT_PATH} -r ${CMAKE_CURRENT_SOURCE_DIR}/thrift/jaeger.thrift
+)
+
+add_library(jae_trace_exporter ${jaeger_thrift_gencpp_SOURCES})
+
+# if(BUILD_TESTING)
+#   add_executable(jaeger_exporter_test )
+#   target_link_libraries(jaeger_exporter_test ${GTEST_BOTH_LIBRARIES})
+#                         ${CMAKE_THREAD_LIBS_INIT} jaeger_trace_exporter)
+# 
+#   gtest_add_tests()
+# endif() # BUILD_TESTING
--- a/exporters/jaeger/CMakeLists.txt
+++ b/exporters/jaeger/CMakeLists.txt
@@ -1,0 +1,25 @@
+# Copyright 2020, OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include_directories(include)
+
+add_library(jae_trace_exporter)
+
+if(BUILD_TESTING)
+  add_executable(jaeger_exporter_test )
+  target_link_libraries(jaeger_exporter_test ${GTEST_BOTH_LIBRARIES})
+                        ${CMAKE_THREAD_LIBS_INIT} jaeger_trace_exporter)
+
+  gtest_add_tests()
+endif() # BUILD_TESTING
--- a/exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
+++ b/exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
@@ -14,14 +14,17 @@
 
 #pragma once
 
+#include <opentelemetry/sdk/trace/exporter.h>
+#include <memory>
+
 OPENTELEMETRY_BEGIN_NANESPACE
 namespace exporter
 {
 namespace jaeger
 {
-enum class TransportFormat
+enum class TransportType
 {
-    THRIFT,
+  THRIFT_UDP,
 };
 
 /**
@@ -29,7 +32,10 @@ enum class TransportFormat
  */
 struct JaegerExporterOptions
 {
-
+  // The endpoint to export to.
+  std::string server_addr      = "localhost";
+  uint16_t server_port         = 9090;
+  TransportType transport_type = THRIFT_UDP;
 };
 
 namespace trace_sdk = opentelemetry::sdk::trace;
@@ -54,21 +60,21 @@ public:
    * @param timeout an option timeout, default to max.
    */
   bool Shutdown(
-      std::chrono::microseconds timeout = std::chrono::microseconds::max() noexcept override;
-      )
+      std::chrono::microseconds timeout = std::chrono::microseconds::max() noexcept override;)
   {
-      return true;
+    return true;
   }
 
 private:
-  void InitializeLocalEndpoint();
+  void InitializeEndpoint();
 
 private:
   // The configuration options associated with this exporter.
   bool is_shutdown_ = false;
   JaegerExporterOptions options_;
-
+  std::unique_ptr<ThriftSender> sender_;
 };
+
 }  // namespace jaeger
 }  // namespace exporter
 OPENTELEMETRY_END_NAMESPACE
--- a/exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
+++ b/exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
@@ -1,0 +1,74 @@
+// Copyright 2020, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+OPENTELEMETRY_BEGIN_NANESPACE
+namespace exporter
+{
+namespace jaeger
+{
+enum class TransportFormat
+{
+    THRIFT,
+};
+
+/**
+ * Struct to hold Jaeger exporter options.
+ */
+struct JaegerExporterOptions
+{
+
+};
+
+namespace trace_sdk = opentelemetry::sdk::trace;
+
+class JaegerExporter final : public trace_sdk::SpanExporter
+{
+public:
+  /**
+   * Create a JaegerExporter using all default options.
+   */
+  JaegerExporter();
+
+  /**
+   * Export a batch of spans.
+   * @param spans a span of unique pointers to span recordables.
+   */
+  trace_sdk::ExportResult Export(
+      const nostd::span<std::unique_ptr<trace_sdk::Recordable>> &spans) noexcept override;
+
+  /**
+   * Shutdown the exporter.
+   * @param timeout an option timeout, default to max.
+   */
+  bool Shutdown(
+      std::chrono::microseconds timeout = std::chrono::microseconds::max() noexcept override;
+      )
+  {
+      return true;
+  }
+
+private:
+  void InitializeLocalEndpoint();
+
+private:
+  // The configuration options associated with this exporter.
+  bool is_shutdown_ = false;
+  JaegerExporterOptions options_;
+
+};
+}  // namespace jaeger
+}  // namespace exporter
+OPENTELEMETRY_END_NAMESPACE
--- a/exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
+++ b/exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
@@ -1,0 +1,57 @@
+// Copyright 2020, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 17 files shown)


exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
```diff
@@ -0,0 +1,80 @@
...
+#include <memory>
+
+OPENTELEMETRY_BEGIN_NANESPACE
+namespace exporter
+{
+namespace jaeger
+{
+enum class TransportType
+{
+  THRIFT_UDP,
```
jsuereth, line 27: Should we also allow PROTO? (can be a follow up issue)

exporters/jaeger/include/opentelemetry/exporters/jaeger/udp_transport.h
```diff
@@ -0,0 +1,48 @@
...
+class UDPTransport : public Transport
+{
+  pubilc : UDPTransport(const std::string &addr, uint16_t port);
+  ~UDPTransport();
+
+  void EmitBatch(const thrift::Batch &batch) override;
+
+private:
+  std::unique_ptr<AgentClient> agent_;
+  std::shared_ptr<TTransport> socket_;
```
jsuereth, line 41: Why are these shared_ptr?

There's ref-counting overhead vs. unique_ptr.   Are these actually shared?

exporters/jaeger/src/udp_transport.cc
```diff
@@ -0,0 +1,40 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace jaeger
+{
+
+UDPTransport::UDPTransport(const std::string &addr, uint16_t port)
+{
+  socket_    = std::shared_ptr<TTransport>(new TSocket(addr, port));
+  transport_ = std::shared_ptr<TTransport>(new TBufferedTransport(socket_));
```
jsuereth, line 26: it looks like none of these are really shared, should they be unique to avoid the ref-counting overhead?

jsuereth (changes requested):
Do you need help figuring out a bazel build?  If so, ping me on gitter, happy to help get this out the door!

seemk:
Are there plans to continue with this draft PR? We might need this for nginx and httpd instrumentation, would gladly help.

ThomsonTan:
> Are there plans to continue with this draft PR? We might need this for nginx and httpd instrumentation, would gladly help.

@seemk thanks. I'll clean and push my remaining changes this week. Your help will be appreciated.

exporters/jaeger/src/recordable.cc
```diff
@@ -0,0 +1,183 @@
...
+    AddTag(std::string{key}, nostd::get<double>(value));
+  }
+  else if (nostd::holds_alternative<nostd::string_view>(value))
+  {
+    AddTag(std::string{key}, std::string{nostd::get<nostd::string_view>(value)});
+  }
+  // TODO: extend other AttributeType to the types supported by Jaeger.
+}
+
+void Recordable::SetIds(trace::TraceId trace_id,
```
lalitb, line 47: This method is changed to SetIdentity(). Probably there would be merge conflict.

exporters/jaeger/src/recordable.cc
```diff
@@ -0,0 +1,183 @@
...
+                              const opentelemetry::common::AttributeValue &value) noexcept
+{
+  PopulateAttribute(key, value);
+}
+
+void Recordable::AddEvent(nostd::string_view name,
+                          core::SystemTimestamp timestamp,
+                          const common::KeyValueIterable &attributes) noexcept
+{
+  // TODO: convert event to Jaeger Log
```
lalitb, line 67: We can have tickets for the TODOs if they are not planned in this PR.

exporters/jaeger/src/recordable.cc
```diff
@@ -0,0 +1,183 @@
...
+                         const common::KeyValueIterable &attributes) noexcept
+{
+  // TODO: convert link to SpanRefernece
+}
+
+void Recordable::SetStatus(trace::StatusCode code, nostd::string_view description) noexcept
+{
+  if (code == trace::StatusCode::kUnset)
+  {
+    return;
```
lalitb, line 80: shouldn't return here, as the description should still be needed to set.

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
```diff
@@ -0,0 +1,77 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace jaeger
+{
+
+using namespace jaegertracing;
+
+class Recordable final : public sdk::trace::Recordable
```
lalitb, line 29: Nit: Not something which needs immediate attention, but probably we should have recordable names after transport types - eg `TUDPRecordable`. This is an issue with Zipkin exporter too.

exporters/jaeger/src/jaeger_exporter.cc
```diff
@@ -0,0 +1,79 @@
...
+
+  return sdk::trace::ExportResult::kSuccess;
+}
+
+void JaegerExporter::InitializeEndpoint()
+{
+  if (options_.transport_format == TransportFormat::kThriftUdpCompact)
+  {
+    auto transport = std::unique_ptr<Transport>(
+        static_cast<Transport *>(new UDPTransport(options_.server_addr, options_.server_port)));
```
lalitb, line 67: Does thrift-upd transport have any authentication support? Or it is actually not needed as the agent is running locally, and actual authentication is to be configured between agent and collector?

exporters/jaeger/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+include_directories(include)
+include_directories(thrift-gen)
+
+find_package(Boost REQUIRED)
```
lalitb, line 4: Do we need this package?

lalitb (commented):
Thank you for adding Jaeger support :)

exporters/jaeger/src/recordable.cc
ThomsonTan: Sure, I'll log issues for TODOs when this PR is ready to merge.

exporters/jaeger/src/recordable.cc
ThomsonTan: The conversion doc says the status (Jaeger span status?) should not be set for `UNSET`, let me know if this is still needed.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#status

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
ThomsonTan: Agree. I logged the same comment for ETW exporter. I think we could create a task for the cleanup for all the exporters? I'll try the renaming anyway.

exporters/jaeger/src/jaeger_exporter.cc
ThomsonTan: That's a good question. For local run, no authentication is configured. Let me log a TODO to figure this out.

exporters/jaeger/CMakeLists.txt
ThomsonTan: Thanks for the catch.

exporters/jaeger/src/jaeger_exporter.cc
ThomsonTan: As per below doc, Jaeger agent is supposed to run in trusted environment as sidecar, and for both UDP and HTTP in client agent scenario, authentication like TLS is not supported.

https://www.jaegertracing.io/docs/1.22/security/

exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
```diff
@@ -0,0 +1,96 @@
...
+
+class ThriftSender;
+
+/**
+ * Struct to hold Jaeger exporter options.
+ */
+struct JaegerExporterOptions
+{
+  // The endpoint to export to.
+  std::string server_addr          = "localhost";
```
seemk, line 40: Perhaps future proofing, but ff Thrift over HTTP is supported, perhaps rename it to endpoint and parse host, port from it? With HTTP it will be necessary to specify the full endpoint, e.g. `localhost:1234/api/traces`.

exporters/jaeger/src/thrift_sender.cc
```diff
@@ -0,0 +1,117 @@
...
+    max_span_bytes -= process_bytes_size_;
+  }
+
+  thrift::Span &jaeger_span = *span->Span();
+  jaeger_span.__set_tags(span->Tags());
+
+  const uint32_t span_size = CalcSizeOfSerializedThrift(jaeger_span);
+  if (span_size > max_span_bytes)
+  {
+    // TODO, log too large span error.
```
seemk, line 59: This would be nice to have, but how does otel-cpp log in general? I.e. I don't think it should write to stdout or stderr

exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
```diff
@@ -0,0 +1,96 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace jaeger
+{
+enum class TransportFormat
+{
+  kThriftUdp,
+  kThriftUdpCompact,
+  kThriftHttp,
```
seemk, line 28: Perhaps only the currently supported transports should be listed here

lalitb (approved):
LGTM. Sorry for keeping it waiting for long. I think we can move it to main-stream faster, and then evolve for any changes.

exporters/jaeger/thrift-gen/Collector.cpp
```diff
@@ -0,0 +1,481 @@
+/**
+ * Autogenerated by Thrift Compiler (0.14.0)
```
lalitb, line 2: I am not fully aware of how `.thrift` files should be handled in general for the build. For `.proto` we compile them as part of our build process. Should same be done for thrift files, and then not check-in cpp and headers ?

exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
ThomsonTan: Yes, I suggest we refactor the option in later PR.

exporters/jaeger/src/thrift_sender.cc
ThomsonTan: Yes, let's unify the SDK internal logging once we have it.

exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
ThomsonTan: Yes, that works too. But as this list is expected to be stable, I just leave it there to make the header file stable (later extension should not need to update the header file, just update the `.cc` file). But let me know if removing the unimplemented formats is preferred.

exporters/jaeger/src/TUDPTransport.cc
```diff
@@ -0,0 +1,113 @@
...
+  if (socket_ != THRIFT_INVALID_SOCKET)
+  {
+    ::THRIFT_CLOSESOCKET(socket_);
+  }
+  socket_ = THRIFT_INVALID_SOCKET;
+}
+
+uint32_t TUDPTransport::read(uint8_t *buf, uint32_t len)
+{
+  uint32_t num_read = recvfrom(socket_,
```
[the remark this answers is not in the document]
ThomsonTan: This is a block call but it is not supposed to be called in the export process (only `sendto` is called). It is here because it is required to implement the full thrift Jaeger class.

exporters/jaeger/src/thrift_sender.cc
```diff
@@ -0,0 +1,117 @@
...
+}
+
+int ThriftSender::Flush()
+{
+  std::lock_guard<std::mutex> guard{lock_};
+
+  return FlushWithLock();
+}
+
+int ThriftSender::FlushWithLock()
```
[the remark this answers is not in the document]
ThomsonTan: Yes, `Append()` could be called from a different thread than the one calling `FlushWithLock()`.

exporters/jaeger/thrift-gen/Collector.cpp
ThomsonTan: There are examples for both, check in `.thrift` file (OpenTelemetry-Python) and check in the generated files (OpenTelemetry-DotNet and [Jaeger-client-cpp](https://github.com/jaegertracing/jaeger-client-cpp/tree/master/src/jaegertracing/thrift-gen)). As the Jaeger [`.thrift` files](https://github.com/jaegertracing/jaeger-idl/) are stable and not in active development (this is very different for OTLP), so I'd prefer check in the generated files to avoid the cpp codegen for each build.

exporters/jaeger/src/thrift_sender.cc
```diff
@@ -0,0 +1,117 @@
...
+{}
+
+int ThriftSender::Append(std::unique_ptr<Recordable> &&span) noexcept
+{
+  if (span == nullptr)
+  {
+    return 0;
+  }
+
+  std::lock_guard<std::mutex> guard{lock_};
```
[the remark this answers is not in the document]
ThomsonTan: As the mutex guarded operation is a little heavy (network IO is involved), I am not preferring using `SpingLockMutex` to perform some busy wait here. Let me know your thoughts on switching to `common::SpinLockMutext`.

exporters/jaeger/include/opentelemetry/exporters/jaeger/udp_transport.h
ThomsonTan: Thanks Josh. Yes, I also prefer `unique_ptr` if possible. For the cases of using `shared_ptr`, that would be needed as it is required to be passed to other functions as `shared_ptr`, so I just created it as `shared_ptr` in the first place.

exporters/jaeger/src/thrift_sender.cc
```diff
@@ -0,0 +1,114 @@
...
+    return 0;
+  }
+
+  byte_buffer_size_ += span_size;
+  if (byte_buffer_size_ <= max_span_bytes)
+  {
+    span_buffer_.push_back(jaeger_span);
+    if (byte_buffer_size_ < max_span_bytes)
+    {
+      return 0;
```
lalitb, line 66: Does it mean if there are very few spans created, they may keep waiting for a long time in the queue. Should there be some timeout, or we somehow ensure that every Export() results in the export over the network?

ThomsonTan: Good point. The few events could be there till the tracer object is destructed. We probably could create a separate thread to do flush periodically. But I haven't seen other implementations have the timeout mechanism. How about filing a ticket to follow up?

reyang: @lalitb great question!

@ThomsonTan this is what OpenTelemetry .NET is doing https://github.com/open-telemetry/opentelemetry-dotnet/blob/98055871511d0b56df75cb7443a1ed9a49918014/src/OpenTelemetry.Exporter.Jaeger/JaegerExporter.cs#L84.

This is the semantic we need to achieve https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch.

reyang: > We probably could create a separate thread to do flush periodically.

This doesn't sound correct to me.

ThomsonTan: Thanks for sharing the spec and OpenTelemetry-DotNet link. The spec clearly requires each batch of spans to be serialized and sent to destination. Add a call to `Flush()` at the end of `Export(batch)` for this.

exporters/jaeger/src/thrift_sender.cc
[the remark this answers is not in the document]
ThomsonTan: Based on the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch) (see below excerpt), I removed the lock on the thrift_sender which belong to the exporter instance.

>Export() will never be called concurrently for the same exporter instance. Export() can be called again only after the current call returns.

exporters/jaeger/src/thrift_sender.cc
lalitb: Yes, that seems a good approach.

exporters/jaeger/src/thrift_sender.cc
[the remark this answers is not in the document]
lalitb: nice.

lalitb:
Looks good. @jsuereth has requested a change. We can pull this in once he re-review it.

exporters/jaeger/include/opentelemetry/exporters/jaeger/jaeger_exporter.h
jsuereth: +1

jsuereth (approved):
This looks good for the working first copy.

Agree with most comments to open follow-on bugs.  Do you need help getting bazel working here?   As for installation instructions for bazel, it'll likely be "Just add this dependency in your bazel build".  I'm a bit swamped with stuff right now, but I'm happy to help with the time I have on bazel-related build issue if needed.

lalitb:
@ThomsonTan  - Seems this is good to merge now. As I understand, you and/or @jsuereth  will take care of bazel build as separate PR.

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
ThomsonTan: Created issue https://github.com/open-telemetry/opentelemetry-cpp/issues/735 for this.

PR #535, opened 2021-01-27
Update StatusCode as per specs

lalitb:
update StatusCode as per specs.  Permitted values : `Unset`, `Ok`, `Error` , and NOT the grpc values.

Specs: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md

--clip--

**Set Status**
Sets the Status of the Span. If used, this will override the default Span status, which is Unset.

Status is structurally defined by the following properties:

StatusCode, one of the values listed below.
Optional Description that provides a descriptive message of the Status. Description MUST only be used with the Error StatusCode value.
StatusCode is one of the following values:

_Unset_
The default status.
_Ok_
The operation has been validated by an Application developer or Operator to have completed successfully.
_Error_
The operation contains an error.
The Span interface MUST provide:

An API to set the Status. This SHOULD be called SetStatus. This API takes the StatusCode, and an optional Description, either as individual parameters or as an immutable object encapsulating them, whichever is most appropriate for the language. Description MUST be IGNORED for StatusCode Ok & Unset values.

### DIFF
--- a/api/include/opentelemetry/trace/span.h
+++ b/api/include/opentelemetry/trace/span.h
@@ -34,9 +34,9 @@ enum class SpanKind
 
 enum class StatusCode
 {
-  Unset,  // default status
-  Ok,     // Operation has completed successfully.
-  Error   // The operation contains an error
+  kUnset,  // default status
+  kOk,     // Operation has completed successfully.
+  kError   // The operation contains an error
 };
 
 /**
--- a/api/include/opentelemetry/trace/span.h
+++ b/api/include/opentelemetry/trace/span.h
@@ -29,6 +29,16 @@ enum class SpanKind
   kProducer,
   kConsumer,
 };
+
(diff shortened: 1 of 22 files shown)


ThomsonTan (approved):

jsuereth (approved):

api/include/opentelemetry/trace/span.h
```diff
@@ -29,6 +29,16 @@ enum class SpanKind
...
   kConsumer,
 };
+
+// StatusCode - Represents the canonical set of status codes of a finished Span.
+
+enum class StatusCode
+{
+  Unset,  // default status
+  Ok,     // Operation has completed successfully.
+  Error   // The operation contains an error
```
pyohannes, line 39: For enum values: we should prefix them with a `k`, according to the code style guide.

lalitb: Thank you. Have done the change

PR #538, opened 2021-01-27
Improve invalid span Id check

ThomsonTan:
The current check of `SpanId::IsValid` constructs a default instance of `SpanId` and then make comparison between the 2 instance. Improve this by assuming the 8 bytes `SpanId` as unsigned 64-bit integer and comparing it with 0 directly.

### DIFF
--- a/api/include/opentelemetry/trace/span_id.h
+++ b/api/include/opentelemetry/trace/span_id.h
@@ -58,7 +58,11 @@ public:
   bool operator!=(const SpanId &that) const noexcept { return !(*this == that); }
 
   // Returns false if the SpanId is all zeros.
-  bool IsValid() const noexcept { return *this != SpanId(); }
+  bool IsValid() const noexcept
+  {
+    static_assert(kSize == 8, "update is needed if kSize is not 8");
+    return *reinterpret_cast<const uint64_t *>(&rep_) != 0ull;
+  }
 
   // Copies the opaque SpanId data to dest.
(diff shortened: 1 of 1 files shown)


jsuereth (approved):

api/include/opentelemetry/trace/span_id.h
```diff
@@ -58,7 +58,11 @@ class SpanId final
   bool operator!=(const SpanId &that) const noexcept { return !(*this == that); }
 
   // Returns false if the SpanId is all zeros.
-  bool IsValid() const noexcept { return *this != SpanId(); }
+  bool IsValid() const noexcept
+  {
+    static_assert(kSize == 8, "update is needed if kSize is not 8");
+    return *reinterpret_cast<const uint64_t *>(&rep_) != 0ull;
```
lalitb, line 64: nice. Just wondering whether this would beat `memcmp`. Also we could do something similar for trace_id:

 `return (!(uint64_t *) (&rep))[0] == 0ull ) &&  ((uint64_t *) (&rep))[1] == 0ull ))`

lalitb (approved):

api/include/opentelemetry/trace/span_id.h
ThomsonTan: This should beat `memcmp` if the `memcmp` call is not inlined, as this just involves a memory read and an integer compare (probably just 2 instructions). The other places may also be improved like this. The other thing which is unique here is a default instance is constructed before comparison which is also eliminated in this PR. I'll revisit other `memcmp` in this file.

Issue #537, opened 2021-01-27
TraceState interface is not compliant with specs.


lalitb:
As per Specs: ( https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracestate ):
 - Tracestate is represented as immutable list of string key/value pairs and form
 - All mutating operations MUST return a new TraceState with the modifications applied
 
Current implementation ( https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/trace_state.h )  modifies the existing `TraceState` instance instead of returning new.

-

lalitb:
#551 closes this

PR #551, opened 2021-02-01
TraceState implementation as per spec

lalitb:
Current tracestate api is not compliant to `opentelemetry` and `w3c trace` specs. Attempt to fix in this PR. 
To Summaries the changes

- As per the opentelemetry specs (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracestate), TraceState object is immutable, and `add`, `update` and `delete` methds should return newly created object. Added/updated needed methods.

- Methods to serialize and de-serialize TraceState object to/from w3c trace header ( as defined here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracestate)

- Much accurate validation check for tracestate list members using regex in non-gcc4.8 compiler.

#537

### DIFF
--- a/api/include/opentelemetry/trace/trace_state.h
+++ b/api/include/opentelemetry/trace/trace_state.h
@@ -132,7 +132,7 @@ public:
       else
       {
         // `end` points to end of current list member
-        end = end - 1;
+        end--;
       }
       auto list_member = TrimString(header, begin, end);  // OWS handling
       if (list_member.size() == 0)
@@ -179,11 +179,11 @@ public:
       {
         header_s.append(",");
       }
-      
+
       auto entry = (entries_.get())[count];
-      header_s.append(entry.GetKey());
-      header_s.append(kKeyValueSeparator);
-      header_s.append(entry.GetValue());
+      header_s.append(std::string(entry.GetKey()));
+      header_s.append(std::string(1, kKeyValueSeparator));
+      header_s.append(std::string(entry.GetValue()));
     }
     return header_s;
   }
@@ -220,15 +220,18 @@ public:
   TraceState Set(const nostd::string_view &key, const nostd::string_view &value)
   {
     TraceState ts;
-    if ((!IsValidKey(key) || !IsValidValue(value)) || num_entries_ == kMaxKeyValuePairs)
+    if ((!IsValidKey(key) || !IsValidValue(value)))
     {
       // max size reached or invalid key/value. Returning empty tracestate
       return ts;  // empty instance
     }
 
-    // add new key-value pair at beginning
-    Entry e(key, value);
-    (ts.entries_.get())[ts.num_entries_++] = e;
+    // add new key-value pair at beginning if list is not reached to its limit
+    if (num_entries_ < kMaxKeyValuePairs)
+    {
+      Entry e(key, value);
+      (ts.entries_.get())[ts.num_entries_++] = e;
+    }
     for (size_t i = 0; i < num_entries_; i++)
     {
       auto entry = (entries_.get())[i];
--- a/api/include/opentelemetry/trace/trace_state.h
+++ b/api/include/opentelemetry/trace/trace_state.h
@@ -16,10 +16,14 @@
 
 #include <cstdint>
 #include <cstring>
-#include <regex>
 #include <string>
 
-#include <iostream>
+#include <regex>
+#if (__GNUC__ == 4 && (__GNUC_MINOR__ == 8))
+#  define HAVE_WORKING_REGEX 0
+#else
+#  define HAVE_WORKING_REGEX 1
+#endif
 
 #include "opentelemetry/nostd/span.h"
 #include "opentelemetry/nostd/string_view.h"
@@ -196,7 +200,7 @@ public:
         return std::string(entry.GetValue());
       }
     }
-    return "";
+    return std::string();
   }
 
   /**
@@ -274,8 +278,11 @@ public:
    */
   static bool IsValidKey(nostd::string_view key)
   {
-    // return IsValidKeyRegEx(key); // uncomment me for non-GCC4.8 C++11 compiler
+#if HAVE_WORKING_REGEX
+    return IsValidKeyRegEx(key);
+#else
     return IsValidKeyNonRegEx(key);
+#endif
   }
 
   /** Returns whether value is a valid value. See https://www.w3.org/TR/trace-context/#value
@@ -284,8 +291,11 @@ public:
    */
   static bool IsValidValue(nostd::string_view value)
   {
-    // return IsValidValueRegEx(value); // uncomment me for non-GCC4.8 C++11 compiler
+#if HAVE_WORKING_REGEX
+    return IsValidValueRegEx(value);
+#else
     return IsValidValueNonRegEx(value);
+#endif
   }
 
 private:
--- a/api/include/opentelemetry/trace/trace_state.h
+++ b/api/include/opentelemetry/trace/trace_state.h
@@ -16,6 +16,8 @@
 
 #include <cstdint>
 #include <cstring>
+#include <regex>
+#include<string>
 
 #include "opentelemetry/nostd/span.h"
 #include "opentelemetry/nostd/string_view.h"
@@ -39,6 +41,8 @@ public:
   static constexpr int kKeyMaxSize       = 256;
   static constexpr int kValueMaxSize     = 256;
   static constexpr int kMaxKeyValuePairs = 32;
+  static constexpr auto kKeyValueSeparator =  '=' ; 
+  static constexpr auto kMembersSeparator = ',';
 
   // Class to store key-value pairs.
   class Entry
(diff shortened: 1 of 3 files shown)


api/include/opentelemetry/trace/trace_state.h
```diff
@@ -139,8 +269,85 @@ class TraceState
...
+    {
+      left++;
+    }
+    return str.substr(left, right - left + 1);
+  }
+
+  static bool IsValidKeyRegEx(nostd::string_view key)
+  {
+    std::regex reg_key("^[a-z0-9][a-z0-9*_\\-/]{0,255}$");
+    std::regex reg_key_multitenant(
```
pyohannes, line 327: Those can be `static`.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -139,8 +269,85 @@ class TraceState
...
+  {
+    // Hex 0x20 to 0x2B, 0x2D to 0x3C, 0x3E to 0x7E
+    std::regex reg_value(
+        "^[\\x20-\\x2B\\x2D-\\x3C\\x3E-\\x7E]{0,255}[\\x21-\\x2B\\x2D-\\x3C\\x3E-\\x7E]$");
+    // Need to benchmark without regex, as a string object is created here.
+    if (std::regex_match(std::string(value), reg_value))
+    {
+      return true;
+    }
+    return false;
```
pyohannes, line 347:
```suggestion
    return std::regex_match(std::string(value), reg_value);
```
But that's more a matter of taste.

pyohannes (commented):
Thanks, that looks very good! I'll give it another detailed review tomorrow.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,157 @@ class TraceState
...
+      }
+      header_s = header_s + kv;
+    }
+    return header_s;
+  }
 
   // Returns false if no such key, otherwise returns true and populates the value parameter with the
   // associated value.
-  bool Get(nostd::string_view key, nostd::string_view &value) const noexcept
+  std::string Get(nostd::string_view key) const noexcept
```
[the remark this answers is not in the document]
lalitb: As per w3c specs ( https://www.w3.org/TR/trace-context/#value) the value can't be empty string ( should have at-least one non-blank character ), and this is validated before storing from header. So returning empty string would mean `key` doesn't exist. But I don't have strong opinion on this, we can switch back to earlier prototype if that's more clear.

api/include/opentelemetry/trace/trace_state.h
lalitb: This is definitely better :)

api/include/opentelemetry/trace/trace_state.h
lalitb: Good point. Fixed this now.

api/include/opentelemetry/trace/trace_state.h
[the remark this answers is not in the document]
lalitb: Have made comments consistent with method prototype.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,163 @@ class TraceState
     }
   };
 
-  // An empty TraceState.
-  TraceState() noexcept : entries_(new Entry[kMaxKeyValuePairs]), num_entries_(0) {}
+  /**
+   * Returns a newly created Tracestate parsed from the header provided.
```
ThomsonTan, line 111: nit: change spelling of `Tracestate` in this comment to `TraceState`?

api/include/opentelemetry/trace/trace_state.h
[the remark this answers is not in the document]
pyohannes: You're right. Your method signature is fine.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,163 @@ class TraceState
...
+   * @param header Encoding of the tracestate header defined by
+   * the W3C Trace Context specification https://www.w3.org/TR/trace-context/
+   *  @return Tracestate A new Tracestate instance or DEFAULT
+   */
+  static TraceState FromHeader(const nostd::string_view &header)
```
pyohannes, line 116:
```suggestion
  static TraceState FromHeader(nostd::string_view header)
```
Just a nit, but to be consistent with the `Get` method below. `string_view` is const anyway, and it's also small enough to be passed by value.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,163 @@ class TraceState
...
     {
+      auto entry = (entries_.get())[num_entries_ - count];
+      auto kv    = std::string(entry.GetKey()) + kKeyValueSeparator + std::string(entry.GetValue());
+      if (--count)
+      {
+        // append "," if not last member
+        kv += ",";
+      }
+      header_s = header_s + kv;
+    }
```
pyohannes, line 187:
```suggestion
    for (size_t count = 0; count < num_entries_; count++)
    {
      if (count != 0)
      {
        header_s.append(",");
      }
      
      auto entry = (entries_.get())[count];
      header_s.append(entry.GetKey());
      header_s.append(kKeyValueSeparator);
      header_s.append(entry.GetValue());
    }
```
Just a nit, maybe a bit clearer.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -139,8 +275,81 @@ class TraceState
...
+    {
+      return true;
+    }
+    return false;
+  }
+
+  static bool IsValidValueRegEx(nostd::string_view value)
+  {
+    // Hex 0x20 to 0x2B, 0x2D to 0x3C, 0x3E to 0x7E
+    std::regex reg_value(
```
pyohannes, line 346:
```suggestion
    static std::regex reg_value(
```

api/test/trace/trace_state_test.cc
```diff
@@ -60,81 +61,129 @@ TEST(EntryTest, SetValue)
...
+  return header;
+}
+
+TEST(TraceStateTest, ValidateHeaderParsing)
+{
+  std::string trace_state_header = "k1=v1";  // valid header
+  EXPECT_EQ(create_ts_return_header(trace_state_header), trace_state_header);
+
+  trace_state_header = "K1=V1";  // invalid header, key can't start with uppercase.
+  EXPECT_EQ(create_ts_return_header(trace_state_header), "");
```
pyohannes, line 94:
```suggestion
  struct {
    const char* input;
    const char* expected;
  } testcases[] = {
    { "k1=v1", "k1=v1" },
    { "K1=V1", "" },
    { "k1=v1,k2=v2,k3=v3", "k1=v1,k2=v2,k3=v3" },
    { "k1=v1,k2=v2,,", "k1=v1,k2=v24" }
    // not all cases covered
  };

  for (auto& testcase : testcases)
  {
    EXPECT_EQ(TraceState::FromHeader(testcase.input).ToHeader(), testcase.expected);
  }
```
Also just a nit. I think this makes it easier to read and to add tests.

api/test/trace/trace_state_test.cc
```diff
@@ -60,81 +61,129 @@ TEST(EntryTest, SetValue)
...
+{
+  std::string trace_state_header = "k1=v1,k2=v2";
+  auto ts1                       = TraceState::FromHeader(trace_state_header);
+  auto ts1_new                   = ts1.Set("k3", "v3");
+  EXPECT_EQ(ts1_new.ToHeader(), "k3=v3,k1=v1,k2=v2");
+
+  trace_state_header = header_with_max_members();
+  auto ts2           = TraceState::FromHeader(trace_state_header);
+  auto ts2_new       = ts2.Set("n_k1", "n_v1");  // adding to max list, should fail and return empty
+  EXPECT_EQ(ts2_new.ToHeader(), "");
```
pyohannes, line 141: What consequence would this behavior have for the user?

Let's say one receives a `tracestate` header with 32 entries, then adds one (this returns an empty `tracestate`), and then passes on the header. If the user doesn't do explicit checks, an empty header will be passed on. I don't think that's ideal.

What do you think about `Set` just returning the unaltered `tracestate` when the limit is exceeded?

api/include/opentelemetry/trace/trace_state.h
lalitb: Yes, this is much cleaner.

api/test/trace/trace_state_test.cc
lalitb: Thanks.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,163 @@ class TraceState
...
+      end = header.find(kMembersSeparator, begin);
+      if (end == std::string::npos)
+      {
+        // last list member. `end` points to end of it.
+        end = header.size() - 1;
+      }
+      else
+      {
+        // `end` points to end of current list member
+        end = end - 1;
```
ThomsonTan, line 135: Just `--end`? Also could it get `-1` if the first character is `,`?

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,163 @@ class TraceState
...
+      Entry e(key, value);
+      (ts.entries_.get())[ts.num_entries_++] = e;
+    }
+    return ts;
+  }
+
+  /**
+   * Returns `new` TransState object after removing the attribute with given key ( if present )
+   * @returns empty TransState object if key is invalid
+   * @returns copy of original TransState object if key is not present (??)
```
ThomsonTan, line 246:
```suggestion
   * @returns copy of original TraceState object if key is not present (??)
```

Seems for now an empty instance is returned if the key is not present, should we return original `TraceState` like mentioned in the comment?

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -139,8 +275,81 @@ class TraceState
...
+
+  // Maintain the number of entries in entries_. Must be in the range [0, kMaxKeyValuePairs].
+  size_t num_entries_;
+
+  // An empty TraceState.
+  TraceState() noexcept : entries_(new Entry[kMaxKeyValuePairs]), num_entries_(0) {}
+
+  static nostd::string_view TrimString(nostd::string_view str, size_t left, size_t right)
+  {
+    while (str[(std::size_t)right] == ' ' && left < right)
```
ThomsonTan, line 319: nit: C++ style `static_cast<std::size_t>`?

api/test/trace/trace_state_test.cc
lalitb: Yes, I didn't find anything in specs on how to address this, and different languages implementation are behaving differently ( python returns empty, JS doesn't do any check on limit, java returns reference to itself ). Said that, it make sense to return unaltered `tracestate` instead of empty. I will do the change. Thanks for the comment.

api/include/opentelemetry/trace/trace_state.h
lalitb: Thanks. changed it now :)

api/test/trace/trace_state_test.cc
lalitb: This is fixed as discussed.

api/include/opentelemetry/trace/trace_state.h
lalitb: Made it `--end`. If would be fine if first char is `,`, as this will make control to go to below condition:
```
      if (list_member.size() == 0)
      {
        // empty list member, move to next in list
        begin = end + 2;  // begin points to start of next member
        continue;
      }
```
and process next list member. I have added test case to validate this.

api/include/opentelemetry/trace/trace_state.h
lalitb: Thanks for suggestion. Have changed it now as part of other changes.
Also, 
`empty instance is returned if the key is not present,`
 - As of now, as per code it return empty instance if key is not valid, and copy of original instance if key is not present.

api/include/opentelemetry/trace/trace_state.h
lalitb: Thanks. fixed it now.

api/include/opentelemetry/trace/trace_state.h
ThomsonTan: If the first char is `,`, then `end` will be `-1`, so the call to `TrimString` will trigger access violation when accessing the input string, like `header[static_cast<size_t>(-1)]` which happens before the empty `list_member` check?

lalitb: Yes, you are right. I am fixing and bazel asan fails because of that. I am fixing it now :) Thanks for pointing out the issue.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -97,37 +107,168 @@ class TraceState
...
+    for (size_t count = 0; count < num_entries_; count++)
     {
+      if (count != 0)
+      {
+        header_s.append(",");
+      }
+
+      auto entry = (entries_.get())[count];
+      header_s.append(std::string(entry.GetKey()));
+      header_s.append(std::string(1, kKeyValueSeparator));
```
ThomsonTan, line 185: No need to construct a temporary `std::string` here, just pass `kKeyValueSeparator` to `std::string::append`?

api/include/opentelemetry/trace/trace_state.h
lalitb: Ok, the easy solution was to add special conditional check for `,` as first char.

api/include/opentelemetry/trace/trace_state.h
lalitb: thanks fixed.

pyohannes (approved):

lalitb:
@ThomsonTan  - Thanks for your review. Can we merge this PR if there are no further comments. I have a another PR dependent on these changes so the ask :)

ThomsonTan (approved):

Issue #550, opened 2021-02-01
Update SpanProcessor.onStart to accept parent Context


lalitb:
as per spec https://github.com/open-telemetry/opentelemetry-specification/blob/7e0c10260178b947777cf9c11d7f7ac4d4700725/specification/trace/sdk.md#onstart
SpanProcessor.on_start should have a parent Context as parameter

PR #552, opened 2021-02-01
Passing parent context to SpanProcessor::OnStart()

lalitb:
as per spec https://github.com/open-telemetry/opentelemetry-specification/blob/7e0c10260178b947777cf9c11d7f7ac4d4700725/specification/trace/sdk.md#onstart
`SpanProcessor.onStart()` should have a parent Context as parameter.

fixes: #550

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
+++ b/sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
@@ -52,8 +52,11 @@ public:
    * NOTE: This method is a no-op.
    *
    * @param span - The span that just started
+   * @param parent_context - The parent context of the span that just started
    */
-  void OnStart(Recordable &span) noexcept override;
+  void OnStart(Recordable &span,
+               const opentelemetry::trace::SpanContext &parent_context =
+                   opentelemetry::trace::SpanContext::GetInvalid()) noexcept override;
 
   /**
    * Called when a span ends.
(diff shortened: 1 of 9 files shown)


sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
```diff
@@ -66,8 +66,11 @@ class BatchSpanProcessor : public SpanProcessor
    * NOTE: This method is a no-op.
    *
    * @param span - The span that just started
+   * @param parent_context - The parent context of the span that just started
    */
-  void OnStart(Recordable &span) noexcept override;
+  void OnStart(Recordable &span,
+               const opentelemetry::trace::SpanContext &parent_context =
+                   opentelemetry::trace::SpanContext::GetInvalid()) noexcept override;
```
pyohannes, line 73: Any reason for having a default value here? According to the spec this parameter is required, so I think we shouldn't make it easy to omit it.

pyohannes (commented):
Thanks for getting us more closer to spec conformance.

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
lalitb: Yes right, it should be required parameter. I have modified the function prototype accordingly.

jsuereth (approved):

pyohannes (approved):

Issue #560, opened 2021-02-05
Adding TraceState to SpanContext


ThomsonTan:
According to below trace spec, `TraceState` is required to be part of `SpanContext` which is currently missing in the [definition](https://github.com/open-telemetry/opentelemetry-cpp/blob/a1dc8bd22cb255caf4f9e7591c29d2347a3e3384/api/include/opentelemetry/trace/span_context.h#L33).

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracestate

PR #561, opened 2021-02-05
Add TraceState to SpanContext

ThomsonTan:
Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/560. `TraceState` is required part of `SpanContext` as per below doc.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracestate


api/include/opentelemetry/trace/trace_state.h
```diff
@@ -107,11 +107,35 @@ class TraceState
...
+  TraceState(const TraceState &other) noexcept
+      : entries_(new Entry[kMaxKeyValuePairs]), num_entries_(other.num_entries_)
+  {
+    for (size_t i = 0; i < num_entries_; i++)
+    {
+      entries_.get()[i] = other.entries_.get()[i];
+    }
+  }
+
+  TraceState operator=(const TraceState &other) noexcept
```
lalitb, line 122: `TraceState` is supposed to be immutable, so assignment operator should be deleted by default. We may need to explicitly add the assignment operator for `SpanContext` which does assignment for rest of it's members ( and not TraceState). As per specs:
``Please note, since SpanContext is immutable, it is not possible to update SpanContext with a new TraceState. ``

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -107,11 +107,35 @@ class TraceState
     }
   };
 
+  // Creates an empty TraceState.
+  TraceState() noexcept : entries_(new Entry[kMaxKeyValuePairs]), num_entries_(0) {}
```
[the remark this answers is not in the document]
ThomsonTan: Make the empty instance a static default is great suggestion, updated accordingly, thanks.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -107,11 +107,38 @@ class TraceState
     }
   };
 
+  static TraceState &GetDefault()
+  {
+    static TraceState ts;
+    return ts;
```
pyohannes, line 113: I think this doesn't give us too much benefit here, considering that the `TraceState` object is copied on assignment anyway. This pattern might make more sense when using a shared pointer:
```c++
static nostd::shared_ptr<TraceState> ts(new TraceState());
return ts;
```

api/include/opentelemetry/trace/span_context.h
```diff
@@ -81,6 +94,7 @@ class SpanContext final
   trace_api::TraceId trace_id_;
   trace_api::SpanId span_id_;
   trace_api::TraceFlags trace_flags_;
+  trace_api::TraceState trace_state_;
```
pyohannes, line 97: I see some benefits of storing this by value. But I'm leaning a bit more toward a `nostd::shared_ptr<TraceState>` approach. I think we'll mostly have two scenarios:
1. No `TraceState` is used at all. In that instance it would be cheaper to have a `shared_ptr` that can be a `nullptr`.
2. The same `TraceState` is used for all spans in a trace. in that case, a `shared_ptr` could be shared for all spans.

I'm still not sure what's better. I'm just a bit concerned about adding ~ 500 bytes to every span (context).

pyohannes, line 97: I'm wondering if we should also consider this new field in `operator==`? Not sure about that either, I'll have a look what other implementations do.

api/include/opentelemetry/trace/trace_state.h
lalitb: Good point. Passing by reference, or shared ptr would have benefit in this case. And if we are planning to store trace_state as `shared_ptr`  in span context, returning as shared_ptr would be better option.

ThomsonTan: As the return `TraceState` is mutable, I'd be hesitate to make it as static. Changed it to create a new empty instance and renamed it as `GetEmpty()`.

api/include/opentelemetry/trace/span_context.h
ThomsonTan: Good point. Seems the `operator==` on `SpanContext` is not used anywhere yet, so I just removed it.

ThomsonTan: Share the state for all the Spans in the same traces is a good point. Changed the type to `nostd::shared_ptr`.

ThomsonTan: Added back `operator==` as it is used in the span_context test. Not sure it is necessary here. Added `TODO` in the `operator==` and will take detail look into it.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -107,11 +107,35 @@ class TraceState
...
+    for (size_t i = 0; i < other.num_entries_; i++)
+    {
+      entries_.get()[i] = other.entries_.get()[i];
+    }
+
+    this->num_entries_ = other.num_entries_;
+
+    return *this;
+  }
+
```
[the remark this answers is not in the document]
lalitb: If we are storing TraceState as shared_ptr within SpanContext, I think it should be good to remove both assignment and copy constructor here. I don't see we using them anywhere, or any use-case in future.

api/include/opentelemetry/trace/trace_state.h
lalitb: Do we have scenario where `TraceState` once created would be modified? If not, it would be good to keep it static. This way, we can avoid creating empty instance with every span creation.

pyohannes: I'm afraid I opened a can of worms here. As far as I understand, any `TraceState` object is read-only, any modifications returns a completely new object.

However, I think to make this interoperable with the changes here we would need those modifiers (`Get` and `Delete`) to return a `shared_ptr` instead of a `TraceState` instance on the stack: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/trace_state.h#L228

api/include/opentelemetry/trace/trace_state.h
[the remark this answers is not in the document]
ThomsonTan: Thanks for catching. Removed both assignment and copy ctor.

api/include/opentelemetry/trace/trace_state.h
lalitb: Agree, both `TraceState::Get` and `TraceState::Delete` should return shared_ptr objects.

ThomsonTan: Ok, updated to return `shared_ptr` from `TraceState::Set`, `TraceState::Delete` and `TraceState::FromHeader`.

lalitb (approved):

api/include/opentelemetry/trace/span_context.h
```diff
@@ -44,33 +43,51 @@ class SpanContext final
...
-  SpanContext &operator=(const SpanContext &ctx) = default;
+  // @returns the trace_state associated with this span_context
+  const nostd::shared_ptr<trace_api::TraceState> trace_state() const noexcept
+  {
+    return trace_state_;
+  }
 
   bool operator==(const SpanContext &that) const noexcept
   {
+    // TODO: check trace_state?
```
pyohannes, line 84: Can we remove this TODO, but rather add a doc comment to the operator method that describes the current behavior (that trace state is not taken into account when comparing)? The current behavior makes sense in my opinion.

pyohannes (approved):

PR #562, opened 2021-02-05
Add API documentation

pyohannes:
This adds Doxygen API documentation for the API part. I only includes the tracing part and leaves out metrics and logs. It also leaves out everything that's not necessary for an API user to know about. The goal is to provide a minimal amount of high quality documentation (currently it's only minimal).

This relies on the default Doxygen stylesheet with only minimal modifications.

You can see the generated documentation here: https://pyohannes.github.io/opentelemetry-cpp-apidoc-draft/

Items that will be addressed in other PRs:
* Actually improving documentation and adding it where it's missing.
* Working out a way to distributed and publish documentation.
* Hook this into the version update script.
* Don't use the existing README, but use a "Getting started" guide instead.


api/docs/Doxyfile
```diff
@@ -0,0 +1,2461 @@
...
+# Configuration options related to the input files
+#---------------------------------------------------------------------------
+
+# The INPUT tag is used to specify the files and/or directories that contain
+# documented source files. You may enter file names like myfile.cpp or
+# directories like /usr/src/myproject. Separate the files or directories with
+# spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
+# Note: If this tag is empty the current directory is searched.
+
+INPUT                  = ../../README.md \
```
ThomsonTan, line 793: Should all files under `../include` be included in the document?

lalitb, line 793: nit: There are some relative links in README.md which won't be accessible from docs. Probably they need to be made absolute :
```
Installation
Please refer to ./INSTALL.md "INSTALL.md".
``

lalitb:
Just curious, how will the docs updated with each PR merge. Should there be a github action to trigger doxygen with each merge, or github magically takes care of it ?

api/docs/Doxyfile
pyohannes: The README for now is just a placeholder. I intend to replace it with a "Getting started" guide. So I think it's not worth the effort sanitizing the README.

pyohannes: I decided against it, because I think there are parts of the API that we don't want to present to the user, for example all the things in `context`, or the `spin_lock_mutex.h`. Those are things we need internally, but we don't really want user (who are instrumenting their libraries) to use this.

Also, I think we should keep metrics and logs out of the documentation for now.

Those were just my thoughts though, I'm open to other opinions.

pyohannes:
>  Just curious, how will the docs updated with each PR merge.

I suggest we only update the documentation for releases, not for each PR merge. This could be a part of the release process then. I think though we can only flesh this out further once we know where documentation will be published.

lalitb:
LGTM.

lalitb (approved):

ThomsonTan (approved):

Issue #68, opened 2020-04-27
Investigate Boost license implications


maxgolov:
A portion of nostd::variant currently requires a Boost-licensed header.
This may be an issue for some commercial products.

Two alternatives to consider:

- research if we can get that portion code contributed under Open Telemetry (Apache-style) license. That way we still satisfy the ABI stability guarantees.

- the other alternative is to compile with std #67 . That approach means that commercial products building from source with standard library cannot safely interop with other prebuilt Open Telemetry modules.

maxgolov:
I'll try to remove MPark variant and replace by Abseil Variant `absl::variant`. There is no other code in this repo that requires Boost License.

maxgolov:
I would propose we totally remove MPark variant - replacing it with a copy of implementation borrowed from `absl::variant`. I need this anyways to start supporting Visual Studio 2015 #314 ... Thus, removing MPark variant we will get rid of the Boost licensing issue.

maxgolov:
Sorry for inactivity on it -- I'll switch to `absl::variant` #572 thus removing the need for Boost License. ETA: April 2021

maxgolov:
Closing as Duplicate of #572 - switching to Abseil Variant would allow us to remove Boost -Licensed code.

PR #565, opened 2021-02-05
Patches to enable Visual Studio 2015 - Update 2 support

### LINKED ISSUE #314: context::ContextValue cannot be compiled with Visual Studio 2015

maxgolov:
Fixes #314 . It does not yet add the build system for supporting vs2015 builds in CI, as it is quite involved (there is no GitHub Actions runner with preinstalled vs2015 and installing one may need some tweaking, and IS SLOW.. unless we cache some docker container image for that). Best we can do is deploy additional vs2015 community build tools. I'll cover CI for vs2015 in a separate PR. I verified the code locally with ETW exporter on Windows with Visual Studio 2015 - Update 2. It compiles and works well.

# Changes

- MPark variant that we use in the SDK is NOT compiling with Visual Studio 2015. Abseil Variant is compiling and working well with Visual Studio 2015 Update 2+. This visual studio version requirement is coming from Abseil itself, but it's a reasonable requirement. The solution is to add a build option `HAVE_ABSEIL_VARIANT` to compile with that one instead of MPark variant.

- Visual Studio 2015 secure development practices does not like `std::copy` and breaks the build. The fix is to replace the `std::copy` with bounds-checked implementation that uses `for-loop`

- Visual Studio 2015 does not like passing context by value. There is a build error related to alignment. Passing by const reference works.

I manually verified that the code compiles and works in Visual Studio 2015 with ETW exporter. I will cover full CI and other exporters (OTLP, Elastic, etc.) in a follow-up PR that would add Visual Studio 2015 to build loop.

### DIFF
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -23,7 +23,7 @@ private:
 
   // A constructor that sets the token's Context object to the
   // one that was passed in.
-  Token(Context context) : context_(context) {}
+  Token(const Context& context) : context_(context) {}
 
   const Context context_;
 };
@@ -49,7 +49,7 @@ public:
    * @param the new current context
    * @return a token for the new current context. This never returns a nullptr.
    */
-  virtual nostd::unique_ptr<Token> Attach(Context context) noexcept = 0;
+  virtual nostd::unique_ptr<Token> Attach(const Context& context) noexcept = 0;
 
   /**
    * Detach the context related to the given token.
@@ -59,7 +59,7 @@ public:
   virtual bool Detach(Token &token) noexcept = 0;
 
 protected:
-  nostd::unique_ptr<Token> CreateToken(Context context) noexcept
+  nostd::unique_ptr<Token> CreateToken(const Context& context) noexcept
   {
     return nostd::unique_ptr<Token>(new Token(context));
   }
@@ -82,7 +82,7 @@ public:
 
   // Sets the current 'Context' object. Returns a token
   // that can be used to reset to the previous Context.
-  static nostd::unique_ptr<Token> Attach(Context context) noexcept
+  static nostd::unique_ptr<Token> Attach(const Context& context) noexcept
   {
     return GetRuntimeContextStorage()->Attach(context);
   }
@@ -98,7 +98,7 @@ public:
   // mind that the current RuntimeContext will not be changed, and the new
   // context will be returned.
   static Context SetValue(nostd::string_view key,
-                          ContextValue value,
+                          const ContextValue& value,
                           Context *context = nullptr) noexcept
   {
     Context temp_context;
@@ -204,7 +204,7 @@ public:
 
   // Sets the current 'Context' object. Returns a token
   // that can be used to reset to the previous Context.
-  nostd::unique_ptr<Token> Attach(Context context) noexcept override
+  nostd::unique_ptr<Token> Attach(const Context& context) noexcept override
   {
     GetStack().Push(context);
     return CreateToken(context);
@@ -254,7 +254,7 @@ private:
 
     // Pushes the passed in context pointer to the top of the stack
     // and resizes if necessary.
-    void Push(Context context) noexcept
+    void Push(const Context& context) noexcept
     {
       size_++;
       if (size_ > capacity_)
@@ -275,7 +275,13 @@ private:
       Context *temp = new Context[new_capacity];
       if (base_ != nullptr)
       {
-        std::copy(base_, base_ + old_size, temp);
+        // vs2015 does not like this construct considering it unsafe:
+        // - std::copy(base_, base_ + old_size, temp);
+        // Ref. https://stackoverflow.com/questions/12270224/xutility2227-warning-c4996-std-copy-impl
+        for (size_t i = 0; i < (std::min)(old_size, new_capacity); i++)
+        {
+         temp[i] = base_[i];
+        }
         delete[] base_;
       }
       base_ = temp;
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -23,7 +23,7 @@ private:
 
   // A constructor that sets the token's Context object to the
   // one that was passed in.
-  Token(const Context& context) : context_(context) {}
+  Token(const Context &context) : context_(context) {}
 
   const Context context_;
 };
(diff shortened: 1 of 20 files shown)


api/include/opentelemetry/context/runtime_context.h
```diff
@@ -275,7 +275,14 @@ class ThreadLocalContextStorage : public RuntimeContextStorage
       Context *temp = new Context[new_capacity];
       if (base_ != nullptr)
       {
-        std::copy(base_, base_ + old_size, temp);
+        // vs2015 does not like this construct considering it unsafe:
```
ThomsonTan, line 278: Check Visual Studio 2015 compiler like via `_MSC_VER` to separate the workaround for it?

api/include/opentelemetry/nostd/absl/README.md
```diff
@@ -0,0 +1,9 @@
+# Notes on Abseil Variant implementation
+
+This is a snapshot of Abseil Variant `absl::variant` from Abseil `v2020-03-03#8`.
```
ThomsonTan, line 3: As this PR includes some sources files form abseil, is it compatible with our current license?

api/include/opentelemetry/nostd/absl/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This code is required to compile OpenTelemetry code in vs2015 because MPark Variant implementation is not compatible with vs2015.
+
+Build option `HAVE_ABSEIL_VARIANT` allows to enable the build with `absl::variant`, `absl::get` and `absl::visit` as defalt implementation for `nostd::` classes.
```
pyohannes, line 7: Is there a reason we have to ship two variant implementations? If `absl` is needed for VS2015 (and we decide we want to support that), can't we just generally use the `absl` implementation (instead of the MPark one)?

maxgolov: I'd like to take a two-staged approach. We should use Abseil everywhere, but I'd like to keep it under the build option right now until I fully test all existing exporters with it. Within the same OS class the ABI compatibility is retained irrespective of what implementation we take, e.g. Mpark works for Linux and Mac no problem. Whereas Abseil works for Windows. I want to send a follow-up PR that adds a build loop for Windows to switch over to Abseil for Variant. Then we can do this for Linux and Mac.

api/include/opentelemetry/nostd/absl/README.md
maxgolov: Yes, it is the same license. Apache License v2.0. It is better than MPark in this regard. Removing MPark eventually, we can get rid of the Boost license requirement.

api/include/opentelemetry/context/runtime_context.h
maxgolov: Does `std::copy` itself perform a bounds checking? i.e. what if new temp size - `new_capacity` is less than the old capacity? Since it's a raw pointer, I think the old code had a problem.

ThomsonTan (approved):

api/include/opentelemetry/nostd/absl/README.md
maxgolov: I think long-term before v1.0 GA we should switch to Abseil for variant. That'd be great for another reason: we can get rid of Boost license in our list of licenses because only MPark variant is Boost License, whereas Abseil is also Apache License v2.0 .. but I'd like to do this switch in a separate PR.

maxgolov: #68

maxgolov: #68

pyohannes: Makes sense. Can you add an additional issue for this (apart from the license one), otherwise this might be lost when just looking at issue titles.

I'd be fine with switching to absl as default, having MPark under a switch (then our CI pipeline would verify how it's working). However, I assume you expect changes to be necessary for it to work on Linux.

pyohannes (approved):
Approved, provided we work toward getting rid of the MPark `variant` implementation.

api/include/opentelemetry/nostd/absl/README.md
lalitb: Have created an issue for this #572 , and assigned to @maxgolov . Feel free to update with relevant details on it.

Issue #567, opened 2021-02-09
Implement InstrumentationLibrary Instance in SDK


lalitb:
As per specs for [Tracer Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/1a0395e482d4fec6faab0ee4b7bc1dc5fdd3e493/specification/trace/sdk.md#tracer-creation) :

`New Tracer instances are always created through a TracerProvider (see API). The name and version arguments supplied to the TracerProvider must be used to create an InstrumentationLibrary instance which is stored on the created Tracer.`

Currently for cpp, we do pass instrumentation library name and version in tracer sdk, but it is never used internally to create `InstrumentationLibrary ` instance, and further passing it to otel (and other) exporters: https://github.com/open-telemetry/opentelemetry-cpp/blob/25a7178a8d703407f4bbf322c16eb044b58c9a2a/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L35-L37 

As per the otel protocol:
https://github.com/open-telemetry/opentelemetry-proto/blob/e73df7c9f3c474c992990e82aa09ee46c39d6aab/opentelemetry/proto/common/v1/common.proto#L72-L78:

```proto
// InstrumentationLibrary is a message representing the instrumentation library information
// such as the fully qualified name and version. 
message InstrumentationLibrary {
  // An empty instrumentation library name means the name is unknown. 
  string name = 1;
  string version = 2;
}
```

 

`

ThomsonTan:
Do we need create a definition of `InstrumentationLibrary` for SDK or reuse the definition from OTLP? Reuse it will introduce dependency on protobuf.

lalitb:
> Do we need create a definition of `InstrumentationLibrary` for SDK or reuse the definition from OTLP? Reuse it will introduce dependency on protobuf.

Yes, we need to create SDK implementation, and then propagate to otlp exporter.

PR #570, opened 2021-02-09
Add Resources to Recordable interface.

lalitb:
This is second part of initial Resource SDK implementation ( https://github.com/open-telemetry/opentelemetry-cpp/pull/502 ) - to propagate resources to Span exporters through Recordable interface.

TODO -> Modify otlp exporter to export resources. This is dependent on #567 , as the spans need to be grouped on basis on Resources and Instrumentation Library before exporting.

@jsuereth  - It doesn't use alternate approach from processor to exporters, as that would need re-work if Resources are later made as mutable / append-only attributes ( as there are some discussion happening on this (and you are already part of that disucssions :) ) : https://github.com/open-telemetry/opentelemetry-specification/issues/1298 ).

Moving to draft, as per comments from Josh.

### DIFF
--- a/exporters/otlp/src/otlp_exporter.cc
+++ b/exporters/otlp/src/otlp_exporter.cc
@@ -23,10 +23,22 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>> 
   auto resource_span       = request->add_resource_spans();
   auto instrumentation_lib = resource_span->add_instrumentation_library_spans();
 
+  // TBD -> Group using Resources and Instrumentation Lib ( version and name)
+  bool resource_added = false;
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
     *instrumentation_lib->add_spans() = std::move(rec->span());
+
+    if (!resource_added)
+    {
+      auto resource = rec->GetResources();
+      if (resource.attributes_size())
+      {
+        resource_span->set_allocated_resource(&resource);
+        resource_added = true;
+      }
+    }
   }
 }
 
--- a/exporters/otlp/src/otlp_exporter.cc
+++ b/exporters/otlp/src/otlp_exporter.cc
@@ -35,7 +35,8 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>> 
       auto resource = rec->GetResources();
       if (resource.attributes_size())
       {
-        resource_span->set_allocated_resource(&resource);
+        // TBD
+        // resource_span->set_allocated_resource(&resource);
         resource_added = true;
       }
     }
--- a/exporters/otlp/src/otlp_exporter.cc
+++ b/exporters/otlp/src/otlp_exporter.cc
@@ -35,7 +35,7 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>> 
       auto resource = rec->GetResources();
       if (resource.attributes_size())
       {
-        // TBD
+        // TODO
         // resource_span->set_allocated_resource(&resource);
         resource_added = true;
       }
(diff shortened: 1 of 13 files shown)


exporters/otlp/src/otlp_exporter.cc
```diff
@@ -23,10 +23,23 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>>
...
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
     *instrumentation_lib->add_spans() = std::move(rec->span());
+
+    if (!resource_added)
+    {
+      auto resource = rec->GetResources();
+      if (resource.attributes_size())
+      {
+        // TODO
```
jsuereth, line 38: Is this ready for review?

exporters/otlp/src/recordable.cc
```diff
@@ -117,6 +117,101 @@ void PopulateAttribute(opentelemetry::proto::common::v1::KeyValue *attribute,
...
+                               const opentelemetry::sdk::common::OwnedAttributeValue &value)
+{
+
+  attribute->set_key(key.data(), key.size());
+
+  if (nostd::holds_alternative<bool>(value))
+  {
+    attribute->mutable_value()->set_bool_value(nostd::get<bool>(value));
+  }
+
```
ThomsonTan, line 131: nit: remove blank line.

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -121,6 +121,16 @@ class SpanData final : public Recordable
    */
   opentelemetry::trace::SpanKind GetSpanKind() const noexcept { return span_kind_; }
 
+  /**
+   * Get Resources.
+   * @return the resource of this span
+   */
+  const std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>
+      &GetResources() const noexcept
```
jsuereth, line 129: Why doesn't this return a `Resource` object?

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -213,6 +223,14 @@ class SpanData final : public Recordable
     span_kind_ = span_kind;
   }
 
+  void SetResourceAttribute(
```
jsuereth, line 226: I'm trying to understand what this method is intended to do and where it is in the spec.

Fro what I gather, this is basically copying all the resource-attributes onto the span.  Can we instead just grab a reference to the resource?

jsuereth (commented):
IIUC, rather than passing resource to exporter, we're copying all the labels across spans and assuming they all share the same set?

This code was a bit harder to follow.    WDYT about one of the following:

1. Keep the `Resource` bundled into that type instead of using the name "resource attribtues"
2. Modify `MakeRecordable` interfaces to take in a resource
3. I'm happy to take a stab at an interface for sending resource from the TracerProvider.  (Specifically I'd add a `setResource` on `Processor` that is called when a processor is added to the `TracerProvider`).

lalitb:
> IIUC, rather than passing resource to exporter, we're copying all the labels across spans and assuming they all share the same set?
> 
> This code was a bit harder to follow. WDYT about one of the following:
> 
> 1. Keep the `Resource` bundled into that type instead of using the name "resource attribtues"
> 2. Modify `MakeRecordable` interfaces to take in a resource
> 3. I'm happy to take a stab at an interface for sending resource from the TracerProvider.  (Specifically I'd add a `setResource` on `Processor` that is called when a processor is added to the `TracerProvider`).

Thanks @jsuereth for your comments.  I would prefer option 3 unless Resource is made append-only/mutable in spec. Will go with option 2 in that case. Would wait for more suggestions before doing changes accordingly.

jsuereth:
@lalitb Sounds like a plan.  I'll get something to you as soon as I can (this week, likely)

jsuereth:
Ok #575 is ready for comment/comparison.

lalitb:
closing as we are looking into alternate options of implementation ref: #580 , #591

PR #580, opened 2021-02-12
Add resource to Recordable, make room for InstrumentationLibrary

jsuereth:
This PR Is the result of discussion/design from #575.

- [x] Adds a `SetResourceRef` to the `Recordable` interface
- [x] Removes `Tracer` methods not in the specification.
- [x] Update `Tracer` to get `Processor`/`Exporter` solely from  `TracerProvider`
- [x] Update `Span` to keep a reference to generating `Tracer` and pull information from it rather than duplicating
- [x] Ensure `SetResourceRef` is called on `Recordable` when creating spans
- [x] Update OTLP exporter to also export `Resource`
- [x] Update OTLP exporter tests to check contents of exported protos
- [x] Fix windows exporters/builds (ETW)

For a separate PR?:
- [ ] Update `SpanProcessor` to be `unique_ptr` on `TracerProvider`.
- [ ] Update `Exporter` to be `unique_ptr` on `SpanProcessor`s.
- [ ] Add `InstrumentationLirbary` class
- [ ] Update `TracerProvider` to construct multiple `Tracer`s, each remembering their `InstrumentationLibrary`

### DIFF
--- a/exporters/otlp/src/otlp_exporter.cc
+++ b/exporters/otlp/src/otlp_exporter.cc
@@ -22,15 +22,16 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>> 
 {
   auto resource_span       = request->add_resource_spans();
   auto instrumentation_lib = resource_span->add_instrumentation_library_spans();
-  bool has_resource = false;
+  bool has_resource        = false;
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
     // We assume all the spans are for the same resource.
-    if (!has_resource) {
+    if (!has_resource)
+    {
       // *resource_span->mutable_resource() = std::move(rec->resource());
       *resource_span->mutable_resource() = rec->resource();
-      has_resource = true;
+      has_resource                       = true;
     }
     *instrumentation_lib->add_spans() = std::move(rec->span());
   }
--- a/exporters/otlp/src/otlp_exporter.cc
+++ b/exporters/otlp/src/otlp_exporter.cc
@@ -22,10 +22,16 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>> 
 {
   auto resource_span       = request->add_resource_spans();
   auto instrumentation_lib = resource_span->add_instrumentation_library_spans();
-
+  bool has_resource = false;
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
+    // We assume all the spans are for the same resource.
+    if (!has_resource) {
+      // *resource_span->mutable_resource() = std::move(rec->resource());
+      *resource_span->mutable_resource() = rec->resource();
+      has_resource = true;
+    }
     *instrumentation_lib->add_spans() = std::move(rec->span());
   }
 }
--- a/sdk/include/opentelemetry/sdk/trace/recordable.h
+++ b/sdk/include/opentelemetry/sdk/trace/recordable.h
@@ -44,14 +44,15 @@ public:
 
   /**
    * Sets the resource associated with a span.
-   * 
+   *
    * <p> The resource is guaranteed to have a lifetime longer than this recordable.  It is
    * owned by the `TracerProvider` which (indirectly) owns the `Exporter` that generates a
    * `Recordable`.
-   * 
+   *
    * @param resource pointer to the Resource for this span.
    */
-  virtual void SetResourceRef(const opentelemetry::sdk::resource::Resource*const resource) noexcept = 0;
+  virtual void SetResourceRef(
+      const opentelemetry::sdk::resource::Resource *const resource) noexcept = 0;
 
   /**
    * Set an attribute of a span.
--- a/sdk/include/opentelemetry/sdk/trace/recordable.h
+++ b/sdk/include/opentelemetry/sdk/trace/recordable.h
@@ -49,10 +49,11 @@ public:
    * owned by the `TracerProvider` which (indirectly) owns the `Exporter` that generates a
    * `Recordable`.
    *
-   * @param resource pointer to the Resource for this span.
+   * @param resource the Resource for this span.
+   *                 Note: this reference will remain stable for the life of the Recordable.
    */
   virtual void SetResourceRef(
-      const opentelemetry::sdk::resource::Resource *const resource) noexcept = 0;
+      const opentelemetry::sdk::resource::Resource &resource) noexcept = 0;
 
   /**
    * Set an attribute of a span.
--- a/sdk/include/opentelemetry/sdk/trace/recordable.h
+++ b/sdk/include/opentelemetry/sdk/trace/recordable.h
@@ -5,6 +5,7 @@
 #include "opentelemetry/core/timestamp.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/common/empty_attributes.h"
+#include "opentelemetry/sdk/resource/resource.h"
 #include "opentelemetry/trace/canonical_code.h"
 #include "opentelemetry/trace/span.h"
 #include "opentelemetry/trace/span_context.h"
@@ -40,6 +41,17 @@ public:
   virtual void SetIds(opentelemetry::trace::TraceId trace_id,
                       opentelemetry::trace::SpanId span_id,
                       opentelemetry::trace::SpanId parent_span_id) noexcept = 0;
+
+  /**
+   * Sets the resource associated with a span.
+   * 
+   * <p> The resource is guaranteed to have a lifetime longer than this recordable.  It is
+   * owned by the `TracerProvider` which (indirectly) owns the `Exporter` that generates a
+   * `Recordable`.
+   * 
+   * @param resource pointer to the Resource for this span.
+   */
+  virtual void SetResourceRef(const opentelemetry::sdk::resource::Resource*const resource) noexcept = 0;
 
   /**
    * Set an attribute of a span.
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -26,7 +26,7 @@ public:
    * Initialize a new tracer.
    * @param parent The `TraceProvider` which owns resource + pipeline for traces.
    */
-  explicit Tracer(TracerProvider* provider) noexcept;
+  explicit Tracer(TracerProvider *provider) noexcept;
 
   nostd::shared_ptr<trace_api::Span> StartSpan(
       nostd::string_view name,
@@ -39,10 +39,10 @@ public:
   void CloseWithMicroseconds(uint64_t timeout) noexcept override;
 
   /** Returns the provider that created this Tracer. */
-  const TracerProvider& GetProvider() const { return *provider_; }
+  const TracerProvider &GetProvider() const { return *provider_; }
 
 private:
-  TracerProvider* provider_;
+  TracerProvider *provider_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -15,36 +15,18 @@ namespace sdk
 {
 namespace trace
 {
+
+// Forward Declaration due to bidirectional dependency
+class TracerProvider;
+
 class Tracer final : public trace_api::Tracer, public std::enable_shared_from_this<Tracer>
 {
 public:
   /**
    * Initialize a new tracer.
-   * @param processor The span processor for this tracer. This must not be a
-   * nullptr.
+   * @param parent The `TraceProvider` which owns resource + pipeline for traces.
    */
-  explicit Tracer(std::shared_ptr<SpanProcessor> processor,
-                  const opentelemetry::sdk::resource::Resource &resource,
-                  std::shared_ptr<Sampler> sampler = std::make_shared<AlwaysOnSampler>()) noexcept;
-
-  /**
-   * Set the span processor associated with this tracer.
-   * @param processor The new span processor for this tracer. This must not be
-   * a nullptr.
-   */
-  void SetProcessor(std::shared_ptr<SpanProcessor> processor) noexcept;
-
-  /**
-   * Obtain the span processor associated with this tracer.
-   * @return The span processor for this tracer.
-   */
-  std::shared_ptr<SpanProcessor> GetProcessor() const noexcept;
-
-  /**
-   * Obtain the sampler associated with this tracer.
-   * @return The sampler for this tracer.
-   */
-  std::shared_ptr<Sampler> GetSampler() const noexcept;
+  explicit Tracer(const TracerProvider& provider) noexcept;
 
   nostd::shared_ptr<trace_api::Span> StartSpan(
       nostd::string_view name,
@@ -57,9 +39,7 @@ public:
   void CloseWithMicroseconds(uint64_t timeout) noexcept override;
 
 private:
-  opentelemetry::sdk::AtomicSharedPtr<SpanProcessor> processor_;
-  const std::shared_ptr<Sampler> sampler_;
-  const opentelemetry::sdk::resource::Resource &resource_;
+  const TracerProvider& provider_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -26,7 +26,7 @@ public:
    * Initialize a new tracer.
    * @param parent The `TraceProvider` which owns resource + pipeline for traces.
    */
-  explicit Tracer(const TracerProvider& provider) noexcept;
+  explicit Tracer(TracerProvider* provider) noexcept;
 
   nostd::shared_ptr<trace_api::Span> StartSpan(
       nostd::string_view name,
@@ -38,8 +38,11 @@ public:
 
   void CloseWithMicroseconds(uint64_t timeout) noexcept override;
 
+  /** Returns the provider that created this Tracer. */
+  const TracerProvider& GetProvider() const { return *provider_; }
+
 private:
-  const TracerProvider& provider_;
+  TracerProvider* provider_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/sdk/src/trace/span.cc
+++ b/sdk/src/trace/span.cc
@@ -108,7 +108,8 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   recordable_->SetSpanKind(options.kind);
   recordable_->SetStartTime(NowOr(options.start_system_time));
   start_steady_time = NowOr(options.start_steady_time);
-  // recordable_->SetResourceRef(tracer_->GetProvider().GetResource()); TODO
+  recordable_->SetResourceRef(&tracer_->GetProvider().GetResource());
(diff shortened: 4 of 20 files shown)


sdk/include/opentelemetry/sdk/trace/recordable.h
```diff
@@ -41,6 +42,18 @@ class Recordable
...
+   * Sets the resource associated with a span.
+   *
+   * <p> The resource is guaranteed to have a lifetime longer than this recordable.  It is
+   * owned by the `TracerProvider` which (indirectly) owns the `Exporter` that generates a
+   * `Recordable`.
+   *
+   * @param resource pointer to the Resource for this span.
+   */
+  virtual void SetResourceRef(
+      const opentelemetry::sdk::resource::Resource *const resource) noexcept = 0;
```
lalitb, line 55: No strong preference, but can we pass const reference to resource here ?

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -56,10 +38,11 @@ class Tracer final : public trace_api::Tracer, public std::enable_shared_from_th
...
   void CloseWithMicroseconds(uint64_t timeout) noexcept override;
 
+  /** Returns the provider that created this Tracer. */
+  const TracerProvider &GetProvider() const { return *provider_; }
+
 private:
-  opentelemetry::sdk::AtomicSharedPtr<SpanProcessor> processor_;
-  const std::shared_ptr<Sampler> sampler_;
-  const opentelemetry::sdk::resource::Resource &resource_;
+  TracerProvider *provider_;
```
lalitb, line 45: storing const reference to the provider here ?

lalitb (commented):
Thanks for implementing the approach. LGTM in general apart from comments using references instead of pointers.

sdk/include/opentelemetry/sdk/trace/recordable.h
```diff
@@ -41,6 +42,18 @@ class Recordable
                       opentelemetry::trace::SpanId span_id,
                       opentelemetry::trace::SpanId parent_span_id) noexcept = 0;
 
+  /**
+   * Sets the resource associated with a span.
+   *
+   * <p> The resource is guaranteed to have a lifetime longer than this recordable.  It is
```
pyohannes, line 48: > The resource is guaranteed to have a lifetime longer than this recordable.

I'm not sure if we can make this statement as is. I'm thinking about a scenario where the user frees the `TracerProvider`, but the exporter is still working on exporting spans (recordables), maybe in another thread.

If we go with plain pointers for resources (instead of `shared_ptr`), we'd need to be very careful, and either explicitly document all implicit assumptions we make, or take additional precautions to avoid problematic scenarios (e. g. making sure that in the `TracerProvider` destructor `Shutdown` is called for all exporters).

pyohannes (commented):
Thanks, I generally like this approach. I'm just trying to think through its implications (some first thoughts in inline comments).

sdk/include/opentelemetry/sdk/trace/recordable.h
jsuereth: I'm planning to change TracerProvider to (a) ensure shutdown and (b) require resource cleanup after shutdown.

Would you be amenable to that change in correlation with this?  I think it's needed to make this safe, and I'm happy to go update the docs.

sdk/include/opentelemetry/sdk/trace/tracer.h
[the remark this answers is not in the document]
jsuereth: (a) We can't enforce the lifetime of `Processor` at all if we share it across `Tracer`.  One goal here is to give each component of the SDK an explicit owner who controls when memory is terminated
(b) `TracerProvider` owns resource AND processor/sampler.  So the idea is that Tracers are tied to the lifetime of TracerProvider and can leverage both of those freely without sacrificing the "known lifetime" issue as you stated.

I'd love to know the Envoy use case.  AFAIK you shouldn't be able to get a `Tracer` without a `TracerProvider`, but I could be wrong.   I think `Tracer` need some work for `InstrumentationLibrary`, and likely that may push things this direction.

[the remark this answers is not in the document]
lalitb: One advantage of not having coupling between the two is that the `TracerProvider` and `Tracer` can be implemented independently by developers, without worrying about the undocumented interfaces provided/to-be-provided by `TracerProvider`.

lalitb:
> Thanks for implementing the approach. LGTM in general apart from comment about using reference instead of pointer for resources.

Ok thinking about the lifetime of resources, and scenarios where tracer provider is shutdown while the export is still happening, references mayn't be the right approach. So please ignore that suggestion.

exporters/otlp/src/otlp_exporter.cc
```diff
@@ -22,10 +22,17 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>>
...
+  bool has_resource        = false;
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
+    // We assume all the spans are for the same resource.
+    if (!has_resource)
+    {
+      // *resource_span->mutable_resource() = std::move(rec->resource());
+      *resource_span->mutable_resource() = rec->resource();
+      has_resource                       = true;
```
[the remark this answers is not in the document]
ThomsonTan: Or we could assert the later found resource equals the first one.

exporters/otlp/src/otlp_exporter.cc
```diff
@@ -22,10 +22,17 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>>
...
   auto instrumentation_lib = resource_span->add_instrumentation_library_spans();
-
+  bool has_resource        = false;
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
+    // We assume all the spans are for the same resource.
+    if (!has_resource)
+    {
+      // *resource_span->mutable_resource() = std::move(rec->resource());
```
ThomsonTan, line 32: nit: replace this comment with TODO support for moving?

sdk/include/opentelemetry/sdk/trace/tracer.h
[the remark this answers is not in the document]
pyohannes: That was one of the discussions mentioning the Envoy use case: https://github.com/open-telemetry/opentelemetry-cpp/pull/35#discussion_r381468576

Currently we can use tracers without tracer providers, I think that's not violating the spec. I don't know how many use cases are out there for that (unfortunately I'm myself not that familiar with Envoy). However, if we remove that ability we'd need to be deliberate and careful about it.

exporters/otlp/src/otlp_exporter.cc
jsuereth: Actually, theoretically it *is* moved, I should have removed it before :)

exporters/otlp/src/otlp_exporter.cc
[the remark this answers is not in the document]
jsuereth: We can't.  I'm using the loop that adds the spans to add the resource.  IF we `break` than we'll be dropping spans.

sdk/include/opentelemetry/sdk/trace/recordable.h
jsuereth: So I can do this, it's a bit "odd".  TL;DR: I didn't want to convert from a reference to a pointer as that causes me to cringe in the notion of "likely bug".  I found a way to not do that, but it's still a bit cringey, PTAL.

sdk/include/opentelemetry/sdk/trace/recordable.h
pyohannes: I'm definitely amenable to you making those changes.

Some other question: did you for performance reasons decide not to use a managed pointer (`std::shared_ptr`) for holding on to the resource in the recordable? And rather have implicit ownership via trace provider/recordable lifetimes?

So far we tried to make all ownership explicit to be on the (memory) safe side.

sdk/include/opentelemetry/sdk/trace/tracer.h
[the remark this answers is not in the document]
jsuereth: > Configuration (i.e., SpanProcessors, IdGenerator, SpanLimits and Sampler) MUST be managed solely by the TracerProvider and it MUST provide some way to configure all of them that are implemented in the SDK, at least when creating or initializing it.

See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracer-creation

So the envoy concern in question is the following:

> I could imagine users wanting to use a different ownership model than the TracerProvider that creates one tracer per named component and ties lifetime to a global.

I think having `Tracer` tie its lifetime *shorter* than a TracerProvider would still be possible here.  `Resource` *would* be global, but tracer could have shorter lifetime.  That's actually a significant API change we'd have to do though, and has tons of implicaitons around `INstrumentationLirbary` implementation.

jsuereth:
@lalitb  I did implement references prior to seeing this.....

lalitb:
> @lalitb I did implement references prior to seeing this.....
Yes, I can see issues with reference, and agree not being the right approach.

jsuereth:
@lalitb I can remove that commit.  However two questions for the group:

1. If make the other proposed changes (Exporter/Processor being `unique_ptr`s on TracerProvider) does that help the reference concern?
2. How do we want to proceed with `Tracer` going forward? (@pyohannes raises good concerns around per-thread `Tracer` but I think this are orthogonal to whether or not we have a central owner for processing tracers)

pyohannes:
> If make the other proposed changes (Exporter/Processor being unique_ptrs on TracerProvider) does that help the reference concern?

My main concern about the approach (which I generally like): 
1. The recordable has a pointer/reference to the resource which is owned by the tracer provider.
2. We pass ownership of the recordable to the exporter.
3. From that point on we don't have control over what happens with the recordable.

After that, the recordable might be batched by the exporter or another component it is forwarded to, at the same time the global tracer provider could be replaced (there's that possibility at least). To take care of that, we would need to come up with lots of implicit lifetime dependencies (like "shutting down an exporter invalidates all recordables created from it"), and we would need to make it a hard requirement for every exporter to implement `Shutdown` properly (currently it's a no-op for the OTLP exporter).

Maybe I'm too paranoid. However, what are your thoughts about this approach:
1. Storing the resource as a `shared_ptr` in the tracer provider.
1. Changing the signature of `SetResourceRef` to `void SetResouce(const std::shared_ptr<Resource>& resource)`.
1. Then each exporter/recordable can itself decide whether to keep a shared pointer (100% safe), or go the optimized route and keep a pointer/reference. So we keep all options open for exporter/recordable implementations, and don't make any assumptions in the API about how resource ownership should be implemented.

lalitb:
> 2\. How do we want to proceed with `Tracer` going forward? (@pyohannes raises good concerns around per-thread `Tracer` but I think this are orthogonal to whether or not we have a central owner for processing tracers)

Probably I am may be biased with existing design and would like to be corrected if so - but how about keeping the current approach of keeping  sampler, processor ( and now also including resource ) as `shared_ptr` within  `TracerProvider`, and let their life-time governed by references to those `shared_ptr`. If this approach provides the loose-coupling between the `TracerProvider` and `Tracer`, of-course at the cost of performance overhead of using shared_ptr, and not so well-defined life-time for processor/exporter.

jsuereth:
@pyohannes / @lalitb Two points:

1. I'm afraid that things like https://github.com/open-telemetry/opentelemetry-specification/pull/1452 will force us to define clear ownership + control of exporters/tracers.  I'm ok updating this to use `shared_ptr` for flexibility now, but it has my hackles up that we may be begging peter to pay paul.
2. I'm starting to debate whether passing `Resource` to a `Recorable is the right thing after all.  If you look at OTLP exporter, we basically have two requirements:
  - Split traces by `TracerProvider`and pull `Resource` *once*
  - Split traces by `Tracer` and pull `InstrumentationLibrary` *once* for the batch.
  - See the [java implementation](https://github.com/open-telemetry/opentelemetry-java/blob/5f8b6c9dbe293d925437e5473056da47392309db/sdk-extensions/otproto/src/main/java/io/opentelemetry/sdk/extension/otproto/SpanAdapter.java#L67)

We're not actually using `Recordable` in its intended way here for `Resource` or `InstrumentationLibrary`.  I think it'd be more accurate to the other languages if we store `Resource`/`InstrumentationLibrary` on [sdk::Span](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/trace/span.h) and update our `Exporter` API to take that.

a. We could make sure however we store it leads to efficient sorting/ordering in Exporters for the OTLP case (i.e. an efficient hash on `Resource` + `InstrumentationLibrary`
b. We wouldn't be forcing `Exporter`s to *all* implement a way of "carrying around the `Resource`".   They just need to use the reference we give them via Span.
c. We're not really leveraging the `Recordable` benefits for Resource today anyway.  The idea was to avoid constructing
   intermediate objects (which we are) and directly instantiate the runtime-output (e.g. protobuf).
d. I think this might be more uniform w/ other language implementations.

jsuereth:
Regarding the concerns around using shared pointer, I think we *have* to change the existing implementation no matter what.

See the specificaiton line:

> Configuration (i.e., SpanProcessors, IdGenerator, SpanLimits and Sampler) MUST be managed solely by the TracerProvider and it MUST provide some way to configure all of them that are implemented in the SDK, at least when creating or initializing it.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracer-creation

sdk/src/trace/span.cc
```diff
@@ -58,19 +59,17 @@ trace_api::SpanId GenerateRandomSpanId()
...
            const opentelemetry::common::KeyValueIterable &attributes,
            const trace_api::SpanContextKeyValueIterable &links,
            const trace_api::StartSpanOptions &options,
-           const trace_api::SpanContext &parent_span_context,
-           const opentelemetry::sdk::resource::Resource &resource) noexcept
+           const trace_api::SpanContext &parent_span_context) noexcept
     : tracer_{std::move(tracer)},
-      processor_{processor},
-      recordable_{processor_->MakeRecordable()},
+      recordable_{nullptr},
```
ThomsonTan, line 68: This initialization is unnecessary?

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -15,36 +15,18 @@ namespace sdk
...
-   */
-  std::shared_ptr<SpanProcessor> GetProcessor() const noexcept;
-
-  /**
-   * Obtain the sampler associated with this tracer.
-   * @return The sampler for this tracer.
+   * @param parent The `TraceProvider` which owns resource + pipeline for traces.
    */
-  std::shared_ptr<Sampler> GetSampler() const noexcept;
+  explicit Tracer(TracerProvider *provider) noexcept;
```
ThomsonTan, line 29: Could we make the ctor of `Tracer` private and only accessible only in `TracerProvider`?

pyohannes:
Just another idea I had. What about putting all `TracerProvider` related state in a separate type, like `TracerContext`:
```c++
struct TracerContext {
  std::unique_ptr<SpanProcessor> processor;
  std::unique_ptr<Sampler> sampler;
  std::shared_ptr<Resource> resource;
};
```

The `TracerProvider` would hold a `std::shared_ptr<TracerContext>`, with setters to change processor and sampler.

The `Tracer`s would also hold a `std::shared_ptr<TracerContext>`, which points to the one on the related `TracerProvider`. The `InstrumentationLibrary` would separately live on each tracer.

In this way, `TracerProvider` and `Tracer` would be decoupled, while still sharing state. `Tracer` could be used without `TracerProvicer` (by using another way to manage `TracerContext`).

I think it cleans up component dependencies in the SDK, however it doesn't give us an efficient way to pass `Resource` and `InstrumentationLibrary` along to the exporter. I think that's a separate problem, also related to the fact that the spec is poorly written in that specific regard.

lalitb:
> Just another idea I had. What about putting all `TracerProvider` related state in a separate type, like `TracerContext`:
> 
> ```c++
> struct TracerContext {
>   std::unique_ptr<SpanProcessor> processor;
>   std::unique_ptr<Sampler> sampler;
>   std::shared_ptr<Resource> resource;
> };
> ```
> 
> The `TracerProvider` would hold a `std::shared_ptr<TracerContext>`, with setters to change processor and sampler.
> 
> The `Tracer`s would also hold a `std::shared_ptr<TracerContext>`, which points to the one on the related `TracerProvider`. The `InstrumentationLibrary` would separately live on each tracer.
> 
> In this way, `TracerProvider` and `Tracer` would be decoupled, while still sharing state. `Tracer` could be used without `TracerProvicer` (by using another way to manage `TracerContext`).
> 
> I think it cleans up component dependencies in the SDK, however it doesn't give us an efficient way to pass `Resource` and `InstrumentationLibrary` along to the exporter. I think that's a separate problem, also related to the fact that the spec is poorly written in that specific regard.

Thanks @pyohannes for the proposal - Just to be clear to me, the proposal is to have cleaner approach for removing the dependency between `TracerProvider` and `Tracer` and still sharing the state. It is not meant to ensure that configuration (processor, sampler ) is solely owned by `TracerProvider`. As with this approach, we can still have multiple providers share the same `TracerContext`.

lalitb:
> We're not actually using `Recordable` in its intended way here for `Resource` or `InstrumentationLibrary`. I think it'd be more accurate to the other languages if we store `Resource`/`InstrumentationLibrary` on [sdk::Span](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/trace/span.h) and update our `Exporter` API to take that.

@jsuereth  - I like this approach of using sdk::span to store the reference to Resource to have consistent implementation with other languages, just wondering how it would be different and efficient from current approach of storing reference to Resource in otlp `Recordable`. We would still need to iterate over spans/recordables to group by Resources/InstrumentationLib in otlp exporter? And not all exporters may want to process resources ( ? ), so in that case we would store it's reference in Span, but won't be used anywhere. Probably, I am not fully able to visualize this approach.

jsuereth:
@pyohannes With the caveat that we do multithreading primitives correctly around updates, I think that could work for decoupling Tracer from TracerProvider, while still meeting the spec requirement around the processor.

Should I take a crack at this now as part of this CL?

jsuereth:
@lalitb 

> We would still need to iterate over spans/recordables to group by Resources/InstrumentationLib in otlp exporter? And not all exporters may want to process resources ( ? ), so in that case we would store it's reference in Span, but won't be used anywhere. Probably, I am not fully able to visualize this approach

yeah, OTLP requires grouping by Resource/InstrumentationLibrary by its design.   Alternatively we can store a map from IL => ILSpans and Resource => ResourceSpans to do that lookup/grouping as we iterate, depends on which ends up being faster.

I won't be able to get to showing this code until thursday, but I'll try to have something then for you to see what I'm suggesting.  We can discuss in the SiG :)

pyohannes:
>  Just to be clear to me, the proposal is to have cleaner approach for removing the dependency between TracerProvider and Tracer and still sharing the state. It is not meant to ensure that configuration (processor, sampler) is solely owned by TracerProvider ?

It ensures that the configuration is owned by a single `TracerContext`. If one uses a `TracerProvider`, the `TracerProvider` and the related `Tracer`s will have shared ownership of the `TraceContext` (and the configuration in it), but it's only writable via the `TracerProvider`. If one doesn't use a `TracerProvider`, one can let any other (custom) component give ownership of the `TraceContext` and take care of managing updates to it.

pyohannes:
> @pyohannes With the caveat that we do multithreading primitives correctly around updates, I think that could work for decoupling Tracer from TracerProvider, while still meeting the spec requirement around the processor.
>
> Should I take a crack at this now as part of this CL?

Maybe let's get other people's opinion first in tomorrow's SIG.

PR #581, opened 2021-02-12
Add .markdownlintignore file for external projects

ThomsonTan:
Exclude the external projects from our MarkdownLint check.


.markdownlintignore
```diff
@@ -0,0 +1,2 @@
+third_party/**
```
reyang, line 1: Is this a way we can put this file in a subfolder (and specify the path in the CI command line)? We would want to keep the top level folder as clean as possible (currently we already have too many files under the top level folder).

ThomsonTan: There is an option `markdownlint -p <ignore_file_full_path>`, but this also tries to help local run and the user don't need to figure out the option and ignore file to do the lint.

Probably I could move this to be part of our format script, then it would be fine to specify the ignore file path there. How do you think?

A good thing is the file starts with `.` is hidden file in Linux and is not listed by default, but I do agree this makes the github page a little mess.

reyang: Perhaps see if this file can be added to the `third_party` folder.
Not blocking though.

reyang (approved):

jsuereth (approved):

Issue #571, opened 2021-02-10
Add Getting Started / Quick Start guide


lalitb:
Probably as separate markdown doc which can be referenced from main page. 
As an example, how it is done by in Java:
https://github.com/open-telemetry/opentelemetry-java/blob/main/QUICKSTART.md

PrathameshAshra:
Can i take care of this ?
 I am planning to create a new .MD file and which will be a better version of
  https://github.com/open-telemetry/opentelemetry-cpp#quick-start
 the inspiration would be taken from 
 https://github.com/open-telemetry/opentelemetry-java/blob/main/QUICKSTART.md

lalitb:
Sure @PrathameshAshra . Thanks for extending the help. Assigning to you. Would be good if you can come up some initial doc sometime next week, and we can then incrementally enhance it subsequently.
cc @pyohannes as he has plan to integrate it with API doc.

PrathameshAshra:
thanks @lalitb, would keep @pyohannes in loop. meanwhile would try to come up with initial doc or something like template ish.

PrathameshAshra:
@pyohannes , any suggestions before I start  ? 
I am  looking forward for something like https://github.com/open-telemetry/opentelemetry-java/blob/main/QUICKSTART.md but for CPP

pyohannes:
The Java QuickStart guide is a good template to follow. We currently only need sections for _Set up_, _Tracing_, and _Tracing SDK Configuration_. We should also cover multi-threaded scenarios (like in this [example](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/multithreaded/main.cc)).

I added a very minimal _Getting started_ as a [main page of the API documentation](https://github.com/open-telemetry/opentelemetry-cpp/pull/582/files), but we also need something more comprehensive.

If you're in doubt it's best you get up a WIP PR with a rough draft or outline, and we can go from there.

lalitb:
Closing this, as we have a Getting started guide in place. Will create separate issues for SDK docs/api and target for GA milestone.

PR #582, opened 2021-02-16
Add main page for API documentation

pyohannes:
This PR adds a main page for the API reference documentation.

It also makes sure the version in the API documentation is updated by our release build scripts.

### DIFF
--- a/api/docs/Overview.md
+++ b/api/docs/Overview.md
@@ -14,7 +14,7 @@ exported and processed.
 The OpenTelemetry C++ API is provided as a header-only library and supports all
 recent versions of the C++ standard, down to C++11.
 
-A single application might dynamically or statically link different libraries
+A single application might dynamically or statically link to different libraries
 that were compiled with different compilers, while several of the linked
 libraries are instrumented with OpenTelemetry. OpenTelemetry C++ supports those
 scenarios by providing a stable ABI. This is achieved by a careful API design,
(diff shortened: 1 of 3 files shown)


lalitb (approved):
LGTM. Thanks for adding the initial document. Probably, we should be able to extend it further through #571 . Also, not related to this PR, but we would be needing Doxygen doc generation for resources and samplers (and few others ? ) from SDK. I will create a ticket for that.

pyohannes:
> Also, not related to this PR, but we would be needing Doxygen doc generation for resources and samplers (and few others ? ) from SDK.

Yes, but I think those should be separate. Instrumenters will only be interested in the API, and I think it would be confusing to have SDK parts mixed in there.

api/docs/Overview.md
```diff
@@ -0,0 +1,79 @@
...
+enables developers to instrument applications and libraries with the
+OpenTelemetry C++ API while being completely agnostic of how telemetry data is
+exported and processed.
+
+## Library design
+
+The OpenTelemetry C++ API is provided as a header-only library and supports all
+recent versions of the C++ standard, down to C++11.
+
+A single application might dynamically or statically link different libraries
```
ThomsonTan, line 17: nit: perhaps `link to`?

api/docs/Overview.md
```diff
@@ -0,0 +1,79 @@
...
+that were compiled with different compilers, while several of the linked
+libraries are instrumented with OpenTelemetry. OpenTelemetry C++ supports those
+scenarios by providing a stable ABI. This is achieved by a careful API design,
+and most notably by providing ABI stable versions of classes from the standard
+library. All those classes are provided in the `opentelemetry:nostd` namespace.
```
ThomsonTan, line 22: nit
```suggestion
library. All those classes are provided in the `opentelemetry::nostd` namespace.
```

ThomsonTan (approved):

jsuereth (approved):

PR #588, opened 2021-02-25
Add ForceFlush to TracerProvider in SDK

### LINKED ISSUE #587: Add ForceFlush to TracerProvider (SDK only)
As below PR merged to spec repo, we need add `ForceFlush` to `TracerProvider` in SDK.

https://github.com/open-telemetry/opentelemetry-specification/pull/1452

ThomsonTan:
Fixes #587 

## Changes

Add `ForceFlush` to TracerProvider as per the spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/1452

### DIFF
--- a/sdk/src/trace/tracer_provider.cc
+++ b/sdk/src/trace/tracer_provider.cc
@@ -48,6 +48,11 @@ bool TracerProvider::Shutdown() noexcept
 {
   return GetProcessor()->Shutdown();
 }
+
+bool TracerProvider::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  return GetProcessor()->Shutdown(timeout);
+}
 }  // namespace trace
 }  // namespace sdk
 OPENTELEMETRY_END_NAMESPACE
--- a/sdk/src/trace/tracer_provider.cc
+++ b/sdk/src/trace/tracer_provider.cc
@@ -51,7 +51,7 @@ bool TracerProvider::Shutdown() noexcept
 
 bool TracerProvider::ForceFlush(std::chrono::microseconds timeout) noexcept
 {
-  return GetProcessor()->Shutdown(timeout);
+  return GetProcessor()->ForceFlush(timeout);
 }
 }  // namespace trace
 }  // namespace sdk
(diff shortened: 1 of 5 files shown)


sdk/src/trace/tracer_provider.cc
```diff
@@ -48,6 +48,11 @@ bool TracerProvider::Shutdown() noexcept
 {
   return GetProcessor()->Shutdown();
 }
+
+bool TracerProvider::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  return GetProcessor()->Shutdown(timeout);
```
reyang, line 54: Curious - do we want to call ForceFlush or Shutdown? (I thought the latter)

lalitb: As per the specs (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush):
`ForceFlush MUST invoke ForceFlush on all registered SpanProcessors.`

So ForceFlush() method needs to be  defined both for SpanProcessor and Exporter. This can be done as separate PR and in that case Shutdown() call can be removed for now from above.

Note: PR to update specs for adding ForceFlush in SpanExporter is still open ( https://github.com/open-telemetry/opentelemetry-specification/pull/1467 )

ThomsonTan: Thanks for the catch, @reyang 

@lalitb I think we only support one `SpanProcessor` now. I'd prefer to add `ForceFlush` to `SpanExporter` in a separate PR.

reyang (approved):
LGTM.

sdk/src/trace/tracer_provider.cc
reyang: > I think we only support one `SpanProcessor` now.

That's fine, I think we could support multiple processors by using a de-mux processor.

Here are some examples:
* https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/CompositeProcessor.cs
* https://github.com/open-telemetry/opentelemetry-python/blob/99128b35d1a9cf2ecb04097c6a87d320a552a3e4/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L127

lalitb (approved):

pyohannes (approved):

PR #591, opened 2021-02-26
Add TraceState in SamplerResult, and use it for creating span context.

### LINKED ISSUE #556: Sampler: SamplingResult should return TraceState
As per the specs (https://github.com/open-telemetry/opentelemetry-specification/blob/ae31114e2d441f1ecd249ce11bb924a6ac54b5d6/specification/trace/sdk.md#sampler):
`SamplingResult` should contain : 
- A sampling Decision.
- A set of span Attributes that will also be added to the Span
- A Tracestate that will be associated with the Span through the new SpanContext.

This needs to be changed in current implementation:
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/trace/sampler.h#L39:L44

```cpp
struct SamplingResult
{
  Decision decision;
  // A set of span Attributes that will also be added to the Span. Can be nullptr.
  std::unique_ptr<const std::map<std::string, opentelemetry::common::AttributeValue>> attributes;
};
```

lalitb:
Fixes #556 

## Changes
Changes includes
   - Adding trace_state in sampling result. For custom samplers provided by sdk, this is trace_state from parent context ( if available), else default.
   - Passing this trace_state in `sdk::span` constructor to be used for creating current span context.
   - Use this trace_state while creating current span context  in `sdk::span`


sdk/src/trace/span.h
```diff
@@ -22,7 +22,9 @@ class Span final : public trace_api::Span
                 const trace_api::SpanContextKeyValueIterable &links,
                 const trace_api::StartSpanOptions &options,
                 const trace_api::SpanContext &parent_span_context,
-                const opentelemetry::sdk::resource::Resource &resource) noexcept;
+                const opentelemetry::sdk::resource::Resource &resource,
+                const nostd::shared_ptr<opentelemetry::trace::TraceState> trace_state =
```
ThomsonTan, line 26: FYI, this might be conflicted with the `trace_flags` change which needs to be resolved.

ThomsonTan (approved):

pyohannes (approved):

PR #595, opened 2021-03-02
Add support for sphinx as a documentation generator

pyohannes:
This PR adds support for sphinx as a documentation generator. sphinx was chosen as it is the framework of choice to publish documentation to https://readthedocs.org. 

Under the hood it still uses doxygen to create reStructuredText files, which are then used by sphinx. So we don't need to change anything with our existing practices for API documentation.

Currently this includes a root project in `docs/public`, which can link specific subprojects. Currently the only subproject linked is the API documentation in `api/docs`.

This is how it looks on _readthedocs_: https://opentelemetry-c-api.readthedocs.io/en/latest/

When this PR is approved, I will set up a _readthedocs_ project based on this repository, the above is just a placeholder to show how the final result will look like.


api/docs/conf.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'OpenTelemetry C++ API'
+copyright = '2021, OpenTelemetry authors'
+author = 'OpenTelemetry authors'
+
+# The full version, including alpha/beta/rc tags
+release = '0.1.0'
```
ThomsonTan, line 25: Update this to 0.2.0?

pyohannes: Good point, I fixed it.

docs/public/requirements.txt
```diff
@@ -0,0 +1,2 @@
+breathe
```
lalitb, line 1: Do we need to add `sphinx` too in install dependencies, or the plugins installation will take care of it automatically ?

lalitb (approved):
Thanks for adding this. LGTM.  Sphinx looks really cool - much better docs from doxygen genear html.  Couple of things:

1. Hopefully it should be possible to extend this documentation further to include member functions as part of class documentation. Ex, to add member functions for `Tracer` class:  https://opentelemetry-c-api.readthedocs.io/en/latest/otel_api/classopentelemetry_1_1trace_1_1Span.html#exhale-class-classopentelemetry-1-1trace-1-1Span

2. As part of release process, do we need to invoke some webhook to let `readthedocs` know about the new changes, so that it can regenerate the doc. In case this is not planned as part of `readthedocs` setup, we can have a github issue? I do plan to have github action for release, and can take care of this.

pyohannes:
> Hopefully it should be possible to extend this documentation further to include member functions as part of class documentation.
This should already be there. I'll check why it doesn't show up.

pyohannes:
> As part of release process, do we need to invoke some webhook to let readthedocs know about the new changes, so that it can regenerate the doc.
The _readthedocs_ build automatically kicks off when we push to registered branches. Branches are registered on the _readthedocs_ site, all the documentation building happens in an environment provided by _readthedocs_, so we do not need a GitHub action on our side.

What we would need to do on our side: for every new release we will need to register the release/branch on _readthedocs_. I assume we will have `main` registered for the version `latest`.

ThomsonTan (approved):

PR #597, opened 2021-03-05
 Add composite propagator

### LINKED ISSUE #541: Implement Composite Propagator
As per specs:

https://github.com/open-telemetry/opentelemetry-specification/blob/11cc73939a32e3a2e6f11bdeab843c61cf8594e9/specification/context/api-propagators.md#composite-propagator

lalitb:
Fixes #541 

## Changes

Add composite propagator as per specs: 
https://github.com/open-telemetry/opentelemetry-specification/blob/14d123c121b6caa53bffd011292c42a181c9ca26/specification/context/api-propagators.md#composite-propagator


api/test/trace/propagation/composite_propagator_test.cc
```diff
@@ -0,0 +1,122 @@
...
+  context::Context ctx1 = context::Context{};
+
+  context::Context ctx2 = composite_propagator_->Extract(Getter, carrier, ctx1);
+
+  auto ctx2_span = ctx2.GetValue(trace::kSpanKey);
+  EXPECT_TRUE(nostd::holds_alternative<nostd::shared_ptr<trace::Span>>(ctx2_span));
+
+  auto span = nostd::get<nostd::shared_ptr<trace::Span>>(ctx2_span);
+
+  // confirm last propagator in composite propagator list ( B3 here) wins for same key
```
reyang, line 98:
```suggestion
  // confirm last propagator in composite propagator list (B3 here) wins for same key
```

reyang (approved):

api/include/opentelemetry/trace/propagation/composite_propagator.h
```diff
@@ -0,0 +1,73 @@
...
+namespace trace
+{
+namespace propagation
+{
+
+template <typename T>
+class CompositePropagator : public TextMapPropagator<T>
+{
+public:
+  CompositePropagator(std::vector<std::shared_ptr<TextMapPropagator<T>>> &propagators)
```
seemk, line 16: A few minor nitpicks:
* Should `CompositePropagator` own the propagators? i.e. use unique_ptr over shared_ptr?
* Pass by value and `std::move`?
* Would it help to add a method to add a propagator to composite propagator?

api/test/trace/propagation/composite_propagator_test.cc
```diff
@@ -0,0 +1,122 @@
+#include "opentelemetry/context/context.h"
+#include "opentelemetry/nostd/shared_ptr.h"
+#include "opentelemetry/nostd/span.h"
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/trace/default_span.h"
+#include "opentelemetry/trace/noop.h"
+#include "opentelemetry/trace/span.h"
+#include "opentelemetry/trace/span_context.h"
+#include "opentelemetry/trace/trace_id.h"
+#include "opentelemetry/trace/tracer.h"
```
seemk, line 10: Are all of these includes needed or only some of them?

api/test/trace/propagation/composite_propagator_test.cc
```diff
@@ -0,0 +1,122 @@
...
+
+#include <map>
+#include <memory>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "opentelemetry/trace/default_span.h"
```
seemk, line 18: Seems a duplicate include.

seemk (approved):

api/include/opentelemetry/trace/propagation/composite_propagator.h
lalitb: > * Should `CompositePropagator` own the propagators? i.e. use unique_ptr over shared_ptr?
> * Pass by value and `std::move`?

Very valid points. I can't think of scenario where we need to share Propagator across. Will fix this, unless someone further comments with some scenario.

> * Would it help to add a method to add a propagator to composite propagator?

This should help if we need to add propagator after creating the initial composite object. Specs ( https://github.com/open-telemetry/opentelemetry-specification/blob/14d123c121b6caa53bffd011292c42a181c9ca26/specification/context/api-propagators.md#composite-propagator ) doesn't ask for it explicitly so would keep it as it for now. We can add later if required.

api/test/trace/propagation/composite_propagator_test.cc
lalitb: Not all, will remove.

api/test/trace/propagation/composite_propagator_test.cc
```diff
@@ -0,0 +1,120 @@
...
+
+static nostd::string_view Getter(const std::map<std::string, std::string> &carrier,
+                                 nostd::string_view trace_type)
+{
+  auto it = carrier.find(std::string(trace_type));
+  if (it != carrier.end())
+  {
+    return nostd::string_view(it->second);
+  }
+  return "";
```
ThomsonTan, line 36: Is it better to return `nullptr` here?

api/include/opentelemetry/trace/propagation/composite_propagator.h
```diff
@@ -0,0 +1,73 @@
...
+   * @param context Context to add values to
+   * @param carrier Carrier from which to extract context
+   */
+  context::Context Extract(Getter getter,
+                           const T &carrier,
+                           context::Context &context) noexcept override
+  {
+    for (auto &p : propagators_)
+    {
+      context = p->Extract(getter, carrier, context);
```
ThomsonTan, line 63: There seems no need to assign all intermediate extracted context to input context. Probably create temp `context::Context` to hold the extracted context and return the last one?

Is it possible to get empty `propagators_` list, then return the input parameter context? Do we need initialize the context to empty in such case?

api/include/opentelemetry/trace/propagation/composite_propagator.h
lalitb: First two points are implemented now.

api/test/trace/propagation/composite_propagator_test.cc
lalitb: done

api/include/opentelemetry/trace/propagation/composite_propagator.h
lalitb: > There seems no need to assign all intermediate extracted context to input context. Probably create temp `context::Context` to hold the extracted context and return the last one?

Good point - I did implement this initially, but then later discovered that all other implementations are modifying the input context and so reverted. Specs doesn't say anything about it, so was unsure. Have changed it back as you suggested now. Let me know if you see any concerns here :)

> Is it possible to get empty propagators_ list, then return the input parameter context? Do we need initialize the context to empty in such case?

Yes it's possible to get empty list, and we should be returning the input context unchanged it that case.

api/test/trace/propagation/composite_propagator_test.cc
lalitb: This will need change across all propagators. We can do it separately if needed, not in current scope of this PR.

PR #598, opened 2021-03-05
Timestamp

Muchene:
Fixes # (issue)
Zipkin api expects the timestamps in microseconds. 
## Changes
Casted timestamp into microseconds


lalitb:
Thanks for contributing the fix. 
  -  You would probably need to  format the file using bazel buildifier before check-in. You can use [format.sh](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/tools/format.sh) for it. 

 - Also seems `third_party/opentelemetry-proto` has been removed by mistake. Some issue in rebasing with main repo ?

ThomsonTan:
The CLA also needs to be fixed (no user Id for commit https://github.com/open-telemetry/opentelemetry-cpp/commit/345f60202715f89176fc93fa71c50238ba8296be?).

Muchene:
Yeah sorry for this, I was using a base that has some commits from the prometheus_exporter (which is where the issues are coming from). I can close this pull request and do it from a clean base

PR #603, opened 2021-03-09
Propagate trace_flags from parent span to child span

### LINKED ISSUE #602: TraceFlags is not propagated from parent span span to child span
When creating new `Span` with parent span context passed, we ignore the trace flags from the parent span context and always pass 0, see below code snippet. This blocks propagation of sampled status from parent span to child span.

https://github.com/open-telemetry/opentelemetry-cpp/blob/58a1e8c1ed712b3f47dc8f0c4660496e6702b077/sdk/src/trace/span.cc#L97

ThomsonTan:
Fixes #602 

## Changes

`trace_flags` is propagated from parent span to child span at span creation time.

### DIFF
--- a/sdk/src/trace/span.cc
+++ b/sdk/src/trace/span.cc
@@ -78,8 +78,9 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   recordable_->SetName(name);
 
   trace_api::TraceId trace_id;
-  trace_api::SpanId span_id = GenerateRandomSpanId();
-  bool is_parent_span_valid = false;
+  trace_api::SpanId span_id         = GenerateRandomSpanId();
+  trace_api::TraceFlags trace_flags = parent_span_context.trace_flags();
+  bool is_parent_span_valid         = false;
 
   if (parent_span_context.IsValid())
   {
@@ -94,7 +95,7 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   }
 
   span_context_ = std::unique_ptr<trace_api::SpanContext>(
-      new trace_api::SpanContext(trace_id, span_id, trace_api::TraceFlags(), false,
+      new trace_api::SpanContext(trace_id, span_id, trace_flags, false,
                                  is_parent_span_valid ? parent_span_context.trace_state()
                                                       : trace_api::TraceState::GetDefault()));
 
--- a/sdk/src/trace/span.cc
+++ b/sdk/src/trace/span.cc
@@ -64,7 +64,8 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
            const trace_api::SpanContextKeyValueIterable &links,
            const trace_api::StartSpanOptions &options,
            const trace_api::SpanContext &parent_span_context,
-           const opentelemetry::sdk::resource::Resource &resource) noexcept
+           const opentelemetry::sdk::resource::Resource &resource,
+           const bool sampled) noexcept
     : tracer_{std::move(tracer)},
       processor_{processor},
       recordable_{processor_->MakeRecordable()},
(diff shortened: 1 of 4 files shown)


sdk/src/trace/span.cc
```diff
@@ -78,8 +78,9 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   recordable_->SetName(name);
 
   trace_api::TraceId trace_id;
-  trace_api::SpanId span_id = GenerateRandomSpanId();
-  bool is_parent_span_valid = false;
+  trace_api::SpanId span_id         = GenerateRandomSpanId();
+  trace_api::TraceFlags trace_flags = parent_span_context.trace_flags();
```
ThomsonTan, line 82: I was thinking about propagating `trace_flags only for valid `SpanContext`, but since we could create a invalid `SpanContext` with non-zero `trace_flags`, so I choose to do propagation even for invalid span. Let me know if you prefer the other case.

ThomsonTan:
> Thanks Tom for the fix. Looks good to me...

Thanks @stevchin for trying opentelemetry-cpp, reporting this issue and helping on troubleshooting!

sdk/src/trace/span.cc
```diff
@@ -94,7 +95,7 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   }
 
   span_context_ = std::unique_ptr<trace_api::SpanContext>(
-      new trace_api::SpanContext(trace_id, span_id, trace_api::TraceFlags(), false,
+      new trace_api::SpanContext(trace_id, span_id, trace_flags, false,
```
lalitb, line 98: Not sure if I am missing something here - Why do we need to copy the parent's trace_flags to child span. For sampling-bit in child span trace_flag it should be determined as below - 
- Parent trace_flags MAY be used for sampling decision to decide whether child span should be sampled.
- In case the child span should be sampled ( as decided by above sampling algo ) - we should set the sampling-bit in child span trace_flag.

Or is there more information in parent's trace_flag ( other than sampling-bit) which we want to copy from parent ? If that is the case, we should copy that  first and then overwrite sampling bit based on decision from above logic.

Though our current implementation is also wrong - as it's assigning 0x00 to trace_flags even if it's is supposed to be sampled. This should be fixed.

ThomsonTan: Do you mean we set `kIsSampled` directly in the above code line because the location implies sampling should be enabled, instead of copying `trace_flags` from parent `SpanContext`? If so, I think we could keep the same sampled value as parent to be consistent and leave the decision to sampler, like even the IsSampled is off, some sampler may still sample it anyway. But I am fine to set the sampled bit here too.

lalitb: If it helps, how it is done in JS:
https://github.com/open-telemetry/opentelemetry-js/blob/71ba83a0dc51118e08e3148c788b81fe711003e7/packages/opentelemetry-tracing/src/Tracer.ts#L96

and python:

https://github.com/open-telemetry/opentelemetry-python/blob/fb1ae06026d648d867461396a6d1b5c37cbb3a31/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L924

lalitb: > Do you mean we set kIsSampled directly in the above code line because the location implies sampling should be enabled, instead of copying trace_flags from parent SpanContext?

Yes, as we are creating the span (and sampling decision happens before that in tracer), that means decision has already been made by sampler to sample it, and we should just be setting sampling-bit in the span's trace_flag irrespective of what is there in parent's trace_flag.

ThomsonTan: @lalitb thanks for the links. Set sampled based sampler decision makes sense. Updated the PR accordingly.

lalitb (approved):
LGTM. Thanks for fixing it.

jsuereth (approved):

PR #609, opened 2021-03-13
Add nuget packaging

maxgolov:
Add nuget packaging.

## Changes

This is preliminary implementation of nuget packaging scripts for OpenTelemetry C++ SDK. Certain organizations would prefer their entire build environment be isolated, with carefully vetted nuget packages and no access to public GitHub. The scripts here allow to prepackage OpenTelemetry source code in a NuGet package. Presently the NuGet package is source code only, but we can discuss further improvements to add prebuilt libraries to NuGet (in a separate PR).

There are no code changes. Included `tools/nuget/README.md` explains the process.

## Short illustration

```console
C:\work\opentelemetry-cpp.nuget\tools>build-nuget.cmd
Creating nuget OpenTelemetry.Cpp.Dev-0.2.0 ...
Attempting to build package from 'OpenTelemetry.Cpp.Dev.nuspec'.
Successfully created package 'C:\Users\maxgo\AppData\Local\Temp\42332e0f-0836-430a-95bd-b09c9ea53d65\OpenTelemetry.Cpp.Dev.0.2.0.nupkg'.
```

The package is also (temporarily) copied to `.\packages` directory in the source tree.

From there developers may push the package(s) to `nuget.org` feed, OR to their organization private feed.

NuGet specification file / attributes / image is resembling that one of OpenTelemetry.NET. Screenshot of the spec:

image

## Extensibility

The script is extensible. Simply dropping in another .nuspec file allows to create more packages tailored for specific needs. I have included the proposed (DRAFT) list of packages we MAY publish in future. This list is up for discussion and is subject to change.


tools/nuget/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Proposed NuGet package flavors
+
+This section is a `DRAFT` - Work-in-Progress.
+
+The following packages MAY be created and published at nuget.org :
+
+| Package Name                 | Desription                                                  | Notes |
+|------------------------------|-------------------------------------------------------------|-------|
+| OpenTelemetry.Cpp.Source     | Complete recursive source code snapshot with dependencies.  |       |
```
reyang, line 33: @cijothomas need to coordinate the package name here.

reyang: @maxgolov FYI https://www.nuget.org/profiles/OpenTelemetry

tools/nuget/OpenTelemetry.Cpp.Dev.nuspec
```diff
@@ -0,0 +1,24 @@
...
+    <id>OpenTelemetry.Cpp.Dev</id>
+    <version>0.0.0</version>
+    <authors>OpenTelemetry Authors</authors>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <license type="expression">Apache-2.0</license>
+    <licenseUrl>https://licenses.nuget.org/Apache-2.0</licenseUrl>
+    <icon>opentelemetry-icon-color.png</icon>
+    <projectUrl>https://opentelemetry.io/</projectUrl>
+    <description>OpenTelemetry C++ SDK - Development snapshot</description>
+    <tags>Observability OpenTelemetry Monitoring Telemetry Tracing native C++ v140 v141 winapp dyn rt-dyn msvcstl windesktop</tags>
```
reyang, line 12: winapp?

tools/nuget/README.md
```diff
@@ -0,0 +1,44 @@
...
+| OpenTelemetry.Cpp.Dev        | Source code snapshot excluding 3rd party dependencies.      |       |
+| OpenTelemetry.Cpp.Api        | API headers package.                                        |       |
+| OpenTelemetry.Cpp.Sdk        | SDK headers and source package.                             |       |
+| OpenTelemetry.Cpp.Ext        | Extended functionality.                                     |       |
+| OpenTelemetry.Cpp.Exporters  | Exporters.                                                  |       |
```
reyang, line 38: Do we want each exporter to have its individual package?

tools/nuget/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+The following packages MAY be created and published at nuget.org :
+
+| Package Name                 | Desription                                                  | Notes |
+|------------------------------|-------------------------------------------------------------|-------|
+| OpenTelemetry.Cpp.Source     | Complete recursive source code snapshot with dependencies.  |       |
+| OpenTelemetry.Cpp.Dev        | Source code snapshot excluding 3rd party dependencies.      |       |
+| OpenTelemetry.Cpp.Api        | API headers package.                                        |       |
+| OpenTelemetry.Cpp.Sdk        | SDK headers and source package.                             |       |
+| OpenTelemetry.Cpp.Ext        | Extended functionality.                                     |       |
```
reyang, line 37: Is this API extension or SDK extension or both?

tools/nuget/README.md
```diff
@@ -0,0 +1,44 @@
...
+| OpenTelemetry.Cpp.Exporters  | Exporters.                                                  |       |
+
+Prebuilt libraries MAY be provided in future in `OpenTelemetry.Cpp.${CompilerABI}` package.
+
+Refer to [Boost libary](https://www.nuget.org/packages?q=boost) nuget packages for example how to create
+a nuget package for a given compiler and runtime. [Fuget Package Explorer](https://www.fuget.org/) may
```
reyang, line 43: Fuget is the best! 😄

tools/nuget/OpenTelemetry.Cpp.Dev.nuspec
maxgolov: cpprest SDK is cross-plat and has this exact set of tags. I think we'd mainly use NuPkg for Windows builds, with Linux relying on `vcpkg` or Bazel.

tools/nuget/README.md
maxgolov: I do not have a strong preference. My assumption that the default we have in the main repo are all included in the `OpenTelemetry.Cpp.Exporters`.

tools/nuget/README.md
maxgolov: It's the contents of `ext` directory:
- zpages
- http client curl
- http server

These are optional components and should not be generally required for building SDK.

tools/nuget/README.md
maxgolov: Yes, I also use the [NuGet PackageExplorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) locally.

tools/nuget/README.md
maxgolov: My main concern here is that some C++ projects name their packages all lower-case, e.g. `boost` or `cpprestsdk` . Whereas I believe for OpenTelemetry, we should probably stick to common CamelCase naming convention. I isolated all C++ packages under `OpenTelemetry.Cpp.*`

tools/nuget/OpenTelemetry.Cpp.Dev.nuspec
maxgolov: ..but I am open to change this to whatever is more appropriate :)

tools/nuget/README.md
reyang: I think we have freedom here to call it `opentelemetry.cpp` or `otel.cpp` or whatever name that sounds native to C++ developers.

tools/nuget/OpenTelemetry.Cpp.Dev.nuspec
lalitb: These tags are for discoverability of these packages in nuget feed  - So say searching for `dyn` or `rt-dyn` would result in  otel-cpp packages in search result ?

tools/nuget/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Proposed NuGet package flavors
+
+This section is a `DRAFT` - Work-in-Progress. The following packages MAY be created and published at nuget.org :
+
+| Package Name                 | Desription                                                  | Notes |
+|------------------------------|-------------------------------------------------------------|-------|
+| OpenTelemetry.Cpp.Source     | Complete recursive source code snapshot with dependencies.  |       |
+| OpenTelemetry.Cpp.Dev        | Source code snapshot excluding 3rd party dependencies.      |       |
+| OpenTelemetry.Cpp.Api        | API headers package.                                        |       |
+| OpenTelemetry.Cpp.Sdk        | SDK headers and source package.                             |       |
```
lalitb, line 34: Would it make sense to add package dependencies in this table - say `OpenTelemetry.Cpp.Sdk` would also need `OpenTelemetry.Cpp.Api`, `OpenTelemetry.Cpp.Exporters` would need `OpenTelemetry.Cpp.Sdk`  and `OpenTelemetry.Cpp.Ext`.

tools/build-nuget.ps1
```diff
@@ -0,0 +1,73 @@
...
+  $parts = $Version.Split('.');
+  # Add missing tuples
+  $i = $parts.Count
+  while($i -lt 4)
+  {
+    $parts += "0"
+    $i++
+  }
+
+  ,$parts
```
ThomsonTan, line 22: What is this line for?

tools/nuget/README.md
```diff
@@ -0,0 +1,42 @@
...
+Learn more about native code NuGet packages [here](https://docs.microsoft.com/en-us/nuget/guides/native-packages).
+
+## Proposed NuGet package flavors
+
+This section is a `DRAFT` - Work-in-Progress. The following packages MAY be created and published at nuget.org :
+
+| Package Name                 | Desription                                                  | Notes |
+|------------------------------|-------------------------------------------------------------|-------|
+| OpenTelemetry.Cpp.Source     | Complete recursive source code snapshot with dependencies.  |       |
+| OpenTelemetry.Cpp.Dev        | Source code snapshot excluding 3rd party dependencies.      |       |
```
ThomsonTan, line 32: Do we also need create package for contrib repo? If yes, probably we could consolidate the scripts just here instead of creating separate set of scripts in the contrib repo.

tools/nuget/OpenTelemetry.Cpp.Dev.nuspec
maxgolov: I followed Visual Studio guidelines for creating native nuget packages. It refers to cpprest SDK as a good example of a nuget package. I copied the tags from there, combining the tags from OpenTelemetry C# SDK with the (applicable) tags from cpprest SDK, following the recommended best practices. My understanding is that NuGet tooling in Visual Studio IDE somehow parses those tags. The article that I followed is here:
https://docs.microsoft.com/en-us/nuget/guides/native-packages

tools/nuget/README.md
maxgolov: No, I would like to avoid that. I considered that and eventually it was my conscious decision to avoid that. Logical reasoning: dependency management is more predictable for C#. Less predictable for native code. Certain systems simply unpack the nuget (as zip) without dependency management at all. Thus, adding dependency management forces someone to manually add multiple packages, one-by-one.

Thus, I'd rather have us a complete source package:
- **Source**
- **Dev** - stripped-down version, i.e. installing 3rd party deps using alternate mechanism.

For **Sdk** vs **Api**, one can take the **Api** package and not take the **Sdk** package, and should still be able to build the SDK. Thus, adding dependencies to **Api** probably does not make sense.

The reverse-dependency ( **Sdk** depends on **Api** ), can be manually-managed (or at least in a system called CoreXT it would have to be manually managed0. We can discuss this automatic dependency management later on. I am not adding any spec at the moment for either of these packages. Also the other logical reasoning for manual dependency management is if you want to mix and match **Sdk-v3.1.0** with **Api-v3.2.0** , for example. If someone needs this managed automatically, they themselves can create their own metapackage. Maybe **OpenTelemetry.Cpp** should be the metapackage that automatically manages dependencies, but no individual mapping between sub-packages.

tools/build-nuget.ps1
maxgolov: It's a PowerShell syntax for returning an array back to the caller.

tools/nuget/README.md
maxgolov: Yes. We could use the same script against another directory with spec files. I can think about it in a separate PR, or when we have something in contrib that we would like to package.

tools/nuget/README.md
maxgolov: @cijothomas @reyang - I think if we use `OpenTelemetry` prefix, then it's gonna be easier to find the package. That way .NET developers already using the package would see that there's been a package created for native code as well, and spread the word.

tools/nuget/README.md
maxgolov: Maybe `OpenTelemetry.Cpp` package should be the _metapackage_ that aggregates all other packages. All standard exporters go to `*.Exporters` package, and all non-standard exporters go into their own package(s).

tools/nuget/README.md
maxgolov: To recap - maybe just one higher-order _metapackage_ `OpenTelemetry.Cpp` with no suffix, that aggregates all packages and only that one performs dependency management. But no dependency management between the children, i.e. one can still mix-and-match the compatible packages individually by force-installing `OpenTelemetry.Cpp.Sdk-vX` and `OpenTelemetry.Cpp.Api-vY`, **as long as developer knows what they are doing**.

lalitb (approved):
LGTM aa initial specs. We can evolve it further in future.

PR #610, opened 2021-03-15
Baggage API : refactor trace_state, and add Baggage implementation.

### LINKED ISSUE #79: Implement Baggage ( prev: CorrelationContext )
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/correlationcontext/api.md

lalitb:
Fixes #79 

## Changes

 - Refactor TraceState implementation
 - Model Baggage API close to TraceState, and both sharing same internal data structure of key-value list.
 
Converting to draft for some more refactoring, and add more tests.


api/include/opentelemetry/baggage/baggage.h
```diff
@@ -0,0 +1,66 @@
...
+            baggage->kv_properties_->AddEntry(e_key, e_value);
+          return true;
+        });
+    return baggage;
+  }
+
+  nostd::shared_ptr<Baggage> Set(const nostd::string_view &key, const nostd::string_view &value)
+  {
+    nostd::shared_ptr<Baggage> baggage(new Baggage(kv_properties_->Size() + 1));
+    baggage->kv_properties_->AddEntry(key, value);
```
nikhil1511, line 50: I have a question here about the key_value pairs in the baggage. With this implementation, we are adding duplicate keys in our baggage (eg. if key already existed in the baggage, we still prepend a new entry with the same key). Do we want to keep this behaviour or we should avoid adding old entry in our returned baggage.
Standard says to keep only one value for each baggage name ref:  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#overview
We are always returning fresh value in getter so in a way it does not seem to be a big issue. If we keep this behaviour same, I think we need to change couple of things in Delete function as size could be less than kv_properties_->Size() - 1 (line 37)

Let me know what do you think about this. This is also important because I was thinking if we could keep baggage entries as linked list instead of array to save space. Please correct me if I am missing anything.

lalitb: Good point, we should first add the k-v pair from current baggage except for key passed as argument ( if already exist). And in the end, add the new k-v. It should be fine to store one less element than actual allocation as long as we track number of elements internally.

nikhil1511: Yes, I think that should be the way. One more thing, I am planning to use linked list (deep clone everytime) in kv_properties for optimum memory usage (as I don't see usecase of any random access). Currently, context also uses linked list but its not deep clone as context does not have delete key api. So, I context code will be unchanged in this refactoring.
Do you agree with the approach ? Wanted to give you heads up before making change to make sure that I am not missing anything.

lalitb: What benefit do you see with linked list as compared to fixed size array ( which we use now )? Is it because of extra processing we may have to do initially to find exact allocation size in case of array? We want to minimize multiple dynamic memory allocation in api. And iterating a linked list would be slow as compared to array due to non-contiguous memory allocation of it's node.

lalitb: @pyohannes  - would like to hear your thoughts here.

nikhil1511: Yeah, I agree that linked list solution will be slow. I think, the main concern behind it is when we parse headers to extract the baggage, we do not know the size of key value pairs. So, solution is that either we parse header two times or maintain dynamic size data structure. I see that, in trace_state, you have made two passes.
We might hit invalid entries in between and so actual entries in baggage might be lesser. So, essentially first pass gives upper bound on number of entries.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -18,13 +18,16 @@
 #include <cstring>
 #include <string>
 
+#include <iostream>
```
seemk, line 21: Think this is unused?

api/include/opentelemetry/common/kv_properties.h
```diff
@@ -0,0 +1,148 @@
...
+#include "opentelemetry/version.h"
+
+#include <string>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace common
+{
+
+// Class to store fixed size array of key-value pairs of string type
+class KeyValueProperties
```
seemk, line 14: Is there reasoning to not use std library containers, e.g. `unordered_map` over `KeyValueProperties`?

lalitb: This is to guarantee ABI stability in api. We don't use std library
 - in api interface as arguments to virtual member functions (https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/abi-policy.md)
 - as member variables inside api classes.

seemk: Thanks, good to know! 🙂

api/include/opentelemetry/baggage/baggage.h
lalitb: Nikhil- That's true. just wondering if you have any further questions on this design ?

nikhil1511: Should I keep the array structure only and not the linked list ?

nikhil1511: I think, I can also clarify thing about baggage metadata here. We have not incorporated baggage metadata here. In my solution, I had kept baggage metadata along with key, value. Its an opaque structure for now so languages like golang are not keeping it separately (keeping it as a part of value). How should we proceed, I am kind of stuck there. We might want to make kv_properties more generic if we decide to keep it separately.

lalitb: Yes that's correct.

lalitb: > I think, I can also clarify thing about baggage metadata. We have not incorporated baggage metadata here. In my solution, I had kept baggage metadata along with key, value. Its an opaque structure with no semantic meaning for now so languages like golang are not keeping it separately (keeping it as a part of value). How should we proceed ? I am kind of stuck there. We might want to make kv_properties more generic if we decide to keep it separately.

I won't be much worried about incorporating metadata as of now, and probably revisit once it's semantics/usage is properly defined by W3C. Don't see languages like python, dotnet supporting it too.

lalitb:
closing for #676

Issue #564, opened 2021-02-05
Add installation instructions for CMake


ThomsonTan:
The install doc is empty which needs to be filled in with details.

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md

maxgolov:
I can add separate instructions for how to build using `CMake` via GUI - Visual Studio 2019 IDE.

lalitb:
#612 fixes this. @maxgolov - we can create separate ticket for build instruction using VS 2019.

PR #612, opened 2021-03-16
standalone cmake build instructions for opentelemetry-cpp

lalitb:
Fixes #

## Changes

Standalone cmake build instructions for opentelemetry-cpp.


INSTALL.md
```diff
@@ -1 +1,114 @@
...
+   -- Installing: /<install-root>/include/opentelemetry//ext/zpages/static/tracez_index.h
+   -- Installing: /<install-root>/include/opentelemetry//ext/zpages/static/tracez_style.h
+   -- Installing: /<install-root>/include/opentelemetry//ext/zpages/threadsafe_span_data.h
+   -- Installing: /<install-root>/lib/libopentelemetry_zpages.a
+   $
+   ```
+
+### Incorporating Into An Existing CMake Project
+
+TBD
```
seemk, line 110: Perhaps we can describe the same method as is used by nginx instrumentation:
```
find_package(opentelemetry-cpp REQUIRED)
find_package(gRPC REQUIRED)
target_include_directories(my_target PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS})
target_link_libraries(my_target PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES} gRPC::grpc++)
```

INSTALL.md
```diff
@@ -1 +1,114 @@
...
+   $
+   ```
+
+2. Navigate to the repository cloned above, and create the `CMake`
+build configuration.
+
+   ```console
+   $ cd opentelemetry-cpp
+   $ mkdir build && cd build
+   $ cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-Werror"
```
seemk, line 53: What do you think about adding a section to enable position independent code on supported compilers? otel-cpp does not enable it by default, might be useful for someone (would have been for me when I started using it) to add a note about enabling it, i.e. `cmake .. -DCMAKE_POSITION_INDEPENDENT_CODE=ON`

And perhaps a note about enabling OTLP: `-DWITH_OTLP=ON`

lalitb: Good comment. Have updated the document accordingly.

INSTALL.md
lalitb: I was thinking to keep it as part of #564 fix. But if it's already used in nginx, let me go though that once. Thanks for sharing the configuration here.

seemk (approved):

INSTALL.md
lalitb: It's done now.

INSTALL.md
```diff
@@ -1 +1,137 @@
...
+build systems for opentelemetry-cpp.
+
+## Build instructions using CMake
+
+### Prerequisites
+
+- A supported platform (eg Windows, macOS, Linux etc).
+Refer [Platforms Supported](./README.md#supported-development-platforms)
+for more information.
+- A compatible C++ compiler supporting atleast C++11.
```
ThomsonTan, line 13:
```suggestion
- A compatible C++ compiler supporting at least C++11.
```

INSTALL.md
```diff
@@ -1 +1,137 @@
...
+Major compilers are supported.
+Refer [Supported Compilers](./README.md#supported-c-versions) for more information.
```
ThomsonTan, line 15:
```suggestion
Refer to [Supported Compilers](./README.md#supported-c-versions) for more information.
```

INSTALL.md
```diff
@@ -1 +1,137 @@
...
+
+### Building as Standalone CMake Project
+
+1. Getting the opentelementry-cpp source:
+
+   ```console
+   # Change to the directory where you want to create the code repository
+   $ cd ~
+   $ mkdir source && cd source
+   $ git clone https://github.com/open-telemetry/opentelemetry-cpp
```
ThomsonTan, line 40: Add option `--recursive` to clone submodule as well?

INSTALL.md
```diff
@@ -1 +1,137 @@
...
+   $
+   ```
+
+2. Navigate to the repository cloned above, and create the `CMake`
+build configuration.
+
+   ```console
+   $ cd opentelemetry-cpp
+   $ mkdir build && cd build
+   $ cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-Werror"
```
ThomsonTan, line 53: No need to specify "Debug" and "-Werror" options? As there could be multi-configuration targets like for Visual Studio. We could rely on the default config for starter (I think it is still "Debug").

INSTALL.md
lalitb: submodules needn't be cloned for api/sdk build. It would be required for exporters, but better to have separate instructions for them in their respective folders.

INSTALL.md
lalitb: Ok, I was thinking about single-configurators like Makefile while writing this. But, DCMAKE_CXX_FLAGS should be valid for Visual Studio too right ? In that case we can keep it as:
`cmake .. -DCMAKE_CXX_FLAGS="-Werror"`

lalitb: ok have removed both -Werror and Debug now. `-Werror` too is specific to g++ compiler.

ThomsonTan (approved):

PR #616, opened 2021-03-19
Set default service.name if missing

lalitb:
Fixes # (issue)

## Changes

The `service.name` attribute of Resource is mandatory and SDK should set the default value if not specified. This attribute is necessary for certain backends to identify the service that produced the spans.
as in [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service):
* service.name
MUST be the same for all instances of horizontally scaled services. If the value was not specified, SDKs MUST fallback to unknown_service: concatenated with process.executable.name, e.g. `unknown_service:bash`. If process.executable.name is not available, the value MUST be set to `unknown_service`.

### DIFF
--- a/sdk/src/resource/resource.cc
+++ b/sdk/src/resource/resource.cc
@@ -9,9 +9,11 @@ namespace sdk
 namespace resource
 {
 
-const std::string kTelemetrySdkLanguage = "telemetry.sdk.language";
-const std::string kTelemetrySdkName     = "telemetry.sdk.name";
-const std::string kTelemetrySdkVersion  = "telemetry.sdk.version";
+const std::string kTelemetrySdkLanguage  = "telemetry.sdk.language";
+const std::string kTelemetrySdkName      = "telemetry.sdk.name";
+const std::string kTelemetrySdkVersion   = "telemetry.sdk.version";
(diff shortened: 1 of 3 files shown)


sdk/src/resource/resource.cc
```diff
@@ -25,15 +27,19 @@ Resource Resource::Merge(const Resource &other) noexcept
 Resource Resource::Create(const ResourceAttributes &attributes)
 {
   static auto otel_resource = OTELResourceDetector().Detect();
-  auto default_resource     = Resource::GetDefault();
+  auto resource = Resource::GetDefault().Merge(otel_resource).Merge(Resource(attributes));
```
ThomsonTan, line 30: Is `attributes.size()` still needs to be checked before creating `Resource` based on it?

lalitb: No, it should be fine to let merge empty resource and remove this check.

ThomsonTan (approved):

sdk/src/resource/resource.cc
lalitb: Added test to validate resource creation with empty attributes.

PR #626, opened 2021-03-24
Split Zpages webserver hosting from Exporter

jsuereth:
- Creates new shared_ptr thread-safe storage of spans between z-pages http server/aggregation and OTel's SDK.
- Update ZPages utilities to allow spinning up multiple span processors and attaching them to multiple tracer providers
- Update ZPages utilties so the server can be instantiated without overwriting the global tracer provider.


maxgolov (approved):

lalitb (approved):
Thanks for the changes. Nicely done :)

PR #628, opened 2021-03-25
ETW exporter how-to - usage instructions document

maxgolov:
Related to #326

## Documentation

This is a documentation that explains how to use OpenTelemetry C++ SDK with ETW exporter on Windows. The main idea is that application is instrumented using OpenTelemetry in a **platform-agnostic way** . And may subsequently utilize ETW exporter, if that application is running on Windows OS. I included a few examples that allow to utilize **vendor-agnostic** flow that is not tied to any Microsoft offering or service. This could be achieved in many various ways, but I am referring to [Microsoft.Azure.EventFow](https://github.com/Azure/diagnostics-eventflow) as an example of a `side-car` that may listen to events emitted over ETW channel. The other alternate project that may be utilized as a `side-car` for this purpose is [SilkETW](https://github.com/fireeye/SilkETW).

There are two operational modes in ETW exporter:

- **ETW / TraceLoggingDynamic mode** - so called `Dynamic Manifest`, already supported by any ETW Listener, best supported in C# / .NET. But it requires a header that we are working on releasing as opensource later this year. The closest analog to what it is doing is available as [GoLang implementation here](https://pkg.go.dev/github.com/microsoft/go-winio/pkg/etw).

- **ETW / MsgPack mode** - this one is entirely opensource at this point. And the usage example that I added in the document shows how it can be used today, with an out-of-proc listener that is capable of receiving ETW events, decoding their MsgPack-encoded payload to structured objects, then subsequently forwarding those objects to any destination of their choice.

I also added a mini `TODO` list in the **Addendum** section, to elaborate on further evolution plans, specifically on how OpenTelemetry C++ API calls in native code can be expressed as ETW events, that may subsequently be processed out-of-proc and forwarded to OpenTelemetry.NET, for example. That way preserving the uninterrupted flow of OpenTelemetry events further to OpenTelemetry-compatible destination, merely using ETW as a IPC channel for data passing on Windows. Because it is robust, efficient and already optimized for that purpose in the most optimal way imaginable.

There are no code changes, just the document. All code changes go in a separate PR here #519 (there are still a few moments that I'd like to finish in code before re-releasing the code for review).

Corresponding E2E example, with sample code, and ETW listener using EventFlow will be added separately to C++ contrib repo. We can discuss what prominent apps and services running on Windows (Docker, Apache, IIS) may be instrumented using that flow. Ideally I would envision the Docker replacing their [ETW Logging Driver](https://docs.docker.com/config/containers/logging/etwlogs/) with OpenTelemetry. We can also discuss how similar approach, concepts, and structure may be applied in the implementation of a similar exporter in GoLang for Windows using OpenTelemetry Go SDK and `go-winio` package (in a separate PR, and in a separate repo). But many concepts shown here in C++ may be universally applicable to other languages in general.


exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+int main(int argc, const char* argv[])
+{
+  // Obtain a Tracer object for instrumentation name.
+  // Each Tracer is associated with unique TraceId.
+  auto tracer = tp.GetTracer(providerName, "TLD");
+
+  // Properties is a helper class in ETW namespace that is otherwise compatible
+  // with Key-Value Iterable accepted by OpenTelemetry API. Using Properties
+  // should enable more efficient data transfer without unnecessary memcpy.
```
pyohannes, line 67: That's a bit misleading in the context of this example. As far as I can see, none of the usages of `Properties` here is more efficient than what the API provides out of the box. There's no `memcpy` involved in this case:
```c++
span->AddEvent(eventName, {
    {"uint32Key", (uint32_t)1234},
    {"uint64Key", (uint64_t)1234567890},
    {"strKey", "someValue"}
});
```

maxgolov: The answer - really depends. When we need to decorate the properties with additional attributes, such as - passing non-const properties, append some common properties... Maybe unique to custom `TracerProvider` ), for example - some sort of custom correlation technique that needs to be appended on every record. Let's say, some custom field that needs to be called `cV` and stamped on every `Properprties`. Then we could have reused the container that was created by the customer code once, then passing it in essence by reference on one thread (not requiring a transform to `Recordable` ), and not needing the key-value iterable for each key-value. That'd be more efficient than for each key-value copy, then transform to recordable, etc.

maxgolov: I'll leave this code reference as an example where accepting the collection, then passing this **original collection** around by reference, decorating it with additional contents, stripping unneded fields, etc. would be more efficient than trying to construct a new object (or a copy of an object) from initializer list. I think I understand your concern that I should've elaborated on exactly what scenarios would (in future) may benefit from passing in the original collection. This should work great for a header-only library, as we later on may completely bypass the copy/iteration/creation of a new object from initializer list, operating on original customer-code created object.

We mostly follow this pattern in this library elsewhere:
https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/api/Logger.cpp
And it's been working well.

maxgolov: Here is another scenario: let's say your `collection does not change` and contains something like `appName=x,status=Failed,scenario=FileOpen`. You'd prepare it once in a static initializer. Then you pass it as a reference to `Properties`, and never reconstruct it instead of constructing it from Key-Value Iterable every time. It appears like this would be more optimal (at least for a statically linked code where there is no passing across ABI boundary needed). But maybe I'm wrong - we need to take a look at both and run some benchmarks. I can remove the performance comment if it sounds wrong to you.

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+These options affect how "embedded in application" OpenTelemetry C++ SDK code is compiled:
+
+| Name        | Description                                                      |
+|---------------------|------------------------------------------------------------------------------------------------------------------------|
+| HAVE_CPP_STDLIB   | Use STL classes for API surface. This option requires at least C++17. C++20 is recommended.              |
+| HAVE_GSL      | Use [Microsoft GSL](https://github.com/microsoft/GSL) for `gsl::span` implementation. Library must be in include path. |
+| HAVE_ABSEIL_VARIANT | Use `absl::variant` instead of `std::variant`                                      |
+| HAVE_TLD      | Use ETW/TraceLogging Dynamic protocol. This is the default implementation.                       |
+| HAVE_CSTRING_TYPE   | Allow passing C-string type `const char *` to API calls.                                 |
+
```
lalitb, line 115: Nicely summarized the build options. Few of the options ( HAVE_ABSEIL_VARIANT, HAVE_CPP_STDLIB, HAVE_ABSEIL_VARIANT ) are applicable as build instructions for  `opentelemetry-cpp`, do you think we can move them there ( https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#building-as-standalone-cmake-project ), give it's reference and keep rest of them here?

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+exporter::ETW::TracerProvider tp;
+
+int main(int argc, const char* argv[])
+{
+  // Obtain a Tracer object for instrumentation name.
+  // Each Tracer is associated with unique TraceId.
+  auto tracer = tp.GetTracer(providerName, "TLD");
+
+  // Properties is a helper class in ETW namespace that is otherwise compatible
```
lalitb, line 65: Sorry if I am missing something, but where is this helper class `Properties` defined in ETW namespace ? Had a quick look to the code, but didn't find it there ?

exporters/etw/README.md
maxgolov: Yes, but I want to have a separate document created that covers ALL options. Maybe a separate section in `INSTALL.md` as you are suggesting. For this one - I tried to keep it scoped to what is relevant in this concrete usage scenario.

exporters/etw/README.md
maxgolov: Sorry, I might have missed the alias. I'll refresh it. It's presently defined here, in a PR which contains the code. I wanted to scope this PR to documentation only. This is the code:

https://github.com/open-telemetry/opentelemetry-cpp/blob/48991c05032cbac31e29c700261761c01598aa8c/exporters/etw/test/etw_tracer_test.cc#L26

```cpp
using Properties       = opentelemetry::exporter::ETW::Properties;
using PropertyValue    = opentelemetry::exporter::ETW::PropertyValue;
using PropertyValueMap = opentelemetry::exporter::ETW::PropertyValueMap;
```

Going forward I want to benchmark the pros and cons of passing a container vs. initializer list, the topic we discussed with Johannes above. In general, the concept is not necessarily unique to ETW. It could be applied to other exporters. Maybe I'd propose a refactor of pulling it in a separate (optional) helper class on API surface. Users may still use the existing initializer lists as well, as this does not break the "API". There is a moment where we have to be careful about "ABI". But since the implementation is header-only and linked straight into the app, with the matching runtime, the "ABI" compatibility issue is not applicable in this scenario. I can elaborate on this offline.

maxgolov: I also want to make sure that even if we pass a map, if there is a need to involve a transform to API call across the ABI boundary, then the container itself is `KeyValueIterable`-interface compatible. Means the SDK code may iterate over that one in ABI safe manner, and perform a copy of all key-value pairs to the actual container beyond the ABI boundary. This will only be needed for other (non-ETW) exporters, and ETW exporter is lucky in this regard. As it's currently done fully as header-only, slim, very few hops to the actual "Event Write" routine, with no shared library.

exporters/etw/README.md
lalitb: Can we add the specific example/scenario as an example in this doc or in unit-test ( and refer it here ) to justify the performance statement. As the current api/sdk implementation, the only place where `memcpy` MAY happen is during Recordable::AddEvent() execution, but that would depend how exporter implements this logic.

maxgolov: > the only place where memcpy MAY happen is during Recordable::AddEvent() execution

Well, that is the most commonly invoked function, actually.

maxgolov: I'm pushing an update to my code PR to explain this further.

lalitb (approved):
LGTM - perhaps we can add scenario for no-memcpy as that would be helpful.

maxgolov:
@pyohannes - I will add a separate benchmark to measure, and try to supplement the document (in a separate PR) with concrete examples where `initializer list`-supplied data is better, and where prepared `container (map)` is better.

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+## API support
+
+These are the features planned to be supported by ETW exporter:
+
+- [x] OpenTelemetry Tracing API and SDK headers are **stable** and moving towards GA.
+- [ ] OpenTelemetry Logging API is work-in-progress, pending implementation of [Latest Logging API spec here](https://github.com/open-telemetry/oteps/pull/150)
+- [ ] OpenTelemetry Metrics API is not implemented yet.
+
+Implementation of OpenTelemetry C++ SDK ETW exporter on Windows OS is `header only` :
```
ThomsonTan, line 20: nit:
```suggestion
Implementation of OpenTelemetry C++ SDK ETW exporter on Windows OS is `header only`:
```

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+All that is required is to point the compiler at the location of the headers, and then `#include` the header
+files into the application source. Compiler's optimizer can do a much better job when all the library's
+source code is available. Several options below may be turned on to optimize the code with the usage of
+standard C++ library, Microsoft Guidelines Support library, Google Abseil Variant library. Or enabling
+support for non-standard features, such as 8-bit byte arrays support that enables performance-efficient
+representation of binary blobs on ETW wire.
+
+## Example project
+
+The following include directories are required, relative to the top-level source tree of OpenTelemetry C++ SDK:
```
ThomsonTan, line 35: Using "OpenTelemetry C++ SDK" may confuse the audience as we have API and SDK package. Perhaps just way "top-level source tree of OpenTelemetry C++ repo"?

exporters/etw/README.md
ThomsonTan: Besides the description for the options, could we add an extra column to the the pros and even cons of using each option to help users to make decision?

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+OpenTelemetry ETW exporter optionally allows to encode the incoming event payload using
+[MessagePack](https://msgpack.org/index.html) compact binary protocol. ETW/MsgPack encoding
+requires [nlohmann/json](https://github.com/nlohmann/json) library to be included in the build
+of OpenTelemetry ETW exporter. Any recent version of `nlohmann/json` is compatible with ETW
+exporter. For example, the version included in `third_party/nlohmann-json` directory may be
+used.
+
+There is currently **no built-in decoder available** for this format. However, there is ongoing
+effort to include the ETW/MsgPack encoder in [Azure/diagnostics-eventflow](https://github.com/Azure/diagnostics-eventflow)
```
ThomsonTan, line 164: Is it "include the ETW/MsgPack decoder"?

exporters/etw/README.md
```diff
@@ -0,0 +1,201 @@
...
+- api/include/
+- exporters/etw/include/
+- sdk/include/
+
+Code that instantiates ETW TracerProvider, subsequently obtaining a Tracer bound to `OpenTelemetry-ETW-Provider`,
+and emitting a span named `MySpan` with attributes on it, as well as `MyEvent` within that span.
+
+```cpp
+
+#include <map>
```
ThomsonTan, line 46: Could the sample code be separated into .cc source file with build scripts support, and insert the link here? In that way we make sure the sample code builds fine and is up-to-date.

maxgolov: This is intentional. We already have about this same code in a test app..
So I thought it'd be nice to have it embedded into markdown document.

exporters/etw/README.md
maxgolov: My bad, yes :)

exporters/etw/README.md
maxgolov: Do we need space before colon in text, I thought no.. `In modern English-language printing, no space is placed before a colon and a single space is placed after it.`.. Since this is end-of-line, there's no space, but end-of-line instead.

exporters/etw/README.md
maxgolov: Sounds good.

exporters/etw/README.md
maxgolov: I like the idea of elaborating on what this option is doing. But maybe in that same description column.

exporters/etw/README.md
maxgolov: Fixed in next iteration.

exporters/etw/README.md
maxgolov: I added a few more sentences. `HAVE_CPP_STDLIB` has its own separate detailed design doc here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/building-with-stdlib.md

exporters/etw/README.md
maxgolov: I'm marking it as `Resolved` / `Won't Fix` since this proposal is not gonna be consistent with modern English-language printing rules :)

ThomsonTan (approved):

PR #630, opened 2021-03-26
Use gmock full path for linking in CMake script

ThomsonTan:
gmock.lib is specified to otlp exporter for linking which cannot be found by linker if it is not in system path. Actually the full path returned by `find_library()` should be used.


lalitb (approved):
Thanks for the fix.

PR #632, opened 2021-03-26
 Add http client/server example

### LINKED ISSUE #631: Add simple http client/server example
As one of the item in beta tracker milestone ( https://github.com/open-telemetry/opentelemetry-cpp/issues/80 ), add simple HTTP client / server example.

lalitb:
Fixes #631 

## Changes
Adds simple client/server example using http client and server library as provided by this repo.

It's very simple functionality for now to create and dump spans at both the end. Next step would be to have nested spans and w3c context propagation.

Also small formatting fix in ostream exporter to dump attributes properly. Currently on xterm, contents get wrapped on same line and not readable properly for larger attributes.


exporters/ostream/test/ostream_span_test.cc
```diff
@@ -145,7 +145,7 @@ TEST(OStreamSpanExporter, PrintChangedSpanCout)
       "  description   : Test Description\n"
       "  span kind     : Client\n"
       "  status        : Ok\n"
-      "  attributes    : attr1: string\n"
+      "  attributes    : \n\tattr1: string\n"
```
TomRoSystems, line 148: Keep separate lines separate?
```suggestion
      "  attributes    : \n\tattr1: string\n"
      "\tattr1: string\n"
```

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
```diff
@@ -110,11 +110,15 @@ class OStreamSpanExporter final : public sdktrace::SpanExporter
     size_t i    = 1;
     for (auto kv : map)
     {
-      sout_ << kv.first << ": ";
+      if (i == 1)
+      {
+        sout_ << std::endl;
+      }
```
TomRoSystems, line 116: I think we could move this out of the loop and get rid of `i` variable completely with small change in `Export` function.

examples/http/server.hpp
```diff
@@ -0,0 +1,49 @@
...
+        server_.setKeepalive(false);
+    }
+
+    void AddHandler(std::string path, HTTP_SERVER_NS::HttpRequestCallback *request_handler){
+        server_.addHandler(path, *request_handler);
+    }
+
+    void Start() {
+        if (!is_running_) {
+            server_.addListeningPort(port_);
```
TomRoSystems, line 31: It's probably not in the scope of this PR but shouldn't we be more careful about error result for this operation? Like port already in use or you don't have permissions for listening below 1024.

examples/http/client.cc
```diff
@@ -0,0 +1,61 @@
+#include "opentelemetry/ext/http/client/http_client_factory.h"
+#include "tracer_common.hpp"
+
+namespace
+{
+
+void sendRequest(std::string url)
```
TomRoSystems, line 7: In this and similar places when passing string args - const& ?

TomRoSystems (commented):
Small suggestions.

examples/http/client.cc
```diff
@@ -0,0 +1,61 @@
...
+{
+  auto http_client = opentelemetry::ext::http::client::HttpClientFactory::CreateSync();
+
+  // start active span
+  opentelemetry::trace::StartSpanOptions options;
+  options.kind = opentelemetry::trace::SpanKind::kClient;  // server
```
ThomsonTan, line 13:
```suggestion
  options.kind = opentelemetry::trace::SpanKind::kClient;  // client
```

examples/http/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+  target_link_libraries(
+    http_client
+    ${CMAKE_THREAD_LIBS_INIT}
+    ${CORE_RUNTIME_LIBS}
+    opentelemetry_trace
+    http_client_curl
+    opentelemetry_exporter_ostream_span
+    ${CURL_LIBRARIES})
+
+  target_link_libraries(
```
ThomsonTan, line 21: Make the above 2 `target_link_libraries` format consistent?

examples/http/README.md
```diff
@@ -0,0 +1,86 @@
...
+   ```
+
+   * Server console
+
+   ```console
+   {
+     name          : handle_request
+     trace_id      : f92ff8f5f6553ead210f2643779e1587
+     span_id       : 195fae1770f514ce
+     parent_span_id: 0000000000000000
```
ThomsonTan, line 73: Seems the client context is not propagated to the server from the above dump?

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
```diff
@@ -110,11 +110,15 @@ class OStreamSpanExporter final : public sdktrace::SpanExporter
     size_t i    = 1;
     for (auto kv : map)
     {
-      sout_ << kv.first << ": ";
+      if (i == 1)
```
ThomsonTan, line 113: Starts `i` from `1` seems not intuitive, perhaps make it starts from `0`?

ext/include/opentelemetry/ext/http/server/socket_tools.h
```diff
@@ -14,6 +14,7 @@
 #pragma once
 
 #include <algorithm>
+#include <atomic>
```
ThomsonTan, line 17: Is this necessary?

lalitb: Yes, there is atomic variable inside, and compilation fails without that. This used to get silently passed earlier due to other files including this header.

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
lalitb: Yes, I did realize this function needs revisit, will fix it.

examples/http/README.md
lalitb: That part is missing for now, as I mentioned in PR description. Will add it in next PR.

>It's very simple functionality for now to create and dump spans at both the end. Next step would be to have nested spans and w3c context propagation.

examples/http/client.cc
lalitb: Thanks :)

ThomsonTan (approved):

examples/http/README.md
```diff
@@ -0,0 +1,86 @@
...
+     duration      : 20600
+     description   :
+     span kind     : Server
+     status        : Unset
+     attributes    :
+           http.req.headerHost: localhost:8800
+           http.req.headerAccept: */*
+           component: http
+           req_method: GET
+           req_url: /helloworld
```
pyohannes, line 84: When giving examples, I think we should stick as close as possible to the [semantic conventions for HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions).

For example:
* Naming the span after the route.
* `req_url` should be `http.route`.
* `http.req.headerHost` should be `http.host`.

There are other inconsistencies. also for the client span.

examples/http/server.cc
```diff
@@ -0,0 +1,71 @@
...
+                options);
+    auto scope = get_tracer("http_server")->WithActiveSpan(span);
+    for (auto &kv : request.headers)
+    {
+      span->SetAttribute("http.req.header" + std::string(kv.first.data()), kv.second);
+    }
+    if (request.uri == "/helloworld")
+    {
+      span->AddEvent("Processing request");
+      response.headers["Content-Type"] = "text/plain";
```
maxgolov, line 32: nit: it'd be good to add a few standard mime types as constants in the HTTP server header.

examples/http/server.hpp
```diff
@@ -0,0 +1,49 @@
...
+            server_.start();
+            is_running_ = true;
```
maxgolov, line 33: This is fairly common issue. I'm sharing it here since I got burned by similar mistake in production before. Since your `is_running_` is atomic, the best practice is to use `if(!is_running.exchange(true)` for start, and the reverse swap `if(is_running.exchange(false)` for stop. Otherwise your check at line 30 and assignment at line 33 is **not atomic** .

examples/http/server.hpp
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    void Stop() {
+        if (is_running_){
```
maxgolov, line 38: Same issue. Would be best to use [exchange](https://en.cppreference.com/w/cpp/atomic/atomic/exchange)

maxgolov (approved):
I'm generally OK with this, but it would be best if you fix the atomic check-run-assign, because someone else may look at this code, and spread the erroneous implementation elsewhere.

exporters/ostream/test/ostream_span_test.cc
lalitb: Thanks, fixed now.

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
lalitb: Good point. Get rid of this variable now.

examples/http/server.hpp
lalitb: Correct. The original purpose of this http server library was only for testing purpose, and so error handling was not there. We can revisit this if it's used at more places.

examples/http/CMakeLists.txt
lalitb: Not sure how, but this is what cmake-format seems to be generating :)

exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
lalitb: thanks, have removed the use of this index variable.

examples/http/README.md
lalitb: Good point. Have fixed both client and server now.

examples/http/server.cc
lalitb: Thanks. We already have few mime types already there in http server lib, have used them now.

examples/http/server.hpp
lalitb: Good point. Was not thinking of scenario of multiple threads trying to start the server. Have fixed this now.

examples/http/server.hpp
lalitb: fixed this.

PR #633, opened 2021-03-26
Fix GMock linkage error on Windows

maxgolov:
@ThomsonTan  - There's a minor regression (build break on Windows) introduced by #630 for Windows-CMake builds:

image

I found this while rebasing my other PR to latest. We should probably add Windows CMake builds with OTLP testing enabled to the loop (maybe in another PR, this one is just to fix the build break).

## Changes

- vcpkg-installed GMock on Windows for **Debug** build lives in `gmockd.dll` , so searching for `gmock` is morally wrong... 😄 This is idiosyncrasy of Google Test library itself, that it names the two flavors differently and does not provide an easy-to-use CMake config for that.
 
- vcpkg-installed GMock on Windows requires `#define GTEST_LINKED_AS_SHARED_LIBRARY 1`

### DIFF
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -33,7 +33,14 @@ if(BUILD_TESTING)
     TARGET recordable_test
     TEST_PREFIX exporter.otlp.
     TEST_LIST recordable_test)
-  find_library(GMOCK_LIB gmock PATH_SUFFIXES lib)
+  if(MSVC)
+    add_definitions(-DGTEST_LINKED_AS_SHARED_LIBRARY=1)
+  endif()
+  if(MSVC AND CMAKE_BUILD_TYPE STREQUAL "Debug")
+    find_library(GMOCK_LIB gmockd PATH_SUFFIXES lib)
+  else()
+    find_library(GMOCK_LIB gmock PATH_SUFFIXES lib)
+  endif()
   if(GMOCK_LIB)
     add_executable(otlp_exporter_test test/otlp_exporter_test.cc)
     target_link_libraries(


maxgolov:
I am not 100% sure if `gmockd` is also to be used for Debug builds on Linux (surely it should be used on Windows)

exporters/otlp/CMakeLists.txt
```diff
@@ -33,7 +33,14 @@ if(BUILD_TESTING)
     TARGET recordable_test
     TEST_PREFIX exporter.otlp.
     TEST_LIST recordable_test)
-  find_library(GMOCK_LIB gmock PATH_SUFFIXES lib)
+  if(MSVC)
+    add_definitions(-DGTEST_LINKED_AS_SHARED_LIBRARY=1)
```
ThomsonTan, line 37: Perhaps `target_compile_definitions` to make the definition target specific?

maxgolov: If we add more tests, unit tests, or benchmarks, then this definition would apply to all other projects in this directory and subdirectories, so then we won't have to do  `target_compile_definition` every time. In that sense `add_definitions` is a bit more universal.

maxgolov: I am thinking `MAYBE` we should even pull it at the top-most level as a build option - that detects whether Google Test and Google Benchmark and Google Mock are installed as DLLs, but I don't want to touch it right now. 😄

exporters/otlp/CMakeLists.txt
```diff
@@ -33,7 +33,14 @@ if(BUILD_TESTING)
...
+  endif()
+  if(MSVC AND CMAKE_BUILD_TYPE STREQUAL "Debug")
+    find_library(GMOCK_LIB gmockd PATH_SUFFIXES lib)
+  else()
+    find_library(GMOCK_LIB gmock PATH_SUFFIXES lib)
+  endif()
   if(GMOCK_LIB)
```
maxgolov, line 44: I also think that this check `if(GMOCK_LIB)` - probably needs to be removed entirely. If we are building with tests enabled, we should `require` that GMock is available. If it's not there, we should fail building `otlp_exporter_test` instead of silently skipping it.

ThomsonTan (approved):

PR #635, opened 2021-03-27
Enable tls authentication for otlp grpc exporter

### LINKED ISSUE #629: Enable tls authentication for otlp grpc exporter
As of now, otlp-grpc exporter doesn't support tls authentication :

https://github.com/open-telemetry/opentelemetry-cpp/blob/9aae3e64bdc71520fd3b737bc054ad8c604aead5/exporters/otlp/src/otlp_exporter.cc#L36-L41

This needs to be supported. The certificate should be made available to exporter programmatically and (optionally) through environment variable:

OTEL_EXPORTER_OTLP_CERTIFICATE 
OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE 
OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE

Reference -
 [specs](
https://github.com/open-telemetry/opentelemetry-specification/blob/f62744a679814937214fd17394ab3fa8a9099424/specification/protocol/exporter.md)

padmaja-nadkarni:
Fixes #629  Enable tls authentication for otlp grpc exporter

## Changes

Add  support for SslCredentials by adding capability to specify the path to cacert pem file via OtlpExporterOptions. Also made changes to the examples/otlp/main.cc and updated the README for that example.


exporters/otlp/src/otlp_exporter.cc
```diff
@@ -30,13 +31,28 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>>
...
+
 /**
  * Create service stub to communicate with the OpenTelemetry Collector.
  */
 std::unique_ptr<proto::collector::trace::v1::TraceService::Stub> MakeServiceStub(
-    std::string endpoint)
+    const OtlpExporterOptions &options)
 {
-  auto channel = grpc::CreateChannel(endpoint, grpc::InsecureChannelCredentials());
+  auto channel = grpc::CreateChannel(options.endpoint, grpc::InsecureChannelCredentials());
```
lalitb, line 49: should we be creating this insecure-channel only when `use_ssl_credentials` is false ?

padmaja-nadkarni: I thought since use_ssl_credentials is false by default, behavior is as original code, please let me know if you still would like it to be in else clause instead.

lalitb: Does it mean that in case of errors like cert file missing or not-readable, we should allow using insecure channel even though the `use_ssl_credentials` is true ?

examples/otlp/README.md
```diff
@@ -10,6 +10,11 @@ SDK](https://github.com/open-telemetry/opentelemetry-cpp). The application then
 calls a `foo_library` which has been instrumented using the [OpenTelemetry
 API](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/api).
 
+To enable tls authentication for OTLP grpc exporter, SslCredentials can be used by
```
ThomsonTan, line 13:
```suggestion
To enable TLS authentication for OTLP grpc exporter, SslCredentials can be used by
```

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_exporter.h
```diff
@@ -15,6 +15,11 @@ struct OtlpExporterOptions
 {
   // The endpoint to export to. By default the OpenTelemetry Collector's default endpoint.
   std::string endpoint = "localhost:4317";
+  // By default when false, uses grpc::InsecureChannelCredentials(); If true,
+  // uses ssl_credentials_cacert_path
+  bool use_ssl_credentials = false;
```
ThomsonTan, line 20: Is it possible to add a test case for `use_ssl_credentials = true`?

exporters/otlp/src/otlp_exporter.cc
padmaja-nadkarni: oh I get it now, will address.

examples/otlp/README.md
```diff
@@ -10,6 +10,11 @@ SDK](https://github.com/open-telemetry/opentelemetry-cpp). The application then
...
+specifying the path to cacert pem file via OtlpExporterOptions. The path to such a
```
lalitb, line 14:
```suggestion
specifying the path to client certificate pem file via OtlpExporterOptions. The path to such a
```

lalitb: Correct me - but ideally this would be client certificate file - containing client cer, client key and CA certificate chain, In that case, we can just say `client certificate` instead of `cacert` ?

maxgolov (commented):
I think, in general, it's a good change as a starting point. At least, appears like it may work for Linux containers... In future, it would _have been nicer_ to have programmatic access / API for obtaining certificate information in a platform-agnostic way. Other OS, e.g. **Windows** and **Mac** would generally manage certificates in a trusted key store. And cloud apps would generally manage certificates using secure APIs, such as [Azure KeyVault](https://docs.microsoft.com/en-us/rest/api/keyvault/), for example... It appears to me that simplifying this to only allow the cert read from a filesystem file is oversimplification versus what a real-world production system gonna eventually need.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_exporter.h
```diff
@@ -15,6 +15,11 @@ struct OtlpExporterOptions
...
+  // ssl_credentials_cacert_path specifies path to .pem file to be used for SSL encryption.
+  std::string ssl_credentials_cacert_path = "";
```
pyohannes, line 22: Could we add an option to add the certificates as strings, and not as a path to a file containing the certificates?

That would make us more flexible, as there are many instances where certificates aren't available on the file system but only in memory.

padmaja-nadkarni: Would you like it as additional option to file system path or you would like to get rid of file system path? If it is the latter, are all reviewers okay with the change?

lalitb: I think this is good idea to read cert from variable instead of file, that way exporter don't need to deal with file related error scenarios. Let application read the cert from file correctly, and inject it to exporter.

padmaja-nadkarni: In latest commit, added one more option for certificate as a string however, let me know if file option needs to be removed - for our current application needs, we were using file system paths and it is working correctly.

exporters/otlp/src/otlp_exporter.cc
padmaja-nadkarni: addressed.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_exporter.h
padmaja-nadkarni: added a test for config.

lalitb (approved):

ThomsonTan (approved):

examples/otlp/README.md
```diff
@@ -10,6 +10,11 @@ SDK](https://github.com/open-telemetry/opentelemetry-cpp). The application then
 calls a `foo_library` which has been instrumented using the [OpenTelemetry
 API](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/api).
 
+To enable TLS authentication for OTLP grpc exporter, SslCredentials can be used by
+specifying the path to client certificate pem file via OtlpExporterOptions. The path to such a
```
lalitb, line 14:
```suggestion
specifying the path to client certificate pem file, or the string containing this certificate via OtlpExporterOptions. The path to such a
```

lalitb: Sorry for one last comment :)

padmaja-nadkarni: no worries, makes sense, forgot to update.

PR #636, opened 2021-03-29
Add opentelemetry.io docs

austinlparker:
Per open-telemetry/opentelemetry.io#472, we're mirroring the docs content on the website to each SIG. When a release occurs and these docs are updated, please make an issue or PR mirroring them to their appropriate location in the website repo (https://github.com/open-telemetry/opentelemetry.io/tree/main/content/en/docs/cpp).

### DIFF
--- a/website_docs/_index.md
+++ b/website_docs/_index.md
@@ -1,0 +1,28 @@
+---
+title: "C++"
+weight: 11
+description: >
+  A language-specific implementation of OpenTelemetry in C++.
+---
+
+This is the OpenTelemetry for C++ documentation. OpenTelemetry is an
+observability framework -- an API, SDK, and tools that are designed to aid in
+the generation and collection of application telemetry data such as metrics,
+logs, and traces. This documentation is designed to help you understand how to
+get started using OpenTelemetry for C++.
+
+## Status and Releases
+
+The current status of the major functional components for OpenTelemetry C++ is
+as follows:
+
+| Tracing   | Metrics   | Logging             |
+| -------   | -------   | -------             |
+| Pre-Alpha | Pre-Alpha | Not Yet Implemented |
+
+The current release can be found [here](https://github.com/open-telemetry/opentelemetry-cpp/releases)
+
+## Further Reading
+
+- [OpenTelemetry for C++ on GitHub](https://github.com/open-telemetry/opentelemetry-cpp)
+- [Examples](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples)


website_docs/_index.md
```diff
@@ -0,0 +1,28 @@
...
+get started using OpenTelemetry for C++.
+
+## Status and Releases
+
+The current status of the major functional components for OpenTelemetry C++ is
+as follows:
+
+| Tracing   | Metrics   | Logging             |
+| -------   | -------   | -------             |
+| Pre-Alpha | Pre-Alpha | Not Yet Implemented |
```
lalitb, line 21: nit - we do have logging api/sdk pre-alpha release.

website_docs/_index.md
```diff
@@ -0,0 +1,28 @@
...
+
+The current release can be found [here](https://github.com/open-telemetry/opentelemetry-cpp/releases)
+
+## Further Reading
+
+- [OpenTelemetry for C++ on GitHub](https://github.com/open-telemetry/opentelemetry-cpp)
+- [Examples](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples)
```
lalitb, line 28: API docs: https://opentelemetry-c-api.readthedocs.io/en/latest/index.html
Contrib repo : https://github.com/open-telemetry/opentelemetry-cpp-contrib

reyang:
@austinlparker would it be okay to put the @open-telemetry/cpp-maintainers as the "CODEOWNER" of this folder https://github.com/open-telemetry/opentelemetry.io/tree/main/content/en/docs/cpp? In this way we don't have to duplicate the contents in this repo.

I can update the [OTel C++ release process](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/RELEASING.md) and put something like "for each release, update the opentelemetry.io docs".

website_docs/_index.md
ThomsonTan: And I think Metrics should be "Not Yet Implemented".

reyang:
I've submitted a PR https://github.com/open-telemetry/opentelemetry.io/pull/505. Once it is merged, I'll update the release process doc https://github.com/open-telemetry/opentelemetry-cpp/blob/main/RELEASING.md then we can close this PR.

reyang (changes requested):
I wish to explore if we could avoid duplicating docs. Mirroring the docs will create problem since we don't have the same CI rule between the OTel C++ and opentelemetry.io repo.

austinlparker:
I'm trying to avoid mirroring by making this be the authoritative source - my goal is to have a github action that will push these changes to the website each release.

austinlparker:
Unfortunately this is one of those things where everyone wants to do this slightly differently :|

reyang:
> I'm trying to avoid mirroring by making this be the authoritative source - my goal is to have a github action that will push these changes to the website each release.

This would be great!

reyang:
@austinlparker I'm closing this PR as a result of #639.
If there is a way to automatically publish doc from this repo to opentelemetry.io (so we can keep all C++ related docs here), please let us know and we're happy to do it.

PR #638, opened 2021-03-29
Refactoring trace_state to reuse common functionality in baggage

### LINKED ISSUE #79: Implement Baggage ( prev: CorrelationContext )

nikhil1511:
Related to #79 

- Refactoring trace_state code to reuse functionality in baggage implementation
- Extracting out common parsing logic from trace_state into kv_properties


api/include/opentelemetry/common/kv_properties.h
```diff
@@ -0,0 +1,239 @@
...
+      else
+      {
+        end--;
+      }
+
+      auto list_member = StringUtil::Trim(str_, index_, end);
+      if (list_member.size() == 0)
+      {
+        // empty list member. Move to next entry. For both baggage and trace_state this is valid
+        // behaviour.
```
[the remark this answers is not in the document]
nikhil1511: Makes sense, I will update the comment. About the flag, I am not sure if we will need this parsing at other places, so thought of not making it too generic.
If we go ahead with the flag approach, do you think we should have that as part of tokenizer's constructor (I feel flag will be same for all next calls) ? In that case we can probably create TokenizerConfigOption struct to be passed to the constructor.

api/include/opentelemetry/common/string_util.h
```diff
@@ -0,0 +1,42 @@
+// Copyright 2021, OpenTelemetry Authors
```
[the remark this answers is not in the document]
nikhil1511: Trim function is also used by baggage keys and values. So, kv_tokenizer trims the whitespaces in pair and after getting key and value pairs, we trim them as well.
So, I felt its better to keep it at generic place.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -122,60 +66,35 @@ class TraceState
...
-    nostd::shared_ptr<TraceState> ts{new TraceState()};
 
-    std::size_t begin{0};
-    std::size_t end{0};
-    while (begin < header.size() && ts->num_entries_ < kMaxKeyValuePairs)
+    size_t cnt = 0;
+    common::KeyValueStringIterator kv_str_itr(header);
+    bool kv_valid;
+    nostd::string_view key, value;
+    while (kv_str_itr.next(kv_valid, key, value) && cnt < kMaxKeyValuePairs)
```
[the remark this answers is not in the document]
nikhil1511: I thought about it, but felt that we are probably doing only one extra find for '=' (which will be of less time complexity than finding next ',').
I don't have very strong opinion here, so either way is fine.

api/test/common/BUILD
```diff
@@ -5,3 +5,25 @@ otel_cc_benchmark(
     srcs = ["spinlock_benchmark.cc"],
     deps = ["//api"],
 )
+
+cc_test(
```
ThomsonTan, line 9: Adding the tests to CMake build script as well?

api/include/opentelemetry/common/kv_properties.h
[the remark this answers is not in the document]
lalitb: Yeah, I actually meant this flag for constructor.

api/include/opentelemetry/trace/trace_state.h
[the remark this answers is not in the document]
lalitb: Ok, Just felt that multiple function calls for `next()` can have performance impact, but probably compiler level optimization may take care of that.

api/include/opentelemetry/common/kv_properties.h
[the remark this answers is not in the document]
nikhil1511: Added flag

api/include/opentelemetry/trace/trace_state.h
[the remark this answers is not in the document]
nikhil1511: Added a function for counting pairs

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -185,40 +103,37 @@ class TraceState
...
+        [&header_s, &first](nostd::string_view key, nostd::string_view value) noexcept {
+          if (!first)
+          {
+            header_s.append(",");
+          }
+          else
+          {
+            first = false;
+          }
+          header_s.append(std::string(key));
```
lalitb, line 117: This is probably my mistake - but I don't think this conversion (nostd::string_view to std::string) will work properly always (eg if `key` contains end of line character). As C++11 std::string constructor won't accept std::string_view type. Better to use proper conversion : https://stackoverflow.com/questions/59424390/how-to-correctly-create-stdstring-from-a-stdstring-view

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -185,40 +103,37 @@ class TraceState
...
+          header_s.append(1, kKeyValueSeparator);
```
lalitb, line 118: Need to check this conversion in other places too

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -185,40 +103,37 @@ class TraceState
...
+          header_s.append(std::string(value));
```
lalitb, line 119: Need to check this conversion in other places too

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -265,35 +181,37 @@ class TraceState
    */
   nostd::shared_ptr<TraceState> Delete(const nostd::string_view &key)
   {
-    nostd::shared_ptr<TraceState> ts{new TraceState()};
-
     if (!IsValidKey(key))
     {
-      return ts;
+      return TraceState::GetDefault();
```
lalitb, line 186: Earlier this was returning same trace_state instance. Why are we returning default now ?

nikhil1511: Earlier also it is returning empty trace state, right ? We are creating new instance and returning it. Am I missing anything here ?

api/test/trace/trace_state_test.cc
```diff
@@ -7,58 +21,14 @@ namespace
 {
 
 using opentelemetry::trace::TraceState;
+namespace nostd = opentelemetry::nostd;
```
ThomsonTan, line 24: This seems not used?

nikhil1511: Its used in the newly added tests (check line 200). I see that there is some inconsistency, at some places opentelemetry::nostd is used and at some nostd is directly used.

api/include/opentelemetry/trace/trace_state.h
lalitb: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/trace_state.h#L270-L273
Maybe I would have done it wrong in my draft PR ( if taken from there), but existing code does return same trace_state instance. Which actually makes sense too.

nikhil1511: The trace_state returned is empty, right ? I am also taking reference from the same code. The ts created is empty.

Also,  expected behaviour is captured in the comments.
@returns empty TraceState object if key is invalid

lalitb: ok yeah, my bad :)

lalitb (approved):
Thanks for the changes. Looks good to me once we take care of string_view to string conversion. Will wait for feedback from others.

api/test/trace/trace_state_test.cc
ThomsonTan: I'd suggest keeping its usage consistent.

ThomsonTan (approved):

PR #644, opened 2021-04-01
Add Global Propagator, Noop Propagator

### LINKED ISSUE #540: Implement global propagators
As per specs : https://github.com/open-telemetry/opentelemetry-specification/blob/11cc73939a32e3a2e6f11bdeab843c61cf8594e9/specification/context/api-propagators.md

**Global Propagators**
The OpenTelemetry API MUST provide a way to obtain a propagator for each supported Propagator type. Instrumentation libraries SHOULD call propagators to extract and inject the context on all remote calls. Propagators, depending on the language, MAY be set up using various dependency injection techniques or available as global accessors.

lalitb:
Fixes #540 

## Changes

Implement [Global Propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#global-propagators) class to set global-textmap propagator to be used in instrumented libraries. It returns noop-propagator if no such propagator is set.


api/include/opentelemetry/trace/propagation/global_propagator.h
```diff
@@ -0,0 +1,53 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace trace
+{
+namespace propagation
+{
+
+/* Stores the singleton TextMapPropagator */
+
+template <typename T>
```
jsuereth, line 21: IIUC:  We'll have a global propogated PER carrier type here, and need to register propogators as such.

I think this actually isn't what the Spec calls for.  We may need to change the TextMapPropogator to *not* be templatized so that there's a single global propogator, and we find an alternative mechanism to hide the carrier.


Specifically something like:

```
public TextMapPropagator
{
public:
  typename Getter = std::function<nostd::string_view(nostd::string_view trace_type)>;
  typename Setter = std::function<void (nostd::string_view, nostd::string_view)>;
  virtual context::Context Extract(Getter get_from_carrier, context::Context &context) noexcept = 0;
  virtual void Inject(Setter set_from_carrier, const context::Context &context) noexcept = 0;
};
```

Or even exposing something like:

```
public TextMapPropagator
{
public:
  class Getter {
  public:
   virtual nostd::string_view Get(nostd::string_view trace_type) const noexcept = 0;
  };
  class Setter {
  public:   
    virtual void Set(nostd::string_view, nostd::string_view) noexcept = 0;
  };  
  virtual context::Context Extract(const Getter& get_from_carrier, context::Context &context) noexcept = 0;
  virtual void Inject(Setter* set_to_carrier, const context::Context &context) noexcept = 0;
};
```


The ideal split here is:

- TextMapPropogator knows *which keys* to look for and how to set them.  This logic can be set globally and shared across carriers.
- Getter/Setter know how to get and set  keys on a carrier.

Right now the current Propagators conflate the two because any generic propagator MUST be templatized and be compiled against its target carrier.  If you want to provide a global TextMapPropogator to be used against multiple carriers you'd need to register against ALL possible carrier types.  While you *could* solve this with a type-level registry and synthesize of text-map propagator, that code becomes unreadable quickly (in my opinion).

For demonstration, think about an example application with both gRPC + HTTP endpoint where we want to use text-map-propogator.   Maybe it's worth adding this to one of our examples so we can see the implications in this api?

lalitb: Thanks @jsuereth for detailed comment and suggestions. You are right - Global propagator shouldn't be needing carrier type while setting and getting them. I did realize that while implementing. And since it needed change in `TextMapPropagator` and all other propagators inhering from it, I was seeing possibility of avoid these changes. Let me work on these changes.

lalitb: @jsuereth  - it seems to be a bit complicated than what I thought initially. As the Getters and Setter method should be stateless as per specs. This means we need to pass the carrier object to these methods. and so TextMapPropagator needs to know the carrier type, which further means we need to make TextMapPropagator as a template class.
This is what I tried to get rid of the template, but it makes Getter/Setter stateful. While it works, but it's not compliant with specs (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#textmap-propagator ).


```cpp

class TextMapCarrier
{

public:
  virtual nostd::string_view Get(const nostd::string_view &key) const noexcept              = 0;
  virtual void Set(const nostd::string_view &key, const nostd::string_view &value) noexcept = 0;
};

class TextMapPropagator
{
public:
  // Returns the context that is stored in the carrier with the getter as extractor.
  virtual context::Context Extract(const TextMapCarrier &carrier,
                                   context::Context &context) noexcept = 0;

  // Sets the context for carrier with self defined rules.
  virtual void Inject(TextMapCarrier &carrier, const context::Context &context) noexcept = 0;
};

```
And then create HTTP carrier :
```cpp
class TextMapCarrierHttp : public trace::propagation::TextMapCarrier
{
public:
  virtual nostd::string_view Get(const nostd::string_view &key) const noexcept override
  {
    auto it = headers_.find(std::string(key));
    if (it != headers_.end())
    {
      return nostd::string_view(it->second);
    }
    return "";
  }
  virtual void Set(const nostd::string_view &key, const nostd::string_view &value) noexcept override
  {
    headers_[std::string(key)] = std::string(value);
  }

  std::map<std::string, std::string> headers_;
};
```

Reference: https://github.com/open-telemetry/opentelemetry-cpp/compare/main...lalitb:global-propagator-2

jsuereth: @lalitb I think, in this case, we should be willing to deviate from the spec to account for C++-isms and challengess.  Specfically, it likely makes sense to raise this against the spec to see if we can relax the requirements to give C++ room to do something good for C++.

I like the solution you outline because it balances "global text map propagator" with "adaptable context carrier" in a way that likely has the least amount of duplicate code for real users.  My only question is whether ytou think the "Get" and "Set" interfaces for TextMapCarrier need to be split in practice?  I.e. are their C++ HTTP frameworks where you only have readonly access?   What you've defined (with const) lines up more with my C++ expectations.

pyohannes: @lalitb I think what you have sketched here is in line with the spec. I think we shouldn't get hung up on stateless getters and setters: the `Get` and `Set` methods you have are actually part of the carrier, and not carrier-independent getter and setter methods (which the spec requires to be stateless):

> Getter and Setter are optional helper components used for extraction and injection respectively, and are defined as separate objects from the carrier to avoid runtime allocations, by removing the need for additional interface-implementing-objects wrapping the carrier in order to access its contents.

So we'll just have to go with "additional interface-implementing-objects wrapping the carrier".

lalitb: thanks @jsuereth @pyohannes for the comments. Will create a PR based on the above sketch, and can discuss it further over there.

seemk: I think `GlobalPropagator` is an unfortunate spec addition that complicates things more it solves, especially since it has the `MUST` keyword.

In fact I don't think propagators should be forced to be classes, given that propagators most likely hold no state at all. They could very well be free functions, i.e.
```c++
template <typename Carrier>
void B3Extract(const Carrier& carrier, context::Context& context) {
   ...
   carrier.get("b3");
   ...
} 
```

In C or C++ instrumentations I don't see myself using `GlobalPropagator` at all if I can just do `B3Propagator().Extract(...)`.

P.S. This comment is geared towards the spec and not this PR's content.

api/include/opentelemetry/trace/propagation/global_propagator.h
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <mutex>
```
seemk, line 3: I think this is unused and can be removed

api/include/opentelemetry/trace/propagation/global_propagator.h
lalitb: @seemk  Regarding `GlobalPropagator` we do need them for the instrumented library, as the libraries don't need to worry about which propagator to use or is configured. It so allows a  given instrumented library to be used in microservice applications using different propagation protocols.

Regarding `inject` and `extract` methods, I agree they can be free functions as they don't hold any state.

BTW, the propagators have been non-templatized, and the global propagator is merged as part of #668  - This may need changes to contrib repo instrumentation. I will close this PR now.

lalitb:
closing as this PR is not relevant after merging: #668

PR #650, opened 2021-04-05
Create shared context for updating span pipeline from TracerProvider and affecting Tracer.

jsuereth:
- Create new TracerContext class that stores (thread-safe) access to Sampler + SpanPrcoessor.
- All calls to SpanProcessor go through TracerContext, ensuring we abide by specification where TracerProvider calls MUST impact previously created Tracers
- Update all examples for new API.
- Adds "stub" API for appending new SpanProcessors to a given TracerContext.

Replaces #590 with only the shared context portion.


lalitb (approved):
LGTM. Thank for the changes.

sdk/src/trace/tracer_context.cc
```diff
@@ -0,0 +1,42 @@
...
+  return resource_;
+}
+
+void TracerContext::RegisterPipeline(std::unique_ptr<SpanProcessor> processor) noexcept
+{
+  // TODO(jsuereth): Implement
+  // 1. If existing processor is an "AggregateProcessor" append the new processor to it.
+  // 2. If the existing processor is NOT an "AggregateProcessor", create a new Aggregate of this and
+  // the other,
+  //    then replace our atomic ptr with the new aggregate.
```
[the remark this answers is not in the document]
jsuereth: Oh, wow, yeah just copy-pasted the documentation.  I'll rephrase this for now until the next PR lands.

seemk (approved):
I apologize for being out of the loop a bit, but what was the problem this (and #590) solved in the context of Resource SDK? Just to get up to speed, thanks!

lalitb:
> I apologize for being out of the loop a bit, but what was the problem this (and #590) solved in the context of Resource SDK? Just to get up to speed, thanks!

@seemk  - This PR is to ensure that the tracer-provider has sole ownership of all the configuration ( sampler, processor, resources ).  You can follow the discussions in #580 specifically these comments:

https://github.com/open-telemetry/opentelemetry-cpp/pull/580#issuecomment-782305522
https://github.com/open-telemetry/opentelemetry-cpp/pull/580#issuecomment-782248462
https://github.com/open-telemetry/opentelemetry-cpp/pull/580#issuecomment-783734011 ( the final proposal which is implemented in this PR).

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -0,0 +1,79 @@
...
+   * Obtain the resource associated with this tracer context.
+   * @return The resource for this tracer context.
+   */
+  const opentelemetry::sdk::resource::Resource &GetResource() const noexcept;
+
+  /**
+   * Force all active SpanProcessors to flush any buffered spans
+   * within the given timeout.
+   */
+  void ForceFlushWithMicroseconds(uint64_t timeout) noexcept;
```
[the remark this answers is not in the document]
jsuereth: doh! yeah, never fully resolved merge conflict, fixed.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * A class which stores the TracerProvider context.
+ *
+ * This class meets the following design criteria:
+ * - A shared reference between TracerProvider and Tracers instantiated.
+ * - A thread-safe class that allows updating/altering processor/exporter pipelines
+ *   and sampling config.
+ * - The owner/destroyer of Processors/Exporters.  These will remain active until
+ *   this class is destroyed.  I.e. Sampling, Exporting, flushing etc. are all ok if this
+ *   object is alive, and they willl work together.   If this object is destroyed, then
```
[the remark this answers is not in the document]
jsuereth: fixed

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -0,0 +1,85 @@
...
+   */
+  Sampler &GetSampler() const noexcept;
+
+  /**
+   * Obtain the (conceptual) active processor.
+   *
+   * Note: When more than one processor is active, this will
+   * return an "aggregate" processor
+   */
+  SpanProcessor &GetActiveProcessor() const noexcept;
```
pyohannes, line 56: What would be an _inactive_ processor?

pyohannes, line 56: We might theoretically run into issue here when some tracer does some work with a processor obtained by `GetActiveProcessor`, while another thread calls `RegisterPipeline` and thus causes the memory pointed to by the obtained the processor to be freed.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * Obtain the resource associated with this tracer context.
+   * @return The resource for this tracer context.
+   */
+  const opentelemetry::sdk::resource::Resource &GetResource() const noexcept;
+
+  /**
+   * Force all active SpanProcessors to flush any buffered spans
```
pyohannes, line 65: I'm a bit confused by _all active_ span processor. Wouldn't we just have a `CompositeSpanProcessor`? Which would encapsulate flushing and shutdown? So that this is a thing the `CompositeSpanProcessor` has to worry about, and not the `TraceContext`?

sdk/include/opentelemetry/sdk/trace/tracer_provider.h
```diff
@@ -26,34 +27,30 @@ class TracerProvider final : public opentelemetry::trace::TracerProvider
...
-   * @return The span processor for this tracer provider.
-   */
-  std::shared_ptr<SpanProcessor> GetProcessor() const noexcept;
-
-  /**
-   * Obtain the sampler associated with this tracer provider.
-   * @return The sampler for this tracer provider.
-   */
-  std::shared_ptr<Sampler> GetSampler() const noexcept;
+  void RegisterProcessor(std::unique_ptr<SpanProcessor> processor) noexcept;
```
pyohannes, line 53: We have `RegisterProcessor` here and `RegisterPipeline` on the `TraceContext`. Could we unify the terminology here? I don't have a strong preference as to which term to use.

sdk/src/trace/span.cc
```diff
@@ -58,18 +58,15 @@ trace_api::SpanId GenerateRandomSpanId()
 }
 
 Span::Span(std::shared_ptr<Tracer> &&tracer,
-           std::shared_ptr<SpanProcessor> processor,
```
pyohannes, line 61: There is a reason why the processor was put onto the span: this is to guarantee that the span/recordable gets ended on the same processor it was created with.

Otherwise, this could happen:
1. A span/recordable _a_ is created from processor/exporter A.
2. The processor/exporter is changed, maybe replaced by a composite processor/exporter B.
3. When the span/recordable _a_ is ended it is passed to the processor/exporter B and cannot be handled in this context (as it was created from processor/exporter A).

We should decide if that's still a use case we want to support. Just putting it out there that this was considered in the initial design.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
pyohannes: Personally, I wouldn't mind leaving `ForceFlush` and `Shutdown` out. `GetActiveProcessor().ForceFlush()` isn't that bad.

pyohannes (commented):
Thanks for doing that work and putting this chunk into its own PR. This makes it much easier to reason about and to review.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
jsuereth: Shouldn't be the case.  The interface is "RegisterPipeline", and it's a shared pointer.  The TracerContext can guarantee that ALL registered processors remain alive for its own lifecycle.   I.e. we're not going to delete a pipeline, we're only going to allow registration of new ones specifically to prevent this issue.

We have the inverse problem around Recordable though, specifically if we have an "active" span that did NOT get a recordable registered when a new pipeline is registered, what happens when that span ends?

If you look at my design around "ExportableSpan" where a pipeline can ask for its instance of Recordable (and handle a missing recordable), that's how I planned to solve that issue.

sdk/src/trace/span.cc
jsuereth: Ah, I had a different proposal where a processor asks for its recordable back later and receives a nullptr (or other "false" flag).

I can revert this for the other design.  I think this *is* a use case that needs support, there's just more than one option for how to do it.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
jsuereth: This is for users of TracerContext to  understand what it does.   Yes the implementation of this behavior may be implemented by "CompositeSpanProcessor" but it's *not* something that would be directly exposed, instead users just see "RegisterPipeline" and "Shutdown" and "ActiveProcessor".  I.e. we still need to document the behavior of this class and how to use its methods.

Note: I'm not tied to "active" processor, I borrowed the name from the Java SDK.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
pyohannes: I think I understand. So if a new pipeline is registered, the old one isn't just overwritten, but kept in store and just made _inactive_?

sdk/src/trace/span.cc
pyohannes: I don't mind which option we take. I'm in favor though of guaranteeing that the `Recordable` passed to the exporter is of the correct type.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
lalitb: My understanding was that all the pipelines(i.e, processors) remain active. And we can only add a new pipeline, but can't remove the existing one? And once we add multi-processor support, `TracerContext` would actually maintain a unique_ptr to (say) `CompositeSpanProcessor` which further manages all the active pipelines/processors?

lalitb: I don't think there is any `inactive` processor. @jsuereth can confirm though :)

jsuereth: Yeah.  i'm adopting the design I've seen elsewhere where you can "append a pipeline", but can't remove an existing pipeline.

I.e. once a pipeline is active, it's forever active.   We don't have a register/unregister mechanism.

jsuereth: 'inactive' would be any constructed but not registered processor.  

I'm not tied to this name, just took it from Java ;)

lalitb:
@jsuereth -  Would like to merge this PR once you and @pyohannes think it's good to go , that will allow me to raise PR for #664 on top of these changes :)

jsuereth:
@lalitb Understood.   So, to make  progress, it'd be nice to know which of @pyohannes comments are blocking and which to address now.


- Active Processor terminology.   I don't see a counter proposal and i haven't thought of another term. I'd vote to leave it similar to Java unless there's an alternative.
- Keeping `SpanPrcoessor` on Span vs. relying on "active span processor" in tracer.  I'm ok going either way here.  my preference is what I've done in this PR but I'm happy to change it.
- I can update to consistently use `RegisterPipeline` everywhere as the spec uses the term 'pipeline" to denote processor+exporter pairing.

WDYT @pyohannes ?

pyohannes:
Please let's use `RegisterPipeline` consistently wherever appropriate.

Otherwise I'm fine with the current approach. I was a bit confused by the missing `RegisterPipeline` implementation, but I think my main concerns will be handled by a proper implementation of this method and aggregate processors.

pyohannes (approved):
Thanks for getting this on the way.

ext/test/zpages/tracez_processor_test.cc
```diff
@@ -178,8 +178,12 @@ class TracezProcessor : public ::testing::Test
     shared_data   = std::shared_ptr<TracezSharedData>(new TracezSharedData());
     processor     = std::shared_ptr<TracezSpanProcessor>(new TracezSpanProcessor(shared_data));
     auto resource = opentelemetry::sdk::resource::Resource::Create({});
+    // Note: we make a *different* processor for the tracercontext. THis is because
```
ThomsonTan, line 181:
```suggestion
    // Note: we make a *different* processor for the tracercontext. This is because
```

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -0,0 +1,85 @@
...
+  explicit TracerContext(std::unique_ptr<SpanProcessor> processor,
+                         opentelemetry::sdk::resource::Resource resource =
+                             opentelemetry::sdk::resource::Resource::Create({}),
+                         std::unique_ptr<Sampler> sampler =
+                             std::unique_ptr<AlwaysOnSampler>(new AlwaysOnSampler)) noexcept;
+  /**
+   * Attaches a span processor to this tracer context.
+   *
+   * @param processor The new span processor for this tracer. This must not be
+   * a nullptr.  Ownership is given to the `TracerContext`.
```
ThomsonTan, line 40:
```suggestion
   * a nullptr. Ownership is given to the `TracerContext`.
```

sdk/include/opentelemetry/sdk/trace/tracer_context.h
ThomsonTan: Perhaps just name it `GetProcessor` to avoid the confusion.

ThomsonTan:
Update SDK change log?

ThomsonTan (approved):
Update change log?

lalitb:
> @lalitb Understood. So, to make progress, it'd be nice to know which of @pyohannes comments are blocking and which to address now.
> 
> * Active Processor terminology.   I don't see a counter proposal and i haven't thought of another term. I'd vote to leave it similar to Java unless there's an alternative.
> * Keeping `SpanPrcoessor` on Span vs. relying on "active span processor" in tracer.  I'm ok going either way here.  my preference is what I've done in this PR but I'm happy to change it.
> * I can update to consistently use `RegisterPipeline` everywhere as the spec uses the term 'pipeline" to denote processor+exporter pairing.
> 
> WDYT @pyohannes ?

Thanks for summarising @jsuereth. I would be happy to discuss these comments or change if required in subsequent PR for multiple processors. Meanwhile, if you could unblock to merge it by fixing the formatting :)

PR #651, opened 2021-04-05
Enable resource detector tests on Windows

ThomsonTan:
## Changes

The resource detector tests were disabled on Windows because of missing C library call `unsetenv`, but it could be achieved by `_putenv("ENV_VAR_NAME=")` on Windows.


sdk/test/resource/resource_test.cc
```diff
@@ -10,6 +10,10 @@
 
 #include <gtest/gtest.h>
 
+#if defined(_WIN32)
+#  define putenv _putenv
```
maxgolov, line 14: I think, ideally, strictly speaking if you are compiling with Visual Studio `#ifdef _MSC_VER` , then it would be appropriate to use `putenv_s` instead: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-160

Without it you'd get Secure Development Lifecycle warning.

sdk/test/resource/resource_test.cc
```diff
@@ -173,13 +174,21 @@ TEST(ResourceTest, OtelResourceDetector)
     }
   }
   EXPECT_EQ(received_attributes.size(), expected_attributes.size());
+#if defined(_WIN32)
+  putenv("OTEL_RESOURCE_ATTRIBUTES=");
```
maxgolov, line 178: I think if you compile at "Warnings Level 4" in IDE, you'd see the warning here. There's another place in the actual code (not in tests) that also has the same issue.

maxgolov: I logged an issue for the other one - #655

sdk/test/resource/resource_test.cc
```diff
@@ -173,13 +174,21 @@ TEST(ResourceTest, OtelResourceDetector)
...
+#else
   unsetenv("OTEL_RESOURCE_ATTRIBUTES");
```
maxgolov, line 180: There is generally a huge issue with the env var approach for Win 10 Universal apps, i.e. someone ever trying to compile this for Xbox is gonna be having some fun time with environment-variables based approach. Not a blocker for this PR, but something we need to revisit in the `putenv` / resource API :

https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/cppcx/crt-functions-not-supported-in-universal-windows-platform-apps.md

sdk/test/resource/resource_test.cc
ThomsonTan: > I think, ideally, strictly speaking if you are compiling with Visual Studio `#ifdef _MSC_VER` , then it would be appropriate to use `putenv_s` instead: [docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-160](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv-s-wputenv-s?view=msvc-160)
> 
> Without it you'd get yet another Secure Development Lifecycle warning in this spot. I do realize that this requirement is unique to Visual Studio C++ compiler, and that other runtimes generally do not implement the "secure" version of this function. So it's foolish to use any of that "secure" stuff in a cross-plat portable code, because there is no conforming runtime library with that available for other OS / compilers... sigh.
> 
> But - you already have `#ifdef _WIN32` here, might as well assume for now that on Windows we only maintain compilers that operate with MSVCRT runtime (_even clang on Windows should be fine when used with MSVCRT_), where the "more secure" function is available?
> 
> The simpler option is to check for `#ifdef _MSC_VER`, and more complicated option to do double-check for both - OS + compiler or extension, OR fallback to insecure:
> 
> * if  ( `_WIN32` and `_MSC_VER` ) or ( `__STDC_WANT_LIB_EXT1__ ` ) , then use "secure" `putenv_s`
> * if `_WIN32` but not `_MSC_VER` (exotic scenarios - `MinGW` or `clang+libstdc++`, runtime where SDL extension functions are not available), then use "insecure" `putenv`
> 
> I agree with critics of this extension. I am one of the critics of this extension. However, real-world projects often require this extension. It is not in our authority to dictate to people to stop using this extension if they are already using it. At scale... So if it's available, I'd rather say we should check if it's there, and we should be using ***_s** (secure Annex K functions) if these are available in the runtime library.

Changed the macro to `_MS_VER`. 

Actually, `_putenv` doesn't trigger C4996 like below for `getenv`. I think it makes sense because it doesn't return any CRT allocated buffer. 
>  warning C4996: 'getenv': This function or variable may be unsafe.

I also looked at the `stdlib.h` from MSVC, and found that `_putenv` is not declared with `_CRT_INSECURE_DEPRECATE`, so we could still use it without hurting the secruity. The difference provided by `_putenv_s` is it separates environment variable name and value while `_putenv` pass them in one string and separated by `=`.

sdk/test/resource/resource_test.cc
ThomsonTan: The same as above, `_putenv` doesn't trigger CRT security warning.

sdk/test/resource/resource_test.cc
ThomsonTan: Thanks for calling it out. Yes, let's revisit this later.

sdk/test/resource/resource_test.cc
maxgolov: Ok, I'm now wondering what is the purpose of `putenv_s` at all 😄

maxgolov (approved):
I'm still not sure what is the purpose of `putenv_s` 🤷 

LGTM

I greatly appreciate your follow-up email to Visual C++ runtime team to create clarity in this regard.

PR #658, opened 2021-04-06
Refactor/duplicate enums

### LINKED ISSUE #420: Remove duplicate enum declarations
Identical enums that are defined in multiple files can be removed and reused instead. An example is the `ExportResult` enum used in the `exporter.h` file in the [trace SDK](https://github.com/open-telemetry/opentelemetry-cpp/blob/master/sdk/include/opentelemetry/sdk/trace/exporter.h#L15) and the [logs SDK](https://github.com/open-telemetry/opentelemetry-cpp/pull/403/files#diff-d506356831038e0a23cefad4aa469b1f73ef30b67d9eb8d8910ad42d6d9bf482R33) folders.  Both implement the enum values `kSuccess` and `kFailure`. This enum could be shared instead of redefined in both places.

anshulkumar19:
Fixes #420 

## Changes
- Refactoring to put ExportResult enum class into sdk/common
- Reusing this class to remove duplication


lalitb (approved):
Thanks for the fix.

PR #659, opened 2021-04-06
Add SpanContext accessor SpanDataLink

### LINKED ISSUE #657: `sdk::Trace::SpanDataLink` doesn't give access to key link information
I was trying to read the links for testing purposes, only to realise that the key part of the link (the span context referenced) isn't accessible. A TODO kinda summarises the gap here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/afa2e79819a2fddb44487db8ffdf95aa62f47fc7/sdk/include/opentelemetry/sdk/trace/span_data.h#L64

There are a few ways of fixing this.
1. As per the TODO, forward all accessors of the held `SpanContext` to the `SpanDataLink`
2. Inherit `SpanDataLink` from `SpanContext` (similar interfaces to the above)
3. Provide an accessor (ideally const &) to the held `SpanContext`

Personally, I suggest 3, given that the specification mentioned that the SpanContext (presumably in its entirety, including trace_state/flags and any future features...) to be part of the Link: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#specifying-links  . This would be the same for 2, but for 1 it suggests that this class needs to be forever paired with SpanContext and co-evolve. 

None available, 3 options of how to address this have been highlighted above.

eyjohn:
Fixes #657

## Changes

Although the context was stored, it was not accessible. I've extended
tests to also validate that a valid context has been passed.

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/span_data.h
+++ b/sdk/include/opentelemetry/sdk/trace/span_data.h
@@ -61,7 +61,6 @@ private:
 
 /**
  * Class for storing links in SpanData.
- * TODO: Add getters for trace_id, span_id and trace_state when these are supported by SpanContext
  */
 class SpanDataLink
 {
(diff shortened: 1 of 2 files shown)


lalitb (approved):
Thank you for the fix.

PR #674, opened 2021-04-13
Resolve "Level 4" warnings about unused parameter in interface definitions in span API headers

maxgolov:
## Changes

Main motivation for this PR is to resolve ALL compilation warnings in Visual Studio, including (somewhat silly) Level 4 warnings.

API headers contain "interface definition" / Default implementation of `Span`. Plus `NullSpanContext` - empty context. In both cases the methods declare parameters that are not used. This is tripping a (fairly useless) [Level 4 warning](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100?view=msvc-160) in Visual Studio.

There are two implementation options here:

- comment out parameter names : this may be causing a potential problem for tools like Doxygen. But it's a universal approach that resolves the issue in a compiler-agnostic way. Approach works well even if the same paranoid warning is enabled with gcc or clang compiler, i.e. universally works for all compilers.

OR

- add compiler-specific warning suppression pragma, i.e. suppress c4100 for Visual Studio and add `#pragma GCC diagnostic ignored ...` for GCC.

I'm implementing the first option above. But open to opinions / suggestions. The second option also looks reasonable to me, as long as we can ultimately resolve this silly warning. 😄

### DIFF
--- a/api/include/opentelemetry/trace/default_span.h
+++ b/api/include/opentelemetry/trace/default_span.h
@@ -17,15 +17,17 @@ public:
 
   bool IsRecording() const noexcept { return false; }
 
-  void SetAttribute(nostd::string_view key, const common::AttributeValue &value) noexcept {}
+  void SetAttribute(nostd::string_view /* key */,
+                    const common::AttributeValue & /* value */) noexcept
+  {}
 
-  void AddEvent(nostd::string_view name) noexcept {}
+  void AddEvent(nostd::string_view /* name */) noexcept {}
 
-  void AddEvent(nostd::string_view name, core::SystemTimestamp timestamp) noexcept {}
+  void AddEvent(nostd::string_view /* name */, core::SystemTimestamp /* timestamp */) noexcept {}
 
-  void AddEvent(nostd::string_view name,
-                core::SystemTimestamp timestamp,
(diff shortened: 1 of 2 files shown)


maxgolov:
lol at `codecov/project — 94.66% (-0.02%) compared to e8e2fe2` again.. sigh

lalitb (approved):
LGTM, no strong preference, just that I saw use of macro-like `#define UNUSED_PARAMETER(...)` and `(void)unused_var` statement in cpp-client-telemetry code.

lalitb:
> lol at `codecov/project — 94.66% (-0.02%) compared to e8e2fe2` again.. sigh

The only difference I see here is that our formatting tool has added extra lines for the `SetAttribute` method, and if codecov calculates percentage based on the number of lines of code executed, it will reduce the percentage if this method is never executed.

pyohannes (approved):
Thanks for fixing this. I prefer the approach you've taken, it's also in line with [ISO C++ guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused).

reyang (approved):

PR #679, opened 2021-04-16
Attribute value cleanup

### LINKED ISSUE #634: Only support signed 64 bit integers as attribute values
For attribute values, the [specification ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes) only asks for one integer value: signed 64 bit integers.

However, currently our API currently supports [four integer types](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/common/attribute_value.h#L15):
* `uint32_t`
* `uint64_t`
* `int32_t`
* `int64_t`

OTLP also [only supports signed 64 bit integers](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/common/v1/common.proto#L33), so we will run into problems with unsigned 64 bit integers there.

I recommend aligning with the specification and to only support `int64_t`. I can put up a PR for that, but this is breaking change, so please give your opinions here.

pyohannes:
Fixes #634.

## Changes

This change is based on discussion we recently had (in #634 and in SIG meetings).

This PR purposely tries to keep the scope of `AttributeValue` very wide, as any change to this `variant` will require us to release a new major version, as any change to the `variant` is a breaking API change.

* `std::span<uint8_t>` is included as there's an [open issue](https://github.com/open-telemetry/opentelemetry-specification/issues/780) for adding this, and there's interest from user side to have this.
* `uint64_t` and `std::span<uint64_t>` are added to provide complete support for primitive C++ types.

The documentation adds a warning that those types are not part of the specification and that supports from exporter side is not guaranteed.

### DIFF
--- a/api/include/opentelemetry/common/attribute_value.h
+++ b/api/include/opentelemetry/common/attribute_value.h
@@ -10,26 +10,47 @@
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace common
 {
+/// OpenTelemetry signals can be enriched by adding attributes. The
+/// \c AttributeValue type is defined as a variant of all attribute value
+/// types the OpenTelemetry C++ API supports.
+///
+/// The following attribute value types are supported by the OpenTelemetry
+/// specification:
+///  - Primitive types: string, boolean, double precision floating point
+///    (IEEE 754-1985) or signed 64 bit integer.
+///  - Homogenous arrays of primitive type values.
+///
+/// \warning 
+/// \parblock The OpenTelemetry C++ API currently supports several attribute
+/// value types that are not covered by the OpenTelemetry specification:
+///  - \c uint64_t
+///  - \c nostd::span<const uint64_t>
+///  - \c nostd::span<uint8_t>
+///
+/// Those types are reserved for future use and currently should not be
+/// used. There are no guarantees around how those values are handled by
+/// exporters.
+/// \endparblock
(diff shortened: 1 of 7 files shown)


lalitb (approved):

maxgolov (approved):

api/include/opentelemetry/common/attribute_value.h
```diff
@@ -10,46 +10,67 @@
...
                    nostd::span<const double>,
-                   nostd::span<const nostd::string_view>>;
+                   nostd::span<const nostd::string_view>,
+                   // Not currently supported by the specification, but reserved for future use.
+                   // Added to provide support for all primitive C++ types.
+                   uint64_t,
+                   // Not currently supported by the specification, but reserved for future use.
+                   // Added to provide support for all primitive C++ types.
+                   nostd::span<const uint64_t>,
+                   // Not currently supported by the specification, but reserved for future use.
```
ThomsonTan, line 53: nit: this comment is duplicated with the above one at line 47?

ThomsonTan (approved):

api/include/opentelemetry/common/attribute_value.h
pyohannes: Just added it to every variant type that is not supported, to avoid any misunderstandings.

PR #681, opened 2021-04-19
Support Custom span-id and trace-id generator

### LINKED ISSUE #677: Custom Trace and Span IDs Generation
Custom IDs would allow users to set IDs that they need right at the beginning of a span. This is important if the downstream services or eventual backend that receives these spans need them to be in a certain format.

As per the specs:
https://github.com/open-telemetry/opentelemetry-specification/blob/b67faa4e0106cac194a837e3d419ae4fab9fc1c4/specification/trace/sdk.md#id-generators
```The SDK MUST provide a mechanism for customizing the way IDs are generated for both the TraceId and the SpanId.```

This is how other languages support it:
JS: https://github.com/open-telemetry/opentelemetry-js/blob/b7d6e74c5bec6a24f5d617d31402ea95c54f126f/packages/opentelemetry-core/src/trace/IdGenerator.ts

Java: https://github.com/open-telemetry/opentelemetry-java/blob/485cc52c6ff1bb3fa12cd950d6eb97d3ceae6e41/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/IdsGenerator.java

### PR DESCRIPTION
**Pull request by lalitb**
Fixes #677 

## Changes
  - Add abstract interface for a span-id and trace-id generation.
  - Convert existing logic for random span-id and trace-id as the default implementation for the above interface.
   - Modify TracerProvider and TracerContext constructor to support injecting the custom id generator.

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/id_generator.h
+++ b/sdk/include/opentelemetry/sdk/trace/id_generator.h
@@ -1,3 +1,17 @@
+// Copyright 2021, OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #pragma once
 #include "opentelemetry/trace/span_id.h"
 #include "opentelemetry/trace/trace_id.h"
--- a/sdk/include/opentelemetry/sdk/trace/id_generator.h
+++ b/sdk/include/opentelemetry/sdk/trace/id_generator.h
@@ -1,0 +1,35 @@
+#pragma once
+#include "opentelemetry/trace/span_id.h"
+#include "opentelemetry/trace/trace_id.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace trace
+{
+
+/** IdGenerator provides an interface for generating Trace Id and Span Id */
+class IdGenerator
+{
+
+public:
+  /** Returns a SpanId represented by opaque 128-bit trace identifier */
+  virtual opentelemetry::trace::SpanId GenerateSpanId() noexcept = 0;
+
+  /** Returns a TraceId represented by opaque 64-bit trace identifier */
+  virtual opentelemetry::trace::TraceId GenerateTraceId() noexcept = 0;
+};
+
+class RandomIdGenerator : public IdGenerator
+{
+public:
+  opentelemetry::trace::SpanId GenerateSpanId() noexcept override;
+
+  opentelemetry::trace::TraceId GenerateTraceId() noexcept override;
+};
+
+}  // namespace trace
+
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/include/opentelemetry/sdk/trace/id_generator.h
+++ b/sdk/include/opentelemetry/sdk/trace/id_generator.h
@@ -28,6 +28,8 @@ class IdGenerator
 {
 
 public:
+  virtual ~IdGenerator() = default;
+
   /** Returns a SpanId represented by opaque 128-bit trace identifier */
   virtual opentelemetry::trace::SpanId GenerateSpanId() noexcept = 0;
 
--- a/sdk/include/opentelemetry/sdk/trace/tracer_provider.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer_provider.h
@@ -9,6 +9,7 @@
 #include "opentelemetry/sdk/trace/processor.h"
 #include "opentelemetry/sdk/trace/samplers/always_on.h"
 #include "opentelemetry/sdk/trace/tracer.h"
+#include "opentelemetry/sdk/trace/tracer_context.h"
 #include "opentelemetry/trace/tracer_provider.h"
 
 OPENTELEMETRY_BEGIN_NAMESPACE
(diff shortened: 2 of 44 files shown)

### CODE REVIEW

File: sdk/test/trace/tracer_test.cc [hunk 1]
```diff
@@ -45,6 +45,24 @@ class MockSampler final : public Sampler
   nostd::string_view GetDescription() const noexcept override { return "MockSampler"; }
 };
 
+/**
+ * A Mock Custom Id Generator
+ */
+class MockIdGenerator : public IdGenerator
```
eyjohn, line 51: nit: Google Mock would be great to use here, I noticed we don't have it in the repo but EXPECT_CALL and return values can all be done in the test body which would be nice

sdk/test/trace/tracer_test.cc
```diff
@@ -132,6 +152,30 @@ TEST(Tracer, StartSpanSampleOff)
...
+  char rec_trace[32], exp_trace[32];
+  char rec_span[16], exp_span[16];
+
+  cur_span_data->GetTraceId().ToLowerBase16(rec_trace);
+  id_generator->GenerateTraceId().ToLowerBase16(exp_trace);
+
+  cur_span_data->GetSpanId().ToLowerBase16(rec_span);
+  id_generator->GenerateSpanId().ToLowerBase16(exp_span);
+
+  ASSERT_EQ(std::string(rec_trace, sizeof(rec_trace)), std::string(exp_trace, sizeof(exp_trace)));
```
eyjohn, line 175: nit: EXPECT_EQ

sdk/include/opentelemetry/sdk/trace/id_generator.h
```diff
@@ -0,0 +1,49 @@
...
+namespace trace
+{
+
+/** IdGenerator provides an interface for generating Trace Id and Span Id */
+class IdGenerator
+{
+
+public:
+  /** Returns a SpanId represented by opaque 128-bit trace identifier */
+  virtual opentelemetry::trace::SpanId GenerateSpanId() noexcept = 0;
```
eyjohn, line 32: Should the contract enforce constness or not?

I can foresee an implementation that does mutate internal state, a simple implementation being an incremental counter that generates an ID based on the counter. But I'm not sure whether this is by design or not?

It seems Sampler likewise doesn't use `const`

sdk/include/opentelemetry/sdk/trace/id_generator.h
```diff
@@ -0,0 +1,49 @@
...
+
+  /** Returns a TraceId represented by opaque 64-bit trace identifier */
+  virtual opentelemetry::trace::TraceId GenerateTraceId() noexcept = 0;
+};
+
+class RandomIdGenerator : public IdGenerator
```
eyjohn, line 38: nit: should the derived class live in a separate file? it does for samplers

eyjohn (approved):
Overall looks great to land, some minor nitpicks,

Minor question about constness

sdk/test/trace/tracer_test.cc
```diff
@@ -132,6 +152,30 @@ TEST(Tracer, StartSpanSampleOff)
...
+
+  tracer->StartSpan("span 1")->End();
+  auto spans          = span_data->GetSpans();
+  auto &cur_span_data = spans.at(0);
+
+  char rec_trace[32], exp_trace[32];
+  char rec_span[16], exp_span[16];
+
+  cur_span_data->GetTraceId().ToLowerBase16(rec_trace);
+  id_generator->GenerateTraceId().ToLowerBase16(exp_trace);
```
pyohannes, line 170: I think we can compare ids directly? Like `EXPECT_EQ(cur_span_data->GetTraceId(), id_generator->GenerateTraceId());`?

sdk/include/opentelemetry/sdk/trace/id_generator.h
pyohannes: I agree. We shouldn't use `const`, as we can't assume the `IdGenerator` to be stateless.

sdk/include/opentelemetry/sdk/trace/tracer_provider.h
```diff
@@ -27,11 +27,14 @@ class TracerProvider final : public opentelemetry::trace::TracerProvider
...
-                              opentelemetry::sdk::resource::Resource::Create({}),
-                          std::unique_ptr<Sampler> sampler =
-                              std::unique_ptr<AlwaysOnSampler>(new AlwaysOnSampler)) noexcept;
+  explicit TracerProvider(
+      std::unique_ptr<SpanProcessor> processor,
+      opentelemetry::sdk::resource::Resource resource =
+          opentelemetry::sdk::resource::Resource::Create({}),
+      std::unique_ptr<Sampler> sampler = std::unique_ptr<AlwaysOnSampler>(new AlwaysOnSampler),
+      std::unique_ptr<opentelemetry::sdk::trace::IdGenerator> id_generator =
+          std::unique_ptr<opentelemetry::sdk::trace::IdGenerator>(
```
ThomsonTan, line 36: Is it better to make `IdGenerator` a global static? Is there any scenario to assign a `TraceProvider` a dedicated `IdGenerator`?

lalitb: As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracer-creation), `IdGenerator` should be managed by TracerProvider
``
Configuration (i.e., SpanProcessors, IdGenerator, SpanLimits and Sampler) MUST be managed solely by the TracerProvider and it MUST provide some way to configure all of them that are implemented in the SDK, at least when creating or initializing it.
``

sdk/test/trace/tracer_test.cc
lalitb: Good point. It is bitwise comparison of `TraceId` data members ( array of uint8_t ) and it should work.

sdk/include/opentelemetry/sdk/trace/id_generator.h
lalitb: We shouldn't enforce constness, which is how `IdGenerator` is currently implemented. Let me know if you foresee any further changes?

sdk/include/opentelemetry/sdk/trace/id_generator.h
lalitb: Good point, moved derived class to separate header.

ThomsonTan (approved):

PR #686, opened 2021-04-21
Get rid of obsolete core namespace in API

pyohannes:
## Changes

The API has a `opentelemetry::core` namespace, which currently contains the classes `SystemTimestamp` and `SteadyTimestamp`. I see no reason for having separate `opentelemetry::core` and `opentelemetry::common` namespace, and I think it would be easier to just have one namespace for components that are shared for multiple OpenTelemetry signals.

This PR simplifies the API namespaces layout by removing the `opentelemetry::core` namespace and moving its items to `opentelemetry::common`.


sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -27,7 +27,7 @@ class SpanDataEvent
 {
 public:
   SpanDataEvent(std::string name,
-                core::SystemTimestamp timestamp,
+                opentelemetry::common::SystemTimestamp timestamp,
```
ThomsonTan, line 30: Define namespace alias to simplify the namespace reference?

ThomsonTan (approved):

lalitb (approved):

Issue #683, opened 2021-04-19
List of components and their dependent component licenses


maxgolov:
**List of components and their Copyright notices / Licenses**

All code contributed in this repo is licensed under **Apache License v2.0** with a few notable exceptions:

- [MPark Variant](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/nostd/mpark/variant.h) - Boost license. This code can be removed entirely and replaced by Google Abseil implementation, that is **Apache License v2.0** ( #572 ) . That ultimately allows all of OpenTelemetry C++ API be licensed under the same **Apache License v2.0**.

- [HTTP client](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/ext/include/opentelemetry/ext/http/client/curl) presently depends on [libcurl](https://curl.se/libcurl/) , which is implemented under its own [MIT/X-alike license](https://curl.se/docs/copyright.html) . This is required for Zipkin and Elastic exporters.

- [nlohmann/json](https://github.com/nlohmann/json) - _Required_ for Zipkin and Elastic exporter. _Required_ for zPages. And _Optional_ for specific subtype of ETW exporter. The library is licensed under [MIT License](https://github.com/nlohmann/json/blob/develop/LICENSE.MIT). Products builds that are shipping with OTLP gRPC exporter or ETW exporter, and NOT including zPages - do not require this module, but it may come in handy for other aspects, e.g. remote provisioning of exporter using JSON configuration.

- [TraceLogging Dynamic for Windows](https://github.com/microsoft/tracelogging-dynamic-windows) library - strongly recommended for ETW exporter, as default transport channel - is also [MIT License](https://github.com/microsoft/tracelogging-dynamic-windows/blob/main/LICENSE). Products shipping on Linux do not require this component.

- Certain other parts of OTLP exporter may require **Boost** ? cc: @ThomsonTan to confirm on where it will be required.

**Proposal**

The goal is to **create a markdown document that captures above detail with all the licenses**; alongside with build options that enable these components mentioned above. Not every build combo requires all the licenses. Having the OSS components document enables any product commercially shipping with OpenTelemetry C++ SDK to easily list all the required licenses in their distro, prepackaged build of SDK. Reference example of [OSS components document](https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/List-of-OSS-Components.md) that we need to add to the repo.

lalitb:
[...]

PR #689, opened 2021-04-21
Add TraceLogging Dynamic header for ETW exporter (optional MIT-licensed header)

maxgolov:
Related to #326 

## Changes

**TraceLogging Dynamic for Windows** is a header-only C++ library that allows emitting so-called "Dynamic Manifest" Tracing Events to remote out-of-proc ETW listener. This is a code donated by Microsoft as Open Source under MIT license. Functional description and High-level API document [available here](https://github.com/microsoft/tracelogging-dynamic-windows/blob/main/README.md).

**So what is it exactly?**

Nearest closest analog of similar concept is `go-winio/etw` library [available here](https://github.com/microsoft/go-winio/tree/master/pkg/etw).. Short recap, is that applications may emit strongly typed key-value pairs (or structs) events, that are efficiently passed via kernel buffers to out-of-proc listener over [Event Tracing for Windows](https://docs.microsoft.com/en-us/windows/win32/etw/event-tracing-portal) channel.  Event Tracing for Windows (ETW) provides application programmers the ability to start and stop event tracing sessions, instrument an application to provide trace events, and consume trace events. Trace events contain an event header and provider-defined data that describes the current state of an application or operation. You can use the events to debug an application and perform capacity and performance analysis.

**How does it benefit Open Source community like OpenTelemetry??**

Developers that by their own choice write and deploy their programs, services, processes on Windows OS - may now utilize OpenTelemetry C++ SDK with ETW exporter, AND via `TraceLogging Dynamic for Windows` library - liberate their tracing events and forward these to vendor-agnostic flows, such as:

- [Microsoft.Diagnostics.EventFlow](https://github.com/Azure/diagnostics-eventflow) - The EventFlow library suite allows applications to define what diagnostics data to collect, and where they should be outputted to in a **vendor-agnostic manner**. Diagnostics data can be anything from performance counters to application traces.... with outputs such as StdOutput (console output), HTTP (json via http), Application Insights, Azure EventHub, Elasticsearch, Azure Monitor Logs. As well as - Dynatrace output, Google Big Query output, SQL Server output, ReflectInsight output, Splunk output.

- **other out-of-proc ETW listeners**. It is possible to listen to ETW events, for example, using [.NET TraceEventLibrary](https://github.com/microsoft/perfview/blob/main/documentation/TraceEvent/TraceEventLibrary.md) . Or using [FireEye SilkETW](https://github.com/fireeye/SilkETW), which itself also provides a number of 3rd party options, such as [Elastic Output](https://www.elastic.co/). Good article about [SilkETW](https://www.fireeye.com/blog/threat-research/2019/03/silketw-because-free-telemetry-is-free.html)

**Example instrumentation**

Example instrumentation is already available [here](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/exporters/etw). What was previously missing is the ability to emit events in TraceLogging Dynamic format, that is compatible with all existing ETW tooling, such as Windows Performance Recorder, PerfView, GoogleUIforETW, etc.

Now by utilizing OpenTelemetry C++ API - developers may instrument their code in a **platform-agnostic way**. Cross-platform C++ code developers no longer need to call any of the Win32 TraceLogging APIs directly. If they run on Windows - they can direct their telemetry to Event Tracing for Windows facility easily. If they run on Linux - they can use some other OpenTelemetry exporter (e.g. `OTLP` or `fluent` or alike). Instrumented code remains cross-plat, invoking OpenTelemetry API calls only.

**Integration Options**

There are two options:

- **Add the Microsoft MIT-licensed code of TraceLogging Dynamic library as git submodule**. This is in alignment with other libraries, such as we consume `nlohmann/json` (also MIT-licensed) as submodule; OR

- **Add the single header into ETW exporter directory**. Because it is only necessary for builds on Windows with ETW exporter enabled. This could have saved us from hassle of yet another recursive cloning. We could have covered this in a separate OSS Licenses markdown doc, such as `If you are consuming OpenTelemetry C++ SDK with ETW exporter, then there is a separate library required, that is developed by Microsoft and is published under MIT license`. This is, in essence, identical to `nlohmann/json` consumption model, where we require it for zPages feature.. However, in `json.hpp` case - we consume it via git submodule.

~~I am presently implementing the 1st option. But if there is no community push-back to drop-in the header into ETW exporter, I'd switch to the 2nd option - by copying the header into ETW exporter directory; and adding a separate `LICENSE` and `README` file that explains when that module is going to be compiled in.~~

**UPDATE**: _discussed below and I'm implementing the 2nd option - dropping in the header file instead of adding dependency on remote submodule_.

Inclusion of this header also enables other programming language to project on top of OpenTelemetry C++ SDK for their own ETW tracing needs.


maxgolov:
@jsuereth @TomRoSystems @seemk - I'd like to get some feedback from non-Microsoft folks, to ask you if it is acceptable to add MIT-licensed code as submodule (as we do for `json.hpp` library); OR it's fine to drop a copy of it in this repo, under ETW exporter with a separate `LICENSE` file. I have not thought thru the Bazel build aspects. It seems like it would have been easier if we dropped the file in there. This code is HIGHLY UNLIKELY to change within the next few years. It is already doing what it is supposed to do (passing binary-packed events to out-of-proc listener), and the protocol how it's done will not change in years. So I do not anticipate any churn here. I am adding it as submodule, as it seems like in alignment with what we did for other similarly licensed modules. But my personal preference would be to just drop it in as a copy (thus, one less submodule repo to clone, slightly faster checkouts, and possibly easier integration with Bazel).

seemk:
I have nothing against the MIT license as they are pretty similar. However, my personal preference (not forcing it on anyone) is not to use submodules and would prefer the file is a dropin. :slightly_smiling_face:

jsuereth:
I only have two comments:

1. If we pull it in, we should make sure there's a working Bazel build for it.  Given it's windows-only we likely want to upgrade to Bazel 4.0 (and finally use https://docs.bazel.build/versions/4.0.0/platforms.html#skipping-incompatible-targets)
2. I think MIT licensed things are ok as dependencies (and that's consistent in other SDKs).

maxgolov:
@seemk @jsuereth - I can drop-in the file instead. It's MIT-licensed, Windows-only, only needed if customers are building with `WITH_ETW` build option. So those who don't want to refer to that license / file, they can exclude it from the build. I also prefer to drop it in instead of submodule. It'll make Bazel build easy. I think we actually had most of that stuff already sorted out, so relative include from ETW exporter would work great... so based on your feedback, I'll switch from submodule to drop-in, and I'll add a `LICENSE` file next to the header.

Also on related note - I will document the summary of all modules+licenses in a separate documentation PR while working on #683

maxgolov:
Updated the PR to drop-in the header, and updated the ETW exporter documentation to mention the license and original source repo where this header was mirrored from.

lalitb (approved):
As per [this ](https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#copyright-notices) section :
`Any contributed third-party code must originally be Apache 2.0-Licensed or must carry a permisive software license that is compatible when combining with Apache 2.0 License. At this moment, BSD and MIT are the only OSI-approved licenses known to be compatible.`

We should be good to include the MIT license code ( though it doesn't talk about copyright explicitly ).

ThomsonTan (approved):

seemk (approved):

PR #692, opened 2021-04-22
Support for multiple processors

### LINKED ISSUE #664: Allow multiple SpanProcessors per TracerProvider
The [Specification states](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor):

- Span processors can be registered directly on SDK TracerProvider and they are invoked in the same order as they were registered.
- Each processor registered on TracerProvider is a start of pipeline that consist of span processor and optional exporter. SDK MUST allow to end each pipeline with individual exporter.

This is a feature request to add a `MultiSpanProcessor` or `AggregateSpanProcessor` which can be used to register more than one pipeline with the SDK for processing spans.

An example (abandoned) pr: #590

lalitb:
Fixes #664

## Changes

- `MultiSpanProcessor` - Implements `SpanProcessor` and  internally stores configured `SpanProcessors`. 
- `MultiRecordable` - implements the Recordable interface, and encapsulates several regular recordables.
- `TracerContext` internally maintains the `MultiSpanProcessor` instance.
- `TracerContext` constructor accepts vector of `SpanProcessors` to be configured and provide methods to add new `SpanProcessors`.  SpanProcessor once configured cannot be deleted.

Example:
(Using `TracerProvider` ):

```cpp
  auto exporter1 = std::unique_ptr<sdktrace::SpanExporter>(new OStreamSpanExporter);
  auto processor1 = std::unique_ptr<sdktrace::SpanProcessor>(new sdktrace::SimpleSpanProcessor(std::move(exporter1)));

  auto exporter2 = std::unique_ptr<sdktrace::SpanExporter>(new InMemorySpanExporter());
  auto processor2 = std::unique_ptr<sdktrace::SpanProcessor>(new sdktrace::SimpleSpanProcessor(std::move(exporter2)));

  auto provider = nostd::shared_ptr<sdk::trace::TracerProvider>(new sdktrace::TracerProvider(std::move(processor1));
  provider->AddProcessor(std::move(processor2));
  // Set the global trace provider
  opentelemetry::trace::Provider::SetTracerProvider(std::move(provider));
```

(Using `TracerContext` ):

```cpp
  auto exporter1 = std::unique_ptr<sdktrace::SpanExporter>(new OStreamSpanExporter);
  auto processor1 = std::unique_ptr<sdktrace::SpanProcessor>(new sdktrace::SimpleSpanProcessor(std::move(exporter1)));

  auto exporter2 = std::unique_ptr<sdktrace::SpanExporter>(new InMemorySpanExporter());
  auto processor2 = std::unique_ptr<sdktrace::SpanProcessor>(new sdktrace::SimpleSpanProcessor(std::move(exporter2)));

  std::vector<std::unique_ptr<sdktrace::SpanProcessor>> processors;
  processors.push_back(std::move(processor1));
  processors.push_back(std::move(processor2));
  auto context = std::make_shared<sdktrace::TracerContext>(std::move(processors));
  auto provider = nostd::shared_ptr<opentelemetry::trace::TracerProvider>(new sdktrace::TracerProvider(context));

  // Set the global trace provider
  opentelemetry::trace::Provider::SetTracerProvider(provider);
```


examples/http/tracer_common.hpp
```diff
@@ -4,7 +4,7 @@
 #include "opentelemetry/sdk/trace/tracer_provider.h"
 #include "opentelemetry/trace/provider.h"
 
-#include <iostream>
+#include<vector>
```
reyang, line 7:
```suggestion
#include <vector>
```

sdk/include/opentelemetry/sdk/trace/multi_span_processor.h
```diff
@@ -0,0 +1,116 @@
...
+    for (auto &processor : processors)
+    {
+      if (processor)
+      {
+        processors_.push_back(std::move(processor));
+      }
+    }
+  }
+
+  void AddProcessor(std::unique_ptr<SpanProcessor> &&processor)
```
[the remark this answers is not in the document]
lalitb: Good comment. My earlier code did have mutex guard against all vector operations in this class, seems somehow missed out check-in it The doubly linked list looks like a good option too, would be good to do some benchmarking to select the optimal one of two.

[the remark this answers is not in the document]
reyang: I can share the thinking from the .NET implementation:

STL vector + mutex would give the ultimate thread-safety in multi-reader-multi-writer situation, the potential downside is that it will introduce contention to the readers (in this case, the readers are the threads that push spans to the processor chain), which might something to avoid.

Doubly linked list only gives single-writer-multi-reader thread-safety, however it is well aligned with the general .NET API guideline (that if a class method doesn't specify thread safety, by default it is not supposed to be called concurrently).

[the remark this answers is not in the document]
lalitb: Ok. now I realize why I removed the mutex guard - as `MultiSpanProcessor` is an atomic instance in `TracerContext` :

https://github.com/open-telemetry/opentelemetry-cpp/blob/f10acac2fe708635dfb0b48760bc598270b832b9/sdk/include/opentelemetry/sdk/trace/tracer_context.h#L86

And so multiple threads shouldn't be able to access it simultaneously. But will wait for comments on others :)

examples/multi_processor/README.md
```diff
@@ -0,0 +1,12 @@
+
+# Simple Trace Example
```
ThomsonTan, line 2: Remove empty first line and change the title to indicate this is for multiple processors?

examples/http/tracer_common.hpp
```diff
@@ -13,8 +13,10 @@ void initTracer() {
       new opentelemetry::exporter::trace::OStreamSpanExporter);
   auto processor = std::unique_ptr<sdktrace::SpanProcessor>(
       new sdktrace::SimpleSpanProcessor(std::move(exporter)));
+  std::vector<std::unique_ptr<sdktrace::SpanProcessor>> processors;
+  processors.push_back(std::move(processor));
   // Default is an always-on sampler.
-  auto context = std::make_shared<sdktrace::TracerContext>(std::move(processor));
+  auto context = std::make_shared<sdktrace::TracerContext>(std::move(processors));
```
jsuereth, line 19: Should there be a convenience constructor that just takes on processor? 

Alternatively, a static helper method to make TracerProvider from a processor?

jsuereth (commented):
Method name improvements LGTM for Processor / AddProcessor.

+1 to Reiley's comments on thread safety, will re-review when that's updated, but like this approach!

examples/http/tracer_common.hpp
reyang: +1.

This is what OTel.NET does https://github.com/open-telemetry/opentelemetry-dotnet/blob/a25741030f05c60c85be102ce7c33f3899290d49/docs/trace/extending-the-sdk/Program.cs#L30.

reyang: Consider taking a look at the other language clients that reached 1.0 stable and see what's the best practice.
One more example from Python https://opentelemetry-python.readthedocs.io/en/latest/getting-started.html.

lalitb: This example was using `TracerContext`. This is how it looks like if we directly use `TracerProvider` to add processors:

```cpp
  auto provider = nostd::shared_ptr<opentelemetry::trace::TracerProvider>(std:move(processor1));
  provider.AddProcessor(std::move(processor2));
  provider.AddProcessor(std::move(processor3));

  opentelemetry::trace::Provider::SetTracerProvider(provider);

```

lalitb:
> Method name improvements LGTM for Processor / AddProcessor.
> 
> +1 to Reiley's comments on thread safety, will re-review when that's updated, but like this approach!

As suggested by @reyang , the following changes are done
  - Modify `sdk::trace::MultiSpanProcessor` class to store processors in a list instead of vector.
  - Remove atomicity from TracerContext::processor_ ( i.e change it's type from  `opentelemetry::sdk::common::AtomicUniquePtr<SpanProcessor>` to `std::unique_ptr<SpanProcessor>`. This is because the list structure in `MultiSpanProcessor` will take care of single write, multiple read scenario for processors.
  -  `TracerProvider::AddProcessor()` and `TracerContext::AddProcessor()` methods are no longer thread safe, and have added that as comment for them.
  - No change but to mention: We don't support removing processors,

examples/http/tracer_common.hpp
lalitb: I have modified the example NOT to use `TracerContext`, instead directly use `TracerProvider` to pass processor(s) as an argument. `TracerProvider` constructor takes either of single processor, or vector of processors as an argument.

examples/multi_processor/README.md
lalitb: Good point, have changed the title, and the description too.

reyang (approved):
LGTM.

examples/multi_processor/main.cc
```diff
@@ -0,0 +1,79 @@
...
+{
+  auto exporter1 = std::unique_ptr<sdktrace::SpanExporter>(
+      new opentelemetry::exporter::trace::OStreamSpanExporter);
+  auto processor1 = std::unique_ptr<sdktrace::SpanProcessor>(
+      new sdktrace::SimpleSpanProcessor(std::move(exporter1)));
+
+  auto exporter2 = std::unique_ptr<sdktrace::SpanExporter>(new InMemorySpanExporter());
+
+  // fetch the exporter for dumping data later
+  memory_span_exporter = dynamic_cast<InMemorySpanExporter *>(exporter2.get());
```
ThomsonTan, line 27: Assign `new InMemorySpanExporter()` to memory_span_exporter` directory before creating unique_ptr, so avoid this `dynamic_cast`?

examples/multi_processor/README.md
```diff
@@ -0,0 +1,14 @@
+# Multi Processor Trace Example
+
+In this example, the application in `main.cc` initializes and registers a tracer
+provider from the [OpenTelemetry
+SDK](https://github.com/open-telemetry/opentelemetry-cpp). The TraceProvider is conneected
```
ThomsonTan, line 5:
```suggestion
SDK](https://github.com/open-telemetry/opentelemetry-cpp). The `TracerProvider` is connected
```

sdk/include/opentelemetry/sdk/trace/multi_recordable.h
```diff
@@ -0,0 +1,141 @@
...
+  void AddRecordable(const SpanProcessor &processor,
+                     std::unique_ptr<Recordable> recordable) noexcept
+  {
+    recordables_[MakeKey(processor)] = std::move(recordable);
+  }
+
+  const std::unique_ptr<Recordable> &GetRecordable(const SpanProcessor &processor) const noexcept
+  {
+    // TODO - return nullptr ref on failed lookup?
+    static std::unique_ptr<Recordable> empty(nullptr);
```
ThomsonTan, line 36: nit: this empty could be moved to the line right above `return empty`? We don't need initialize this value if recordable is found.

ThomsonTan: Or we could make the `empty` variable as class static as it is also useful in `ReleaseRecordable`?

ThomsonTan (approved):

examples/multi_processor/main.cc
lalitb: Good point. Yeah, it can be done, but I felt it's more important to have `exporter2` created the way the application developer would create, as `exporter2` needs to be added to `TracerProvider` afterward.  The statement:

`auto exporter2 = std::unique_ptr<sdktrace::SpanExporter>(memory_span_exporter );`

would be confusing for application devs as they would never create an exporter this way. Other than that, I don't have any issue in changing the code as suggested :)

sdk/include/opentelemetry/sdk/trace/multi_recordable.h
lalitb: > this empty could be moved to the line right above return empty? We don't need initialize this value if recordable is found. 

done

> Or we could make the empty variable as class static as it is also useful in ReleaseRecordable?

That won't help as ReleaseRecordable needs to return unique_ptr, not it's a reference so class static can't be used. But control would never reach both places as we don't support removing processors once configured.

ThomsonTan (approved):

lalitb:
Will see if there are comments from @jsuereth, @pyohannes, and others before closing it,

sdk/include/opentelemetry/sdk/trace/multi_recordable.h
```diff
@@ -0,0 +1,141 @@
...
+public:
+  void AddRecordable(const SpanProcessor &processor,
+                     std::unique_ptr<Recordable> recordable) noexcept
+  {
+    recordables_[MakeKey(processor)] = std::move(recordable);
+  }
+
+  const std::unique_ptr<Recordable> &GetRecordable(const SpanProcessor &processor) const noexcept
+  {
+    // TODO - return nullptr ref on failed lookup?
```
jsuereth, line 35: Nit: remove the TODO

sdk/include/opentelemetry/sdk/trace/multi_span_processor.h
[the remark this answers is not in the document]
jsuereth: Atomic just means that you can compare-and-swap the pointer.  Users can still access it simultaneously (after reading the pointer in memory).

*if* you have `AddProcessor` return a new MultiProcessor instance and compare-and-swap the Atomic Ptr then I think you'd be ok, but as is, you still have a (possible) threading issue if more than thread calls AddProcessor.  It's unlikely but still possible.

jsuereth (changes requested):
I think there's still a threading issue around AddProcessor.

You could possible solve this via fully-copying the processors array in the "AddProcessor" method of TracerContext (and then use a compare-and-swap to flip instances) meaning no threads would have an inconsistent processor.   However, the code, as it stands, has a very unlikely threading issue.

lalitb:
> I think there's still a threading issue around AddProcessor.
> 
> You could possible solve this via fully-copying the processors array in the "AddProcessor" method of TracerContext (and then use a compare-and-swap to flip instances) meaning no threads would have an inconsistent processor. However, the code, as it stands, has a very unlikely threading issue.

Thank you @jsuereth for the comments. Just to have a better understanding of the issue -  the current approach of switching to the linked-list structure still has race condition around `AddProcessor`,  and another approach would be moving back to vector, and do full-copying followed by flip?

sdk/include/opentelemetry/sdk/trace/multi_span_processor.h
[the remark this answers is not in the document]
lalitb: OK, I think I got your point, in that case, we should have Atomic Ptr for Span Processor, and create new instance of MultiProcessor everytime AddProcessor() is called. It is already documented that AddProcessor() is not thread safe, and shouldn't be called simultaneously.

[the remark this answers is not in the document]
jsuereth: Ah, I missed that AddProcessor was marked not thread-safe as well.   This means my point is not important.

If the changes to make AddProcessor thread-safe aren't too painful, I think it's worth it.  I think it's a low-frequency method call, so it's ok to be "expensive" to AddProcessor as long as the "read" hot path is fast (which you already have AFAICT).   Removing my requested changes.  I'd prefer to see AddProcessor thread-safe but this is fine as-is.

jsuereth (approved):

jsuereth:
@lalitb 

> Thank you @jsuereth for the comments. Just to have a better understanding of the issue - the current approach of switching to the linked-list structure still has race condition around AddProcessor, and another approach would be moving back to vector, and do full-copying followed by flip?

Yeah, I think you have two options:

1. In AddProcessor, just do a big compare-and-swap of the whole (immutable-ish) Processor.   This should have the advantage that reads are fast, writes are consistent and it should be less work overall.
2. The double-linked list approach could work, but you likely need to do some memory-barriers / compare-and-swap routines on top of the code you already have to ensure no one views the linked list in an inconsistent state (which, having two pointers, means there's a slight chance).

reyang:
> Ah, I missed that AddProcessor was marked not thread-safe as well.   This means my point is not important.
> 
> If the changes to make AddProcessor thread-safe aren't too painful, I think it's worth it.  I think it's a low-frequency method call, so it's ok to be "expensive" to AddProcessor as long as the "read" hot path is fast (which you already have AFAICT).   Removing my requested changes.  I'd prefer to see AddProcessor thread-safe but this is fine as-is.

FYI - a slightly different perspective for consideration, in C# we chose to make AddProcessor not thread safe (see my explanation [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/692#discussion_r619406393)) for couple reasons:

1. AddProcessor has side effect since processors are meant to be called in FIFO sequence. Having multiple threads adding processors simultaneously is normally a indication of application bug as the sequence is undeterministic.
2. In C# the convention is that member functions are not thread-safe unless explicitly specified. I think C++ STL follows the same convention.

lalitb:
> 1. AddProcessor has side effect since processors are meant to be called in FIFO sequence. Having multiple threads adding processors simultaneously is normally an indication of application bug as the sequence is undeterministic.

That's a logical reason to enforce non-thread-safe behavior.

lalitb:
@reyang @jsuereth  - I am merging it in the double linked-list-based approach based on the discussions here.  Atm ptr is not used as AddProcessor() is not thread-safe. Will make it explicitly in sdk document.  Also will create a ticket to discuss and revisit its design in the future if we need to.

PR #693, opened 2021-04-22
Add InstrumentationLibrary to Tracer

### LINKED ISSUE #567: Implement InstrumentationLibrary Instance in SDK

ThomsonTan:
Fixes #567 and #694.

## Changes

Add `InstrumentationLibrary` to SDK and return a new tracer for a different `InstruementationLibrary`.


sdk/src/trace/tracer_provider.cc
```diff
@@ -23,6 +23,9 @@ opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer> TracerProvider::G
     nostd::string_view library_name,
     nostd::string_view library_version) noexcept
 {
+  tracer_->SetInstrumentationLibrary(
+      opentelemetry::instrumentationlibrary::InstrumentationLibrary::create(library_name,
```
lalitb, line 27: Each tracer is identified by `library_name`. And `TracerProvider` should create separate instances of tracer for each library_name as input to `TracerProvider::GetTracer`. And further returns the same tracer if it already exists for input `library_name`. Default (non-no-op) tracer should be created if the `library_name` is empty.  This can be achieved by `TracerProvider` maintaining a list of tracers (mapped with their `library_name`).

Though not related to this PR. Will create a separate ticket for that.

ThomsonTan: Yes, I was thinking on this but thought it may be good to do it in a separate PR, as currently `TracerProvider` contains only 1 `Tracer`. Opened below ticket and I'll look into it.

https://github.com/open-telemetry/opentelemetry-cpp/issues/694

jsuereth: +1 to solving this.

jsuereth (approved):
Change looks good assuming two things:

1. We create new `Tracer`s for different libraries/versions (you have #694) for that.
2. We fill out `InstrumentationLibrary` in the OTLP exporter.   Is there a bug open to track that work or do we need to open one?

lalitb:
> 2\. We fill out `InstrumentationLibrary` in the OTLP exporter.   Is there a bug open to track that work or do we need to open one?

Created #695  for this.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -162,6 +166,25 @@ class Tracer
...
+  InstrumentationLibrary *GetInstrumentationLibrary() const noexcept
+  {
+    return instrumentation_library_.get();
+  }
+
+  /**
+   * Set instrumentation library.
+   * @param instrumentation library to set on the current tracer.
+   */
+  void SetInstrumentationLibrary(
```
lalitb, line 182: As discussed, it should be possible to remove this method and instead pass `InstrumentationLibrary` instance in sdk::tracer constructor through sdk::tracerprovider::gettracer() . This would further need move creation of tracer instance in sdk::tracerprovider::gettracer() method.

ThomsonTan:
> Change looks good assuming two things:
> 
> 1. We create new `Tracer`s for different libraries/versions (you have #694) for that.
> 2. We fill out `InstrumentationLibrary` in the OTLP exporter.   Is there a bug open to track that work or do we need to open one?

Create new `Tracer` is also included in this PR (fixed #694). I'll work on filling out `InstrumentationLibrary` as a separate ticket (#695). We need populate this info to most exporters like OTLP, Zipkin, Jaeger, etc.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -162,6 +166,25 @@ class Tracer
     }
   }
 
+  /**
+   * Get the instrumentation library associated with the current tracer.
+   * @return the instrumentation library for current tracer.
+   */
+  InstrumentationLibrary *GetInstrumentationLibrary() const noexcept
```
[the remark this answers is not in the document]
ThomsonTan: Done.

sdk/src/trace/tracer_provider.cc
```diff
@@ -19,11 +19,28 @@ TracerProvider::TracerProvider(std::unique_ptr<SpanProcessor> processor,
...
     nostd::string_view library_version) noexcept
 {
-  return opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer>(tracer_);
+  // if (library_name == "") {
+  //   // TODO: log invalid name.
+  // }
+
+  const std::lock_guard<std::mutex> guard(lock_);
+
+  auto lib = InstrumentationLibrary::create(library_name, library_version);
```
seemk, line 32: If `GetTracer` were to be used in an instrumentation to retrieve a tracer when creating spans, this potentially creates 3 allocations and deallocs on a hot path.

Would it make sense to avoid heap allocations here for the usual case when a tracer already exists?

I mean using `GetTracer()` every time could perhaps be considered bad use of the API as opposed to caching a tracer in the instrumentation, but I think it might be too easy too misuse the API and just default to `GetTracer()`

ThomsonTan: @seemk thanks. I think this is a valid concern. I updated the PR and delayed the heap allocation to the time when creating new `Tracer` is necessary.

sdk/src/trace/tracer_provider.cc
```diff
@@ -19,11 +19,29 @@ TracerProvider::TracerProvider(std::unique_ptr<SpanProcessor> processor,
...
     nostd::string_view library_version) noexcept
 {
-  return opentelemetry::nostd::shared_ptr<opentelemetry::trace::Tracer>(tracer_);
+  // if (library_name == "") {
+  //   // TODO: log invalid library_name.
+  // }
+
+  const std::lock_guard<std::mutex> guard(lock_);
+
+  for (auto &tracer : tracers_)
```
lalitb, line 32: Nit: Is it somehow possible to avoid this search loop by using a map instead of a vector say something like:
std::map<std::pair<name, version>, tracer> tracers_;

lalitb (approved):
LGTM. Thanks for adding this feature.

seemk (approved):

sdk/src/trace/tracer_provider.cc
ThomsonTan: I assumed there would not be too many tracers in one process (like less than a hundred) so linear scan is fine compared to the hash or sort overhead. But we could discuss this a little more.

lalitb: Yes, I too think so. If it's not a major performance benefit with std::map, and it adds more complexity to current design, we should be using vector.

PR #699, opened 2021-04-27
Visual Studio 2019 fixes for API headers

maxgolov:
## Changes

There are three fundamental changes here:

- Better detection of compiler features in Visual Studio 2019. For example, you can't just assume that if `<span>` exists, then the feature is available. It is shielded beyond `c++latest` flag, so have to check for that flag too. That way if you build with Visual Studio 2019 IDE and latest headers, but using `c++17` flag - the code would still compile successfully.

- compiler error fix in case if `__cplusplus` has incorrect value. This is long-standing issue, that even [has its own Reddit thread here](https://www.reddit.com/r/cpp/comments/bqsyss/zc_cplusplus_by_default/) ... We solve this in CMake and Bazel builds by defining corresponding build flag (`/Zc:__cplusplus`). But other build systems and libraries may not have that flag set. This presents a challenge when they include our API headers. Thus, it'd be best to use smarter compiler feature logic for MSVC, based off MSVC-specific compiler language detection.

- improve handling of `HAVE_CPP_STDLIB` build option: if we are building with `c++17` and use `std::variant`, we can still fallback to use `nostd::span` implementation that comes with OpenTelemetry. This happens if you are building with Visual Studio 2019 with `c++17` standard and not with the latest. That is when you would prefer to use `std::variant` (neither Abseil nor nostd), but you do not want to use `gsl::span` due to whatever reason.

This is a short summary. I will cover user-focused summary: build flavors, options, when one can use `HAVE_CPP_STDLIB`, and when one gets the `std::span` versus `gsl::span` versus `nostd::span` (e.g. Abseil Span) in a separate documentation-only PR.

### DIFF
--- a/api/include/opentelemetry/std/span.h
+++ b/api/include/opentelemetry/std/span.h
@@ -33,7 +33,7 @@
 #    endif
 #    // Check for other compiler span implementation
 #    if !defined(_MSVC_VALUE) && __has_include(<span>)
-       // This works as long as compiler standard is set to C++20
+// This works as long as compiler standard is set to C++20
 #      define HAVE_SPAN
 #    endif
 #  endif
@@ -51,7 +51,7 @@ template <class ElementType, std::size_t Extent = gsl::dynamic_extent>
 using span = gsl::span<ElementType, Extent>;
 }
 OPENTELEMETRY_END_NAMESPACE
-#  define HAVE_SPAN
+#    define HAVE_SPAN
 #  else
 // No `gsl::span`, no `std::span`, fallback to `nostd::span`
 #  endif
--- a/api/include/opentelemetry/std/span.h
+++ b/api/include/opentelemetry/std/span.h
@@ -26,13 +26,16 @@
 #      define HAVE_SPAN
 #    endif
 #  endif
-#  if __has_include(<span>) && !defined(HAVE_SPAN)  // Check for span
-#    define HAVE_SPAN
-#  endif
-#  if !__has_include(<string_view>)  // Check for string_view
-#    error \
-        "STL library does not support std::span. Possible solution:"                   \
-         " - #undef HAVE_CPP_STDLIB // to use OpenTelemetry nostd::string_view"
+#  if !defined(HAVE_SPAN)
+#    // Check for Visual Studio span
+#    if defined(_MSVC_VALUE) && _HAS_CXX20
+#      define HAVE_SPAN
+#    endif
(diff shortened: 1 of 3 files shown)


api/include/opentelemetry/std/span.h
```diff
@@ -26,13 +26,16 @@
...
-#  if __has_include(<span>) && !defined(HAVE_SPAN)  // Check for span
-#    define HAVE_SPAN
-#  endif
-#  if !__has_include(<string_view>)  // Check for string_view
-#    error \
-        "STL library does not support std::span. Possible solution:"                   \
-         " - #undef HAVE_CPP_STDLIB // to use OpenTelemetry nostd::string_view"
+#  if !defined(HAVE_SPAN)
+#    // Check for Visual Studio span
+#    if defined(_MSVC_VALUE) && _HAS_CXX20
```
ThomsonTan, line 31: Where is `_MSVC_VALUE` defined?

maxgolov: `_MSVC_VALUE` is a predefined macro added since Visual Studio 2017+ and is [described here](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160)

`_MSVC_LANG Defined as an integer literal that specifies the C++ language standard targeted by the compiler. It's set only in code compiled as C++. The macro is the integer literal value 201402L by default, or when the /std:c++14 compiler option is specified. The macro is set to 201703L if the /std:c++17 compiler option is specified. It's set to a higher, unspecified value when the /std:c++latest option is specified. Otherwise, the macro is undefined. The _MSVC_LANG macro and /std (Specify Language Standard Version) compiler options are available beginning in Visual Studio 2015 Update 3.` ... 

Based on my understanding, we can check that macro is defined. Means we may get native `std::span` support. And if it is defined (means we are running with at least Visual Studio 2017+), we subsequently check if `_HAS_CXX20=1`, if the build tools have support for `c++20`, and only then we may `#include <span>`. If we don't do that check and simply include the `<span>` if it exists, then msvc compiler fails with an error.

maxgolov: Good catch! It was one of those moments. Fixed it.

maxgolov:
Please ignore `codecov/project — 94.80% (-0.02%) compared to 8ea888a`

lalitb (approved):
LGTM

ThomsonTan (approved):

PR #700, opened 2021-04-27
Populate InstrumentationLibrary to exporters

### LINKED ISSUE #695: Propagate instrumentation library information to exporters
Instrumentation library information -  name and version - should be propagated to exporters. 
This is dependent on #567, once we have InstrumentationLibrary instance created for each tracer.

ThomsonTan:
Fixes #695 

## Changes

Added `SetInstrumentationLibrary` to `Recordable` interface which is used to populate instrumentation library to exporters.

### DIFF
--- a/exporters/zipkin/src/recordable.cc
+++ b/exporters/zipkin/src/recordable.cc
@@ -216,11 +216,11 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept
 }
 
 void Recordable::SetInstrumentationLibrary(
-    std::shared_ptr<const opentelemetry::sdk::instrumentationlibrary::InstrumentationLibrary>
-        &&instrumentation_library) noexcept
+    const opentelemetry::sdk::instrumentationlibrary::InstrumentationLibrary
+        &instrumentation_library) noexcept
 {
-  span_["tags"]["otel.library.name"]    = instrumentation_library->GetName();
-  span_["tags"]["otel.library.version"] = instrumentation_library->GetVersion();
+  span_["tags"]["otel.library.name"]    = instrumentation_library.GetName();
+  span_["tags"]["otel.library.version"] = instrumentation_library.GetVersion();
 }
 
 void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept {}
--- a/exporters/zipkin/src/recordable.cc
+++ b/exporters/zipkin/src/recordable.cc
@@ -219,8 +219,8 @@ void Recordable::SetInstrumentationLibrary(
     std::shared_ptr<const opentelemetry::sdk::instrumentationlibrary::InstrumentationLibrary>
         &&instrumentation_library) noexcept
 {
-  span_["otel.library.name"]    = instrumentation_library->GetName();
-  span_["otel.library.version"] = instrumentation_library->GetVersion();
+  span_["tags"]["otel.library.name"]    = instrumentation_library->GetName();
+  span_["tags"]["otel.library.version"] = instrumentation_library->GetVersion();
 }
 
 void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept {}
--- a/exporters/zipkin/src/recordable.cc
+++ b/exporters/zipkin/src/recordable.cc
@@ -215,6 +215,14 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept
   span_["duration"] = duration.count();
 }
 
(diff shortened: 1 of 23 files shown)


exporters/zipkin/src/recordable.cc
```diff
@@ -215,6 +215,14 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept
   span_["duration"] = duration.count();
 }
 
+void Recordable::SetInstrumentationLibrary(
+    std::shared_ptr<const opentelemetry::sdk::instrumentationlibrary::InstrumentationLibrary>
+        &&instrumentation_library) noexcept
+{
+  span_["otel.library.name"]    = instrumentation_library->GetName();
+  span_["otel.library.version"] = instrumentation_library->GetVersion();
```
lalitb, line 223: As per the [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#instrumentationlibrary)these should be reported in `tags`:

`OpenTelemetry Span's InstrumentationLibrary MUST be reported as tags to Zipkin using the following mapping.`

Similar to what is done for status:
https://github.com/open-telemetry/opentelemetry-cpp/blob/516d02d8bc19f778731d58c7162d16e08340207a/exporters/zipkin/src/recordable.cc#L194

sdk/include/opentelemetry/sdk/trace/recordable.h
```diff
@@ -125,6 +129,13 @@ class Recordable
    * @param duration the duration to set
    */
   virtual void SetDuration(std::chrono::nanoseconds duration) noexcept = 0;
+
+  /**
+   * Set the instrumentation library of the span.
+   * @param instrumentation_library the instrumentation library to set
+   */
+  virtual void SetInstrumentationLibrary(
```
lalitb, line 137: Thinking out loud -  should we be instead passing nostd::string_view for `name` and `version`  - to ensure that exporter(s) won't store this shared_ptr reference locally?

exporters/zipkin/src/recordable.cc
ThomsonTan: Thanks for the catch. I misread the `tags` part in the spec. Fixed it.

sdk/include/opentelemetry/sdk/trace/recordable.h
ThomsonTan: Yes, this is a good thought. I was trying to avoid dispersing the internal structure of `InstruemntationLibrary` to places other than the class itself as the [spec](https://github.com/open-telemetry/oteps/blob/main/text/0083-component.md#future-possibilities) mentioned it could be expected in future.

I tried define the pointer as `std::wake_ptr` in the exporter to avoid taking ownership. I could make this change if it looks good to you.

lalitb: If we want to keep the structure intact, it would be much easier to pass const reference to  `InstruemntationLibrary`, as Tracer::GetInstrumentationLibrary() already returns that.

ThomsonTan: Take const reference could cause use after free if the tracer provider is destructed before the exporter, or the exporter needs to make a copy of the `InstrumentationLibrary` which I'd like to avoid.

lalitb: I don't think that would happen. As the shutdown is propagated from TracerProvider -> Processor -> Exporter.

ThomsonTan: This looks reasonable. Could there be case like the exporter runs asynchronously then it is still working after the call of `TracerProvider::Shutdown`?

Updated the PR to pass const reference of `InstrumentationLibrary` to exporters.

lalitb: That shouldn't happen if the exporter is implemented correctly. `Exporter::Shutdown()` should ensure all cleanups ( that includes pending/ongoing synchronous/asynchronous calls ) are completed before returning. And `TracerProvider::Shutdown()` would return only after.

ThomsonTan: But `TracerProvicer::Shutdown()` could return fail or timeout as the spec mentioned, will `TracerProvider` continue to destruct in these cases?

lalitb (approved):
LGTM. 
We can see if there are comments from @jsuereth, @pyohannes, and others before merging.

Issue #649, opened 2021-04-05
Proper handling of Span Links in ETW exporter


maxgolov:
There is one missing API path, where ETW exporter is not accepting Span Links properly.

This issue is reported to address the concern found during code review.

https://github.com/open-telemetry/opentelemetry-cpp/pull/519#discussion_r606348661

It is of _lower priority_ since only specific path is affected, the core functionality is operational.

maxgolov:
Also need to address the other couple minor feedbacks from #702

PR #702, opened 2021-04-27
ETW exporter warnings clean-up, compliance, implementing 2 TODOs

maxgolov:
This change is isolated to ETW exporter. Mostly compliance and warnings fixes.

## Changes

There are no _significant_ changes, only bugfix, plus adding missing features / resolving TODOs.

List of fixes:

- **PropertyValue Container**: perform `memcpy` of original `const char *` buffer into `std::string` -owning object because original buffer MAY refer to temporary. We can tune this later on if this extra `memcpy` ever becomes an issue. Favoring `API usability` and `stability` over perf at this point. Without this fix the `const char *` and `string literal` parameters may not be processed correctly in some scenarios.

- `arrays` - are still not supported yet by ETW exporter. Adding `UNREFERENCED_PARAMETER(vec);` to avoid the compilation L4 warning that may be a problem for envs where warnings are treated as errors. Will address the arrays support as feature work in a separate PR.

- disable warning `5054` when compiling with Visual Studio 2019. This warning is triggered in "external" project `TraceLoggingDynamic.h` header that I am not going to touch. The warning itself is benign and unique to 2019. Older compilers do not emit it. It is a benign [conformance warning](https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements) that has rather low practical value in our use case.

- remove aliases for `using Properties = exporter::etw::Properties; using PropertyType = exporter::etw::PropertyType;` - redeclaring these in another namespace was causing ambiguity for Visual Studio 2015 - Update 2. This does not functionally change anything, and it was not an issue for 2017 (`C++14`) and 2019 (`C++17`) that we have in the build loop. Thus, it wasn't immediately caught.

- **remove `HAVE_TLD` Feature Gate**: we now always include `TraceLoggingDynamic.h` that's been open-sourced and merged in another PR #689  (Yay!) . So the option becomes irrelevant.

- **Other L4 warnings clean-up**: when compiling without `HAVE_MSGPACK` - since (see above), TLD is always included, we add `UNREFERENCED_PARAMETER` macro to resolve L4 warnings. This code is presently Windows-only. And the macro is a standard macro for Visual Studio. Since the same code may be compiled with / without the feature, we have to preserve the parameter names. I could have used `std::ignore`, but `UNREFERENCED_PARAMETER` should be Okay - within the constraints of a typical Windows-only code (which ETW exporter is).

- **GetTracer API** : second parameter _used to_ specify the encoding type (MsgPack or TraceLoggingDynamic). Since this was NOT in alignment with spec - which suggests the second param is a `library_version`, I'm moving the encoding type to ETW Tracer Provider configuration option called `TracerProviderConfiguration::encoding` . That means now all Tracers obtained from a given ETW provider instance - use the same binary encoding (TLD as default). In most cases - since TLD is the better, preferred encoding anyways, we do not sacrifice too much (if anything). Whatever it takes to keep the API surface in alignment with the spec! Second parameter in case of ETW exporter is optional (ignored completely) since ETW exporter is `header-only`. We can discuss how to implement that parameter, should there be ever the need for inclusion of ETW exporter in a shared library. For now - this is unnecessary. All of it is lean and instrumentation may get compiled straight into customer code with no external DLL.

- **Status API**: implement `SetStatus` API.

- **Timestamps**: add option `HAVE_FIELD_TIME` that allows to stamp an additional time field encoded as text (alongside with standard ETW time that comes at envelope level); plus make time conversion routines static inline, don't export them; future-proof ISO8601 time formatting routine to include up to 6 digits after the dot instead of 3. This may help in a flow where a backend is expecting a format with 6 digits precision.

I will describe the new compile- and runtime- options, such as ETW `TracerProviderConfiguration` concepts, `HAVE_FIELD_TIME`, `ETW_FIELD_ENDTTIME`, etc. in a separate "documentation-only" examples + PR.


exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -104,6 +106,72 @@ static inline void GetOption(const TracerProviderOptions &options,
...
+        break;
+
+      case CONST_HASHCODE(TLD):
+        // nobrk
+      case CONST_HASHCODE(tld):
+        // nobrk
+        evtFmt = ETWProvider::EventFormat::ETW_MANIFEST;
+        break;
+
+      default:
```
ThomsonTan, line 166: Assert here instead of ignore silently?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -791,6 +881,27 @@ class Span : public trace::Span
...
 public:
+  /**
+   * @brief Update Properties object with current Span status
+   * @param evt
+   */
+  void UpdateStatus(Properties &evt)
+  {
+    /* Should we avoid populating this extra field if status is unset? */
+    if ((status_code_ == trace::StatusCode::kUnset) || (status_code_ == trace::StatusCode::kOk))
+    {
```
ThomsonTan, line 892: Why report `kUnset` as success, for other exporter like [Zipkin](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#status), the status is not reported for `kUnset`.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1033,44 +1147,8 @@ class TracerProvider : public trace::TracerProvider
   nostd::shared_ptr<trace::Tracer> GetTracer(nostd::string_view name,
                                              nostd::string_view args = "") override
```
ThomsonTan, line 1148: nit: rename the second argument as `span`?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -399,9 +473,18 @@ class Tracer : public trace::Tracer
...
+      // ETW has its own precise timestamp at envelope layer for every event.
+      // However, in some scenarios it is easier to deal with ISO8601 strings.
+      // In that case we convert the app-created timestamp and place it into
+      // Payload[$ETW_FIELD_TIME] field. The option configurable at compile-time.
+      evt[ETW_FIELD_ENDTTIME] = utils::formatUtcTimestampMsAsISO8601(endTimeMs);
+#endif
       // Duration of Span in milliseconds
       evt[ETW_FIELD_DURATION] = endTimeMs - startTimeMs;
+      // Presently we assume that all spans are server spans
+      evt[ETW_FIELD_SPAN_KIND] = uint32_t(trace::SpanKind::kServer);
```
ThomsonTan, line 486: Cannot we read this from the span instead of using a hard-code kind?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Presently it is falling back to default (`ETW`). I am wondering if we should add a comment here that maybe we should log a warning. Similar to what is done when `nullptr` is passed as name to `GetTracer`, i.e. it would probably be in spirit of the rest of API to ignore with some internal trace warning.. When we have a system for actually logging the warnings.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Okay, I will change that. I was not sure.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: I think the second argument should be `library_version` most likely. The first one is `instrument` or `instrumentation` name, and the second is presently unused.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
maxgolov: Let me get back to you on this.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -104,6 +106,72 @@ static inline void GetOption(const TracerProviderOptions &options,
...
+        break;
+
+      case CONST_HASHCODE(XML):
+        // nobrk
+      case CONST_HASHCODE(xml):
+        evtFmt = ETWProvider::EventFormat::ETW_XML;
+        break;
+
+      case CONST_HASHCODE(TLD):
+        // nobrk
```
lalitb, line 160: nit: remove this comment

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
lalitb: We can also return noop tracer if the encoding is incorrect/not-supporter.

lalitb (approved):

ThomsonTan (approved):

lalitb:
@maxgolov I have merged this PR. There were few open comments, if it needs changes please raise separate PR for that.

maxgolov:
@lalitb @ThomsonTan - absolutely. I'll address these minor comments when I add support for Span Links #649

PR #704, opened 2021-04-27
Fixing Visual Studio 2019 with `c++latest` (C++20) and `WITH_STL` build option

### LINKED ISSUE #703: C++20 readiness (build with STL classes is failing)
**This is C++20 compatibility clean-up issue**

The issue is unique to the build with:
- Visual Studio 2019 - `latest`
- C++20 enabled
- `HAVE_CPP_STDLIB` build option enabled (to use all API surface classes from STL)

This build combo may become relevant in future once we default to `C++20` and above support. Main problem is [described here in this thread](https://stackoverflow.com/questions/33874772/why-cant-i-construct-a-gslspan-with-a-brace-enclosed-initializer-list)... Basically, this code:

```cpp
    char trace_identity[trace_id_length + span_id_length + 6];
    span_context.trace_id().ToLowerBase16({&trace_identity[0], trace_id_length});
```

is incorrect when compiled with STL. Because `std::span` has no constructor that takes in `initializer list`. And in this scenario - Visual Studio 2019 treats the two parameters passed to constructor as initializer list, for some reason.

Due to whatever magical reason our own "in house" implementation of `nostd::span` - is way more relaxed. And it does not trip on this. I am sincerely hoping that `std::span` issue I observe here is NOT due to some aliases that I have written from `nostd::` to `std::` for STDLIB build flavor? 😁 ... And my honest understanding is we can fix it by simply adding an explicit conversion. Thus, I'm sending in a PR to fix it. There are couple other similar quirks like this in other `/propagation/` headers.

Also while testing the entire build with Visual Studio 2019 and `HAVE_CPP_STDLIB` ad-hoc (_since we do not have this in our Build loop yet_), I found a few other minor issues:
- `#include <array>` was missing in `ostream_log_test.cc`
- OTLP exporter test fails to compile inside Gmock (known issue with a slightly older Gmock and C++20)
- `#include <array>` was missing in `logger_provider_sdk_test.cc`

Main motivation why I'd like to get all these issues sorted out, is I'm working on a document that explains how to build the SDK using Visual Studio 2019 and CMake. Now since we give a build recipe - how to build with `HAVE_CPP_STDLIB` via UI in a click, it would be confusing if some components are not compiling... Also going forward - it would be natural in a year or two to expect that the entire API surface / OpenTelemetry DLL build relies on _standard_ STL classes more often, as within the confines of Windows OS we should expect a gradual shift to 2019 and above. Visual Studio 2022 public preview will be released this summer, so 2019 becomes "yesterday's version" and 2017 becomes "obsolete" as we speak. And we can more freely embrace the toolchain + STL native implementation of corresponding classes, like `std::variant` and `std::span` from STL. As we do that, we must follow the contracts described in C++20 spec.

maxgolov:
Fixes #703

## Changes

There are no API changes. Clean-up of the build failures with Visual Studio 2019, C++20 and `WITH_STL` ( `-DHAVE_CPP_STDLIB` ) option. Not a _significant_ contribution. However, we could mention that now the entire SDK can be compiled using vs2019 w/c++latest.

### DIFF
--- a/exporters/otlp/test/otlp_exporter_test.cc
+++ b/exporters/otlp/test/otlp_exporter_test.cc
@@ -1,16 +1,32 @@
-#include "opentelemetry/exporters/otlp/otlp_exporter.h"
+#ifndef HAVE_CPP_STDLIB
+// Unfortunately as of 04/27/2021 the fix is NOT in the vcpkg snapshot of Google Test.
+// Remove above `#ifdef` once the GMock fix for C++20 is in the mainline.
+//
+// Please refer to this GitHub issue for additional details:
+// https://github.com/google/googletest/issues/2914
+// https://github.com/google/googletest/commit/61f010d703b32de9bfb20ab90ece38ab2f25977f
+//
+// If we compile using Visual Studio 2019 with `c++latest` (C++20) without the GMock fix,
+// then the compilation here fails in `gmock-actions.h` from:
+//   .\tools\vcpkg\installed\x64-windows\include\gmock\gmock-actions.h(819):
+//   error C2653: 'result_of': is not a class or namespace name
+//
+// That is because `std::result_of` has been removed in C++20.
 
-#include "opentelemetry/exporters/otlp/protobuf_include_prefix.h"
+#  include "opentelemetry/exporters/otlp/otlp_exporter.h"
 
-#include "opentelemetry/proto/collector/trace/v1/trace_service_mock.grpc.pb.h"
+#  include "opentelemetry/exporters/otlp/protobuf_include_prefix.h"
 
-#include "opentelemetry/exporters/otlp/protobuf_include_suffix.h"
+// Problematic code that pulls in Gmock and breaks with vs2019/c++latest :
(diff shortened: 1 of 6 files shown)


exporters/otlp/test/otlp_exporter_test.cc
```diff
@@ -1,16 +1,32 @@
...
+//
+// If we compile using Visual Studio 2019 with `c++latest` (C++20) without the GMock fix,
+// then the compilation here fails in `gmock-actions.h` from:
+//   .\tools\vcpkg\installed\x64-windows\include\gmock\gmock-actions.h(819):
+//   error C2653: 'result_of': is not a class or namespace name
+//
+// That is because `std::result_of` has been removed in C++20.
 
-#include "opentelemetry/exporters/otlp/protobuf_include_prefix.h"
+#  include "opentelemetry/exporters/otlp/otlp_exporter.h"
```
maxgolov, line 16: Hmm, a bit confused. Need to fix this - probably merge issue, I did not intend to change those two lines.

lalitb (approved):

ThomsonTan (approved):

PR #706, opened 2021-04-27
Propagate resource to exporters

### LINKED ISSUE #334: Adding resource to TracerProvider

lalitb:
Fixes #334

## Changes


sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -244,6 +259,7 @@ class SpanData final : public Recordable
   std::vector<SpanDataEvent> events_;
   std::vector<SpanDataLink> links_;
   opentelemetry::trace::SpanKind span_kind_{opentelemetry::trace::SpanKind::kInternal};
+  opentelemetry::sdk::resource::ResourceAttributes resourceAttributes_;
```
jsuereth, line 262: What is this used for?  Is it worth copying the resource attributes in every span?

jsuereth:
Generally, LGTM.

sdk/include/opentelemetry/sdk/trace/span_data.h
lalitb: Thanks for noticing it. Will make it reference/pointer to resource :)

exporters/zipkin/include/opentelemetry/exporters/zipkin/recordable.h
```diff
@@ -67,6 +71,7 @@ class Recordable final : public sdk::trace::Recordable
 
 private:
   ZipkinSpan span_;
+  std::string service_name_;
```
maxgolov, line 74: I'm thinking here about some optimization opportunity: how often do you expect `service_name` value to change, i.e. do you have to copy it here for every recordable, then again into every `JSON` (ZipkinSpan) object?

exporters/zipkin/src/recordable.cc
```diff
@@ -213,6 +213,16 @@ void Recordable::SetName(nostd::string_view name) noexcept
   span_["name"] = name.data();
 }
 
+void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resource) noexcept
```
maxgolov, line 216: Confirming re. thread-safety of `SetResource` vs `GetServiceName` -- both cannot be called at once from two different threads?

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -50,6 +50,9 @@ class Tracer final : public trace_api::Tracer, public std::enable_shared_from_th
     return *instrumentation_library_;
   }
 
+  /** Returns the currently configured resource **/
+  const opentelemetry::sdk::resource::Resource &GetResource() { return context_->GetResource(); }
```
maxgolov, line 54: No setter - `SetResource` in API, only getter - `GetResource` ? What are the thread-safety guarantees of this method on API surface?

lalitb: The resource is a property of `TracerProvider` and should be set during the initialization of TracerProvider in the main thread. It's not re-settable again once initialized.

exporters/zipkin/include/opentelemetry/exporters/zipkin/recordable.h
lalitb: We can think of some optimization of not passing resource and instrumentation library as part of every span, but still, these data need to be copied to every ZipkinSpan. ( https://zipkin.io/zipkin-api/#/default/post_spans ).

exporters/zipkin/src/recordable.cc
lalitb: `SetResource` would be called when Span using `Tracer::StartSpan()`. And GetServiceName() would be when Span is ending : `Span::End() -> Processor::OnEnd() -> Exporter::Export()`. Both these need to be sequential in same thread.

exporters/zipkin/include/opentelemetry/exporters/zipkin/recordable.h
maxgolov: That's fine for now. I am looking at Zipkin exporter as example for Fluentd exporter, I can share some thoughts how I would've rearranged this in the Fluentd exporter.

exporters/zipkin/src/recordable.cc
maxgolov: I would let it go for now. But we may need to describe some of these expectations where we do not explicitly shield the access with a mutex, esp. in the other place where we return an object (that may get changed in another thread) by reference.

maxgolov (approved):

sdk/include/opentelemetry/sdk/trace/tracer.h
maxgolov: I approved.

From a practical standpoint, however, we do have scenarios as follows:
- whatever Monitoring Agent is upgraded, OR remote configuration is upgraded.
- your `resource` contents now may need to change - to indicate that a new configuration is applied without application service restart.

In those cases, I think, we may need to consider how to make some of this thread-safe. Current logic of returning `Resource` by reference in SDK may break if `Resource` needs to change in a different thread.

I am not sure how to track my concern. Maybe we need to have a generic umbrella issue that captures thread-safety guarantees at both levels:
- API thread safety guarantees
- SDK thread safety guarantees

In a separate markdown document, dev-focused document for exporter developers.

exporters/zipkin/src/recordable.cc
lalitb: Good point. Specifically for Resources, these are immutable as per the specs, with their ownership transferred to TracerProvider while it's initialization. So there won't be race condition arising while accessing them within the pipeline.

sdk/include/opentelemetry/sdk/trace/tracer.h
lalitb: > * your `resource` contents now may need to change, e.g. updated configuration of the `TracerProvider` - to indicate that a new configuration is now applied. Without application service restart. Your threads continue running, while accessing resource, while the resource object is maybe changing at the same time.. which may lead to crash.

The ownership model of the Resource object is well defined. Once created, the ownership of the Resource object is transferred to TracerProvider, and there is no setters to update it at runtime.

exporters/zipkin/include/opentelemetry/exporters/zipkin/recordable.h
lalitb: Thanks, will look forward to that :)

sdk/include/opentelemetry/sdk/trace/tracer.h
lalitb: Also to add, as per the specs : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md

``
When used with distributed tracing, a resource can be associated with the TracerProvider when the TracerProvider is created. That association cannot be changed later. When associated with a TracerProvider, all Spans produced by any Tracer from the provider MUST be associated with this Resource.
``

The resource once set can't be changed later.

exporters/otlp/include/opentelemetry/exporters/otlp/recordable.h
```diff
@@ -19,6 +19,9 @@ class Recordable final : public sdk::trace::Recordable
 public:
   const proto::trace::v1::Span &span() const noexcept { return span_; }
 
+  /** Dynamically converts the resource of this span into a proto. */
+  proto::resource::v1::Resource resource() const noexcept;
```
ThomsonTan, line 23: Probably call it `proto_resource` to differentiate it with SDK resource?

exporters/otlp/src/recordable.cc
```diff
@@ -213,6 +324,8 @@ void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept
   span_.set_kind(proto_span_kind);
 }
 
+void SetResource(const opentelemetry::sdk::resource::Resource &resource) noexcept {}
```
ThomsonTan, line 327: Is this global function needed?

sdk/src/trace/span.cc
```diff
@@ -102,7 +102,7 @@ Span::Span(std::shared_ptr<Tracer> &&tracer,
   recordable_->SetSpanKind(options.kind);
   recordable_->SetStartTime(NowOr(options.start_system_time));
   start_steady_time = NowOr(options.start_steady_time);
-  // recordable_->SetResource(tracer_->GetResoource()); TODO
+  recordable_->SetResource(tracer_->GetResource());
   // recordable_->SetResource(tracer_->GetInstrumentationLibrary()); TODO
```
ThomsonTan, line 106: Should this line be removed?

exporters/otlp/include/opentelemetry/exporters/otlp/recordable.h
lalitb: Sure, changed it.

exporters/otlp/src/recordable.cc
lalitb: Thanks for noticing it. No idea why I added it, removed it now.

sdk/src/trace/span.cc
lalitb: Yes it should be, removed now. Thanks.

ThomsonTan (approved):

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -225,6 +232,11 @@ class SpanData final : public Recordable
     span_kind_ = span_kind;
   }
 
+  void SetResource(const opentelemetry::sdk::resource::Resource &resource) noexcept override
+  {
+    resource_ = &resource;
```
maxgolov, line 237: @lalitb - this is problematic spot. It won't work too well with temporaries because it's assigning a pointer to reference. If `Resource` is compiled with `std::variant`, and a temporary is passed to API, e.g. `SetResource("key", "value")` - it'd break..

PR #708, opened 2021-04-28
Set Spankind for Zipkin exporter

lalitb:
## Changes

Small fix to populate spankind attribute for Zipkin exporter.
specs : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/zipkin.md#spankind

### DIFF
--- a/exporters/zipkin/src/recordable.cc
+++ b/exporters/zipkin/src/recordable.cc
@@ -231,10 +231,6 @@ void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept
   {
     span_["kind"] = span_iter->second;
   }
-  else
-  {
-    span_["kind"] = nullptr;
-  }
 }
 }  // namespace zipkin
 }  // namespace exporter
--- a/exporters/zipkin/src/recordable.cc
+++ b/exporters/zipkin/src/recordable.cc
@@ -16,13 +16,21 @@
 
 #include "opentelemetry/exporters/zipkin/recordable.h"
 
-#include <iostream>
+#include <map>
+#include <string>
 
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter
 {
 namespace zipkin
 {
+
+std::map<opentelemetry::trace::SpanKind, std::string> kSpanKindMap = {
+    {opentelemetry::trace::SpanKind::kClient, "CLIENT"},
+    {opentelemetry::trace::SpanKind::kServer, "SERVER"},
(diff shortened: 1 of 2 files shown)


exporters/zipkin/src/recordable.cc
```diff
@@ -215,7 +224,17 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept
   span_["duration"] = duration.count();
 }
 
-void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept {}
+void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept
+{
+  if (kSpanKindMap.find(span_kind) != kSpanKindMap.end())
+  {
+    span_["kind"] = kSpanKindMap.at(span_kind);
```
ThomsonTan, line 231: nit: save the `find()` result to a variable to avoid search the map twice?

ThomsonTan (approved):

exporters/zipkin/src/recordable.cc
lalitb: Good point. fixed that.

exporters/zipkin/src/recordable.cc
```diff
@@ -215,7 +224,18 @@ void Recordable::SetDuration(std::chrono::nanoseconds duration) noexcept
...
-void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept {}
+void Recordable::SetSpanKind(opentelemetry::trace::SpanKind span_kind) noexcept
+{
+  auto span_iter = kSpanKindMap.find(span_kind);
+  if (span_iter != kSpanKindMap.end())
+  {
+    span_["kind"] = span_iter->second;
+  }
+  else
+  {
```
maxgolov, line 235: I'm just wondering if you really need the `nullptr` assignment here, since the set is complete in the map... Also I think if you try to access the `span_["kind"]` property on JSON object elsewhere, it should already return you `nullptr` or an empty string anyways.... The only problem here might be is if there's a new type of `SpanKind` added to API, but not added to the map above.

maxgolov (approved):

exporters/zipkin/src/recordable.cc
lalitb: Good point. Will remove it.

maxgolov: Sounds good.

PR #709, opened 2021-04-28
Remove std::getenv in Zipkin test as reported by MSVC

ThomsonTan:
Similar to #698, replacing std::getenv by getenv_s to stop the warning emitted by MSVC.


lalitb (approved):
LGTM. 
Seems some reusable function would be handy as we need to do this in lots of places.

ThomsonTan:
> LGTM.
> Seems some reusable function would be handy as we need to do this in lots of places.

Yes, probably we could create a wrapper for `getenv_s` and return `std::unique_ptr` from the wrapper, so make it looks like `std::getenv` but doesn't return system managed memory.

I'll consider add this to utility if we see more calls of `std::getenv`.

PR #712, opened 2021-04-29
Convert null to empty string for tracer name

ThomsonTan:
## Changes

Below spec change required tracer name to be converted to empty string if it is null. As we are using `nostd::string_view` to represent null, this change converts it to empty string in `TracerProvider::GetTracer(name, version)`.

https://github.com/open-telemetry/opentelemetry-specification/pull/1654


sdk/src/trace/tracer_provider.cc
```diff
@@ -23,6 +23,9 @@ nostd::shared_ptr<opentelemetry::trace::Tracer> TracerProvider::GetTracer(
     nostd::string_view library_name,
     nostd::string_view library_version) noexcept
 {
+  if (library_name.data() == nullptr) {
+      library_name = "";
+  }
   // if (library_name == "") {
```
maxgolov, line 29: Do you still intend keeping the previous `TODO`, since now the code intentionally forces that condition?

maxgolov (approved):

sdk/src/trace/tracer_provider.cc
```diff
@@ -23,6 +23,9 @@ nostd::shared_ptr<opentelemetry::trace::Tracer> TracerProvider::GetTracer(
     nostd::string_view library_name,
     nostd::string_view library_version) noexcept
 {
+  if (library_name.data() == nullptr) {
```
lalitb, line 26: Passing a nullptr to GetTrace() is the application code issue, and the right behavior should be to let it crash so that the application developer can fix this. Probably I need to understand the spec change more clearly.

ThomsonTan: The latest [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md) mentioned this in explicit that null name should not crash and should be treat like empty name.
> In case an invalid name (null or empty string) is specified, a working Tracer implementation MUST be returned as a fallback rather than returning null or throwing an exception, its name property SHOULD be set to an empty string,

sdk/src/trace/tracer_provider.cc
ThomsonTan: The TODO is still needed I think, as the spec requires message reporting. Yes, I assume it is internal SDK logging.

>message reporting that the specified value is invalid SHOULD be logged

lalitb (approved):

lalitb:
@ThomsonTan  - Can you look into CI failures so that we can merge this ?

ThomsonTan:
> @ThomsonTan - Can you look into CI failures so that we can merge this ?

Thanks for the reminder, fixed the format.

PR #723, opened 2021-05-04
Cleanup: Get Span from Context, Set Span to Context 

### LINKED ISSUE #711: context: support getting and setting span without a key
It is not possible get and set the span on a context without using the span key, e.g. current usage of getting a span:

```c++
ctx.GetValue(trace::kSpanKey);
```
According to the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#context-interaction):
```
The API MUST provide the following functionality to interact with a Context instance:
- Extract the Span from a Context instance
- Insert the Span to a Context instance

The functionality listed above is necessary because API users SHOULD NOT have access to the Context Key used by the Tracing API implementation.
```

The key based access is used everywhere internally and externally, for example nginx instrumentation has a helper function for the same purpose: https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/instrumentation/nginx/src/propagate.cpp#L102

We do have a helper function called `GetCurrentSpan` in propagators' detail namespace:
https://github.com/open-telemetry/opentelemetry-cpp/blob/d1f0b136e3cf88e3d13b2b979520fafe32705319/api/include/opentelemetry/trace/propagation/detail/context.h#L12-L20

Similar function in the JS API:
https://github.com/open-telemetry/opentelemetry-js-api/blob/b5784d30e51f5f7ffbdba0e6b83c07dd23b50c8d/src/context/context.ts#L44-L46

I propose renaming `GetCurrentSpan` in `trace::propagation::detail` to `GetSpan` (also making it return a span not span context) and move it to a proper header and namespace. In a similar way `SetSpan` should be added.

lalitb:
Fixes #711 

## Changes

1. Make `Tracer::GetCurrentSpan()` and `Tracer::WithActiveSpan()` methods static, as they are not dependent on Tracer instance ( inspired from dotnet tracer implementation:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Trace/Tracer.cs#L41-L67

2. Remove `tracer.cc:GetCurrentSpanContext()` helper method, and instead use `Tracer::GetCurrentSpan()` wherever required.

3. Fix for #711 : Implement `tracer::propagation::detail::GetSpanFromContext()` and `tracer::propagation::detail::SetSpanToContext()` functions. @seemk can you please review this change. The header `tracer/propagation/detail.h` is already part of API installation so these functions can be used from there.
( these changes are inspired from python implementation: https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/propagation/__init__.py )


ThomsonTan (approved):

reyang (approved):
LGTM.

api/include/opentelemetry/trace/propagation/detail/context.h
```diff
@@ -9,15 +10,22 @@ namespace propagation
...
-inline trace::SpanContext GetCurrentSpan(const context::Context &context)
+inline nostd::shared_ptr<trace::Span> GetSpanFromContext(const context::Context &context)
 {
   context::ContextValue span = context.GetValue(trace::kSpanKey);
   if (nostd::holds_alternative<nostd::shared_ptr<trace::Span>>(span))
   {
-    return nostd::get<nostd::shared_ptr<trace::Span>>(span).get()->GetContext();
+    return nostd::get<nostd::shared_ptr<trace::Span>>(span);
   }
+  static nostd::shared_ptr<Span> invalid_span{new DefaultSpan(SpanContext::GetInvalid())};
```
maxgolov, line 20: Question / concerns about the combination of `inline` and `magic static` here:

- magic statics may present a challenge on app termination, i.e. if someone tries to call into `GetSpanFromContext` in `std::atexit` handler. You'd most likely experience a crash.

- since you are returning a `shared_ptr` to what appears to be `invalid_span`, what prevents me to overwrite the contents of that ptr with my own span then. And the next caller then would actually get my span, not the invalid span, i.e. appears like previously the code was safe. Now I can hijack and override the invalid span with my own span.

Can you double-check on this?

lalitb: Both are valid points. Thanks for noticing it. I don't think we should be returning static in case of invalid_span. We also don't do it here: 

https://github.com/open-telemetry/opentelemetry-cpp/blob/c7f49debf3331ef912a8a20e0b1f29b0a67317ae/api/include/opentelemetry/trace/tracer.h#L162

Will fix it accordingly.

api/include/opentelemetry/trace/propagation/detail/context.h
```diff
@@ -9,15 +10,22 @@ namespace propagation
 namespace detail
 {
 
-inline trace::SpanContext GetCurrentSpan(const context::Context &context)
+inline nostd::shared_ptr<trace::Span> GetSpanFromContext(const context::Context &context)
```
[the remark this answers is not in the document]
lalitb: Sure, that's a good point.

api/include/opentelemetry/trace/propagation/detail/context.h
maxgolov: I checked that I might have been wrong about what'd happen. With clang c++11 appears like my 2nd concern is maybe wrong.. But most likely you can simply remove the `static` keyword at line 20. This would be nicer to read. I like your last suggestion!

maxgolov (approved):
Looks good. Approved with suggestion: please consider removing `static` keyword and implementing what we discussed in a thread above.

api/include/opentelemetry/trace/propagation/detail/context.h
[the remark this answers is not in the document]
lalitb: @seemk @maxgolov  - I have renamed these functions as `GetSpan` and `SetSpan`, however, they are now part of the `trace::propagation` namespace instead of `trace`, as these functions are specific to `propagation`. Let me know if there are any concerns.

seemk (approved):
Thanks, LGTM 👍

PR #727, opened 2021-05-06
Enhance existing http example to support w3c trace context propagation

lalitb:
## Changes

This PR enhances the existing HTTP client/serve example with http propagation. The `trace_id` and `parent_span_id` are propagated from client to server. 
Also nit change on renaming the header file extension as per the convention (`hpp` -> `h`)


examples/http/server.cc
```diff
@@ -19,6 +19,15 @@ class RequestHandler : public HTTP_SERVER_NS::HttpRequestCallback
     options.kind          = opentelemetry::trace::SpanKind::kServer;  // server
     std::string span_name = request.uri;
 
+    // extract context from http header
+    const HttpTextMapCarrier<std::map<std::string, std::string>> carrier(
+        (std::map<std::string, std::string> &)request.headers);
+    auto prop = opentelemetry::context::propagation::GlobalTextMapPropagator::GetGlobalPropagator();
+    auto current_ctx = opentelemetry::context::RuntimeContext::GetCurrent();
+    auto new_context = prop->Extract(carrier, current_ctx);
+    options.parent   = GetSpanFromContext(new_context)->GetContext();
```
pyohannes, line 28: I think this shouldn't be strictly necessary, as `Extract` sets the span as the currently active span in the context?

examples/http/tracer_common.h
```diff
@@ -0,0 +1,95 @@
...
+#include "opentelemetry/trace/provider.h"
+
+#include "opentelemetry/context/propagation/global_propagator.h"
+#include "opentelemetry/context/propagation/text_map_propagator.h"
+#include "opentelemetry/trace/propagation/http_trace_context.h"
+
+#include <cstring>
+#include <iostream>
+#include <vector>
+#include "opentelemetry/ext/http/client/http_client.h"
```
maxgolov, line 14: I wonder if we can _eventually_ generalize some of this, since the same / similar logic would be necessary for gRPC metadata. I like how it was handled in dapr examples:

https://docs.dapr.io/developing-applications/building-blocks/observability/w3c-tracing/w3c-tracing-howto/

maxgolov (approved):
I'm good with this for now. We may need to refactor this to generalize the common parts to be agnostic of whether it's HTTP client headers or gRPC metadata.

examples/http/server.cc
lalitb: @pyohannes - `Extract` would return a new context with the `span` created from extracted `span_context`, and won't make it active. This is as per the [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#textmap-extract):

``
Returns a new Context derived from the Context passed as argument.
``

examples/http/tracer_common.h
lalitb: Yes, we should eventually reusable the code for both HTTP and grpc.

PR #728, opened 2021-05-06
Minor api documentation cleanup

pyohannes:
## Changes

* Remove documentation references for internal `common` classes that are not relevant to users.
* Remove documentation references for `nostd` classes. We shouldn't confuse the users with documentation for those classes, as we must strive for 100% compatibility with the standard library anyway.
* Remove `NullKeyValueIterable`. It isn't used anywhere in the code.

Cleanup of documentation in the `opentelemetry::trace` namespace will come in a separate PR.

### DIFF
--- a/api/docs/Doxyfile
+++ b/api/docs/Doxyfile
@@ -839,6 +839,8 @@ RECURSIVE              = YES
 
 EXCLUDE                = ../include/opentelemetry/common/spin_lock_mutex.h \
                          ../include/opentelemetry/common/key_value_iterable_view.h \
+                         ../include/opentelemetry/common/kv_properties.h \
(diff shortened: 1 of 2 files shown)


api/docs/Doxyfile
```diff
@@ -839,8 +839,10 @@ RECURSIVE              = YES
 
 EXCLUDE                = ../include/opentelemetry/common/spin_lock_mutex.h \
                          ../include/opentelemetry/common/key_value_iterable_view.h \
+                         ../include/opentelemetry/common/kv_properties.h \
```
ThomsonTan, line 842: Could we move these files into a separate folder like `internal` to make it clear that they are not intended for end-user?

ThomsonTan (approved):

PR #729, opened 2021-05-08
Simple instrumented gRPC example

Hablapatabla:
## Changes

Add a simple instrumented client/server example using gRPC. Instructions on running the example are held in the README. The example requires gRPC and Protobuf as dependencies. This should be no problem but in the project-level CMakeLists, they are only loaded when -DWITH_OTLP=ON, which I will look out for.


examples/grpc/client.cc
```diff
@@ -0,0 +1,109 @@
...
+    // and because it could still be useful to a debugger in the future.
+    std::string span_name = "GreeterClient/Greet";
+    auto span             = get_tracer("grpc-client")
+                    ->StartSpan(span_name,
+                                {{"rpc.system", "grpc"},
+                                 {"rpc.service", "grpc-example.GreetService"},
+                                 {"rpc.method", "Greet"},
+                                 {"net.peer.ip", ip},
+                                 {"net.peer.port", port},
+                                 {"rpc.grpc.status_code", 0}},
```
lalitb, line 51: status-code value should be based on the response received from the server. We should set it after that receiving server response using `span::SetAttribute(key, value)`

examples/grpc/client.cc
```diff
@@ -0,0 +1,109 @@
...
+                                options);
+    auto scope = get_tracer("grpc-client")->WithActiveSpan(span);
+
+    gRPCMapCarrier carrier;
+    carrier.gRPCMapCarrier::Set("http.header.stub", "temporarily-stubbed");
+
+    opentelemetry::context::Context ctx1 = opentelemetry::context::Context{"current-span", span};
+    opentelemetry::context::Context ctx2 = propagator->Extract(carrier, ctx1);
```
lalitb, line 59: Client should take the current (span)context, and inject that into carrier. Why are we doing `Extract` here ?

examples/grpc/experimental/greeter_interceptor.cc
```diff
@@ -0,0 +1,199 @@
...
+    const char *method                           = info->method();
+    grpc::experimental::ServerRpcInfo::Type type = info->type();
+  }
+
+  void Intercept(grpc::experimental::InterceptorBatchMethods *methods) override
+  {
+    if (methods->QueryInterceptionHookPoint(
+            grpc::experimental::InterceptionHookPoints::PRE_SEND_INITIAL_METADATA))
+    {
+      auto *map = methods->GetSendInitialMetadata();
```
lalitb, line 46: The real use case of interceptors for instrumentation would be to intercept the requests, and create spans out of the request attributes to be exported to external endpoint. While it is ok for now to have `cout` here, please update the README for this as todo/comming soon.

examples/grpc/grpc_foo_lib/foo_split.cc
```diff
@@ -0,0 +1,54 @@
...
+
+  gRPCMapCarrier carrier;
+  carrier.gRPCMapCarrier::Set("http.header.stub", "temporarily-stubbed");
+
+  auto span                            = get_tracer("grpc-lib")->StartSpan("splitlib");
+  auto scope                           = get_tracer("grpc-lib")->WithActiveSpan(span);
+  opentelemetry::context::Context ctx1 = opentelemetry::context::Context{"current-span", span};
+  opentelemetry::context::Context ctx2 = propagator->Extract(carrier, ctx1);
+  gRPCMapCarrier carrier2;
+  propagator->Inject(carrier2, ctx2);
```
lalitb, line 48: We can just fetch the current context from run-time, and inject that into the carrier. We don't need `Extract` to get the current context.

lalitb (commented):
Thank you for the PR. This is definitely a good start.

tools/ports/benchmark/portfile.cmake
```diff
@@ -1,49 +1,49 @@
...
-
-vcpkg_fixup_cmake_targets(CONFIG_PATH lib/cmake/benchmark)
-
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/include)
-file(REMOVE_RECURSE ${CURRENT_PACKAGES_DIR}/debug/share)
-
-# Handle copyright
-file(COPY ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/benchmark)
-file(RENAME ${CURRENT_PACKAGES_DIR}/share/benchmark/LICENSE ${CURRENT_PACKAGES_DIR}/share/benchmark/copyright)
+if(VCPKG_CMAKE_SYSTEM_NAME STREQUAL "WindowsStore")
```
maxgolov, line 1: I think you need to revert this file since you added it due to issue you had with line endings.

tools/ports/benchmark/CONTROL
```diff
@@ -1,5 +1,5 @@
-Source: benchmark
-Version: 1.5.1
-Homepage: https://github.com/google/benchmark
-Description: A library to support the benchmarking of functions, similar to unit-tests.
+Source: benchmark
```
maxgolov, line 1: Same here. No changes, unrelated to your work, revert this one.

examples/grpc/grpc_foo_lib/foo_split.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include <sstream>
+#include <string>
+#include <vector>
+#include <map>
+#include "../tracer_common.h"
```
maxgolov, line 6: This relative path `../` looks a bit unclean. You'd rather add the parent directory to include paths.

lalitb:
@Hablapatabla  - Hope you are fine with me taking over these changes, as we need them urgently now. I will try to see if I have permission to commit to your branch, else will move these changes to my fork and update.

lalitb:
> @Hablapatabla - Hope you are fine with me taking over these changes, as we need them urgently now. I will try to see if I have permission to commit to your branch, else will move these changes to my fork and update.

Ok, I have fixed the propagation, need to add cmake option to optionally enable building this example ( as this needs grpc ), and also need to fix README. Will do that too.

tools/ports/benchmark/portfile.cmake
lalitb: yes removed file now.

examples/grpc/grpc_foo_lib/foo_split.h
lalitb: removed this code, as we don't need this for simple example.

examples/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 if(WITH_OTLP)
   add_subdirectory(otlp)
+  add_subdirectory(grpc)
```
lalitb, line 3: We can use a separate cmake option too - but I felt that it's better to build this along with otlp as both have dependency on grpc library.

lalitb:
@open-telemetry/cpp-approvers  - Can you please review this PR. This is now in working condition with context propagation using grpc metadata.

jsuereth (approved):
Great work!

This looks great, can you open a bug to add a bazel build for this?

PR #730, opened 2021-05-08
Allow to build with alternate vcpkg.exe by respecting VCPKG_ROOT variable

### LINKED ISSUE #701: OpenTelemetry C++ SDK build with vcpkg / OTLP-gRPC: 260 symbols MAX_PATH limitation
This is NOT OpenTelemetry C++ SDK issue. But rather a build issue with ninja, described here:

https://github.com/ninja-build/ninja/issues/1900

**Problem Description**

Imagine a project that builds OpenTelemetry C++ SDK as a dependent git submodule in its own build tree. OpenTelemetry C++ SDK on Windows is easy to build using vcpkg. However, there is a file path length limitation of 260 symbols that is coming into play here if you are building with OTLP+gRPC enabled.

The following layout:
`C:\agent\worker\MyInstrumentationFrameworkProject\third_party\opentelemetry-cpp\tools\vcpkg\buildtrees\grpc`

fails to build miserably while trying to install OTLP-gRPC on Windows with the following error:

`ninja: error: Stat(C:/agent/worker/MyInstrumentationFrameworkProject/third_party/opentelemetry-cpp/tools/vcpkg/buildtrees/grpc/src/2948941d75-78cfc6f1b1.clean/src/core/ext/upbdefs-generated/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.upbdefs.c): Filename longer than 260 characters`

Since it is rather hard to predict where the source code is going to be located, especially if it gets pulled into another project built somewhere on a build pipeline as a submodule... we do not have much options. We may try to assume that a custom `%TEMP%` variable is set to short path. And we host `vcpkg` build tree inside `%TEMP%` rather than hosting it inside the `tools/vcpkg` directory? That may work. However, this approach has some security implications. Ideally the full build has to be isolated in a single work tree, without needing any external directories, especially external temporary directories which may be harder to isolate.

Immediate recommendation we can give that the origin name of `opentelemetry` working directory MUST BE shorter than `260 - 5 - strlen("/tools/vcpkg/buildtrees/grpc/src/2948941d75-78cfc6f1b1.clean/src/core/ext/upbdefs-generated/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.upbdefs.c")`, that is **no longer than 71 symbols**.

In the above example -  the path **C:/agent/worker/MyInstrumentationFrameworkProject/third_party/opentelemetry-cpp/** is already too long. As it will fail `vcpkg install gRPC:x64-windows` installation steps within the same build tree, because it is exceeding the allowed base path length limit. There could be other issues coming into play later on: as new extensions could be added to gRPC source, there is no such limitation on Unix OS, and filenames could be even longer than the example above.

**Quick recap - TL;DR**

Don't use names that are "too long" as base path for building OpenTelemetry C++ SDK on Windows.

Use something short, like `C:\otel\` - is a good name. And then the `vcpkg gRPC` build required for OTLP - passes okay. 😁

maxgolov:
Fixes #701 

## Changes

This is quite small and simple change to start respecting [`VCPKG_ROOT` variable](https://github.com/microsoft/vcpkg/blob/master/docs/users/config-environment.md#vcpkg_root). This fix (partially) addresses the corner-case, when it turns out that the path to gRPC build tree is too long on Windows.

Developers may provide a short path to their own `vcpkg` installation. Such as `set VCPKG_ROOT=c:\vcpkg`, then the `setup-buildtools.cmd` is going to honor that short path. gRPC is then building successfully.

One side-effect is that developer-provided snapshot of `vcpkg` is going to be used in that case instead of opentelemetry submodule-provided version of `vcpkg`.

## Why this fix is needed

In my case I'm running an Azure pipeline build, where the project name and, subsequently, the path to project directory is relatively long. And OpenTelemetry itself lives in a `third_party` submodule of its parent project build tree. Plus, gRPC project has very long file names and directory names, that aggravates the [file length limit issue on Windows](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation). 🤷  An adjustment to my pipeline then is to checkout the `vcpkg` separately in its own location to `C:\vcpkg` (shorter path). And use `VCPKG_ROOT` to direct the build script to use that instance of it instead.

I believe this fix would be beneficial to other OpenTelemetry devs as well.

## Alternate ideas considered

It is possible to use `mklink /J` command to create a directory junction from `C:\vcpkg` to OpenTelemetry snapshot of `vcpkg` submodule directory, as a shorter path-alias. But this is just too smart.. Has its own side-effects. Thus, I'm proposing a dumber and simpler quick fix instead. The entire quick fix is optional. If someone does not like it, then they just don't set `VCPKG_ROOT` variable at all.

On a positive note, this also means now you can reuse one instance of `vcpkg` installation across several checkouts of opentelemetry build tree instead of building it each and every time.

@meastp - please let me know if this quick fix seems reasonable to you.

### DIFF
--- a/tools/build.cmd
+++ b/tools/build.cmd
@@ -26,8 +26,8 @@ if ("%CMAKE_ARCH%"=="") (
 )
 
 REM Use preinstalled vcpkg if installed or use our local
-if "%VCPKG_INSTALLATION_ROOT%" neq "" (
-  set "VCPKG_CMAKE=%VCPKG_INSTALLATION_ROOT%\scripts\buildsystems\vcpkg.cmake"
+if "%VCPKG_ROOT%" neq "" (
+  set "VCPKG_CMAKE=%VCPKG_ROOT%\scripts\buildsystems\vcpkg.cmake"
 ) else (
   set "VCPKG_CMAKE=%CD%\vcpkg\scripts\buildsystems\vcpkg.cmake"
 )
--- a/tools/setup-buildtools.cmd
+++ b/tools/setup-buildtools.cmd
@@ -1,5 +1,6 @@
 @echo off
 set "PATH=%ProgramFiles%\CMake\bin;%~dp0;%~dp0vcpkg;%PATH%"
+if "%VCPKG_ROOT%" NEQ "" set "PATH=%VCPKG_ROOT%;%PATH%"
 pushd %~dp0
 
 net session >nul 2>&1


maxgolov:
My good old friend `code coverage bot` is wrong again! No, changing the build script does not affect the code coverage 😀

Issue #572, opened 2021-02-10
Switch to abseil::variant as default variant class across api


lalitb:
opentelemetry-cpp api currenty brings 3 variants of `variant` class switchable at build time.
1. std variant ( `HAVE_CPP_STDLIB` switch :  https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/std/variant.h

2. abseil variant ( `HAVE_ABSEIL_VARIANT` switch:  https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry//nostd/absl/types/variant.h )

3. mark variant ( enabled by default: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/nostd/mpark/variant.h )

Going forward - it would be better to converge to single variant class ( possibly abseil::variant as it is usable with vc2015 ), and get rid of rest of the libraries. 

Refer to discussions in this PR thread: https://github.com/open-telemetry/opentelemetry-cpp/pull/565

PR #731, opened 2021-05-09
Make ContextValue default constructible with monostate

ThomsonTan:
## Changes

Add `nostd::monostate` to `variant` type `ContextValue` so it can be default constructed with empty value. With this change, we don't need return a valid `ContextValue` for a non-exist key.


api/include/opentelemetry/context/context_value.h
```diff
@@ -13,7 +13,8 @@
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace context
 {
-using ContextValue = nostd::variant<bool,
+using ContextValue = nostd::variant<nostd::monostate,
```
maxgolov, line 16: We don't currently run CI around `std::` substitute for `nostd::` . But I think this would have to be added around here: https://github.com/open-telemetry/opentelemetry-cpp/blob/d8467a78b3cf3accfa3a0a2569bfd3b90822ad02/api/include/opentelemetry/nostd/variant.h#L18

For the build flavor that uses either `absl::` or `std::` for `variant` implementation.

maxgolov: If you can try building your change with `tools\build.cmd` on Windows just to check - it builds both `nostd` flavor and STL flavor (the one that aliases all `nostd::` classes to `std::` for vs2019 in C++20 mode).. Otherwise if all other folks are OK with the change, I can fix it if you are breaking `WITH_STL` option.

lalitb: +1 to support absl::monostate and std::monostate. 
@maxgolov  - What do you think about supporting only one implementation of variant (#572)

ThomsonTan: Thanks @maxgolov and @lalitb , added alias for `nostd::monostate` to both `std::monostate` and `absl::monostate`.

lalitb (approved):

api/include/opentelemetry/context/context_value.h
maxgolov: I'd say we switch to Abseil by default.

PR #734, opened 2021-05-10
potential fix for std::variant implicit conversion in C++17 STL

### LINKED ISSUE #733: Avoid pointer to reference to temporaries in `SetResource` API
**Repro Steps**

- Compile with standard library (STL) classes.
- Run all tests: 335 tests passed; 10 tests failed (all related to `SetResource` API).

image

@lalitb - this is problematic spot. It won't work too well with temporaries because it's assigning a pointer to reference.

If `Resource` is compiled with `std::variant`. And a temporary is passed to API, e.g.

`SetResource("key", "value")`

It'd break.. Sorry, I found it just now. Haven't noticed during code review originally.. :( Now I tagged the spot in review below:

https://github.com/open-telemetry/opentelemetry-cpp/pull/706#r629059711_

**Exact Problematic Test Spot**

image

Same issue across all 10 tests.

**How to Fix**

We need to avoid a pointer to temporary here. Since `Resource` values are not changing often, perhaps a copy is more appropriate than a pointer?

lalitb:
Fixes ##733

## Changes

@maxgolov  This is not for merge, but to showcase the potential changes needed to fix #733, As mentioned in the comment there, this is not related to resource sdk.


lalitb:
> I'm Okay with this as a temporary measure. I think we could have handled it in the `AttributeValue` class.

This is not related to `string_view`. As an example, resource sdk doesn't use `AttributeValue`, but directly uses `OwnedAttributeValue`. The problem is with the implicit conversion of `char *` to `bool` in C++17, which seems to be corrected in C++20. Please see my comment on the issue you created.

exporters/ostream/test/ostream_log_test.cc
```diff
@@ -150,7 +150,7 @@ TEST(OStreamLogExporter, LogWithStringAttributesToCerr)
   auto record = exporter->MakeRecordable();
 
   // Set resources for this log record only of type <string, string>
-  record->SetResource("key1", "val1");
+  record->SetResource("key1", std::string("val1"));
```
ThomsonTan, line 153: Could we provide an explicit overload to `SetResource` with `const char *` to avoid potential future error?

maxgolov (changes requested):
I don't like that. I'm proposing to re-add `const char *` to variant. Just verified that it works, and including it in the Abseil Variant PR.

This re-add of `const char *` will give us much nicer user / developer experience, when we get to C++17 standard lib impl of variant.

lalitb:
> I don't like that. I'm proposing to re-add `const char *` to variant. Just verified that it works, and including it in the Abseil Variant PR.
> 
> This re-add of `const char *` will give us much nicer user / developer experience, when we get to C++17 standard lib impl of variant.

Agree, it's not fully user-friendly, but it keeps api/sdk stable. Unless we plan to add `const char *` to `OwnedAttributeValue`, these changes need to go through.

Issue #741, opened 2021-05-11
`HAVE_CPP_STDLIB` and  `HAVE_GSL` should be auto added when using `-DWITH_STL=ON` with cmake.


owent:
**Describe your environment**
Windows, CMake

**Steps to reproduce**
Using `cmake -DWITH_STL=ON ...` to build project.

**What is the expected behavior?**
`HAVE_CPP_STDLIB` and `HAVE_GSL` should be added into `INTERFACE_COMPILE_DEFINITIONS` of exported target `opentelemetry-cpp::api` .

**What is the actual behavior?**
Nothing here.

**Additional context**

owent:
Closed by #742

PR #742, opened 2021-05-11
Fix issues #740 and #741

### LINKED ISSUE #740: Can not use prebuilt protobuf of parent project with cmake.
**Describe your environment** 
CMake, Windows

**Steps to reproduce**
Having a prebuilt of protobuf.

Do not trigger vcpkg.

Using vcpkg to installl protobuf again.

**Additional context**
Should use `Protobuf_FOUND` instead of `protobuf_FOUND` .

owent:
Fixes #740 #741

## Changes

+ Move `HAVE_CPP_STDLIB` and `HAVE_GSL` into `INTERFACE_COMPILE_DEFINITIONS` of `opentelemetry_api`.
+ Replace `protobuf_FOUND` with `Protobuf_FOUND`.


ThomsonTan (approved):

lalitb:
@owent  - can you rebase this so that we can merge it

owent:
> @owent - can you rebase this so that we can merge it

Done.

CMakeLists.txt
```diff
@@ -49,8 +49,6 @@ endif()
 
 if(WITH_STL)
   # Require at least C++17. C++20 is needed to avoid gsl::span
-  add_definitions(-DHAVE_CPP_STDLIB -DHAVE_GSL)
```
maxgolov, line 52: I have minor doubt here. Don't we need this definition in the main common place for other projects that may be including API headers directly, i.e. if I build a test app that consumes API, do I need to add a dependency on `opentelemetry_api` target everywhere? If another target is created, such as that would aggregate multiple targets into one, would that new target also have to add those two definitions?

Trying to understand the _cons_ of having this possibly defined in both spots, main `CMakeLists.txt`, as well as in the api?

owent: You mean we should just add `HAVE_CPP_STDLIB` and `HAVE_GSL` into `INTERFACE_COMPILE_DEFINITIONS` of `opentelemetry_resources` in `opentelemetry_sdk`? It seems better, I will test it later.

lalitb:
@owent  - Can you please rebased once again. There was one unresolved comment so couldn't merge last time. Looks good to merge now.

halayli:
I am a bit confused by this change. What does STL has to do with GSL? What if someone wants STL but doesn't want GSL?

I feel this trickled from msvc environment. How about having this behavior for windows only at least?

owent (commented):
> I am a bit confused by this change. What does STL has to do with GSL? What if someone wants STL but doesn't want GSL?
> 
> I feel this trickled from msvc environment. How about having this behavior for windows only at least?



> I am a bit confused by this change. What does STL has to do with GSL? What if someone wants STL but doesn't want GSL?
> 
> I feel this trickled from msvc environment. How about having this behavior for windows only at least?

I'm not familiar with GSL. It seems to import some experimental libraries of STL, and alias to the implement if STL when STL support them.

halayli:
I feel this needs to be a flag. automagically importing a third party library when specifying `WITH_STL` will surprise developers.

lalitb:
> I feel this needs to be a flag. automagically importing a third party library when specifying `WITH_STL` will surprise developers.

We only use `gsl::span` from GSL. But I agree with you.  If the WITH_STL option is specified, it should be using `std::span` if available ( for C++20), and fall back to `nostd::span` instead of `gsl::span`. Use `gsl::span` only if developers specifies WITH_GSL option.

Will wait for @maxgolov to reply, probably `gsl::span` is a more complete implementation than `nostd::span` and so this logic.

owent:
> I feel this needs to be a flag. automagically importing a third party library when specifying `WITH_STL` will surprise developers.

Agree too.

maxgolov:
@halayli @owent @lalitb 
`std::span` is only available since C++20. The only official _1-1 conforming implementation_ of `span` for C++17 is `gsl::span`. This is described in [Cpp Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines)

The option (including GSL span) is described in greater level of detail here in this document in our repo:
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/building-with-stdlib.md

I have already submitted a change that automagically falls-back to `nostd::span` in case if GSL span is not available in the build environment, as part of this PR #771 .

GSL is friendly-licensed available as a [submodule here](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/third_party). GSL span is cross-platform.

PR #744, opened 2021-05-12
Enforce the C++ standard in CMake when CXX_STANDARD is specified

iblancasa:
Fixes

## Changes

Ensure the desired C++ standard is used when building the libraries. This is done enabling **[`CMAKE_CXX_STANDARD_REQUIRED`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD_REQUIRED.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ else()
   )
 endif()
 
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 if(NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 11)
 endif()


lalitb (approved):

ThomsonTan (approved):

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ else()
   )
 endif()
 
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
maxgolov, line 30: Does it work well with construct at line below?

```cmake
  if(CMAKE_MINOR_VERSION VERSION_GREATER "3.18")
    # Ask for 20, may get anything below
    set(CMAKE_CXX_STANDARD 20)
  else()
    # Ask for 17, may get anything below
    set(CMAKE_CXX_STANDARD 17)
  endif()
```

i.e. if I'm asking for 20, but my compiler only supports 17. Would I get 17 without any error?

maxgolov (changes requested):
https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED
`If this property is set to ON, then the value of the CXX_STANDARD target property is treated as a requirement. If this property is OFF or unset, the CXX_STANDARD target property is treated as optional and may "decay" to a previous standard if the requested is not available. For compilers that have no notion of a standard level, such as MSVC 1800 (Visual Studio 2013) and lower, this has no effect.`

**If this property is OFF or unset, the CXX_STANDARD target property is treated as optional and may "decay" to a previous standard**

^ this was actually the intent, to decay to lower version automagically. We require any compiler that supports at least `C++11`. But depending on standard set - we may more optimally build with `C++20`, or safely decay to `C++17` or top `C++14`. But it still would build just fine.

Can you kindly elaborate on a practical need to make it a **requirement**? Can you make it a requirement in the outside external build script by passing corresponding option?

If we are making this change, could you please recommend the solution for the following scenario:

- OS may have different toolchain installed `gcc-7` or `clang-8` - those did not have support for C++20. For example, 2-3 old Mac machine, with an older Xcode toolchain, that may have either older tooling (C++17) or more recent tooling (C++20).

- build `WITH_STL` option previously assumed the soft "decay" to lower standard happening automagically in that scenario, i.e. trying to get all standard template features from 20. But if not available, safely decay / downgrade to 17.

Please address this issue in the other spots where `CXX_STANDARD` is being used in the same file.

iblancasa:
The next lines are:
```cmake
if(NOT DEFINED CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 11)
endif()
```

What I was trying to avoid is: if you try to add support to a different system or try to compile this project with a compiler not supporting C++11 or newer, the build will fail too late. I mean: CMake should raise you need, at least, support for C++11 in your compiler.

A possible solution for this situation could be checking the [`CMAKE_CXX_KNOWN_FEATURES`](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#cmake-cxx-known-features) CMake variable to know the "newer supported standard" by the used toolchain, store it in a variable and use it when configuring the targets. I mean, something like (this is kind of pseudo-code):

```cmake
for(standard ${list_of_supported_standards_from_newer_to_older})
   if(standard IN CMAKE_CXX_KNOWN_FEATURES)
       set(TOOLCHAIN_NEWER_STANDARD CXX_STANDARD ${standard})
   endif()
endfor()

set_target_properties(my_target
    PROPERTIES
        ${TOOLCHAIN_NEWER_STANDARD}
)
```

ThomsonTan:
@maxgolov I am not sure the "decay" you mentioned here is done by CMAKE, the sounds like the feature detection should be in the `.cc/.h` source files instead of CMake build script.

>build WITH_STL option previously assumed the soft "decay" to lower standard happening automagically in that scenario

I'd not recommend [CMAKE_CXX_KNOWN_FEATURES](https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#cmake-cxx-known-features) if we trying to use it to simulate the decay feature.

iblancasa:
What CMake does to know if one standard can be used or not with one toolchain is to check the value of `CMAKE_CXX_KNOWN_FEATURES`. When the requested standard is not available and `CMAKE_CXX_STANDARD_REQUIRED` is set to `OFF`, it will try to use the previous standard version until find one supported. 

I think letting CMake guess what standard use is not really intuitive when building because the error is not shown until you are building. Those kind of errors, I think, should be raised while generating the build system.

maxgolov:
> What I was trying to avoid is: if you try to add support to a different system or try to compile this project with a compiler not supporting C++11 or newer...

The answer to that `just don't...` 😃 

> I think letting CMake guess what standard use is not really intuitive when building because the error is not shown until you are building

Technically this is purely 'document'-able / administrative issue, where we can mention the list of compilers that we support. We require C++11. This is clearly mentioned in the [README](https://github.com/open-telemetry/opentelemetry-cpp#supported-c-versions) , the first thing people should see. Even `gcc-4.8.x` being marginally enough, needing quite a bit of patches - because there are some known quirks with OTLP-gRPC at the moment. Latest gRPC needing a compiler newer than `gcc-4.8.x`.

I'd say this change adds more hassle than the sheer potential volume of problems it is trying to solve.

We can create a matrix document [similar to this](https://github.com/microsoft/cpp_client_telemetry#build-environments) that precisely lists all compilers that we tried, known to work, with tests all passing, and the configurations we are eager to support. Anything unlisted may be supported if someone contributes a PR, then updates the support matrix accordingly.

I'd say a document would be better than trying to fix it in `CMakeLists.txt`

iblancasa:
I'm not really convinced about that approach but I understand your concerns. I have been working on an environment supporting more than 84 combinations of toolchain-operating system and I faced some situations (specially with third-party) when it was not so obvious that the code was not supported (even having documentation). I just was trying to help other developers to avoid wasting their time (as I did for different tools during some years 😅). But yes, I understand your approach can be enough. Thanks for your time to explain your point of view 😄!

> We can create a matrix document similar to this that precisely lists all compilers that we tried, known to work, with tests all passing, and the configurations we are eager to support. Anything unlisted may be supported if someone contributes a PR, then updates the support matrix accordingly.

If you want, I can create a new issue and work on it.

lalitb:
> If you want, I can create a new issue and work on it.

Thank @iblancasa, it would be nice to have. Probably this can replace the current section for [supported platforms](https://github.com/open-telemetry/opentelemetry-cpp#supported-development-platforms), as that only covers the platforms part of our CI pipeline.

maxgolov:
> If you want, I can create a new issue and work on it.

It would be great to have a good matrix document. I am not against another alternate approach that may ensure that we require C++11 as a minimum. Maybe solution that uses `CMAKE_CXX_KNOWN_FEATURES` only initially to check if compiler has all the C++11 features we need, and if not - failing the build early with a descriptive error.

maxgolov:
@iblancasa - I assume we can close this PR without merging for now? I see you started working on the other issue to document the compilers. I can help a bit with validation using those docker images we have in repo today, to confirm what works with what `standard toolchain` that comes with modern Ubuntu distros. It'd be also great to have similar setup / coverage for RedHat distros.

PR #745, opened 2021-05-12
Add preview feature-flag/macro for metrics 

lalitb:
## Changes

The metrics implementation is not complete ( and may need a complete rewrite ), so it needs to be behind the feature-flag as we are planning for 1.0.0 RC ( 1.0.0 release is only for tracing signal, not metrics and logs).

As per the feature flag policy defined in [Versioning.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/Versioning.md#policy) - This PR just wraps all the metrics code (api, sdk, tests, exporter, example ) under ENABLE_METRICS_PREVIEW hashdef.
```
#ifdef ENABLE_METRICS_PREVIEW
..
..
#endif
```

The changes look huge as the clang-format does the indentation of preprocessor directives within this wrapper.


reyang (approved):

maxgolov:
I'm OK with this change in general. But it could have also been handled by a feature flag in CMake, i.e. NOT including any of these files as part of the build. It would have been a single option and a relatively isolated change in `CMakeLists.txt`, without needing to change the source code at all (because you are excluding files entirely). I would have done this differently for CMake..

I approve this only because I do not know how to handle the exclusion of directory in Bazel in more flexible way. Maybe you could have simply added the entirety of metrics to `.bazelignore`, because it is subject to rework anyways, i.e.
`sdk/test/metrics
...
`
to `.bazelignore`, and a corresponding `WITH_METRICS` flag to the root level `CMakeLists.txt`.

maxgolov (approved):

exporters/prometheus/test/prometheus_exporter_utils_test.cc
```diff
@@ -469,3 +470,9 @@ TEST(PrometheusExporterUtils, TranslateToPrometheusMultipleAggregators)
                prometheus_client::MetricType::Gauge, 1, vals);
 }
 OPENTELEMETRY_END_NAMESPACE
+#else
```
maxgolov, line 473: Yeah, you wouldn't have needed it if you excluded the entire folder. It would've been just 2-3 files change instead of 51.

lalitb:
@maxgolov Yes it would be definitely easy to ignore entire 'metrics' related folders in CMake build and also probably some way to do through bazel. But our releases are source archives, and idea is to add preview flag explicitly in source to ensure that metrics is not part of build irrespective of whatever build system is used for this released source (not just bazel or cmake).

pyohannes:
Could we maybe additionally ignore folders related to metrics? Otherwise we end up with [empty targets for metrics](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/CMakeLists.txt#L1), which might be confusing for users.

pyohannes:
Also, we should think about enabling `ENABLE_METRIC_PREVIEW` for our CI builds. Otherwise we won't catch regressions in the metric code.

lalitb:
> Could we maybe additionally ignore folders related to metrics? Otherwise we end up with [empty targets for metrics](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/CMakeLists.txt#L1), which might be confusing for users.

Yes Johannes, I did realize that after raising the PR. As you mentioned, better would be enabling metrics preview for our CI builds.

api/test/metrics/meter_provider_test.cc
```diff
@@ -39,3 +40,9 @@ TEST(Provider, MultipleMeterProviders)
 
   ASSERT_NE(Provider::GetMeterProvider(), tf);
 }
+#else
+TEST(Provider, DummyTest)
+{
+  // empty
```
ThomsonTan, line 46: Why do we need the dummy test if metrics is not enabled in build?

ThomsonTan (approved):

api/include/opentelemetry/metrics/async_instruments.h
```diff
@@ -1,7 +1,8 @@
-#pragma once
+#ifdef ENABLE_METRICS_PREVIEW
+#  pragma once
```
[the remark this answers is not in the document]
lalitb: Yes, good point. have changed it now :) Had written a bash script to automate this task, and it was easy to put ifdef as the first and last uncommented line :)

api/test/metrics/meter_provider_test.cc
lalitb: It was required as test targets were failing without that. Now I have changed the cmake conf to ignore metrics-specific directories if `WITH_METRICS_PREVIEW` is not set. So could remove these empty tests also

pyohannes (approved):

PR #747, opened 2021-05-13
#574: add instructions about how to build with Bazel

### LINKED ISSUE #574: Add Installation instructions for Bazel.
The install doc is empty which needs to be filled in with details.

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md

Similar issue for cmake: #564

iblancasa:
Fixes #574

## Changes
Add some instructions about how to build with Bazel.


lalitb:
Thanks @iblancasa for the PR. We can get some feedback from @eyjohn and @jsuereth also here. As @eyjohn had some plans to reorganize the build/install document for cmake and bazel.

INSTALL.md
```diff
@@ -134,4 +134,55 @@ target_link_libraries(foo PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES})
...
+   ...
+   Resolving deltas: 100% (3225/3225), done.
+   $
+   ```
+
+2. Download the dependencies and build the source code:
+
+   ```console
+   $ bazel build //...
+   bazel build -- //... -//exporters/otlp/... -//exporters/prometheus/...
```
maxgolov, line 172: Technically the exclusion here isn't right. It all builds in its entirety right now, including `otlp` - which is going to be the main thing. Helper script for windows:
https://github.com/open-telemetry/opentelemetry-cpp/blob/414a7314f7ab343533042f70eaea5810764fdd5d/tools/build-bazel.cmd#L4

maxgolov:
I think you should be good to re-add `otlp` exporter. If you fix that, it looks good then. Approved with suggestion.

INSTALL.md
maxgolov, line 172:
```suggestion
   bazel build //...
```

ThomsonTan: How about just invoke `tools/buildbazel.cmd` here which will keep both consistent?

maxgolov (approved):

maxgolov:
> CI / markdown-lint (pull_request) Failing after 7s — markdown-lint

Please run the linter.

INSTALL.md
maxgolov: Maybe we need to add a script for Linux then, like `tools/build-bazel.sh` . But I actually like this document too.

INSTALL.md
```diff
@@ -134,4 +134,55 @@ target_link_libraries(foo PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES})
...
+Refer to [Platforms Supported](./README.md#supported-development-platforms)
+for more information.
+- A compatible C++ compiler supporting at least C++11.
+Major compilers are supported.
+Refer to [Supported Compilers](./README.md#supported-c-versions) for more information.
+- [Git](https://git-scm.com/) for fetching opentelemetry-cpp source code from repository.
+To install Git, consult the [Set up Git](https://help.github.com/articles/set-up-git/)
+guide on GitHub.
+- [Bazel](https://www.bazel.build/) for building opentelemetry-cpp API,
+SDK with their unittests. We use 3.7.2 in our build system.
```
maxgolov, line 149: Can we somehow omit `3.7.2.` here. It's 10 seconds to follow the guidelines, even if we don't mention the version explicitly. Maybe suggesting to check what version is being used:
https://github.com/open-telemetry/opentelemetry-cpp/blob/414a7314f7ab343533042f70eaea5810764fdd5d/.bazelversion#L1

Something like: **Please install the exact version of bazel mentioned in `.bazelversion`** ?


mgolovanov@xps:/mnt/c/work/opentelemetry-cpp.main$ bazel build //...
ERROR: The project you're trying to build requires Bazel 3.7.2 (specified in /mnt/c/work/opentelemetry-cpp.main/.bazelversion), but it wasn't found in /usr/bin.

You can install the required Bazel version via apt:
  sudo apt update && sudo apt install bazel-3.7.2

INSTALL.md
maxgolov: OTLP-gRPC takes forever to build though 😄

lalitb (approved):

eyjohn (approved):
Thanks for this contribution, it covers the process well.

I still have (slow) plans to restructure the install/usage docs focused on using otel-cpp in projects (which may/not include installing it) but I think this will build on top of these base instructions.

lalitb:
@iblancasa - This is good to merge once the CI check for markdown lint is fixed. CI uses [markdownlint-cli](https://www.npmjs.com/package/markdownlint-cli) to validate the document using [this ](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/.markdownlint.json) configuration.

INSTALL.md
```diff
@@ -133,4 +133,55 @@ target_link_libraries(foo PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES})
...
+
+To install Bazel, consult the [Installing Bazel](https://docs.bazel.build/versions/3.7.0/install.html) guide.
+
+### Building as Standalone Bazel Project
```
jsuereth, line 152: Isn't this more useful for contributors, less so installation?


E.g. I'd expect most bazel users to want  to do something like pulling in OTel CPP via `WORKSPACE` definition using `http_archive` or some other mechanism.

Not that this isn't useful, just not how I'd expect folks to consume.

jsuereth (commented):
One meta question on instructions, otherwise, thanks much for taking a crack at this!

INSTALL.md
jsuereth: Yeah, I think we should likely encourage folks to install/use Bazelisk, as it can auto-correct versions on the fly and leads to a much smoother developer experience when you move back/forth between different bazel projects.

INSTALL.md
jsuereth: e.g. https://github.com/GoogleCloudPlatform/opentelemetry-operations-cpp/blob/main/WORKSPACE#L46

Or you can see:

https://github.com/googleapis/google-cloud-cpp/pull/1230/files for an example usage of OpenCensus C++

lalitb: This is a good point. We can have a separate section (say) "Incorporating Into An Existing Project", which includes the above-mentioned steps, similar to how we document for `CMake`.

lalitb:
@iblancasa  - we would like to pull these changes in. But the markdown-lint check is failing for now. Once you fix that, we can merge the changes.

PR #748, opened 2021-05-13
Upgrade Google Benchmark to latest v1.5.3 in git submodule and for CMake-vcpkg build on Windows

maxgolov:
## Changes

- Upgrade Google Benchmark to latest release [v1.5.3](https://github.com/google/benchmark/releases/tag/v1.5.3)

- vcpkg on Windows in their mainline repo still has some build and installation issues with Google Benchmark since v1.5.1. We were previously building it via our custom patched port file to overcome these mainline issues. Now I created a new fresh port file that works well for vcpkg on Windows. My port is implementing a temporary patch for this issue here: https://github.com/google/benchmark/issues/1046 ... And also resolves some funky build issue with `CMake+ninja` on Windows, where CMake cannot find vcpkg package because it has version with arch bitness suffix, like `1.5.3 (64bit)`, that completely confuses CMake. Since `vcpkg` maintains all packages in their own directory split by architecture `[x64-windows|x86-windows]`, the arch check of version is completely unnecessary. I need this patch to illustrate the use of Ninja build on Windows in a follow-up PR.

I'm considering to send in the port file above into mainline `vcpkg` as well. They are seemingly stuck on version from 2020 right now. It'd be of a great benefit for them to do the upgrade to latest. But since it'd take a bit longer to commit to `vcpkg`, and no energy from my end to drive it in there just yet - I'm applying our local port fixes as overlay here only.


lalitb (approved):
LGTM

ThomsonTan (approved):

PR #755, opened 2021-05-14
How to build FAST with Visual Studio 2019+CMake+Ninja on Windows

maxgolov:
## Changes

This is a set of changes + documentation needed to switch from `MSBuild` to `Ninja` for Windows build.

What it gives:
- it is dramatically improving your productivity: gives you back 4 minutes of your lifetime for every clean build. Timing details are in the markdown doc.
- your incremental builds now are seconds (5s per build configuration).
- if you use Visual Studio IDE, now your test run time for selective tests are also seconds. Markdown doc shows how.

This is rather trivial change. No code changes. Just build infra. See `building-with-vs2019.md` markdown for all the details.

I'll send a separate follow-up PR that adds this process ( `setup-buildtools.cmd+build.cmd`) as part of our CI/CD loop, with a goal in mind to continuously validate both `nostd` and `stdlib` build flavor on **Windows**. This process is flexible, as it supports any Visual Studio version and any compiler. In fact, it even works with CLang-LLVM if Visual Studio is not installed (verified that). I have not tested if similar process works for MinGW or Emscripten. But from previous personal experience, same build script should largely work with minor adjustments for other compilers. Linux process doc comes after in a separate PR.


CMakeLists.txt
```diff
@@ -84,6 +92,12 @@ option(WITH_JAEGER "Whether to include the Jaeger exporter" OFF)
 
 option(BUILD_TESTING "Whether to enable tests" ON)
 if(WIN32)
+  if(BUILD_TESTING)
+    if(MSVC)
+      # GTest bug: https://github.com/google/googletest/issues/860
+      add_compile_options(/wd4275)
```
ThomsonTan, line 98: Could we scope this option to only test code compilation?

CMakeLists.txt
```diff
@@ -65,7 +65,15 @@ if(WITH_STL)
 
   # Optimize for speed to reduce the hops
   if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    set(CMAKE_CXX_FLAGS_SPEED "/O2")
+    if(CMAKE_BUILD_TYPE MATCHES Debug)
+      # Turn off optimizations for DEBUG
+      set(CMAKE_CXX_FLAGS_SPEED "/Od")
```
ThomsonTan, line 70: Sounds like `CMAKE_CXX_FLAGS_SPEED` is just used as a temporary variable? Could we give it another name as the `CMAKE_` prefix may imply it is used by CMake?

CMakeLists.txt
maxgolov: Unfortunately the issue is that we have so many different test projects, and it appears like it would have to be applied to all project that includes Google Test. The fix is in the latest Google Test. But that latest Google Test is not getting pulled via `vcpkg`, so I see this error. It'd have to be added to many files. I can do that if you insist

CMakeLists.txt
maxgolov: Yes, I agree and will adjust that.

maxgolov: Changed the name of it to `MSVC_CXX_OPT_FLAG`

lalitb (approved):
LGTM

ThomsonTan (approved):

PR #758, opened 2021-05-15
Populate resource to OTLP proto data

### LINKED ISSUE #757: Resource is not populated to OTLP proto data
In previous PR https://github.com/open-telemetry/opentelemetry-cpp/pull/706, resource is passed to exporter recordable structure, but it is still not populated to `proto::collector::trace::v1::ExportTraceServiceRequest`, so the OTLP receiver will not get any resource from the OpenTelemetry C++ SDK.

**Steps to reproduce**
Run example_otlp and OpenTelemetry collector with OTLP collector.

We should see "Resource labels" in the receive data.

No "Resource labels" sent to OpenTelemetry collector.

**Additional context**

ThomsonTan:
Fixes #757 

## Changes

Populate resource to OTLP proto data and also update the readme for OTLP receiver with easier instructions to run and test it end-to-end with docker.


exporters/otlp/src/otlp_exporter.cc
```diff
@@ -28,8 +28,18 @@ void PopulateRequest(const nostd::span<std::unique_ptr<sdk::trace::Recordable>>
...
     *instrumentation_lib->add_spans() = std::move(rec->span());
+
+    if (!has_resource)
+    {
+      std::unique_ptr<proto::resource::v1::Resource> proto_resource{
+          new proto::resource::v1::Resource};
+
+      rec->PopulateProtoResource(proto_resource.get());
+
+      resource_span->set_allocated_resource(proto_resource.release());
```
lalitb, line 40: nit - Just wondering if line 35 - 40 can be replaced with single line:

``
resource_span->mutable_resource() = rec->ProtoResource();
``

using older implementation of `ProtoResource`.

lalitb (approved):
Thanks for adding the support. Looks good, perhaps we can add unit tests for the resources too.

exporters/otlp/src/otlp_exporter.cc
ThomsonTan: I chose `set_allocated_resource` which could avoid allocate the `Resource` twice or more, but yes, it look simpler with `mutable_resource`. Updated the PR to the older implementation of `ProtoResource` as this optimization may not deserve the extra complexity.

ThomsonTan:
> Thanks for adding the support. Looks good, perhaps we can add unit tests for the resources too.

Added unit resource test to OTLP recordable test case. We may need more tests if we'd like to validate the OTLP exported result.

PR #761, opened 2021-05-16
Rename Recordable under OTLP exporter to OtlpRecordable

ThomsonTan:
## Changes

There is `class Recordable` from API, SDK and exporters. Share the same name could cause confuse and hinder extending it with multiple variants in one exporter.

### DIFF
--- a/exporters/otlp/BUILD
+++ b/exporters/otlp/BUILD
@@ -22,9 +22,9 @@ cc_library(
         "src/otlp_recordable.cc",
     ],
     hdrs = [
+        "include/opentelemetry/exporters/otlp/otlp_recordable.h",
         "include/opentelemetry/exporters/otlp/protobuf_include_prefix.h",
         "include/opentelemetry/exporters/otlp/protobuf_include_suffix.h",
-        "include/opentelemetry/exporters/otlp/otlp_recordable.h",
     ],
     strip_include_prefix = "include",
     deps = [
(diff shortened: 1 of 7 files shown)


maxgolov (approved):
I'm Okay with this change. But do you think a similar change has to happen / applied to all other exporters? I'm a bit hesitant to do this for ETW exporter at this time.

ThomsonTan:
@maxgolov I think this change is not a must have. Zipkin has part of the change, and I will do the same thing for Jaeger in another PR. It should fine for ETW to have its own convention is it doesn't follow the current Recordable approach.

maxgolov:
My main logical reasoning is to avoid the churn for now with first customers onboarding. I can consider refactoring and providing an alias when the dust settles.

lalitb:
Should we make it slightly more granular - say `OtlpgRPCRecordable ` (to indicate it is OTLP over gRPC), and later it is also possible to add Recordable for OTLP over HTTP.

ThomsonTan:
> Should we make it slightly more granular - say `OtlpgRPCRecordable ` (to indicate it is OTLP over gRPC), and later it is also possible to add Recordable for OTLP over HTTP.

This is a good point. Support other protocols is optional, and also there could be refactor like extracting a base recordable for all OTLP recordables, so I'd suggest leave this to future extension. I would not insist the current naming in this PR, but `OtlpgRPCRecordable` looks a little over complicated for now. How do you think @lalitb.

jsuereth (approved):

PR #771, opened 2021-05-20
Switch from MPark variant to `absl::variant` as default

### LINKED ISSUE #572: Switch to abseil::variant as default variant class across api

maxgolov:
Fixes #572 #733

## Changes

Why:
- MPark Variant is:
a) Boost-licensed not Apache License v2 (forcing us to depend on Boost License unnecessarily)
b) Does not compile with Visual Studio 2015 at all.

How to fix:
- use `absl::variant` , because it is:
a) Apache License v2
b) It compiles well with Visual Studio 2015 Update 2.

**UPDATE 1:** there is a change needed in `common::AttributeValue` to re-add `const char *`, because neither `std::variant` nor `absl::variant` provide an implicit conversion of that type to `string_view`. What happens then, it by-default gets converted to `bool`. So any static initializer in the map, for a pair such as `{{"key, "value"}}` is gonna break, by assigning the variant value with type `bool` instead of `string_view`. I'm implementing a fix to resolve this. Related issue #733 #734

**UPDATE 2:** ~~unfortunately we also need to add `const char *` attribute value constructor to `OwnedAttributeValue` since otherwise it'd break the SDK `ResourceAttribute` implementation. In most cases the exporters do not have to deal with `const char *` because transform from `AttributeValue` variant to `OwnedAttributeValue` variant presently implies a memcpy of `const char *` into owning `std::string` object. However, I did adjust all default exporters to potentially deal with `const char *`. This is 3 lines of code per exporter, and I would argue the maintenance cost of that is zero.~~

**UPDATE 3:** - as discussed, I removed `const char *` from `OwnedAttributeValue`. I also refactored `ResourceAttribute` to use a map that allows to transform incoming `const char *` attributes assigned using initializer lists into `std::string`.

I added a separate document that explains the process of updating Abseil snapshot in this PR #797


lalitb (approved):
LGTM. Thanks for the cleanup :)

ThomsonTan (dismissed):

ThomsonTan:
Probably `#include "absl/base/config.h"` should be `#include "../types.config.h"` in `api/include/opentelemetry/nostd/absl/types/variant.h`. Does this only fail CI build? @maxgolov

maxgolov:
@lalitb @ThomsonTan ..

Folks, please don't merge it yet. I pushed the latest update. Unfortunately there are two outstanding issues:

- it appears that `absl::variant` has the same issue we discussed here that happens with `std::variant`. #733 #734

Because I see the same EXACT tests failing now with Abseil, because the variant does not copy temporary 😢

Screenshot:

image

- the other second issue is shipping `absl/` directory inside `api/include/nostd`. I might need to fix paths in our snapshot of Abseil, to use paths relative to `api/include/opentelemetry` , because I am not sure yet how we can add include directory of `api/include/opentelemetry/nostd` to dependent projects. There is also something related to being able to build with original Abseil. What I was trying to avoid is linking to original Abseil. Because Abseil variant implementation otherwise works great as a header-only library. The only problem with that is implementation of exception handler.

I need a bit more work on that. I might need an advise from @owent how to export our local Abseil headers, in case if we bundle Abseil within, not needing an external snapshot of it.. to keep the entire API still pure header-only. That's a requirement from someone else. And I need the same (header-only API) for ETW exporter build.

maxgolov:
I can try to fix the tests by making sure we pass `std::string` rather than `const char *`. But we kinda have a long-term issue on API surface now, that's maybe because we did not add `const char *` to variant... 🤷 ... If we had it on `AttributeValue`, then I guess we wouldn't have had the implicit conversion to `bool` issue?? I can try patching that..

maxgolov:
I got down to only 5 test failures by re-adding `const char *` to `AttributeValue`. It solves the original issue described in #733 #734. The change is NOT SMALL, and it's been a struggle so far. I'm pushing the latest update. The only remaining tests that are failing are now isolated to `resources.ResourceTest.*` , all other tests are now passing.

Basically, for both - `std::variant` since C++17, and for `absl::variant`, in order to provide convenient initialization from temporary `const char *`, we have to re-add `const char *` explicitly into variant. Because it does not get implicitly converted to `nostd::string_view`, and falls back to `bool`. Which breaks a simple construct like `{ { "key, "value } }` in static initializer lists.

Refreshed PR with latest stuff, only 5 tests are now failing, down from 10.

owent:
> @lalitb @ThomsonTan ..
> 
> Folks, please don't merge it yet. I pushed the latest update. Unfortunately there are two outstanding issues:
> 
> * it appears that `absl::variant` has the same issue we discussed here that happens with `std::variant`. #733 #734
> 
> Because I see the same EXACT tests failing now with Abseil, because the variant does not copy temporary 😢 ... OR treats it as `bool` type, whatever it is, it messes up the outcome.
> 
> Screenshot:
> 
> image
> 
> * the other second issue is shipping `absl/` directory inside `api/include/nostd`. I might need to fix paths in our snapshot of Abseil, to use paths relative to `api/include/opentelemetry` , because I am not sure yet how we can add include directory of `api/include/opentelemetry/nostd` to dependent projects (which would be required for `#include "absl/base/config.h" to work, I temporarily added a hack in root-level `CMakeLists.txt` to run the tests, but that hack is not good as a permanent solution).
> 
> There is also something related to being able to build with original Abseil. What I was trying to avoid is linking to original Abseil. Because Abseil variant implementation otherwise works great as a header-only library. The only problem with that is implementation of exception handler.
> 
> I need a bit more work on that. I might need an advise from @owent how to export our local Abseil headers, in case if we bundle Abseil within, not needing an external snapshot of it.. to keep the entire API still pure header-only. That's a requirement from someone else. And I need the same (header-only API) for ETW exporter build. It would be sad to lose that pure header-only feature of API target.

We could add these cmake scripts below into ```api/CMakeLists.txt``` to replace ```include_directories(api/include/opentelemetry/nostd)``` .

```cmake
if(WITH_ABSEIL)
  target_compile_definitions(opentelemetry_api INTERFACE HAVE_ABSEIL)
  target_link_libraries(opentelemetry_api INTERFACE ${CORE_RUNTIME_LIBS})
else()
  target_include_directories(
    opentelemetry_api
    INTERFACE
      "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include/opentelemetry/nostd>"
      "$<INSTALL_INTERFACE:include/opentelemetry/nostd>")
endif()
```

maxgolov:
@owent  - I just pushed an update to use relative paths for local Abseil. I still leave an option for end-users / devs to decide if they want the full Abseil, e.g. `WITH_ABSEIL`. I like your suggestion. I do not exactly like what I did with relative paths in latest...

exporters/otlp/CMakeLists.txt
```diff
@@ -37,12 +37,13 @@ if(BUILD_TESTING)
...
   if(MSVC)
     add_definitions(-DGTEST_LINKED_AS_SHARED_LIBRARY=1)
-  endif()
-  if(GMOCK_LIB)
-    # unset GMOCK_LIB to force find_library to redo the lookup, as the cached
-    # entry could cause linking to incorrect flavor of gmock and leading to
-    # runtime error.
-    unset(GMOCK_LIB CACHE)
+  else()
+    if(GMOCK_LIB)
```
maxgolov, line 41: @ThomsonTan - Tom, sorry, I kept it in here because I still have that issue popping up occasionally. I don't know why :(

owent:
> @owent - I just pushed an update to use relative paths for local Abseil. I still leave an option for end-users / devs to decide if they want the full Abseil, e.g. `WITH_ABSEIL`. I like your suggestion. I do not exactly like what I did with relative paths in latest...

I like the `WITH_ABSEIL` option to allow me to use my own abseil.But I don't see any options for bazel. Should `opentelemetry_api` use abseil imported by gRPC when using bazel?

maxgolov:
>  Should opentelemetry_api use abseil imported by gRPC when using bazel?

I think it's a good call. Because gRPC uses Abseil anyways, if we are building SDK optimized for that same Abseil, we do not bloat with unnecessary local copy. What I am not certain --- since this is in API, and since this is in the main part of API -- `common::AttributeValue` used for Spans, Events, etc.. How do we pre-package the necessary headers for customer to use, OR do we need to document that the customer themselves need to take a dependency on that exact same version of Abseil that gRPC does? I am not sure how we neatly document that..

lalitb:
> Basically, for both - `std::variant` since C++17, and for `absl::variant`, in order to provide convenient initialization from temporary `const char *`, we have to re-add `const char *` explicitly into `common::AttributeValue` variant on API surface. Because otherwise without that explicit conversion - it get implicitly converted to `bool` instead of `nostd::string_view`..



> I got down to only 5 test failures by re-adding `const char *` to `AttributeValue`. It solves the original issue described in #733 #734. The change is NOT SMALL, and it's been a struggle so far. I'm pushing the latest update. The only remaining tests that are failing are now isolated to `resources.ResourceTest.*` , all other tests are now passing.
> 
> Basically, for both - `std::variant` since C++17, and for `absl::variant`, in order to provide convenient initialization from temporary `const char *`, we have to re-add `const char *` explicitly into `common::AttributeValue` variant on API surface. Because otherwise without that explicit conversion - it get implicitly converted to `bool` instead of `nostd::string_view`..
> 
> Which breaks a simple construct like `{ { "key, "value } }` in static initializer lists. I don't think this is good dev experience at all.
> 
> Refreshed PR with latest stuff. Only 5 tests are now failing, down from 10. Hopefully I can sort out the rest of them in a day.

FYI - this issue has been fixed in later versions of std::variant as below:

Resolution is described here: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0608r3.html

And this issue (P0608R3) is fixed in:
 clang 9.0 - https://libcxx.llvm.org/docs/Cxx2aStatus.html, 
 GCC 10.1 - https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2020
microsoft/STL - https://github.com/microsoft/STL/pull/1629

I remember there were ownership issues pointed out by @pyohannes earlier for adding support for `const char *` in `AttributeValue`.  Will try to dig more on that.  Should we switch back to default mpart::variant in that case, as ideally, the fix should come from these libraries?

lalitb:
> Will try to dig more on that.

The earlier discussion for the context :

https://github.com/open-telemetry/opentelemetry-cpp/pull/519#discussion_r561368313

maxgolov:
@lalitb - it would take years before we get to `C++20`. I verified the fix with gcc and msvc in this PR. Will fix Bazel separately.

lalitb:
> @lalitb - it would take years before we get to `C++20`. I verified the fix with gcc and msvc in this PR. Will fix Bazel separately.

how about switching back to mpark::variant, if that way we can remove `const char *` from AttributeValue. Sorry if I am missing the context, but what benefit we are getting with absl:::variant over mpark::variant here?

maxgolov:
> how about switching back to mpark::variant, if that way we can remove const char * from AttributeValue

We can't. It does not work with Visual Studio 2015.

maxgolov:
All tests pass with msvc and gcc.
There's only one issue with Mac OS X Apple clang, related to passing attributes via initializer list.
I'll take a look at it tomorrow.

lalitb:
> Why are you opposed to `const char *`? It seems to be explicit and natural to me. I fixed all the tests.

Yes, `const char *` will indeed fix the tests.  My concern is it just makes the system more error-prone as we don't have control of when the pointer gets deleted. Specifically, in batch span-processors or batch exporters where the actual export happens in batches, we don't know how long the pointer will remain valid. Also, it would be good to have a minimal set of possible attribute-types.

As I understand these are the possibilities:
1. Switch to absl::variant, and add `const char *` as another type in AttributeValue. This is current PR.
2. Switch to absl::variant and let devs explicitly convert char pointers or string literals to std::string before calling our api to avoid unexpected results. This can be documented.
3. Keep two variants of  `variant` as it exists currently - `mpark::variant` as default, and `absl::variant` for vs 2015.
4. Switch to another variant implementation which supports char * to std::string conversion say from Micrsoft/STL - this is going to be a bigger task and not preferable specifically near GA.

Again I don't have a strong opinion against `const char *` (first option above) and fine if we all agree with this.

maxgolov:
> Yes, const char * will indeed fix the tests. My concern is it just makes the system more error-prone as we don't have control of when the pointer gets deleted. Specifically, in batch span-processors or batch exporters where the actual export happens in batches, we don't know how long the pointer will remain valid.

This is quite artificial concern. `const char *` exists only on API surface `common::AttributeValue`. Recordable does not require `const char *` per se.

> Switch to absl::variant and let devs explicitly convert char pointers or string literals to std::string before calling our api to avoid unexpected results. This can be documented.

This is very unnatural. In most cases this is against what a developer would expect our API to look like. Please use `nlohmann/json` as example: you can initialize your JSON objects with key-value from a pair of `<const char* key, const char *value>`. I think this is great pattern. You should naturally expect that your `string` in C++ can be initialized from `const char *`, this is given. We should not try to take away that _convenience_.

> Keep two variants of variant as it exists currently - mpark::variant as default, and absl::variant for vs 2015.

This defeats the purpose of **ABI compatibility** on Windows. If some vendor requires ABI stable SDK on Windows for dynamic loading, then they have to use the lowest of vc14x ABI stable primitives. That will be `absl::variant` only (because MPark won't compile). And they'd most likely build a DLL with Visual Studio 2015 to cover compatibility with 2015/2017/2019, avoiding the use of MPark Variant and avoiding the use of STD lib.

maxgolov:
I'm nearly there. I'll fix Mac OS X, we get to `100% pass rate`, we can discuss. I would love to address your concern about the extra type on API surface; at the same time there is no impact to `Recordable` and `OwningAttribute` - those internal SDK types behind the ABI boundary, used by async batching exporters can have `std::string` storage. Don't need `const char *`. So the extent of change is limited to adding `const char *` as C-string type on API. I'd argue it's a `string` type - totally within the constraints of OpenTelemetry specification.

maxgolov:
All tests are now passing.

maxgolov:
@jsuereth @eyjohn - I would like to get some feedback from you. The PR is big.

It addresses the two key issues:
- provides portable variant implementation that works for ALL compilers, **including vs2015** (MPark won't work there at all).
- implements the changes necessary to allow STDLIB build with C++17 or C++20 `std::variant`. That implementation, albeit flawed, comes with any today's compiler. And it requires `const char *` on API surface, in order to disambiguate with `bool` type. Without that - the parameters passed as `const char *` or string literals, would be treated as `bool` (wrong).

Issues we discussed with @ThomsonTan during SIG meeting:
- [x] removing MPark variant.
- [x] make absl::variant default; local header-only snapshot of Abseil (`OTABSL_*` prefix) - in order to avoid clash with Abseil library used in gRPC / protobuf-generated code. I will submit a follow-up PR with a script and process how to refresh our own local header-only snapshot of Abseil.
- [x] optimization option to reuse Abseil that comes with gRPC. This option should work in case if we are statically linking the SDK. That'll minimize the library size. But it won't be ABI-stable. @owent
- [x] adding C-string (const char *) on API surface. Not adding to `OwnedAttributeValue`. This is required to avoid 'wrong' conversion to `bool` that could happen in stock implementation of C++17 `std::variant`
- [ ] follow-up PR : add build loop for vs2015+nostd
- [ ] follow-up PR : add build loop for vs2017+stdlib
- [ ] follow-up PR : add build loop for vs2019+stdlib
- [ ] follow-up PR : add build loop for gcc-latest (C++20) + stdlib
- [x] remove notion of Boost license. All code now is Apache License v2. No Boost in API headers.

**Separate topics not covered right now**
- [Tom] Support VS 2022.

Going forward it would make sense for Windows OpenTelemtry C++ SDK **.dll flavor** be built with vs2015+Abseil+other nostd classes. That'd keep it compatible with vs2017, vs2019, and (hopefully) vs2022... Whereas statically linked / header-only customers would be best with pure STDLIB, no `nostd`.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -29,6 +29,7 @@ using OwnedAttributeValue = nostd::variant<bool,
                                            uint32_t,
                                            int64_t,
                                            double,
+                                           const char *,
```
lalitb, line 32: Is this intended. As char pointer is not owned attribute?

maxgolov: You are right. It should not be needed because we have a converter to translate `common::AttributeValue(const char *)` to `OwnedAttributeValue(std::string)`. Let me double-check again, and I should be able to remove that.. What I don't remember - if you are using this class for `Resrouce`? If you are, then it's still needed. Resource, however, has a separate method in constructor that allows to transform all `const char *` to `std::string`.

maxgolov: Yeah, you are using it for `ResourceAttributes` here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/12e56f9f4ad2e2a870482d89b999a9cf7c493e85/sdk/include/opentelemetry/sdk/resource/resource.h#L22

So if we do not add it to `OwnedAttributeValue`, then only `ResourceAttributes` scenario is affected. And you'd still hit the issue described in #733 !

However, when all exporters deal with `OwnedAttributeValue`, - I believe they should not be receiving any objects that contain `const char *` . Because in the transformer from `AttributeValue` to `OwnedAttributeValue` we now initialize the string value as `std::string` here:

https://github.com/open-telemetry/opentelemetry-cpp/blob/fb5393b1c4151ef3a9dcc059c1c8973191853dca/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L80

That means to downstream exporters, they'd never normally have to deal with `const char *` much. So I would rather keep it. Or if you have alternate suggestions how to solve the `ResourceAttributes` issue in a nicer way, that'd be much appreciated?

One issue that I think we may want to consider, is if in future we allow initialization from map of `const char *` (which in my opinion, is natural and expected, and most people would like to see that) - **we could have avoided a memcpy in synchronous non-batching exporter scenario, akin to ETW - you do not have to own the attribute value when it is immediately serialized on user thread**.

maxgolov: So I'm not changing it, because changing it breaks the `ResourceAttributes` scenario.

api/include/opentelemetry/std/span.h
```diff
@@ -26,6 +26,9 @@
 #      define HAVE_SPAN
 #    endif
 #  endif
+#  if !__has_include(<gsl/gsl>)
```
maxgolov, line 29: @lalitb #813 - if it's not available, the option is undefined.

lalitb: Thanks @maxgolov. Seems fine. We can discuss this PR changes (i.e making absl::variant as default) in today's community meeting.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
lalitb: @maxgolov  - Would you be removing `const char *` from `OwnedAttributeValue` in this PR, as we are changing `ResourceAttributes` implementation to directly use` common::AttributeValue` (#817- thanks for raising this issue) in separate PR? I understand we may have to ( temporarily ) convert resource string literals to std::string() in unit-tests which should be fine right?

CMakeLists.txt
```diff
@@ -125,13 +129,13 @@ option(
   OFF)
 option(WITH_EXAMPLES "Whether to build examples" ON)
 
-option(WITH_METRICS_PREVIEW "Whether to build metrics preview" OFF)
+option(WITH_METRICS_PREVIEW "Whether to build metrics preview" ON)
```
maxgolov, line 132: Apologies for enabling metrics and logs preview, just want to make sure that ALL of existing code successfully compiles with the new implementation of variant.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -114,19 +115,32 @@ class AttributeMap
         });
   }
 
+  // Construct map from initializer list by applying `SetAttribute` transform for every attribute
```
maxgolov, line 118: @lalitb - I switched Resource Attributes to use this `AttributeMap` class we had in common utils. I also added this method here that safely transforms from non-owning API/ABI stable collection of `common::AttributeValue` to owning `OwnedAttributeValue`, using `nostd::visit` converter that ensures that for each attribute - we store these as `std::string`.

I removed `const char *` from `OwnedAttributeValue` itself. Fixes #817 , as discussed during community meeting.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -114,19 +115,32 @@ class AttributeMap
...
+  // Convert non-owning key-value to owning std::string(key) and OwnedAttributeValue(value)
   void SetAttribute(nostd::string_view key,
                     const opentelemetry::common::AttributeValue &value) noexcept
   {
-    attributes_[std::string(key)] = nostd::visit(converter_, value);
+    (*this)[std::string(key)] = nostd::visit(converter_, value);
   }
 
 private:
-  std::unordered_map<std::string, OwnedAttributeValue> attributes_;
```
maxgolov, line 129: I removed internal `std::unordered_map` here. It is much nicer to keep all characteristics of `unordered_map` retained on `AttributeMap` object itself, i.e. you can now range-for-loop iterate over that, etc.. I see this as a reusable helper class we can consume within the SDK more often.

sdk/include/opentelemetry/sdk/resource/resource.h
```diff
@@ -18,8 +21,7 @@ namespace sdk
 namespace resource
 {
 
-using ResourceAttributes =
-    std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>;
+using ResourceAttributes = opentelemetry::sdk::common::AttributeMap;
```
maxgolov, line 24: Switching to "smarter" map that performs transform of incoming `const char *` values to `std::string`.

maxgolov:
@pyohannes @lalitb - could you have a quick look?

Latest changes address the issues we discussed during SIG meeting:

- API layer `common::AttributeValue` now has `const char *`.

- SDK layer `OwnedAttributeValue` - does not need `const char*` anymore. It retains only `std::string`. Visitor pattern converter helps to transform from `const char *` (on API surface) to `std::string` (in SDK).

- I refactored Resource Attribute + revamped the existing helper class `AttributeMap`, that one now facilitates the visitor pattern transform of incoming `AttributeValue(const char *)` (API layer) into `OwnedAttributeValue(std::string)` (SDK layer). Resource Attributes now safely be initialized from initializer lists (with C-strings on them), and internally they store all values as `std::string`.

api/include/opentelemetry/std/span.h
lalitb: So, if `gsl:span` is available, and `HAVE_CPP_STDLIB`option is set, for C++11/14/17, we use gsl::span instead of nostd::span. I don't think that is the correct approach. We should fallback to nostd::span unless customer explicitly wants to use any other library?

api/include/opentelemetry/trace/tracer.h
```diff
@@ -85,6 +85,19 @@ class Tracer
...
+      nostd::string_view name,
+      std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>> attributes,
+      const StartSpanOptions &options = {}) noexcept
+  {
+    return this->StartSpan(name,
+                           nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>{
+                               attributes.begin(), attributes.end()},
+                           {}, options);
+  }
+
```
lalitb, line 100: Why is this method templatized?

lalitb:
Thanks, @maxgolov  - Looks good to me now. Had one comment related to a different issue as I think `gsl::span` would still get used if it's installed even if customer doesn't want to use it for C++11/14/17 compilers.

@pyohannes, @ThomsonTan   - would you like to review changes specific to `const char *` for common::AttributeValue, and `sdk::common::OwnedAttributeType` as what we discussed in yesterday's meeting?

CMakeLists.txt
lalitb: We don't need to enable it at CI pipeline: https://github.com/open-telemetry/opentelemetry-cpp/blob/1c70f0e45ea9009c9d5989e3c5cfcf45b8a54004/ci/do_ci.sh#L41
 
Won't that be enough for validation?

api/include/opentelemetry/std/span.h
lalitb: Ok. have gone through #821 - should be fine as long as we are documenting it well.

CMakeLists.txt
maxgolov: This is minor thing. I can revert those flags before merge.

api/include/opentelemetry/std/span.h
maxgolov: I ti

api/include/opentelemetry/trace/tracer.h
maxgolov: From what I remember - I had run into needing this for Mac OS X clang only, for `examples/http/server.cc` below.

CMakeLists.txt
maxgolov: Fixed.

api/include/opentelemetry/trace/tracer.h
lalitb: Ok. but the template parameter `T` is not used anywhere inside the function definition. Does it really need to be a template function or normal function will do?

maxgolov: Let me try it locally again before I revert this..

maxgolov: You are right. That was a left-over from the other debug effort I had to do, to ensure that the types we pass to `absl::variant` are actually well-defined in that variant. The other fix I did in `server.cc` was sufficient. I removed the unused template.

sdk/src/common/core.cc
```diff
@@ -7,8 +7,13 @@
 #include "opentelemetry/nostd/variant.h"
 // clang-format on
 
-#if defined(HAVE_ABSEIL) && !defined(HAVE_ABSEIL_VARIANT)
-
+#if defined(HAVE_ABSEIL)
```
maxgolov, line 10: @owent - once this PR merged, could you please share some instructions, did we need this for MinGW? I think maybe we're not gonna be needing this code anymore. If we are building with `HAVE_ABSEIL` in future, we can demand that the implementation of `ThrowBadVariantAccess` comes from Abseil lib linked into the project. That way we can remove it from `core.cc` here.. Basically, I'm no longer using this exception handler for our own Abseil private variant, and most likely the code between lines 10 to 36 could be simply removed (in a separate PR 😄 ).

maxgolov:
@pyohannes @ThomsonTan @jsuereth 

I believe I have addressed ALL the questions / issues. Would appreciate feedback on this. Integrating it should allow me to set up the build loops for vs2015-vs2017-vs2019, using Abseil Variant. As well as all tests for `STDLIB` are now passing, so I can add a CI loop with `STDLIB` for other platforms as well.

ThomsonTan (approved):
The PR and the `const char *` change LGTM.

PR #774, opened 2021-05-20
remove context::Token() default constructor since it's implicitly deleted by const member

halayli:
## Changes

explicitly defining default constructor for context::Token() generates a compiler warning because it's implicitly deleted by its const member `const Context context_`


halayli:
sorry github spaces screwed up and created a garbage PR

PR #777, opened 2021-05-20
[WIP] Integrate api docs for opentelemetry-sdk to readthedocs

lalitb:
Integrate api documentation for opentelementry-sdk to readthedocs.


lalitb:
There seems to be some issues related to links and formatting, moving to draft to resolve that.

maxgolov (approved):
Quick question: so while we integrate this, is there some way we can knit the markdown contents of `/docs` folder into that the new process as well, at some point later on? There are some good write-ups we'd probably want to incorporate as `Additional documentation`.

lalitb:
> Quick question: so while we integrate this, is there some way we can knit the markdown contents of `/docs` folder into that the new process as well, at some point later on? There are some good write-ups we'd probably want to incorporate as `Additional documentation`.

@maxgolov  - Yes that's the goal, I am trying to see if each folder brings the docs/*rst files, and it should be possible to merge them in root /docs/ folder under proper sections - API docs, SDK docs, individual exporter docs, installation docs - similar to what we have for python ( https://opentelemetry-python.readthedocs.io/en/latest/ )

PR #784, opened 2021-05-21
fix for sampling root span

### LINKED ISSUE #782: Root span is not sampled correctly
TraceIdRatioBasedSampler always samples the root span because TraceId==0 when the sampling decision is made in the [C++ version of StartSpan](https://github.com/open-telemetry/opentelemetry-cpp/blob/976b5caa64ce11d175ea46a2cf7176c2657b23d5/sdk/src/trace/tracer.cc#L28).

To fix, for root spans, a TraceId should be generated before calling ShouldSample, e.g. in [the Java SDK](https://github.com/open-telemetry/opentelemetry-java/blob/b22d125acde86e3cb9d4761a69860bfe953ce307/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkSpanBuilder.java#L179).

lalitb:
Fixes #782 

## Changes

Move span_context creation from sdk::Span to sdk::Tracer, and pass it in sdk::Span constructor. This also allows passing valid trace_id of root span to Sampler, and in accordance with how other languages implement.

### DIFF
--- a/sdk/test/trace/tracer_test.cc
+++ b/sdk/test/trace/tracer_test.cc
@@ -48,6 +48,8 @@ public:
     }
     else
     {
+      // we should never reach here
+      assert(false);
       return {Decision::DROP};
     }
   }
--- a/sdk/test/trace/tracer_test.cc
+++ b/sdk/test/trace/tracer_test.cc
@@ -21,25 +21,35 @@ using opentelemetry::exporter::memory::InMemorySpanExporter;
 using opentelemetry::trace::SpanContext;
 
 /**
- * A mock sampler that returns non-empty sampling results attributes.
+ * A mock sampler with ShouldSample returning:
+ *  Decision::RECORD_AND_SAMPLE if trace_id is valid
+ *  Decision::DROP otherwise.
  */
 class MockSampler final : public Sampler
 {
 public:
   SamplingResult ShouldSample(
       const SpanContext & /*parent_context*/,
-      trace_api::TraceId /*trace_id*/,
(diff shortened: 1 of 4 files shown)


sdk/test/trace/tracer_test.cc
```diff
@@ -21,25 +21,35 @@ using opentelemetry::exporter::memory::InMemorySpanExporter;
...
+      // Return two pairs of attributes. These attributes should be added to the
+      // span attributes
+      return {Decision::RECORD_AND_SAMPLE,
+              nostd::unique_ptr<const std::map<std::string, opentelemetry::common::AttributeValue>>(
+                  new const std::map<std::string, opentelemetry::common::AttributeValue>(
+                      {{"sampling_attr1", 123}, {"sampling_attr2", "string"}}))};
+    }
+    else
+    {
+      return {Decision::DROP};
```
ThomsonTan, line 51: Should we `assert(false)` here? Seems invalid trace Id should not be passed to sampler.

lalitb: Instead of asserting in unit-test, better would be to fail the particular unit-test and let other tests run? Like in this case, `ValidTraceIdToSampler` test will fail.

ThomsonTan: Yes, fail this particular unit-test works too.

lalitb: ok got it. Have changed it.

ThomsonTan (approved):

maxgolov (approved):

PR #786, opened 2021-05-22
Add Jaeger exporter to CMake CI build

### LINKED ISSUE #785: Jaeger exporter is not enabled in CI build
Jaeger exporter is currently missing in CI build.

ThomsonTan:
Fixes #785 

## Changes

Follow the Zipkin exporter, this PR add Jaeger exporter to CMake CI build without OTLP exporter. OTLP exporter takes a lot of time to build dependencies, group all non-OTLP exporters in a separate CI build will reduce the total CI build time.


ci/setup_thrift.sh
```diff
@@ -0,0 +1,42 @@
...
+wget https://github.com/apache/thrift/archive/refs/tags/v${THRIFT_VERSION}.tar.gz
+tar -zxvf v${THRIFT_VERSION}.tar.gz
+cd thrift-${THRIFT_VERSION}
+mkdir -p out
+pushd out
+cmake -G Ninja .. \
+    -DBUILD_COMPILER=OFF \
+    -DBUILD_CPP=ON \
+    -DBUILD_LIBRARIES=ON \
+    -DBUILD_NODEJS=OFF \
```
ThomsonTan, line 30: All the options turned off here are turned on by default in thrift repo. They are not used by our C++ build so turn them off explicitly would introducing more dependence to our CI build.

https://github.com/apache/thrift/blob/70992f1e74e525461121fb9e607000b19f31a4ca/build/cmake/DefineOptions.cmake#L115

.github/workflows/ci.yml
```diff
@@ -19,7 +19,9 @@ jobs:
         sudo ./ci/setup_cmake.sh
         sudo ./ci/setup_ci_environment.sh
     - name: run cmake tests (without otlp-exporter)
-      run: ./ci/do_ci.sh cmake.test
+      run: |
+        sudo ./ci/setup_thrift.sh
```
lalitb, line 23: We don't need `sudo`, as github actions are run by default as `root` user.
Also, if we move thrift installation as part of do_ci.sh script, this script will remain usable outside of github workflow, without separately installing thrift ( similar to how it is done for `prometheus`). Probably `setup_thrift.sh`  can expose a function to install thrift, and that function can be called from do_ci.sh for actual installation if called with "cmake.test" argument.

ThomsonTan: I also believe `sudo` is not necessary, just follow other scripts invocation like `setup_grpc.h`.

I don't prefer moving the installation logic to `do_ci.sh` which would make `do_ci.sh` too complicated with the third-party dependence setup instructions, and also hard to share with non-CI environment. Expose a setup function in `setup_thrift.sh` looks good, but we still need include `setup_thrift.sh` to invoke the function. So it may be simpler to just invoke the script?

ThomsonTan: @lalitb do all github action run as root by default? I just removed all the `sudo` in our `ci.yml` but got permission denied in the CI run.

lalitb: I thought unless explicitly specified with `--user` options, all jobs run as `root` user. Probably revert it back if there is a permission denied error. Sorry for the confusion.

lalitb (approved):

PR #788, opened 2021-05-23
Return Scope object instead of unique_ptr<Scope> from Tracer::WithActiveSpan()

### LINKED ISSUE #787: why nostd::unique_ptr<Scope>  not Scope
Any reason why WithActiveSpan returns `nostd::unique_ptr<Scope>` instead of simply returning Scope?

lalitb:
Fixes #787 

## Changes

- As mentioned in #787 , `Tracer::WithActiveSpan()` can avoid memory allocation on heap by returning newly created `Scope` object instead of `unique_ptr<Scope>`. The copy constructor of `Scope` is implicitly deleted because of `unique_ptr` member variable, so this ensures multiple copies are not possible for `Scope`.

- Also modified etw::tracer::StartSpan() method to not make span active immediately after creating as this deviates from the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation). `tracer::WithActiveSpan()` should be called explicitly by instrumentation library to make it active. ( cc @maxgolov )

### DIFF
--- a/exporters/etw/test/etw_tracer_test.cc
+++ b/exporters/etw/test/etw_tracer_test.cc
@@ -74,7 +74,7 @@ TEST(ETWTracer, TracerCheck)
     std::this_thread::sleep_for (std::chrono::seconds(1));
     {
       auto outerSpan = tracer->StartSpan("MySpanL2", attribs);
-      auto outerScope = tracer->WithActiveSpan(outerScope);
+      auto outerScope = tracer->WithActiveSpan(outerSpan);
 
       // Create nested span. Note how we share the attributes here.
       // It is Okay to either reuse/share or have your own attributes.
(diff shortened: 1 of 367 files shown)


lalitb:
@maxgolov - ETW exporter build is failing with this change due to the way we are using the `Scope` object in Tracer SDK there. Can you please review this PR whether it is a blocker for ETW?

pyohannes (approved):

exporters/etw/test/etw_tracer_test.cc
```diff
@@ -116,13 +116,13 @@ TEST(ETWTracer, TracerCheck)
         };
         EXPECT_NO_THROW(innerSpan->AddEvent(eventName3, event3));
         std::this_thread::sleep_for (std::chrono::seconds(1));
-        innerSpan->End();
```
ThomsonTan: Wondering why this is broken?

jsuereth (approved):
We may want to have a secondary method to apply active-span to context that DOESN'T have auto-scope (for advanced use cases).

Like this cleanup!

exporters/etw/test/etw_tracer_test.cc
lalitb: Looks like this comment is on an older / outdated commit?

ThomsonTan: Yes. I was commenting on the line which added `EXPECT_NO_THROW`.

ThomsonTan (approved):

exporters/etw/test/etw_tracer_test.cc
lalitb: Ok got it. That was to keep with consistent with other End() calls,

PR #792, opened 2021-05-24
Update readthedocs api guide: nested spans and context propagator

lalitb:
Added Nested spans and Context Propagation in API Getting started guide.

Here is how it looks like:

https://labhas-opentelemetry-cpp.readthedocs.io/en/latest/GettingStarted.html

### DIFF
--- a/docs/public/GettingStarted.rst
+++ b/docs/public/GettingStarted.rst
@@ -68,7 +68,7 @@ Create nested Spans
 
 
 Spans can be nested, and have a parent-child relationship with other spans.
-When a given span is active,the newly created span inherits the active span's
+When a given span is active, the newly created span inherits the active span's
 trace ID, and other context attributes.
 
 Context Propagation
--- a/docs/public/GettingStarted.rst
+++ b/docs/public/GettingStarted.rst
@@ -60,10 +60,10 @@ Create nested Spans
     {
         auto inner_span = tracer->StartSpan("Inner operation");
         auto inner_scope = tracer->WithActiveSpan(inner_span);
-        // inner operation
+        // ... perform inner operation
         inner_span->End();
     }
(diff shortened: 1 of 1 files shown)


docs/public/GettingStarted.rst
```diff
@@ -48,5 +48,57 @@ related span is ended.
 
 The concept of an active span is important, as any span that is created
 without explicitly specifying a parent is parented to the currently
-active span.
+active span. A span without a parent is called root span.
```
ThomsonTan, line 51:
```suggestion
active span. A span with a null parent is called root span.
```

docs/public/GettingStarted.rst
```diff
@@ -48,5 +48,57 @@ related span is ended.
...
+
+.. code:: cpp
+
+    auto outer_span = tracer->StartSpan("Outer operation");
+    auto outer_scope = tracer->WithActiveSpan(outer_span);
+    {
+        auto inner_span = tracer->StartSpan("Inner operation");
+        auto inner_scope = tracer->WithActiveSpan(inner_span);
+        // ... perform inner operation
+        inner_span->End();
```
ThomsonTan, line 64: Is this call to `->End()` optional? The destruction of `inner_span` would call `End()`.

lalitb: It's optional, as ~Span destructor calls it anyway. But the `Span` object returned by `Tracer::StartSpan` is shared_ptr and in case instrumentation code is referring to this object at some other places, the ~Span destructor will be called when all the references are cleaned up. It's always good ( and no harm ) to call it explicitly to export the Span. And when the later destructor is called, it won't perform this again.

ThomsonTan (approved):

PR #797, opened 2021-05-25
Explain the logical reasoning behind local Abseil snapshot + namespace renaming

maxgolov:
No code changes.

This PR is a supplementary doc that explains the logical reasoning / merits of the process used to create / refresh the Abseil snapshot in PR #572

Since some changes were not trivial, a separate section is added to elaborate on these changes.

Hopefully we should not employ this (semi-manual) process too often.


maxgolov:
We seriously have to do something about `codecov/project`. It is:
- rarely adds much (or any) value
- noisy
- creates an impression of CI failure when, in fact, CI is completely clean.

We may need to edit the associated GitHub action to avoid triggering red marks, and only treat it as optional, always pass showing the code coverage rate changes. Maybe there's an option to configure a min threshold that trips the failure, excluding anything that is under 1%.

Also, perhaps, if the list of changed files covers only Markdown or in `/docs` folder, we should avoid triggering any builds at all. That'd save environment and electrons that are unnecessarily disturbed.

lalitb:
incorrectly closed by bot.

PR #805, opened 2021-05-27
Fixes #804

### LINKED ISSUE #804: Header files in ostream exporter are not installed when building with cmake.
**Describe your environment** 
Header files in ostream exporter are not installed when building with cmake.

**Steps to reproduce**

Header files installed.

Nothing in `opentelemetry/exporters/ostream`

**Additional context**
Tag: v0.7.0

owent:

Fixes #804

## Changes

1. Fix install rule for header files of ostream exporter
2. Add more export targets for cmake

### DIFF
--- a/exporters/elasticsearch/CMakeLists.txt
+++ b/exporters/elasticsearch/CMakeLists.txt
@@ -1,10 +1,28 @@
-include_directories(include)
-include_directories(${CMAKE_SOURCE_DIR}/ext/include)
-
 add_library(opentelemetry_exporter_elasticsearch_logs src/es_log_exporter.cc)
+
+set_target_properties(opentelemetry_exporter_elasticsearch_logs
+                      PROPERTIES EXPORT_NAME elasticsearch_log_exporter)
+
+target_include_directories(
+  opentelemetry_exporter_elasticsearch_logs
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+         "$<INSTALL_INTERFACE:include>")
 
 target_link_libraries(opentelemetry_exporter_elasticsearch_logs
                       PUBLIC opentelemetry_trace http_client_curl)
+
+install(
+  TARGETS opentelemetry_exporter_elasticsearch_logs
+  EXPORT "${PROJECT_NAME}-target"
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+install(
+  DIRECTORY include/opentelemetry/exporters/elasticsearch
+  DESTINATION include/opentelemetry/exporters/
+  FILES_MATCHING
+  PATTERN "*.h")
 
 if(BUILD_TESTING)
   add_executable(es_log_exporter_test test/es_log_exporter_test.cc)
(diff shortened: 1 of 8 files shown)


lalitb (approved):
Thank you for the fix. Looks food.

exporters/elasticsearch/CMakeLists.txt
```diff
@@ -1,31 +1,36 @@
...
-
 add_library(opentelemetry_exporter_elasticsearch_logs src/es_log_exporter.cc)
 
+set_target_properties(opentelemetry_exporter_elasticsearch_logs
+                      PROPERTIES EXPORT_NAME elasticsearch_log_exporter)
+
+target_include_directories(
+  opentelemetry_exporter_elasticsearch_logs
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+         "$<INSTALL_INTERFACE:include>")
```
ThomsonTan, line 9: Should this line to `${CMAKE_SOURCE_DIR}/ext/include` above?

exporters/jaeger/CMakeLists.txt
```diff
@@ -26,6 +34,19 @@ if(MSVC)
   endif()
 endif()
 
+install(
```
ThomsonTan, line 37: I am wondering whether we need install the header files for SDK exporters. I think we don't want any user to include `opentelemetry-cpp` via `find_package()` to include them? @open-telemetry/cpp-approvers

exporters/elasticsearch/CMakeLists.txt
owent: `opentelemetry_exporter_elasticsearch_logs` has already link `http_client_curl` , which will add `${CMAKE_SOURCE_DIR}/ext/include` into include directories.

exporters/jaeger/CMakeLists.txt
owent: How should we import SDK exporters then? I have a project with prebuilt opentelemetry-cpp and use these codes below to use opentelemetry-cpp.

```cmake
find_package(opentelemetry-cpp)

add_executable(example_otlp main.cc)
target_link_libraries(
  example_otlp
  opentelemetry-cpp::trace
  opentelemetry-cpp::otlp_exporter
  opentelemetry-cpp::ostream_span_exporter
  opentelemetry-cpp::api)
```

ThomsonTan: The exported library target could be used for linking so it makes sense to install and export, but there is no need to install the header file as we don't expect the user to call the exporter functions directly?

lalitb: @ThomsonTan  - Won't user need to include these exporters headers too ( as here: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/simple/main.cc#L10 ). So they need to be installed?

owent: I found all examples use the headers of exporters(e.g. https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/jaeger/main.cc#L22).
Is there any example about using a exporter without calling the exporter functions directly?

ThomsonTan: You are right, the derived `SpanExporter` class definition should still be part of the header file. I was thinking about the `Recordable` stuff which is also installed by should not be directly visible to the user. I am fine to install the headers, perhaps we could move the non-public headers for exporter out of the installation path.

ThomsonTan (approved):

exporters/jaeger/CMakeLists.txt
owent: I like the idea of hiding the code details about how to initialize SDK. Maybe we can provider a API in SDK to use a configure file to `initTracer()` in the future, and then all compoments are no need to exported then?
I'm not sure if this should be done.It's seems not any part of specification and the similar job is already done by `opentelemetry-collector` .

lalitb: Looking at the otlp, elasticsearch,  and jaeger exporter code,  we actually don't need to install their recordable.h, and so it can be excluded. Users would only use exporter headers. Zipkin exporter header is incorrectly including its recordable header, which needs to be fixed before excluding.
If we plan to handle this in separate PR, we can raise a issue for that.

PR #806, opened 2021-05-27
Exclude header files for metrics from header-only cmake targets

lalitb:
## Changes

If metrics is disabled during cmake build ( i.e, WITH_METRICS_PREVIEW=OFF ), the two header-only targets - `opentelemetry-api` and `opentelemetry-sdk` still contain the metrics-related header files. This should be removed.

After this fix:
```console
$ cmake .. -DBUILD_TESTING=OFF -DWITH_METRICS_PREVIEW=OFF && make && make install
...
$ ls /usr/local/include/opentelemetry/
baggage  common  config.h  context  detail  exporters  ext  logs  nostd  plugin  sdk  std  trace  version.h
$ ls /usr/local/include/opentelemetry/sdk/
common  instrumentationlibrary  logs  resource  trace  version
$
$ cmake .. -DBUILD_TESTING=OFF -DWITH_METRICS_PREVIEW=ON && make && make install
...
$ ls /usr/local/include/opentelemetry/
baggage  common  config.h  context  detail  exporters  ext  logs  metrics  nostd  plugin  sdk  std  trace  version.h
$ ls /usr/local/include/opentelemetry/sdk/
common  instrumentationlibrary  logs  metrics  resource  trace  version
$ 
```


api/CMakeLists.txt
```diff
@@ -13,11 +13,17 @@ install(
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
+set(EXCLUDE_PATTERN "")
+if(NOT WITH_METRICS_PREVIEW)
+  set(EXCLUDE_PATTERN "metrics")
```
ThomsonTan, line 18: Just wondering this pattern could exclude some unexpected header files, like some potential internal metrics for SDK?

lalitb: Yes, if there are any header files with the name containing `metrics` under the non-metrics directory under `include/opentelemetry/sdk/`, it will exclude them, but I don't see why should someone create that.

ThomsonTan: Could we include the full or partial path? Agree that the name conflict chance is very low and not a blocker though.

ThomsonTan (approved):

api/CMakeLists.txt
lalitb: Yes, good point. I will add that.

lalitb: ok, I did some quick tests. The current pattern will only remove files/directories with the name "metrics", and not those containing "metrics". So we should be good here.

PR #809, opened 2021-05-28
Make our codecov settings more forgiving

maxgolov:
This is a humble attempt to make our codecov settings more forgiving.

Main motivation for this PR is that our current codecov settings are flaky and trigger A LOT of false-positives.

Practical illustration. See those red marks - false-positives mostly associated with badly tuned codecov settings:

image

Hopefully the new settings are going to:
- drop the codecov to informational.
- allow some float / slack in the settings.
- devs should still treat the generated report with respect and dignity, and use their best judgement to decide if they need to improve their newly added code code coverage.

### DIFF
--- a/.github/.codecov.yaml
+++ b/.github/.codecov.yaml
@@ -1,13 +1,22 @@
 codecov:
-  require_ci_to_pass: yes
+  require_ci_to_pass: false
   max_report_age: off
 
 coverage:
   precision: 2
-  round: down
+  round: up
   range: "80...100"
   status:
-    patch: off
+    project:
+      default:
+        informational: true
+        target: auto
+        threshold: 10%
+    patch:
+      default:
+        informational: true
+        target: auto
+        threshold: 10%
 
 parsers:
   gcov:
@@ -20,7 +29,7 @@ parsers:
 comment:
   layout: "reach,diff,flags,tree"
   behavior: default
-  require_changes: no
+  require_changes: false
 
 # Relative file path fixing.
 # CI file paths must match Git file paths.


maxgolov:
I refreshed my other PR with these settings to validate that it works as expected:
https://github.com/open-telemetry/opentelemetry-cpp/pull/771

.github/.codecov.yaml
```diff
@@ -1,13 +1,22 @@
...
-  round: down
+  round: up
   range: "80...100"
   status:
-    patch: off
+    project:
+      default:
+        informational: true
+        target: auto
+        threshold: 10%
```
ThomsonTan, line 14: Is 10% threshold too large than needed?

maxgolov: I think right now it affects only the color of the report, because either way it is now marked as optional. In my case that I quoted (e.g. Abseil) since I've been adding / touching size'able number of files, the diff was about 7%. I think so is the case for any mass clean-up, like adding headers, or any borrowed library. MAYBE 5% is a better number?

The other thing is default settings quoted by codecov are 80%. Assuming we strive to reach 100%, we may be able to regress down from 100% to -10%, 90% not less than that, with anticipation that a follow-up PR would fix that..

So 90% +/- **10%** should probably be fine. Again, it is informational only. It should only affect the color of the report now.

maxgolov: To explain the reason why I saw -7% on my other PR, I re-enabled the metrics, which immediately tanked the score:
https://app.codecov.io/gh/open-telemetry/opentelemetry-cpp/compare/771/tree/api/include/opentelemetry/metrics

Maybe we could indeed deal with something that is closer to 5% instead of 10%.
Please let me know your thoughts on that.

ThomsonTan (approved):

maxgolov:
We can strive to tune it down to maybe 2% _eventually_ . We can look at what sort of coverage we have per-component, and assuming we get to 80-100% zone for all of them, we can tune the threshold setting to 1-2%?

PR #810, opened 2021-05-30
Add OTLP/HTTP+JSON Protocol exporter

owent:

Fixes # 803

## Changes

Implementation of HTTP(OTLP/HTTP+JSON Protocol) exporter.

~This is a uncompleted PR and it's used to discuss the details of API and changes for now. I will add tests and CHANGELOG some time later.~

Almost finished.


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
```diff
@@ -0,0 +1,113 @@
...
+
+private:
+  // Stores if this exporter had its Shutdown() method called
+  bool is_shutdown_ = false;
+
+  // The configuration options associated with this exporter.
+  const OtlpHttpExporterOptions options_;
+
+  // Object that stores the HTTP sessions that have been created
+  std::unique_ptr<opentelemetry::ext::http::client::curl::HttpOperation> curl_operation_;
```
lalitb, line 108: Any specific reason for directly using curl implementation, and not the http_client as we use in zipkin: https://github.com/open-telemetry/opentelemetry-cpp/blob/f793e6e45e181e8224e38871b837746b878bf821/exporters/zipkin/src/zipkin_exporter.cc#L6

This will allow using platform-specific HTTP client implementation - e.g, WinSock on Windows.

lalitb:
> I add a OtlpHttpExporter , should we rename OtlpExporter into OtlpGrpcExporter and rename otlp_exporter.h|.cpp into otlp_exporter.h/.cpp ? It's a break change.

I agree we should rename OtlpExporter into OtlpGrpcExporter. We are not GA yet, so breaking changes are expected, and should be acceptable.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
owent: I want to create session/request from URL directly, without split it into host and port and support HTTPS. The API of `Session` and `HttpClient` have not such APIs by now  and I don't want to use `HttpClientSync`.
I think we can switch to `HttpClient` when it supports this feature.

lalitb: Really sorry if I am missing something here, as I still have some doubts:

> I don't want to use HttpClientSync

What is the benefit of an async request here? As `SpanExporter::Export()` doesn't support callback, so we need to anyway wait to get a response before returning from this method. Also, `libcurl` doesn't have async api.

> support HTTPS 

I think both `HttpClient` and `HttpClientSync` support HTTPS, as the authentication data can be added as part of HTTP headers. Also, with `HttpClientSync` we don't need to split host and port.

Unless really required, it's better not to force application owners to install the `libcurl` library. Applications running on Windows/macOS may prefer using native libraries for network communication.  Plan is to provide [HttpClientFactory](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/include/opentelemetry/ext/http/client/http_client_factory.h) implementation for these native libraries too similar to [curl](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/src/http/client/curl/http_client_factory_curl.cc). And let the HttpClient library pick up implementation as available for that platform.

owent: > What is the benefit of an async request here? As SpanExporter::Export() doesn't support callback, so we need to anyway wait to get a response before returning from this method. Also, libcurl doesn't have async api.

Just like `ElasticsearchLogExporter` , using async request allow us to cancle the request and exit gracefully here. libcurl has [multi interface](https://curl.se/libcurl/c/libcurl-multi.html) that can be used to perform asynchronous requests. I have some project use libuv and libcurl to send HTTP requests asynchronously.(e.g. [http_request](https://github.com/atframework/atframe_utils/blob/main/src/network/http_request.cpp) )

> support HTTPS

The codes in `opentelemetry::ext::http::client::curl::Session` is

```cpp
public:
  Session(HttpClient &http_client, const std::string &host, uint16_t port = 80)
      : http_client_(http_client), is_session_active_(false)
  {
    if (host.rfind("http://", 0) != 0 && host.rfind("https://", 0) != 0)
    {
      host_ = "http://" + host;  // TODO - https support
    }
    else
    {
      host_ = host;
    }
    host_ += ":" + std::to_string(port) + "/";
  }
```

It seems to reset `https://` to `http://`. 

I also think it's better to switch to `HttpClient` after HttpClient has API of create request only by URL. Or shall I use `HttpClientSync` for now and switch to `HttpClient` later?

lalitb: > Just like ElasticsearchLogExporter , using async request allow us to cancle the request and exit gracefully here

`ElasticsearchLogExporter` was implemented when sync request support was missing in the HttpClient library ( https://github.com/open-telemetry/opentelemetry-cpp/pull/444#discussion_r542112168 ), and there was a plan to migrate it on sync request. But I agree request cancellation is something we can use with the async request.


> I also think it's better to switch to HttpClient after HttpClient has API of create request only by URL. Or shall I use HttpClientSync for now and switch to HttpClient later?

I would suggest using `HttpClientSync` for now instead of `libcurl` directly. I agree `HttpClient` does need changes for supporting HTTPS and full-URL, and I will add a ticket for that. We can move to that once it's fixed.

owent: > I would suggest using HttpClientSync for now instead of libcurl directly. I agree HttpClient does need changes for supporting HTTPS and full-URL, and we can create a ticket for it. And then move to that once fixed.

Done. And rename `OtlpExporter`, `OtlpExporterOptions`, `otlp_exporter.h`, `otlp_exporter.cpp` to `OtlpGrpcExporter`, `OtlpGrpcExporterOptions`, `otlp_grpc_exporter.h`, `otlp_grpc_exporter.cpp` .

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
```diff
@@ -0,0 +1,107 @@
...
+public:
+  /**
+   * Create an OtlpHttpExporter using all default options.
+   */
+  OtlpHttpExporter();
+
+  /**
+   * Create an OtlpHttpExporter using the given options.
+   */
+  OtlpHttpExporter(const OtlpHttpExporterOptions &options);
```
lalitb, line 75: Nit - can we make this `explicit`. I think this is missing for `OtlpGrpcExporter` too which also should be modified.

lalitb (approved):
LGTM. Thanks for the contribution :)

bazel/repository.bzl
```diff
@@ -32,10 +32,10 @@ def opentelemetry_cpp_deps():
     maybe(
         http_archive,
         name = "com_github_grpc_grpc",
-        sha256 = "d6277f77e0bb922d3f6f56c0f93292bb4cfabfc3c92b31ee5ccea0e100303612",
-        strip_prefix = "grpc-1.28.0",
+        sha256 = "2060769f2d4b0d3535ba594b2ab614d7f68a492f786ab94b4318788d45e3278a",
+        strip_prefix = "grpc-1.33.2",
         urls = [
-            "https://github.com/grpc/grpc/archive/v1.28.0.tar.gz",
+            "https://github.com/grpc/grpc/archive/v1.33.2.tar.gz",
```
ThomsonTan, line 38: Why upgrade this to 1.33? I think 1.34 is used in our CI.

https://github.com/open-telemetry/opentelemetry-cpp/blob/0f6199f9f78f11f7192bf66f28e12e80d0556164/ci/setup_grpc.sh#L16

owent: I think 1.33 is the only version support both bazel 4+ and legacy gcc 4.8. 
` setup_grpc.sh` is only used when using cmake, not bazel.
I found grpc 1.34 upgrade abseil-cpp to `2020923.2` and it uses something gcc 4.8 do not support.
I submit another PR #666 and try to upgrade gRPC but failed on legacy toolchain. Maybe some one can help to switch the 
grpc to old version or just disable OTLP Exporters when using bazel?I'm not familar with bazel.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
owent: My negligence. I will add `explicit` later and add some tests.

examples/otlp/README.md
```diff
@@ -42,7 +43,8 @@ docker run --rm -it -p 4317:4317 -v "%cd%/examples/otlp":/cfg otel/opentelemetry
 
 Note that the OTLP exporter connects to the Collector at `localhost:4317` by
 default. This can be changed with first argument from command-line, for example:
-`./example_otlp gateway.docker.internal:4317`.
+`./example_otlp_grpc gateway.docker.internal:4317` and
+`./example_otlp_http gateway.docker.internal:4317`..
```
ThomsonTan, line 47: nit: 2 dots at the end of line 47.

ThomsonTan (approved):

bazel/repository.bzl
maxgolov: I assume for Windows, while eventually getting to prebuilt DLL / SDK, we can most likely use whatever recent version of gRPC from vcpkg? Mine right now is 1.37.0

```console
C:\work\opentelemetry-cpp\tools\vcpkg>vcpkg list grpc                                                                                                                                                                                                                           grpc:x64-windows                                   1.37.0#2         An RPC library and framework                     
```

I hope that one works with Visual Studio 2015 Update 2..
```

File: examples/CMakeLists.txt [hunk 5]
```diff
@@ -1,4 +1,4 @@
-if(WITH_OTLP)
+if(WITH_OTLP_GRPC OR WITH_OTLP_HTTP)
```
maxgolov, line 1: Do we really need that level of granularity, can one flag build both?

maxgolov (approved):

examples/CMakeLists.txt
ThomsonTan: I think it is rare for one build to have both, so 2 flags make sense. Or we could introduce a third flag `WITH_OTLP` to include both but then it seems over complicated.

owent:
Please do not merge this PR right now. I will add test codes later.

examples/CMakeLists.txt
owent: Sometimes we want only one of `WITH_OTLP_GRPC` and `WITH_OTLP_HTTP` .
Set `WITH_OTLP=ON` will try to turn `WITH_OTLP_GRPC` on when grpc is found and  turn `WITH_OTLP_HTTP` on when both libcurl and nlohmann_json are found in this PR.
The `WITH_OTLP` option is just for backward compatibility, and the actually options take effect is  `WITH_OTLP_GRPC` and `WITH_OTLP_HTTP` now.

bazel/repository.bzl
owent: I tested several toolchains with different versions of gRPC, and here is the result:

+ When using bazel to build opentelemetry-cpp
  + gRPC 1.33(which upgrade protobuf to 1.34) support bazel 4
  + gRPC 1.28-1.32 only support bazel 3
  + gRPC 1.28-1.33 support legacy gcc 4.8 while 1.34 or upper (which upgrade abseil to 20200923.X) don't
+ When using cmake to build opentelemetry-cpp
  + Just like bazel, and the official support of vcpkg is gcc 6+.
  + The protobuf (3.15) in the latest release of vcpkg(2021.05.12) do not support the latest MSVC 1929(VS 16.10)
    > Maybe we need wait for newer version to work with MSVC 1929(VS 16.10) or upper.

Right now, I just update gRPC in bazel to 1.33 for compatiability. Maybe it can be upgraded to newer version in the future and in another PR.

examples/otlp/README.md
owent: Thanks.

owent: Fixed.

exporters/otlp/test/otlp_http_exporter_test.cc
```diff
@@ -0,0 +1,323 @@
...
+  OtlpHttpExporterOptions opts;
+  opts.json_bytes_mapping = JsonBytesMappingKind::kHex;
+  std::unique_ptr<OtlpHttpExporter> exporter(new OtlpHttpExporter(opts));
+  EXPECT_EQ(GetOptions(exporter).json_bytes_mapping, JsonBytesMappingKind::kHex);
+}
+
+}  // namespace otlp
+}  // namespace exporter
+OPENTELEMETRY_END_NAMESPACE
+#endif
```
lalitb, line 323: Thank you for adding these real tests using the inbuilt HTTP server. We should be doing something similar for Zipkin too. I will add an issue for same.

lalitb (approved):
LGTM. Thanks for adding the tests.

PR #814, opened 2021-06-01
Reinstate the TracerContext constructor for single processor

maxgolov:
## Changes

Recent refactor removed the constructor for single processor, assuming that there is always a vector of processors #692 . However, the most common trivial scenario is when a customer provides a single processor rather than a vector of them. The intent of this change is to reinstate the previous constructor that got (inadvertently?) removed, keeping the simpler constructor in place where it was before.

No significant changes. Just reinstating what was accidentally removed.

Since the constructor is rather trivial (empty), implementation is included inline in the header, not added to separate .cc file. I do not have objection if there is an ask to move the empty body into separate compilation unit. But it seems okay to me as-is.

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/tracer_context.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer_context.h
@@ -33,6 +33,18 @@ class TracerContext
 {
 public:
   explicit TracerContext(
+      std::unique_ptr<SpanProcessor> processor,
+      opentelemetry::sdk::resource::Resource resource =
+          opentelemetry::sdk::resource::Resource::Create({}),
+      std::unique_ptr<Sampler> sampler = std::unique_ptr<AlwaysOnSampler>(new AlwaysOnSampler),
+      std::unique_ptr<IdGenerator> id_generator =
+          std::unique_ptr<IdGenerator>(new RandomIdGenerator())) noexcept
+      : processor_(std::move(processor)),
+        resource_(resource),
+        sampler_(std::move(sampler)),
+        id_generator_(std::move(id_generator)){};
+
+  explicit TracerContext(
       std::vector<std::unique_ptr<SpanProcessor>> &&processor,
       opentelemetry::sdk::resource::Resource resource =
           opentelemetry::sdk::resource::Resource::Create({}),


ThomsonTan (approved):

lalitb (approved):

lalitb:
> the most common trivial scenario is when a customer provides a single processor rather than a vector of them

We can recommend customers to use the `TracerProvider` interface as it supports both single and multiple processors unless they are directly using the `Tracer` object. 
The problem with the current change is that `TracerContext` maintains object of type `MultiSpanProcessor ` internally, and so calling `TracerContext::AddProcessor()` will fail after this change.

lalitb (changes requested):
This will fail `TracerContext::AddProcessor` as mentioned in comments.

maxgolov:
> The problem with the current change is that TracerContext maintains object of type MultiSpanProcessor internally, and so calling TracerContext::AddProcessor() will fail after this change.

@lalitb  - can you elaborate on a meaning of `AddProcessor` API in case if the system is configured to run with single span processor?

lalitb:
> @lalitb - can you elaborate on a meaning of `AddProcessor` API in case if the system is configured to run with a single span processor? How often it's called, and what's the need of that API? The only use-case I see is in a test case for the Multi Processor.

`AddProcessor()` should add the processor passed as an argument to the list of configured processors irrespective of where it is called. And below example will now fail for a multi-processor system ( as the customer may now want to add processors one by one ).

```cpp
  auto processor1 = std::unique_ptr<SpanProcessor>(new SimpleSpanProcessor(std::move(exporter1)));
  auto context = std::make_shared<TracerContext>(std::move(processor));
  auto processor2 = std::unique_ptr<SpanProcessor>(new SimpleSpanProcessor(std::move(exporter2)));
  context->AddProcessor(processor2);
```

maxgolov:
I don't think it fails if you invoke the constructor intended to be used for multi-processor scenario. Also assuming that if customer used the first constructor (single-processor), they are not going to attempt to call `AddProcessor`. We can possibly rework the single processor implementation to populate a vector with one single entry in it. Otherwise it appears on API surface of `TracerContext` that we operate everywhere on base class of `SpanProcessor`. Then somewhere in internal implementation detail we do a `static_cast` to child, which seems dangerous. It would have been nicer to have this done with `dynamic_cast`, and auto-rewire from single- to multi- processor only when necessary. Thus, not dealing with walking the list of processor in a trivial case.

lalitb:
>  Then somewhere in internal implementation detail we do a static_cast to child, which seems dangerous. It would have been nicer to have this done with dynamic_cast, and auto-rewire from single- to multi- processor only when necessary.

I don't see `static_cast` conversion here dangerous as long as we know that the object being converted is of child type. It's used internally and should be safe. As you rightly said, `dynamic_cast` uses RTTI and we shouldn't use it within the library. Not every application would want to enable RTTI.
I did try earlier to have approach like this, somehow didn't work, but would be worth trying out:

```
TracerContext::TracerContext(sstd::unique_ptr<SpanProcessor> &processor,
                             opentelemetry::sdk::resource::Resource resource,
                             std::unique_ptr<Sampler> sampler,
                             std::unique_ptr<IdGenerator> id_generator) noexcept
    : processor_(std::unique_ptr<SpanProcessor>(new MultiSpanProcessor(<somehow convert to vector of processor>))),
      resource_(resource),
      sampler_(std::move(sampler)),
      id_generator_(std::move(id_generator))
{}
``

lalitb:
incorrectly closed by bot.

lalitb:
Closing. The changes are not correct as discussed in the review comments.

PR #815, opened 2021-06-02
Allow to build `WITH_JAEGER` in Visual Studio 2019 + CMake

maxgolov:
## Changes

Can't build `WITH_JAEGER` on Windows via Visual Studio 2019, and/or via `tools/build.cmd`

Changes:
- install Thrift via vcpkg in `setup-buildtools.cmd`
- add Thrift detection logic in the main `CMakeLists.txt`
- respect the path that is auto-detected by vcpkg-installed Thrift : it's 1-level up from what thrift-gen is expecting, i.e. the include dir name contains `thrift/` but the code also includes `thrift/*.h`, so we need to append a dir that's 1-level up to make thrift-gen happy.

These changes allow to build `WITH_JAGER` on Windows.

UNRELATED ISSUE: Thrift is extremely bulky and requires Boost.. takes quite a bit of time to build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,20 @@ cmake_policy(SET CMP0057 NEW)
 
 project(opentelemetry-cpp)
 
+# Mark variables as used so cmake doesn't complain about them
+mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
+
+# Autodetect vcpkg toolchain
+if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
+  set(CMAKE_TOOLCHAIN_FILE
+      "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
+      CACHE STRING "")
+endif()
+
+if(VCPKG_CHAINLOAD_TOOLCHAIN_FILE)
+  include("${VCPKG_CHAINLOAD_TOOLCHAIN_FILE}")
+endif()
+
 file(READ "${CMAKE_CURRENT_LIST_DIR}/api/include/opentelemetry/version.h"
      OPENTELEMETRY_CPP_HEADER_VERSION_H)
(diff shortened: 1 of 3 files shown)


lalitb (approved):
LGTM. Thanks for the fix.

CMakeLists.txt
```diff
@@ -6,6 +6,20 @@ cmake_policy(SET CMP0057 NEW)
 
 project(opentelemetry-cpp)
 
+# Mark variables as used so cmake doesn't complain about them
+mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
```
ThomsonTan, line 10: What's the CMake complain here? User would not be able to see it in CMake GUI if it is marked as advanced.

exporters/jaeger/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 include_directories(thrift-gen)
 
 find_package(Thrift REQUIRED)
+# vcpkg config recipe points to THRIFT_INCLUDE_DIR=...\thrift Ensure that the
```
ThomsonTan, line 4: nit: missing period before `Ensure`?

ThomsonTan (approved):

exporters/jaeger/CMakeLists.txt
maxgolov: This was on two separate lines, but auto-formatter (`./ci/do_ci.sh format`) broke it.. I can add a period in between.

CMakeLists.txt
maxgolov: This is done to suppress "unused variable" on non-Windows OS. If you click on `[x] Advanced` in CMake GUI, then you can still show this variable. It'd be in the same class as a number of other important variables, like `CMAKE_CXX_COMPILER`, `CMAKE_C_COMPILER`, `CMAKE_C_FLAGS` ... I borrowed this idea from vcpkg.

ThomsonTan: Thanks, the mark of advanced makes sense then.

PR #816, opened 2021-06-02
Feature/span limits

### LINKED ISSUE #365: Adding span collection limits
As per the recent [spec ](https://github.com/open-telemetry/opentelemetry-specification/pull/942) change : 

Max attributes per span : 1000
Max links per span : 1000
Max events per span : 1000

The upper bound needs to be validated.

anshulkumar19:
Fixes #365 

## Changes

The changes made in this PR addresses the feature of imposing span limits parameters as defined in spec. 


sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -114,11 +114,21 @@ class AttributeMap
         });
   }
 
+  const std::size_t GetAttributeMapSize() const noexcept
```
maxgolov, line 117: I don't think we need this. Please check my PR here, as it'd be nicer if we simply expose all properties of real `std::map`-alike container here:

https://github.com/open-telemetry/opentelemetry-cpp/blob/752c5251734a90386d857928eeb73bea53926ff1/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L102

maxgolov (changes requested):
I think it would be much more elegant if that's just a usual `std::map` container. I am doing it in the other PR I mentioned in the comment. #771

anshulkumar19:
> I think it would be much more elegant if that's just a usual `std::map` container. I am doing it in the other PR I mentioned in the comment. #771

sdk/include/opentelemetry/sdk/trace/recordable.h
```diff
@@ -146,6 +147,25 @@ class Recordable
...
+
+protected:
+
+  SpanLimits GetSpanLimits()
+  {
+    return span_limit_;
+  }
+
+private:
+  SpanLimits span_limit_;
```
lalitb, line 168: I don't think we should add `SpanLimits` as a data member here. It will get allocated every time a new recordable is created ( i.e, every time a new span is created ). And SpanLimits values are not going to change.

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -197,7 +197,10 @@ class SpanData final : public Recordable
   void SetAttribute(nostd::string_view key,
                     const opentelemetry::common::AttributeValue &value) noexcept override
   {
-    attribute_map_.SetAttribute(key, value);
+    if (attribute_map_.GetAttributeMapSize() < GetSpanLimits().AttributeCountLimit
+          || attribute_map_.KeyExists(key)) {
+      attribute_map_.SetAttribute(key, value);
+    }
```
lalitb, line 203: Not sure if this is a good idea to drop the attributes silently.

sdk/include/opentelemetry/sdk/trace/span_limits.h
```diff
@@ -0,0 +1,35 @@
...
+	int AttributeCountLimit = 128;
+	int EventCountLimit = 128;
+	int LinkCountLimit = 128;
+	int AttributePerEventCountLimit = 128;
+	int AttributePerLinkCountLimit = 128;
+};
+
+} // namespace trace
+} // namespace sdk
+
```
lalitb, line 34: Can you this be a class with static methods to return the limits ?

sdk/include/opentelemetry/sdk/trace/recordable.h
anshulkumar19: This is not intended change, maybe didn't sync correctly.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
anshulkumar19: I will pick your changes from attribute_utils.h and modify accordingly.

sdk/include/opentelemetry/sdk/trace/span_data.h
anshulkumar19: Agreed. 
We can either log every time when attribute is dropped, but that would be excessive logging and is not recommended in the spec.
Second way, we can store the information of how many attributes are added in the span and so the number of attributes dropped. This should be enough information, any opinion on this?

sdk/include/opentelemetry/sdk/trace/span_limits.h
anshulkumar19: Since we are having a global tracer provider and span limits are associated with that only, so that should be better as it will simplify arguments passing. Will make the changes in next commit.

lalitb:
@anshulkumar19 -  While I am thinking about the approach to follow for supporting the span limits, the problem I see is that our implementation doesn't copies/owns any of the span properties inside api/sdk, and these properties (links, events, attributes) are propagated directly to exporters. `SpanData` is one sample implementation of Recordable which does owns data but is meant to be used by the in-memory exporter. This means SDK can't limit the number of these elements going to the exporter. The only way to impose the limit would be within the exporter code, and this is not what is expected from the specification.
Let me know your thoughts on this, and if you see any feasible approach for implementation. As this is an optional feature, we should go ahead with implementation only if it doesn't involve major changes in the way we propagate span data to exporters.

lalitb:
@anshulkumar19 - While I am thinking about the approach to follow for supporting the span limits, the problem I see is that our implementation doesn't copies/owns any of the span properties inside api/sdk, and these properties (links, events, attributes) are propagated directly to exporters. `SpanData` is one sample implementation of Recordable which does owns data but is meant to be used by the in-memory exporter. This means SDK can't limit the number of these elements going to the exporter. The only way to impose the limit would be within the exporter code, and this is not what is expected from the specification.
Let me know your thoughts on this, and if you see any feasible approach for implementation. As this is an optional feature, we should go ahead with implementation only if it doesn't involve major changes in the way we propagate span properties to exporters.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
maxgolov: Sounds good. That works. I'll reset my status to unblock your review.

sdk/include/opentelemetry/sdk/trace/span_data.h
maxgolov: Yeah, this won't really work... Because I'd see that class being used not necessarily only for Span Attributes. But for Resource Attributes as well. Do resource attributes have the same limit? I mean - can we move this check elsewhere.. Can we keep it a generic converter-helper class to transform attributes from API AttributeValue to OwnedAttributeValue, and just keep it as generic as possible, without binding it to Span specifics.

sdk/include/opentelemetry/sdk/trace/span_limits.h
```diff
@@ -0,0 +1,119 @@
...
+    return spanLimits;
+  }
+
+  static opentelemetry::common::SpinLockMutex &GetLock() noexcept
+  {
+    static opentelemetry::common::SpinLockMutex lock;
+    return lock;
+  }
+
+  int AttributeCountLimit         = 128;
```
maxgolov, line 110: Sorry for naïve question :

1. are we anticipating Span limits for different exporters (Jaeger, Zipkin, OTLP, or some other custom exporter, say FluentD-MsgPack) to be the same, and following these exact defaults?

2. If the answer to Q1 is `No`, then can we have these limits make somehow tunable, and be part of some form of Exporter configuration? This is becoming a bit tricky though. In case of multi-processor, in case if we want to route our telemetry to different exporter, and if every exporter has its own limits..

Apologies if it's a dumb question.

maxgolov:
Reading the spec - these are all **soft limits** - `the recommended limit of 1000 elements. SDKs MAY provide a way to change this limit...`, so it is **recommended** and it **MAY change**. Question here, so let's say my system imposes a different limit. And I fork my data via multi-processor to two different exporters at once. For example:
- console exporter.
- my (custom) MsgPack exporter.

It seems like then the issue is to configure the soft limits and intercept these _externally_ , beyond the multi-processor , once the event reaches concrete exporter, that may impose a soft limit.. In that case if the limit is exceeded, there has to be a mechanism to notify the caller about the failure.

That notification itself would most likely be asynchronous. Because if we are "forking" events to different destinations in a batching thread in the background, that is the only moment we'd know which concrete exporter gonna drop and which one gonna let it go.

I'd like to avoid imposing those soft limit in the API, and find a good way to solve this elsewhere. I think your limiter class, in general, looking Okay, but I'm not sure how to plug it best.

maxgolov:
Regarding my last point on async error callback - notification. I'd like to propose this being added as a generic facility into OpenTelemetry C++ SDK, following this pattern we have here:

https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/include/public/DebugEvents.hpp

https://github.com/microsoft/cpp_client_telemetry/blob/18fb1a1cf1a73077a32140d482719271153eff5a/lib/include/public/DebugEvents.hpp#L137

https://github.com/microsoft/cpp_client_telemetry/blob/18fb1a1cf1a73077a32140d482719271153eff5a/lib/include/public/DebugEvents.hpp#L173

It's a pub-sub model, when caller may receive an "interrupt" error notification - **DebugEvent**- sent from SDK worker thread(s) in case if something went bad. Caller listens for those **DebugEvent** notifications via registering **DebugEventListener** . That way the caller gets their `OnDebugEvent(x)` invoked, and depending on contents x - e.g. `SPAN_LIMIT_EXCEEDED` - may decide what to do. As long as they handle those errors wisely, i.e. can't call recursively call SDK from interrupt / error handler thread. 

They may ignore these notifications. Or they may transform these into additional layer of telemetry on SDK errors, if they wish to do so.

lalitb:
> Regarding my last point on async error callback - notification. I'd like to propose this being added as a generic facility into OpenTelemetry C++ SDK, following this pattern we have here:
> 
> https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/include/public/DebugEvents.hpp
> 
> https://github.com/microsoft/cpp_client_telemetry/blob/18fb1a1cf1a73077a32140d482719271153eff5a/lib/include/public/DebugEvents.hpp#L137
> 
> https://github.com/microsoft/cpp_client_telemetry/blob/18fb1a1cf1a73077a32140d482719271153eff5a/lib/include/public/DebugEvents.hpp#L173
> 
> It's a pub-sub model, when caller may receive an "interrupt" error notification - **DebugEvent**- sent from SDK worker thread(s) in case if something went bad. Caller listens for those **DebugEvent** notifications via registering **DebugEventListener** . That way the caller gets their `OnDebugEvent(x)` invoked, and depending on contents x - e.g. `SPAN_LIMIT_EXCEEDED` - may decide what to do. As long as they handle those errors wisely, i.e. can't call recursively call SDK from interrupt / error handler thread. The answer then is - we don't print anything, we don't log anything, we let the caller decide what they want to do with those errors. They may ignore these notifications. Or they may transform these into additional layer of telemetry on SDK errors, if they wish to do so. Using whatever choice, either logging these locally or (async) re-dispatching these errors back into local exporter, e.g. `stdout`.

@maxgolov  - Thanks for bringing the error handling scenario, as it has been discussed in past too. Request you to please add your recommendations here #408 . Specs also provide guidance on supporting custom error handlers [here](https://github.com/open-telemetry/opentelemetry-specification/blob/609a63784b2da7d59efe2a39822a23b09ae5b637/specification/error-handling.md#self-diagnostics).

sdk/include/opentelemetry/sdk/trace/span_data.h
lalitb: The limit is specific to Span attributes, links, and events. SDK doesn't own/copy any of these properties internally, so it is not possible to impose this limit within sdk. All these properties are propagated to exporters, and it is up to the exporters to impose the limits. The exporters may choose not to transform these properties to `OwnedAttributeValue` and instead transform/serialize to (say) JSON format for performance reasons.

maxgolov:
@lalitb - guidance is rather soft, it's not a requirement but rather a recommendation, example how: _These are examples of how end users might register custom error handlers. **Examples are for illustration purposes only. Language library authors are free to deviate from these provided that their design matches the requirements outlined above**._

I can send in a PR with a proposal for the **DebugEvent** emitter / **DebugEventListener** pattern implementation. It allows both - local subscriptions for the errors, say at Exporter (or maybe at TracerProvider level); or single global handler for global lifecycle / state management errors that could occur outside of specific TracerProvider or Exporter.

lalitb:
> I can send in a PR with a proposal for the **DebugEvent** emitter / **DebugEventListener** pattern implementation. It allows both - local async subscriptions for SDK errors. Say at Exporter (or maybe at TracerProvider level). Or a subscription to single global debug/error event handler for global lifecycle / state management errors that could occur outside of specific TracerProvider or Exporter.

Let's discuss the proposal on issue #408 before implementation. In general, we can start with specs guidelines and provide the functionality to let applications implement custom error handlers, and otel sdk providing functions to set and get these error handlers. This will also allow the applications to plugin the otel logging api as a custom error handler.

anshulkumar19:
> @anshulkumar19 - While I am thinking about the approach to follow for supporting the span limits, the problem I see is that our implementation doesn't copies/owns any of the span properties inside api/sdk, and these properties (links, events, attributes) are propagated directly to exporters. `SpanData` is one sample implementation of Recordable which does owns data but is meant to be used by the in-memory exporter. This means SDK can't limit the number of these elements going to the exporter. The only way to impose the limit would be within the exporter code, and this is not what is expected from the specification.
> Let me know your thoughts on this, and if you see any feasible approach for implementation. As this is an optional feature, we should go ahead with implementation only if it doesn't involve major changes in the way we propagate span properties to exporters.

I was about to add the similar constraints in exporter side "SetAttributes" functions so that these limits would be imposed there. Since the spec talks imposing these limits only to the sdk side, so that becomes tricky.
I can see the "sdk_exporters" are defined in the structure of spec (https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace). Is it suggesting that these sdk constraints can be implied on exporters as well, because eventually exporters are attached to the spans/traces via sdk?

sdk/include/opentelemetry/sdk/trace/span_limits.h
anshulkumar19: Spec mentions these limits set to tracer provider and we have one global tracer provider, so in the current scenario all of these limits should be same. 
Moreover, according to my understanding since these limits are there for the safety from erroneous code adding absurd number of attributes to a span so it should be related to span related limits rather than depending upon any exporters. So, once these limits are defined they should be followed for each kind of span (sdk/exporters).

lalitb:
> I was about to add the similar constraints in exporter side "SetAttributes" functions so that these limits would be imposed there. Since the spec talks imposing these limits only to the sdk side, so that becomes tricky.
> I can see the "sdk_exporters" are defined in the structure of spec (https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace). Is it suggesting that these sdk constraints can be implied on exporters as well, because eventually exporters are attached to the spans/traces via sdk?

 With the current otel-cpp design, the only place to filter these span properties would be at the exporter level. And if that is the case, a better approach would be to pass these limit configuration during exporter startup along with other exporter options. Eg in Zipkin, it would be here:

https://github.com/open-telemetry/opentelemetry-cpp/blob/3de8f354755b0b67d1c9e14abf889999e96bcd4f/exporters/zipkin/include/opentelemetry/exporters/zipkin/zipkin_exporter.h#L47-L55

Each exporter would have different memory requirements, and thus different configuration values. This is something not defined by specs, and let exporters decide how they want to implement it.

But for now, I don't think we should implement this optional requirement at the SDK level when there is nothing to filter there at SDK.  

@pyohannes, @maxgolov  - any thoughts?

sdk/include/opentelemetry/sdk/trace/span_data.h
maxgolov: @lalitb - I think we should have that trigger (limit enforcement) somewhere when we transform into `OwnedAttributeValue`. I don't have concrete proposals. Maybe we should merge the foundation class for some enforcement, proposed here in this PR. Then find out a good way to bind to that class.

maxgolov:
> With the current otel-cpp design, the only place to filter these span properties would be at the exporter level..

Yes, I agree with that.

anshulkumar19:
@lalitb @maxgolov As per suggestions for not implementing this optional feature at SDK level, closing the PR for now.

lalitb:
Thanks for your work @anshulkumar19 - as @maxgolov suggested this may be utilized in the future if we plan to have an upper bound for `OwnedAttributeValue` for a specific use-case. We can revisit this PR at that time.

PR #820, opened 2021-06-03
Fixes to compile SDK with Visual Studio 2019 Update 16.10 (C++20)

### LINKED ISSUE #819: `opentelemetry::proto::common::v1::_KeyValue_default_instance_`: illegal initialization of `constinit` entity with a non-constant expression (C++20)
Building with latest version of Visual Studio 2019 - update **16.10.0** , C++20 - using `-std:c++latest` , generated protobufs fail with this error:

image

Something related to this code here:

```cpp
// Our use of constinit does not yet work with GCC:
// https://github.com/protocolbuffers/protobuf/issues/8310
#if defined(__cpp_constinit) && !defined(__GNUC__)
#define PROTOBUF_CONSTINIT constinit
#elif defined(__has_cpp_attribute)
#if __has_cpp_attribute(clang::require_constant_initialization)
#define PROTOBUF_CONSTINIT [[clang::require_constant_initialization]]
#endif
#endif
#ifndef PROTOBUF_CONSTINIT
#define PROTOBUF_CONSTINIT
#endif
```

Since it used to work before, I wonder if it's related to **16.10.0** now defining `__cpp_constinit`. My speculation at this point that I'm using slightly older version of all dependencies (via vcpkg), and maybe missing this protobuf fix:

```console
2021-04-06 version 3.16.0 (C++/Java/Python/PHP/Objective-C/C#/Ruby/JavaScript)
...
  C++ 
  * Fix PROTOBUF_CONSTINIT macro redefinition (#8323)
```

And most likely the code still works well with Visual Studio 2019 update 16.9.x. Need to dig deeper to understand the best solution, most likely starting with pulling the latest protobuf. And checking if latest vs2019 compiler works well with latest protobuf. If not, possibly applying a patch to disable `PROTOBUF_CONSTINIT` how it's done for `__GNUC__`.

maxgolov:
Fixes #819 

## Changes

Quick recap:

- latest update to Visual Studio 2019 compiler (16.10) introduces support for `constinit`. However, this new feature breaks protobuf ( https://github.com/protocolbuffers/protobuf/issues/8688 ). Either because the feature is not fully implemented, or because protobuf uses it in some interesting way.

- I am verifying that with the fix all code now builds and work well with `WITH_STL` ( `-DHAVE_CPP_STDLIB`) in C++20. And that's where we now get to compile protobuf-generated stubs for OTLP with C++20. The issue was that it all worked literally last week with 16.9, got broken in 16.10. Which is very frustrating.. Because GitHub Action runner images have also all already been updated to pull the latest Visual Studio 2019 Update 16.10. There is no way back. Means if we don't fix it now, we can't verify C++20 builds with msvc-2019-16.10 in our CI.

## Fixes

Local patch here includes:
- a local overlay port / backport of `vcpkg` fix, that I submitted here in `vcpkg` : https://github.com/microsoft/vcpkg/pull/18251
- same fix proposed in the main `protobuf` : https://github.com/protocolbuffers/protobuf/pull/8690 

It may take some time before either of those fixes get merged.

There are no changes to API. No functionality changes expected other than fixing the build break. All files in `tools/ports/protobuf` are original files from mainline `vcpkg` port. With a minor fix (mainline PRs above) - applied on top, pretty much to disable the usage of `constinit`. Since it appears to be not ready to be used yet. At least not in Protobuf codebase.

`tools/setup-buildtools.cmd` now covers overlays for:
- Google Benchmark (was there before)
- Protobuf

Pinning both of them to versions that work well; with patches on top to make sure it works well even with latest vs2019.

I also updated `vcpkg` submodule snapshot to:

```text
commit 5568f110b509a9fd90711978a7cb76bae75bb092 (HEAD, tag: 2021.05.12)
Author: Victor Romero <romerosanchezv@gmail.com>
Date:   Thu May 13 12:53:39 2021 -0700
```

That's the latest `vcpkg` release published [here](https://github.com/microsoft/vcpkg/releases), so I figured the patched overlays must be compatible with that exact latest version of `vcpkg` since it is giving us the latest stable versions of everything.


lalitb:
@maxgolov  - I don't fully understand vcpkg yet. Just to be clear, these are the temporary changes and supposed to be removed once the mainstream patches are available for vcpkg, and/or fix is provided by protobuf?

maxgolov:
> @maxgolov - Just to be clear, these are the temporary changes and supposed to be removed once the mainstream patches are available for vcpkg, and/or fix is provided by protobuf?

Yes.

- same (identical) patch is now merged in protobuf master.
- same exact patch is still on review in vcpkg.
- it may take several weeks-to-months before protobuf posts the new release AND this release is picked up by vcpkg.

maxgolov:
Identical fix approved in `vcpkg` repo as well. But we'd need to upgrade the vcpkg submodule to master once this fix is merged.

lalitb (approved):
LGTM to have a temporary overlay. We can have a separate issue for cleanup once vcpkg brings the new protobuf version.

maxgolov:
We need to handle the removal of it in a few months from now (ETA: 3-5 months), when the next major `vcpkg` is released with the patch.

ThomsonTan:
If the same PR is already under review in vcpkg repo, could we just update our submodule once it is merged there? Instead of dup the fix here?

maxgolov:
> If the same PR is already under review in vcpkg repo, could we just update our submodule

We were mostly trying to anchor to their major releases, so once it's merged - I am not sure how soon we have to wait for their major release. That's why I was thinking to live with the overlay for another few months. vcpkg PR is this one:
https://github.com/microsoft/vcpkg/pull/18251

It's approved. All checks passed. But not merged yet.

ThomsonTan (approved):

maxgolov:
This test is flaky (unrelated to my change, have been failing before):

```console
	179 - trace.CircularBufferTest.Simulation (Failed)
```

PR #821, opened 2021-06-03
Update building-with-stdlib.md

maxgolov:
Provide / clarify some answers regarding `gsl::span`, why GSL and why not Abseil, and why it is needed if statically linking with compiler lower than C++17 for `WITH_STL` build flavor.  The document also now takes into consideration that MPark Variant is going away and Abseil Variant is becoming the default `nostd::variant` implementation for anything C++11 and above, if compiled without STL/stdlib classes.


lalitb (approved):
LGTM. Thanks for documenting :)

PR #832, opened 2021-06-07
Fix endianness of Jaeger IDs for transmission

### LINKED ISSUE #831: Jaeger should transmit all IDs in big endian
According to Jaeger exporter spec below, all IDs should be transmitted in big endian.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids

ThomsonTan:
Fixes #831

## Changes

Detect the host endianness and convert all IDs to big endian if needed.


exporters/jaeger/src/recordable.cc
```diff
@@ -39,12 +39,25 @@ void Recordable::PopulateAttribute(nostd::string_view key, const common::Attribu
 void Recordable::SetIdentity(const trace::SpanContext &span_context,
                              trace::SpanId parent_span_id) noexcept
 {
+  // IDs should be converted to big endian before transmission.
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids
+#if JAEGER_IS_LITTLE_ENDIAN == 1
```
maxgolov, line 44: I'm wondering if this is something that should be handled by `htons`?

https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-htons
https://linux.die.net/man/3/htons

maxgolov: Or, to be more precise, something like this?
```cpp
#define htonll(x) ((1==htonl(1)) ? (x) : ((uint64_t)htonl((x) & 0xFFFFFFFF) << 32) | htonl((x) >> 32))
#define ntohll(x) ((1==ntohl(1)) ? (x) : ((uint64_t)ntohl((x) & 0xFFFFFFFF) << 32) | ntohl((x) >> 32))
```

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
```diff
@@ -7,12 +7,46 @@
...
+#if (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && \
+     __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+#  define JAEGER_IS_LITTLE_ENDIAN 1
+#elif defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && \
+    __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#  define JAEGER_IS_LITTLE_ENDIAN 0
+#elif defined(_WIN32)
+#  define JAEGER_IS_LITTLE_ENDIAN 1
+#else
+#  error "Endian detection needs to be set up for your compiler"
```
lalitb, line 19: how about using run-time determination if it's not set by compiler, something like:
```
    const int value { 0x01 };
    bool isLittleEndian = ( *(static_cast<const unsigned char *>(static_cast<const void *>(&value))) == 0x01);
```

exporters/jaeger/src/recordable.cc
lalitb: nicely done. this will check the endianness to before performing conversion

ThomsonTan: Thanks. I considered `htonl` as the first option, but I'd like to avoid the runtime check on endianness and preferred the compiler intrinsic to perform byte order swap for simplicity and slight perf win.

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
ThomsonTan: I'd like to avoid the runtime check as it is a static property known at compile time. I think our current definition should cover most real cases if not all.

exporters/jaeger/src/recordable.cc
ThomsonTan: I could change the function name to `to_big_endian` and define it as as direct return for `big endian` host.  In this way, we don't need check the endianness at callsite of this function.

ThomsonTan: Second thought, leave the `JAEGER_IS_LITTLE_ENDIAN` in the callsite could save us to create `bswap_128`.

lalitb (approved):
LGTM.

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
bogdandrutu: For a reference implementation see https://github.com/abseil/abseil-cpp/blob/master/absl/base/internal/endian.h I think you can use that code (not directly if avoiding dependency is needed) but is a good implementation.

exporters/jaeger/src/recordable.cc
maxgolov: I was wondering if it's something you can move into a separate static inline utility function, then it can be reused elsewhere? I think it might be common requirement, to transmit the buffer in big endian, not unique to Jaeger?

ThomsonTan: So far I only saw the requirement in Jaeger. I'd be happen to move this to common utility if it is used in other places.

maxgolov (approved):
I'm Okay with the essence of this change. But I think it would have been nicer to refactor the implementation into separate utility header. This refactor won't anyhow degrade the perf, if you make the function `static inline`. By the looks of it what you are doing is `htonllll` , which should really be _standard_ utility function. I believe we'd need this in other exporters.

Issue #812, opened 2021-06-01
HTTPClient : Support for full URL argument instead of host & port


lalitb:
As discussed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/810#discussion_r642303340):

- [HttpClient::CreateSession](https://github.com/open-telemetry/opentelemetry-cpp/blob/ac0e6174e74f845e7d375a156b5abea33956c056/ext/include/opentelemetry/ext/http/client/http_client.h#L229) method should support passing full URL as an argument instead of host and port (or perhaps support both). This would be then consistent with `HttpClientSync::Get(url)` and `HttpClientSync::Post(url)` methods.
- Validate if HTTPS URL is parsed successfully, and supported for curl implementation. Remove `TODO` from here once `https` works/fixed:

https://github.com/open-telemetry/opentelemetry-cpp/blob/ac0e6174e74f845e7d375a156b5abea33956c056/ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h#L123

ohamuy:
Hey! Could I pick up this issue?

ohamuy:
@lalitb Would you rather have me allow for [HttpClient::CreateSession](https://github.com/open-telemetry/opentelemetry-cpp/blob/ac0e6174e74f845e7d375a156b5abea33956c056/ext/include/opentelemetry/ext/http/client/http_client.h#L229) to have the template have multiple parameters, one for host/port and one for a full url? Or would you rather have me change the current implementation to simply fix https and work for full url

ohamuy:
I’d like to work on this issue. Can you please assign this to me? @alolita

lalitb:
Thanks @ohamuy  - assigned to you now. We can use function overloading to support both unless you see benefit with templates?

ohamuy:
Yea @lalitb I could definitely do function overloading. Also how do we know whether the user wants http or https appended if they only provide the host and port? Also for testing would I use Bazel or is there some other way to test my code?

lalitb:
> Yea @lalitb I could definitely do function overloading. Also how do we know whether the user wants http or https appended if they only provide the host and port? Also for testing would I use Bazel or is there some other way to test my code?

default should be always `http` unless use explicitly specified `https`. This should be irrespective of whether port number is 80 or 443.

ohamuy:
[...]

PR #833, opened 2021-06-07
Http client add url support

ohamuy:
HTTPClient : Support for full URL argument

Fixes # [812](https://github.com/open-telemetry/opentelemetry-cpp/issues/812)

Overloaded the CreateSession function allowing for url input:
Added it in the header file http_client.h
Implementation in http_client_curl.h
The function allows for url input instead of just host & port, utilizing the url parser in common folder. It returns an empty/non-existent host, and default port(80) if the parsing fails, and continues similarly to the original function otherwise.
One Unit test was changed to test the overloading function, but the overall functionality is not changed as the other original function parameters are tested in other unit tests. All tests passed, file builds successfully.
Lastly, the default port of 80 was removed from the original function because if the user ends up not providing a port, it will go through the other overloaded function and the url parser gives it a default port of 80

@lalitb @reyang


ThomsonTan:
@ohamuy thanks for your contribution. Could you please fix the format?

ohamuy:
@ThomsonTan Done! Also should I add to the change log for this commit?

ext/include/opentelemetry/ext/http/client/http_client.h
```diff
@@ -227,8 +227,11 @@ class HttpClient
 {
 public:
   virtual std::shared_ptr<Session> CreateSession(nostd::string_view host,
-                                                 uint16_t port = 80) noexcept = 0;
-  virtual bool CancelAllSessions() noexcept                                   = 0;
+                                                 uint16_t port) noexcept = 0;
```
maxgolov, line 230: Do we still need the port parameter here, if the URL itself can contain scheme and port?
Should we rename the `host` parameter to `url` or `uri` ?

ext/include/opentelemetry/ext/http/client/http_client.h
```diff
@@ -227,8 +227,11 @@ class HttpClient
...
+
+  virtual std::shared_ptr<Session> CreateSession(nostd::string_view url) noexcept = 0;
```
maxgolov, line 232: Okay, I see what you are doing. But I am not sure - maybe we should just deprecate the first method?

ohamuy: Yes exactly, I just overloaded the function because that is what the @lalitb (the original issue creator) said might have been good. If its alright I can just simply remove the old parameter and just leave the new one since its also consistent with the `GET` and `POST` commands. But that is definitely a decision for the maintainers so when they give their input we can see whats the best course of action. @lalitb  @reyang

lalitb: Yes, it should be fine removing the original existing method as @maxgolov suggested.

lalitb (approved):
LGTM once the proposed changes are done. Thanks for the contribution.

ohamuy:
@lalitb @maxgolov So I implemented the changes. Final change list:

-Deleted old `CreateSession` implementation
-Changed HTTPClient tests to match the correct parameters of the new function (all tests passed)
-Original constructor for `Session` was greatly simplified to one line as now we can pass in the scheme of `parsedUrl` 
-Added scheme parameter for `Session` constructor

ohamuy:
Ok so apparently removing the original function broke a lot of tests, what should I do?

.vscode/settings.json
```diff
@@ -0,0 +1,89 @@
+{
```
maxgolov: Do we need this?

maxgolov (approved):
LGTM, but I think your Visual Studio code settings is probably something that we'd rather not add in here?

maxgolov:
> Ok so apparently removing the original function broke a lot of tests, what should I do?

Doesn't look like _a lot_ - 3 files with tests... Fix them? 😃 

image

.vscode/settings.json
ohamuy: I removed this in the most recent commit! it was an accidental file commited

ThomsonTan (approved):

ThomsonTan (approved):

lalitb:
Thank you for your contribution. Looks good to merge now :)

Issue #828, opened 2021-06-07
Can the API avoid returning shared_ptr<Span> 


bogdandrutu:
See: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/tracer.h#L39

By having this design an atomic ref-counting will always be needed. This will block in the future a possibly more performant "streaming" implementation where Span == SpanContext and all "record" apis will stream an event. The reason for "blocking" is because will still pay the refcounting and heap allocation cost even though is not needed.

The SDK can implement the Span by having an internal shared_ptr using a pimpl pattern.

bogdandrutu:
Same comment should apply for "Baggage" as well. It should allow an alternative implementation (even though is not a proper implementation but could be a header replace) that does not require Baggage to be heap allocated and refcounted.

jsuereth:
We (Google) would be a fan of this change, giving us freedom to control performance going forward with differing implementations

pyohannes:
This should be doable.

I'm not sure if we can preserve the ability to automatically end a span when it is deallocated (or goes out of scope). We might end up requiring explicit calls to `End` for all spans that are not managed by a [Scope](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/scope.h).

bogdandrutu:
You can achieve exactly the same behavior by having internally a `shared_ptr<SpanImpl>` and have the copy ctor/assignment do ref and destructor unref, and when out of scope you call End. It is just a variant that does not change the behavior but changes the ownership of the reference object, and allows alternative implementations that can avoid using reference counting.

lalitb:
Had a quick look into span api code, the problem I see here is that `api::Span` is an abstract class, which means we can't return a copy of it,  so we have to return a reference or pointer of it ( holding the instance of `sdk::Span`), and this will introduce the typical ownership/memory issues where application/instrumentation-code is still holding the reference/pointer to object which has been already deleted. 

For the rest of the components like Baggage / TraceState, we can use the Pimpl idiom to get rid of returning shared_ptr as implemented here: #836

lalitb:
[...]

PR #836, opened 2021-06-08
[POC] Baggage API : return `Baggage` instead of `shared_ptr<Baggage>` using  PImpl idiom

lalitb:
Not to merge for now - this is proof of concept for Baggage Implementation using Pimpl Idiom, as part of discussion #828 


jsuereth:
Nice work!

One question:

Did we have performance benchmarks for this like we do for span processors?  If not we should open a bug going forward to check performance of baggage propagation.

lalitb:
> Did we have performance benchmarks for this like we do for span processors? If not we should open a bug going forward to check performance of baggage propagation.

Good point. Will raise a bug for this.

api/include/opentelemetry/baggage/baggage.h
```diff
@@ -18,276 +19,61 @@ namespace baggage
...
-      else
-      {
-        err = 1;
-        return "";
-      }
-    }
-
-    return ret;
-  }
+  std::string ToHeader() const { return baggage_impl_->ToHeader(); }
```
maxgolov, line 72: Since this is in API, is this `std::string` type ever going to be exposed across DLL / Shared Library boundary? Even if it's a private method, isn't this still something where we'd rather return a `nostd::string_view` instead of `std::string` ?

api/include/opentelemetry/baggage/baggage_impl.h
```diff
@@ -0,0 +1,292 @@
...
+        ret.push_back(to_hex(c >> 4));
+        ret.push_back(to_hex(c & 15));
+      }
+    }
+
+    return ret;
+  }
+
+  // Uri decode key value pairs after extracting from header
+  static std::string UrlDecode(nostd::string_view str, bool &err)
```
maxgolov, line 242: Need to check what happens when `static std::string` is delcared as a static method on base class of a class that you are exporting. I don't think this inheritance is safe. I believe it has to be PIMPL without inheritance!

api/include/opentelemetry/baggage/baggage.h
```diff
@@ -18,276 +19,61 @@ namespace baggage
 class Baggage
 {
 public:
-  static constexpr size_t kMaxKeyValuePairs = 180;
-  static constexpr size_t kMaxKeyValueSize  = 4096;
-  static constexpr size_t kMaxSize          = 8192;
-  static constexpr char kKeyValueSeparator  = '=';
-  static constexpr char kMembersSeparator   = ',';
-  static constexpr char kMetadataSeparator  = ';';
+  Baggage() : baggage_impl_(new BaggageImpl()) {}
```
maxgolov, line 22: I don't think you can inherit from IMPL, it has to be private PIMPL.

api/include/opentelemetry/baggage/baggage.h
```diff
@@ -18,276 +19,61 @@ namespace baggage
...
-
-    return ret;
-  }
+  std::string ToHeader() const { return baggage_impl_->ToHeader(); }
 
 private:
-  // Store entries in a C-style array to avoid using std::array or std::vector.
-  nostd::unique_ptr<opentelemetry::common::KeyValueProperties> kv_properties_;
+  Baggage(BaggageImpl *baggage_impl) : baggage_impl_(baggage_impl) {}
+  nostd::shared_ptr<BaggageImpl> baggage_impl_;
```
pyohannes, line 76: Maybe I'm missing something: the `shared_ptr` is wrapped now and not exposed, but isn't it the goal to provide a way to get completely rid of `shared_ptr` usage? How would that work with this solution?

api/include/opentelemetry/baggage/baggage.h
lalitb: It's a non-virtual function, so it should be safe to return `std::string`, it won't break ABI stability. Though we can change it as below had it been a virtual function:

```
  void ToHeader(nostd::string &header) {
}
```

api/include/opentelemetry/baggage/baggage_impl.h
lalitb: Sorry, I didn't get this. Neither of  `Baggage` and `BaggageImpl` is base class or derived class here.

api/include/opentelemetry/baggage/baggage_impl.h
```diff
@@ -0,0 +1,292 @@
...
+{
+
+static constexpr size_t kMaxKeyValuePairs = 180;
+static constexpr size_t kMaxKeyValueSize  = 4096;
+static constexpr size_t kMaxSize          = 8192;
+static constexpr char kKeyValueSeparator  = '=';
+static constexpr char kMembersSeparator   = ',';
+static constexpr char kMetadataSeparator  = ';';
+
+class BaggageImpl
```
bogdandrutu, line 25: Can this be hidden? Maybe have an "internal" as part of the namespace and we add a rule that everything in the internal package is private? "baggage::internal::BaggageImpl"?

bogdandrutu (commented):
Nice work, thanks.

api/include/opentelemetry/baggage/baggage.h
lalitb: You are right @pyohannes. I was additionally trying to make the copy constructor & assignment operator for `Baggage` private, to ensure that we increment ref count only for `BaggageImpl` internally when needed using `friend` class, but seems we can't have non-copyable alternatives in `std::variant`.

maxgolov: So, in essence, we are not really making it better-perf for any of the default exporters.. we making it worse by creating yet another abstraction layer with a shared_ptr. Okay. 🤔

maxgolov: I'd like to see a Google Benchmark before and after.

lalitb: Yes true. In fact, it adds another layer of redirection, along with a small cost incurred for storing <shared_ptr> as a data member. This should be a solution only if we plan not to use smart pointers across API. It's more of a design decision than a performance solution.

maxgolov: Let's run some numbers. I am also interested in synchronous, non-batching, streaming exporter. And in good numbers.

pyohannes: > So, in essence, we are not really making it better-perf for any of the default exporters.. we making it worse by creating yet another abstraction layer with a shared_ptr. Okay.

I kind of agree, I don't see how this improves things. We still have a baked-in shared pointer which we hide from the user, and the user has no way to change that.

I don't see any benefits of this refactor. If it's a purely stylistic effort, I have no strong opinion about that.

> Do you see any better approach to handle it?

I see two possible ways:
1. Implementing a purely stack-based solution ourselves. This would require a re-implementation of the underlying key/value data structure.
2. Giving the user means so they can provide their own baggage implementation, which follows their own memory management strategy.

Maybe there are other ways not coming to my mind. If we go with (2), the same approach could also be used to avoid using shared pointers for spans.

maxgolov:
I think perf numbers presently show that this refactor isn't making anything better, plus takes away the partial ownership semantics. 

I think it'd be best if we carefully allow additional API, maybe in a separate namespace, to address the need without refactoring of what we already got working in `v1.0`.

lalitb:
These are the perf numbers for baggage api ( the earlier one posted at #828 was for Span create api ) with and without PIMP idiom:

**Existing Implementation ( without PIMPL):**
```
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
Load Average: 0.38, 0.44, 0.35
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
BM_CreateBaggageFromTenEntries         14671 ns        14671 ns        44127
BM_ExtractBaggageHavingTenEntries        361 ns          361 ns      1853059
BM_CreateBaggageFrom180Entries        271756 ns       271757 ns         2541
BM_ExtractBaggageWith180Entries         5815 ns         5815 ns       115602
BM_SetValueBaggageWithTenEntries        2344 ns         2344 ns       301103
BM_SetValueBaggageWith180Entries       42179 ns        42179 ns        16987
BM_BaggageToHeaderTenEntries            3992 ns         3992 ns       178561
BM_BaggageToHeader180Entries           69452 ns        69451 ns         9263
```

**New Implementation ( this PR) using PIMPL** 
```
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
Load Average: 0.39, 0.44, 0.35
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
BM_CreateBaggageFromTenEntries         13780 ns        13780 ns        45221
BM_ExtractBaggageHavingTenEntries        378 ns          378 ns      1886558
BM_CreateBaggageFrom180Entries        263451 ns       263452 ns         2631
BM_ExtractBaggageWith180Entries         5634 ns         5634 ns       113505
BM_SetValueBaggageWithTenEntries        2394 ns         2394 ns       298951
BM_SetValueBaggageWith180Entries       40432 ns        40432 ns        16882
BM_BaggageToHeaderTenEntries            3988 ns         3988 ns       173536
BM_BaggageToHeader180Entries           70706 ns        70706 ns         8979
```

There is no significant performance difference in both the approaches, which is expected as we still use `shared_ptr<Baggage>` internally as a data member in the PIMPL approach.

PR #839, opened 2021-06-08
Build farm for Windows

maxgolov:
This PR implements a `CMake` + `ninja` build farm setup for the most common compilers used on modern versions of Windows.

Both - developer -focused and CI/CD pipeline -focused, to cover the following compilers:
- Visual Studio 2015 (v140) - `C++11/C++14` - targeting Windows 8.1 SDK and above.
- Visual Studio 2017 (v141) - `C++17` - targeting Windows 10 and above.
- Visual Studio 2019 (v142) - `C++20` - targeting Windows 10 and above.
- LLVM Clang - targeting whatever version of Windows SDK. Tested with 11 and 12.

Using the following build systems:
- `CMake`
- `Ninja` - fastest, with optional fallback to MSBuild.
- `vcpkg` - used for dependency management.

The process is largely documented [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/building-with-vs2019.md) . Except that now we also support not only 2019. But older compilers too, plus LLVM Clang which we never had in our build system for Windows. Ad-hoc tested that the code works well with Clang-11 and Clang-12. This comes in handy, because things like Chromium on Windows - build with clang.

Build scripts contain enough comments for further customization and tweaking as-necessary. Someone implementing their own custom build pipeline may use these scripts as a reference example. Devs may also use these scripts to quickly build and test the SDK side-by-side with a set of different compilers installed on one box. Output from the build is laid out as follows:

image

Running something like:

```console
tools\build-vs2015.cmd
tools\build-vs2017.cmd
tools\build-vs2019.cmd
tools\build-clang-12.cmd
```

Produces a set of build artifacts in one run:
- `nostd` classes - for ABI compat across different compilers.
- `WITH_STL` - for static linking, with native STL classes.
for all compilers above.

Note that `vcpkg` installation for Visual Studio 2015 must always be stored separately from the rest. Because if you use `vs2019` to compile and link `vcpkg` libraries -- these packages won't link to SDK being built with `vs2015` compiler and linker. That limitation described in Visual Studio compiler documentation [here](https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-160#restrictions) :

image


api/include/opentelemetry/std/span.h
```diff
@@ -39,9 +39,10 @@
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace nostd
 {
+using gsl::dynamic_extent;
```
maxgolov, line 42: That's a fix from #846 - I needed it to verify Visual Studio 2017 build.

maxgolov:
@lalitb @ThomsonTan - this PR is only dev build scripts necessary for all versions of Visual Studio. If you can have a quick look and merge, that'd be great!

lalitb (approved):

ThomsonTan (approved):

PR #840, opened 2021-06-08
More fixes to build all with Visual Studio 2015

### LINKED ISSUE #756: Remove `std::copy` from ETW exporter
This is related to controversial topic of Annex K bound checkers added as an optional extension in C11. This thread [here](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1967.htm) provides some background detail, with reasonably written community feedback about the usefulness of Annex K.

ETW exporter uses `std::copy` here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/32f10c7eea5b64b0fb75e14517ddf4cdd7f6b0ac/exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h#L305

This triggers a false-positive with some strict code analysis tooling when [Secure Development Lifecycle](https://www.microsoft.com/en-us/securityengineering/sdl) checks are turned on. I already recommended the customer to temporarily suppress the warning just around this compilation unit. Since it is false-positive in this context.

My goal is to fix this for ETW exporter. Unfortunately fixing it one thing at a time won't necessarily bring us anyhow closer to totally making code `std::copy`-free, since the same pattern is actively employed by `nlohmann/json.hpp` and some parts of Google Abseil library.

Perhaps part of the fix should be to establish a CI/CD pipeline for Windows, to ensure that the entire codebase (core parts of API and SDK) remain compliant with secure coding practices on Windows. It is progressively harder for many customers to keep suppressing these (perhaps unnecessarily strict) warnings in their own builds.

maxgolov:
These are all the changes needed to compile OpenTelemetry C++ SDK with Visual Studio 2015.
I manually tested ad-hoc that the entire SDK works well with all optional component included.
Fixing warnings, errors, and warnings treated as errors.

Fixes #756 

Partially addresses #215 , manually tested ad-hoc. No CI yet. CI loop (scripts, no code changes) will be added in #839

## Changes

- `CMAKE_VERSION` check was not properly working on very OLD cmake lower than 3.3.

- `b3_propagator.h` triggers an error about narrowing cast of `kTraceIdHexStrLength`. Add explicit static cast to `size_t`.

- `nostd::variant_alternative_t` test does not compile in Visual Studio 2015. I suspect we are missing one of the headers. The rest of it works. And it's a lower priority - since we do not use any comparison of variant type in SDK, there should be no decrease in code coverage by disabling that test.

- ETW exporter: vs2015/C++ has no `__has_include` feature. And since `TraceLoggingDynamic.h` is now always available - Microsoft open-sourced it under friendly license, we unconditionally include it. It's the best and standard compact binary protocol we should use on Windows for ETW, that can be easily decoded out-of-proc by .NET listener, for example.

- ETW exporter: `using` behavior in C++11 triggers ambiguity when the same class name is being aliased as `using` vs something else that is imported via `using namespace`. That doesn't happen on more recent versions of compilers. Explicitly assume that the ETW-specific wrapper classes (Properties,PropertyValue,PropertyValueMap) get imported via `using namespace` instead of alias, then it all works great.

- ETW exporter: avoid using `std::copy` since older compilers were picky about this due to Secure Lifecycle Development process. New compilers are more relaxed. But, unfortunately, we strive to run with zero compliance warnings on old compiler.

- `__cplusplus` flag is not supported by old Visual Studio 2015. Add version check to avoid 'unknown flag` warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if(WITH_STL)
   add_definitions(-DHAVE_CPP_STDLIB)
   add_definitions(-DHAVE_GSL)
   # Require at least C++17. C++20 is needed to avoid gsl::span
-  if(CMAKE_MINOR_VERSION VERSION_GREATER "3.18")
+  if(CMAKE_VERSION VERSION_GREATER 3.18.0)
     # Ask for 20, may get anything below
     set(CMAKE_CXX_STANDARD 20)
   else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,8 +92,7 @@ if(WITH_STL)
         set(MSVC_CXX_OPT_FLAG "/O2")
       endif()
     endif()
-    set(CMAKE_CXX_FLAGS
-        "${CMAKE_CXX_FLAGS} /Zc:__cplusplus ${MSVC_CXX_OPT_FLAG}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${MSVC_CXX_OPT_FLAG}")
   endif()
 endif()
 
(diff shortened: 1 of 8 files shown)


maxgolov:
Appears like this test is occasionally tripping:

```console
//exporters/otlp:otlp_grpc_exporter_test                                 FAILED in 0.3s
  /home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/io_opentelemetry_cpp/bazel-out/k8-tsan-fastbuild/testlogs/exporters/otlp/otlp_grpc_exporter_test/test.log
  ```
  
Doesn't appear to be related to my change. Rerunning it.

lalitb (approved):
LGTM

ThomsonTan (approved):

lalitb:
> Appears like this test is occasionally tripping:
> 
> ```
> //exporters/otlp:otlp_grpc_exporter_test                                 FAILED in 0.3s
>   /home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/io_opentelemetry_cpp/bazel-out/k8-tsan-fastbuild/testlogs/exporters/otlp/otlp_grpc_exporter_test/test.log
> ```
> 
> Doesn't appear to be related to my change. Rerunning it.

Yes, it's failing frequently now. Need to check if it's coming from any latest merge.

PR #855, opened 2021-06-14
Add absl/stl CI builds using cmake

### LINKED ISSUE #854: Add CI job to enable build using STL/ABSEIL for modern C++ features.
Enable CI build to use ABSEIL and STL for modern C++ features ( std::string_view, std::variant, std::span, std::shared_ptr, std::unique_ptr) .

lalitb:
Fixes #854 

## Changes

Add CI builds for WITH_STL and WITH_ABSEIL cmake options.
WITH_STL build is failing, so it is commented-out.


.github/workflows/ci.yml
```diff
@@ -56,6 +72,8 @@ jobs:
         sudo ./ci/setup_cmake.sh
     - name: run tests
       run: ./ci/do_ci.sh cmake.c++20.test
+    #- name: run tests (enable stl)
```
ThomsonTan, line 75: Log an issue to add this back?

ci/install_abseil.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+export DEBIAN_FRONTEND=noninteractive
+
+if ! type cmake > /dev/null; then
+    #cmake not installed, exiting
```
ThomsonTan, line 10: Log an error message to the console?

ci/install_abseil.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+fi
+export BUILD_DIR=/tmp/
```
ThomsonTan, line 13: Is `export` necessary here? Does shell variable work?

ci/install_abseil.sh
```diff
@@ -0,0 +1,26 @@
...
+export INSTALL_DIR=/usr/local/
+pushd $BUILD_DIR
+git clone --depth=1 -b 20210324.0 https://github.com/abseil/abseil-cpp.git
```
ThomsonTan, line 16: Is `20210324.0` a tag name or branch name? And define it in a shell variable?

lalitb: It's a tag name, Made it a shell variable.

ci/install_abseil.sh
lalitb: No, it's not required. Made it a shell variable.

ci/install_abseil.sh
lalitb: good point. Done that

.github/workflows/ci.yml
lalitb: Yes, created now- #859

maxgolov: And I already fixed it. Please merge the fix - #860

ci/install_abseil.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+export DEBIAN_FRONTEND=noninteractive
+
+if ! type cmake > /dev/null; then
+    echo "cmake not installed. Exiting.." >&2
```
maxgolov, line 10: Technically we do have at least two scripts that allow to install cmake if necessary:
- `/ci/setup_cmake.sh` - that's only Linux x86_64. But installs fast.
- `/tools/setup-cmake.sh` - that's everything. But it builds it from source. And it's building slowly.

I'd suggest you don't even check for cmake in here. Assume it's given.

ci/install_abseil.sh
```diff
@@ -0,0 +1,27 @@
...
+INSTALL_DIR=/usr/local/
+TAG=20210324.0
+pushd $BUILD_DIR
+git clone --depth=1 -b ${TAG} https://github.com/abseil/abseil-cpp.git
+cd abseil-cpp
+mkdir build && pushd build
+cmake -DBUILD_TESTING=OFF -DCMAKE_CXX_STANDARD=11 \
+    -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR \
+    ..
+make -j $(nproc)
```
maxgolov, line 23: If you have ninja, see `/tools/setup-ninja.sh`, that'll give you at least x3-x5 times faster build.

ci/do_ci.sh
```diff
@@ -55,6 +68,20 @@ elif [[ "$1" == "cmake.c++20.test" ]]; then
...
   exit 0
+elif [[ "$1" == "cmake.c++20.stl.test" ]]; then
+  install_prometheus_cpp_client
+  cd "${BUILD_DIR}"
+  rm -rf *
+  cmake -DCMAKE_BUILD_TYPE=Debug  \
+        -DWITH_METRICS_PREVIEW=ON \
+        -DWITH_LOGS_PREVIEW=ON \
+        -DCMAKE_CXX_FLAGS="-Werror" \
+        -DCMAKE_CXX_STANDARD=20 \
```
maxgolov, line 79: You don't have to manually specify `CMAKE_CXX_STANDARD=20`. I fixed it so that it gets properly asked-for and detected in #860  . You can just omit it because `WITH_STL=ON` is going to set it either to 17 or 20, depending on CMake version. Older CMake versions did not know about 20.

maxgolov (approved):
It looks good, but if you merge #860 first - you should be able to uncomment the loop with gcc-9 and c++20 that you had to comment out. All tests pass with my fix ( at least on my machine 😄 ).

ci/install_abseil.sh
maxgolov: One interesting thing here is that gRPC / protobuf, for example, should already be installing Abseil ❓ Maybe it's installed private in gRPC / Protobuf.

ci/install_abseil.sh
maxgolov: I think one good option, or at least optimization opportunity, would be to see how gRPC is installing it. Because then we can just anchor to the same exact Abseil used by gRPC/protobuf/OTLP, to minimize code / workspace size... That'll be though non-ABI-portable min-size static build of SDK.

ci/install_abseil.sh
maxgolov: I'd say - don't even check for CMake in this script. We have so many other scripts, and we can just say that CMake is a major pre-requisite to get anything built, esp. if you are setting up a build loop for CMake - it is already making assumption that CMake has to be there.

ci/install_abseil.sh
lalitb: Yes, the plan is to reuse Abseil from gRPC install once we add these tests for otlp expoter. As of now, it is not testing otlp exporter.

ci/install_abseil.sh
lalitb: Sure will remove it.

ci/install_abseil.sh
lalitb: Yes, I was thinking to add this as part of the existing `cmake_otprotocol_test` job. As part of this PR, we are not building otlp exporter using Abseil.

ci/do_ci.sh
lalitb: nice, will do tha.

ThomsonTan (approved):

PR #853, opened 2021-06-14
Optimize OtlpHttpExporter to support timeout and cancle

owent:
Signed-off-by: owent <admin@owent.net>

Fixes #852

## Changes

Use `HttpClient` to replace `HttpClientSync` .

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [x] Changes in public API reviewed

PR #858, opened 2021-06-15
Fix missing notify of conditional variable in OTLP HTTP exporter test

### LINKED ISSUE #857: Conditional variable cv_got_events in otlp_http_exporter_test is never notified
**Describe your environment**
otlp_http_exporter_test use [conditional variable](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/test/otlp_http_exporter_test.cc#L45) to sync between main thread and HTTP response handling thread, but the conditional variable is never notified even when the HTTP response is completed processing.

**Steps to reproduce**
Run `exporter.otlp.OtlpHttpExporterTestPeer.ExportJsonIntegrationTest`.

The test should pass shortly.

The test passed, but it also causes timeout.

**Additional context**

ThomsonTan:
Fixes #857 

## Changes

Added `notify_one` call to `cv_got_events` right after HTTP response processing.

### DIFF
--- a/exporters/otlp/test/otlp_http_exporter_test.cc
+++ b/exporters/otlp/test/otlp_http_exporter_test.cc
@@ -226,7 +226,7 @@ TEST_F(OtlpHttpExporterTestPeer, ExportJsonIntegrationTest)
     report_trace_id.assign(trace_id_hex, sizeof(trace_id_hex));
   }
 
-  ASSERT_TRUE(waitForRequests(20, old_count + 1));
+  ASSERT_TRUE(waitForRequests(2, old_count + 1));
   auto check_json                   = received_requests_json_.back();
   auto resource_span                = *check_json["resource_spans"].begin();
   auto instrumentation_library_span = *resource_span["instrumentation_library_spans"].begin();
--- a/exporters/otlp/test/otlp_http_exporter_test.cc
+++ b/exporters/otlp/test/otlp_http_exporter_test.cc
@@ -87,6 +87,8 @@ public:
       }
     }
 
+    int response_status = 0;
+
     if (request.uri == kDefaultTracePath)
     {
       response.headers["Content-Type"] = "application/json";
@@ -103,7 +105,7 @@ public:
         else
         {
           response.body = "{\"code\": 400, \"message\": \"Parse binary failed\"}";
-          return 400;
+          response_status = 400;
         }
       }
       else if (nullptr != request_content_type && *request_content_type == kHttpJsonContentType)
@@ -113,7 +115,7 @@ public:
         if (json.is_discarded())
         {
           response.body = "{\"code\": 400, \"message\": \"Parse json failed\"}";
-          return 400;
+          response_status = 400;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


exporters/otlp/test/otlp_http_exporter_test.cc
```diff
@@ -124,18 +126,22 @@ class OtlpHttpExporterTestPeer : public ::testing::Test, public HTTP_SERVER_NS::
...
     else
     {
       std::unique_lock<std::mutex> lk(mtx_requests);
       response.headers["Content-Type"] = "text/plain";
       response.body                    = "404 Not Found";
-      return 200;
+      response_status = 200;
     }
+
+    cv_got_events.notify_one();
```
lalitb, line 142: why is this needed ? `cv.wait_for` was supposed to be unblocked with a timeout. I think this will fail now if the test is modified so that `cv.wait_for` has multiple requests to wait for, as now it will be unblocked after a single request?

ThomsonTan: Timeout in `cv.wait_for` is the last resort. `cv` should be waked up anytime its checking condition could be changed, so its condition could be re-checked instead of waiting for timing out. `notify_one` just tell `cv` to recheck so it should not trigger failure behavior?

cc @owent

lalitb: Does it mean that without notification, the `cv` will keep blocked for 2sec (as configured) even though the condition is satisfied early?

ThomsonTan: Yes, this is the case the PR trying to fix.

lalitb: Asking because we are using this construct in other places too (https://github.com/open-telemetry/opentelemetry-cpp/blob/7fca2c39586696d71de4f1feeb9ee6705a4e0261/ext/test/http/curl_http_test.cc#L124), probably they need a change too ( in separate PR ).

lalitb (approved):

exporters/otlp/test/otlp_http_exporter_test.cc
ThomsonTan: You are right, it is needed in curl_http_test.cc as well. Added the change to the PR and avoid another 2 timeout to happen there. Thanks for the catch.

owent: > Timeout in `cv.wait_for` is the last resort. `cv` should be waked up anytime its checking condition could be changed, so its condition could be re-checked instead of waiting for timing out. `notify_one` just tell `cv` to recheck so it should not trigger failure behavior?
> 
> cc @owent

You are right. LGTM.

owent:
Could you please remove the `TODO` I omit in #853 ? I think there is no need to create a PR just to remove the `TODO` comment.

ThomsonTan:
> Could you please remove the `TODO` I omit in #853 ? I think there is no need to create a PR just to remove the `TODO` comment.

Sure, just removed that.

PR #865, opened 2021-06-17
If external nlohmann::json is not found, then use one from our submodule

### LINKED ISSUE #843: Windows compile with CMake fails - Can't find googletest or nlohmann_json
**Describe your environment** 
Windows VS 2019 (16.10.0)

**Steps to reproduce**
Clone repository with --recursive.
Run CMake with arguments -G "Visual Studio 16 2019" -A "x64" "-DBUILD_TESTING=OFF" "c:\dev\AA3rdPartyExt\opentelemetry-cpp" "-DCMAKE_TOOLCHAIN_FILE=C:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DWITH_PROMETHEUS=ON

Clean compile

Build-CMake-Project-Windows : Error Invoking cmake to create the project files failed - level 1
 Building with nostd types...
CMake Error at C:/dev/vcpkg/scripts/buildsystems/vcpkg.cmake:263 (_find_package):
    nlohmann_jsonConfig.cmake
    nlohmann_json-config.cmake
  has been installed.

**Additional context**
I was able to work around this by cloning and building both googletest and nlohmann_json and setting the CMake CMAKE_PREFIX_PATH environment variable to the target path of the respective builds.

DCMAKE_PREFIX_PATH=c:\dev\AA3rdPartyExt\Build\nlohmann_json\VS16\Win64;c:\dev\AA3rdPartyExt\Build\googletest\VS16\Win64"

**Note:** It appears that the submodules were downloaded as per the log output below:
Processing project Open Telemetry from repository https://github.com/open-telemetry/opentelemetry-cpp in local folder c:\dev\AA3rdPartyExt\opentelemetry-cpp.
Cloning files from GitHub repository https://github.com/open-telemetry/opentelemetry-cpp into c:\dev\AA3rdPartyExt\opentelemetry-cpp
Submodule path 'third_party/benchmark': checked out 'c05843a9f622db08ad59804c190f98879b76beba'
Submodule path 'third_party/googletest': checked out 'a6dfd3aca7f2f91f95fc7ab650c95a48420d513d'
Submodule path 'third_party/ms-gsl': checked out '6f4529395c5b7c2d661812257cd6780c67e54afa'
Submodule path 'third_party/nlohmann-json': checked out 'db78ac1d7716f56fc9f1b030b715f872f93964e4'
Submodule path 'third_party/opentelemetry-proto': checked out '795cc815ce9bb438b46a4a7f6ecb465b52d50935'
Submodule path 'third_party/prometheus-cpp': checked out 'bb017ec15a824d3301845a1274b4b46a01d6d871'
Submodule path 'third_party/prometheus-cpp/3rdparty/civetweb': checked out '8e243456965c9be5212cb96519da69cd54550e3d'
Submodule path 'third_party/prometheus-cpp/3rdparty/googletest': checked out '703bd9caab50b139428cea1aaff9974ebee5742e'
Submodule path 'tools/vcpkg': checked out 'e9f8cc67a5e5541973e53ac03f88adb45cc1b21b'

maxgolov:
Fixes #843 ( partially 😄 )

Quick recap: despite the fact that we clone `nlohmann::json` library as one of our submodules, we do not attempt to use that by default. Neither in Linux nor in Windows. This is a bit illogical, because for Windows - we expect it to be provided by `vcpkg`, on Mac we install it with `brew`, and on Ubuntu we expect it to be installed via `apt`... Despite the fact that we actually have it right there - in the submodule!

## Changes

The following components depend on / require `nlohmann/json.hpp` :
- ETW exporter (optional for MsgPack binary payload encoding)
- OTLP HTTP exporter
- Elastic logs exporter
- Zipkin exporter
- w3c Trace Context tests
- zPages

Given that:
- in majority of scenarios the library is required.
- and the library is (header-only) always there - one of our submodules.

I'd like to consolidate the usage of it as follows:
- stop requiring it in all these modules above
- and instead require it unconditionally once in top-level `CMakeLists.txt`
- try to find nlohmann_json package (e.g. installed via brew or vcpkg) first
- if the package is not found, then use our local submodule snapshot (that's always there and will always be found)

The library itself is "pay-per-play". It's header only. If someone builds a build combo without it, - fine, we're good, we're just not including it. We require it. But we don't include it. This approach makes the overall SDK build / consumption experience much smoother.. And we no longer need to explain _why we have JSON right there in the submodule but it's not found during the build_. Making this world quite a bit more sane and reasonable place.

Note that we presently don't install / export / deploy the json.hpp library together with OpenTelemetry SDK headers. The library remains private in most cases. If we ever need to export that as part of OpenTelemetry SDK, i.e. install it to user include directories, then we could set the `JSON_Install` option to ON. This installs our local snapshot of nlohmann/json.hpp.. But in cases where users really feel like their code depends on json.hpp, they might as well install their own version themselves via pkg manager.

Similar approach could be implemented for `benchmark`, `googletest` and `ms-gsl`... But for for now I'm tackling this only for `nlohmann/json.hpp`.


CMakeLists.txt
```diff
@@ -228,6 +228,24 @@ endif()
...
+if(NOT nlohmann_json_FOUND)
+  message("Using local nlohmann::json from submodule")
+  set(JSON_BuildTests
+      OFF
+      CACHE INTERNAL "")
+  set(JSON_Install
+      OFF
+      CACHE INTERNAL "")
+  # This option allows to link nlohmann_json::nlohmann_json target
+  add_subdirectory(${PROJECT_SOURCE_DIR}/third_party/nlohmann-json)
```
lalitb, line 241: Will it fail `cmake` configure if submodules are not checked out? Just thinking of the scenarios if the user wants to build api and sdk which ideally doesn't have json/curl requirements, we shouldn't let configure fail in those cases.

maxgolov: Since it is required for _most_ build configurations, I can add a check if directory exists. If it does not exist, I can trigger a warning recommending to do recursive clone.

lalitb: Yes agreed. How about not breaking the build if `WITH_API` is enabled and `json` is not installed? Though I don't have strong opinion on this :)

lalitb (approved):

ThomsonTan (approved):

CMakeLists.txt
maxgolov: @lalitb - Sounds good. I can improve it in one of my other build infra PRs!

PR #868, opened 2021-06-21
Adding semantic-convention attributes for trace

### LINKED ISSUE #566: Add Semantic Conventions
From spec:

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions

Example existing solutions:
[Java](https://github.com/open-telemetry/opentelemetry-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java)

lalitb:
Fixes #566 

## Changes

This PR adds semantics as defined here: 
https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions

This will let us use uniform attributes/events/exceptions names across services.

This is currently used in the HTTP and gRPC example. These may be used in exporters too, would create PR to replace there.


api/include/opentelemetry/trace/semantic_conventions.h
```diff
@@ -0,0 +1,190 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/version.h"
+
+#define OTEL_TRACE_GENERATE_SEMCONV_METHOD(attribute_name, attribute_value) \
+  static const nostd::string_view Get##attribute_name() noexcept            \
```
maxgolov, line 10: Ideally this should be expressible as `constexpr` - to return the value of it at compile time, without incurring the overhead of a static method call. I'm not sure if the getters give significant benefit, as in - better structure or something? From Google C++ style document: `Global strings: if you require a global or static string constant, consider using a simple character array, or a char pointer to the first element of a string literal. String literals have static storage duration already and are usually sufficient.`

examples/http/server.cc
```diff
@@ -31,17 +35,16 @@ class RequestHandler : public HTTP_SERVER_NS::HttpRequestCallback
...
-                {{"http.server_name", server_name},
-                 {"net.host.port", server_port},
-                 {"http.method", request.method},
-                 {"http.scheme", "http"},
-                 {"http.request_content_length", static_cast<uint64_t>(request.content.length())},
-                 {"http.client_ip", request.client}},
-                options);
+    auto span = get_tracer("http-server")
+                    ->StartSpan(span_name,
+                                {{SemanticConventions::GetAttributeHttpServerName(), server_name},
```
maxgolov, line 40: It looks rather bulky, in my opinion.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: How can we do it in an all header way? C++11 won't allow initializing`std::string`/`nostd::string` within class definition.

examples/http/server.cc
lalitb: Agree. Do we have a better way of handling this?

api/include/opentelemetry/trace/semantic_conventions.h
```diff
@@ -0,0 +1,190 @@
...
+ * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md"/>.
+ */
+class SemanticConventions final
+{
+public:
+  // The set of constants matches the specification as of this commit.
+  // https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md
+
+  // Genetal network connection attributes
```
ThomsonTan, line 32:
```suggestion
  // General network connection attributes
```

api/include/opentelemetry/trace/semantic_conventions.h
```diff
@@ -0,0 +1,190 @@
...
+  OTEL_TRACE_GENERATE_SEMCONV_METHOD(AttributeNetTransport, "net.transport")
```
ThomsonTan, line 33: Expand `Attribute` in name `AttributeNetTransport` in the macro of `OTEL_TRACE_GENERATE_SEMCONV_METHOD` instead of duplicating it for every macro call?

api/include/opentelemetry/trace/semantic_conventions.h
```diff
@@ -0,0 +1,190 @@
...
+  OTEL_TRACE_GENERATE_SEMCONV_METHOD(AttributeNetPeerIp, "net.peer.ip")
```
ThomsonTan, line 34: Could we generate these definitions from the YAML definition of semantic conventions such as https://github.com/open-telemetry/opentelemetry-specification/blob/main/semantic_conventions/trace/http.yaml?

api/include/opentelemetry/trace/semantic_conventions.h
maxgolov: You just don't use getters. Use constants with `constexpr` definition?

api/include/opentelemetry/trace/semantic_conventions.h
maxgolov: ...and could we generate them, and make them `constexpr` string constants? These would be encouraged by Google C++ coding style, and provide the most optimal compile-time optimization. `Global strings: if you require a global or static string constant, consider using a simple character array, or a char pointer to the first element of a string literal. String literals have static storage duration already and are usually sufficient.` .. constexpr string literal or string array should work, with all of these constants having unique prefix. Rather than grouping them in class - we group them in namespace. There is also a compile time flag that allows string pooling, we use that to avoid bloating in case if the same header gets included in different compilation units. For Visual Studio compiler that is `/GF (Eliminate Duplicate Strings)`. And for gcc / clang I think it's on by default?

api/include/opentelemetry/trace/semantic_conventions.h
maxgolov: Or if you use getters, return straight the literal constant type without any class:

image

examples/http/server.cc
maxgolov: There are several options:
- make methods `constexpr` and return a string literal constant, inited as const char * or char array
- or just define all string constants in a dedicated namespace, that enables `using` for those who want to write it in a shorter form. Then for gcc / clang this scenario is already optimized, and for msvc we need to enable string pooling to combine all strings into one region of memory. These constants all have global static duration, there is no need to "get" them, they'd be efficiently compile-time optimized. Using `string_view` for that is an overkill.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: > Could we generate these definitions from the YAML definition of semantic conventions such as

I did think of writing a `script` to parse the yaml file and generate C++ code containing these `consts` from that and add it as part of CI. But this data is not going to change on daily basis, and instead, it would be fine to modify our code a couple of times before GA to be consistent. 
For otel-python and otel-go, they do have a script to generate code for these constants, along with other semantic data ( eg creating enums for allowable values for attributes, defining constraints of the list of attributes to be used etc) which we don't need to do. The current approach in this PR is as done by `otel-dotnet`.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: Good point. Will do  that.

examples/http/server.cc
lalitb: Thanks for the suggestion. I did initially wanted to use the second option but wasn't aware of the compiler optimization of this static global across multiple translation units. Let me try that.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: Haha, thanks for catching this, it doesn't looks good :)

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: > Or if you use getters, return straight the literal constant type without any class:

Thanks for your suggestions. I have modified the code to create the `constrxpr` function and return literal.  For another suggestion on using the `constrxpr` variable, somehow, I don't see these getting optimized in different translation units by gcc compiler ( small pingy code to demonstrate that: https://www.onlinegdb.com/IjRRjLjZo ). But anyway preferred the `constexpr` function as we don't need to depend on linker/compiler optimization here.

examples/http/server.cc
lalitb: As mentioned in prev comment, have created the `constexpr` function now.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: This is done now.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: To summarise the available options:
 **option1**:  `constexpr static class methods`: The advantage of using these is that only those methods are generated (in .text segment) at compile time which are used across all the translation units. The rest of the unused methods are ignored.
**option2**: `constexpr static global functions in the namespace`: This will create a copy of the used functions in each translation unit while ignoring the rest of the unused functions. 
**option3**: `constexpt static variables in the namespace`: This will create a local copy of all the variables in each of the translation units including the header, and it seems that the gcc compiler doesn't optimize it.

The right solution comes from the C++17 `inline` variables, but `option1` seems to be the correct approach for C++11 which is now implemented in this PR. We can discuss if there are any issues with the approach.

api/include/opentelemetry/trace/semantic_conventions.h
lalitb: @ThomsonTan  - Have created issue for automatic generation of these definitions : https://github.com/open-telemetry/opentelemetry-cpp/issues/873
Once we finalize the approach to use, first ( and probably initial few ) definitions can be committed manually while someone works on script to generate code as per the finalised approach.

lalitb:
@maxgolov @ThomsonTan I have modified the current implementation in this PR as per the logic used for sem-conv for Resources in #872. Please review it again. Thanks.,

lalitb:
@open-telemetry/cpp-approvers. @maxgolov  - Need help reviewing it. Changes are similar to what we implemented for already merged PR for resources: #872. Thanks.

ThomsonTan (approved):

Issue #869, opened 2021-06-21
Can't build project (either API_ONLY or default options) using CMake from release tarball


lidavidm:
**Describe your environment** 

Ubuntu 18.04 x86_64, OpenTelemetry v1.0.0-rc2

```
$ uname -a
Linux partita 5.4.0-74-generic #83~18.04.1-Ubuntu SMP Tue May 11 16:01:00 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
$ cmake --version
cmake version 3.19.6
$ g++ --version
g++ (crosstool-NG 1.24.0.133_b0863d8_dirty) 9.3.0
```

**Steps to reproduce**

```
$ wget https://github.com/open-telemetry/opentelemetry-cpp/archive/refs/tags/v1.0.0-rc2.tar.gz
$ tar xf v1.0.0-rc2.tar.gz
$ cd opentelemetry-cpp-1.0.0-rc2/
$ mkdir build && cd build
$ cmake .. -DWITH_API_ONLY=ON # or omit
```

**What is the expected behavior?**

We should be able to configure OpenTelemetry.

**What is the actual behavior?**

CMake fails because nlohmann-json isn't included in the release tarball:

```
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /home/lidavidm/miniconda3/envs/arrow4/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/lidavidm/miniconda3/envs/arrow4/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building for architecture ARCH=x64
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
Using local nlohmann::json from submodule
CMake Error at CMakeLists.txt:241 (add_subdirectory):
  The source directory

[...]

PR #870, opened 2021-06-21
If `nlohmann_json` is missing in customer's build system, then install it from submodule

### LINKED ISSUE #843: Windows compile with CMake fails - Can't find googletest or nlohmann_json

maxgolov:
Fixes #843

## Problem Description

Previous fix ( #865 ) works well for Linux and Mac. But missed one important detail for Windows: `nlohmann::json` package is unconditionally required by ETW exporter on Windows, because `json.hpp` header is included in **public** ETW exporter headers. And because it is added as INTERFACE package. When it is added as INTERFACE (not PRIVATE), means it must to be installed on customer machine into CMake cache. Otherwise CMake won't let the build to proceed. Actually the same is true for zPages, but for zPages we just did not enforce it via `target_link_libraries ... INTERFACE nlohmann_json` - so it was "passing" without a hard-check.

There are at least two workarounds available:
- use `vcpkg install nlohmann_json` to install the package before the buiild ; OR
- pass `-DJSON_Install=ON` to CMake when compiling for Windows. That'll install the `nlohmann_json` before building the rest of code.

With the fix, the workaround of passing `-DJSON_Install=ON` is no longer needed.

## Changes

If `nlohmann::json` is not installed on customer system, then we use our instance of it from our submodule. **And we install it**. Because it is unconditionally required by ETW exporter, zPages, both part of default build. I also improved the logic to check that submodule directory exists. We discussed it in a previous related PR. If submodule does not exist and the package is not installed from elsewhere, then the build fails with a recommendation / instructions how to fix it by recursively cloning with submodules :

image

Customer reporting the issue confirmed that the fix / idea behind the fix is working well:

image

UPDATE: also when we package the source tarball, allow the API-only build to avoid including `nlohmann_jsonhpp` library.


CMakeLists.txt
```diff
@@ -235,13 +235,21 @@ if(NOT nlohmann_json_FOUND)
       OFF
       CACHE INTERNAL "")
   set(JSON_Install
-      OFF
+      ON
       CACHE INTERNAL "")
```
ThomsonTan, line 239: Should we remove the `INTERNAL` attribute on the variable `JSON_Install` as we expect that the user could pass it into command line? Also provide a simple doc string here instead of leaving it as empty?

maxgolov: I borrowed this exact syntax based on `nlohmann::json` documentation. Described here:
https://json.nlohmann.me/integration/cmake/#supporting-both

Internal does not matter much - you can still provide an option to toggle it `OFF` in command line. It works well. And I think the only case when you'd want that OFF is MAYBE for Linux and Mac? And only if you build without zPages. This is minor detail, so I'd rather not bother changing it.

ThomsonTan (approved):

CMakeLists.txt
```diff
@@ -235,13 +235,21 @@ if(NOT nlohmann_json_FOUND)
...
+    # This option allows to add header to include directories
+    include_directories(
+      ${PROJECT_SOURCE_DIR}/third_party/nlohmann-json/single_include)
+  else()
+    message(
+      FATAL_ERROR
+        "\nnlohmann_json package was not found. Please either provide it manually or clone with submodules. "
+        "To initialize, fetch and checkout any nested submodules, you can use the following command:\n"
+        "git submodule update --init --recursive")
+  endif()
```
lalitb, line 252: while we are here, can we also add the change discussed earlier : https://github.com/open-telemetry/opentelemetry-cpp/pull/865#discussion_r654111208 ?
We do have an issue also for that now: https://github.com/open-telemetry/opentelemetry-cpp/issues/869, and I don't think we should need to checkout submodules for api only build.

lalitb (approved):

CMakeLists.txt
maxgolov: I'll address this for `nlohmann` for API-only build. But I anticipate other issues (both related to `nlohmann` json and unrelated to it) with building SDK from source tarball that does not cover recursive dependencies of SDK.

maxgolov: @lalitb - please take a look at latest iteration. This should now pass the API-only build, fixing #869

CMakeLists.txt
```diff
@@ -228,22 +228,33 @@ endif()
...
-      OFF
-      CACHE INTERNAL "")
-  # This option allows to link nlohmann_json::nlohmann_json target
-  add_subdirectory(${PROJECT_SOURCE_DIR}/third_party/nlohmann-json)
-  # This option allows to add header to include directories
-  include_directories(
-    ${PROJECT_SOURCE_DIR}/third_party/nlohmann-json/single_include)
-else()
-  message("Using external nlohmann::json")
+if(NOT WITH_API_ONLY)
```
maxgolov, line 231: @lidavidm - I hope that should fix the build for API-only build.

lidavidm: I can confirm this fixes the API-only build, thanks!

It would be nice if we could still build the rest of the library that doesn't require json without this dependency, though (e.g. the SDK), though I can understand if that would get complicated.

maxgolov: @lidavidm - it happens that in the majority of cases (`80% scenarios` I'd say), we do actually require `nlohmann/json.hpp`, e.g. for ETW on Windows, for zPages (internal monitor of SDK statistics), for Zipkin exporter, as well as for OTLP-HTTP exporter. Plus for all tests, and generally for examples too. Since we got you covered for the case when the build is done only for API and nothing else, I think it's gonna be too much of a hassle to make json optional.... as you noticed, it gets complicated - given that most components require it. And we give two alternatives - either consume json library from user-provided deps. Or use the one we cloned from recursive submodules, both no work cross-plat.

We can probably consider further improvements in a follow-up PR, but as low priority. Since all immediate scenarios have been addressed.

lalitb: I agree on having further cleanup in a follow-up PR. I think the most common usage scenario would be instrumentation libraries devs using otel-cpp api, and application developers using otel-gRPC exporter. And both these scenarios don't depend on JSON and so shouldn't fail. Though I do see the hassle involved here to work it cross-platform.
 
Thanks for fixing the API only build :)

maxgolov: @lalitb - My worry is that OTLP-HTTP exporter may also be common. This one does require JSON. Same as Zipkin. Also I think the zPages project is kinda common across all languages. The idea is we use zPages for debugging OpenTelemetry in-proc.. that build flavor of final SDK dll / shared object bit, with zPages in it, would require json.hpp

While I agree we may not need it for OTLP-gRPC only build without zPages, in most other configurations we end up needing json.hpp anyways.

lidavidm: I totally understand. I think the biggest thing is just that the release tarballs don't include the submodules, but that's a Github issue.

Issue #713, opened 2021-04-29
Allow to build SDK + contrib repo via `BUILD_CONTRIB` build option


maxgolov:
**Is your feature request related to a problem?** _Not really a problem..._

OpenTelemetry C++ SDK contrib repo:

https://github.com/open-telemetry/opentelemetry-cpp-contrib

may contain additional "non-standard" exporters that are maintained by community, e.g. `fluentd` exporter, for example.

**Describe the solution you'd like**

I would like to implement a build option for the main repo that allows to build contrib - `BUILD_CONTRIB` :
- by turning on a custom build option; OR
- by specifying the path to local checked out contrib snapshot, OR to alternate "contrib" overlay.

**Describe alternatives you've considered**

We could have added contrib repo as a submodule. Adding it as submodule presents other maintenance challenges.

**Additional context**

- **CMake** : when the SDK is built `BUILD_CONTRIB`, it should include the `CMakeLists.txt` from contrib path using `add_subdirectory(/path/to/contrib)` , that way the entire build tree of contrib repo is also included in the build.

- **Bazel** : contrib may be added as symlink (Unix) or directory juniction (Windows) at `/contrib` path. Bazeil would pick it up.
What it gives is ability to build a distro of SDK with additional exporters.

Note that the `contrib`:
- could be our contrib with community provided OSS exporters (mentioned above); or
- could be just totally different PRIVATE CUSTOM vendor contrib repository, with private closed-source non-standard vendor exporters.

**What we achieve is a single build that builds the SDK with custom additions?**

- Continuous CI of contrib against the main repository.
- Running all tests of contrib plus the main repository.
- Opportunity to customize the build targets and provide your own combined build target.
- Contrib knows the main repo targets; it can cover extra steps, such as custom binary packaging (e.g. nuget)

I'd like to illustrate how that would work for the following example:
- `fluentd` exporter lives in contrib repo.
- build option in the main allows to build with a snapshot of contrib repo overlaid on top of the main SDK repo.

**Bonus points**

Once the change is merged, we could have added a CI that implements an _OPTIONAL_ build check: verify that refactor in the main is not breaking anything significant in contrib. By building the main `BUILD_CONTRIB` against the latest `https://github.com/open-telemetry/opentelemetry-cpp-contrib` (main) branch snapshot of contrib. That way if refactor happens in the main that breaks any of the contrib exporters, we would know right away.

eyjohn:
[...]

PR #871, opened 2021-06-22
[WIP] foundation for building with contrib + sockets library - main repo part I

maxgolov:
**Changes**

PR implements the necessary plumbing for `fluentd` / `fluentbit` exporters that live in contrib repo. Since we already have a library for basic cross-plat sockets support (used by HTTP server), I refactored that library to support TCP, UDP and Unix Domain sockets. Subsequently, `fluentd` exporter uses that library to talk to end-point over TCP, UDP or Unix Domain.

To explain the structure:
- `/ext/include/opentelemetry/ext/net/common` - network library, socket client, socket server. Reusable, (still) used by HTTP server. Now with more sockets support, i.e. in addition to TCP (that was already there) - I added UDP and Unix Domain.
- `/ext/test/sockets` - cross-plat tests for the sockets.
- `CMakeLists.txt` - that allows to build with contrib specified via local directory, or by `FetchContent` from contrib repo.

Subsequently, contrib repo part II relies on some generic / general purpose code that lives in the main, e.g. in order to avoid reimplementing the same socket library again. The code is Apache License 2.0, all copyright OpenTelemetry Authors. No need to borrow any alternate BSD or Boost -licensed network library.

@lalitb - corresponding PR for the `fluentd` exporter is in the contrib repo is here:
https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/42


CMakeSettings.json
```diff
@@ -121,6 +121,86 @@
           "type": "BOOL"
         }
       ]
-    }
-  ]
+        },
```
maxgolov, line 124: I will remove this. This is just an illustration how to build with contrib from Visual Studio. I'll move this into separate document.

exporters/etw/test/etw_tracer_test.cc
```diff
@@ -14,6 +14,8 @@ using namespace OPENTELEMETRY_NAMESPACE;
 
 using namespace opentelemetry::exporter::etw;
```
maxgolov, line 15: I will revert this, as I submitted these changes in a separate PR #882

lalitb:
I would review it more thoroughly once this is ready for review, the general comments I have for now are:

- `socket library` - Very nicely done and this would be definitely useful for applications requiring cross-platform network libraries, we need to discuss whether the `main` repo is the right place for this. While I see the possibility of using the socket library internally for Jaeger thrift implementation (  cc @ThomsonTan  ), still not a convincing reason for exposing it as api/header to external systems. As an example, we do have an internal HTTP client and (test) server implementation, but this can't be used from external systems. The `main` repo should provide components for opentelemetry-api, opentelemetry-sdk, various core exporters, and not more than that. It can be in a separate repo so that both fluentd and thrift can use from there, or else let's keep it within fluentd exporter for now?

-  `For the building with Contrib repo` - This has been discussed in community meetings and in issue #713 with some good comments from @eyjohn, would be good to have wider feedback from @open-telemetry/cpp-approvers and @open-telemetry/cpp-maintainers, as my thoughts may be biased on my general perception of the build configuration :)

lalitb:
Closing this PR. Plan is to consolidate these changes with fluentd exporter in the `contrib` repo.

PR #872, opened 2021-06-22
Adding semantic-convention attributes for resources

lalitb:
This PR adds semantics as defined here:
https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions

This will let us use uniform attributes/events/exceptions names across services.

### DIFF
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -120,7 +120,7 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
 {
   // only service.name attribute is supported by specs as of now.
   auto attributes = resource.GetAttributes();
-  if (attributes.find(SemanticConventions::GetAttributeSericeName()) != attributes.end())
+  if (attributes.find(SemanticConventions::GetAttributeServiceName()) != attributes.end())
   {
     service_name_ =
         nostd::get<std::string>(attributes[SemanticConventions::GetAttributeSericeName()]);
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -122,8 +122,7 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
   auto attributes = resource.GetAttributes();
   if (attributes.find(OTEL_CPP_GET_ATTR(AttrServiceName)) != attributes.end())
   {
-    service_name_ =
-        nostd::get<std::string>(attributes[OTEL_CPP_GET_ATTR(AttrServiceName()]);
+    service_name_ = nostd::get<std::string>(attributes[OTEL_CPP_GET_ATTR(AttrServiceName)]);
   }
 }
 
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -1,8 +1,10 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-#include <opentelemetry/exporters/jaeger/recordable.h>
+#include "opentelemetry/exporters/jaeger/recordable.h"
+#include "opentelemetry/sdk/resource/semantic_conventions.h"
 
+using namespace opentelemetry::sdk::resource;
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter
 {
@@ -117,9 +119,10 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
 {
   // only service.name attribute is supported by specs as of now.
   auto attributes = resource.GetAttributes();
-  if (attributes.find("service.name") != attributes.end())
+  if (attributes.find(SemanticConventions::GetAttributeSericeName()) != attributes.end())
   {
-    service_name_ = nostd::get<std::string>(attributes["service.name"]);
+    service_name_ =
+        nostd::get<std::string>(attributes[SemanticConventions::GetAttributeSericeName()]);
   }
 }
 
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -4,12 +4,13 @@
 #include "opentelemetry/exporters/jaeger/recordable.h"
 #include "opentelemetry/sdk/resource/semantic_conventions.h"
 
-using namespace opentelemetry::sdk::resource;
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter
 {
 namespace jaeger
 {
+
+using namespace opentelemetry::sdk::resource;
 
 Recordable::Recordable() : span_{new thrift::Span} {}
 
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -123,7 +123,7 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
   if (attributes.find(SemanticConventions::GetAttributeServiceName()) != attributes.end())
   {
     service_name_ =
-        nostd::get<std::string>(attributes[SemanticConventions::GetAttributeSericeName()]);
+        nostd::get<std::string>(attributes[SemanticConventions::GetAttributeServiceName()]);
   }
 }
 
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -120,10 +120,10 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
 {
   // only service.name attribute is supported by specs as of now.
   auto attributes = resource.GetAttributes();
-  if (attributes.find(SemanticConventions::GetAttributeServiceName()) != attributes.end())
+  if (attributes.find(OTEL_CPP_GET_ATTR(AttrServiceName)) != attributes.end())
   {
     service_name_ =
-        nostd::get<std::string>(attributes[SemanticConventions::GetAttributeServiceName()]);
+        nostd::get<std::string>(attributes[OTEL_CPP_GET_ATTR(AttrServiceName()]);
   }
 }
 
--- a/sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
+++ b/sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
@@ -1,0 +1,136 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "opentelemetry/version.h"
+
+#define OTEL_RESOURCE_GENERATE_SEMCONV_METHOD(attribute_name, attribute_value) \
+  static constexpr const char *GetAttribute##attribute_name() noexcept { return attribute_value; }
+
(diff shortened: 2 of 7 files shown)


lalitb:
@maxgolov @ThomsonTan - Semantic convention for the `Resources` is also done similar to that for traces in #868 , ie using `constexpr` static methods defined in the class. Though this is sdk, and we have the flexibility to define it within the source file, and reference it through `extern`, I still felt that the `constexpr` method approach is better for the reasons mentioned here: https://github.com/open-telemetry/opentelemetry-cpp/pull/868#discussion_r655967079.

Feel free to comment on a better approach to follow here and/or for traces. thanks.

exporters/jaeger/src/recordable.cc
```diff
@@ -117,9 +120,10 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
 {
   // only service.name attribute is supported by specs as of now.
   auto attributes = resource.GetAttributes();
-  if (attributes.find("service.name") != attributes.end())
+  if (attributes.find(SemanticConventions::GetAttributeSericeName()) != attributes.end())
```
pyohannes, line 123:
```suggestion
  if (attributes.find(SemanticConventions::GetAttributeServiceName()) != attributes.end())
```

sdk/test/resource/resource_test.cc
```diff
@@ -17,30 +18,29 @@
...
-  opentelemetry::sdk::resource::ResourceAttributes expected_attributes = {
+  ResourceAttributes expected_attributes = {
       {"service", "backend"},
       {"version", (uint32_t)1},
       {"cost", 234.23},
-      {"telemetry.sdk.language", "cpp"},
-      {"telemetry.sdk.name", "opentelemetry"},
-      {"telemetry.sdk.version", OPENTELEMETRY_SDK_VERSION},
-      {"service.name", "unknown_service"}};
+      {SemanticConventions::GetAttributeTelemetrySdkLanguage(), "cpp"},
```
maxgolov, line 36: I honestly would prefer an API model similar to Qt5 "translation resources", which solves the same problem - just in a different knowledge domain. More on that [here](https://doc.qt.io/qt-5/qtlinguist-hellotr-example.html)

To elaborate:

- let's say you have a namespace called `MyAttributeNames` , and either `constexpr` or `extern` that looks like Qt5 `tr(name)` ... Let's name it `attr(name)`

- then in order to obtain the actual name for a given system, instead of `SemanticConventions::GetAttributeTelemetrySdkLanguage()` , you'd write `attr("sdk.language").... Moreover, since you are not gonna be using all attributes at once, it is Okay to declare these as literals, e.g. OTEL_ATTR_SDK_LANGUAGE

Then instead of having N setters, which will also present a challenge - should you ever need to project this C++ code to C#, Python or Javascript/Node, etc... you'd have ONE method - called `AttributeNames::attr` and you'd invoke it as:

```cpp
using namespace AttributeNames;
...
{attr(OTEL_ATTR_SDK_LANGUAGE), "cpp"}
... // or
{attr("sdk.language", "cpp")}
```

Then we should strive the transform to be done at compile time, without generating those 50 methods. It does not scale. It'd scale nicely if you have one method / one utility function, hopefully `constexpr`. And a table of literals, `resources` in a classic definition of this world in C++ (as in translation resources, resource compiler, etc.).

Again, `tr()` from Qt5 seems like the most reasonable way to achieve that.

sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
```diff
@@ -0,0 +1,136 @@
...
+ * specifications. <see
+ * href,"https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md"/>.
+ */
+
+class SemanticConventions final
+{
+
+public:
+  // service attributes
+  OTEL_RESOURCE_GENERATE_SEMCONV_METHOD(ServiceName, "service.name");
```
maxgolov, line 27: You can do this instead. Take this compile-time hash function:

```cpp
/// <summary>
/// Compile-time constexpr djb2 hash function for strings
/// </summary>
static constexpr uint32_t hashCode(const char *str, uint32_t h = 0)
{
  return (uint32_t)(!str[h] ? 5381 : ((uint32_t)hashCode(str, h + 1) * (uint32_t)33) ^ str[h]);
}

#define CONST_UINT32_T(x) std::integral_constant<uint32_t, (uint32_t)x>::value

#define CONST_HASHCODE(name) CONST_UINT32_T(OPENTELEMETRY_NAMESPACE::utils::hashCode(#name))
```

Then for every attribute, generate a table of values. At compile time - by listing all of them. Sorry for using this [example](https://github.com/microsoft/cpp_client_telemetry/blob/master/lib/tracing/api/module_list) - something like this:
```cpp
// Hash function below is computing module tag from literal at compile-time using C++11 constexpr
#define DBG_MODULE_CTAG(name)       CONST_UINT32_T(djb2_constexpr_hash(#name))
#define DBG_MODULE_TAG(name)        djb2_hash(name)
#define DBG_LOG_NAMETAG(name)       { DBG_MODULE_CTAG(name) , #name } ,
```

So - you created a compile time table that maps your attribute names to unique IDs. You can still refer to original attributes by their aliases. Like `ATTR_ServiceName`, macro-magic permitting.

The goal is that you can do tricks like switch-case, constexpr, and mapping from attribute to unique attribute id. Then instead of many methods, you end up with `attr(name)` or `attr(id)`. That's how Qt5 `tr()` works, and how translations / resources work in Android pretty much. Except that `tr()` requires string literal, and Android resources are all revolving around `getResource(id)`.

In our model, I think, it'd be more appropriate to perform compile-time mapping from `attr("my.attribute.name")` -> returns actual attribute name within given semantic convention. Not many getters. One getter. Lookup table. And preferably, compile-time lookup, not method call.

sdk/test/resource/resource_test.cc
```diff
@@ -61,17 +61,17 @@ TEST(ResourceTest, create_without_servicename)
...
 {
-  opentelemetry::sdk::resource::ResourceAttributes expected_attributes = {
+  ResourceAttributes expected_attributes = {
       {"version", (uint32_t)1},
       {"cost", 234.23},
-      {"telemetry.sdk.language", "cpp"},
-      {"telemetry.sdk.name", "opentelemetry"},
-      {"telemetry.sdk.version", OPENTELEMETRY_SDK_VERSION},
-      {"service.name", "backend"},
+      {SemanticConventions::GetAttributeTelemetrySdkLanguage(), "cpp"},
```
maxgolov, line 67: Actually to elaborate on my idea - you probably do not even need to change anything. Keep passing original string literals. We have an assignment operator in the `ResourceAttributes` implementation, that could have done the semantic transformation depending on what exporter / system is being used. Then you'd just completely avoid calling getters. You'd pre-configure your `ResourceAttributes` by specifying destination convention (like, imagine - your `ResourceAttributes` class can translate from English to French, and you pass in English keys, but they get magically translated to French, because your system "locale" is French).

If you need concrete example, I can send some WIP code to illustrate the model?

lalitb: Thanks @maxgolov for the elaboration. I can try implementing this model, but if you can provide a more concrete implementation example it would be helpful.

maxgolov: We can definitely avoid getters. OR - at least, it all boils down to a single method that allows to transform from schema-agnostic common `attribute name` or `attribute id` to semantic- / schema- specific attribute name, the way it shows up on wire. I feel like more concrete code example (based off above constexpr hash calculation) would be helpful?

lalitb: @maxgolov  I have done the changes (hopefully) as per your suggestion :)
Have kept both the alternatives for now so that we can compare both and finalize.
`semantic_conventions.h` -> the suggested change.
`semantic_conventions_old.h` -> Initial implementation

maxgolov (approved):
Yes, thank you for implementing it that way. This implementation is in alignment with what I strongly believe is reasonable way of doing it! ❤️ We may think about generating the table automatically from schema in a separate follow-up PR, later when this is needed. Manually changing it as a table is also looking much cleaner.

lalitb:
> Yes, thank you for implementing it that way. This implementation is in alignment with what I strongly believe is reasonable way of doing it! ❤️ We may think about generating the table automatically from schema in a separate follow-up PR, later when this is needed. Manually changing it as a table is also looking much cleaner.

Make sense. Will do some cleanup in this PR - removing old semantic code, and move the hash function to `api` so that it can be utilized for both `traces` and `resources`. And then modify PR #868 similarly.

exporters/jaeger/src/recordable.cc
```diff
@@ -117,9 +120,10 @@ void Recordable::SetResource(const opentelemetry::sdk::resource::Resource &resou
 {
   // only service.name attribute is supported by specs as of now.
   auto attributes = resource.GetAttributes();
-  if (attributes.find("service.name") != attributes.end())
+  if (attributes.find(OTEL_CPP_GET_ATTR(AttrServiceName)) != attributes.end())
   {
-    service_name_ = nostd::get<std::string>(attributes["service.name"]);
+    service_name_ =
+        nostd::get<std::string>(attributes[OTEL_CPP_GET_ATTR(AttrServiceName()]);
```
ThomsonTan, line 126: Is parenthesis needed in `AttrServiceName()`?

sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
```diff
@@ -0,0 +1,139 @@
+#include <stdint.h>
+#include <iostream>
+#include <type_traits>
+#include <unordered_map>
+#include "opentelemetry/version.h"
+
+// typedef uint32_t Tag;
```
ThomsonTan, line 7: Remove this unused code?

sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
```diff
@@ -0,0 +1,139 @@
...
+
+static constexpr uint32_t hashCode(const char *str, uint32_t h = 0)
+{
+  return (uint32_t)(!str[h] ? 5381 : ((uint32_t)hashCode(str, h + 1) * (uint32_t)33) ^ str[h]);
```
ThomsonTan, line 11: Is there any reference for this hash function? Put a link here if so.

exporters/jaeger/src/recordable.cc
lalitb: No, it was a typo. I should start testing Jaeger too locally now :). Fixed now.

sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
lalitb: Done.

sdk/include/opentelemetry/sdk/resource/semantic_conventions.h
lalitb: Good point. Have moved this function to `api/include/opentelemetry/common/string_util.h` and the reference of the hash algorithm.

lalitb:
@ThomsonTan  - I have resolved the comments from you. I hope it's fine to merge it, as I need to update #868 PR on top of these changes. If you have some other comments, I can take them in separately.

PR #879, opened 2021-06-29
Associate valid SpanId/SpanContext with Spans irrespective of the Sampling decision.

### LINKED ISSUE #878: Parent-based sampler disconnected family trees
**Describe your environment**
Using a 1.0.0 release candidate.

**Steps to reproduce**
Create a parent-based sampler with a ratio-based delegate sampler and a 0.5 sampling ratio.  Create multiple generations of spans.  You'll find that instead of always getting either a single trace with all the spans in it (from a sampled ancestor) or no traces (from a non-sampled ancestor), you randomly get multiple traces (a disconnected family tree).

The parent-based sampler should delegate sampling decisions to a different sampler only if a sampling decision was not previously made for a parent context (i.e., if the parent context is invalid).  Repeated use of the parent sampler on descendants of a non-sampled span should result in a single, non-exported trace, just as it results in a single, exported trace from a sampled ancestor.

Descendants of non-sampled spans have invalid context instead.  This results in many "orphan spans" when the delegate sampler, which handles invalid span contexts, kicks in.

**Additional context**
I think the correct behavior should be to always create spans with valid IDs.  Although I'm not so familiar with the Go and Java SDKs, it seems like they always generate spans with valid IDs.

lalitb:
Fixes #878 

## Changes

As per [Specs](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/trace/sdk.md#sdk-span-creation), non-sampled / dropped spans should still have the valid SpanId. The child span of the non-sampled span should have the same trace_id associated.

``
Generate a new span ID for the Span, independently of the sampling decision. This is done so other components (such as logs or exception handling) can rely on a unique span ID, even if the Span is a non-recording instance.
``

This PR ensures that a valid span-id is associated with the non-sampled spans.


maxgolov (approved):

api/include/opentelemetry/trace/noop.h
```diff
@@ -55,11 +60,11 @@ class NoopSpan final : public Span
 
   bool IsRecording() const noexcept override { return false; }
 
-  SpanContext GetContext() const noexcept override { return span_context_; }
+  SpanContext GetContext() const noexcept override { return *span_context_.get(); }
```
ThomsonTan, line 63:
```suggestion
  SpanContext GetContext() const noexcept override { return span_context_->get(); }
```

lalitb: This is what we are doing here : `*(span_context.get())` . Don't think `span_context_->get()` will work.

lalitb: @ThomsonTan - I hope you are fine with the comment. Else, we will resolve this in another PR.

PR #885, opened 2021-06-30
Cleanup of context from Runtime storage

### LINKED ISSUE #884: Scope object doesn't release the ownership of Span's shared_ptr
**Describe your environment** 
gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)

**Steps to reproduce**
Sample code:
```cpp
TEST(Tracer, TestAlwaysOnSampler)
{
  std::unique_ptr<InMemorySpanExporter> exporter(new InMemorySpanExporter());
  std::shared_ptr<InMemorySpanData> span_data = exporter->GetData();
  auto tracer_on                              = initTracer(std::move(exporter));
  // Testing AlwaysOn sampler.
  // Create two spans for each tracer. Check the exported result.
  {
    auto span_on_1 = tracer_on->StartSpan("span 1");
    auto span_on_2 = tracer_on->StartSpan("span 2");
    {
      Scope scope(span_on_1);
    }
  }
  auto spans = span_data->GetSpans();
  ASSERT_EQ(1, spans.size());
  ASSERT_EQ("span 2", spans.at(0)->GetName());  // span 2 ends.
  //ASSERT_EQ("span 1", spans.at(1)->GetName()); // span 1 does not end even though it is out of scope.
}
```

`  ASSERT_EQ(1, spans.size());`  there should be two spans in memory exporter in the end.

`  ASSERT_EQ(1, spans.size());`  there is only one span in memory exporter in the end.

What did you see instead?
The actual behavior is seen which is incorrect.

**Additional context**
Runtime context management needs to be fixed.

lalitb:
Fixes #884 

## Changes

 [`ThreadLocalContextStorage` ](https://github.com/open-telemetry/opentelemetry-cpp/blob/c0b459e75d73bc51137f51c3c88871cf53608a40/api/include/opentelemetry/context/runtime_context.h#L175 )is used as default runtime storage, and is internally implemented as a `Stack`. The [Pop()](https://github.com/open-telemetry/opentelemetry-cpp/blob/c0b459e75d73bc51137f51c3c88871cf53608a40/api/include/opentelemetry/context/runtime_context.h#L227) operation doesn't perform the cleanup of stored `Context` object, but only decrements the stack size.  The actual cleanup happens only when that storage is later re-utilized with a new Context. The side effect of this approach is that any `shared_ptr` object stored in the popped-out Context will not be cleaned up at the time of Pop() operation.

The fix is to replace the current context (meant to be popped out) with an empty `Context` object before decrementing the size.
 This will release the ownership of any `shared_ptr` object stored in this current context.

Also, Stack::Pop() shouldn't be returning the removed object, to be consistent with std::stack api.


api/include/opentelemetry/context/runtime_context.h
```diff
@@ -223,15 +223,19 @@ class ThreadLocalContextStorage : public RuntimeContextStorage
...
       if (size_ == 0)
       {
-        return Context();
+        return;
       }
+      // Store empty Context before decrementing `size`, to ensure
+      // the shared_ptr object (if stored in prev context object ) are released.
+      // The stack is not resized, and the unused memory would be reutilised
+      // for subsequent context storage.
+      base_[size_ - 1] = Context();
```
ThomsonTan, line 237: How could someone get a `shared_ptr` of the `Context` stored in `ThreadLocalContextStorage`?

lalitb: Line 237 is actually resetting the `base[size - 1]` with empty Context() object. This is to ensure that previously-stored shared_ptr<Span> gets released. The new empty Context() is not supposed to be used again.

ThomsonTan (approved):

lalitb:
@maxgolov - Would like to include these changes in the next release. Would it be good to merge, as you wanted to do some multi-threaded tests with these changes?

lalitb:
@maxgolov Merging this. If there are issues found in your testing, we can address them separately. Thanks.

PR #887, opened 2021-07-01
Change Zoom link for Community meeting.

lalitb:
Fix

## Changes

Changing Zoom link for C++ Community meeting to ensure we don't share the same meeting link with the prior meeting of Agent/Collector SIG. Basically, we are swapping our meeting ID with .Net Auto Instrumentation SIG.

Community README is updated: https://github.com/open-telemetry/community/pull/766
Calendar and meeting notes have been updated. 

This new meeting link would be used for Monday's and Wednesday's meetings.

PR #891, opened 2021-07-02
Remove meeting link in contributing doc

ThomsonTan:
## Changes

We updated Zoom meeting in README in https://github.com/open-telemetry/opentelemetry-cpp/pull/887, but we also have the same link in contributing doc which needs to be fixed as well. Instead, this PR removes the duplicated link in the contributing doc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,7 +142,7 @@ on each other), the owner should try to get people aligned by:
 
 If none of the above worked and the PR has been stuck for more than 2 weeks, the
 owner should bring it to the OpenTelemetry C++ SIG meeting. See
-[README.md](README.md#Contributing) for the meeting link.
+[README.md](README.md#contributing) for the meeting link.
 
 ## Useful Resources
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,8 +141,8 @@ on each other), the owner should try to get people aligned by:
   split it up.
 
 If none of the above worked and the PR has been stuck for more than 2 weeks, the
-owner should bring it to the [OpenTelemetry C++ SIG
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -141,8 +141,8 @@ on each other), the owner should try to get people aligned by:
   split it up.
 
 If none of the above worked and the PR has been stuck for more than 2 weeks, the
-owner should bring it to the [OpenTelemetry C++ SIG
-meeting](https://zoom.us/j/8203130519). The meeting passcode is _77777_.
+owner should bring it to the OpenTelemetry C++ SIG meeting. See
+[README.md](README.md#Contributing) for the meeting link.
```
reyang, line 145: `README.md#Contributing` or `README.md#contributing` or doesn't matter (uppercase or lowercase)?

ThomsonTan: Seems casing doesn't matter in URL. [this](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/README.md#CoNtrIbutinG) URL works as well in my browser.

reyang: I don't think so.

https://www.w3.org/TR/WD-html40-970708/htmlweb.html

_URLs in general are case-sensitive (with the exception of machine names). There may be URLs, or parts of URLs, where case doesn't matter, but identifying these may not be easy. **Users should always consider that URLs are case-sensitive.**_

reyang (approved):
LGTM.

CONTRIBUTING.md
ThomsonTan: Thanks for the sharing. It makes sense for the URL to be case-sensitive as it shows up in many short URLs. Perhaps github does the toleration.

PR #911, opened 2021-07-19
Fix for resource deletion after tracer provider shutdown

### LINKED ISSUE #904: BatchSpanProcessor accesses deleted resources after tracer destruction
**Describe your environment**
Using a version after v1.0.0-rc2, including recent fixes to scopes and `StartSpan`.

**Steps to reproduce**
Destroy a tracer provider connected to a batch span processor while it is processing spans.

Even on tracer provider destruction, which could happen if a user wants to change the tracing configuration, the batch span processor will emit spans with resources intact.

Resource attributes are absent from spans emitted by the batch span processor on tracer provider destruction.

**Additional context**
I've observed the following sequence of events.
1. `BatchSpanProcessor::OnEnd` adds the span to its circular buffer.  The resource has the expected attributes.
2. `BatchSpanProcessor::OnEnd` does not notify the worker thread.
3. The tracer's destructor gets called, due to intentional destruction of the global tracer provider.
4. That tracer's resource destructor gets called (due to tracer destruction).
5. `BatchSpanProcessor::DrainQueue` gets called.
6. `BatchSpanProcessor::Export` consumes the circular buffer.  The resource (still at the same address) has no attributes.

Perhaps the destruction order needs to be adjusted, or recordables should not store [bare pointers](https://github.com/open-telemetry/opentelemetry-cpp/blob/79c4c718d576c6e66adf91bd74cf49bd3d394ee9/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_recordable.h#L61) to resources.

See the [Slack discussion](https://cloud-native.slack.com/archives/C01N3AT62SJ/p1625599122068700) for a similar problem occurring.

lalitb:
Fixes #904 

## Changes

Change in the order of declarations of tracer-context, instrumentation library, resources, and processor to ensure that resources and instrumentation library objects are destroyed after processor.
Adding test-case is a bit tricky ( would be doing it before merging the PR): but this is the batch-processor example output when tracer provider is deleted ( before and after changes):

Before change:
```
Shutting down and draining queue....
{
  name          : Span 1
  trace_id      : 738d6e1ac63df645a59696e9e8b3bc7b
  span_id       : 2bb3160e4738beae
  tracestate    :
  parent_span_id: 0000000000000000
  start         : 1626709154132876300
  duration      : 1700
  description   :
  span kind     : Internal
  status        : Unset
  attributes    :
  events        :
  links         :
  resources     :
  instr-lib     : "VُY7`Y"V`0"V."V."VA0@"V"V>"V>"VA]f(JY¤C`Y"V`0"V."VSY;`?"VSpan 1?"V?"V?(4"V05"VA"Vp+"VA>"]5@'@'?ײE
`Y"V`0"V."VTY;@"VSpan A"VHA"V?(4"V05"VAp@"]$Xj[W=}4N`Y"V`0"V."V'UY; B"VSpan 3HB"V
```

After change:
```
Shutting down and draining queue....
{
  name          : Span 10
  trace_id      : faa0a602c715e0019e0c9da8cac147bc
  span_id       : 18cd2ae1bc8d2724
  tracestate    :
  parent_span_id: 0000000000000000
  start         : 1626708952913928200
  duration      : 1600
  description   :
  span kind     : Internal
  status        : Unset
  attributes    :
  events        :
  links         :
  resources     :
        service.name: unknown_service
        telemetry.sdk.version: 1.0.0-rc3
        telemetry.sdk.name: opentelemetry
        version: 1
        telemetry.sdk.language: cpp
        service: test_service
  instr-lib     : foo_library
}

```

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -60,6 +60,8 @@ public:
   Sampler &GetSampler() { return context_->GetSampler(); }
 
 private:
+  // order of declaration is important here - instrumentation library should destroy after
+  // tracer-context.
   std::shared_ptr<InstrumentationLibrary> instrumentation_library_;
   std::shared_ptr<sdk::trace::TracerContext> context_;
 };
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -60,8 +60,8 @@ public:
   Sampler &GetSampler() { return context_->GetSampler(); }
 
 private:
-  std::shared_ptr<sdk::trace::TracerContext> context_;
   std::shared_ptr<InstrumentationLibrary> instrumentation_library_;
+  std::shared_ptr<sdk::trace::TracerContext> context_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -29,6 +29,7 @@ load("@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")
 
 grpc_extra_deps()
 
+load("@com_github_grpc_grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")
 load("@upb//bazel:workspace_deps.bzl", "upb_deps")
 
 upb_deps()
--- a/docs/building-with-vcpkg.md
+++ b/docs/building-with-vcpkg.md
@@ -102,6 +102,18 @@ dependencies:
   C++14 or C++17 compiler.
 - `nlohmann-json` - required when building with zPages module.
 - `prometheus-cpp` - required for Prometheus exporter.
+- `gRPC` and `Protobuf` - required for OTLP exporter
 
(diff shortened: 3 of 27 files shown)


seemk (approved):
Stumbled upon this, tested this PR and can confirm this fixes the use after free.

Thanks!

lalitb:
@open-telemetry/cpp-approvers need help reviewing this. Its small change to reorder declaration for the fix.

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -60,8 +60,8 @@ class Tracer final : public trace_api::Tracer, public std::enable_shared_from_th
   Sampler &GetSampler() { return context_->GetSampler(); }
 
 private:
-  std::shared_ptr<sdk::trace::TracerContext> context_;
   std::shared_ptr<InstrumentationLibrary> instrumentation_library_;
```
ThomsonTan, line 63: Could we document the order requirement in the comment as well?

ThomsonTan (approved):

sdk/include/opentelemetry/sdk/trace/tracer.h
lalitb: good point, documented it now.

Issue #656, opened 2021-04-06
Allow to programmatically populate settings for OTELResourceDetector


maxgolov:
Presently resource detector uses environment variable here:

https://github.com/open-telemetry/opentelemetry-cpp/blob/484700a0da2547852653271d2bdc10b50830bb86/sdk/src/resource/resource_detector.cc#L15

This approach won't work for Windows Universal Store (e.g. Xbox) applications:
https://github.com/MicrosoftDocs/cpp-docs/blob/master/docs/cppcx/crt-functions-not-supported-in-universal-windows-platform-apps.md

Because these applications cannot use environment variables. There may be some alternative approach available to populate the variable programmatically? e.g. using something like `OTELResourceDetector::SetResourceAttributes(...)` , that way the `Detect` method can still return the `Resource` object successfully after. But this time - it is built from programmatically-passed value, not obtaining it from env var.

Then UWP store app may use some remote config provisioning technique to obtain the settings, e.g. from remote JSON file first. Then pass these resource values to `OTELResourceDetector` via `SetResourceAttributes`. Then the rest of code should still be able to call into `OTELResourceDetector().Detect()` as before. Auto-detecting the settings passed to it via API, not from env var.

This small improvement should make the SDK compile'able for Universal Store / Xbox apps. This is not affecting system services, or web-services. But may become relevant if SDK is used in client apps at some point in future. We can add the necessary `#ifdef WINDOWS_UWP` to avoid `getenv` if targeting Xbox Games or Windows Universal Store apps.

@lalitb

lalitb:
@maxgolov  there are three approaches to specify the resource attributes according to  specs 
- [Resource::Create](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#create): Resource attributes can be specified as an argument to this function. This is something that can be leveraged by UWP.
- [Custom resource detectors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#create): Custom resource detectors can be created for platforms ( eg, UWP here) outside SDK. This can be created in a way not depend on the env variable. 
[...]

PR #924, opened 2021-07-26
Move env-var read logic to common dir, and optional reading of env-var

lalitb:
## Changes

1. Move env var reading logic to sdk::common namespace, this allows it to be reused at multiple places.
2. Optional reading of env variables based on macro -DNO_GETENV. The env var is not read if the macro is set. This allows using sdk with platforms like UWP which doesn't support env variables. ( as discussed in #656 ).
3. Make required otlp-grpc exporter options to be read through env variable, which further allows loading this exporter at run-time using the dynamic-load plugin. This would be further added for other exporters also.


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter.h
```diff
@@ -11,26 +11,56 @@
...
+
+  auto endpoint = opentelemetry::sdk::common::GetEnvironmentVariable(kOtlpGrpcEndpointEnv);
+  return endpoint.size() ? endpoint : kOtlpGrpcEndpointDefault;
+}
+
+inline const bool GetOtlpGrpcDefaultIsSslEnable()
+{
+  constexpr char kOtlpGrpcIsSslEnableEnv[] = "OTEL_EXPORTER_OTLP_GRPC_SSL_ENABLE";
+  auto ssl_enable = opentelemetry::sdk::common::GetEnvironmentVariable(kOtlpGrpcIsSslEnableEnv);
+  if (ssl_enable == "True" || ssl_enable == "TRUE" || ssl_enable == "true" || ssl_enable == "1")
```
ThomsonTan, line 35: Convert it to lower case before comparison, to make the comparison case insensitive?

ThomsonTan (approved):

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter.h
lalitb: Good point will change that.

lalitb: Ok. I realize there is no method available in `stdlib` to perform this conversion. we have to iterate over the string to convert it.  The existing comparison looks simple as compared to that :)

PR #925, opened 2021-07-26
Fix exporting of Jaeger exporter target name

seemk:
I believe it was introduced in https://github.com/open-telemetry/opentelemetry-cpp/pull/918

When doing `make install` the generated `opentelemetry-cpp-config.cmake` was trying to find a non-existent target, thus the `OPENTELEMETRY_CPP_LIBRARIES` variable did not have Jaeger exporter set.

### DIFF
--- a/exporters/jaeger/CMakeLists.txt
+++ b/exporters/jaeger/CMakeLists.txt
@@ -11,14 +11,14 @@ set(JAEGER_THRIFT_GENCPP_SOURCES
 
 set(JAEGER_EXPORTER_SOURCES
     src/jaeger_exporter.cc src/thrift_sender.cc src/udp_transport.cc
-    src/recordable.cc src/TUDPTransport.cc)
+    src/recordable.cc src/TUDPTransport.cc src/http_transport.cc src/THttpTransport.cc)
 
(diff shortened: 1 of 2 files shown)


lalitb:
Sorry, my bad. Thanks for raising the PR for the fix. 
Just wondering if we should instead revert the EXPORT_NAME to `jaeger_trace_exporer` to be inconsistent with how others are named in `OPENTELEMETRY_CPP_LIBRARIES`  variable.

https://github.com/open-telemetry/opentelemetry-cpp/blob/1f313419e77e8bd762ba5f0e6739a73aca52a211/exporters/jaeger/CMakeLists.txt#L21

seemk:
> Sorry, my bad. Thanks for raising the PR for the fix.
> Just wondering if we should instead revert the EXPORT_NAME to `jaeger_trace_exporer` to be consistent with how other targets are listed in `OPENTELEMETRY_CPP_LIBRARIES` variable.
> 
> https://github.com/open-telemetry/opentelemetry-cpp/blob/1f313419e77e8bd762ba5f0e6739a73aca52a211/exporters/jaeger/CMakeLists.txt#L21

Thanks! Changed the export name instead :+1:

lalitb (approved):

ThomsonTan (approved):

PR #927, opened 2021-07-28
Use latest grpc version (v1.39.0) for cmake build of otlp exporter.

lalitb:
## Changes

- Use the latest grpc version ( v1.39.0) for cmake builds of otlp exporter when compiled with g++5.1 or greater
- Use the compatible grpc version (v1.33.2) for cmake builds of otlp exporter when compiled with g++4.8 or g++5.0 compiler.
- Add github workflow to build otlp exporter using g++4.8 compiler ( with grpc v1.33.2 )
- Install abseil C++ library -  as grpc v1.39.0 service api has a dependency on this library  ( and we use this  for otlp mock)

we still need similar support for bazel. As of now, bazel builds uses grpc v1.33.2. 


ci/do_ci.sh
```diff
@@ -92,6 +92,22 @@ elif [[ "$1" == "cmake.legacy.test" ]]; then
...
+  export BUILD_ROOT="${BUILD_DIR}"
+  ${SRC_DIR}/tools/build-gtest.sh
+  ${SRC_DIR}/tools/build-benchmark.sh
+  cmake -DCMAKE_BUILD_TYPE=Debug  \
+        -DCMAKE_CXX_STANDARD=11 \
+        -DWITH_OTLP=ON \
+        "${SRC_DIR}"
+  grpc_cpp_plugin=`which grpc_cpp_plugin`
+  proto_make_file="CMakeFiles/opentelemetry_proto.dir/build.make"
+  sed -i "s~gRPC_CPP_PLUGIN_EXECUTABLE-NOTFOUND~$grpc_cpp_plugin~" ${proto_make_file} #fixme
```
ThomsonTan, line 107: What is this for?

ci/do_ci.sh
```diff
@@ -92,6 +92,22 @@ elif [[ "$1" == "cmake.legacy.test" ]]; then
...
   exit 0
+elif [[ "$1" == "cmake.legacy.exporter.otprotocol.test" ]]; then
+  cd "${BUILD_DIR}"
+  rm -rf *
+  export BUILD_ROOT="${BUILD_DIR}"
+  ${SRC_DIR}/tools/build-gtest.sh
+  ${SRC_DIR}/tools/build-benchmark.sh
+  cmake -DCMAKE_BUILD_TYPE=Debug  \
+        -DCMAKE_CXX_STANDARD=11 \
+        -DWITH_OTLP=ON \
```
ThomsonTan, line 103: Do you think we could set `WITH_OTLP` to `ON` by default?

ci/do_ci.sh
lalitb: This is actually copied from here (existing otlp build action on ubuntu-latest):
https://github.com/open-telemetry/opentelemetry-cpp/blob/ef16d0048a0ccd6d664b86da1c95294633965bad/ci/do_ci.sh#L103

Somehow, cmake is not able to find the `grpc_cpp_plugin` executable during configuration and so it has to be replaced manually in generated `make` file before doing the actual `make`.
I will debug this separately and fix in another PR.

ci/do_ci.sh
lalitb: You mean make it default here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/ef16d0048a0ccd6d664b86da1c95294633965bad/CMakeLists.txt#L144

I will raise an issue for this, and we can handle it separately as this is not related to this PR?

ThomsonTan: Sounds good.

ThomsonTan (approved):

PR #937, opened 2021-08-03
Fix casting to microseconds

srgbtl:
## Changes
std::chrono::duration can be converted to integral value via count() function


lalitb (approved):
Thanks for the fix.

ThomsonTan (approved):

PR #938, opened 2021-08-03
[WIP] HTTP exporter support ssl credentials

pavanshahm:
Fixes # (issue)

## Changes

We needed to include ssl credentials if we wanted to use the http exporter so we added an additional option in http_exporter_options in order to allow it. 

We also needed to update the curl.BUILD files in order to build cURL with the https protocol supported. This was done with a lot of trial and error and copied from the tensorflow build file, so I'm not sure which options are actually necessary and which ones are not. 

Unit Tests are pending, but let me know if this is the right approach before I write those tests. 


lalitb:
> I have thought using [civetweb](https://github.com/civetweb/civetweb/) to setup a temporary HTTP server for unit test. Because [prometheus-cpp](https://github.com/jupp0r/prometheus-cpp) also depend it. We can also use it without import another library or binary.
> It maybe need more discussion about how to test HTTP requests. @maxgolov @lalitb

@owent  - This looks good - production-ready embeddable web-server, but just wondering whether we really need this for unit tests. What advantage do you think we will get from this over our internal [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) specifically for unit-tests?

owent (commented):
> > I have thought using [civetweb](https://github.com/civetweb/civetweb/) to setup a temporary HTTP server for unit test. Because [prometheus-cpp](https://github.com/jupp0r/prometheus-cpp) also depend it. We can also use it without import another library or binary.
> > It maybe need more discussion about how to test HTTP requests. @maxgolov @lalitb
> 
> @owent - This looks good - production-ready embeddable web-server, but just wondering whether we really need this for unit tests. What advantage do you think we will get from this over our internal [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) specifically for unit-tests?

We need HTTPS here.If implementation SSL support for [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) need a lot codes, maybe it's easier to use a third party library?

lalitb:
@pavanshahm - Thanks for this PR. It would be good to have SSL certificate check integrated. A couple of comments here:
 1. Can we avoid adding code to build curl with SSL support as part of the otel-cpp build? I am not a bazel expert, but with cmake, we can build curl + SSL through vcpkg toolchain.
 2. instead of adding a certificate path, can we leverage the EventHandler class to pass the certificate:
 https://github.com/open-telemetry/opentelemetry-cpp/blob/dc25ea686707dee2ea993f49664abf2b88e6f58a/ext/include/opentelemetry/ext/http/client/http_client.h#L205

lalitb:
> We need HTTPS here.If implementation SSL support for [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) need a lot codes, maybe it's easier to use a third party library?

OK. I thought we don't really need to test the ssl-handshake as part of the unit test here, as it should work correctly with `curl` as long as the correct certificate is passed.  We can locally add unit-test tests to ensure that the SSL certificate is set correctly through our HTTP client library. But we can discuss if you feel it is really required?

owent:
> > We need HTTPS here.If implementation SSL support for [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) need a lot codes, maybe it's easier to use a third party library?
> 
> OK. I thought we don't really need to test the ssl-handshake as part of the unit test here, as it should work correctly with `curl` as long as the correct certificate is passed. We can locally add unit-test tests to ensure that the SSL certificate is set correctly through our HTTP client library. But we can discuss if you feel it is really required?

I don't think we need to test ssl-handshake  here. But I think we should test whether the new options (`sslCertPath`) works. Is there any better suggestion to approach it?

lalitb:
> > > We need HTTPS here.If implementation SSL support for [http server](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.0.0-rc3/ext/include/opentelemetry/ext/http/server/http_server.h) need a lot codes, maybe it's easier to use a third party library?
> > 
> > 
> > OK. I thought we don't really need to test the ssl-handshake as part of the unit test here, as it should work correctly with `curl` as long as the correct certificate is passed. We can locally add unit-test tests to ensure that the SSL certificate is set correctly through our HTTP client library. But we can discuss if you feel it is really required?
> 
> I don't think we need to test ssl-handshake either. But I think we should test whether the new options (`sslCertPath`) works. Is there any better suggestion to approach it?

ok, I was thinking to avoid testing the HTTPS request path as part of the unit test if it needs us to build and link with a third-party library. On second thought - I had a quick look at `civetweb` usage in prometheus-cpp. Perhaps it can be a good replacement for internal http_server for both unit tests ( for http/https requests) and zpages.

lalitb:
@pavanshahm  - just wondering if you have plans to continue on this PR? As this has been at WIP state for some time now.

lalitb:
> @pavanshahm - just wondering if you have plans to continue on this PR? As this has been at WIP state for some time now.

@pavanshahm - Checking one more time if you have plans to work on this PR. While it would be really good to get it merged, we also don't want to keep it stale for long.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
```diff
@@ -68,6 +68,9 @@ struct OtlpHttpExporterOptions
 
   // TODO: Enable/disable to verify SSL certificate
   std::chrono::milliseconds timeout = std::chrono::milliseconds(30000);
+
+  // filepath of the sll certifcate, this is empty if there is none. 
+  std::string sslCertPath;
```
sirzooro, line 73: mTLS support (requested by someone else here in another issue) will require two more fields for paths to private and public key. To avoid confusion please rename this field to make it more clear that this is path to trusted root CA certificates, e.g. `sslCaCertPath`.
In my issue I also asked about way to specify minimum TLS version (`CURLOPT_SSLVERSION`) and allowed ciphers (`CURLOPT_SSL_CIPHER_LIST`). Below I also mentioned another one to disable certificate validation. All of them would need corresponding fields here.
BTW, you also have typo in comment above - sll -> ssl

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -58,12 +58,20 @@ class Request : public http_client::Request
     timeout_ms_ = timeout_ms;
   }
 
+  void IncludePermissionsFilePath(std::string filepath) noexcept override
```
sirzooro, line 61: Other methods here uses `nostd::string_view` for parameters instead of `std::string`, please make it consistent.

PR #943, opened 2021-08-10
Not set parentId in case parentId is empty

andriy-fd:
Fixes # (issue)
The root Span is not displayed in SignalFX (Splunk APM) when data is exported via Zipkin exporter.
According to https://zipkin.io/zipkin-api/#/ (ListOfSpans seection) the **parentId** element should be absent for root span

## Changes

added logic into Zipkin's recordable implementation to add parentId element only when parentId is valid.


lalitb (approved):
thanks for the fix. You would have to authorize easyCLA before it can be merged. 

Also to format code before commit, run below from root directory:
$ ./ci/do_ci.sh format

PR #966, opened 2021-09-01
 make span context management api public

### LINKED ISSUE #950: Helper methods for accessing Span directly from Context in Span API
The 1.0.0 specification states:

```
This section defines all operations within the Tracing API that interact with the Context.

The API MUST provide the following functionality to interact with a Context instance:

- Extract the Span from a Context instance
- Insert the Span to a Context instance
The functionality listed above is necessary because API users SHOULD NOT have access to the Context Key used by the Tracing API implementation.

If the language has support for implicitly propagated Context (see here), the API SHOULD also provide the following functionality:

- Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the Span from the context.
- Set the currently active span to the implicit context. This is equivalent to getting the implicit context, then inserting the Span to the context.
```

I was looking for documentation or example (e.g. context propagation) where this is done.  In some languages, they provide helper methods directly on Span which will extract/insert into context (both explicitly or implicitly).  

Specifically, I think this may need to be public: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/propagation/detail/context.h#L14

It looks like these methods may be on Tracer: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/tracer.h#L146.  This is inconsistent with other APIs.

Ideally the propogation:detail + tracer methods would get reconciled into a standard header with Span <-> Context interaction.

lalitb:
Fixes #950 

## Changes

Move span-context helper methods from  `trace::propagation` to `trace` namespace. They are now treated as being in public api namespace. It's not possible to move them to Span API ( i.e, making them static within api::Span class ) as the complete implementation of  `DefaultSpan` is not visible/available in that class.

This is how header structure looks like:

context/context.h - contains implementation for context data-structure logic
trace/tracer.h - span context management ( implicit )
trace/context.h - span context management ( explicit )

The implicit context management are still static methods in `trace::tracer` class to ensure not to introduce a breaking change, and at the same time remain specs compliant.

cc @jsuereth 


api/test/trace/span_benchmark.cc
```diff
@@ -112,7 +112,7 @@ void BM_SpanCreationWitContextPropagation(benchmark::State &state)
     auto outer_span_context = SpanContext(trace_id, span_id, trace_api::TraceFlags(), false);
     auto outer_span =
         nostd::shared_ptr<trace_api::Span>(new trace_api::DefaultSpan(outer_span_context));
-    trace_api::propagation::SetSpan(current_ctx, outer_span);
+    opentelemetry::trace::SetSpan(current_ctx, outer_span);
```
ThomsonTan, line 115: Probably just use `trace_api` as it is already used a lot?

lalitb:
closing this PR

PR #969, opened 2021-09-02
Support determining parent span from Context while creating new Span

### LINKED ISSUE #955: Span Parent should be determined from Context (and take this as a parameter.
The [specification api](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/trace/api.md#span-creation) states:

> The parent Context or an indication that the new Span should be a root Span. The API MAY also have an option for implicitly using the current Context as parent as a default behavior. This API MUST NOT accept a Span or SpanContext as parent, only a full Context.
>
> The semantic parent of the Span MUST be determined according to the rules described in [Determining the Parent Span from a Context](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/trace/api.md#determining-the-parent-span-from-a-context).

Here's the current API: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/span.h#L67

lalitb:
Fixes #955 

## Changes
This adds support for passing `Context` wrapping parent `Span` object while creating new Span. 

- In order to not avoid breaking the existing customers code, the legacy api supporting passing `SpanContext` object is also kept in place. Also to avoid storing two different variables for `Context` and `SpanContext` inside [StartSpanOptions](https://github.com/open-telemetry/opentelemetry-cpp/blob/da5c79af98ae7c0e177566362f3fe25a3896d693/api/include/opentelemetry/trace/span.h#L49) struct, variant object is introduced within this struct.
- Slight restructuring in header files was required to avoid cyclic dependency in header files ( i.e `span.h` requiring `context/context.h` and `context/context.h` requiring `span.h` through `context/context_value.h`) : this required adding `span_startoptions.h` and `span_metadata.h`.

cc @jsuereth 


jsuereth (approved):

lalitb:
build failed due to merge conflict with #966 , have resolved them now.

ThomsonTan (approved):

PR #970, opened 2021-09-03
Add note on experimental semantic convention implementation

lalitb:
Fixes 968

## Changes

- Add comment for semantic convention for resource and trace - to indicate it's based on experimental specs, and may change in future.
- Rename experimental headers with `experimenta_*` prefix
- Rename few src files having `.cpp` extension to `.cc` to be consistent with other src files.
- 
cc @jsuereth 


ThomsonTan:
@jsuereth do you think this declaration for semantic convention is sufficient, or we also need to declare the status in the release document?

lalitb:
> @jsuereth do you think this declaration for semantic convention is sufficient, or we also need to declare the status in the release document?

We should be adding it in release description for 1.0.0 as done by other SIGs : 
Eg Java https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.5.0

api/include/opentelemetry/trace/semantic_conventions.h
```diff
@@ -1,6 +1,12 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+// NOTE:
```
jsuereth, line 4: Can you also add a note in the "versioning" document that denotes that headers considered unstable will be documented as such?

lalitb: This is done now.

Versioning.md
```diff
@@ -65,6 +65,12 @@ Refer to the [ABI Policy](./docs/abi-policy.md) for more details. To summarise:
   allowed to break existing stable interfaces. Feature flags will be removed
   once we have a stable implementation for the signal.
 
+* As an exception, small experimental features in otherwise stable signals/components
+  mayn't necessarily be released under feature flag. These would be flagged as experimental
+  by adding a `NOTE` in it's header file - either at the beginning of file, or as the comment for
```
jsuereth, line 70: non-blocking comment -

Would it make sense to call the header file `experimental_*` to further clarify?

I think you could do this and still provide backwards compatibility later by forward-chain including the new non-experimental folder when it stabilizes.   Just a thought.

jsuereth (approved):

Versioning.md
lalitb: Yes this does makes sense. Will incorporate this.

lalitb: This is done now.

PR #973, opened 2021-09-07
remove internally used functionality from public API documentation

### LINKED ISSUE #972: detail namespace should be kept compatible or denoted as internal-only usage
There's a few undocumented "detail" namespaces.  I'm assuming these are meant to be "changeable" header-only implementation components we don't want to maintain 100% compatibility for.

Here's an [example](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/span_context_kv_iterable_view.h)

All such "detail" namespaces should have documentation denoting these are internal-only packages (or we need to commit to not changing any interface exposed here).

Additionally, can also document this [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/Versioning.md#versioning)

lalitb:
Fixes #972 

## Changes

Small change to ensure that all internally used api's shouldn't be part of public API documentation.
As required by #972 all functionality under `::detail` namespace is already excluded from public API doc. This PR adds exclusion for additional missing headers.

cc @jsuereth 


lalitb:
@jsuereth @ThomsonTan  - To make this more explicit
 - Documented use of `detail` namespace both in header, and Versioning doc.
 - Ensure we remove all the internal implementation from public API doc.

jsuereth (approved):

ThomsonTan (approved):

PR #974, opened 2021-09-08
Make Otlp exporter configuration environment variables specs-compliant

### LINKED ISSUE #971: OTLP Environment variables do not match specification
Currently, [OTLP environment variables (OTEL_EXPORTER_OTLP_GRPC_ENDPOINT)](https://github.com/open-telemetry/opentelemetry-cpp/blob/5067e49bce39e8dd4c3986042e558e965c73a253/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter.h#L24) does not [match the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/protocol/exporter.md).

Copied relevant sections:

## Configuration Options

The following configuration options MUST be available to configure the OTLP exporter. Each configuration option MUST be overridable by a signal specific option.

| Configuration Option | Description                                                  | Default           | Env variable                                                 |
| -------------------- | ------------------------------------------------------------ | ----------------- | ------------------------------------------------------------ |
| Endpoint             | Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using `OTEL_EXPORTER_ENDPOINT` with OTLP/HTTP, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. `v1/traces` or `v1/metrics`). The per-signal endpoint configuration options take precedence and can be used to override this behavior. See the [OTLP Specification][otlphttp-req] for more details. | `https://localhost:4317` | `OTEL_EXPORTER_OTLP_ENDPOINT` `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` |
| Certificate File     | Path to certificate file for TLS credentials of gRPC client. Should only be used for a secure connection. | n/a               | `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` `OTEL_EXPORTER_OTLP_METRICS_CERTIFICATE` |
| Headers              | Key-value pairs to be used as headers associated with gRPC or HTTP requests. See [Specifying headers](./exporter.md#specifying-headers-via-environment-variables) for more details.                   | n/a               | `OTEL_EXPORTER_OTLP_HEADERS` `OTEL_EXPORTER_OTLP_TRACES_HEADERS` `OTEL_EXPORTER_OTLP_METRICS_HEADERS` |
| Compression          | Compression key for supported compression types. Supported compression: `gzip`| No value              | `OTEL_EXPORTER_OTLP_COMPRESSION` `OTEL_EXPORTER_OTLP_TRACES_COMPRESSION` `OTEL_EXPORTER_OTLP_METRICS_COMPRESSION` |
| Timeout              | Max waiting time for the backend to process each spans or metrics batch. | 10s               | `OTEL_EXPORTER_OTLP_TIMEOUT` `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT` `OTEL_EXPORTER_OTLP_METRICS_TIMEOUT` |

Supported values for `OTEL_EXPORTER_OTLP_*COMPRESSION` options:

- If the value is missing, then compression is disabled.
- `gzip` is the only specified compression method for now. Other options MAY be supported by language SDKs and should be documented for each particular language.

Example 1

The following configuration sends all signals to the same collector:

```bash
export OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4317
```

Example 2

Traces and metrics are sent to different collectors:

```bash
export OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://collector:4317

export OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=https://collector.example.com/v1/metrics
```

### Specify Protocol

Currently, OTLP has more than one transport protocol it can support, e.g.
`grpc`,  `http/json`, `http/protobuf`.   As of 1.0 of the specification, there
*is no specified default, or configuration via environment variables*.  We
reserve the following environment variables for configuration of protocols in
the future:

- `OTEL_EXPORTER_OTLP_PROTOCOL`
- `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL`
- `OTEL_EXPORTER_OTLP_METRICS_PROTOCOL`

SDKs have an unspecified default, if no configuration is provided.

### Specifying headers via environment variables

The `OTEL_EXPORTER_OTLP_HEADERS`, `OTEL_EXPORTER_OTLP_TRACES_HEADERS`, `OTEL_EXPORTER_OTLP_METRICS_HEADERS` environment variables will contain a list of key value pairs, and these are expected to be represented in a format matching to the [W3C Correlation-Context](https://github.com/w3c/baggage/blob/master/baggage/HTTP_HEADER_FORMAT.md), except that additional semi-colon delimited metadata is not supported, i.e.: key1=value1,key2=value2. All attribute values MUST be considered strings.

lalitb:
Fixes #971 

## Changes

This PR makes otlp exporter configuration environment variables specs compliant.
Also add one new env variable ( not part of specs ) -  `OTEL_EXPORTER_OTLP_CERTIFICATE_STRING`  to specify certificate as string. 

Also introduce new env-variables ( not part of specs ) `


exporters/otlp/README.md
```diff
@@ -42,11 +42,27 @@ options.url = "localhost:12345";
...
-| ------------ |----------------- |
-| `endpoint`   | `localhost:4317` |
+| Option       | Env Variable |Default          |   Description  |
+| ------------ |---------------|------------ |----------------|
+| `endpoint`   | `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT`    |`localhost:4317` | The OTLP GRPC endpoint to connect to |
+| `use_ssl_credentials` | `OTEL_EXPORTER_OTLP_GRPC_SSL_ENABLE`  | `false`  |  Whether the endpoint is SSL enabled |
+| `ssl_credentials_cacert_path`  |  n/a | `""` |   SSL Certificate file path |
+| `ssl_credentials_cacert_as_string` | `OTEL_EXPORTER_OTLP_GRPC_SSL_CERTIFICATE` | `""`  |   SSL Certifcate as in-memory string |
+
+Note: These env variables names are not same as in [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/protocol/exporter.md).
```
ThomsonTan, line 54:
```suggestion
Note: These environment variable names are not same as in [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/protocol/exporter.md).
```

exporters/otlp/README.md
```diff
@@ -42,11 +42,27 @@ options.url = "localhost:12345";
...
+
+### Configuration options ( OTLP HTTP Exporter )
+
+| Option       | Env Variable |Default          | Description |
+| ------------ |-----|------------ |------|
+| `url`   | n/a    | `http://localhost:4317/v1/traces`  | The OTLP HTTP endpoint to connect to |
+| `content_type` | n/a  | `application/json`  |   Data format used - JSON or Binary |
+| `json_bytes_mapping`  |  n/a | `JsonBytesMappingKind::kHexId` | Encoding used for trace_id and span_id |
+| `use_json_name` | n/a | `false`  | Whether to use json name of protobuf field to set the key of json |
+| `timeout`  | n/a  | `3000 ms` | http timeout |
```
ThomsonTan, line 65:
```suggestion
| `timeout`  | n/a  | `30000 ms` | http timeout |
```
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h#L70

lalitb: good catch :)

jsuereth (changes requested):
Commented on the bug as well.

I don't think this is sufficient for a 1.0 against the specification.

I think there are two enviornment variables here that you can leave for legacy users BUT you should provide the new environment variables as alternatives, specifically:

`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE`

lalitb:
@jsuereth  - It was creating more noise in code to support both legacy and compliant environment variables, and would further increase in future once we start supporting `OTEL_EXPORTER_OTLP_[TRACES_]PROTOCOL`. Keeping that in mind, have removed legacy environment variables. This would be breaking change ( hopefully won't affect lots of users ), and we will document this in release description.

exporters/otlp/src/otlp_grpc_exporter.cc
```diff
@@ -59,6 +59,7 @@ std::unique_ptr<proto::collector::trace::v1::TraceService::Stub> MakeServiceStub
     const OtlpGrpcExporterOptions &options)
 {
   std::shared_ptr<grpc::Channel> channel;
+
```
lalitb, line 62: will remove this file later.

lalitb:
@owent - As this env variables were initially modified by you, would like to have a look into this PR. Thanks.

owent:
> @owent - As these env variables were initially modified by you, would like to have a look into this PR. Thanks.

LGTM.

Should we also make the same changes for `OtlpHttpExporter` ?

lalitb:
> LGTM.
> 
> Should we also make the same changes for `OtlpHttpExporter` ?                     

Thanks for reviewing @owais. `OtlpHttpExporter` doesn't support environment variables yet, so we are not non-compliant there :) Will create a github issue for adding env variables there.

owent:
> > LGTM.
> > Should we also make the same changes for `OtlpHttpExporter` ?
> 
> Thanks for reviewing @owais. `OtlpHttpExporter` doesn't support environment variables yet, so we are not non-compliant there :) Will create a github issue for adding env variables there.

OK. I think I can add support of these environment variables for `OtlpHttpExporter` then:

+ `OTEL_EXPORTER_OTLP_ENDPOINT`
+ `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+ `OTEL_EXPORTER_OTLP_HEADERS`
+ `OTEL_EXPORTER_OTLP_TRACES_HEADERS`

But `OTEL_EXPORTER_OTLP_CERTIFICATE` and `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` can be available only after #938 is finished.

lalitb:
> OK. I think I can add support of these environment variables for `OtlpHttpExporter` then:
> 
> * `OTEL_EXPORTER_OTLP_ENDPOINT`
> * `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
> * `OTEL_EXPORTER_OTLP_HEADERS`
> * `OTEL_EXPORTER_OTLP_TRACES_HEADERS`
> * `OTEL_EXPORTER_OTLP_TIMEOUT`
> * `OTEL_EXPORTER_OTLP_TRACES_TIMEOUT`

That would be awesome :)

> But OTEL_EXPORTER_OTLP_CERTIFICATE and OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE can be available only after #938 is finished.

Agree.

jsuereth (approved):

PR #1005, opened 2021-09-30
Adding new Logger:log() method

lalitb:
## Changes

This is to add the `Logger: Log()` method for the below use-case:

```cpp
  Properties attribs = {{"attrib1", 1}, {"attrib2", 2}};
  logger.log(severity, name, attribs);
```
where `Properties` class implements `common::KeyValueIterable`.

Similar method is supported by `Traccer::StartSpan()`:
https://github.com/open-telemetry/opentelemetry-cpp/blob/21a441e2be80ca335a7d6abd033af2f1cdf477ab/api/include/opentelemetry/trace/tracer.h#L60-L65

Also, a minor change to use full `tracer` namespace in logger sdk, in order to avoid including unnecessary tracer headers for logger instrumentation.

### DIFF
--- a/api/include/opentelemetry/logs/logger.h
+++ b/api/include/opentelemetry/logs/logger.h
@@ -143,7 +143,7 @@ public:
    */
   template <class T,
             nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr>
-  inline void Log(Severity severity, const T &attributes) noexcept
+  void Log(Severity severity, const T &attributes) noexcept
   {
     this->Log(severity, "", "", std::map<std::string, std::string>{}, attributes, {}, {}, {},
               std::chrono::system_clock::now());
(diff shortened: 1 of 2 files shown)


api/include/opentelemetry/logs/logger.h
```diff
@@ -189,6 +189,23 @@ class Logger
...
   }
 
+  /**
+   * Writes a log.
+   * @param severity The severity of the log
+   * @param name The name of the log
+   * @param attributes the attributes, stored as a 2D list of key/value pairs, that are associated
+   * with the log event
+   */
+  void Log(Severity severity,
```
reyang, line 199: `inline`?

reyang (approved):

api/include/opentelemetry/logs/logger.h
```diff
@@ -189,6 +189,23 @@ class Logger
     this->Log(severity, name, "", {}, attributes, {}, {}, {}, std::chrono::system_clock::now());
   }
 
+  /**
+   * Writes a log.
+   * @param severity The severity of the log
+   * @param name The name of the log
+   * @param attributes the attributes, stored as a 2D list of key/value pairs, that are associated
```
ThomsonTan, line 196:
```suggestion
   * @param attributes The attributes, stored as a 2D list of key/value pairs, that are associated
```

ThomsonTan (approved):

api/include/opentelemetry/logs/logger.h
lalitb: member function inside class body is `inline` by default. I now noticed that couple of existing member functions have explicit `inline` specifier. Will remove it from there to avoid any confusion.

PR #1020, opened 2021-10-19
cmake: add Find<package>.cmake find module method for thrift

ideepika:
In some cases, when using jaeger-cpp-client, we use thrift provided from
distro packages, these does not contain thriftConfig.cmake.
For such scenarios, we pivot to do not make find_package CONFIG mode as
default, instead we use basic signature[0]. We make use of
`CMAKE_MODULE_PATH` to search for FindThrift.cmake, which will find and
set all thrift::thrift's target properties.

[0] https://cmake.org/cmake/help/latest/command/find_package.html
The command arguments determine which of the above modes is used. When
the basic signature is used, the command searches in Module mode first.
If the package is not found, the search falls back to Config mode.

Signed-off-by: Deepika Upadhyay <dupadhya@redhat.com>

Fixes # (issue)

## Changes


ideepika:
> Out of curiosity, which are these distributions providing thrift package without cmake config files?

Sure, I can help there, I think most does, you can check them here: https://pkgs.org/search/?q=thrift
for eg (Ubuntu 20.04/Centos/Fedora): https://ubuntu.pkgs.org/20.04/ubuntu-universe-arm64/libthrift-dev_0.13.0-2build2_arm64.deb.html

PR #1041, opened 2021-10-29
fix: update curl lib build_file

miiila:
When trying to import otlp as a dependency through Bazel, build_file can't be found in current repository. This commit standardize it to @io_opentelemetry_cpp as it is in other deps.

Fixes # (issue)

## Changes

Updates path to curl build file


lalitb:
#1046 fixes this.

PR #1053, opened 2021-11-04
Move old metrics implementation to different directory, and rename targets to *_deprecated

lalitb:
## Changes

As the effort required to update our existing metrics SDK implementation as per new specs is huge, it's easier to re-write it from the ground-up ( #1033 ). This means we could
  - Either remove the old implementation, which breaks any customers who would be using it, and leave them with no alternative till the new implementation is ready.
  - Or let both old and new implementation exists together for some time till new implementation is ready. The old implementation exists under the feature flag WITH_METRICS_PREVIEW. And the new implementation would be active/build only if the feature flag is disabled.

We discussed both approaches in the community meeting and agreed on second.  This PR makes necessary changes for it, basically
 - Move the old implementation to `*/_metrics/*` directory, and new development continue under `*/metrics/*`.
 - Rename cmake and bazel target for old implementation to `opentelemetry_metrics_deprecated`, `prometheus_exporter_deprecated`

ps. -  We can also use this PR on discussing both approaches if required. The first approach does provide cleaner repo and easy maintainability. 


reyang (approved):

api/include/opentelemetry/_metrics/async_instruments.h
```diff
@@ -4,8 +4,8 @@
 #pragma once
 #ifdef ENABLE_METRICS_PREVIEW
 
-#  include "instrument.h"
-#  include "observer_result.h"
+#  include "opentelemetry/_metrics/instrument.h"
```
ThomsonTan, line 7: Could we add implication that the folder `_metrics` is deprecated, like name it as `metrics_deprecated`?

ThomsonTan (approved):

api/include/opentelemetry/_metrics/async_instruments.h
lalitb: That's a good point. I started with `metrics_deprecated` and it looked weirdly long to me including as `#include` in headers and srcs. I borrowed `_metrics` directory notation from python SIG the way they are handling it. 
To make it more clear, have updated the status in README.md along with specifying this directory name.

Issue #1051, opened 2021-11-04
Building OTLP exporter from the release tarball fails for opentelemetry-proto dependency 


lalitb:
**Describe your environment** 
ubuntu 20.04

**Steps to reproduce**
Download the release tarball, and try building OTLP exporter with cmake option WITH_OTLP=ON
The problem is that the release tarball contains the `third_party/opentelemetry-proto` directory but it is empty as it is a git submodule and the CI/build is not including this as part of the tarball).

**What is the expected behavior?**
Build to be successful.

**What is the actual behavior?**
Build fails

**Additional context**
Few ways I can think to fix it:

1 Let cmake download the opentelemetry-proto dependency at build time using (say) `FetchContent` if the `third_party/opentelemetry-proto` directory is empty. `FetchContent` is introduced in cmake 3.14, so we may have to upgrade the cmake version for this.
2. Pre-compile proto source(cpp) and headers (hpp) using protocol buffer compiler and check-in them in the repo ( similar to how opentelemetry-python does it - https://github.com/open-telemetry/opentelemetry-python/tree/main/opentelemetry-proto ). We can also have a script to generate these files, to be used every time the version needs to be bumped.

lalitb:
@esigo  - fyi, have assigned it to you as discussed in today's meeting.

esigo:
@lalitb seems I can't assign it to myself. I will see if I can come up with a solution that doesn't require cmake version change.

esigo:
I will use `ExternalProject_Add` as it is available in cmake 3.1.

esigo:
There is a similar issue with `nlohmann_json`. I try to fix both.

PR #1055, opened 2021-11-06
checkout opentelemetry-proto v0.11.0

esigo:
Related to #1051 (issue)

## Changes

checkout opentelemetry-proto v0.11.0 and store the third_party release tags in a file

### DIFF
--- a/third_party_release
+++ b/third_party_release
@@ -1,0 +1,7 @@
+benchmark=v1.5.3
+googletest=release-1.8.0-2523-ga6dfd3ac
+ms-gsl=v1.0.0-393-g6f45293
+nlohmann-json=v3.9.1
+opentelemetry-proto=v0.11.0
+prometheus-cpp=v0.11.0
+vcpkg=2020.04-2702-g5568f110b
--- a/third_party_release
+++ b/third_party_release
@@ -1,3 +1,6 @@
+gRPC=v1.39.1
+thrift=0.14.1
+abseil=20210324.0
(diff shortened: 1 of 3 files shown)


third_party_release
```diff
@@ -0,0 +1,7 @@
+benchmark=v1.5.3
+googletest=release-1.8.0-2523-ga6dfd3ac
+ms-gsl=v1.0.0-393-g6f45293
+nlohmann-json=v3.9.1
+opentelemetry-proto=v0.11.0
+prometheus-cpp=v0.11.0
+vcpkg=2020.04-2702-g5568f110b
```
lalitb, line 7: Sorry if I am missing something. What is the purpose of adding this file, and also adding an option in `do_ci.sh` to create this file? Is it meant to be a central place to list versions of all the otel-cpp dependencies?

esigo: I believe it's a good idea to have the versions documented somewhere so the users can download the versions of the dependencies that a release has been published with. This would be obsolete if the plan is to not support older versions though.

lalitb: Ok, that makes sense. Just wondering if this should also contain the dependencies not coming from submodules, eg gRPC(v1.39.1), thrift (0.14.1), abseil(20210324.0)?

esigo: yes absolutely. Just added them to the script.

lalitb (approved):

ThomsonTan (approved):

PR #1056, opened 2021-11-06
Building otlp exporter from the release tarball

### LINKED ISSUE #1051: Building OTLP exporter from the release tarball fails for opentelemetry-proto dependency

esigo:
Fixes #1051 (issue)

## Changes

download opentelemtery-proto if it's not stored.
Read opentelemtery-proto tag from file if not available use main.


CMakeLists.txt
```diff
@@ -20,6 +20,16 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+if(EXISTS "${CMAKE_SOURCE_DIR}/third_party_release")
+  file(STRINGS "${CMAKE_SOURCE_DIR}/third_party_release" third_party_tags)
+  foreach(third_party ${third_party_tags})
+    string(REGEX REPLACE "^[ ]+" "" third_party ${third_party})
+    string(REGEX MATCH "^[^=]+" third_party_name ${third_party})
+    string(REPLACE "${third_party_name}=" "" third_party_tag ${third_party})
+    set(${third_party_name} "${third_party_tag}")
+  endforeach()
+endif()
+
```
lalitb, line 32: Trying to understand the logic here. We iterate the file and extract the name and version from each entry. In the end,  `third_party_name` and `third_party_tag` contain name and version for last entry of file ?

cmake/opentelemetry-proto.cmake
```diff
@@ -1,6 +1,36 @@
-include(${PROJECT_SOURCE_DIR}/cmake/proto-options-patch.cmake)
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opentelemetry-proto/.git)
+    set(PROTO_PATH "${CMAKE_CURRENT_SOURCE_DIR}/third_party/opentelemetry-proto")
+    set(needs_proto_download FALSE)
+else()
+    if("${opentelemetry-proto}" STREQUAL "")
```
lalitb, line 5: how would this be defined to point to a specific version/branch?

CMakeLists.txt
esigo: No we are storing all the versions not just the last one. e.g. `${nlohmann-json}` has `v3.9.1`, or `${opentelemetry-proto} `has, `v0.11.0`.
In this PR I use ${opentelemetry-proto} to download it from github.
We can use the same technique for all the other dependencies when they are not available in the `third_party` folder.

lalitb: Ah got it, I missed the `set` command.

lalitb (approved):

ThomsonTan (approved):

CMakeLists.txt
```diff
@@ -20,6 +20,16 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules/")
 endif()
 
+if(EXISTS "${CMAKE_SOURCE_DIR}/third_party_release")
```
ThomsonTan, line 23: How does the file `${CMAKE_SOURCE_DIR}/third_party_release` get created?

ThomsonTan: Never mind, just saw it was created in #1055

PR #1060, opened 2021-11-11
Mark cURL as required dependence for building Zipkin exporter

ThomsonTan:
## Changes

`http_client_curl` is required to build Zipkin exporter. It is better to report error when running CMake files if cURL is not found for Zipkin, or it will be surfaced as link error during later linking stage.

### DIFF
--- a/exporters/zipkin/CMakeLists.txt
+++ b/exporters/zipkin/CMakeLists.txt
@@ -13,12 +13,7 @@
 # limitations under the License.
 
 include_directories(include)
-find_package(CURL)
-if(NOT CURL_FOUND)
-  message(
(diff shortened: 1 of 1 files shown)


lalitb (approved):

exporters/zipkin/CMakeLists.txt
```diff
@@ -14,9 +14,12 @@
 
 include_directories(include)
 find_package(CURL)
-if(CURL_FOUND)
-  add_definitions(-DWITH_CURL)
+if(NOT CURL_FOUND)
```
lalitb, line 17: how about using `find_package(CURL REQUIRED)`?

ThomsonTan: Yes, this looks better.

PR #1068, opened 2021-11-16
Improvement: span_id should not break strict aliasing.

### LINKED ISSUE #1067: API break strict aliasing.
It's easy to understand so didn't follow the template.

```c++
// api/include/opentelemetry/trace/span_id.h
  bool IsValid() const noexcept
  {
    static_assert(kSize == 8, "update is needed if kSize is not 8");
    return *reinterpret_cast<const uint64_t *>(&rep_) != 0ull;
  }
```

would generate this compiler warning:

```text
warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
```

`-O2` optimizing would enable `-fstrict-aliasing`. API header file should not make such a breaking.

hcoona:
Fixes #1067

### DIFF
--- a/api/include/opentelemetry/trace/span_id.h
+++ b/api/include/opentelemetry/trace/span_id.h
@@ -60,8 +60,8 @@ public:
   // Returns false if the SpanId is all zeros.
   bool IsValid() const noexcept
   {
-    static_assert(kSize == 8, "update is needed if kSize is not 8");
-    return *reinterpret_cast<const uint64_t *>(&rep_) != 0ull;
+    static constexpr uint8_t kEmptyRep[kSize] = {0};
+    return memcmp(rep_, kEmptyRep, kSize) != 0;
   }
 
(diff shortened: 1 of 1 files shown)


lalitb (approved):

api/include/opentelemetry/trace/span_id.h
```diff
@@ -61,7 +61,8 @@ class SpanId final
   bool IsValid() const noexcept
   {
     static_assert(kSize == 8, "update is needed if kSize is not 8");
```
ThomsonTan, line 63: This `static_assert` could be removed then as the code doesn't make such assumption anymore?

ThomsonTan (approved):

api/include/opentelemetry/trace/span_id.h
hcoona: fixed

PR #1074, opened 2021-11-17
build release tarball when nlohmann-json not installed

### LINKED ISSUE #981: Building from the release tarball fails if nlohmann-json not installed on the system
**Describe your environment** 
```
Ubuntu 20.04
Linux pace-desktop 5.11.0-34-generic #36~20.04.1-Ubuntu SMP Fri Aug 27 08:06:32 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
nlohmann-json3-dev package is NOT installed system wide
```

**Steps to reproduce**
 * In reality I am trying to add nlohmann-json as a dependency into my own cmake module using ExternalProject and listing the release tarball as the desired version.  The below steps however, are simpler, and just pretend I wanted to download and install the release into my system.
 * Download release tarball
 * Extract release tarball
 * cd <extracted_dir>
 * mkdir build
 * cd build
 * cmake ..

Ideally, cmake would download nlohmann-json using ExternalProject or some other mechanism (or at least have a flag to enable this).  Or, nlohmann-json's headers would be included as part of the release tarball.  However, that does not seem to be implemented yet, which is fine.

Given this, I would expect to see the standard CMake find_package failure message:

```
CMake Error at CMakeLists.txt:245 (find_package):

    nlohmann_jsonConfig.cmake
    nlohmann_json-config.cmake

  has been installed.
```

What did you see instead?

```
CMake Error at CMakeLists.txt:256 (add_subdirectory):
  The source directory

    /home/pace/Downloads/opentelemetry-cpp-1.0.0-rc4/third_party/nlohmann-json

  does not contain a CMakeLists.txt file.
```

**Additional context**

The problem is that the release tarball contains the `nlohmann-json` directory but it is empty (likely because it is a git submodule and the CI/build is not including this as part of the tarball).

If I delete that directory then I still don't get the error message I would expect.  Instead I get:

```
  nlohmann_json package was not found.  Please either provide it manually or
  clone with submodules.  To initialize, fetch and checkout any nested
  submodules, you can use the following command:

  git submodule update --init --recursive
```

This is a little better.  I imagine it is a helpful message added for new developers.  However, since I am downloading the release tarball, it is more or less meaningless (I didn't clone any git repository, there is no action I can take here).  If I mark the package as required then I get the error message I would expect (at which point I clearly realize that I need to provide a system install of the nlohmann-json module and I can resolve that).

This might be opinionated, and I really do appreciate the work that you have done here, but it is rather odd behavior from a CMake module and any deviation from the CMake norm seems to send me down an unfortunate trail of debugging.

esigo:
Fixes #981 (issue)

## Changes

clones nlohmann-json if not present in third_party folder.


CMakeLists.txt
```diff
@@ -294,15 +295,14 @@ if((NOT WITH_API_ONLY)
...
+        set(nlohmann_json_clone TRUE)
+        include(cmake/nlohmann-json.cmake)
+        message("\nnlohmann_json package was not found. Cloning from github")
       endif()
     else()
-      # Let's fail with default find_package error
-      find_package(nlohmann_json REQUIRED)
+      set(nlohmann_json_clone TRUE)
+      include(cmake/nlohmann-json.cmake)
+      message("\nnlohmann_json package was not found. Cloning from github")
```
lalitb, line 305: Just wondering if we can bit simplify from L281 - L308 as ( not tested ) :

```
set(nlohmann_json_clone FALSE)
if (nlohmann_json_FOUND)
  message("Using external nlohmann::json")
elseif( EXISTS ${PROJECT_SOURCE_DIR}/.git AND EXISTS ${PROJECT_SOURCE_DIR}/third_party/nlohmann-json/CMakeLists.txt)
       message("Trying to use local nlohmann::json from submodule")
	...
        include_directories(
          ${PROJECT_SOURCE_DIR}/third_party/nlohmann-json/single_include)
else()
      set(nlohmann_json_clone TRUE)
      include(cmake/nlohmann-json.cmake)
      message("\nnlohmann_json package was not found. Cloning from github")
endif()
```

ext/test/w3c_tracecontext_test/CMakeLists.txt
```diff
@@ -10,4 +10,8 @@ else()
     PRIVATE ${CMAKE_THREAD_LIBS_INIT} opentelemetry_trace http_client_curl
             opentelemetry_exporter_ostream_span ${CURL_LIBRARIES}
             nlohmann_json::nlohmann_json)
+  if(lohmann_json_clone)
```
lalitb, line 13: should it be `nlohmann_json_clone` ? I think CI uses external nlohmann_json package, so it won't get caught there :)

exporters/otlp/CMakeLists.txt
```diff
@@ -48,6 +48,11 @@ if(WITH_OTLP_HTTP)
     opentelemetry_exporter_otlp_http_client
     PUBLIC opentelemetry_sdk opentelemetry_proto http_client_curl
            nlohmann_json::nlohmann_json)
+  if(nlohmann_json_clone)
+    add_dependencies(opentelemetry_exporter_otlp_http_client
+                     nlohmann_json::nlohmann_json)
+    list(APPEND OPENTELEMETRY_OTLP_TARGETS nlohmann_json_)
```
lalitb, line 54: Why is this step required?

esigo: it's needed for install. `nlohmann_json::nlohmann_json` is an alias for `nlohmann_json_` here.

CMakeLists.txt
esigo: thanks. done

ext/test/w3c_tracecontext_test/CMakeLists.txt
esigo: thanks. fixed

exporters/otlp/CMakeLists.txt
lalitb: ok, thanks. Should that be an install target for Zipkin and ETW too, or else just have it in `cmake/nlohmann-json.cmake` ?

lalitb (approved):

ThomsonTan (approved):

exporters/otlp/CMakeLists.txt
esigo: good idea. done

PR #1075, opened 2021-11-18
Mark tags to bazel targets

### LINKED ISSUE #1072: Mark tags to bazel targets

The project could build with GCC4.8 + CMake but failed to build with GCC4.8+Bazel. I cannot understand it although I know the higher version of GRPC could not compile with GCC4.8.

I think most of the core components rely on nothing than some extremely basic libraries such as abseil, gsl, gtest & google benchmark. We at least should support bazel build for these targets.

I'd like to mark the bazel targets with tags, so we can easily to exclude some components. For example, excluding OTLP would significantly simplify dependencies.

** Example **

```python
cc_test(
    name = "otlp_recordable_test",
    srcs = ["test/otlp_recordable_test.cc"],
    deps = [
        ":otlp_recordable",
        "@com_google_googletest//:gtest_main",
    ],
    tags = ["otlp"],
)
```

```bash
bazel build //... --build_tag_filters=-otlp
```

esigo:
Fixes #1072 (issue)

## Changes
mark tags to bazel targets.

can exclude targets e.g.
`bazel build //... --build_tag_filters=-otlp,-es,-etw,-prometheus,-zipkin`

### DIFF
--- a/bazel/otel_cc_benchmark.bzl
+++ b/bazel/otel_cc_benchmark.bzl
@@ -1,4 +1,4 @@
-def otel_cc_benchmark(name, srcs, deps):
+def otel_cc_benchmark(name, srcs, deps, tags = [""]):
     """
     Creates targets for the benchmark and related targets.
 
@@ -23,7 +23,7 @@ def otel_cc_benchmark(name, srcs, deps):
         name = name,
         srcs = srcs,
         deps = deps + ["@com_github_google_benchmark//:benchmark"],
-        tags = ["manual"],
+        tags = tags + ["manual"],
     )
 
     # The result of running the benchmark, captured into a text file.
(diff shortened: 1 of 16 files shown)


lalitb:
Thanks for adding it. Can we also add separate tags for ( fine to add it separately from this PR if this is a bigger task )?
 - API ( this would be helpful for the instrumentation-only build, similar to what we do through cmake [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/9e8be676cbf27645a3c534f99b7f8ee167daa5a3/CMakeLists.txt#L206)
 - Separate tags for otlp_grpc and otlp_http ?
 - Tags for tests and/or examples, to provide way to filter building tests and examples.

esigo:
> Thanks for adding it. Just thinking if we can we also add separate tags for:
> 
> * API ( this would be helpful for the instrumentation-only build, similar to what we do through cmake [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/9e8be676cbf27645a3c534f99b7f8ee167daa5a3/CMakeLists.txt#L206) )
> * Separate tags for otlp_grpc and otlp_http ?
> * Tags for tests and/or examples, to provide way to filter building tests and examples.
> 
> It's fine to add it separate from this PR if this is a bigger task ( or raise a github issue for someone to pick it up ).

@lalitb thanks for the feedback.
Yeah sounds good. Let's merge this. I'll raise a separate PR to cover the rest.

lalitb (approved):

ThomsonTan (approved):

PR #1097, opened 2021-11-24
Remove missing target in legacy prometheus

### LINKED ISSUE #1096: Can not build with cmake when enable prometheus exporter
**Describe your environment**

Linux

**Steps to reproduce**

Enable legacy prometheus exporter and prebuilt prometheus-cpp.

Build success

```bash
Building with standard library types...
CMake Error at exporters/prometheus/CMakeLists.txt:34 (install):
  install TARGETS given target "pull" which does not exist.


-- Configuring incomplete, errors occurred!
```

owent:
Signed-off-by: owentou <owentou@tencent.com>

Fixes #1096 

## Changes

Remove missing target and fix document for exported config file.

### DIFF
--- a/cmake/opentelemetry-cpp-config.cmake.in
+++ b/cmake/opentelemetry-cpp-config.cmake.in
@@ -20,30 +20,30 @@
 #   OPENTELEMETRY_VERSION           - Version of opentelemetry-cpp.
 #
 # ::
-#   opentelemetry-cpp::api                        - Imported target of opentelemetry-cpp::api
-#   opentelemetry-cpp::sdk                        - Imported target of opentelemetry-cpp::sdk
-#   opentelemetry-cpp::ext                        - Imported target of opentelemetry-cpp::ext
-#   opentelemetry-cpp::version                    - Imported target of opentelemetry-cpp::version
-#   opentelemetry-cpp::common                     - Imported target of opentelemetry-cpp::common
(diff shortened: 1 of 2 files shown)


lalitb (approved):
Thanks for fixing this, looks like it got introduced while deprecating the old matrix implementation. Not related to this PR, but I feel it would be good to add the `cmake install` test too in the CI pipeline to catch these errors.

owent:
> Thanks for fixing this, looks like it got introduced while deprecating the old matrix implementation. Not related to this PR, but I feel it would be good to add the `cmake install` test too in the CI pipeline to catch these errors.

Need use prebuilt prometheus-cpp to catch this problem. It maybe will take a long time to build all dependencies.
In https://github.com/atframework/cmake-toolset/actions/runs/1499413005, it takes about 50 minutes to finish the job, most package in it is the dependencies of opentelemetry-cpp. So I wonder if we should do the same thing here, or should we  cache the prebuilt files when the versions of dependencies not changed?

lalitb:
> Need use prebuilt prometheus-cpp to catch this problem.

I am not completely sure if that would have helped here. As I understand, the missing target error would have got caught during the otel-cpp installation, which we don't do in our CI pipeline. 
But to speed up the CI pipeline, I feel it would be good to use some form of caching ( either using pre-built binaries of external dependencies as you mentioned or using cmake cache).
While I would merge this PR, we can continue discussing in this thread for any optimization we can do to speed-up pipeline.

PR #1099, opened 2021-11-25
fix data race in BM_ThreadYieldSpinLockThrashing

### LINKED ISSUE #1098: data race in BM_ThreadYieldSpinLockThrashing
**Describe your environment** Describe any aspect of your environment relevant to the problem,
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

**Steps to reproduce**
`bazel test --copt=-DENABLE_METRICS_PREVIEW --copt=-DENABLE_LOGS_PREVIEW --test_output=errors --cache_test_results=no --config=tsan //api/test/common:spinlock_benchmark_smoketest`
no race
What did you expect to see?
no race
What did you see instead?
a data race
**Additional context**

esigo:
Fixes #1098 (issue)

## Changes

The lock function, should return only and only if it actually acquired the lock. This was not the case if the locking was not successful after a limit amount of tries.
see [here](https://github.com/open-telemetry/opentelemetry-cpp/issues/1098#issuecomment-982804027) for more info the issue.


ThomsonTan:
@esigo thanks for looking into it and the fix. Could you please share a little information on how the race condition happens in the original implementation?

api/test/common/spinlock_benchmark.cc
```diff
@@ -103,29 +103,17 @@ static void BM_ProcYieldSpinLockThrashing(benchmark::State &s)
...
-          if (i % 32 == 0)
-          {
-            std::this_thread::yield();
-          }
+          ;
         }
         std::this_thread::yield();
       },
-      [](std::atomic<bool> &l) { l.store(false, std::memory_order_release); });
+      [](std::atomic_flag &l) { l.clear(std::memory_order_release); });
```
lalitb, line 116: Thanks for the fix. would you help understand this change (probably enrich PR description) how it is fixing the race condition.

esigo: I updated the PR description. Please let me know if it needs more information is needed.

lalitb: Thanks for the update. Should we either
 - Change the test comment ( // SpinLock thrashing with thread::yield() after N spins ) as there is no more yield happening.
 - Add the yield after every few tries in the tight loop to be in sync with the benchmark comment.

lalitb (approved):

ThomsonTan (approved):

ThomsonTan:
Thanks @esigo  for the explanation.

PR #1106, opened 2021-11-30
Jaeger exporter: extend supported attributes types

### LINKED ISSUE #1104: Jaeger exporter: attributes of unsupported types are silently forgotten

Actually, it does not matter, it's a consistent bug.

- Windows 10, MSVC 2022, c++17
- Ubuntu 21.10, gcc 11.2, c++17
- CMake 3.21.3
- opentelemetry-cpp 1.1.0
- thrift 0.15.0

**Steps to reproduce**

Add attribute of `int32_t type to a span. e.g.:

- in examples/common/foo_library/foo_library.cc change function f1() as follows:
```
void f1()
{
  auto span = get_tracer()->StartSpan("f1");
  auto scoped_span = trace::Scope(span);
  int32_t myValue = 42;
  span->AddAttribute("some.attribute", myValue);
}
```
- have some jaeger agent running and listening on localhost:6831
- have jaeger collector and query working
- run examples/jaeger

What did you expect to see?

- Jaeger UI shows `some.attribute` with value 42 among other span's attributes.

What did you see instead?

- There is no such attribute in Jaeger UI
- No warning during compilation
- No assert nor log nor anything during runtime

**Additional context**

- in exporters/jaeger/src/recordable.cc
  - in function `JaegerRecordable::PopulateAttribute()` only some types of attributes are stored as tags.
  - unsupported types are just: `// TODO: extend other AttributeType to the types supported by Jaeger.`

There should be some assert at least.
Please consider support at least all non-vector types. This actually means static_cast of few integer types of different size and with/without the sign.
Vector-type attributes may be stored as a list of normal attributes with index suffix in name, i.e. `my.vector.0`, `my.vector.1`, etc. Just a thought.

esigo:
Fixes #1104 (issue)

## Changes

adds support for the missing types. if not logs an error.
Also bazel unit test build added.
Fixed compiler warnings in the unit test.


exporters/jaeger/src/recordable.cc
```diff
@@ -39,14 +56,97 @@ void JaegerRecordable::PopulateAttribute(nostd::string_view key,
...
+  }
+  else if (nostd::holds_alternative<int32_t>(value))
+  {
+    AddTag(std::string{key}, int64_t{nostd::get<int32_t>(value)}, tags);
+  }
+  else if (nostd::holds_alternative<uint32_t>(value))
+  {
+    AddTag(std::string{key}, int64_t{nostd::get<uint32_t>(value)}, tags);
+  }
+  else if (nostd::holds_alternative<uint64_t>(value))
```
reyang, line 145: Curious - is there specific ordering? (I saw uint32 was after int32, while uint64 was before int64, hence the ask).

esigo: I made it consistent. The order should not matter though.

reyang: Thanks!

reyang (approved):
LGTM once the changelog is updated. It might be nice to have some test cases.

exporters/jaeger/src/recordable.cc
```diff
@@ -19,10 +20,26 @@ void JaegerRecordable::PopulateAttribute(nostd::string_view key,
                                          const common::AttributeValue &value,
                                          std::vector<thrift::Tag> &tags)
 {
-  if (nostd::holds_alternative<int64_t>(value))
+  if (nostd::holds_alternative<int>(value))
```
lalitb, line 23: shouldn't we be comparing only for the supported types from here: https://github.com/open-telemetry/opentelemetry-cpp/blob/b1b1f6409a3a95e783f746920ba48006d8d32926/api/include/opentelemetry/common/attribute_value.h#L37

So, int, uint64_t are not valid attribute types.

esigo: thanks, removed unsupported types

esigo:
> LGTM once the changelog is updated. It might be nice to have some test cases.

@reyang done

lalitb (approved):

lalitb:
Can be merged once the conflict is resolved.

ThomsonTan (approved):

PR #1110, opened 2021-12-03
Metrics SDK: View API

### LINKED ISSUE #1088: Metrics SDK : Add Views

lalitb:
Fixes #1088 

## Changes

- Implements View API: View, View Registry,  InstrumentSelector, MeterSelector
- Placeholder for Aggregator and Aggregation.


lalitb:
@jsuereth  -  As @reyang is on vacation this month, would need help from you on reviewing this. This is still not the final code and will change as we progress further on Aggregation and Aggregator APIs.

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
```diff
@@ -0,0 +1,96 @@
...
+    if (matchAll_)
+    {
+      return true;
+    }
+    else
+    {
+#  if HAVE_WORKING_REGEX
+      return std::regex_match(str.data(), reg_key_);
+#  else
+      return false;  // not supported
```
jsuereth, line 52: Seems like this should fail in some fashion.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -0,0 +1,34 @@
...
+class AttributesProcessor
+{
+public:
+  virtual MetricAttributes process(
+      const opentelemetry::common::KeyValueIterable &attributes) noexcept = 0;
+};
+
+class DefaultAttributesProcessor : public AttributesProcessor
+{
+  MetricAttributes process(
```
jsuereth, line 23: IIUC - by necessity, this API will need to take in an Iterator and return a concrete object.  It's likely this would get used on the "hot path" of metrics.  Does it make sense to have this API work only against `AttributeMap`, that way in the event you're just passing it through, there isn't a sort + unique step (I forget the details of attribute map).

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
```diff
@@ -0,0 +1,96 @@
...
+  virtual ~Predicate()                                                        = default;
+  virtual bool Match(opentelemetry::nostd::string_view string) const noexcept = 0;
+};
+
+class PatternPredicate : public Predicate
+{
+public:
+  PatternPredicate(opentelemetry::nostd::string_view pattern) : reg_key_{pattern.data()}
+  {
+    if (pattern == "*")
```
jsuereth, line 32: I don't think you need to support `*` in all your predicates.  You could have ONE predicate which is the "star" predicate and then a "predicate factory" that returns either a new `PatternPredicate` or `MatchEverything` predicate, if you wanted.

Only commenting because I noticed this logic duplicate in the "ExactPredicate" code as well.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
```diff
@@ -0,0 +1,104 @@
...
+        if (!callback(*(registered_view->view_.get())))
+        {
+          return false;
+        }
+      }
+    }
+    // return default view if none found;
+    if (!found)
+    {
+      static View view(kDefaultViewName, kDefaultViewDescription);
```
jsuereth, line 69: [blocking] The default name and description *should* come from the instrument descriptor's name/description.

jsuereth (commented):
Overall looks great!

- One major comment on default view name/description (the default metric from an instrument with new view is its "natural" aggregation (marked TODO) and the instrument name/description)
- A couple questions/nits that can get resolved as implementation/benchmarking land to think about.

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
lalitb: Yes, this will be supported eventually. For now, we can add the LOG_ERROR to log the error while returning `false`.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
lalitb: Yes, your understanding is correct here. The metrics API takes non-owning KeyValueIterable object as [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/9dd3da33b03fac66af275c417938e0dc69f4eda6/api/include/opentelemetry/metrics/sync_instruments.h#L40), and the plan is to make a copy of this only when required in hot-path. Attribute-processor could be one of the places to create a copy in case of passthrough.

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
lalitb: Good point. Will change accordingly.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
lalitb: Should we return default view ( with empty name and description ) as done in Java [here](https://github.com/open-telemetry/opentelemetry-java/blob/32f7138131d18b5f1d13667ae31d35dc015d21ed/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/view/ViewRegistry.java#L55). This will allow us to return a default `static` immutable object always when a view is not found.

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
lalitb: This is changed now.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
lalitb: Have made changes as per my suggestion above. Let me know if that works.

sdk/include/opentelemetry/sdk/metrics/view/predicate.h
lalitb: Added LOG_ERROR. Let me know if you have a better way to fail. This will eventually return a default View. We don't throw exceptions in sdk. Let me know if you have a better way to fail.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
jsuereth: I expect this to (commonly) be a passthrough case.   Can deal with it after initial implementation :)

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
```diff
@@ -0,0 +1,102 @@
...
+        if (!callback(*(registered_view->view_.get())))
+        {
+          return false;
+        }
+      }
+    }
+    // return default view if none found;
+    if (!found)
+    {
+      static View view("");
```
jsuereth, line 67: Shouldn't this still pull name/description from the instrument?

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
jsuereth: Ah, I see, you're planning to *then* select the name of the metric by checking for empty description or empty name.

I understand wanting to keep this static.  You'll need to *either* use Java's "MetricDescriptor" trick to figure out what to name the underlying metric, or keep this logic elsewhere.

That's entirely reasonable to do, but remember you'll still be instantiating something to remember the name/description of the resulting metric.   For C++ it may be easier to do so view the View object, but you can decide once you work on the "storage" aspect of views :)

jsuereth (approved):

PR #1111, opened 2021-12-03
Add JSON support to OStreamSpanExporter

### LINKED ISSUE #202: Add JSON formatting to OStreamExporters
There is currently no way for the user to change the format that OStreamExporters export the data. If the user wants to send their data directly to a json file, it will not be in the correct format.

Have the user pass in a boolean for json printing in the constructor. Then in export() format the data what the user has specified. Since the user won't directly call export() it makes sense to pass the boolean in the constructor.

An alternative is to support json formatting as the default and only printing. This doesn't seem ideal as most users will only print to stdout and will have a harder time reading the output.

**Additional context**
The two exporters are [218](https://github.com/open-telemetry/opentelemetry-cpp/pull/218) and [183](https://github.com/open-telemetry/opentelemetry-cpp/pull/183)

deejgregor:
Fixes #202.

## Changes

I modeled this change on the Python implementation of [ReadableSpan.to_json()](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L443).

Right now, this only contains support for OStreamSpanExporter, but I'll see if I can add support for the other OStream exporters when I'm on end of year break. Any feedback on my approach in the meantime would be helpful before I do the other exporters. I'll also note that I'm barely familiar with C++, so please don't hesitate to point out even minor issues.

## TODO

* [ ] Complete handling of all `OwnedAttributeValue` variants in `PopulateAttribute`.
* [ ] Support JSON for `OStreamLogExporter` and `OStreamMetricsExporter`.
* [ ] Any other refactoring opportunities?

### DIFF
--- a/exporters/ostream/CMakeLists.txt
+++ b/exporters/ostream/CMakeLists.txt
@@ -8,7 +8,13 @@ target_include_directories(
   PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>")
 
 target_link_libraries(opentelemetry_exporter_ostream_span
-                      PUBLIC opentelemetry_trace)
+                      PUBLIC opentelemetry_trace
+                             nlohmann_json::nlohmann_json)
+
+if(nlohmann_json_clone)
+  add_dependencies(opentelemetry_exporter_otlp_http_client
+                   nlohmann_json::nlohmann_json)
+endif()
 
 install(
   TARGETS opentelemetry_exporter_ostream_span
--- a/exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
+++ b/exporters/ostream/include/opentelemetry/exporters/ostream/span_exporter.h
@@ -10,8 +10,11 @@
 #include "opentelemetry/version.h"
 
 #include <iostream>
+#include <list>
 #include <map>
 #include <sstream>
+
+#include "nlohmann/json.hpp"
 
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace exporter
@@ -30,7 +33,7 @@ public:
    * export() function will send span data into.
    * The default ostream is set to stdout
    */
-  explicit OStreamSpanExporter(std::ostream &sout = std::cout) noexcept;
+  explicit OStreamSpanExporter(std::ostream &sout = std::cout, bool isJson = false) noexcept;
 
   std::unique_ptr<opentelemetry::sdk::trace::Recordable> MakeRecordable() noexcept override;
 
@@ -43,6 +46,7 @@ public:
 
 private:
   std::ostream &sout_;
+  bool isJson_;
   bool is_shutdown_ = false;
   mutable opentelemetry::common::SpinLockMutex lock_;
   bool isShutdown() const noexcept;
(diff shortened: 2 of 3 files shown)


lalitb:
Thanks for taking up this item. A couple of points, we can discuss if they don't make sense:
 - As of now, core otel-cpp components ( API, SDK, OStream, and Memory exporters ) don't have any external library dependencies. Do you think we should enable/disable JSON support using macro, as this will not introduce a dependency on nlohmann_json by default?
 -  Exporters create their implementation of `Recordable` to get the data in a format they need. As an example, Zipkin exporter has its own `Recordable` implementation to format data in JSON format ( https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/zipkin/src/recordable.cc). OStream exporter as of now uses `SpanData` as default recordable implementation for console output. Do you think we should introduce a new Recordable for JSON?

esigo:
Thanks for your contribution. 
As `OStreamSpanExporter` is the entry point for the new users, I believe we should keep it as simple as possible (ideally with no external dependencies and code pollution).
Do you think we can add the JSON support to [contrib repo](https://github.com/open-telemetry/opentelemetry-cpp-contrib)?

deejgregor:
Thanks @lalitb and @esigo. I'll be working on this more over the holidays, and your feedback will be very helpful. Thanks, again!

lalitb:
@deejgregor - Do you have any updates here. Thanks.

deejgregor:
> * As of now, core otel-cpp components ( API, SDK, OStream, and Memory exporters ) don't have any external library dependencies. Do you think we should enable/disable JSON support using macro, as this will not introduce a dependency on nlohmann_json by default?

Ah, yes. For some reason, I thought `nlohmann_json` was required by default, and it is not.

> * Exporters create their implementation of `Recordable` to get the data in a format they need. As an example, Zipkin exporter has its own `Recordable` implementation to format data in JSON format ( https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/zipkin/src/recordable.cc). OStream exporter as of now uses `SpanData` as default recordable implementation for console output. Do you think we should introduce a new Recordable for JSON?

+1. That will be a nicer way to go

@lalitb thanks for your comments above, and after (finally) poking around a bit and understanding more, I think it makes sense to do both of these.

deejgregor:
> Thanks for your contribution. As `OStreamSpanExporter` is the entry point for the new users, I believe we should keep it as simple as possible (ideally with no external dependencies and code pollution). Do you think we can add the JSON support to [contrib repo](https://github.com/open-telemetry/opentelemetry-cpp-contrib)?

@esigo I agree about not adding complexity to OStreamSpanExporter and I'll make a new exporter.

What do you think about having the JSON exporter in the main repo, though, instead of contrib as you suggest? On one hand, its only dependency is nlohmann_json which is already used by other exporters in the main repo. On the other hand, since this exporter won't be abled by default since nlohmann_json isn't a base requirement, I'm not sure if this exporter will get much use, so maybe the contrib repo is the better place for it.

Frankly, as I've *finally* started using a real exporter to Jaeger, my own need for a JSON exporter for testing is pretty much gone. I'm starting to wonder how much use a JSON exporter will be for others.... in particular, is this useful to contribute back at all, regardless of which repo it goes in? Curious on your thoughts (and anyone else's).

PR #1113, opened 2021-12-03
Fix otlp generates null span ids

### LINKED ISSUE #1107: opentelemetry-cpp generates null span ids
https://github.com/open-telemetry/opentelemetry-cpp/blob/b1b1f6409a3a95e783f746920ba48006d8d32926/exporters/otlp/src/otlp_recordable.cc#L23

Hi all,

When creating root tracing spans in C++, I see the following in my tracing vendor (Honeycomb):

image

After inspecting the span in Honeycomb, the problem is that it has a parent spanid of `0000000000000000`. If I make the following change, I get a normal-looking span. See change and normal-looking span below:

```cpp
if (parent_span_id.IsValid()) {
  span_.set_parent_span_id(reinterpret_cast<const char *>(parent_span_id.Id().data()),
                       trace::SpanId::kSize);
}
```

image

esigo:
Fixes #1107 (issue)

## Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Increment the:
 
 ## [Unreleased]
 
+* [EXPORTER] Fix otlp generates null span ids ([#1106](https://github.com/open-telemetry/opentelemetry-cpp/pull/1106))
+
 ## [1.1.0] 2021-11-19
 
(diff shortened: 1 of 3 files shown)


lalitb (approved):
Thanks.

ThomsonTan (approved):

PR #1116, opened 2021-12-04
`jaeger::TUDPTransport::write` crash when `getaddrinfo` returns error

### LINKED ISSUE #1114: jaeger::TUDPTransport::write crash when wrong endpoint used
**Describe your environment** 

opentelemetry-cpp ci build image

**Steps to reproduce**

run `jaeger` example with wrong endpoint e.g.:
`bazel-bin/examples/jaeger/example_jaeger host.docker.internal:16386`

no crash, ideally an error message or an exception

What did you expect to see?

no crash

no crash

What did you see instead?

crash

**Additional context**

esigo:
Fixes #1114 (issue)

## Changes

when `getaddrinfo` used in `jaeger::TUDPTransport::open` returns error, we still tend to dereference `server_addr_info_` in member methods while it is `null`.
The PR adds the null check to avoid `SIGSEGV`.


exporters/jaeger/src/TUDPTransport.cc
```diff
@@ -75,6 +75,10 @@ void TUDPTransport::close()
 
 uint32_t TUDPTransport::read(uint8_t *buf, uint32_t len)
 {
+  if (!server_addr_info_)
```
ThomsonTan, line 78: Thanks for the catch. Could we also add internal log here?

esigo: I was afraid that the log here would be too much if the the function is called often.
I add the log in a sec.

ThomsonTan: Ok, perhaps add it right after the call to `getaddrinfo`.

esigo: There is already a log there:

`[Error] File: exporters/jaeger/src/TUDPTransport.cc:57Jaeger Exporter: getaddrinfo failed with error: -2`

Shall I change the message?

lalitb (approved):

exporters/jaeger/src/TUDPTransport.cc
ThomsonTan: This current message is fine.

ThomsonTan (approved):

PR #1119, opened 2021-12-07
Document dependencies

### LINKED ISSUE #1108: Add minimum version of Thrift in Jaeger exporter build instructions
**Describe your environment**

Ubuntu 16.04 with Apache Thrift 0.12.0 and trying to build opentelemetry-cpp v1.1.0

**Steps to reproduce**

```
cmake -DWITH_JAEGER=ON -DBUILD_TESTING=OFF -DWITH_EXAMPLES=OFF -DWITH_PROMETHEUS=ON ..
make -j
```

Building should succeed or should make clear why it doesn't work

The build fails with this error:
```
opentelemetry-cpp-1.1.0/exporters/jaeger/src/THttpTransport.h:23:8: error: ‘bool opentelemetry::v1::exporter::jaeger::THttpTransport::isOpen() const’ marked ‘override’, but does not override
   23 |   bool isOpen() const override;
```

**Additional context**
So the reason this fails is that Apache Thrift 0.12.0 doesn't mark `isOpen()` as const so the override fails. Thrift 0.13.0 does mark it const but there may be other build issues with this version. I was able to build with 0.15.0 at least. Could you add some information about what versions of Thrift work in the documentation? Or assert the version in the CMake build?

esigo:
Fixes #1108 (issue)

## Changes

Adds dependencies to INSTALL.md.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,7 +5,7 @@ build systems for opentelemetry-cpp.
 
 ## Dependencies
 
-You can link OpenTelemetry with libraries provided in [dependencies.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/dependencies.md) (complete list of libraries with versions used in our CI can be found [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release)).
+You can link OpenTelemetry C++ SDK with libraries provided in [dependencies.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/dependencies.md) (complete list of libraries with versions used in our CI can be found [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release)).
 
 ## Build instructions using CMake
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -5,12 +5,13 @@ build systems for opentelemetry-cpp.
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -3,6 +3,16 @@
 [CMake](https://cmake.org/) and [Bazel](https://bazel.build) are the official
 build systems for opentelemetry-cpp.
 
+## Dependencies
+
```
lalitb, line 7: There is also a more detailed dependencies doc here ( https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/dependencies.md).
Can we add a reference to the above doc along with https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release for versions in this section?

esigo: oh, how I missed this file :). 
Thanks removed the libs.

lalitb (approved):

INSTALL.md
```diff
@@ -3,6 +3,10 @@
...
+You can link OpenTelemetry with libraries provided in [dependencies.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/dependencies.md) (complete list of libraries with versions used in our CI can be found [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release)).
```
ThomsonTan, line 8: Make it specific to "You can link OpenTelemetry C++ SDK ...“?

esigo: done

ThomsonTan (approved):

PR #1134, opened 2021-12-13
Jaeger exporter Windows and Mac bazel

### LINKED ISSUE #1101: Jaeger bazel Windows and Mac
Jaeger bazel Windows and Mac require boost installed.

ideally `bazel` should use `rules_foreign_cc` and add `boost` as dependency.

**Additional context**

esigo:
Fixes #1101 (issue)

## Changes


WORKSPACE
```diff
@@ -16,9 +16,45 @@ workspace(name = "io_opentelemetry_cpp")
...
+openssl_repositories()
+
+# Load prometheus C++ dependencies.
+load("@com_github_jupp0r_prometheus_cpp//bazel:repositories.bzl", "prometheus_cpp_repositories")
+
+prometheus_cpp_repositories()
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
owent, line 32: I'm not sure, but is `bazel_skylib` the internal module of bazel? Is it necessary to load it from github?

esigo: It's needed [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/610193c563e7c85a98f11dbbdf588f2f583a372d/bazel/openssl_repositories.bzl#L9).

owent: Thanks for explain

bazel/openssl_repositories.bzl
```diff
@@ -0,0 +1,58 @@
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
```
owent, line 1: Is it possible to let grpc to use the same ssl library as jaeger exporter?If we have a program which links both otlp grpc exporter and jaeger exporter, there may be symbol conflicts of libressl(used by grpc by default) and openssl.

WORKSPACE
```diff
@@ -16,9 +16,45 @@ workspace(name = "io_opentelemetry_cpp")
...
+    ],
+)
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
```
owent, line 47: Should all `git_repository` and `http_archive` moved into `bazel/repository.bzl` ?

bazel/openssl_repositories.bzl
esigo: `gRPC` is using `boringssl`. There is a [trick](https://groups.google.com/g/grpc-io/c/nBjdFT-d1Bo/m/QQ1VuOhCAQAJ) to make gRPC use openssl. Though we are already using the boringssl that comes form gRPC in our [CI](https://github.com/open-telemetry/opentelemetry-cpp/blob/f380fcba45112b91deb53d14daae785cdbf320b5/ci/setup_grpc.sh#L43), while using `openssl` for Jaeger exporter.

WORKSPACE
esigo: yeah thanks I cleaned the WORKSPACE.

bazel/openssl_repositories.bzl
owent: boringssl and openssl have some symbols with same name. I just wonder whether they have 100% ABI compatibility with each other. If so, it's fine to use one implementation of some functions and global variables and use the rest functions and global variables of another.

esigo: I believe they are not [compatible](https://boringssl.googlesource.com/boringssl/+/HEAD/PORTING.md).
I would suggest investigating this in another issue/PR. We should have the same issue in CMake.

owent: In cmake, we can use `-DgRPC_SSL_PROVIDER=package` and `-DOPENSSL_ROOT_DIR=<OpenSSL installation directory>` when building grpc and grpc will use openssl to replace boringssl. But I'm not familiar with bazel and not know how to do this in bazel.

esigo: yes that should be the fix for `CMake`. For `bazel` we should define a new library with the name `boringssl` before importing `gRPC`.
I can propose the fix in another PR for both `CMake` and `bazel`?

owent: Sounds good and thanks.

WORKSPACE
```diff
@@ -16,9 +16,25 @@ workspace(name = "io_opentelemetry_cpp")
 
 # Load our direct dependencies.
 load("//bazel:repository.bzl", "opentelemetry_cpp_deps")
+load("//bazel:openssl_repositories.bzl", "openssl_repositories")
 
 opentelemetry_cpp_deps()
 
+openssl_repositories()
```
lalitb, line 23: Is opessl a dependency for boost? I mean does thrift uses the features of boost which require openssl ( i know boost::asyncio requires openssl).

lalitb (commented):
Have some questions before closing on it. Let's discuss this quickly during the community meeting (if you plan to attend this week), else we can continue the discussion here.

WORKSPACE
```diff
@@ -16,9 +16,25 @@ workspace(name = "io_opentelemetry_cpp")
...
+# Load prometheus C++ dependencies.
+load("@com_github_jupp0r_prometheus_cpp//bazel:repositories.bzl", "prometheus_cpp_repositories")
+
+prometheus_cpp_repositories()
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
+
+load("@com_github_nelhage_rules_boost//:boost/boost.bzl", "boost_deps")
```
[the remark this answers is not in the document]
esigo: It is not official but it is recommended in bazel's [doc](https://docs.bazel.build/versions/1.2.0/rules.html).

WORKSPACE
esigo: It is needed for [thrift](https://github.com/esigo/opentelemetry-cpp/blob/867af702b2df856b79ed65f8d44f86c490984c88/exporters/jaeger/BUILD#L212).

INSTALL.md
```diff
@@ -193,6 +193,15 @@ To install Bazel, consult the [Installing Bazel](https://docs.bazel.build/versio
 
    ```
 
+   Please consider installing `Jaeger` dependencies using [`setup_thrift.ps1`](https://github.com/esigo/opentelemetry-cpp/tree/Jaeger-exporter-Windows-and-Mac-bazel/ci/setup_thrift.ps1) and passing `vcpkg` path as environment variables on Windows:
```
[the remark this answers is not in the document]
esigo: Thank, I fixed the link

esigo:
> Have some questions before closing on it. Let's discuss this quickly during the community meeting (if you plan to attend this week), else we can continue the discussion here.

Thanks for the review, let's discuss in the meeting.

lalitb:
@esigo Thanks, the changes are very nicely done. I have one suggestion here -

- Can we restrict this PR ( or maybe a smaller separate PR  ) to enable Jaeger bazel build for windows using vcpkg.
- For Linux and macS, we can install boost and OpenSSL using apt-get and brew respectively, and then continue thrift installation using bazel ( as we have support for it thanks to you :) ). We already document the dependencies [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#dependencies) for users to ensure these are met before building.

What do you think about it ?

esigo:
> @esigo Thanks, the changes are very nicely done. I have one suggestion here -
> 
> * Can we restrict this PR ( or maybe a smaller separate PR  ) to enable Jaeger bazel build for windows using vcpkg.
> * For Linux and macS, we can install boost and OpenSSL using apt-get and brew respectively, and then continue thrift installation using bazel ( as we have support for it thanks to you :) ). We already document the dependencies [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#dependencies) for users to ensure these are met before building.
> 
> What do you think about it ?

Thanks for the idea, the issue is with Mac, a config similar to [this](https://github.com/esigo/opentelemetry-cpp/blob/e903e1676f0b413a620b5c126b99db473243391a/exporters/jaeger/BUILD#L205) is needed, which it didn't work using brew.
One option would be to install dependencies using vcpkg and do the same thing for Mac, Windows (and maybe also Linux). The downside is that vcpkg install will be slower compared to directly using boost rule. Also in this PR we are treating Mac and Linux the same most of the time, with vcpkg we will need to treat each OS differently.
@lalitb Shall I try vcpkg for both Mac and Windows?

lalitb:
> @lalitb Shall I try vcpkg for both Mac and Windows?

Let me think more on Mac, not sure if using vcpkg would be a good idea. Meanwhile, do you think we are in a good shape to enable bazel build for Jaeger exporter on Windows (using vcpkg)? If we can separate out that in a PR, while we continue discussing for Mac.

bazel/repository.bzl
```diff
@@ -103,3 +112,56 @@ def opentelemetry_cpp_deps():
         strip_prefix = "curl-7.73.0",
         urls = ["https://curl.haxx.se/download/curl-7.73.0.tar.gz"],
     )
+
+    # boost
```
ThomsonTan, line 116: Is boost optional as libthrift is optoin?

ci/setup_thrift.ps1
```diff
@@ -0,0 +1,15 @@
+$ErrorActionPreference = "Stop"
+trap { $host.SetShouldExit(1) }
+
+# TODO: consider moving this to .gitmodules
+git clone https://github.com/Microsoft/vcpkg.git
```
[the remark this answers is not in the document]
ThomsonTan: And checkout a stable branch instead of the latest?

esigo:
> > @lalitb Shall I try vcpkg for both Mac and Windows?
> 
> Let me think more on Mac, not sure if using vcpkg would be a good idea. Meanwhile, do you think we are in a good shape to enable bazel build for Jaeger exporter on Windows (using vcpkg)? If we can separate out that in a PR, while we continue discussing for Mac.

@lalitb yes that should be OK. I'll make a clean PR to support Jaeger bazel on Windows with boost and openssl coming from vcpkg.

esigo:
> > > @lalitb Shall I try vcpkg for both Mac and Windows?
> > 
> > 
> > Let me think more on Mac, not sure if using vcpkg would be a good idea. Meanwhile, do you think we are in a good shape to enable bazel build for Jaeger exporter on Windows (using vcpkg)? If we can separate out that in a PR, while we continue discussing for Mac.
> 
> @lalitb yes that should be OK. I'll make a clean PR to support Jaeger bazel on Windows with boost and openssl coming from vcpkg.

@lalitb my idea didn't work because boost is still needed for consuming thrift. I'm afraid simplification is not possible.

If you are ok with this PR, I'll push fix for other comments.

lalitb:
> @lalitb my idea didn't work because boost is still needed for consuming thrift. I'm afraid simplification is not possible.
> 
> If you are ok with this PR, I'll push fix for other comments.

Sorry, it's not clear to me. For Windows `setup_thrift.ps1` should take care of boost and openssl installation, and then we have thrift installation using bazel - will this not work?

esigo:
> > @lalitb my idea didn't work because boost is still needed for consuming thrift. I'm afraid simplification is not possible.
> > If you are ok with this PR, I'll push fix for other comments.
> 
> Sorry, it's not clear to me. For Windows `setup_thrift.ps1` should take care of boost and openssl installation, and then we have thrift installation using bazel - will this not work?

oh I see the issue I missed. So the boost headers are needed for consuming thrift, and also boost library files  for linking thrift itself. I have mixed them now, so Windows uses vcpkg for linking and the rule's boost for the headers. Thanks for pointing this out, let me see if I can find a way to sort it out.

esigo:
> > @lalitb Shall I try vcpkg for both Mac and Windows?
> 
> Let me think more on Mac, not sure if using vcpkg would be a good idea. Meanwhile, do you think we are in a good shape to enable bazel build for Jaeger exporter on Windows (using vcpkg)? If we can separate out that in a PR, while we continue discussing for Mac.

closing this PR as discussed. #1227 will bring jaeger bazel for Windows.

PR #1136, opened 2021-12-14
Improve the instructions for Bazel build.

vasireddy99:
This PR is to propose the changes in instructions for building with bazel and adding the note to not to have GoogleTest installed.

Bazel will download, build and install gtest version from [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/5e258b0c0c192620fe3410b9e56c7480fa278a82/bazel/repository.bzl#L27) and it will fail if it gets different gtest headers from system defined path.

Please feel free to add review comments.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -137,13 +137,9 @@ target_link_libraries(foo PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES})
 
 ## Build instructions using Bazel
 
-<<<<<<< HEAD
 NOTE: Experimental, and not supported for all the components. Make sure the
 [GoogleTest](https://github.com/google/googletest) installation may fail if there is a different version of googletest already installed in system-defined path.
-=======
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -137,7 +137,8 @@ target_link_libraries(foo PRIVATE ${OPENTELEMETRY_CPP_LIBRARIES})
 
 ## Build instructions using Bazel
 
-NOTE: Experimental, and not supported for all the components.
+NOTE: Experimental, and not supported for all the components. Make sure the 
+[GoogleTest](https://github.com/google/googletest) is not installed prior to using the Bazel.
```
lalitb, line 141: We can re-phrase it as `[GoogleTest](https://github.com/google/googletest) installation may fail if there is a different version of googletest already installed in system-defined path.`

lalitb (approved):
LGTM. Thanks.

ThomsonTan (approved):

PR #1153, opened 2021-12-19
ZipkinExporter shutdown

esigo:
Fixes ZipkinExporter shutdown

## Changes

ZipkinExporter shutdown should set shutdown before return

### DIFF
--- a/exporters/zipkin/src/zipkin_exporter.cc
+++ b/exporters/zipkin/src/zipkin_exporter.cc
@@ -98,6 +98,12 @@ void ZipkinExporter::InitializeLocalEndpoint()
   local_end_point_["port"] = url_parser_.port_;
 }
 
+bool ZipkinExporter::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  isShutdown_ = true;
+  return true;
+}
+
 }  // namespace zipkin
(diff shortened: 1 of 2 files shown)


lalitb (approved):

ThomsonTan (approved):

exporters/zipkin/src/zipkin_exporter.cc
```diff
@@ -98,6 +98,12 @@ void ZipkinExporter::InitializeLocalEndpoint()
   local_end_point_["port"] = url_parser_.port_;
 }
 
+bool ZipkinExporter::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  isShutdown_ = true;
```
lalitb, line 103: Not related to PR, but probably we need to discuss if this should be an atomic operation for all exporters to protect against Shutdown() and Export() called from separate threads. While specs don't allow Export() to be called concurrently ( https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch ), it doesn't talk about Export() and Shutdown().

esigo: Agree it should be protected somehow.

PR #1154, opened 2021-12-20
Logger: Propagating resources through LoggerProvider

### LINKED ISSUE #1086: Logger: Propagating resources through LoggerProvider
The implementation of logger API/SDK was inspired by the initial trace implementation. And there was no support of resources at that time. This support needs to be added to Logger. It would be good to replicate the implementation from the latest trace.
 - Introduce LoggerContext to manager ownership of resources and processor ( similar to [TracerContext ](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/trace/tracer_context.h))
 - Inject resource through LoggerProvider, as done by TracerProvider: https://github.com/open-telemetry/opentelemetry-cpp/blob/089b04d77fc9f2e0e4e4d87b1fff5b10a79ab3e1/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L41 
 - Propagating resources to exporters through recordable ( Need to validate if this is fully supported, or any changes required).

owent:
Signed-off-by: owentou <owentou@tencent.com>

Fixes #1086 

## Changes

+ Add `logs::LoggerContext`, `logs::MultiLogProcessor`, `logs::MultiRecordable`
+ Remove resource from `Logger::Log`
+ Remove `TracerProvider::GetProcessor` (Without implementation before.)
+ \[BREAK CHANGES\] Change the declaration of `sdk::logs::Recordable::SetResource`
+ \[BREAK CHANGES\] Change the `LoggerProvider::SetProcessor` to `LoggerProvider::AddProcessor`
+ Share `LoggerContext` between loggers
+ Add `api/include/opentelemetry/common/macros.h` to provide `OPENTELEMETRY_DEPRECATED`, `OPENTELEMETRY_DEPRECATED_MESSAGE` and `OPENTELEMETRY_MAYBE_UNUSED`

Is there any way to convert `opentelemetry::sdk::common::OwnedAttributeValue` into `opentelemetry::common::AttributeValue` ?
It's difficult to reuse and dump data from `opentelemetry::sdk::common::OwnedAttributeValue`.

### DIFF
--- a/api/include/opentelemetry/common/macros.h
+++ b/api/include/opentelemetry/common/macros.h
@@ -9,45 +9,45 @@
 
 /// \brief Declare class, variable or functions as deprecated
 /// usage:
-///   OPENTELEMETRY_DEPRECATED_ATTR int a;
-///   class OPENTELEMETRY_DEPRECATED_ATTR a;
-///   OPENTELEMETRY_DEPRECATED_ATTR int a();
+///   OPENTELEMETRY_DEPRECATED int a;
+///   class OPENTELEMETRY_DEPRECATED a;
+///   OPENTELEMETRY_DEPRECATED int a();
 /// usage:
-///   OPENTELEMETRY_DEPRECATED_MSG("there is better choose") int a;
+///   OPENTELEMETRY_DEPRECATED_MESSAGE("there is better choose") int a;
 ///   class DEPRECATED_MSG("there is better choose") a;
-///   OPENTELEMETRY_DEPRECATED_MSG("there is better choose") int a();
+///   OPENTELEMETRY_DEPRECATED_MESSAGE("there is better choose") int a();
 ///
 
 #if defined(__cplusplus) && __cplusplus >= 201402L
-#  define OPENTELEMETRY_DEPRECATED_ATTR [[deprecated]]
+#  define OPENTELEMETRY_DEPRECATED [[deprecated]]
 #elif defined(__clang__)
-#  define OPENTELEMETRY_DEPRECATED_ATTR __attribute__((deprecated))
+#  define OPENTELEMETRY_DEPRECATED __attribute__((deprecated))
 #elif defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
-#  define OPENTELEMETRY_DEPRECATED_ATTR __attribute__((deprecated))
+#  define OPENTELEMETRY_DEPRECATED __attribute__((deprecated))
 #elif defined(_MSC_VER) && _MSC_VER >= 1400  // vs 2005 or higher
 #  if _MSC_VER >= 1910 && defined(_MSVC_LANG) && _MSVC_LANG >= 201703L
-#    define OPENTELEMETRY_DEPRECATED_ATTR [[deprecated]]
+#    define OPENTELEMETRY_DEPRECATED [[deprecated]]
 #  else
-#    define OPENTELEMETRY_DEPRECATED_ATTR __declspec(deprecated)
+#    define OPENTELEMETRY_DEPRECATED __declspec(deprecated)
 #  endif
 #else
-#  define OPENTELEMETRY_DEPRECATED_ATTR
+#  define OPENTELEMETRY_DEPRECATED
 #endif
 
 #if defined(__cplusplus) && __cplusplus >= 201402L
-#  define OPENTELEMETRY_DEPRECATED_MSG(msg) [[deprecated(msg)]]
+#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
 #elif defined(__clang__)
-#  define OPENTELEMETRY_DEPRECATED_MSG(msg) __attribute__((deprecated(msg)))
+#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __attribute__((deprecated(msg)))
 #elif defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
-#  define OPENTELEMETRY_DEPRECATED_MSG(msg) __attribute__((deprecated(msg)))
+#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __attribute__((deprecated(msg)))
 #elif defined(_MSC_VER) && _MSC_VER >= 1400  // vs 2005 or higher
 #  if _MSC_VER >= 1910 && defined(_MSVC_LANG) && _MSVC_LANG >= 201703L
-#    define OPENTELEMETRY_DEPRECATED_MSG(msg) [[deprecated(msg)]]
+#    define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
 #  else
-#    define OPENTELEMETRY_DEPRECATED_MSG(msg) __declspec(deprecated(msg))
+#    define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __declspec(deprecated(msg))
 #  endif
 #else
-#  define OPENTELEMETRY_DEPRECATED_MSG(msg)
+#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg)
 #endif
 
 /// \brief Declare variable as maybe unused
--- a/api/include/opentelemetry/common/macros.h
+++ b/api/include/opentelemetry/common/macros.h
@@ -7,49 +7,6 @@
 
 #include "opentelemetry/version.h"
 
-/// \brief Declare class, variable or functions as deprecated
-/// usage:
-///   OPENTELEMETRY_DEPRECATED int a;
-///   class OPENTELEMETRY_DEPRECATED a;
-///   OPENTELEMETRY_DEPRECATED int a();
-/// usage:
-///   OPENTELEMETRY_DEPRECATED_MESSAGE("there is better choose") int a;
-///   class DEPRECATED_MSG("there is better choose") a;
-///   OPENTELEMETRY_DEPRECATED_MESSAGE("there is better choose") int a();
-///
-
-#if defined(__cplusplus) && __cplusplus >= 201402L
-#  define OPENTELEMETRY_DEPRECATED [[deprecated]]
-#elif defined(__clang__)
-#  define OPENTELEMETRY_DEPRECATED __attribute__((deprecated))
-#elif defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
-#  define OPENTELEMETRY_DEPRECATED __attribute__((deprecated))
-#elif defined(_MSC_VER) && _MSC_VER >= 1400  // vs 2005 or higher
-#  if _MSC_VER >= 1910 && defined(_MSVC_LANG) && _MSVC_LANG >= 201703L
-#    define OPENTELEMETRY_DEPRECATED [[deprecated]]
-#  else
-#    define OPENTELEMETRY_DEPRECATED __declspec(deprecated)
-#  endif
-#else
-#  define OPENTELEMETRY_DEPRECATED
-#endif
-
-#if defined(__cplusplus) && __cplusplus >= 201402L
-#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
-#elif defined(__clang__)
-#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __attribute__((deprecated(msg)))
-#elif defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
-#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __attribute__((deprecated(msg)))
-#elif defined(_MSC_VER) && _MSC_VER >= 1400  // vs 2005 or higher
-#  if _MSC_VER >= 1910 && defined(_MSVC_LANG) && _MSVC_LANG >= 201703L
-#    define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) [[deprecated(msg)]]
-#  else
-#    define OPENTELEMETRY_DEPRECATED_MESSAGE(msg) __declspec(deprecated(msg))
-#  endif
-#else
-#  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg)
-#endif
-
 /// \brief Declare variable as maybe unused
 /// usage:
 ///   OPENTELEMETRY_MAYBE_UNUSED int a;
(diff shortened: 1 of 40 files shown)


lalitb:
> Is there any way to convert opentelemetry::sdk::common::OwnedAttributeValue into opentelemetry::common::AttributeValue ?
It's difficult to reuse and dump data from opentelemetry::sdk::common::OwnedAttributeValue.

Where do you see its usage? We can possibly create a converter similar to AttributeValue to OwnedAttributeValue -https://github.com/open-telemetry/opentelemetry-cpp/blob/7b9132c9d25b4795f46ce33854551432b8cf840f/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L65

owent:
> > Is there any way to convert opentelemetry::sdk::common::OwnedAttributeValue into opentelemetry::common::AttributeValue ?
> > It's difficult to reuse and dump data from opentelemetry::sdk::common::OwnedAttributeValue.
> 
> Where do you see its usage? We can possibly create a converter similar to AttributeValue to OwnedAttributeValue -
> 
> https://github.com/open-telemetry/opentelemetry-cpp/blob/7b9132c9d25b4795f46ce33854551432b8cf840f/sdk/include/opentelemetry/sdk/common/attribute_utils.h#L65

I have wrote a example to share resource between logs before.And the codes is like these:

```cpp
class KeyValueIterableViewForResourceAttributes final : public opentelemetry::common::KeyValueIterable {
 public:
  explicit KeyValueIterableViewForResourceAttributes(
      const opentelemetry::sdk::resource::ResourceAttributes& container) noexcept
      : container_{&container} {}

  // KeyValueIterable
  bool ForEachKeyValue(
      opentelemetry::nostd::function_ref<bool(nostd::string_view, opentelemetry::common::AttributeValue)> callback)
      const noexcept override {
    auto iter = std::begin(*container_);
    auto last = std::end(*container_);
    for (; iter != last; ++iter) {
      if (!callback(iter->first, ConvertOwnedAttribute(iter->second))) {
        return false;
      }
    }
    return true;
  }

  size_t size() const noexcept override { return nostd::size(*container_); }

  static opentelemetry::common::AttributeValue ConvertOwnedAttribute(
      const opentelemetry::sdk::common::OwnedAttributeValue& value) {
    if (nostd::holds_alternative<bool>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<bool>(value)};
    } else if (nostd::holds_alternative<int32_t>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<int32_t>(value)};
    } else if (nostd::holds_alternative<int64_t>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<int64_t>(value)};
    } else if (nostd::holds_alternative<uint32_t>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<uint32_t>(value)};
    } else if (nostd::holds_alternative<uint64_t>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<uint64_t>(value)};
    } else if (nostd::holds_alternative<double>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<double>(value)};
    } else if (nostd::holds_alternative<std::string>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::string>(value)};
    } else if (nostd::holds_alternative<std::vector<bool>>(value)) {
      // TODO: Add support for vector<bool>
      return opentelemetry::common::AttributeValue{};
    } else if (nostd::holds_alternative<std::vector<int32_t>>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::vector<int32_t>>(value)};
    } else if (nostd::holds_alternative<std::vector<uint32_t>>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::vector<uint32_t>>(value)};
    } else if (nostd::holds_alternative<std::vector<int64_t>>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::vector<int64_t>>(value)};
    } else if (nostd::holds_alternative<std::vector<uint64_t>>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::vector<uint64_t>>(value)};
    } else if (nostd::holds_alternative<std::vector<double>>(value)) {
      return opentelemetry::common::AttributeValue{nostd::get<std::vector<double>>(value)};
    } else if (nostd::holds_alternative<std::vector<std::string>>(value)) {
      // TODO: Add support to convert vector<std::string> tp vector<nostd::string_view>
      return opentelemetry::common::AttributeValue{};
    }

    return opentelemetry::common::AttributeValue{};
  }

 private:
  const opentelemetry::sdk::resource::ResourceAttributes* container_;
};

void Report(const opentelemetry::sdk::resource::Resource& shared_resource) {
  auto span = GetTracer()->StartSpan("cpp-test-span-otlp");
  auto ctx = span->GetContext();
  auto logger = GetLogger();
  auto shared_resource_attributes = KeyValueIterableViewForResourceAttributes(shared_resource.GetAttributes());
  using attribute_span = nostd::span<const std::pair<nostd::string_view, opentelemetry::common::AttributeValue>>;
  logger->Log(opentelemetry::logs::Severity::kInfo, "log name", "log body", shared_resource_attributes,
              opentelemetry::common::KeyValueIterableView<attribute_span>{attribute_span{}}, ctx.trace_id(),
              ctx.span_id(), ctx.trace_flags(), std::chrono::system_clock::now());
}
```

It can not convert all types from `OwnedAttributeValue` to `AttributeValue` . Even atfer move resource into provider, I think it's still useful if we want to `Recordable::SetAttribute` from data with lifetime(I want to reuse `OwnedAttributeValue` to store complex data structures and use it as attributes of trace or log).

lalitb:
> It can not convert all types from `OwnedAttributeValue` to `AttributeValue` . Even atfer move resource into provider, I think it's still useful if we want to `Recordable::SetAttribute` from data with lifetime(I want to reuse `OwnedAttributeValue` to store complex data structures and use it as attributes of trace or log).

Thanks for the explanation with a nicely written example. As you correctly mentioned, the scenario is no more relevant once we start sharing resources through the provider. And passing resource as argument through Logger::log() should be removed. This is a breaking change, but acceptable as Logs Signal is still in preview. For log attributes, it would be better to pass them as KeyValueIterableView. Let me know if I am missing something here.

owent:
> > It can not convert all types from `OwnedAttributeValue` to `AttributeValue` . Even atfer move resource into provider, I think it's still useful if we want to `Recordable::SetAttribute` from data with lifetime(I want to reuse `OwnedAttributeValue` to store complex data structures and use it as attributes of trace or log).
> 
> For log attributes, it would be better to pass them as KeyValueIterableView. Let me know if I am missing something here.

Yes, this is what I mean, and my question is how to create `KeyValueIterableView` of `OwnedAttributeValue` ?

lalitb:
> Yes, this is what I mean, and my question is how to create `KeyValueIterableView` of `OwnedAttributeValue` ?

I am still thinking why should this be under the scope of otel-cpp. Isn't this the responsibility of the library developer to ensure to call `Span::AddEvent(name, attributes)`, `Logger::Log(..., attributes,...)` correctly? They may choose to use SDK::common::AttributMap as storage, or something as simple as std::map<std::string, std::string>. As long as it is converted into Key

> Yes, this is what I mean, and my question is how to create `KeyValueIterableView` of `OwnedAttributeValue` ?

Which conversion is a problem in your above example - `vector<std::string>` to `span<nostd::string_view>` ? I think `vector<bool>` to span<bool> should be implicitly possible.

owent:
> > Yes, this is what I mean, and my question is how to create `KeyValueIterableView` of `OwnedAttributeValue` ?
> 
> I am still thinking why should this be under the scope of otel-cpp. Isn't this the responsibility of the library developer to ensure to call `Span::AddEvent(name, attributes)`, `Logger::Log(..., attributes,...)` correctly? They may choose to use SDK::common::AttributMap as storage, or something as simple as std::map<std::string, std::string>. As long as it is converted into Key
> 
> > Yes, this is what I mean, and my question is how to create `KeyValueIterableView` of `OwnedAttributeValue` ?
> 
> Which conversion is a problem in your above example - `vector<std::string>` to `span<nostd::string_view>` ? I think `vector<bool>` to `span<bool>` should be implicitly possible.

`vector<bool>` has a special implementation, it is more likely a bitset and the elements in it have no unique address. `span<bool>` need contiguouse addresses for every element to initialize.

lalitb:
> `vector<bool>` has a special implementation, it is more likely a bitset and the elements in it have no unique address. `span<bool>` need contiguous addresses for every element to initialize.

yes agree, I see a problem with conversion for both vector<bool>and vector<string>, don't have a solution for now. Let me think about this. Now I also remember vaguely remember discussing this with @maxgolov during ETW implementation [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/9a6c745781a5c4c8dfa3a55236e77ff976b049a4/exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h#L271).

owent:
@lalitb This PR is almost finished. Could you please review the changes some time later?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -67,6 +70,39 @@ class ElasticSearchRecordable final : public sdk::logs::Recordable
...
+        json_[name][key.data()] = opentelemetry::nostd::get<uint64_t>(value);
+        return;
+      case common::kTypeDouble:
+        json_[name][key.data()] = opentelemetry::nostd::get<double>(value);
+        return;
+      case common::kTypeString:
+        json_[name][key.data()] = opentelemetry::nostd::get<std::string>(value).data();
+        return;
+      default:
+        return;
```
lalitb, line 102: log not-supported warning?

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -75,7 +111,18 @@ class ElasticSearchRecordable final : public sdk::logs::Recordable
...
     // Convert the severity enum to a string
-    json_["severity"] = opentelemetry::logs::SeverityNumToText[static_cast<int>(severity)];
+    int severity_index = static_cast<int>(severity);
+    if (severity_index < 0 ||
+        severity_index >= std::extent<decltype(opentelemetry::logs::SeverityNumToText)>::value)
+    {
+      std::stringstream sout;
+      sout << "Invalid severity(" << severity_index << ")";
+      json_["severity"] = sout.str();
+    }
```
lalitb, line 121: Thanks for fixing this.

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -26,7 +26,7 @@ class Logger final : public opentelemetry::logs::Logger
    * @param logger_provider The logger provider that owns this logger.
    */
   explicit Logger(opentelemetry::nostd::string_view name,
-                  std::shared_ptr<LoggerProvider> logger_provider) noexcept;
+                  std::shared_ptr<LoggerContext> logger_provider) noexcept;
```
lalitb, line 29: variable name as `logger_context`?

lalitb (commented):
thanks for the PR. Changes are nicely done. have added initial comments. It's a large PR, will have another scan on the multi-processor, multi-recordable part.

sdk/include/opentelemetry/sdk/logs/logger.h
owent: Thanks and I will fix this later.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
owent: I just keep the same behaviour as it dose before. Maybe we can create a separate issue to add warning for this exporter?
I don't use `ElasticsearchLogExporter` and I'm not sure whether should we add additional data

api/include/opentelemetry/logs/logger.h
```diff
@@ -57,48 +56,72 @@ class Logger
...
                    trace::TraceId trace_id,
                    trace::SpanId span_id,
                    trace::TraceFlags trace_flags,
                    common::SystemTimestamp timestamp) noexcept = 0;
 
+  OPENTELEMETRY_DEPRECATED_MESSAGE("resource is removed and ignored")
+  void Log(Severity severity,
+           nostd::string_view name,
+           nostd::string_view body,
+           OPENTELEMETRY_MAYBE_UNUSED const common::KeyValueIterable &resource,
```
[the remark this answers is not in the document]
owent: I think it's helpful for users to have some time to upgrade the usage of Log API instead of just `compile error`.

sdk/include/opentelemetry/sdk/logs/logger.h
owent: Fixed

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
lalitb: With logs in experimental, I think breaking changes should be acceptable and users should be prepared for it. This will keep the API much cleaner. We should definitely mention this as a breaking change in the Release description.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
lalitb: Yes, we can do it separately.

sdk/src/logs/multi_log_processor.cc
```diff
@@ -0,0 +1,153 @@
...
+  for (auto &processor : processors_)
+  {
+    if (!processor->ForceFlush(std::chrono::duration_cast<std::chrono::microseconds>(timeout_ns)))
+    {
+      result = false;
+    }
+    start_time = std::chrono::system_clock::now();
+    if (expire_time > start_time)
+    {
+      timeout_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(expire_time - start_time);
```
lalitb, line 99: What are we trying to achieve here - to ensure that the total time taken by all the processors shouldn't exceed `timeout` ? For traces, logic is a bit different, as the same `timeout` is used for all processors. This looks good too to me.

owent: Yes, I think it may confuse ther users if they set timeout to some value, but actually take a more longer time.

E.g. when closing a service process, we may force to flush data, wait a limited time and then force to exit no matter it's done or not.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
owent: OK, I can remove these later, should I keep `api/include/opentelemetry/common/macros.h` ? Maybe it's useful in the future?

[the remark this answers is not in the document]
lalitb: We can remove macros for now, can add them back later if required. Thanks.

sdk/src/logs/multi_log_processor.cc
lalitb: Yes agree.

lalitb (approved):
LGTM once we remove the deprecated functionality.

owent:
> LGTM once we remove the deprecated functionality.

Done. I keep the `OPENTELEMETRY_MAYBE_UNUSED` and it's used in `sdk/include/opentelemetry/sdk/common/empty_attributes.h` . Some compiler will warning unused static function when it's included(maybe included by `recordable.h`) by source file and do not use it.

lalitb:
> > LGTM once we remove the deprecated functionality.
> 
> Done. I keep the `OPENTELEMETRY_MAYBE_UNUSED` and it's used in `sdk/include/opentelemetry/sdk/common/empty_attributes.h` . Some compiler will warning unused static function when it's included(maybe included by `recordable.h`) by source file and do not use it.

Thanks. Probably all unused functionality can be removed separately some other time. Changes look good now. Thanks for working on this.

api/include/opentelemetry/common/macros.h
```diff
@@ -0,0 +1,30 @@
...
+///   class OPENTELEMETRY_MAYBE_UNUSED a;
+///   OPENTELEMETRY_MAYBE_UNUSED int a();
+///
+#if defined(__cplusplus) && __cplusplus >= 201703L
+#  define OPENTELEMETRY_MAYBE_UNUSED [[maybe_unused]]
+#elif defined(__clang__)
+#  define OPENTELEMETRY_MAYBE_UNUSED __attribute__((unused))
+#elif defined(__GNUC__) && ((__GNUC__ >= 4) || ((__GNUC__ == 3) && (__GNUC_MINOR__ >= 1)))
+#  define OPENTELEMETRY_MAYBE_UNUSED __attribute__((unused))
+#elif defined(_MSC_VER) && _MSC_VER >= 1700  // vs 2012 or higher
```
ThomsonTan, line 22: Why check vs2012 here? Seems the only valid case here is `_MSVC_LANG >= 201703L`. Also if this is true, should this be covered by the first check of `__cplusplus >= 201703L`?

owent: Yes, it's duplicated, I have removed it.Thanks.

lalitb:
@ThomsonTan - Is it good to merge?

owent:
Rebase on main again

ThomsonTan (approved):

PR #1155, opened 2021-12-20
ZipkinExporter unit-tests

esigo:
Fixes #1143 (issue)

## Changes

Add unit tests for ZipkinExporter.
For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1160, opened 2021-12-21
Fix leak in Jaeger exporter

bjosv:
During our introduction of Jaeger and using the `JaegerExporter` we found this memory leak.
There is currently no unittest for the `JaegerExporter` in opentelemetry-cpp that could have caught this, but maybe we should add something similar to the ongoing `ZipkinExporter` test (#1155)?

## Changes

The jaeger_span was never destroyed after its release from the JaegerRecordable.
This change let a `std::unique_ptr` handle the destruction when going out of scope, first allowing the`thrift::Span` to be copied to the `span_buffer_`.

### DIFF
--- a/exporters/jaeger/src/thrift_sender.cc
+++ b/exporters/jaeger/src/thrift_sender.cc
@@ -37,11 +37,11 @@ int ThriftSender::Append(std::unique_ptr<JaegerRecordable> &&span) noexcept
     max_span_bytes -= process_bytes_size_;
   }
 
-  thrift::Span &jaeger_span = *span->Span();
-  jaeger_span.__set_tags(span->Tags());
-  jaeger_span.__set_logs(span->Logs());
+  auto jaeger_span = std::unique_ptr<thrift::Span>(span->Span());
+  jaeger_span->__set_tags(span->Tags());
+  jaeger_span->__set_logs(span->Logs());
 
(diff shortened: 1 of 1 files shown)


lalitb (approved):
LGTM. Thanks for the fix. 
Agree, Jaeger exporter tests need to be improved.

esigo (approved):
thanks for the fix.
LGTM.
I will add more tests without mock to cover `Thrift` part.

bjosv:
> I will add more tests without mock to cover `Thrift` part.
Great! Was thinking of looking into this, but just let me know If I can be of any help.

esigo:
We can catch the leak in the new [test](https://github.com/open-telemetry/opentelemetry-cpp/runs/4607890718?check_suite_focus=true#step:5:222).

lalitb:
@bjosv  - can you update the branch to main, to allow this to be merged.

bjosv:
@lalitb  The PR has now been updated to the latest from main.

bjosv:
CI step `Bazel on MacOS` failed with
`ERROR: Analysis of target '//exporters/elasticsearch:es_log_exporter_test' failed; build aborted: Analysis of target '@local_config_cc//:tvos_x86_64' failed`

It doesn't seem to be related to my correction according to the output, but I'm not familiar with the test.

ThomsonTan:
Thanks for catching and fixing.

PR #1163, opened 2021-12-23
fix: otlp http exporter block thread (#1141)

wxl374:
Fixes #1141

## Changes

fix deadlock when notify_all() called before wait().

### DIFF
--- a/exporters/otlp/src/otlp_http_client.cc
+++ b/exporters/otlp/src/otlp_http_client.cc
@@ -100,7 +100,7 @@ public:
 
       // Set the response_received_ flag to true and notify any threads waiting on this result
       response_received_ = true;
-      stop_waiting_ = true;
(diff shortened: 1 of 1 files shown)


exporters/otlp/src/otlp_http_client.cc
```diff
@@ -233,6 +254,9 @@ class ResponseHandler : public http_client::EventHandler
   std::condition_variable cv_;
   std::mutex mutex_;
 
+  // Whether notify has been called
+  bool stop_waiting_ = false;
+
   // Whether the response from Elasticsearch has been received
```
owent, line 260: Could you please alose fix the error message in this comment?

owent:
LGTM

lalitb (approved):
Thanks for the fix. LGTM.

PR #1165, opened 2021-12-28
[WIP] multi-platform scheduled builds

esigo:
This PR provides multi-platform scheduled builds, so we can test build on different CPU architectures e.g. arm64 or OS versions.

example build command (building with [gcc 8.5](https://hub.docker.com/_/gcc)):
`docker build --build-arg BASE_IMAGE=gcc:8.5 -f ../docker/ubuntuLatest/Dockerfile . -t otel-cpp`

## Changes

uses `setup_grpc.sh` and `setup_thrift.sh` scripts in [`Dockerfile`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docker/ubuntuLatest/Dockerfile)in multiple stages so that Docker can build the stages in parallel.
The ARM build can take up to 3 hours because of the emulation.
If the image built with the scheduled job, pushed to some registry, it can be directly used on ARM machines as well.
 As the build using this method is very slow, a `cron` job is used. 


owent:
Should we build our development docker image instead install and build dependencies every time?It may reduce a lot time to all CI jobs.

esigo:
> Should we build our development docker image instead install and build dependencies every time?It may reduce a lot time to all CI jobs.

Thanks, yeah I totally agree. The question is where to host. We discussed this in one of the meetings and there is no plan to host docker images.
The PR relies on caching though, those gRPC and thrift layers will be cached and the subsequent builds will have ok speed.

PR #1167, opened 2022-01-03
importing gsl::span if std::span is not available

### LINKED ISSUE #813: WITH_STL shouldn't try importing gsl::span if std::span is not available.
As discussed in #742:
If the `WITH_STL` cmake option is enabled, the build should be using std::span if available (for C++20), and fall back to `nostd::span` instead of `gsl::span` ( or perhaps fail the build). Use `gsl::span` only if developers explicitly specify the `WITH_GSL` build option.

esigo:
Fixes #813 (issue)

## Changes

If the `WITH_STL` cmake option is enabled, the build will use `std::span` if available (for C++20), and fall back to `nostd::span` instead of `gsl::span` (or perhaps fail the build). Uses `gsl::span` only if explicitly specified the `WITH_GSL` build option.


CMakeLists.txt
```diff
@@ -165,6 +155,20 @@ if(WITH_STL)
   endif()
 endif()
 
+if(WITH_GSL)
+  add_definitions(-DHAVE_GSL)
```
lalitb, line 159: Just to clarify, with this PR GSL library will only be used if the `WITH_GSL` flag is enabled?

esigo: yes it is enabled only when the user explicitly asks for it.

api/include/opentelemetry/std/span.h
```diff
@@ -11,7 +11,7 @@
 #if defined __has_include
 #  if __has_include(<version>)  // Check for __cpp_{feature}
 #    include <version>
-#    if defined(__cpp_lib_span)
+#    if defined(__cpp_lib_span) && __cplusplus > 201703L
```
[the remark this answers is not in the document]
esigo: because  `__cpp_lib_span` is defined for C++17 on g++-10:`g++-10 (Ubuntu 10.3.0-1ubuntu1~20.04) 10.3.0`.
I use these CMake options for build:
`-DCMAKE_CXX_COMPILER=g++-10 -DCMAKE_CXX_STANDARD=17`

CMakeLists.txt
lalitb: Ok, I am concerned seeing this part:
https://github.com/open-telemetry/opentelemetry-cpp/blob/05d43d61ed162eea063cbd4c0ea19728d106b6b3/api/CMakeLists.txt#L48-L51

As it looks we are enabling HAVE_GSL macro with WITH_STL option, which would mean try using GSL from submodule if STL doesn't have std::span implementation.

esigo: that is covered [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/bc1b469c72f12e068af10c73829f1486eb006a2d/api/include/opentelemetry/std/span.h#L29). We shouldn't define it though, I cleaned it from the CMake file.

api/include/opentelemetry/std/span.h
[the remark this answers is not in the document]
lalitb: Which should be ok right? If the `HAVE_CPP_STDLIB` option is specified, and the standard library provides the implementation ( stable or experimental) it should get selected. What is the behavior now with these changes on `g++-10/C++17` - compilation failure as std::span is not available?

CMakeLists.txt
lalitb: ok, thanks for the clarification.

lalitb (approved):
LGTM with a minor suggestion.

api/include/opentelemetry/std/span.h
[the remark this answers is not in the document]
esigo: The behavior for `g++-10/C++17` is that `nostd::span` will be picked if the `WITH_GSL` option was not used. I think we shouldn't use `std::span` if the user explicitly asks for `C++17`.
Shall I enable `std::span` if `C++17` was used and `std::span` is available?

PR #1169, opened 2022-01-03
Fix deadlock for ES

lalitb:
Fixes #1141 

## Changes

Elastic search exporter also suffers from deadlock issue similar to otlp, fix similar to #1163 


exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -100,6 +120,9 @@ class ResponseHandler : public http_client::EventHandler
   std::condition_variable cv_;
   std::mutex mutex_;
 
+  // Whether notify has been called
+  bool stop_waiting_ = false;
```
ThomsonTan, line 124: Does this need to be atomic as it could be read/set from multiple threads?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -73,6 +76,23 @@ class ResponseHandler : public http_client::EventHandler
   // Callback method when an http event occurs
   void OnEvent(http_client::SessionState state, nostd::string_view reason) noexcept override
   {
+    // need to modify stop_waiting_ under lock before calling notify_all
+    switch (state)
+    {
+      case http_client::SessionState::CreateFailed:
+      case http_client::SessionState::ConnectFailed:
```
esigo, line 83: Would it be possible to add the enumeration values which are not handled?
[enumeration value 'CreateFailed' not handled in switch [-Wswitch]](https://github.com/open-telemetry/opentelemetry-cpp/runs/4694407563?check_suite_focus=true#step:5:117)

exporters/elasticsearch/src/es_log_exporter.cc
owent: I think `stop_waiting_` is already protected by `mutex_`, there's no need to declare `stop_waiting_` as atomic.

PR #1174, opened 2022-01-16
Continuous benchmark tests as part of the CI

### LINKED ISSUE #1170: benchmark tests as part of the CI
It is hard to overstate the importance of benchmarking and monitoring the benchmarks over time. This becomes even more important to us as it is expected/desired that telemetry operations should have minimal adverse impact on performance. In order to quantify and measure how well we're doing there, benchmark tests have to be added as part of the CI.

This can be done using [continuous-benchmark](https://github.com/marketplace/actions/continuous-benchmark) where benchmark results can be stored on every merge to the main branch. To avoid noising the main branch, [gh-pages](https://stackoverflow.com/questions/4750520/git-branch-gh-pages) branch can be used to push the benchmark results. This has already been used in [python SIG](https://open-telemetry.github.io/opentelemetry-python/benchmarks/index.html).

Tasks which require privilege:
- Activating githbub-pages
- Creating gh-pages branch

**Additional context**

esigo:
Fixes #1170 (issue)

## Changes

Runs  benchmark tests at every push to the `main` branch and pushes the results to `gh-pages` branch. The results are grouped into api, sdk and exporters.

An example of working graphs can be seen [here](https://esigo.github.io/opentelemetry-cpp/benchmarks/index.html).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -68,6 +68,5 @@ jobs:
           alert-threshold: '200%'
           comment-on-alert: true
           fail-on-alert: true
-          alert-comment-cc-users: '@esigo'
           gh-pages-branch: gh-pages
           benchmark-data-dir-path: benchmarks
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,73 @@
+name: OpenTelemetry-cpp benchmarks
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  deployments: write
+
+jobs:
+  benchmark:
+    name: Run OpenTelemetry-cpp benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+      - name: Mount Bazel Cache
+        uses: actions/cache@v2
+        env:
+          cache-name: bazel_cache
+        with:
+          path: /home/runner/.cache/bazel
+          key: bazel_benchmark
+      - name: setup
+        run: |
+          sudo ./ci/setup_cmake.sh
+          sudo ./ci/setup_ci_environment.sh
+      - name: Run benchmark
+        id: run_benchmarks
+        run: |
+          ci/do_ci.sh bazel.benchmark
+          mkdir -p benchmarks
+          mv api-benchmark_result.json benchmarks
+          mv sdk-benchmark_result.json benchmarks
+          mv exporters-benchmark_result.json benchmarks
+      - uses: actions/upload-artifact@master
+        with:
+          name: benchmark_results
+          path: benchmarks
+  store_benchmark:
+    needs: benchmark
+    strategy:
+      matrix:
+        components: ["api", "sdk", "exporters"]
+    name: Store benchmark result
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@master
+        with:
+          name: benchmark_results
+          path: benchmarks
+      - name: Print json files
+        id: print_json
+        run: |
+          cat benchmarks/*
+      - name: Push benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: OpenTelemetry-cpp ${{ matrix.components }} Benchmark
+          tool: 'googlecpp'
+          output-file-path: benchmarks/${{ matrix.components }}-benchmark_result.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '200%'
+          comment-on-alert: true
+          fail-on-alert: true
+          alert-comment-cc-users: '@esigo'
+          gh-pages-branch: gh-pages
+          benchmark-data-dir-path: benchmarks
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -164,6 +164,7 @@ elif [[ "$1" == "cmake.exporter.otprotocol.test" ]]; then
 elif [[ "$1" == "bazel.with_abseil" ]]; then
   bazel $BAZEL_STARTUP_OPTIONS build $BAZEL_OPTIONS --//api:with_abseil=true //...
   bazel $BAZEL_STARTUP_OPTIONS test $BAZEL_TEST_OPTIONS --//api:with_abseil=true //...
+  exit 0
 elif [[ "$1" == "cmake.test_example_plugin" ]]; then
   # Build the plugin
   cd "${BUILD_DIR}"
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -17,6 +17,43 @@ function install_prometheus_cpp_client
   popd
 }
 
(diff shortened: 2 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Push benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          name: OpenTelemetry-cpp ${{ matrix.components }} Benchmark
+          tool: 'googlecpp'
+          output-file-path: benchmarks/${{ matrix.components }}-benchmark_result.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          auto-push: true
+          # Show alert with commit comment on detecting possible performance regression
+          alert-threshold: '200%'
```
lalitb, line 68: This is the threshold of comparison with the previous result from the main branch? Trying to understand the flow here.

ci/do_ci.sh
```diff
@@ -17,6 +17,43 @@ function install_prometheus_cpp_client
...
+    cd bazel-bin
+    find . -name \*_result.json -exec bash -c \
+      'echo "$@" && mkdir -p "$BENCHMARK_DIR/$(dirname "$@")" && \
+       cp "$@" "$BENCHMARK_DIR/$@" && chmod +w "$BENCHMARK_DIR/$@"' _ {} \;
+  )
+
+  # collect benchmark results into one array
+  components=(api sdk exporters)
+  pushd $BENCHMARK_DIR
+  components=(api sdk exporters)
```
lalitb, line 42: nit - this is already defined.

exporters/otlp/test/otlp_grpc_exporter_benchmark.cc
```diff
@@ -4,6 +4,23 @@
 #include "opentelemetry/exporters/otlp/otlp_grpc_exporter.h"
 #include "opentelemetry/exporters/otlp/otlp_recordable.h"
 
+#include <benchmark/benchmark.h>
+#include "opentelemetry/exporters/otlp/otlp_grpc_exporter.h"
```
lalitb, line 8: nit - this is included above.

.github/workflows/benchmark.yml
esigo: yes there will be an alert similar to [this](https://github.com/esigo/opentelemetry-cpp/commit/6b3c8ec0c9b2baee058a0c738866c729a6f090e6#commitcomment-63913248), when there is an slowdown bigger than `200%`.
Smaller threshold could be an issue, as the machines used for CI are not always the same.

ci/do_ci.sh
esigo: thanks, cleaned.

exporters/otlp/test/otlp_grpc_exporter_benchmark.cc
esigo: thanks, cleaned.

.github/workflows/benchmark.yml
lalitb: Ok, will `fail-on-alert` block PR merge if the threshold is higher? If yes, is it possible to let CI job fail, but allow merge on case to case basis?

esigo: > Ok, will `fail-on-alert` block PR merge if the threshold is higher? If yes, is it possible to let CI job fail, but allow merge on case to case basis?

No the merge will go through, only the job will fail with some alert as a comment on the commit.

exporters/otlp/test/otlp_grpc_exporter_benchmark.cc
```diff
@@ -171,4 +187,30 @@ BENCHMARK(BM_OtlpExporterDenseSpans);
...
+  auto exporter  = std::unique_ptr<trace_sdk::SpanExporter>(new otlp::OtlpGrpcExporter(opts));
+  auto processor = std::unique_ptr<trace_sdk::SpanProcessor>(
+      new trace_sdk::SimpleSpanProcessor(std::move(exporter)));
+  auto provider =
+      nostd::shared_ptr<trace::TracerProvider>(new trace_sdk::TracerProvider(std::move(processor)));
+  // Set the global trace provider
+  trace::Provider::SetTracerProvider(provider);
+}
+
+void BM_otlp_grpc_with_collector(benchmark::State &state)
```
lalitb, line 205: Do we need to measure benchmark with the actual collector, or it would be sufficient to have results using faking the service stub, as we are more interested in the stats resulting from the otel-cpp code?

esigo: I think it's a good-to-have number for the users.
Can we have both with and without the actual collector?

lalitb: I would ideally like to keep the stats with an actual collector. We have lately seen CI failures because of transient network timeout issues, I am just concerned if testing with real collector instance shouldn't add to that. Also, whether adding docker instances to the VM consume more resources and slows the CI jobs. And we are spawning another docker instance for `jq` parsing. We can keep them if you don't see any such slowness in CI with multiple iterations.

esigo: The test with the collector was pretty stable in my test CI. This can't be guaranteed to be the case always though. The job will be executed only when we merge a commit to the main branch, so we will see the issues if any only after merge to main.
This can't be part of checks for RPs as it will be noisy. 
Shall we keep it as it is, in case we got failures, I can raise a PR to use mock.

lalitb: > Shall we keep it as it is, in case we got failures, I can raise a PR to use mock.

Should be fine for me. Let's wait for suggestions from @ThomsonTan too.

ci/do_ci.sh
```diff
@@ -17,6 +17,42 @@ function install_prometheus_cpp_client
...
+  )
+
+  # collect benchmark results into one array
+  pushd $BENCHMARK_DIR
+  components=(api sdk exporters)
+  for component in "${components[@]}"
+  do
+    out=$component-benchmark_result.json
+    find ./$component -type f -name "*_result.json" -exec cat {} \; > $component_tmp_bench.json
+    cat $component_tmp_bench.json | docker run -i --rm itchyny/gojq:0.12.6 -s \
```
lalitb, line 46: nit - one minor comment, if it's not a major rework, can we use [jq](https://stedolan.github.io/jq/download/) instead of `gojq`, it's more lightweight without any external dependencies.

lalitb (approved):

ci/do_ci.sh
esigo: I've tried jq, it was showing different behavior on my local (Ubuntu 20.4) and on the GHA vm (Ubuntu latest). Collecting the benchmark results of all the tests into one array didn't work on the jq of GHA vm. So I switched to gojq which has a docker image.

exporters/otlp/test/otlp_grpc_exporter_benchmark.cc
ThomsonTan: Agree with @lalitb , I prefer to use mock to avoid that the result could be affected by environment workload and network traffic, but we could do this in later PRs.

ThomsonTan (approved):

PR #1177, opened 2022-01-18
Fix build test of Zipkin exporter on Windows

ThomsonTan:
## Changes

Zipkin exporter doesn't build on Windows but Jaeger exporter builds fine. This PR ports the necessary Jaeger build scripts to Zipkin exporter as well.

### DIFF
--- a/exporters/zipkin/CMakeLists.txt
+++ b/exporters/zipkin/CMakeLists.txt
@@ -37,6 +37,8 @@ install(
   PATTERN "recordable.h" EXCLUDE)
 
 if(BUILD_TESTING)
+  add_definitions(-DGTEST_LINKED_AS_SHARED_LIBRARY=1)
+
   add_executable(zipkin_recordable_test test/zipkin_recordable_test.cc)
 
   target_link_libraries(
--- a/exporters/zipkin/test/zipkin_exporter_test.cc
+++ b/exporters/zipkin/test/zipkin_exporter_test.cc
@@ -12,7 +12,7 @@
 #  include "opentelemetry/trace/provider.h"
 
 #  include <gtest/gtest.h>
-#  include <gmock/gmock.h>
+#  include "gmock/gmock.h"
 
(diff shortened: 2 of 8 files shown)


exporters/zipkin/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ install(
   PATTERN "recordable.h" EXCLUDE)
 
 if(BUILD_TESTING)
+  add_definitions(-DGTEST_LINKED_AS_SHARED_LIBRARY=1)
```
lalitb, line 40: So in windows, we only have gtest dll?

exporters/zipkin/test/zipkin_exporter_test.cc
```diff
@@ -3,7 +3,6 @@
 
 #ifndef HAVE_CPP_STDLIB
 
-#  define _WINSOCKAPI_  // stops including winsock.h
```
esigo, line 6: Not related to the PR:
This is not part of CMake test for Windows CI.

exporters/zipkin/CMakeLists.txt
ThomsonTan: There is some unresolved global lock variable in gtest/gmock reported if built without this flag.

exporters/zipkin/test/zipkin_exporter_test.cc
ThomsonTan: I moved this `define` to spin lock header which includes `windows.h` as well. So this should be unnecessary. Let me know you prefer keep this define here.

lalitb (approved):

exporters/zipkin/test/zipkin_exporter_test.cc
esigo: thanks, spin lock header is actually the right place to have the define.

esigo (approved):

PR #1173, opened 2022-01-14
[Not For Review] Metrics Aggregation SDK

lalitb:
not to be reviewed Splitting this draft PR into smaller PRs for ease of review.
Fixes # (issue)

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1178, opened 2022-01-21
Add Aggregation as part of metrics SDK.

### LINKED ISSUE #1090: Metrics SDK: Add Aggregation

lalitb:
Fixes #1090

## Changes
Splitting #1173 into smaller PRs for ease of review. This PR adds aggregation for Sum, LastValue, and Histogram.

Implementation is based on the specs here:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#aggregation


sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation.h
```diff
@@ -0,0 +1,37 @@
...
+private:
+  bool is_monotonic_;
+};
+
+class Aggregation
+{
+public:
+  virtual void Aggregate(long value, const PointAttributes &attributes = {}) noexcept = 0;
+
+  virtual void Aggregate(double value, const PointAttributes &attributes = {}) noexcept = 0;
```
lalitb, line 29: Preferred overloaded methods over the template ( for double and long type) to avoid making all header-only libraries. The template use had a cascading effect here, forcing to make all inter-related classes as templates. We can discuss if you see any issues here.

sdk/test/metrics/BUILD
```diff
@@ -9,3 +9,27 @@ cc_test(
         "@com_google_googletest//:gtest_main",
     ],
 )
+
+cc_test(
+    name = "view_registry_test",
+    srcs = [
+        "view_registry_test.cc",
+    ],
```
esigo, line 17: Would it be possible to add tags here?
```suggestion
    ],
    tags = [
        "metrics",
        "test",
    ],
```

sdk/test/metrics/BUILD
```diff
@@ -9,3 +9,27 @@ cc_test(
...
+        "//sdk/src/resource",
+        "@com_google_googletest//:gtest_main",
+    ],
+)
+
+cc_test(
+    name = "aggregation_test",
+    srcs = [
+        "aggregation_test.cc",
+    ],
```
esigo, line 29:
```suggestion
    ],
    tags = [
        "metrics",
        "test",
    ],
```

sdk/src/metrics/aggregation/lastvalue_aggregation.cc
```diff
@@ -0,0 +1,59 @@
...
+void LongLastValueAggregation::Aggregate(long value, const PointAttributes &attributes) noexcept
+{
+  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
+  is_lastvalue_valid_ = true;
+  value_              = value;
+}
+
+PointType LongLastValueAggregation::Collect() noexcept
+{
+  if (!is_lastvalue_valid_)
```
esigo, line 28: Don't we need to protect `is_lastvalue_valid_` here?

lalitb: Good point. I thought it should be fine to return an invalid value in given Collect() call, as long as the subsequent call would return a valid value. Worst case, we may miss a given metric in a series of Collect() calls, but this way we can avoid a lock.

sdk/test/metrics/aggregation_test.cc
```diff
@@ -0,0 +1,122 @@
...
+{
+  LongHistogramAggregation aggr;
+  auto data = aggr.Collect();
+  ASSERT_TRUE(nostd::holds_alternative<HistogramPointData>(data));
+  auto histogram_data = nostd::get<HistogramPointData>(data);
+  ASSERT_TRUE(nostd::holds_alternative<long>(histogram_data.sum_));
+  ASSERT_TRUE(nostd::holds_alternative<std::vector<long>>(histogram_data.boundaries_));
+  EXPECT_EQ(nostd::get<long>(histogram_data.sum_), 0);
+  EXPECT_EQ(histogram_data.count_, 0);
+  EXPECT_NO_THROW(aggr.Aggregate(12l, {}));   // lies in third bucket (indexed at 0)
```
esigo, line 82: The comments are not correct. also in `DoubleHistogramAggregation`

sdk/src/metrics/aggregation/lastvalue_aggregation.cc
esigo: I'm afraid this would be a UB, and the lock for reading is unavoidable.

lalitb: I thought UB should be fine as long as  `is_lastvalue_valid_` returns either true/false (was not expecting strong consistency here). As in one of the next few iterations, it would be true eventually. But have added the lock if it's causing confusion.

sdk/test/metrics/BUILD
lalitb: Thanks. Added it.

sdk/test/metrics/BUILD
lalitb: Added.

sdk/test/metrics/aggregation_test.cc
lalitb: Do you mean 12 does lie in 3rd bucket? The default buckets here are: [-infinity to 0], [0 to 5], [5 to10], [10 to 25] , and so on (refer - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#histogram-aggregation ). So 12 will lie in the third bucket.

esigo: No I mean `indexed at 0` part, this should be 2 or I'm getting sth wrong here?

lalitb: Yes, have only kept the bucket number now. Thanks for noticing it.

esigo (approved):
Thanks for the PR. LGTM

esigo (approved):

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation.h
```diff
@@ -0,0 +1,37 @@
...
+#ifndef ENABLE_METRICS_PREVIEW
+#  include "opentelemetry/nostd/string_view.h"
+#  include "opentelemetry/sdk/metrics/data/metric_data.h"
+#  include "opentelemetry/sdk/metrics/data/point_data.h"
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace metrics
+{
+class InstrumentMonotonicityAwareAggregation
```
ThomsonTan, line 14: Seems `InstrumentMonotonicityAwareAggregation` should be a subclass of `Aggretation`?

lalitb: Yes or else we can also move this functionality within the Aggregation class. I thought of keeping it a separate class as not sure right now how to use it for Asynchronous aggregation. We can refactor this later once we have more complete implementation.

PR #1184, opened 2022-01-28
Sync and Async Instruments SDK

lalitb:
## Changes

This PR adds below classes for metrics SDK:
 - synch instruments implementation
 -  async instruments implementation
 - Measurement Processor - This is analogous to Span Processor for traces, and all the measurements flow through this.
 - Metric Storage - To store the metrics for the readers. These would call the aggregation API ( #1178) to do the actual aggregation, to generate metrics data.

Some of these classes are placeholders and would be enriched once more flow gets added.


sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -0,0 +1,68 @@
...
+                    const opentelemetry::common::KeyValueIterable &attributes) noexcept override
+  {
+    // TBD
+    // aggregation->recordDouble(value, attributes);
+  }
+
+  bool Collect(AggregationTemporarily aggregation_temporarily,
+               nostd::function_ref<bool(MetricData)> callback) noexcept override
+  {
+    if (callback(MetricData()))
```
esigo, line 57: nit: can be simplified to 
```cpp
    return callback(metric_data);
```

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -36,6 +36,19 @@ struct InstrumentDescriptor
...
+{
+  // TBD - This enum to be removed once #1178 is merged.
+  kUnspecified,
+  kDelta,
+  kCummulative
+};
+
+class MetricData
+{
+  // TBD - This class to be removed once #1178 is merged
```
esigo, line 49: #1178 is merged.

sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -94,13 +95,16 @@ class Meter final : public opentelemetry::metrics::Meter
       nostd::string_view unit        = "") noexcept override;
 
   /** Returns the associated instruementation library */
-  const sdk::instrumentationlibrary::InstrumentationLibrary &GetInstrumentationLibrary()
+  const sdk::instrumentationlibrary::InstrumentationLibrary *GetInstrumentationLibrary()
       const noexcept;
 
+  /** Returns the associated instruementation library */
```
esigo, line 101: comment has to be updated.

sdk/include/opentelemetry/sdk/metrics/async_instruments.h
```diff
@@ -0,0 +1,176 @@
...
+      : name_(name),
+        instrumentation_library_{instrumentation_library},
+        measurement_processor_{measurement_processor},
+        callback_(callback),
+        description_(description),
+        unit_(unit)
+  {}
+
+protected:
+  std::string name_;
```
esigo, line 36: Why don't we use `nostd::string_view` for the members?

sdk/include/opentelemetry/sdk/metrics/measurement_processor.h
```diff
@@ -0,0 +1,108 @@
...
+
+#  include <map>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace metrics
+{
+
+static std::size_t MakeKey(const MetricReader &metric_reader)
```
esigo, line 20: Do we need this function introduced to the namespace?

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -4,33 +4,73 @@
...
 class MetricReader
 {
 public:
-  virtual ~MetricReader() = default;
+  MetricReader(
+      AggregationTemporarily aggregation_temporarily = AggregationTemporarily::kCummulative)
+      : aggregation_temporarily_(aggregation_temporarily), measurement_processor_callback_({})
+  {}
+
+  virtual ~MetricReader() { measurement_processor_callback_ = {}; }
```
esigo, line 28: Why do we need this?

sdk/src/metrics/meter.cc
```diff
@@ -23,25 +25,24 @@ Meter::Meter(std::shared_ptr<MeterContext> context,
              std::unique_ptr<sdk::instrumentationlibrary::InstrumentationLibrary>
                  instrumentation_library) noexcept
     : context_{context}, instrumentation_library_{std::move(instrumentation_library)}
+
```
esigo, line 28: nit: new line

sdk/test/metrics/async_instruments_test.cc
```diff
@@ -0,0 +1,22 @@
...
+#  include "opentelemetry/sdk/instrumentationlibrary/instrumentation_library.h"
+#  include "opentelemetry/sdk/metrics/measurement_processor.h"
+
+#  include <gtest/gtest.h>
+
+using namespace opentelemetry;
+using namespace opentelemetry::sdk::instrumentationlibrary;
+using namespace opentelemetry::sdk::metrics;
+
+std::string library_name     = "opentelemetry-cpp";
```
esigo, line 15: why do we need these variables?

sdk/include/opentelemetry/sdk/metrics/async_instruments.h
lalitb: It's better to own these instrument properties. The default ownership of these string_view arguments passed in the constructor is with the instrumentation library, and it may choose to deallocate them anytime.

sdk/include/opentelemetry/sdk/metrics/measurement_processor.h
lalitb: Wanted to avoid a global static function in the default namespace, and since it calculates hash on MetricReader pointer, thought it is better to make it part of opentelemetry::metrics namespace. Let me know if it doesn't look correct?

sdk/include/opentelemetry/sdk/metrics/instruments.h
lalitb: thanks, removed now.

sdk/include/opentelemetry/sdk/metrics/meter.h
lalitb: thanks. updated.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
lalitb: Good catch. We don't need it. Removed now.

sdk/src/metrics/meter.cc
lalitb: fixed.

sdk/test/metrics/async_instruments_test.cc
lalitb: No, removed it now. thanks.

sdk/include/opentelemetry/sdk/metrics/async_instruments.h
esigo: thanks for the clarification.

esigo (approved):
Thanks for the PR. LGTM

sdk/include/opentelemetry/sdk/metrics/async_instruments.h
ThomsonTan: Could the ownership be moved here from the InstrumentationLibrary? Just want to check whether it is possible to avoid the copy here.

lalitb: Yes, it is possible by changing the API for creating metric (with passing `std::string &&` instead of `nostd::string_view`) which would be  ABI breaking change, as the implementation of `std::string` is not ABI compatible.
The meter is created only during SDK configuration and we store these three meter attributes (`name`, `unit`, and `description`) during that time. It is not allocated in the hot path for metric creation and collection.

Issue #1146, opened 2021-12-16
Evaluate removing curl dependency for Apple platform - Adding support for CFNetwork API


lalitb:
Apple comes with CFNetwork API to enable applications to interact with FTP and HTTP protocols to access Internet resources. This means it doesn't explicitly need to use curl as an HTTP client library.

otel-cpp is designed to plugin any external client library by extending abstract classes defined in the ext::http::client namespace.
And HttpClientFactory factory is responsible to pick up the correct library based on the platform it is running.

CURL implementation is available here: https://github.com/open-telemetry/opentelemetry-cpp/tree/main/ext/include/opentelemetry/ext/http/client/curl
Similar implementation needs to be provided to support CFNetwork API.

This issue is created to evaluate whether CFNetwork can be supported, and probably raise another issue for the implementation.

Issue #1145, opened 2021-12-16
Remove curl dependency for Windows - Adding support for WinINet API


lalitb:
Windows comes with WinINet API to enable applications to interact with FTP and HTTP protocols to access Internet resources. This means it doesn't explicitly need to use curl as an HTTP client library.

otel-cpp is designed to plugin any external client library by extending abstract classes defined in the [ext::http::client](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/include/opentelemetry/ext/http/client/http_client.h) namespace.
And [HttpClientFactory](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/src/http/client/curl/http_client_factory_curl.cc) factory is responsible to pick up the correct library based on the platform it is running.

CURL implementation is available here: https://github.com/open-telemetry/opentelemetry-cpp/tree/main/ext/include/opentelemetry/ext/http/client/curl
Similar implementation needs to be provided to support WinINet API.

PR #1185, opened 2022-01-30
Mock for http exporters

### LINKED ISSUE #1049: Unit test failing in CI: OtlpHttpExporterTestPeer.ExportJsonIntegrationTest
One of the failure instances:
https://github.com/open-telemetry/opentelemetry-cpp/runs/4083453037?check_suite_focus=true

```
[ RUN      ] OtlpHttpExporterTestPeer.ExportJsonIntegrationTest
exporters/otlp/test/otlp_http_exporter_test.cc:240: Failure
Value of: waitForRequests(2, old_count + 1)
  Actual: false
Expected: true
[  FAILED  ] OtlpHttpExporterTestPeer.ExportJsonIntegrationTest (4446 ms)
```

esigo:
Fixes #1049 (issue)

## Changes
removes `http` server from exporter tests, and uses mock for the clients.


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -98,14 +104,16 @@ class OtlpHttpClient
...
+#endif
 
   /**
    * Export
    * @param message message to export, it should be ExportTraceServiceRequest,
    * ExportMetricsServiceRequest or ExportLogsServiceRequest
    */
-  sdk::common::ExportResult Export(const google::protobuf::Message &message) noexcept;
-
+  VIRTUAL_TEST sdk::common::ExportResult Export(const google::protobuf::Message &message) noexcept;
```
lalitb, line 116: `googletest` suggests to mock the non-virtual function using template - https://chromium.googlesource.com/external/github.com/google/googletest/+/refs/tags/release-1.8.0/googlemock/docs/CookBook.md#mocking-nonvirtual-methods, we can avoid using ENABLE_TEST macro in that case.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -98,14 +104,16 @@ class OtlpHttpClient
    * Create an OtlpHttpClient using the given options.
    */
   explicit OtlpHttpClient(OtlpHttpClientOptions &&options);
+#ifdef ENABLE_TEST
+  VIRTUAL_TEST ~OtlpHttpClient() {}
```
lalitb, line 108: I am thinking if mocking `ext::http::client::HttpClient` would be a better approach (the way we do it for Zipkin) as this will allow us to test most of the existing code in OtlpHttpClient::Export() method. We may have to add ext::http::client::HttpClient object as dependency injection to OtlpHttpClient to make this feasible.  What do you think?

esigo: I didn't use template because I thought it will be too much of noise in code.
Thanks for the idea, I will use dependency injection.

exporters/otlp/test/otlp_http_exporter_test.cc
```diff
@@ -11,12 +11,14 @@
 
 #  include "opentelemetry/exporters/otlp/protobuf_include_suffix.h"
 
+#  include "opentelemetry/ext/http/client/curl/http_client_curl.h"
```
lalitb, line 14: Sorry for responding late. The changes look good in general. I am just thinking if there is an alternate way of mocking the http_client library without directly referring to `ext::http::client::curl` in test code. Once we start supporting more HTTP libraries (#1146 & #1145), and if the libcurl is missing (say for windows if WinINet is used), the build for tests will fail with unresolved symbols for curl. As of now, we don't directly refer to http::client::curl in any of the exporters, but bring the right client through [HttpClientFactory](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/include/opentelemetry/ext/http/client/http_client_factory.h). It would be good if we can continue the same way in tests too. Though I don't see any easy way of avoiding it ( maybe bringing a curl mock through HttpClientFactoryMock or something). Sorry if I am confusing here, let me think more and add more details.

esigo: good point. so `HttpClientFactoryMock` which creates a `NosendHttpClient` with `MockSession` as its session and no inheritance from the curl client?

lalitb: Yes that's what I was thinking.

ext/include/opentelemetry/ext/http/client/nosend/http_client_nosend.h
```diff
@@ -0,0 +1,184 @@
...
+    host_ = scheme + "://" + host + ":" + std::to_string(port) + "/";
+  }
+
+  std::shared_ptr<opentelemetry::ext::http::client::Request> CreateRequest() noexcept override
+  {
+    http_request_.reset(new Request());
+    return http_request_;
+  }
+
+  MOCK_METHOD(void,
```
lalitb, line 122: Do we need to mock the SendRequest(), or can we create a empty (200 OK) response, and call the callback from same thread. Something like this 
```cpp
auto response   = std::unique_ptr<Response>(new Response());
.. populate response object..
callback.OnResponse(response);
```

This will allow us to remove dependency of gtest/gmock from this target?

lalitb (approved):
Thanks for the PR. Changes look good, apart from minor issues to address
 - Remove mock dependency for nosend HTTP interface.
 - Small change in HTTP factory implementation (discussing offline)

ext/include/opentelemetry/ext/http/client/nosend/http_client_nosend.h
esigo: I have used Mock for testing the request that is going to be sent. The other option would be to make the client test itself via:
```cpp
std::function<void(http_client::nosend::Session *)> CreateJsonTester(
    const std::string report_trace_id)
{
  return [report_trace_id](http_client::nosend::Session *mock_session) {
    auto check_json    = nlohmann::json::parse(mock_session->GetRequest()->body_, nullptr, false);
    // extract received_trace_id from request
    EXPECT_EQ(received_trace_id, report_trace_id);
  };
}
```

Then store this function into session, so whenever there is a request going to be sent the session tests itself:
```cpp
  void SendRequest(opentelemetry::ext::http::client::EventHandler &callback) noexcept override
  {
    if (test_input_)
    {
      test_input_(this);
    }
    // let the otlp_http_client to continue
    Response response;
    callback.OnResponse(response);
  }
```
Shall I keep gMock or let the session test itself?

lalitb: Thank. Let's keep the gMock for now, we can fix it later.

I am just thinking another solution to change signature for [EventHandler::OnEvent](https://github.com/open-telemetry/opentelemetry-cpp/blob/ef3a623a3bedd3fded167bade625d7fa392d2b88/ext/include/opentelemetry/ext/http/client/http_client.h#L203) to something like:

`virtual void OnEvent(SessionState, void* data = nullptr, size_t size = 0)) noexcept = 0;`

The method in its current form is not usable. Sending back client implementation-specific `data` would be usable in real-world scenarios like server certificate validation during SSL Handshake. And NoSend client can then use this to get the Request object to be tested, something like

```
void SendRequest(EventHandler &callback) noexcept override
{
    auto response   = std::unique_ptr<Response>(new Response());
    callback.OnEvent(EventState::Connected, (void*)http_request_.get(), size);    
      .. populate response object..
    callback.OnResponse(response);
}
```

But don't want to delay this PR more, let's have this fix separately. We can have a github ticket to add this support / or probably remove mock from nosend client.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
```diff
@@ -98,7 +98,15 @@ class OtlpHttpExporter final : public opentelemetry::sdk::trace::SpanExporter
   const OtlpHttpExporterOptions options_;
 
   // Object that stores the HTTP sessions that have been created
-  OtlpHttpClient http_client_;
+  std::unique_ptr<OtlpHttpClient> http_client_;
+  // For testing
+  friend class OtlpHttpExporterTestPeer;
```
ThomsonTan, line 103: Duplicated with line 95?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_exporter.h
```diff
@@ -97,7 +97,15 @@ class OtlpHttpLogExporter final : public opentelemetry::sdk::logs::LogExporter
   const OtlpHttpLogExporterOptions options_;
 
   // Object that stores the HTTP sessions that have been created
-  OtlpHttpClient http_client_;
+  std::unique_ptr<OtlpHttpClient> http_client_;
+  // For testing
+  friend class OtlpHttpLogExporterTestPeer;
```
ThomsonTan, line 102: Duplicated with line 94.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h
esigo: thanks, cleaned.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_exporter.h
esigo: thanks, cleaned

PR #1191, opened 2022-02-02
Metrics SDK: Filtering metrics attributes

### LINKED ISSUE #1190: Metrics SDK: Attributes filtering through View.
As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view), Views should be able to configure the filtering of attributes reported on metrics, including removing all the attributes.

lalitb:
## Fixes 
#1190 

## Changes
As per the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view), Views should be able to configure the filtering of attributes reported on metrics, including removing all the attributes. This PR adds a filter attribute-processor for this. The list of attributes to be included in the metrics can be specified through this filter.


sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -11,13 +11,25 @@ namespace metrics
 {
 using MetricAttributes = opentelemetry::sdk::common::AttributeMap;
 
+/**
+ * The AttributesProcessor is responsible for customizing which
```
reyang, line 15: I wonder if this can be made internal/private to the SDK for now.

The specification only allows very limited manipulation on attributes (e.g. remove certain attributes from the measurement), this generic solution could work but it comes with perf cost (e.g. the virtual `AttributesProcessor::process` call).

Doesn't have to be a blocker though.

reyang:
Doesn't have to be in this PR, we need to understand the performance characteristics (e.g. heap allocation, memory fragmentation, contention if there is any, etc.) in order to guide the SDK design/implementation.

Here are some general guidelines regarding metrics memory usage https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#memory-management.

Consider two types of performance tests:
1. benchmark - which tells us when we call instrument.Add/Record, what's the heap allocation in bytes and CPU cycles for each call
2. stress test - while running the SDK and utilize all the CPU cores to emit as much as of measurements to the SDK (and probably scrape from the pull exporter at the same time), what's the total amount of memory usage, do we see stable calls/second, what's the scalability on SMP when the number of cores grow, etc.

reyang (approved):
I think it's fine to unblock this so we can get a working prototype end-to-end.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
lalitb: Yes, I do suspect some perf cost associated with virtual and attributes copy. I have added a benchmark test to get cost statistics and would be helpful if have to consider performance improvement.

Also, I realized that we need to calculate the hash on the attribute map to store metrics data. Have added functionality to calculate the hash based on `std::hash` and boost::hash_combine`. As hash should be the same irrespective of the order of key/values in the AttributeMap, I have changed the internal storage for AttributeMap from unordered_map to map. This way we can avoid sorting of keys for every measurement.

Let me see how can we make this class `internal/private` to SDK, as C++ doesn't have a direct way of doing it. We can make `AttributeProcessor::process()` as private, and other SDK classes as its friend to it but would like to see a better approach if possible.

lalitb:
> Doesn't have to be in this PR, we need to understand the performance characteristics (e.g. heap allocation, memory fragmentation, contention if there is any, etc.) in order to guide the SDK design/implementation.
> 
> Here are some general guidelines regarding metrics memory usage https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#memory-management.
> 
> Consider two types of performance tests:
> 
> 1. benchmark - which tells us when we call instrument.Add/Record, what's the heap allocation in bytes and CPU cycles for each call, and what's the contribution from each component (provider, processor/reader, exporter)
> 2. stress test - while running the SDK and utilize all the CPU cores to emit as much as of measurements to the SDK (and probably scrape from the pull exporter at the same time), what's the total amount of memory usage, do we see stable calls/second, what's the scalability on SMP when the number of cores grow, etc.

Thanks, this is useful. We will discuss these performance attributes in our community meeting and will come up with plan to measure them in a better way.

PR #1209, opened 2022-02-13
Concurrency HTTP sessions for `OtlpHttpClient`

### LINKED ISSUE #1176: Do not `waitForResponse` in `OtlpHttpClient::Export` ?

Mentioned in #1175 . 

Do not `waitForResponse` in `OtlpHttpClient::Export` and destroy session only after `ResponseHandler` got some events.
We also need control parallel thread spawned by `HttpOperation::SendAsync` here.

This can also assign to me.

owent:
Fixes #1176 

## Changes

+ Allow multiple running http sessions in `OtlpHttpClient`
+ Add `concurrency_sessions` to control the max concurrency http sessions.

### DIFF
--- a/api/include/opentelemetry/common/timestamp.h
+++ b/api/include/opentelemetry/common/timestamp.h
@@ -169,5 +169,39 @@ public:
 private:
   int64_t nanos_since_epoch_;
 };
+
+class DurationUtil
+{
+public:
+  template <class Rep, class Period>
+  static std::chrono::duration<Rep, Period> AdjustWaitForTimeout(
+      std::chrono::duration<Rep, Period> timeout,
+      std::chrono::duration<Rep, Period> indefinite_value) noexcept
+  {
+    // Do not call now() when this duration is max value, now() may have a expensive cost.
+    if (timeout == std::chrono::duration<Rep, Period>::max())
+    {
+      return indefinite_value;
+    }
+
+    // std::future<T>::wait_for, std::this_thread::sleep_for, and std::condition_variable::wait_for
+    // may use steady_clock or system_clock.We need make sure now() + timeout do not overflow.
+    auto max_timeout = std::chrono::duration_cast<std::chrono::duration<Rep, Period>>(
+        std::chrono::steady_clock::time_point::max() - std::chrono::steady_clock::now());
+    if (timeout >= max_timeout)
+    {
+      return indefinite_value;
+    }
+    max_timeout = std::chrono::duration_cast<std::chrono::duration<Rep, Period>>(
+        std::chrono::system_clock::time_point::max() - std::chrono::system_clock::now());
+    if (timeout >= max_timeout)
+    {
+      return indefinite_value;
+    }
+
+    return timeout;
+  }
+};
+
 }  // namespace common
 OPENTELEMETRY_END_NAMESPACE
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
@@ -11,6 +11,7 @@
 
 #include "opentelemetry/common/spin_lock_mutex.h"
 #include "opentelemetry/ext/http/client/http_client.h"
+#include "opentelemetry/nostd/variant.h"
 #include "opentelemetry/sdk/common/exporter_utils.h"
 
 #include "opentelemetry/exporters/otlp/otlp_environment.h"
@@ -19,6 +20,7 @@
 #include <chrono>
 #include <condition_variable>
 #include <cstddef>
+#include <functional>
 #include <list>
 #include <memory>
 #include <mutex>
@@ -112,11 +114,21 @@ public:
   ~OtlpHttpClient();
 
   /**
-   * Export
+   * Sync export
    * @param message message to export, it should be ExportTraceServiceRequest,
    * ExportMetricsServiceRequest or ExportLogsServiceRequest
    */
   sdk::common::ExportResult Export(const google::protobuf::Message &message) noexcept;
+
+  /**
+   * Async export
+   * @param message message to export, it should be ExportTraceServiceRequest,
+   * ExportMetricsServiceRequest or ExportLogsServiceRequest
+   * @param result_callback callback to call when the exporting is done
+   */
+  void Export(
+      const google::protobuf::Message &message,
+      std::function<bool(opentelemetry::sdk::common::ExportResult)> &&result_callback) noexcept;
 
   /**
    * Shut down the HTTP client.
@@ -134,14 +146,49 @@ public:
   void ReleaseSession(const opentelemetry::ext::http::client::Session &session) noexcept;
 
 private:
+  struct HttpSessionData
+  {
+    std::shared_ptr<opentelemetry::ext::http::client::Session> session;
+    std::shared_ptr<opentelemetry::ext::http::client::EventHandler> event_handle;
+
+    inline HttpSessionData() = default;
+
+    inline explicit HttpSessionData(
+        std::shared_ptr<opentelemetry::ext::http::client::Session> &&input_session,
+        std::shared_ptr<opentelemetry::ext::http::client::EventHandler> &&input_handle)
+    {
+      session.swap(input_session);
+      event_handle.swap(input_handle);
+    }
+
+    inline explicit HttpSessionData(HttpSessionData &&other)
+    {
+      session.swap(other.session);
+      event_handle.swap(other.event_handle);
+    }
+
+    inline HttpSessionData &operator=(HttpSessionData &&other) noexcept
+    {
+      session.swap(other.session);
+      event_handle.swap(other.event_handle);
+      return *this;
+    }
+  };
+
+  /**
+   * @brief Create a Session object or return a error result
+   *
+   * @param message The message to send
+   */
+  nostd::variant<sdk::common::ExportResult, HttpSessionData> createSession(
+      const google::protobuf::Message &message,
+      std::function<bool(opentelemetry::sdk::common::ExportResult)> &&result_callback) noexcept;
+
   /**
    * Add http session and hold it's lifetime.
-   * @param session the session to add
-   * @param event_handle the event handle of this session
+   * @param session_data the session to add
    */
-  void addSession(
-      std::shared_ptr<opentelemetry::ext::http::client::Session> session,
-      std::shared_ptr<opentelemetry::ext::http::client::EventHandler> event_handle) noexcept;
+  void addSession(HttpSessionData &&session_data) noexcept;
 
   /**
    * @brief Real delete all sessions and event handles.
@@ -164,28 +211,6 @@ private:
    */
   OtlpHttpClient(OtlpHttpClientOptions &&options,
                  std::shared_ptr<ext::http::client::HttpClient> http_client);
-
-  struct HttpSessionData
-  {
-    std::shared_ptr<opentelemetry::ext::http::client::Session> session;
-    std::shared_ptr<opentelemetry::ext::http::client::EventHandler> event_handle;
-
-    inline HttpSessionData() = default;
-
-    inline explicit HttpSessionData(
-        std::shared_ptr<opentelemetry::ext::http::client::Session> &&input_session,
-        std::shared_ptr<opentelemetry::ext::http::client::EventHandler> &&input_handle)
-    {
-      session.swap(input_session);
-      event_handle.swap(input_handle);
-    }
-
-    inline explicit HttpSessionData(HttpSessionData &&other)
-    {
-      session.swap(other.session);
-      event_handle.swap(other.event_handle);
-    }
-  };
 
   // Stores if this HTTP client had its Shutdown() method called
   bool is_shutdown_;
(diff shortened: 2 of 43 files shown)


lalitb:
Thanks for the PR. Would be reviewing this early next week.

owent:
> Thanks for the PR. Would be reviewing this early next week.

After #1185 is merged, there are no unit test for concurrency sessions now. I wonder whether should I add unit test for this, if so, the connecting problem (in `CI / Bazel valgrind`) may happen again.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -280,18 +280,30 @@ class HttpClient : public opentelemetry::ext::http::client::HttpClient
...
   bool CancelAllSessions() noexcept override
   {
-    for (auto &session : sessions_)
+    // CancelSession may change sessions_, we can not change a container while iterating it.
+    while (!sessions_.empty())
     {
-      session.second->CancelSession();
+      std::map<uint64_t, std::shared_ptr<Session>> sessions;
+      sessions.swap(sessions_);
+      for (auto &session : sessions)
```
esigo, line 288: why iterating over `sessions_` is not ok but iterating over `sessions` is ok here

exporters/otlp/test/otlp_http_log_exporter_test.cc
```diff
@@ -256,6 +259,8 @@ TEST_F(OtlpHttpLogExporterTestPeer, ExportBinaryIntegrationTest)
         http_client::nosend::Response response;
         callback.OnResponse(response);
```
esigo, line 260: don't we need `response.Finish(callback);` here?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -71,20 +76,25 @@ struct OtlpHttpClientOptions
   // Additional HTTP headers
   OtlpHeaders http_headers = GetOtlpDefaultHeaders();
 
+  // Concurrency sessions
+  std::size_t concurrency_sessions = 8;
```
esigo, line 80: nit: Is't it better to use `concurrent_sessions`?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -654,22 +734,38 @@ opentelemetry::sdk::common::ExportResult OtlpHttpClient::Export(
...
-  for (auto &header : options_.http_headers)
   {
-    request->AddHeader(header.first, header.second);
-  }
-  request->SetUri(http_uri_);
-  request->SetTimeoutMs(std::chrono::duration_cast<std::chrono::milliseconds>(options_.timeout));
-  request->SetMethod(http_client::Method::Post);
-  request->SetBody(body_vec);
-  request->ReplaceHeader("Content-Type", content_type);
+    std::lock_guard<std::recursive_mutex> guard{session_manager_lock_};
```
esigo, line 738: Why we need to do all the processing before checking for Shutdown?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -680,38 +776,143 @@ opentelemetry::sdk::common::ExportResult OtlpHttpClient::Export(
...
+    }
+  }
+
+  return !gc_sessions_.empty();
+}
+
 bool OtlpHttpClient::isShutdown() const noexcept
 {
-  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
   return is_shutdown_;
```
esigo, line 916: Without lock this read would have race condition.

owent: `is_shutdown_` and `isShutdown()` are protected by `session_manager_lock_` now.

exporters/otlp/src/otlp_http_client.cc
owent: It's used to protect `isShutdown()`, `is_shutdown_`, `running_sessions_` and `http_client_`. They are all not thread-safe.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
owent: Thanks, I'm not native english speaker.I will change the name here.

exporters/otlp/test/otlp_http_log_exporter_test.cc
owent: I want to reuse it to trigger `OnEvent` and call `OnResponse`. The codes do not trigger `OnEvent` before.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
owent: The call stack `HttpClient::CancelAllSessions` -> `Session::CancelSession` -> `HttpClient::CleanupSession` will crash here.
`CleanupSession` will change `sessions_` when iterating it.

owent:
The timeout test seems not a problem of this PR

owent:
We have a internal benmark report which shows this PR increase QPS of `OtlpHttpLogExporter` from 4k/second to 13k/second.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -680,38 +777,130 @@ opentelemetry::sdk::common::ExportResult OtlpHttpClient::Export(
...
   }
-  bool write_successful = handler->waitForResponse();
 
-  // End the session
-  session->FinishSession();
+  // Wait for any session to finish if there are to many sessions
+  std::unique_lock<std::mutex> lock(session_waker_lock_);
+  bool wait_successful = session_waker_.wait_for(lock, options_.timeout, [this] {
+    std::lock_guard<std::recursive_mutex> guard{session_manager_lock_};
+    return running_sessions_.size() <= options_.concurrent_sessions;
```
lalitb, line 785: What will happen if number of current running sessions are less than max-configured? `Export` will return success even though there is no real successful export ?

lalitb:
> We have a internal benmark report which shows this PR increase QPS of OtlpHttpLogExporter from 4k/second to 13k/second.

Good to know that. I had quick glance at the Export() function, will review it more thoroughly tomorrow. Sorry for the delay on this.

exporters/otlp/src/otlp_http_client.cc
owent: Yes

lalitb: Yes, I can see the improvement we can achieve with these changes. The concern I have is that this deviates from the specs (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#interface-definition-1), which states that the Export method should return the result of transfer of the data over the wire. This can allow the Span Processor to take decision accordingly say retrying the failed transactions.
We may want to change the Export method to return the result callback to make these changes compliant to sepcs ?

api/include/opentelemetry/common/timestamp.h
```diff
@@ -169,5 +169,39 @@ class SteadyTimestamp
...
+
+class DurationUtil
+{
+public:
+  template <class Rep, class Period>
+  static std::chrono::duration<Rep, Period> AdjustWaitForTimeout(
+      std::chrono::duration<Rep, Period> timeout,
+      std::chrono::duration<Rep, Period> indefinite_value) noexcept
+  {
+    // Do not call now() when this duration is max value, now() may have a expemsive cost.
```
reyang, line 181:
```suggestion
    // Do not call now() when this duration is max value, now() may have expensive cost.
```

api/include/opentelemetry/common/timestamp.h
```diff
@@ -169,5 +169,39 @@ class SteadyTimestamp
...
+    if (timeout == std::chrono::duration<Rep, Period>::max())
+    {
+      return indefinite_value;
+    }
+
+    // std::future<T>::wait_for, std::this_thread::sleep_for, and std::condition_variable::wait_for
+    // mey use steady_clock or system_clock.We need make sure now() + timeout do not overflow.
```
reyang, line 188:
```suggestion
    // may use steady_clock or system_clock. We need make sure now() + timeout do not overflow.
```

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -114,19 +126,33 @@ class OtlpHttpClient
...
 
-  // The configuration options associated with this HTTP client.
-  const OtlpHttpClientOptions options_;
+  /**
+   * @brief Real delete all sessions and event handles.
+   * @note This function is called in the same thread where we create sessions and handles
+   *
+   * @return return true if there are more sessions to delete
+   */
+  bool cleanupGCSessions() noexcept;
```
reyang, line 152: What is "GC"?

exporters/otlp/src/otlp_http_client.cc
reyang: I think we should make `isShutdown` thread safe, and short circuit here (e.g. return if isShutdown() is true).

exporters/otlp/src/otlp_http_client.cc
lalitb: Also, had a quick look into how other SIGs are handling this. Java returns export status to processor asynchronously using [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html), while JS returns status using callback.

reyang: FYI https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-concurrent-requests

lalitb:
@owent - This is a great initiative, thanks once again for the PR. Do you think it would make sense to have `exporter` return export status as callback and let the `processor` manage the logic for the number of simultaneous ongoing exports?  This will ensure we are specs compliant, and no separate changes would be required for the Zipkin exporter. Also later, we can make it truly asynchronous by adding support of `multi_handle` for curl client.

exporters/otlp/src/otlp_http_client.cc
owent: > Also, had a quick look into how other SIGs are handling this. Java returns export status to processor asynchronously using [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html), while JS returns status using callback.

Good idea. Maybe we can use `nostd::variant<std::future<sdk::common::ExportResult>>` here. But it's a break change for exporters's APIs.

owent: > FYI https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlpgrpc-concurrent-requests

The `OtlpGrpcExporter` and `OtlpGrpcLogExporter` have the same problem, but I didn't try to use asynchronize API or test the performance yet.

lalitb: > But it's a break change for exporters's APIs. I think it need more discussion about it.

Yes, I was thinking about that too. We have to keep supporting the existing `Exporter::Export()` interface and add new to support callback/futures. But it needs some thought and discussion.

lalitb: > The OtlpGrpcExporter and OtlpGrpcLogExporter have the same problem, but I didn't try to use asynchronize API or test the performance yet.

Agree, they too suffer from the same issue.

api/include/opentelemetry/common/timestamp.h
owent: Fixed and thanks

api/include/opentelemetry/common/timestamp.h
owent: Fixed and thanks

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -99,6 +109,8 @@ class OtlpHttpClient
    */
   explicit OtlpHttpClient(OtlpHttpClientOptions &&options);
 
+  ~OtlpHttpClient();
```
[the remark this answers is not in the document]
owent: GC here mean garbage collection of http sessions.When a http session is finished, we can not destroy it immediately right now, because it may still be visited in some codes. (e.g. When we remove a session in OnResponse, it will be visited in OnEvent later).
So I move the finished sessions into `gc_sessions_` and real destroy it soon later.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
owent: GC here mean garbage collection of http sessions.When a http session is finished, we can not destroy it immediately right now, because it may still be visited in some codes. (e.g. When we remove a session in OnResponse, it will be visited in OnEvent later).
So I move the finished sessions into gc_sessions_ and real destroy it soon later.

lalitb:
@owent  - can we close this PR if the changes are already in feature branch?

lalitb:
@owent  - can we close this PR if the changes are already in feature branch?

owent:
> @owent - can we close this PR if the changes are already in feature branch?

Yes, this is already moved to [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes)

lalitb:
@owent fyi - https://github.com/open-telemetry/opentelemetry-specification/pull/2452#issuecomment-1115973257

PR #1214, opened 2022-02-15
allow extension of the lifetime of ContextStorage.

### LINKED ISSUE #1211: Expose static RuntimeContext storage for lifecycle management

I am working on the Apache Arrow project and this is related to an issue we have: [ARROW-15604](https://issues.apache.org/jira/browse/ARROW-15604)

The problem is that Arrow has a thread pool stored in its own static state.  We create spans in tasks that run on that thread pool.  At shutdown, there is no way to control which static state gets destroyed first.  Usually it is OpenTelemetry's state that is destroyed first because the thread pool tries to wait until its worker threads have finished.  Those worker threads are usually finishing up tasks and, at the end, destroy a scope.  The scope's destructor attempts to access the static runtime context storage but it has already been destroyed.

We can work around this by initializing the runtime context storage ourselves and using `opentelemetry::context::RuntimeContext::SetRuntimeContextStorage` but this is only a workaround because we are a library and not an application so we risk overriding the user's runtime context storage.  Instead we'd like to be able to make a copy of OTel's shared pointer so that we can keep it alive until our threads finish.

If the method `opentelemetry::context::RuntimeContext::GetRuntimeContextStorage()` were public that would solve our problem.

I think we're open to any alternatives that we can use to extend the lifetime of the runtime context storage.

**Additional context**
I've created a simple reproduction of this issue here: https://github.com/westonpace/otel-storage-accessor-repr

esigo:
Fixes #1211 (issue)

## Changes

The specifications require the context to be [immutable](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/context.md#overview), but does not talk about whether the users are allowed to have access to the storage.
The change in this PR allows the users to get a shared_ptr to const so they can't do anything with the pointer but extending the lifetime of the storage.
 
### DIFF
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -150,6 +150,18 @@ public:
     GetStorage() = storage;
   }
 
+  /**
+   * Provide a pointer to const runtime context storage.
+   *
+   * The returned pointer can only be used for extending the lifetime of the runtime context
+   * storage.
+   *
+   */
+  static nostd::shared_ptr<const RuntimeContextStorage> GetConstRuntimeContextStorage() noexcept
+  {
+    return GetRuntimeContextStorage();
+  }
+
 private:
   static nostd::shared_ptr<RuntimeContextStorage> GetRuntimeContextStorage() noexcept
   {


lalitb (approved):

ThomsonTan:
Could we provide some dedicated helper method to increase the lifetime of RuntimeStorage? As the user cannot deduce the purpose of this API directly from the API name.

esigo:
> Could we provide some dedicated helper method to increase the lifetime of RuntimeStorage? As the user cannot deduce the purpose of this API directly from the API name.

Could you please elaborate?

ThomsonTan:
> Could you please elaborate?

I meant some explicit API like `AddRef/ReleaseRef` but this seems requiring 2 APIs. The current one looks good to through.

PR #1219, opened 2022-02-18
[Metric SDK] Synchronous Instruments - Aggregation Storage(s) creation for configured views

### LINKED ISSUE #1217: [Metric SDK] - Sync Instruments - Aggregation Storage(s) creation for configured views.
While creating a Synchronous Instrument through [Meter API](https://github.com/open-telemetry/opentelemetry-cpp/blob/3a9d0870bc846f79f97c81796ce70440d944aa88/api/include/opentelemetry/metrics/meter.h#L39), the SDK should create multiple Aggregation Storage using the below logic:

 - For each of the configured views in View Registry
   -  Find a list of matching views ( same instrument type, name, and description)  using [ViewRegistry::FindViews](https://github.com/open-telemetry/opentelemetry-cpp/blob/3a9d0870bc846f79f97c81796ce70440d944aa88/sdk/include/opentelemetry/sdk/metrics/view/view_registry.h#L46)
   - Create Synchronous Aggregation Storage for each matching view.
   - Register all matching storage against the instrument object, and the meter.

lalitb:
Fixes #1217 

## Changes
Create and store Aggregation Storage(s) as per the logic described in #1217 

Refer to the flow in architecture ( https://docs.google.com/document/d/17PiDI91MQy07kkiSJ7lvlnogvMD3FV85g__WEackj1M ) which shows Aggregation Storage as SyncMetricStorage and AsyncMetricStorage, and the same name is used in the code.


sdk/include/opentelemetry/sdk/metrics/view/view.h
```diff
@@ -21,26 +22,22 @@ class View
...
 
   virtual std::string GetName() const noexcept { return name_; }
 
   virtual std::string GetDescription() const noexcept { return description_; }
 
-  virtual const opentelemetry::sdk::metrics::Aggregation &GetAggregation() const noexcept
-  {
-    return *aggregation_.get();
-  }
+  virtual const AggregationType &GetAggregationType() const noexcept { return aggregation_type_; }
```
esigo, line 40: Why do we need to return `const&` here?

sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -105,7 +106,11 @@ class Meter final : public opentelemetry::metrics::Meter
   // order of declaration is important here - instrumentation library should destroy after
   // meter-context.
   std::unique_ptr<sdk::instrumentationlibrary::InstrumentationLibrary> instrumentation_library_;
-  std::shared_ptr<sdk::metrics::MeterContext> context_;
+  std::shared_ptr<sdk::metrics::MeterContext> meter_context_;
+  std::unordered_map<std::string, std::shared_ptr<MetricStorage>> storage_registry_;
```
esigo, line 110: It would be nice to have the key documented here.

sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -105,7 +106,11 @@ class Meter final : public opentelemetry::metrics::Meter
...
+
+  std::unique_ptr<WritableMetricStorage> RegisterMetricStorage(
```
esigo, line 112: nit: I think for private methods we should use `registerMetricStorage`?

sdk/src/metrics/meter.cc
```diff
@@ -149,7 +179,30 @@ const sdk::instrumentationlibrary::InstrumentationLibrary *Meter::GetInstrumenta
...
+}
+
+std::unique_ptr<WritableMetricStorage> Meter::RegisterMetricStorage(
+    InstrumentDescriptor &instrument_descriptor)
+{
+
+  auto view_registry = meter_context_->GetViewRegistry();
+  std::unique_ptr<WritableMetricStorage> storages(new MultiMetricStorage());
+
+  auto result = view_registry->FindViews(
```
esigo, line 192: What `storages` will be if we didn't find `view` here?
also `result` is not used.

sdk/include/opentelemetry/sdk/metrics/view/view.h
lalitb: Good point. Earlier, this method was returning `Aggregation` object and so was `const &`. . It is not required now, so removed.

sdk/include/opentelemetry/sdk/metrics/meter.h
lalitb: Yes, have added the documentation now.

sdk/include/opentelemetry/sdk/metrics/meter.h
lalitb: Google-style [guide](https://google.github.io/styleguide/cppguide.html#Function_Names) doesn't seem to differentiate in naming private and public methods. I have been sticking to `CamelCase` for both private and public. Looking into the repo, I think we are not consistent and have been using both `CamelCase` and `camelCase`. Probably good to discuss and use one form.

sdk/src/metrics/meter.cc
lalitb: [ViewRegistry::FindViews](https://github.com/open-telemetry/opentelemetry-cpp/blob/9157fd474883b4d90dab2e4f5aa887a705c2ab94/sdk/include/opentelemetry/sdk/metrics/view/view_registry.h#L65) returns a default view (local static) of drop aggregation type if no view is found.

esigo: Thanks for the clarification.

esigo (approved):
LGTM
Thanks for the PR.

PR #1224, opened 2022-02-18
thrift boost deps

esigo:
No need to install  libboost-all-dev for CI
This should also fix CI bazel build failures happening [recently]( libboost-all-dev install from the CI).
## Changes

We need `boost locale` only for building thrift. This PR installs boost locale and removes `libboost-all-dev` install from the CI.


lalitb:
Thanks for the fix. Do you know if anything changed in github env causing this failure?

PR #1225, opened 2022-02-19
Run apt update before install thrift dependencies

ThomsonTan:
## Changes

The current apt cache could be stale when running `setup_thrift.sh`, so update the cache before installing any dependency.

Fixes build error like https://github.com/open-telemetry/opentelemetry-cpp/runs/5255793986?check_suite_focus=true.

### DIFF
--- a/ci/setup_thrift.sh
+++ b/ci/setup_thrift.sh
@@ -4,6 +4,8 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 export THRIFT_VERSION=0.14.1
 
+apt update
+
 if ! type cmake > /dev/null; then
     #cmake not installed, exiting
     exit 1


lalitb:
Interesting, I thought it would be enough to do it in [setup_ci_environment.sh](https://github.com/open-telemetry/opentelemetry-cpp/blob/cf7f9e28ab6610e81a9fc2799b31bbcbef7cf613/ci/setup_ci_environment.sh#L7), as it is run before thrift installation.

lalitb:
> Also we run `setup_thrift.sh` before `setup_ci_environment.sh`.

Good point, should we change the order instead?

lalitb:
Please ignore my last comment. No harm in adding `apt update` irrespective of which order we run. Have merged the  PR :)

ThomsonTan:
> > Also we run `setup_thrift.sh` before `setup_ci_environment.sh`.
> 
> Good point, should we change the order instead?

Yeah, I thought about this, but don't want to add this implicit dependence of our setup scripts.

PR #1251, opened 2022-03-08
Add support for span links to Jaeger export.

deejgregor:
~~Fixes # (issue)~~ (no existing issue found)

## Changes

Add support for span links to Jaeger export.

* ~~[ ] Changes in public API reviewed~~ (no API changes)

### DIFF
--- a/exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
+++ b/exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
@@ -58,6 +58,7 @@ public:
   std::vector<thrift::Tag> Tags() noexcept { return std::move(tags_); }
   std::vector<thrift::Tag> ResourceTags() noexcept { return std::move(resource_tags_); }
   std::vector<thrift::Log> Logs() noexcept { return std::move(logs_); }
+  std::vector<thrift::SpanRef> References() noexcept { return std::move(references_); }
   const std::string &ServiceName() const noexcept { return service_name_; }
 
   void SetIdentity(const opentelemetry::trace::SpanContext &span_context,
@@ -109,6 +110,7 @@ private:
   std::vector<thrift::Tag> tags_;
   std::vector<thrift::Tag> resource_tags_;
   std::vector<thrift::Log> logs_;
+  std::vector<thrift::SpanRef> references_;
   std::string service_name_;
 };
 
--- a/exporters/jaeger/src/recordable.cc
+++ b/exporters/jaeger/src/recordable.cc
@@ -193,7 +193,33 @@ void JaegerRecordable::SetInstrumentationLibrary(
 void JaegerRecordable::AddLink(const trace::SpanContext &span_context,
                                const common::KeyValueIterable &attributes) noexcept
 {
-  // TODO: convert link to SpanRefernece
+  // Note: "The Link’s attributes cannot be represented in Jaeger explicitly."
+  // -- https://opentelemetry.io/docs/reference/specification/trace/sdk_exporters/jaeger/#links
+  //
+  // This implementation does not (currently) implement the optional conversion to span logs.
+
+  thrift::SpanRef reference;
+
+  reference.__set_refType(thrift::SpanRefType::FOLLOWS_FROM);
+
+  // IDs should be converted to big endian before transmission.
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#ids
+#if JAEGER_IS_LITTLE_ENDIAN == 1
+  reference.__set_traceIdHigh(
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()))));
+  reference.__set_traceIdLow(
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1)));
+  reference.__set_spanId(
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data()))));
+#else
+  reference.__set_traceIdLow(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data())));
+  reference.__set_traceIdHigh(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1));
+  reference.__set_spanId(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data())));
+#endif
+
+  references_.push_back(reference);
 }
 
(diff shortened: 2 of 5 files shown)


deejgregor:
One important note: I'm not really a C++ developer, so please look over this with a fine-toothed comb.

exporters/jaeger/src/recordable.cc
```diff
@@ -193,7 +193,33 @@ void JaegerRecordable::SetInstrumentationLibrary(
...
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data()))));
+#else
+  reference.__set_traceIdLow(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data())));
+  reference.__set_traceIdHigh(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1));
+  reference.__set_spanId(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data())));
+#endif
+
+  references_.push_back(reference);
```
lalitb, line 222: how will this list of links get exported to the backend? Does some code need to be added in [ThriftSender::Append()](https://github.com/open-telemetry/opentelemetry-cpp/blob/c96a3e32f0b5bf17e58b1300383f9f205f33558f/exporters/jaeger/src/thrift_sender.cc#L23) to do that?

exporters/jaeger/src/recordable.cc
```diff
@@ -193,7 +193,33 @@ void JaegerRecordable::SetInstrumentationLibrary(
...
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1)));
+  reference.__set_spanId(
+      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data()))));
+#else
+  reference.__set_traceIdLow(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data())));
+  reference.__set_traceIdHigh(
+      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1));
+  reference.__set_spanId(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data())));
+#endif
```
lalitb, line 220: nit - can this code be shared between [JaegerRecordable::SetIdentity()](https://github.com/open-telemetry/opentelemetry-cpp/blob/c96a3e32f0b5bf17e58b1300383f9f205f33558f/exporters/jaeger/src/recordable.cc#L135) and here?

lalitb (approved):

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
```diff
@@ -58,6 +58,7 @@ class JaegerRecordable final : public sdk::trace::Recordable
   std::vector<thrift::Tag> Tags() noexcept { return std::move(tags_); }
   std::vector<thrift::Tag> ResourceTags() noexcept { return std::move(resource_tags_); }
   std::vector<thrift::Log> Logs() noexcept { return std::move(logs_); }
+  std::vector<thrift::SpanRef> References() noexcept { return std::move(references_); }
```
ThomsonTan, line 61: Seems `SpanRef`s are not attached to `jaeger_span`?

exporters/jaeger/src/recordable.cc
deejgregor: Good catch! I missed a line when I pulled the changes over. Added: https://github.com/open-telemetry/opentelemetry-cpp/pull/1251/files#diff-3fc4e4d4cb58766986df79016e29275bf3712c2fdcda74c598152f42598bf546R43

exporters/jaeger/include/opentelemetry/exporters/jaeger/recordable.h
deejgregor: Yup! I goofed and missed bringing over a line. Added: https://github.com/open-telemetry/opentelemetry-cpp/pull/1251/files#diff-3fc4e4d4cb58766986df79016e29275bf3712c2fdcda74c598152f42598bf546R43

(and made sure nothing else is missing)

exporters/jaeger/src/recordable.cc
deejgregor: I considered that, but there are two things that made me lean away from doing it:
1. The target objects are different (`thrift::Span` vs. `thrift::SpanRef`), so I think I'd need to do it with a macro instead of a function. And things are already fairly complex here (at least for my neophyte tastes), which made me avoid the macro. If there's a better way to do it, I'm happy to take pointers.
2. We need to do slightly different work in each case--`thrift::Span` needs the `parent_span_id` copied, but that doesn't exist on `thrift::SpanRef`.

Neither of them are showstoppers, so I'm happy to take a whack at merging it (except for the `parent_span_id`, of course).

I tried expanding the inline otel_bswap_64 function and/or adding a macro function to do a little bit more of the heavy lifting that happens on each line and ran into problems with both. In particular, the varying size of the data, how indexing changes, and where the high/low pieces go made it fairly complex.

The only thing I can think that would make sense would be moving the bits for trace ID and span ID into a macro that would be called like this.... and this works (at least the tests passed):
```
# define FOO(span_context, dst) \
  (dst)->__set_traceIdHigh( \
      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data())))); \
  (dst)->__set_traceIdLow( \
      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1))); \
  (dst)->__set_spanId( \
      otel_bswap_64(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data()))));
#else
# define FOO(span_context, dst) \
  (dst)->__set_traceIdLow( \
      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()))); \
  (dst)->__set_traceIdHigh( \
      *(reinterpret_cast<const int64_t *>(span_context.trace_id().Id().data()) + 1)); \
  (dst)->__set_spanId(*(reinterpret_cast<const int64_t *>(span_context.span_id().Id().data())));
#endif
...
FOO(span_context, span_); // JaegerRecordable::SetIdentity
...
FOO(span_context, &reference); //  JaegerRecordable::AddLink
```

lalitb: Yes I can see the issue here, don't think this to be blocker. We can leave it for now.

deejgregor:
Just force pushed c4f902134a2f373374c5378c6381b1a57221415b with a clang-format fixes for what was identified in https://github.com/open-telemetry/opentelemetry-cpp/runs/5474270958?check_suite_focus=true.

Force pushed a fix earlier for the missing line in https://github.com/open-telemetry/opentelemetry-cpp/pull/1251#discussion_r822229377

ThomsonTan (approved):

Issue #1239, opened 2022-03-01
Async/callback based interface between Span/Log Processor and Span/Log Exporter


lalitb:
As of now, the flow of spans from Span Processor to Exporter is synchronous, which means the processor has to wait for the upload/export status of the ongoing request before invoking a new request. 

``Application -> Span::OnEnd() -> SimpleProcessor::OnEnd -> Mutex Lock -> Exporter::Export()``

The application has to wait for the span to be uploaded to continue further with business logic. 

 The Batch processor simplifies this by batching/caching the spans in the processor and uploading them in a separate thread.
```
Application -> Span::OnEnd() -> BatchProcessor::OnEnd() -> CircularBuffer:Add(span)
            -> Span::OnEnd() -> BatchProcessor::OnEnd() -> CircularBuffer:Add(span)
            -> Span::OnEnd() -> BatchProcessor::OnEnd() -> CircularBuffer:Add(span)   
                                        ProcessorThread-> CircularBuffer::Drain() -> Export::Export()
Application -> Span::OnEnd() -> BatchProcessor::OnEnd() -> CircularBuffer:Add(span)  
```
This allows the Application to continue processing business logic without waiting for spans to export, and also Span::OnEnd() will not block even when export is ongoing in a separate thread ( there is no lock due to the use of  CircularBuffer to store Spans).

With BatchProcessor, while the interface between application and processor is fast (no wait), the interface between processor and exporter is still synchronous. As the processor needs to wait for the ongoing span(s) export to finish before invoking new export. This causes a bottleneck if the application is creating spans faster than the time taken by the exporter to export the spans, and eventually, the CircularBuffer gets full. Once the CircularBuffer gets full, Span::OnEnd() needs to wait/block on the buffer to be available for writing.

The current design for the interface between processor and exporter is driven by [specification ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-exporter) as below
 - Export() will never be called concurrently for the same exporter instance. Export() can be called again only after the current call returns.
 - Success - The batch has been successfully exported. For protocol exporters this typically means that the data is sent over the wire and delivered to the destination server.
 - Note: this result may be returned via an async mechanism or a callback, if that is idiomatic for the language implementation.

So, the solution could be to return the exporter::Export() result as a callback or async mechanism,
[...]

PR #1252, opened 2022-03-09
Added changes for async callback mechanism from Processor to Exporter

DebajitDas:
…porters

Fixes # (https://github.com/open-telemetry/opentelemetry-cpp/issues/1239)

## Changes
Includes changes related to interface between processor to exporter such that async callback can be handled in the processor.

### DIFF
--- a/exporters/elasticsearch/src/es_log_exporter.cc
+++ b/exporters/elasticsearch/src/es_log_exporter.cc
@@ -110,7 +110,6 @@ private:
   bool console_debug_ = false;
 };
 
-
 /**
  * This class handles the async response message from the Elasticsearch request
  */
@@ -121,20 +120,18 @@ public:
    * Creates a response handler, that by default doesn't display to console
    */
   AsyncResponseHandler(
-    std::shared_ptr<ext::http::client::Session> session,
-    nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback,
-    bool console_debug = false)
-    : console_debug_{console_debug}
-    , session_{std::move(session)}
-    , result_callback_{result_callback} {}
+      std::shared_ptr<ext::http::client::Session> session,
+      nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback,
+      bool console_debug = false)
+      : console_debug_{console_debug},
+        session_{std::move(session)},
+        result_callback_{result_callback}
+  {}
 
   /**
    * Cleans up the session in the destructor.
    */
-  ~AsyncResponseHandler()
-  {
-    session_->FinishSession();
-  }
+  ~AsyncResponseHandler() { session_->FinishSession(); }
 
   /**
    * Automatically called when the response is received
@@ -147,10 +144,12 @@ public:
     if (body_.find("\"failed\" : 0") == std::string::npos)
     {
       OTEL_INTERNAL_LOG_ERROR(
-        "[ES Trace Exporter] Logs were not written to Elasticsearch correctly, response body: "
-        << body_);
+          "[ES Trace Exporter] Logs were not written to Elasticsearch correctly, response body: "
+          << body_);
       result_callback_(sdk::common::ExportResult::kFailure);
-    } else {
+    }
+    else
+    {
       result_callback_(sdk::common::ExportResult::kSuccess);
     }
   }
@@ -193,7 +192,6 @@ private:
   // Whether to print the results from the callback
   bool console_debug_ = false;
 };
-
 
 ElasticsearchLogExporter::ElasticsearchLogExporter()
     : options_{ElasticsearchExporterOptions()},
@@ -284,8 +282,9 @@ sdk::common::ExportResult ElasticsearchLogExporter::Export(
 }
 
 void ElasticsearchLogExporter::Export(
-      const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>> &records,
-      nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback) noexcept
+    const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>>
+        &records,
+    nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback) noexcept
 {
   // Return failure if this exporter has been shutdown
   if (isShutdown())
@@ -322,8 +321,8 @@ void ElasticsearchLogExporter::Export(
   request->SetBody(body_vec);
 
   // Send the request
-  auto handler = std::make_shared<AsyncResponseHandler>(
-    session, result_callback, options_.console_debug_);
+  auto handler =
+      std::make_shared<AsyncResponseHandler>(session, result_callback, options_.console_debug_);
   session->SendRequest(handler);
 }
 
--- a/exporters/elasticsearch/src/es_log_exporter.cc
+++ b/exporters/elasticsearch/src/es_log_exporter.cc
@@ -110,6 +110,82 @@ private:
   bool console_debug_ = false;
 };
 
+
+/**
+ * This class handles the async response message from the Elasticsearch request
+ */
+class AsyncResponseHandler : public http_client::EventHandler
+{
+public:
+  /**
+   * Creates a response handler, that by default doesn't display to console
+   */
+  AsyncResponseHandler(
+    std::shared_ptr<ext::http::client::Session> session,
+    nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback,
+    bool console_debug = false)
+    : console_debug_{console_debug}
+    , session_{std::move(session)}
+    , result_callback_{result_callback} {}
+
+  /**
+   * Automatically called when the response is received
+   */
+  void OnResponse(http_client::Response &response) noexcept override
+  {
+
+    // Store the body of the request
+    body_ = std::string(response.GetBody().begin(), response.GetBody().end());
+    session_->FinishSession();
+    if (body_.find("\"failed\" : 0") == std::string::npos)
+    {
+      OTEL_INTERNAL_LOG_ERROR(
+        "[ES Trace Exporter] Logs were not written to Elasticsearch correctly, response body: "
+        << body_);
+      result_callback_(sdk::common::ExportResult::kFailure);
+    } else {
+      result_callback_(sdk::common::ExportResult::kSuccess);
+    }
+  }
+
+  // Callback method when an http event occurs
+  void OnEvent(http_client::SessionState state, nostd::string_view reason) noexcept override
+  {
+    // If any failure event occurs, release the condition variable to unblock main thread
+    switch (state)
+    {
+      case http_client::SessionState::ConnectFailed:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Connection to elasticsearch failed");
+        break;
+      case http_client::SessionState::SendFailed:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Request failed to be sent to elasticsearch");
+
+        break;
+      case http_client::SessionState::TimedOut:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Request to elasticsearch timed out");
+
+        break;
+      case http_client::SessionState::NetworkError:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Network error to elasticsearch");
+        break;
+    }
+    result_callback_(sdk::common::ExportResult::kFailure);
+  }
+
+private:
+  // Stores the session object for the request
+  std::shared_ptr<ext::http::client::Session> session_;
+  // Callback to call to on receiving events
+  nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback_;
+
+  // A string to store the response body
+  std::string body_ = "";
+
+  // Whether to print the results from the callback
+  bool console_debug_ = false;
+};
+
+
 ElasticsearchLogExporter::ElasticsearchLogExporter()
     : options_{ElasticsearchExporterOptions()},
       http_client_{new ext::http::client::curl::HttpClient()}
@@ -162,8 +238,8 @@ sdk::common::ExportResult ElasticsearchLogExporter::Export(
   request->SetBody(body_vec);
 
   // Send the request
-  std::unique_ptr<ResponseHandler> handler(new ResponseHandler(options_.console_debug_));
-  session->SendRequest(*handler);
+  auto handler = std::make_shared<ResponseHandler>(options_.console_debug_);
+  session->SendRequest(handler);
 
   // Wait for the response to be received
   if (options_.console_debug_)
@@ -196,6 +272,50 @@ sdk::common::ExportResult ElasticsearchLogExporter::Export(
   }
 
   return sdk::common::ExportResult::kSuccess;
+}
+
+void ElasticsearchLogExporter::Export(
+      const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>> &records,
+      nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback) noexcept
+{
+  // Return failure if this exporter has been shutdown
+  if (isShutdown())
+  {
+    OTEL_INTERNAL_LOG_ERROR("[ES Log Exporter] Exporting "
+                            << records.size() << " log(s) failed, exporter is shutdown");
+    return;
+  }
+
+  // Create a connection to the ElasticSearch instance
+  auto session = http_client_->CreateSession(options_.host_ + std::to_string(options_.port_));
+  auto request = session->CreateRequest();
+
+  // Populate the request with headers and methods
+  request->SetUri(options_.index_ + "/_bulk?pretty");
+  request->SetMethod(http_client::Method::Post);
+  request->AddHeader("Content-Type", "application/json");
+  request->SetTimeoutMs(std::chrono::milliseconds(1000 * options_.response_timeout_));
+
+  // Create the request body
+  std::string body = "";
+  for (auto &record : records)
+  {
+    // Append {"index":{}} before JSON body, which tells Elasticsearch to write to index specified
+    // in URI
+    body += "{\"index\" : {}}\n";
+
+    // Add the context of the Recordable
+    auto json_record = std::unique_ptr<ElasticSearchRecordable>(
+        static_cast<ElasticSearchRecordable *>(record.release()));
+    body += json_record->GetJSON().dump() + "\n";
+  }
+  std::vector<uint8_t> body_vec(body.begin(), body.end());
+  request->SetBody(body_vec);
+
+  // Send the request
+  auto handler = std::make_shared<AsyncResponseHandler>(
+    session, result_callback, options_.console_debug_);
+  session->SendRequest(handler);
 }
 
 bool ElasticsearchLogExporter::Shutdown(std::chrono::microseconds timeout) noexcept
--- a/exporters/elasticsearch/src/es_log_exporter.cc
+++ b/exporters/elasticsearch/src/es_log_exporter.cc
@@ -175,6 +175,8 @@ public:
       case http_client::SessionState::NetworkError:
         OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Network error to elasticsearch");
         break;
+      default:
+        break;
     }
     result_callback_(sdk::common::ExportResult::kFailure);
   }
--- a/exporters/elasticsearch/src/es_log_exporter.cc
+++ b/exporters/elasticsearch/src/es_log_exporter.cc
@@ -129,6 +129,14 @@ public:
     , result_callback_{result_callback} {}
 
   /**
+   * Cleans up the session in the destructor.
+   */
+  ~AsyncResponseHandler()
+  {
+    session_->FinishSession();
+  }
+
+  /**
    * Automatically called when the response is received
    */
   void OnResponse(http_client::Response &response) noexcept override
@@ -136,7 +144,6 @@ public:
 
     // Store the body of the request
     body_ = std::string(response.GetBody().begin(), response.GetBody().end());
-    session_->FinishSession();
     if (body_.find("\"failed\" : 0") == std::string::npos)
     {
       OTEL_INTERNAL_LOG_ERROR(
--- a/exporters/otlp/src/otlp_http_client.cc
+++ b/exporters/otlp/src/otlp_http_client.cc
@@ -668,8 +668,8 @@ opentelemetry::sdk::common::ExportResult OtlpHttpClient::Export(
   request->ReplaceHeader("Content-Type", content_type);
 
   // Send the request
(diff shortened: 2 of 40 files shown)


esigo:
you may use `./ci/run_docker.sh  tools/format.sh` for fixing the format.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -143,19 +143,19 @@ class Session : public opentelemetry::ext::http::client::Session
   }
 
   virtual void SendRequest(
-      opentelemetry::ext::http::client::EventHandler &callback) noexcept override
+      std::shared_ptr<opentelemetry::ext::http::client::EventHandler> callback) noexcept override
```
lalitb, line 146: Can we transfer the callback ownership, if that helps us avoid shared_ptr?

exporters/zipkin/src/zipkin_exporter.cc
```diff
@@ -93,6 +93,14 @@ sdk::common::ExportResult ZipkinExporter::Export(
   return sdk::common::ExportResult::kSuccess;
 }
 
+void ZipkinExporter::Export(
+  const nostd::span<std::unique_ptr<sdk::trace::Recordable>> &spans,
+  nostd::function_ref<bool(sdk::common::ExportResult)> result_callback)
+  noexcept
+{
+
```
lalitb, line 101: For exporters not supporting async, we can internally have these methods call sync interface along with logging a warning. Something like:

```cpp
void ZipkinExporter::Export(
  const nostd::span<std::unique_ptr<sdk::trace::Recordable>> &spans,
  nostd::function_ref<bool(sdk::common::ExportResult)> result_callback)
  noexcept
{
   log_warning();
   auto status = Export(spans);
   result_callback(status);
}

```

lalitb, line 101: For exporters not supporting async, we can internally have these methods call sync interface along with logging a warning. Something like:

```cpp
void ZipkinExporter::Export(
  const nostd::span<std::unique_ptr<sdk::trace::Recordable>> &spans,
  nostd::function_ref<bool(sdk::common::ExportResult)> result_callback)
  noexcept
{
   log_warning();
   auto status = Export(spans);
   result_callback(status);
}

```

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
DebajitDas: Moving the callback ownership to http_client, would involve storing the unique_ptr to some kind of map (similar to session) and clearing the entry in the map when the async thread completes. But I could not figure out the simplest way to let know http_client that the async thread is completed and map needs to be cleaned. However, cleanup could be achieved by running a separate cleanup thread which cleans up on getting notified by async thread.

Do you have anything else in mind?

The other approach I was thinking was to move the ownership to the lambda function defined in Session::SendRequest, because this lambda function would run in async thread and callback should be valid till lambda function ends. This could have been doable in C++14 as it supports std::move in lambda capture. But I could not think of achieving the same in C++11. 

Also, since we have maintained two versions of Export implementation in es_log_exporter.cc, the scope of Export outlives the async thread and in another version the async thread outlives the scope of Export, unique_ptr needs to outlive both of them.

Let me know your thoughts.

exporters/zipkin/src/zipkin_exporter.cc
DebajitDas: Added as per review comments.

sdk/src/logs/batch_log_processor.cc
```diff
@@ -151,19 +153,28 @@ void BatchLogProcessor::Export(const bool was_force_flush_called)
                       return true;
                     });
                   });
+  if (is_export_async_ == false || was_force_flush_called == true) {
```
lalitb, line 156: I think async export should also be called in case of force flush if `is_export_async` is true. The callback should do the notify in that case?

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
```diff
@@ -37,6 +37,12 @@ struct BatchSpanProcessorOptions
    * equal to max_queue_size.
    */
   size_t max_export_batch_size = 512;
+
+  /**
+   * Determines whether the export happens asynchronously.
+   * Default implementation is synchronous.
+   */
+  bool is_export_async = false;
```
lalitb, line 45: Should the async interface also added for simple_span_processor?

sdk/src/logs/batch_log_processor.cc
DebajitDas: Agreed on force flush. What is your thoughts in case of Shutdown?
Since shutdown is called during destructor, do we need use async export during shutdown.

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
DebajitDas: Agreed on this, will make the change.

sdk/src/logs/batch_log_processor.cc
lalitb: No, there should be no new upload during the shutdown, it should just wait on the thread for pending uploads to finish.

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -110,6 +110,89 @@ class ResponseHandler : public http_client::EventHandler
...
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Request failed to be sent to elasticsearch");
+
+        break;
+      case http_client::SessionState::TimedOut:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Request to elasticsearch timed out");
+
+        break;
+      case http_client::SessionState::NetworkError:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Network error to elasticsearch");
+        break;
```
owent, line 177: Missing `CreateFailed` and `Cancelled` ? Should we add `default: break;` to avoid warnings?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -668,8 +668,8 @@ opentelemetry::sdk::common::ExportResult OtlpHttpClient::Export(
   request->ReplaceHeader("Content-Type", content_type);
 
   // Send the request
-  std::unique_ptr<ResponseHandler> handler(new ResponseHandler(options_.console_debug));
-  session->SendRequest(*handler);
+  auto handler = std::make_shared<ResponseHandler>(options_.console_debug);
+  session->SendRequest(handler);
```
owent, line 672: This is good enough by now. The async version of `OtlpHttpClient` is already finished in #1209 . I will rebase from this PR when it's merged.
There are also some other problems when `OtlpHttpClient` run on multiple threads and they are also fixed in #1209 .

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
owent: I was thinking whether to add a background thread to deal with resources of async HTTP call?
We will also need a thread to run [`curl_multi_perform`](https://curl.se/libcurl/c/[curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html).html) when I start to implement the real async version of `curl::Session::SendRequest`.
With a background worker thread. I think we can allocate the unique_ptr<EventHandler> on stack in sync Export and move it into the background worker thread in the async Export.

> This could have been doable in C++14 as it supports std::move in lambda capture. But I could not think of achieving the same in C++11.
 We can use parameter to move unique_ptr in C++11. Here is a example https://godbolt.org/z/T5sG7K9fn

sdk/src/logs/batch_log_processor.cc
owent: Shouldn't we wait util all data to be exported when we call ForceFlush or Shutdown when in async mode?

owent:
Could we create a feature branch for this PR and #1209 ? The #1209 also fix some problems run under multiple threads when I implement the async version of OtlpHttpClient. There will be many conflicts between these two PRs if we work standalone and everytime when we both make changes to sdk.

sdk/src/logs/batch_log_processor.cc
DebajitDas: Agreed @owent, we should wait.

exporters/elasticsearch/src/es_log_exporter.cc
DebajitDas: Sure, will add this today

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
DebajitDas: > We can use parameter to move unique_ptr in C++11. Here is a example https://godbolt.org/z/T5sG7K9fn

Thanks @owent  for sharing this.

DebajitDas: > Also, since we have maintained two versions of Export implementation in es_log_exporter.cc, the scope of Export outlives the async thread and in another version the async thread outlives the scope of Export, unique_ptr needs to outlive both of them.

@owent But according to above reasoning, I think we can start with shared_ptr for now.

sdk/src/logs/batch_log_processor.cc
lalitb: Agree we should wait in both cases. The [notification ](https://github.com/open-telemetry/opentelemetry-cpp/blob/49648bd8f741a28a2b111c8d8ecad75a070ea649/sdk/src/trace/batch_span_processor.cc#L163) logic can be called from callback for forceflush(), and similar logic can be added for shutdown() ?

DebajitDas: Yes, @lalitb in case of shutdown, we need to add  a new notification logic. I  will do that and raise a PR again

lalitb: > Yes, @lalitb in case of shutdown, we need to add a new notification logic. I will do that and raise a PR again

On further thought, this will work for the export triggered by the forceflush/shutdown. But how about the already ongoing async requests. We need to wait for their completion too. This means that the processor needs to keep track of the status of those requests.

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -110,6 +110,89 @@ class ResponseHandler : public http_client::EventHandler
...
+
+  /**
+   * Automatically called when the response is received
+   */
+  void OnResponse(http_client::Response &response) noexcept override
+  {
+
+    // Store the body of the request
+    body_ = std::string(response.GetBody().begin(), response.GetBody().end());
+    if (body_.find("\"failed\" : 0") == std::string::npos)
```
ThomsonTan, line 144: Could we check response code instead of searching string to detect failure here?

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -110,6 +110,89 @@ class ResponseHandler : public http_client::EventHandler
...
+  }
+
+  // Callback method when an http event occurs
+  void OnEvent(http_client::SessionState state, nostd::string_view reason) noexcept override
+  {
+    // If any failure event occurs, release the condition variable to unblock main thread
+    switch (state)
+    {
+      case http_client::SessionState::ConnectFailed:
+        OTEL_INTERNAL_LOG_ERROR("[ES Trace Exporter] Connection to elasticsearch failed");
```
ThomsonTan, line 164:
```suggestion
        OTEL_INTERNAL_LOG_ERROR("[ES Log Exporter] Connection to Elasticsearch failed");
```

And similar cases on this switch statement.

exporters/elasticsearch/src/es_log_exporter.cc
lalitb: I think this part of code is actually called with the response is a success ( 200 OK). The original implementor of this exporter knows best what failure error it is further trying to catch from the response body ( or need to look into elastic search API details). :)

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -110,6 +110,89 @@ class ResponseHandler : public http_client::EventHandler
...
+  }
+
+private:
+  // Stores the session object for the request
+  std::shared_ptr<ext::http::client::Session> session_;
+  // Callback to call to on receiving events
+  nostd::function_ref<bool(opentelemetry::sdk::common::ExportResult)> result_callback_;
+
+  // A string to store the response body
+  std::string body_ = "";
```
ThomsonTan, line 190:
```suggestion
  std::string body_;
```
Utilize the default ctor which creates empty string as well.

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -198,6 +281,51 @@ sdk::common::ExportResult ElasticsearchLogExporter::Export(
...
+  request->AddHeader("Content-Type", "application/json");
+  request->SetTimeoutMs(std::chrono::milliseconds(1000 * options_.response_timeout_));
+
+  // Create the request body
+  std::string body = "";
+  for (auto &record : records)
+  {
+    // Append {"index":{}} before JSON body, which tells Elasticsearch to write to index specified
+    // in URI
+    body += "{\"index\" : {}}\n";
```
ThomsonTan, line 313: Could al the records share a single {"index":{}}?

DebajitDas: This part of code was copied from the other Export function. The original implementer would be in better position to answer this.

exporters/elasticsearch/src/es_log_exporter.cc
```diff
@@ -110,6 +110,89 @@ class ResponseHandler : public http_client::EventHandler
...
+   */
+  void OnResponse(http_client::Response &response) noexcept override
+  {
+
+    // Store the body of the request
+    body_ = std::string(response.GetBody().begin(), response.GetBody().end());
+    if (body_.find("\"failed\" : 0") == std::string::npos)
+    {
+      OTEL_INTERNAL_LOG_ERROR(
+          "[ES Trace Exporter] Logs were not written to Elasticsearch correctly, response body: "
```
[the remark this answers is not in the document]
DebajitDas: Sure, will incorporate this change.

exporters/elasticsearch/src/es_log_exporter.cc
DebajitDas: Sure, will incorporate this change.

exporters/elasticsearch/src/es_log_exporter.cc
DebajitDas: Sure, will make the change.

sdk/src/logs/batch_log_processor.cc
DebajitDas: Thats a valid point @lalitb. What suggestions do you have on achieving this?  Should the processor create thread and call async export on the new thread?

lalitb: This probably needs more design thinking before finalizing. The callback for the status of export would be invoked by a thread outside of the processor, and this should somehow update the status to the processor. Will give more thoughts on this before writing :)

owent: > This probably needs more design thinking before finalizing. The callback for the status of export would be invoked by a thread outside of the processor, and this should somehow update the status to the processor. Will give more thoughts on this before writing :)

The same problem is already solved in `OtlpHttpClient::Export` of #1209 . When there are too many sessions, `OtlpHttpClient::Export` will wait for some some sessions to finish and then return.

As it in  https://github.com/open-telemetry/opentelemetry-cpp/pull/1209/files#diff-eb7f99ebc152d9cc0f4781cc0a0388907e4d4b58271dba85eba99fd16b880674R782 .The thread which call `OtlpHttpClient::Export` will use one lock and one condition variable to solve the concurrency problem. The lock is used to protect the management of running sessions, and the condition variable is used to be waked up when any session is finished.We allow any thread to call `Shutdown/ForceFlush` or finish and wake up the exporting thread. According to the spec. only one thread will call `Export`.

I think we can do the same changes here.

**BTW:** Can we create a feature branch for #1209 and this PR and merge #1209 to this new branch? We both change  codes for concurrency IO. There will be more and more conflicts if we do more changes later.The later we finish these two PRs, the harder to merge the conflicts.

lalitb: > The same problem is already solved in OtlpHttpClient::Export of https://github.com/open-telemetry/opentelemetry-cpp/pull/1209 . When there are too many sessions, OtlpHttpClient::Export will wait for some some sessions to finish and then return.

Thanks, if this ensures that only one thread will call Export() and works well for new export interface :) Good idea to have both the changes in the feature branch.

DebajitDas: Sure, I would look up the changes in OtlpHttpClient::Export  and would bring in those concept to Processor. 
Let me know If the feature branch is created and step forward to merge them.

DebajitDas: I have looked into the changes done in #1209. Since OtlpHttpClient had to maintain Sessions and EventHandler object until the export completes, it had achieved this by having map of running sessions and list of completed sessions to deal with it and its associated logic to clean up. Thanks @owent for sharing this PR.

From the Processor perspective, we do not need any object to keep it alive till the completion of async Export. All it needs is to wait for uncompleted async Exports to complete during shutdown and we can know this only by the callback. To achieve this, we can add an atomic counter -- which can be increased on every async Export call and would be decreased on the callback function. And Shutdown would wait until this counter is zero. 
        Counter = 0   ---> means no async export running and shutdown can proceed forward
        Counter = n, (n > 0). ----> means n no of async export/thread running concurrently.
        
 @owent @lalitb Please provide your arguments on this.

owent: I suggest to maintain the exporting recordable objects in process.When any exporter have BUGs which try to finish some futures more than once or do not finish some futures for ever, the Shutdown/ForceFlush may wait for ever.

Please let me known if you have any thought about this.

DebajitDas: Agreed @owent. We should maintain the recordable object and also not wait forever during Shutdown/ForceFlush. Thanks.
What is the plan to merge both PRs into a feature branch?

lalitb: > What is the plan to merge both PRs into a feature branch?

We can create a feature branch on the repo ( not on its fork), and you can keep merging the PRs on that. Finally, we can merge that feature to the main. I don't think creating branch on main repo is restricted, otherwise let me know I can create the branch @owent @DebajitDas

DebajitDas: @lalitb I think I do not have rights to create a branch. Can you please help creating a branch?

lalitb: @owent @DebajitDas  have created a brach `async-changes` in the main repo. You can create the PR on this branch ( instead of main) to push your changes there. Anyone should be able to merge to this branch once PR is created. PR to merge from `async-changes` to `main` branch can be created once the changes are ready.

owent: I have create #1274 to merge "Cocurrency otlp http session" into `[async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes)` .

DebajitDas: Hi @owent, I have raised https://github.com/open-telemetry/opentelemetry-cpp/pull/1275, but how do we want to proceed? Shall I go merge my changes first?

owent: @lalitb Could you please review #1274 or #1275 and merge any of them into [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) first. And then we can continue to rebase the other one.
@DebajitDas I think it's fine to merge any of these two PRs first.

lalitb: Sure, have merged #1274. Looks like need to fix the conflict for #1275 to merge that.

DebajitDas: Resolved conflict on #1275. Please let me know if this can be merged. @lalitb

sdk/src/trace/batch_span_processor.cc
```diff
@@ -157,8 +158,30 @@ void BatchSpanProcessor::Export(const bool was_force_flush_called)
...
+  */
+  if (is_export_async_ == false)
+  {
+    exporter_->Export(nostd::span<std::unique_ptr<Recordable>>(spans_arr.data(), spans_arr.size()));
 
+    NotifyForceFlushCompletion(was_force_flush_called);
+  }
+  else
+  {
+    exporter_->Export(nostd::span<std::unique_ptr<Recordable>>(spans_arr.data(), spans_arr.size()),
```
owent, line 172: Should we declare async Export of `LogExporter` and `SpanExporter` as `void Export(const nostd::span<std::unique_ptr<Recordable>> &records, std::function<bool(ExportResult)>&& result_callback) noexcept override` but not `void Export(const nostd::span<std::unique_ptr<Recordable>> &records, nostd::function_ref<bool(ExportResult)> result_callback) noexcept override`.
These code capture some data and this lambda object can not be destroyed before `result_callback` be called. We need store this `result_callback` some where and call it later, but with `function_ref`, we can not move it.

@lalitb @DebajitDas 

> As metioned in http://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p0792r5.html , `function_ref` is designed as a parameter, but not for storage.

sdk/src/trace/batch_span_processor.cc
```diff
@@ -157,8 +158,30 @@ void BatchSpanProcessor::Export(const bool was_force_flush_called)
...
+                      [this, was_force_flush_called](sdk::common::ExportResult result) {
```
owent, line 173: Should we declare async `Export` of `LogExporter` and `SpanExporter` as `void Export(const nostd::span<std::unique_ptr<Recordable>> &records, std::function<bool(ExportResult)>&& result_callback) noexcept override` ?
These code capture some data and this lambda object can not be destroyed before `result_callback` be called. We need store this `result_callback` some where and call it later, but with `function_ref`, we can not move it.

@lalitb @DebajitDas 

> As metioned in http://open-std.org/JTC1/SC22/WG21/docs/papers/2019/p0792r5.html , `function_ref` is designed as a parameter, but not for storage.

lalitb: yes good point. we can't transfer ownership of `function_ref`.
`std::function<bool(ExportResult)>&& result_callback` would be right callback to use.

owent: @DebajitDas Could you please raise another PR to modify the type of `result_callback` and push to `async-changes`? I will rebase from it merge the async version of `OtlpHttpClient::Export` then.

DebajitDas: Sure @owent, will raise by EOD

DebajitDas: Raised this PR https://github.com/open-telemetry/opentelemetry-cpp/pull/1278 which has the changes for type change

owent: > Raised this PR #1278 which has the changes for type change

Thanks I have raise #1281 to implement async `OtlpHttpExporter` and `OtlpHttpLogExporter`. Both async and sync unit tests are also added for these two exporters.

DebajitDas:
Closing this PR as the same has been merged to feature branch async-changes

PR #1283, opened 2022-03-23
span SetAttribute crash

### LINKED ISSUE #1282: span SetAttribute occur crash
When I execute Span::End, if I execute Span::SetAttribute again, a crash will occur. I checked the implementation and found that Span::SetAttribute does not execute to determine whether the recordable_ is null, but other interfaces have been executed.
image
wecom-temp-459b18201251f1904171a86c44fe64e5

esigo:
Fixes #1282 (issue)

## Changes

adds null check and unit test


sdk/src/trace/span.cc
```diff
@@ -94,7 +94,10 @@ Span::~Span()
 void Span::SetAttribute(nostd::string_view key, const common::AttributeValue &value) noexcept
 {
   std::lock_guard<std::mutex> lock_guard{mu_};
-
+  if (recordable_ == nullptr)
+  {
+    return;
+  }
```
reyang, line 100: nit, consider adding an empty line between L100 and L101

reyang (approved):
LGTM, perhaps worth a changelog entry?

sdk/src/trace/span.cc
esigo: thanks, done

sdk/src/trace/span.cc
```diff
@@ -94,6 +94,10 @@ Span::~Span()
 void Span::SetAttribute(nostd::string_view key, const common::AttributeValue &value) noexcept
 {
   std::lock_guard<std::mutex> lock_guard{mu_};
+  if (recordable_ == nullptr)
+  {
```
ThomsonTan, line 98: Do we need emit a log message in this case (`recordable_ == nullptr`)?

ThomsonTan (approved):

sdk/src/trace/span.cc
esigo: We don't emit in the other methods.

owent (approved):
LGTM. Should we also add unit tests to call `AddEvent` , `SetStatus` and `UpdateName` after `End`?
They have simular problems.

esigo:
> LGTM. Should we also add unit tests to call `AddEvent` , `SetStatus` and `UpdateName` after `End`? They have simular problems.

Thanks, added more tests after `End` to cover other methods.

PR #1284, opened 2022-03-23
Disable benchmark action failure

esigo:
Disable benchmark action failure

## Changes

As the benchmark tests are not so stable because of the machines used by github actions, we decided to disable failure but keep the tests.
Also fixes the documentation for the benchmark.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -67,6 +67,6 @@ jobs:
           # Show alert with commit comment on detecting possible performance regression
           alert-threshold: '200%'
           comment-on-alert: true
-          fail-on-alert: true
+          fail-on-alert: false
           gh-pages-branch: gh-pages
           benchmark-data-dir-path: benchmarks


lalitb (approved):

.github/workflows/benchmark.yml
```diff
@@ -67,6 +67,6 @@ jobs:
           # Show alert with commit comment on detecting possible performance regression
           alert-threshold: '200%'
           comment-on-alert: true
-          fail-on-alert: true
+          fail-on-alert: false
```
ThomsonTan, line 70: With `fail-on-alert` set to `false`, will we still get notified with regression, or we need to visit the benchmark page to get the result?

ThomsonTan (approved):

.github/workflows/benchmark.yml
esigo: The workflow will not be marked as failed but still there will be some comment on the commits as we have now.

Issue #1292, opened 2022-03-28
Windows Ninja build fails with 'C:\Program' is not recognized as an internal or external command


juandemanjon:
In a Windows environment where Ninja binary is a path with spaces the build fails and the error messages is:
```'C:\Program' is not recognized as an internal or external command```

The fix is pretty straight forward, just wrapping the %NINJA% call with quotes, which I am going to send a PR

**Steps to reproduce**
To reproduce it, ninja has to be in a path with spaces. In my environment I have:
```
where ninja
C:\ProgramData\chocolatey\bin\ninja.exe
C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe
```

ThomsonTan:
@juandemanjon in your environment, I assume the ninja path issue will not happen because the first one should be used which doesn't include space in the full path, is this correct?

juandemanjon:
@ThomsonTan Actually happens because the script iterates over the _where_ results and the last one is actually the one applied

PR #1293, opened 2022-03-28
Fix for #1292 Ninja path with spaces in Windows

juandemanjon:
Quick fix for #1292


reyang (approved):

lalitb (approved):

tools/build.cmd
```diff
@@ -170,5 +170,5 @@ REM
 REM ##########################################################################################
 :build_config_ninja
 cmake -G "Ninja" -DCMAKE_MAKE_PROGRAM="!NINJA!" -DCMAKE_TOOLCHAIN_FILE="!VCPKG_CMAKE!" !CONFIG! "!ROOT!"
-%NINJA%
+"%NINJA%"
```
ThomsonTan, line 173: Would this work if the full path of ninja has already been quoted?

reyang: > Would this work if the full path of ninja has already been quoted?

No, it won't work. The action (setting `%NINJA%` to something that contains double quotes) here is already sounding like a bug to me.

PR #1297, opened 2022-03-29
Don't create NoopMeterProvider for GetProvider

ThomsonTan:
## Changes

In legacy metrics implementation, `GetProvider` is not exposed, and it returns `shared_ptr` only for `SetMeterProvider` which always overwrites the stored provider, so no need to create `NoopMeterProvider`, because it is never used.


esigo (approved):

lalitb:
I thought the NoopMeterProvider is required for scenario where SDK is not installed.

ThomsonTan:
> I thought the NoopMeterProvider is required for scenario where SDK is not installed.

I'll update this PR to break the case where the default provider is required.

PR #1306, opened 2022-04-01
 Added max async export support using separate AsyncBatchSpan/LogProcessor

### LINKED ISSUE #1239: Async/callback based interface between Span/Log Processor and Span/Log Exporter

DebajitDas:
Fixes #1239 

## Changes: 
 1. Added separate asyncbatchspan/log processor as discussed.
 2. Added max async export support, where the number of running exports are tracked with unique id.
 3. If the running exports are busy and no callback is receieved, any new recordables are discarded.
 4. Added test case for AsyncShutdown if no callback is received. Shutdown should not block for ever.
        

sdk/src/logs/batch_log_processor.cc
```diff
@@ -236,45 +243,61 @@ void BatchLogProcessor::Export()
...
 #  ifdef ENABLE_ASYNC_EXPORT
-void BatchLogProcessor::WaitForShutdownCompletion()
+bool BatchLogProcessor::CleanUpGarbageAsyncData()
 {
-  // Since async export is invoked due to shutdown, need to wait
-  // for async thread to complete.
-  if (is_export_async_)
-  {
-    std::unique_lock<std::mutex> lk(synchronization_data_->async_shutdown_m);
-    while (true)
```
owent, line 265: Should we still need a indefinite loop to wait all running exports to finish in destructor? Or we may start to destroy a exporter when it's still running.

exporters/otlp/test/otlp_http_log_exporter_test.cc
```diff
@@ -103,9 +103,13 @@ class OtlpHttpLogExporterTestPeer : public ::testing::Test
     std::string attribute_storage_string_value[] = {"vector", "string"};
 
     auto provider = nostd::shared_ptr<sdk::logs::LoggerProvider>(new sdk::logs::LoggerProvider());
-    provider->AddProcessor(
-        std::unique_ptr<sdk::logs::LogProcessor>(new sdk::logs::BatchLogProcessor(
-            std::move(exporter), 5, std::chrono::milliseconds(256), 5, is_async)));
+    provider->AddProcessor(std::unique_ptr<sdk::logs::LogProcessor>(
+        new sdk::logs::BatchLogProcessor(std::move(exporter), 5, std::chrono::milliseconds(256), 5
+#    ifdef ENABLE_ASYNC_EXPORT
```
ThomsonTan, line 108: Could we add a CI build for `ENABLE_ASYNC_EXPORT`?

DebajitDas: Yes, I will add on this PR

sdk/src/logs/batch_log_processor.cc
```diff
@@ -292,13 +315,6 @@ void BatchLogProcessor::NotifyCompletion(
     synchronization_data->is_force_flush_notified.store(true, std::memory_order_release);
     synchronization_data->force_flush_cv.notify_one();
   }
-
```
owent, line 295: We need to call `async_export_waker.notify_all()` here, or all threads call Export and Shutdown will be blocked util timeout.

sdk/src/logs/batch_log_processor.cc
```diff
@@ -334,6 +344,17 @@ bool BatchLogProcessor::Shutdown(std::chrono::microseconds timeout) noexcept
...
   }
+#  ifdef ENABLE_ASYNC_EXPORT
+  // wait for running async exports <= 0
+  std::unique_lock<std::mutex> lock(synchronization_data_->async_export_waker_m);
+  synchronization_data_->async_export_waker.wait_for(lock, timeout, [this] {
+    std::lock_guard<std::mutex> lk(synchronization_data_->async_export_data_m);
+    return export_data_storage_->running_async_exports.size() <= 0;
+  });
+
+  while (CleanUpGarbageAsyncData() == false)
```
owent, line 355: Should we use async_export_waker to wait a event instead of busy wait here?

sdk/test/logs/batch_log_processor_test.cc
```diff
@@ -176,18 +187,24 @@ TEST_F(BatchLogProcessorTest, TestAsyncShutdown)
...
 
   // Create a few test log records and send them to the processor
-  const int num_logs = 3;
+  const int num_logs = 500;
 
   for (int i = 0; i < num_logs; ++i)
   {
+    if (i % 20 == 0)
+    {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
owent, line 205: Why do we need sleep here?I think it's better to use a condition_variable to get notify if we want to know when jobs are done and it's ready to send next round datas.
Sometimes the CI runner is slow and can not finish this job in 1 millisecond.

owent (changes requested):
There are same problems between log processor and span processor.
Could you please add timeout mechanism for these two proccesors in case of the exporter miss the callback and add some tests about it to ensure the processors will not wait for ever?

owent:
@lalitb Could we add `async-changes` into `.github/workflows/ci.yml` ?


> ```yaml
> pull_request:
>     branches: [ main, async-changes]
> ```

sdk/src/logs/batch_log_processor.cc
```diff
@@ -236,45 +243,61 @@ void BatchLogProcessor::Export()
     }
     else
     {
+      std::unique_ptr<AsyncExportData> export_data(new AsyncExportData());
+      export_data->recordables =
```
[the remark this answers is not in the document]
owent: I think we can't check which callback is missing or timeout, so we can't recover it if we only has the atomic counter.
According to https://github.com/open-telemetry/opentelemetry-specification/issues/2434#issuecomment-1080924492_ and https://github.com/open-telemetry/opentelemetry-specification/issues/2434, it's the responsibility of exporters to do the retry logic. I think we can just save a sequence or a id related to the callback , but not every recordable.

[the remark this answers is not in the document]
DebajitDas: I have one more query:
Lets assume max_async_export is set to 8.  Now, no callbacks are received on all 8 async export calls. Now, what should we do before we make the 9th async export - Should  we discard the recordables and do not make any export call ? What is your suggestion?

sdk/src/logs/batch_log_processor.cc
DebajitDas: Waiting logic to wait on all running exports to finish is added in Shutdown function.

sdk/src/logs/batch_log_processor.cc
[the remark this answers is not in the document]
owent: I think we can just drop it, just like the sync mode.

lalitb:
> @lalitb Could we add `async-changes` into `.github/workflows/ci.yml` ?
> 
> > ```yaml
> > pull_request:
> >     branches: [ main, async-changes]
> > ```

Yes, we should add it, and can be removed once the feature branch is merged to main.

sdk/src/logs/batch_log_processor.cc
DebajitDas: Yeah will do this.

sdk/test/logs/batch_log_processor_test.cc
DebajitDas: ok will take care this.

sdk/src/logs/batch_log_processor.cc
DebajitDas: Are you talking about async_export_waker.notify_all() to be called in other areas?

sdk/test/logs/batch_log_processor_test.cc
owent: I mean `while (CleanUpGarbageAsyncData() == false);` is a busy wait. `CleanUpGarbageAsyncData` only lock `async_export_data_m`, and this lock has a very small critial section, it's almost busy wait when worker thread is exited when exporter is still exporting data in other thread.

owent:
> > @lalitb Could we add `async-changes` into `.github/workflows/ci.yml` ?
> > > ```yaml
> > > pull_request:
> > >     branches: [ main, async-changes]
> > > ```
> 
> Yes, we should add it, and can be removed once the feature branch is merged to main.

Thanks and I create #1309

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
```diff
@@ -134,22 +141,41 @@ class BatchSpanProcessor : public SpanProcessor
   void DrainQueue();
 
 #ifdef ENABLE_ASYNC_EXPORT
-  /* In case of async export, wait and notify for shutdown to be completed.*/
-  void WaitForShutdownCompletion();
+  struct AsyncExportData
+  {
+    nostd::span<std::unique_ptr<Recordable>> recordables;
```
lalitb, line 146: Why do we need a span of recordable here? It is just a reference and may get invalidated.

sdk/src/logs/batch_log_processor.cc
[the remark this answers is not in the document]
lalitb: Just to be clear of the question, if the max number of ongoing async requests is ongoing, and the circular queue is full, we should drop the incoming recordable ( as in sync mode).

sdk/src/logs/batch_log_processor.cc
owent: I mean `while (CleanUpGarbageAsyncData() == false);` is a busy wait. `CleanUpGarbageAsyncData` only lock `async_export_data_m`, and this lock has a very small critial section, it's almost busy wait when worker thread is exited when exporter is still exporting data in other thread.

.github/workflows/ci.yml
```diff
@@ -150,6 +150,28 @@ jobs:
...
+      with:
+        path: /home/runner/.cache/bazel
+        key: bazel_test
+    - name: setup
+      run: |
+        sudo ./ci/setup_thrift.sh dependencies_only
+        sudo ./ci/setup_ci_environment.sh
+        sudo ./ci/install_bazelisk.sh
+    - name: run tests
+      run: ./ci/do_ci.sh bazel.with_async_export
```
owent, line 173: Could you enable `-DENABLE_ASYNC_EXPORT_PREVIEW` for all jobs but left one without it?I think we also need benchmark,address sanitizer, thread sanitizer and so on for this.

sdk/src/logs/batch_log_processor.cc
```diff
@@ -41,6 +41,14 @@ BatchLogProcessor::BatchLogProcessor(std::unique_ptr<LogExporter> &&exporter,
   synchronization_data_->is_force_flush_pending.store(false);
   synchronization_data_->is_force_flush_notified.store(false);
   synchronization_data_->is_shutdown.store(false);
+#  ifdef ENABLE_ASYNC_EXPORT
+  export_data_storage_->export_ids_flag.reserve(max_export_async_);
```
owent, line 45: Is `export_data_storage_->export_ids_flag.resize(max_export_async_, true);` better?

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
```diff
@@ -141,22 +141,16 @@ class BatchSpanProcessor : public SpanProcessor
...
+    std::queue<size_t> export_ids;
+    std::vector<bool> export_ids_flag;
   };
   std::shared_ptr<ExportDataStorage> export_data_storage_;
 
-  bool CleanUpGarbageAsyncData();
-
   const bool is_export_async_;
   const size_t max_export_async_;
+  static constexpr int kInvalidExportId = 0;
```
owent, line 153: Maybe we can use `constexpr size_t kInvalidExportId = static_cast<size_t>(-1);` here? Which is just like `std::string::npos`. And some compiler will warn comparison between sign and unsigned interger, so we should use size_t for better.

owent (changes requested):

owent:
> @owent I have not added the test case of async forceflush when no callback is received. Because currently forceflush block indefinitely if no callback is recieved. It blocks at wait_result, because wait_result never becomes true. Need your help in Force Flush.

On my suggestion, we can use a LRU map to replace `export_ids` and `export_ids_flag` to resolve this problem. After calling `Export();` in `DoBackgroundWork()`, we can peek all running but timeout exporting and force to call callback.

We can use `list` to store exporting `id` and the state to check if the callback already called(in case of we call callback in `DoBackgroundWork()` and then it be called again by exporters), and we can use `id` as key of a unordered_map and use the iterator of the `list` before as value to implement this LRU map, 
here is  a sample of [lru_map](https://github.com/atframework/atframe_utils/blob/main/include/mem_pool/lru_map.h) in my project.

DebajitDas:
> On my suggestion, we can use a LRU map to replace `export_ids` and `export_ids_flag` to resolve this problem. After calling `Export();` in `DoBackgroundWork()`, we can peek all running but timeout exporting and force to call callback.

If we force call callback for all timed-out exports, we would free up the ids to be re-used by new exports. This way we would be allowing number of running exports > max_export_async. Do you have any thoughts on this?

owent:
> > On my suggestion, we can use a LRU map to replace `export_ids` and `export_ids_flag` to resolve this problem. After calling `Export();` in `DoBackgroundWork()`, we can peek all running but timeout exporting and force to call callback.
> 
> If we force call callback for all timed-out exports, we would free up the ids to be re-used by new exports. This way we would be allowing number of running exports > max_export_async. Do you have any thoughts on this?

Maybe we can add a interface to notify exporter to abort timeout exporting? What your thought about it ?
@lalitb

sdk/src/trace/batch_span_processor.cc
```diff
@@ -218,48 +228,58 @@ void BatchSpanProcessor::Export()
...
+      {
+        std::weak_ptr<ExportDataStorage> export_data_watcher = export_data_storage_;
+
+        std::weak_ptr<SynchronizationData> synchronization_data_watcher = synchronization_data_;
+        exporter_->Export(
+            nostd::span<std::unique_ptr<Recordable>>(spans_arr.data(), spans_arr.size()),
+            [notify_force_flush, synchronization_data_watcher, export_data_watcher,
+             id](sdk::common::ExportResult result) {
+              // TODO: Print result
+              if (synchronization_data_watcher.expired())
```
lalitb, line 254: Seems I am missing something. When will the weak_ptr to synchronization_data_ and export_data_storage_ be expired?

owent: These callback may be moved to another thread by another exporter, an in this situation, processor may be destroyed before invoking callback.

lalitb:
> Maybe we can add a interface to notify exporter to abort timeout exporting? What your thought about it ?
@lalitb

If there are `max_export_async` exports running forever, we should start ignoring the new exports. This is also the current behavior in the case of sequential exports. There may be underlying issues with the network, export endpoint, or the exporter logic, I don't think the processor should try to recover the situation.

sdk/src/trace/batch_span_processor.cc
lalitb: ok, thanks for the explanation. Makes sense.

DebajitDas:
> > Maybe we can add a interface to notify exporter to abort timeout exporting? What your thought about it ?
> > @lalitb
> 
> If there are `max_export_async` exports running forever, the processor should start ignoring the new exports. This is also the current behavior in the case of sequential exports. There may be underlying issues with the network, export endpoint, or the exporter logic, I don't think the processor should try to recover the situation. These are my view, but feel free to comment if they don't seem correct :)

Just to confirm, I am re-iterating what the behaviour is now in this PR
1. In case of normal export calls, if all async exports are busy and no callbacks are recieved, any new exports would be discarded.
2. In case of force-flush is being called and no callback is received, force-flush blocks indefinitely.
3. In case shutdown or destructor is called,  if any running exports do not return within timeout, shutdown will block till timeout and continue.

lalitb:
> Just to confirm, I am re-iterating what the behaviour is now in this PR
> 
> 1. In case of normal export calls, if all async exports are busy and no callbacks are recieved, any new exports would be discarded.
> 2. In case of force-flush is being called and no callback is received, force-flush blocks indefinitely.
> 3. In case shutdown or destructor is called,  if any running exports do not return within timeout, shutdown will block till timeout and continue.

Thanks for summarising - 1 and 3 look good to me. I need to check the force-flush code once again, but can't we make it return/fail (without changing the exporter interface) if exports don't return within the timeout. It should definitely not block indefinitely.

DebajitDas:
> > Just to confirm, I am re-iterating what the behaviour is now in this PR
> > 
> > 1. In case of normal export calls, if all async exports are busy and no callbacks are recieved, any new exports would be discarded.
> > 2. In case of force-flush is being called and no callback is received, force-flush blocks indefinitely.
> > 3. In case shutdown or destructor is called,  if any running exports do not return within timeout, shutdown will block till timeout and continue.
> 
> Thanks for summarising - 1 and 3 look good to me. I need to check the force-flush code once again, but can't we make it return/fail (without changing the exporter interface) if exports don't return within the timeout. It should definitely not block indefinitely.

@owent I am not clear if there is any specific reason to keep ForceFlush blocking (even though there is timeout)?  [wait_result](https://github.com/open-telemetry/opentelemetry-cpp/blob/async-changes/sdk/src/trace/batch_span_processor.cc#:~:text=//%20for%20ever-,wait_result,-%3D%20synchronization_data_%2D%3Eforce_flush_cv.wait_for) never becomes true even if timeout expires, without callback being called.

lalitb:
> > > Just to confirm, I am re-iterating what the behaviour is now in this PR
> > > 
> > > 1. In case of normal export calls, if all async exports are busy and no callbacks are recieved, any new exports would be discarded.
> > > 2. In case of force-flush is being called and no callback is received, force-flush blocks indefinitely.
> > > 3. In case shutdown or destructor is called,  if any running exports do not return within timeout, shutdown will block till timeout and continue.
> > 
> > 
> > Thanks for summarising - 1 and 3 look good to me. I need to check the force-flush code once again, but can't we make it return/fail (without changing the exporter interface) if exports don't return within the timeout. It should definitely not block indefinitely.
> 
> @owent I am not clear if there is any specific reason to keep ForceFlush blocking (even though there is timeout)? [wait_result](https://github.com/open-telemetry/opentelemetry-cpp/blob/async-changes/sdk/src/trace/batch_span_processor.cc#:~:text=//%20for%20ever-,wait_result,-%3D%20synchronization_data_%2D%3Eforce_flush_cv.wait_for) never becomes true even if timeout expires, without callback being called.

Even in the initial sync implementation for BatchSpanProcessor, force-flush block indefinitely, though [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-1) clearly states it should abort/complete within the timeout. So seems we were never specs compliant :)

DebajitDas:
> compliant

In that case, let me not touch ForceFlush for now in this PR. We can take up in separate PR to fix that. Do you agree?

owent:
> > compliant
> 
> In that case, let me not touch ForceFlush for now in this PR. We can take up in separate PR to fix that. Do you agree?

Agree. some exporters implement the timeout/abort while some do not, Ithink we can finish it in the future.

lalitb:
> 



> > > compliant
> > 
> > 
> > In that case, let me not touch ForceFlush for now in this PR. We can take up in separate PR to fix that. Do you agree?
> 
> Agree. some exporters implement the timeout/abort while some do not, Ithink we can finish it in the future.

Agree.

sdk/src/logs/batch_log_processor.cc
DebajitDas: Agreed.

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
DebajitDas: Agreed.

.github/workflows/ci.yml
DebajitDas: Could you please point me to some examples to refer to for benchmark, address sanitizer and thread sanitizer for this?

owent (approved):
LGTM

lalitb:
> LGTM

@owent There was a discussing with @DebajitDas on this PR in today's community meeting. We felt it would be better from maintainability prospective to introduce a separate AsyncBatchSpanProcessor / AsyncBatchLogProcessor and have the async changes confined to there. We can still have feature flag to enable/disable these two components. Let us know if that would be fine? We can afterwards see if similar thing can be done for Exporter.

owent:
> > LGTM
> 
> @owent There was a discussing with @DebajitDas on this PR in today's community meeting. We felt it would be better from maintainability prospective to introduce a separate AsyncBatchSpanProcessor / AsyncBatchLogProcessor and have the async changes for processor confined there. We can still have feature flag to enable/disable these two components. Let us know if that would be fine? We can afterwards see if similar thing can be done for Exporter.

That's fine. I think we need think about how to share codes with aync and async exporter by this way.

owent (changes requested):
There a a lot similar code between async and sync processors and tests. Could we add some classes to maintain these codes instead of jusy copy them?

DebajitDas:
> There a a lot similar code between async and sync processors and tests. Could we add some classes to maintain these codes instead of jusy copy them?

Can we make AsyncBatchSpanProcessor derived from BatchSpanProcessor ? This way we can reduce lots of duplicate code. What do you think?

owent:
> > There a a lot similar code between async and sync processors and tests. Could we add some classes to maintain these codes instead of jusy copy them?
> 
> Can we make AsyncBatchSpanProcessor derived from BatchSpanProcessor ? This way we can reduce lots of duplicate code. What do you think?

I think it's fine right now. `AsyncBatchSpanProcessor` has all members of `BatchSpanProcessor` in current implement.
What's your opinion about this way? @lalitb

lalitb:
> I think it's fine right now. `AsyncBatchSpanProcessor` has all members of `BatchSpanProcessor` in current implement.
> What's your opinion about this way? @lalitb

It looks me good idea to derive Async if that allows us to eliminate lots of duplicate code. Unless @owent you see any issue in this, will leave it to you to decide :)
I will be spending time over this weekend for in-depth review of exporter and processor changes.

sdk/src/logs/async_batch_log_processor.cc
```diff
@@ -0,0 +1,393 @@
...
+#    include <vector>
+using opentelemetry::sdk::common::AtomicUniquePtr;
+using opentelemetry::sdk::common::CircularBufferRange;
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace logs
+{
+AsyncBatchLogProcessor::AsyncBatchLogProcessor(
```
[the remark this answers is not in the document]
DebajitDas: Thanks, added in the PR

exporters/otlp/test/otlp_http_log_exporter_test.cc
```diff
@@ -103,13 +104,15 @@ class OtlpHttpLogExporterTestPeer : public ::testing::Test
     std::string attribute_storage_string_value[] = {"vector", "string"};
 
     auto provider = nostd::shared_ptr<sdk::logs::LoggerProvider>(new sdk::logs::LoggerProvider());
-    provider->AddProcessor(std::unique_ptr<sdk::logs::LogProcessor>(
-        new sdk::logs::BatchLogProcessor(std::move(exporter), 5, std::chrono::milliseconds(256), 5
 #    ifdef ENABLE_ASYNC_EXPORT
-                                         ,
-                                         is_async
+    provider->AddProcessor(
+        std::unique_ptr<sdk::logs::LogProcessor>(new sdk::logs::AsyncBatchLogProcessor(
```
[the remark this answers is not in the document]
DebajitDas: Thanks, done in the next PR

sdk/include/opentelemetry/sdk/logs/batch_log_processor.h
```diff
@@ -77,21 +77,21 @@ class BatchLogProcessor : public LogProcessor
                              const BatchLogProcessorOptions &options);
 
   /** Makes a new recordable **/
-  std::unique_ptr<Recordable> MakeRecordable() noexcept override;
+  virtual std::unique_ptr<Recordable> MakeRecordable() noexcept override;
```
owent, line 80: We should not add `virtual` keyword for overried methods. Clang will warning this.

sdk/include/opentelemetry/sdk/trace/async_batch_span_processor.h
```diff
@@ -25,20 +25,7 @@ namespace trace
...
-
-  /* The time interval between two consecutive exports. */
-  std::chrono::milliseconds schedule_delay_millis = std::chrono::milliseconds(5000);
-
-  /**
-   * The maximum batch size of every export. It must be smaller or
-   * equal to max_queue_size.
-   */
-  size_t max_export_batch_size = 512;
+  BatchSpanProcessorOptions options;
```
owent, line 28: Just want to discuss it, is it simpiler to declare `struct AsyncBatchSpanProcessorOptions : public BatchSpanProcessorOptions` ?

sdk/src/logs/async_batch_log_processor.cc
```diff
@@ -378,13 +161,7 @@ bool AsyncBatchLogProcessor::Shutdown(std::chrono::microseconds timeout) noexcep
...
 }
 
-AsyncBatchLogProcessor::~AsyncBatchLogProcessor()
-{
-  if (synchronization_data_->is_shutdown.load() == false)
-  {
-    Shutdown();
-  }
-}
+AsyncBatchLogProcessor::~AsyncBatchLogProcessor() {}
```
owent, line 164: We should shutdown here.If it's shutdown in destructor of base class. the `AsyncBatchLogProcessor::Shutdown` will not be called.

sdk/src/logs/async_batch_log_processor.cc
```diff
@@ -362,8 +145,8 @@ bool AsyncBatchLogProcessor::Shutdown(std::chrono::microseconds timeout) noexcep
   GetWaitAdjustedTime(timeout, start_time);
   // wait for  all async exports to complete and return if timeout reached.
   {
-    std::unique_lock<std::mutex> lock(synchronization_data_->async_export_data_m);
-    synchronization_data_->async_export_waker.wait_for(lock, timeout, [this] {
+    std::unique_lock<std::mutex> lock(export_data_storage_->async_export_data_m);
```
owent, line 148: I think we should loop wair_for `scheduled_delay_millis_` in a loop until timeout is reached.When `async_export_waker.notify_all()` is called between predicate callback and the real lock operation.This may wait for ever.

owent (changes requested):

owent:
BTW:

```cmake
if(WITH_METRICS_PREVIEW)
  add_definitions(-DENABLE_METRICS_PREVIEW)
endif()
```

These scripts in `CMakeLists.txt` should be moved into `api/CMakeLists.txt` and be replaced by

```cmake
if(WITH_METRICS_PREVIEW)
  target_compile_definitions(opentelemetry_api INTERFACE ENABLE_METRICS_PREVIEW)
endif()
```

Other options are already be moved in `main` branch. And we could merge from `main` some times later, could you please move `ENABLE_METRICS_PREVIEW` in this PR?

sdk/src/logs/async_batch_log_processor.cc
DebajitDas: I am not clear on this. Could you please elaborate on how this might wait forever or share some sample code which fixes this?

sdk/src/logs/async_batch_log_processor.cc
DebajitDas: Agreed

sdk/include/opentelemetry/sdk/logs/batch_log_processor.h
DebajitDas: ok, will make the change.

sdk/include/opentelemetry/sdk/trace/async_batch_span_processor.h
DebajitDas: Thanks, this is much better.

DebajitDas:
> BTW:
> 
> ```cmake
> if(WITH_METRICS_PREVIEW)
>   add_definitions(-DENABLE_METRICS_PREVIEW)
> endif()
> ```
> 
> These scripts in `CMakeLists.txt` should be moved into `api/CMakeLists.txt` and be replaced by
> 
> ```cmake
> if(WITH_METRICS_PREVIEW)
>   target_compile_definitions(opentelemetry_api INTERFACE ENABLE_METRICS_PREVIEW)
> endif()
> ```
> 
> Other options are already be moved in `main` branch. And we could merge from `main` some times later, could you please move `ENABLE_METRICS_PREVIEW` in this PR?

You must be talking about moving WITH_ASYNC_EXPORT_PREVIEW instead.

owent:
> > BTW:
> > ```cmake
> > if(WITH_METRICS_PREVIEW)
> >   add_definitions(-DENABLE_METRICS_PREVIEW)
> > endif()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > These scripts in `CMakeLists.txt` should be moved into `api/CMakeLists.txt` and be replaced by
> > ```cmake
> > if(WITH_METRICS_PREVIEW)
> >   target_compile_definitions(opentelemetry_api INTERFACE ENABLE_METRICS_PREVIEW)
> > endif()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Other options are already be moved in `main` branch. And we could merge from `main` some times later, could you please move `ENABLE_METRICS_PREVIEW` in this PR?
> 
> You must be talking about moving WITH_ASYNC_EXPORT_PREVIEW instead.

Yes, sorry for my mistake.

sdk/src/logs/async_batch_log_processor.cc
DebajitDas: Thanks, for the clarification. I understood your concern here.
I re-looked into the BatchLogProcessor::ForceFlush at https://github.com/DebajitDas/opentelemetry-cpp/blob/export-changes/sdk/src/logs/batch_log_processor.cc#L112. So it handles this case as follows:
1. when the timeout is set as std::chrono::duration<Rep, Period>::max() or too large, we are waiting on conditiion variable for scheduled_delay_millis_ in loop untill the predicate is true.
2. When the timeout is set as some finite value, we wait till timeout.

Now, if we introduce the above concept in Shutdown, then the predicate in case 1 above would never be true if no callback is called and hence it would block indefinitely. Is this valid then?

owent: The problem is the predicate counld be true, but the condition_variable may not be notified any more when shutdown.
Let's looks into the implement of `wait_for` in libstdc++ (GCC 11.2.0).

```cpp
template<typename _Clock, typename _Duration, typename _Predicate>
bool wait_until(unique_lock<mutex>& __lock, const chrono::time_point<_Clock, _Duration>& __atime, _Predicate __p) {
  while (!__p())
    if (wait_until(__lock, __atime) == cv_status::timeout) // The real lock operation
      return __p();
  return true;
}

template<typename _Rep, typename _Period>
cv_status  wait_for(unique_lock<mutex>& __lock, const chrono::duration<_Rep, _Period>& __rtime) {
  using __dur = typename steady_clock::duration;
  return wait_until(__lock, steady_clock::now() + chrono::__detail::ceil<__dur>(__rtime));
}
```

And the threads run as below:

| The thread call `Shutdown()`                             | Background woker thread                                 |                                                              |
| -------------------------------------------------------- | ------------------------------------------------------- | ------------------------------------------------------------ |
| `while (!__p())`                                         | -                                                       | `__p()` returns `false`                                      |
| -                                                        | `export_data->export_ids.push(id);`                     |                                                              |
| -                                                        | `export_data->export_ids_flag[id - 1] = true;`          | `__p()` should return `true` now                             |
| -                                                        | `export_data_storage->async_export_waker.notify_all();` | `notify_all()` is just ignored because it not locked yet.    |
| `if (wait_until(__lock, __atime) == cv_status::timeout)` | -                                                       | `async_export_waker` is locked now, and wait until `__atime` | 

What I mean is just let `async_export_waker` have a change to run `__p()` again and break this loop.

owent (approved):
LGTM now.

PR #1314, opened 2022-04-03
Move public definitions to `opentelemetry_api`.

### LINKED ISSUE #1311: We should move all public definitions in `CMakeLists.txt` to `target_compile_definitions(opentelemetry_api INTERFACE ...)`
**Describe your environment**

System: Windows
Build System: cmake

**Steps to reproduce**

`cmake <...> -DWITH_ETW=ON`

Now, some definitions are added by `add_definitions(...)` in `CMakeLists.txt` and so that these definitions are private. But we should keep the same behaviour between compiling opentelemetry-cpp and using it by users. We can add these public definitions by `target_compile_definitions(opentelemetry_api INTERFACE ...)` and then these definitions will be auto passed to all cmake targets depend `opentelemetry_api`.

Some definitions are missing and library get a different behavior.

**Additional context**

Should we add prefix to `NO_GETENV` and `NOMINMAX` in case of them will conflict with other libraries?

It this's acceptable, it can be assigned to me please.

owent:

Fixes #1311 

## Changes

Move definitions to `opentelemetry_api` and set them as `INTERFACE` definitions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,24 +125,11 @@ if(NOT DEFINED CMAKE_CXX_STANDARD)
   endif()
 endif()
 
-if(WITH_ABSEIL)
-  find_package(absl CONFIG REQUIRED)
-
-  set(CORE_RUNTIME_LIBS absl::bad_variant_access absl::any absl::base
-                        absl::bits absl::city)
-
-  # target_link_libraries(main PRIVATE absl::any absl::base absl::bits
-  # absl::city)
-endif()
-
 if(WITH_STL)
   # These definitions are needed for test projects that do not link against
   # opentelemetry-api library directly. We ensure that variant implementation
   # (absl::variant or std::variant) in variant unit test code is consistent with
-  # the global project build definitions.
-  add_definitions(-DHAVE_CPP_STDLIB)
-
-  # Optimize for speed to reduce the hops
+  # the global project build definitions. Optimize for speed to reduce the hops
   if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     if(CMAKE_BUILD_TYPE MATCHES Debug)
       # Turn off optimizations for DEBUG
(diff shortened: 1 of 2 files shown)


lalitb (approved):
LGTM, Thanks for the cleanup.

CMakeLists.txt
```diff
@@ -125,24 +125,11 @@ if(NOT DEFINED CMAKE_CXX_STANDARD)
   endif()
 endif()
 
-if(WITH_ABSEIL)
-  find_package(absl CONFIG REQUIRED)
-
-  set(CORE_RUNTIME_LIBS absl::bad_variant_access absl::any absl::base
```
ThomsonTan, line 131: Where is the definition of `CORE_RUNTIME_LIBS` moved to?

CMakeLists.txt
```diff
@@ -215,17 +180,8 @@ option(
 option(WITH_EXAMPLES "Whether to build examples" ON)
 
 option(WITH_METRICS_PREVIEW "Whether to build metrics preview" OFF)
-
-if(WITH_METRICS_PREVIEW)
-  add_definitions(-DENABLE_METRICS_PREVIEW)
```
ThomsonTan, line 220: Should this and `ENABLE_LOGS_PREVIEW` below be moved to api/CMakeLists.txt as well?

CMakeLists.txt
lalitb: The variable is actually removed, and these libraries are now directly linked to `opentelemetry-api`

test

CMakeLists.txt
owent: These two definitions are already in `api/CMakeLists.txt`. It's duplicate here.

ThomsonTan (approved):

lalitb:
@owent - Can you rebase it enable merge. Thanks

owent:
> Reference

Done

PR #1317, opened 2022-04-05
Cocurrency otlp http session

### LINKED ISSUE #1243: Using `libcurl-multi` to implement `opentelemetry::ext::http::client::curl::HttpClient`

To improve IO performance as mentioned in #1175 and #1176 

Using  `libcurl-multi` instead of invoking threads for `HttpClient`.

**Additional context**

+ https://curl.se/libcurl/c/libcurl-multi.html
+ https://curl.se/libcurl/c/multi-double.html

owent:
Fixes #1243 

## Changes

+ Using `curl_multi_handle` to improve performance of `curl::HttpClient`
+ Keep the same behavior as before when using `curl::HttpOperation::Send` (sync mode), only malloc async data and bind `curl_easy_handle` to `curl_multi_handle` for `curl::HttpOperation::SendAsync`.
+ Add background thread to call `curl_multi_perform` and `curl_multi_poll` in `curl::HttpClient`, and if no more pending request, the thread will exit and we will create background thread again if there are more requests in the future.
+ Do not use system `close(fd)` to abort request, use `curl_multi_wakeup()` to wakup and do the real removing the the background thread.Also we allow to break connecting, sending or receiving data from callback of curl.

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1,0 +1,684 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "opentelemetry/ext/http/client/curl/http_operation_curl.h"
+
+#include "opentelemetry/ext/http/client/curl/http_client_curl.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace ext
+{
+namespace http
+{
+namespace client
+{
+namespace curl
+{
+
+size_t HttpOperation::WriteMemoryCallback(void *contents, size_t size, size_t nmemb, void *userp)
+{
+  HttpOperation *self = reinterpret_cast<HttpOperation *>(userp);
+  if (nullptr == self)
+  {
+    return 0;
+  }
+
+  self->raw_response_.insert(self->raw_response_.end(), static_cast<char *>(contents),
+                             static_cast<char *>(contents) + (size * nmemb));
+
+  if (self->WasAborted())
+  {
+    return 0;
+  }
+
+  if (self->GetSessionState() == opentelemetry::ext::http::client::SessionState::Connecting)
+  {
+    self->DispatchEvent(opentelemetry::ext::http::client::SessionState::Connected);
+  }
+
+  if (self->GetSessionState() == opentelemetry::ext::http::client::SessionState::Connected)
+  {
+    self->DispatchEvent(opentelemetry::ext::http::client::SessionState::Sending);
+  }
+
+  return size * nmemb;
+}
+
+size_t HttpOperation::WriteVectorHeaderCallback(void *ptr, size_t size, size_t nmemb, void *userp)
+{
+  HttpOperation *self = reinterpret_cast<HttpOperation *>(userp);
+  if (nullptr == self)
+  {
+    return 0;
+  }
+
+  const unsigned char *begin = (unsigned char *)(ptr);
+  const unsigned char *end   = begin + size * nmemb;
+  self->response_headers_.insert(self->response_headers_.end(), begin, end);
+
+  if (self->WasAborted())
+  {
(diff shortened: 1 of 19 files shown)


owent:
While #1309 is already merged, why there still is only one check here?

lalitb:
> While #1309 is already merged, why there still is only one check here?

Probably the event is triggered form the workflow file present in the branch where merge is happening, which in this case is `async-changes` branch :)

owent:
> > While #1309 is already merged, why there still is only one check here?
> 
> Probably the event is triggered form the workflow file present in the branch where merge is happening, which in this case is `async-changes` branch :)

Thanks, I merged `main` branch, and it works now.

lalitb:
> Thanks, I merged `main` branch, and it works now.

Looks like this has messed up the PR, I see 56 files modified :).

owent:
> > Thanks, I merged `main` branch, and it works now.
> 
> Looks like this has messed up the PR, I see 56 files modified :).

I have merged main branch, it has a lot change between `async-changes` and `main`. Maybe I should create a PR to merge `main` to `async-changes` first and then create this PR?

lalitb:
> > > Thanks, I merged `main` branch, and it works now.
> > 
> > 
> > Looks like this has messed up the PR, I see 56 files modified :).
> 
> I have merged main branch, it has a lot change between `async-changes` and `main`. Maybe I should create a PR to merge `main` to `async-changes` first and then create this PR?

Yes I think that would be better. Sorry about all this trouble.

owent:
> I have merged main branch, it has a lot change between `async-changes` and `main`. Maybe I should create a PR to merge `main` to `async-changes` first and then create this PR?

This PR is base on #1320 now, I think the change list will be simple after #1320 is merged.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -0,0 +1,685 @@
...
+        curl_easy_reset(curl_resource_.easy_handle);
+      }
+      session->GetHttpClient().ScheduleRemoveSession(session->GetSessionId(),
+                                                     std::move(curl_resource_));
+    }
+
+    callback.swap(async_data_->callback);
+    if (callback)
+    {
+      async_data_->callback_thread == std::this_thread::get_id();
```
DebajitDas, line 371: Are we trying to assign here instead of using "==" here?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -0,0 +1,685 @@
...
+      callback(*this);
+      async_data_->callback_thread == std::thread::id();
```
DebajitDas, line 373: I am not clear why are we assigning a new thread id here?

ext/src/http/client/curl/http_operation_curl.cc
owent: Yes and thanks.it's fixed now.

ext/src/http/client/curl/http_operation_curl.cc
owent: It's not necessary, but I think it's more clear to reset `callback_thread` after callback is finished.

owent:
@lalitb This PR is ready to be reviewed now. The **Performance Alert** is expired and please just ignore them.

lalitb:
> @lalitb This PR is ready to be reviewed now. The **Performance Alert** is expired and please just ignore them.

Thanks @owent. Will review this over the weekend.

DebajitDas (approved):
LGTM

lalitb:
@owent one test is failing. Can we fix that before merging to feature branch.

owent:
> @owent one test is failing. Can we fix that before merging to feature branch.

Done. The failed test is network error. I have click rerun and it pass now.

lalitb:
Merging it now. Are we good to raise PR from feature branch to main?

PR #1322, opened 2022-04-08
[ETW Exporter] - ETW provider handle cleanup

lalitb:
Fixes:

ETW provider `handle` is not cleaned-up properly when ETW logger is destroyed. While the handle is deregistered, it is not removed properly from the list of active providers ( because deletion is done wrongly on the copy of the provider map). The provider map gets corrupted and results in event loss if the logger with the same name is created again.

## Changes

Tiny one character change to ensure deletion of provider handle happens through the reference of the provider list, and not from the copy. 

### DIFF
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
@@ -203,6 +203,7 @@ public:
   {
     std::lock_guard<std::mutex> lock(m_providerMapLock);
 
+    // use reference to provider list, NOT it' copy.
     auto &m = providers();
     auto it = m.begin();
     while (it != m.end())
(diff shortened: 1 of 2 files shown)


exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
```diff
@@ -203,7 +203,7 @@ class ETWProvider
   {
     std::lock_guard<std::mutex> lock(m_providerMapLock);
 
-    auto m  = providers();
+    auto &m = providers();
```
ThomsonTan, line 206: Add a comment? Or could we disallow `providers()` get copied?

ThomsonTan (approved):

exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
lalitb: Thanks, added a comment.

PR #1329, opened 2022-04-15
Remove redundant tail / in DESTINATION in CMake install

ThomsonTan:
Remove the redundant `/` at the end of DESTINATION in cmake function `install()`.

Also merged 2 identical CMake `install()` which only diff in exclude pattern.

## Changes

### DIFF
--- a/exporters/elasticsearch/CMakeLists.txt
+++ b/exporters/elasticsearch/CMakeLists.txt
@@ -21,7 +21,7 @@ install(
 
 install(
   DIRECTORY include/opentelemetry/exporters/elasticsearch
-  DESTINATION include/opentelemetry/exporters/
+  DESTINATION include/opentelemetry/exporters
   FILES_MATCHING
   PATTERN "*.h"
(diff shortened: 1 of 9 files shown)


exporters/elasticsearch/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ install(
 
 install(
   DIRECTORY include/opentelemetry/exporters/elasticsearch
-  DESTINATION include/opentelemetry/exporters/
+  DESTINATION include/opentelemetry/exporters
```
lalitb, line 24: Thanks for the fix, just wondering if this fails build in any scenario?

lalitb (approved):

exporters/elasticsearch/CMakeLists.txt
ThomsonTan: This don't fail anything so far. I just observed there are path like `include/opentelemetry/exporters//etw/config.h` which may confuse.

esigo (approved):

PR #1330, opened 2022-04-16
metrics histogram example

esigo:
Adds metrics histogram example.
## Changes

### DIFF
--- a/examples/common/metrics_foo_library/foo_library.cc
+++ b/examples/common/metrics_foo_library/foo_library.cc
@@ -2,24 +2,43 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #ifndef ENABLE_METRICS_PREVIEW
+#  include "foo_library.h"
 #  include <chrono>
+#  include <list>
 #  include <thread>
 #  include "opentelemetry/metrics/provider.h"
 
 namespace nostd       = opentelemetry::nostd;
 namespace metrics_api = opentelemetry::metrics;
 
-void foo_library(const std::string &name)
+void foo_library::counter_example(const std::string &name)
 {
-  // Get the Meter from the MeterProvider
   auto provider                               = metrics_api::Provider::GetMeterProvider();
   nostd::shared_ptr<metrics_api::Meter> meter = provider->GetMeter(name, "1.2.0");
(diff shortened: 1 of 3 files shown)


examples/common/metrics_foo_library/foo_library.cc
```diff
@@ -2,24 +2,43 @@
...
+  auto provider                               = metrics_api::Provider::GetMeterProvider();
+  nostd::shared_ptr<metrics_api::Meter> meter = provider->GetMeter(name, "1.2.0");
+  auto histogram_counter                      = meter->CreateDoubleHistogram(name);
+  auto context                                = opentelemetry::context::Context{};
+  std::map<std::string, std::string> labels   = {{"key", "value"}};
+  auto labelkv = opentelemetry::common::KeyValueIterableView<decltype(labels)>{labels};
+  while (true)
+  {
+    double val = (rand() % 700) + 1.1;
+    histogram_counter->Record(val, labelkv, context);
```
lalitb, line 39: May be as an enhancement later, just like `val`, we can also have `attributes` passed randomly. Define 4-5 kv-attributes, and pass one of them randomly in `Record` method.

examples/metrics_simple/metrics_ostream.cc
```diff
@@ -44,21 +44,49 @@ void initMetrics(const std::string &name)
...
-  foo_library(name);
+
+  if (example_type == "counter")
+  {
+    foo_library::counter_example(name);
+  }
+  else
+  {
+    foo_library::histogram_example(name);
+  }
```
lalitb, line 89: Would also be a good use-case to create both counter and histogram if no argument is specified - to demonstrate getting metrics from multiple instruments simultaneously.

lalitb (approved):
Thanks for adding the support. LGTM with minor comments.

examples/common/metrics_foo_library/foo_library.cc
esigo: thanks, added random attribute.

examples/metrics_simple/metrics_ostream.cc
esigo: thanks, added multiple instruments simultaneously.

PR #1333, opened 2022-04-17
dependencies image as artifact

### LINKED ISSUE #393: Common platform builds: enhance CI to build in nested docker images
@lalitb suggested this approach during code review.

Rather than targeting 'legacy' or 'latest' compiler and a default set of GitHub action images, we can implement better (and faster) test coverage with the following CI loop in nested Docker images:

| OS | Compiler |
|----|------------|
| Ubuntu 14.04 | gcc-4.8 |
| Ubuntu 16.04 | gcc-5.3.1 |
| Ubuntu 18.04 | gcc-7.4.0 |
| Ubuntu 20.04 | gcc-9.x.x (and gcc-10.x.x) |

Draft PR that adds these containers. But does not yet add these to the CI yet is here:  #377 

Above configurations are tested and known to work, passing all core tests. We can also add `RHEL` and `CentOS` images (yum / rpm-based distros) as well. The aim is to iteratively build and run these Docker images on a GitHub Actions Linux runner, using GitHub cache (up to 5GB per repo) - to save the cached images. This GitHub action seems to be implementing what's needed: 
https://github.com/marketplace/actions/build-docker-images-using-cache

Linux docker allows for nested containers: running docker container in GitHub action (which itself is run in docker container).

We can organize a build farm for up to 10 popular Linux repos using one set of build tools: `tools/build.sh`

Anyone can later reuse this effort to build OpenTelemetry C++ SDK PPA for Ubuntu or alike. Personal Package Archives (PPA) allow you to upload Ubuntu source packages to be built and published as an apt repository by Launchpad.

esigo:
Fixes #393 (issue)

## Changes
Adds a weekly job to build dependencies and store them in an image. The image will be stored as an artifact (see an example [here](https://github.com/esigo/opentelemetry-cpp/actions/runs/2181104518)) and can be loaded in our CI, to both speed up the current builds and also enable tests on other platforms that would be very slow without dependencies already installed.

An example action is provided [here](https://github.com/esigo/opentelemetry-cpp/blob/artifacts/.github/workflows/deps_download.yml) which runs our cmake-test using the image downloaded as artifact. 

### DIFF
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -45,3 +45,4 @@ jobs:
       with:
         name: otel-cpp-deps
         path: /opt/otel-cpp-deps.tar
+        retention-days: 14
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -1,0 +1,47 @@
+name: 'OpenTelemetry-cpp dependencies image'
+on:
+  schedule:
+    - cron: "0 3 * * 6"
+
+jobs:
+  docker_image:
+    name: Docker Image
+    runs-on: ubuntu-latest
+    timeout-minutes: 300
+    steps:
+    -
+      name: checkout
+      uses: actions/checkout@v2
+    -
+      name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    -
+      name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+    -
+      name: Build Image
+      uses: docker/build-push-action@v2
+      with:
+        builder: ${{ steps.buildx.outputs.name }}
+        context: ci/
+        file: ./docker/Dockerfile
+        build-args: BASE_IMAGE=ubuntu:latest
+        platforms: linux/amd64
+        # platforms: linux/amd64,linux/arm64
+        push: false
+        tags: otel-cpp
+        load: true
+        cache-from: type=gha
+        # cache-to: type=gha,mode=max
+    -
+      name: Save Image
+      run: |
+        docker images
+        docker save -o /opt/otel-cpp-deps.tar otel-cpp
+    -
+      name: Upload Image
+      uses: actions/upload-artifact@v3
+      with:
+        name: otel-cpp-deps
+        path: /opt/otel-cpp-deps.tar
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -30,7 +30,7 @@ jobs:
         platforms: linux/amd64
         # platforms: linux/amd64,linux/arm64
         push: false
-        tags: otel-cpp
+        tags: otel-cpp-deps
         load: true
         cache-from: type=gha
         # cache-to: type=gha,mode=max
@@ -38,7 +38,7 @@ jobs:
       name: Save Image
       run: |
         docker images
-        docker save -o /opt/otel-cpp-deps-debian.tar otel-cpp
+        docker save -o /opt/otel-cpp-deps-debian.tar otel-cpp-deps
     -
       name: Upload Image
       uses: actions/upload-artifact@v3
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -32,8 +32,6 @@ jobs:
         push: false
         tags: otel-cpp-deps
         load: true
-        cache-from: type=gha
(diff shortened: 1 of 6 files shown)


.github/workflows/dependencies_image.yml
```diff
@@ -0,0 +1,47 @@
...
+        push: false
+        tags: otel-cpp
+        load: true
+        cache-from: type=gha
+        # cache-to: type=gha,mode=max
+    -
+      name: Save Image
+      run: |
+        docker images
+        docker save -o /opt/otel-cpp-deps.tar otel-cpp
```
lalitb, line 41: nit - We can have platform/distribution specific artifacts, maybe we can name it accordingly (as a separate PR if this is not straightforward to achieve)

.github/workflows/dependencies_image.yml
```diff
@@ -0,0 +1,47 @@
...
+    -
+      name: Upload Image
+      uses: actions/upload-artifact@v3
+      with:
+        name: otel-cpp-deps
+        path: /opt/otel-cpp-deps.tar
```
lalitb, line 47: Do we have the estimated size of these tar-ed docker images, as GitHub cache is limited to 10GB, and we are already using some of it for Bazel builds? We can later also see if these artifacts can be stored separately in azure blob - 
https://github.com/marketplace/actions/upload-a-build-artifact-to-azure-blob-storage

esigo: I didn't find the size limit in the [documentation](https://github.com/actions/upload-artifact/issues/9). I think the same rule as logs should apply here. I'm going to reduce the retention period  for the artifact to two weeks.

The size of the artifact is about [1GB](https://github.com/esigo/opentelemetry-cpp/actions/runs/2181104518).

We can reduce this size (probably to 200MB) by storing `gRPC` and `thrift` only in the image.

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN ./setup_grpc.sh -v GRPC_VERSION
+
+FROM base as thrift
+RUN apt-get install -y --no-install-recommends wget
+
+# install thrift
+ARG THRIFT_VERSION=0.14.1
+ADD setup_thrift.sh .
+RUN ./setup_thrift.sh
+
```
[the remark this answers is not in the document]
esigo: Let's add it as separate PR.

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG BASE_IMAGE=ubuntu:latest
+ARG CORES=${nproc}
+
+FROM ${BASE_IMAGE} as base
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y build-essential autoconf \
```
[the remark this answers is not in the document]
esigo: I made it debian only

.github/workflows/dependencies_image.yml
esigo: Let's do it in a separate PR, I need to see if that's possible.

.github/workflows/dependencies_image.yml
lalitb: Ok interesting. I also don't see it is documented too. I was thinking it to be part of Cache size limit.
What all we store right now apart from gRPC and thrift ?

docker/Dockerfile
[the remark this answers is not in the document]
lalitb: ok got it.

.github/workflows/dependencies_image.yml
esigo: the image had everything for building the library git, cmake,...
I changed the docker file. now it has only `gRPC` and `thrift`.

.github/workflows/dependencies_image.yml
```diff
@@ -0,0 +1,48 @@
...
+        builder: ${{ steps.buildx.outputs.name }}
+        context: ci/
+        file: ./docker/Dockerfile
+        build-args: BASE_IMAGE=ubuntu:latest
+        platforms: linux/amd64
+        # platforms: linux/amd64,linux/arm64
+        push: false
+        tags: otel-cpp-deps
+        load: true
+        cache-from: type=gha
```
lalitb, line 35: Is this effective if `cache-to` is not used ?

lalitb (approved):
LGTM. Thanks for adding separate dependency image.

.github/workflows/dependencies_image.yml
esigo: no it's not, I cleaned it.

PR #1346, opened 2022-04-22
Fix output time in metrics OStream exporter

### LINKED ISSUE #1345: Metrics OStream exporter prints time into local time zone
The OStream metrics exporter prints time to string based on local time zone, which fails tests like `OStreamMetricsExporter.ExportDropPointData` with below message.

```console
391: With diff:
391: @@ -3,6 +3,6 @@
391:    schema url\t:
391:    version\t: 1.2.0
391: -  start time\t: Wed Dec 31 16:00:00 1969
391: -  end time\t: Wed Dec 31 16:00:00 1969
391: +  start time\t: Thu Jan  1 00:00:00 1970
391: +  end time\t: Thu Jan  1 00:00:00 1970
391:    description\t: description
391:    unit\t\t: unit
391:
```

**Steps to reproduce**
Run test `OStreamMetricsExporter.ExportDropPointData` on a machine not in UTC timezone.

The test should pass.

The test failed with above output log.

**Additional context**

ThomsonTan:
Fixes #1345

## Changes

Print time to string in UTC instead of of current local time zone.


esigo (commented):
Can we use [strftime](https://en.cppreference.com/w/cpp/chrono/c/strftime) and [gtime_s](https://en.cppreference.com/w/c/chrono/gmtime) here?

ThomsonTan:
> Can we use [strftime](https://en.cppreference.com/w/cpp/chrono/c/strftime) and [gtime_s](https://en.cppreference.com/w/c/chrono/gmtime) here?

Yes, we can. Is there concern here that `std::asctime` not thread-safe?

esigo:
> > Can we use [strftime](https://en.cppreference.com/w/cpp/chrono/c/strftime) and [gtime_s](https://en.cppreference.com/w/c/chrono/gmtime) here?
> 
> Yes, we can. Is there concern here that `std::asctime` not thread-safe?

Yes it's not thread safe.

owent:
> Can we use [strftime](https://en.cppreference.com/w/cpp/chrono/c/strftime) and [gtime_s](https://en.cppreference.com/w/c/chrono/gmtime) here?

`gmtime_s` is not available on some environment we support. I think we can use `gmtime_s` for MSVC and `gmtime` for the rest.

lalitb:
We can also print epoch time as done by log and span exporter:
https://github.com/open-telemetry/opentelemetry-cpp/blob/bf8f433cae41ad682cf9e390650158a2751373c1/exporters/ostream/src/span_exporter.cc#L77

lalitb (approved):
LGTM, minor comment- if we can keep it consistent with how time is printed in traces and logs.

lalitb:
> > LGTM, minor comment- if we can keep it consistent with how time is printed in traces and logs.
> 
> Do you mean using the same style, like std::gmtime_s and std::strftime?

Sorry was not clear, as mentioned here:

https://github.com/open-telemetry/opentelemetry-cpp/pull/1346#issuecomment-1106246663

ThomsonTan:
> > > LGTM, minor comment- if we can keep it consistent with how time is printed in traces and logs.
> > 
> > 
> > Do you mean using the same style, like std::gmtime_s and std::strftime?
> 
> Sorry was not clear, as mentioned here:
> 
> [#1346 (comment)](https://github.com/open-telemetry/opentelemetry-cpp/pull/1346#issuecomment-1106246663)

Makes sense, file issue https://github.com/open-telemetry/opentelemetry-cpp/issues/1361 for tracking.

PR #1363, opened 2022-04-29
Disable test on prometheus-cpp which not need

ThomsonTan:
The Prometheus exporter could build with submodule prometheus-cpp, but this submodule defines its own GTest::gmock_main target which could cause conflict with google test referenced by opentelemetry-cpp. Fixes this by disabling test in prometheus-cpp as it is not needed in this repo.

Also run the current premetheus exporter test only for `WITH_METRICS_PREVIEW`.

## Changes


CMakeLists.txt
```diff
@@ -361,7 +364,8 @@ if(BUILD_TESTING)
         ${CMAKE_BINARY_DIR}/lib/libgmock.a)
   elseif(WIN32)
     # Make sure we are always bootsrapped with vcpkg on Windows
-    find_package(GTest)
+    # set(CMAKE_FIND_DEBUG_MODE TRUE)
```
esigo, line 367: do we need this?

ThomsonTan: > do we need this?

Thanks for catching, this comment was used for local debugging, removed it.

esigo (approved):

PR #1334, opened 2022-04-18
Enable metric collection for Async Instruments - Delta and Cumulative 

lalitb:
Fixes #1236 #1234 

## Changes

 AsyncMetricStorage invokes the instrumented observation callback and waits for it to be done. After the observation completes, it resets the active storage stash and builds metric data for the metric instrument.
Also created TemporalMetricStorage inspired from [otel-java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/TemporalMetricStorage.java) to store the history of metrics. This TemporalMetricStorage would be later re-used for synchronous storage too.

@ThomsonTan @esigo - Updated the description, so should be good to review now. Thanks

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1369, opened 2022-05-04
Reuse TemporalMetricStorage for both SyncMetricStorage and AsyncMetricStorage

lalitb:
## Changes

TemporalMetricStorage was introduced as part of #1334 to store metrics history ( unreported and last reported metrics data). This can be reused for SyncMetricStorage, as the same logic is used to populate these metrics data in both sync and async storage.


esigo (approved):
LGTM
thanks for the PR :)

PR #1373, opened 2022-05-05
Fix empty tracestate header propagation

### LINKED ISSUE #1366: Empty headers injected by HTTP and Baggage propagator
If the HTTP propagator is used on an invalid context or the baggage propagator is used on an empty baggage, the resulting header map will contain empty values. I don't know if this is intended or not, but I think it'd be more user friendly if empty values wouldn't be injected, so the user wouldn't have to remove or check them manually before using them as HTTP request headers.

yzsolt:
Fixes #1366

## Changes

- Fixed Tracestate propagation tests
- Fixed empty Tracestate header propagation


lalitb (approved):
LGTM. Thanks for the fix.

ThomsonTan (approved):

esigo (approved):

PR #1374, opened 2022-05-05
Add noexcept/const qualifier at missing places for Trace API.

lalitb:
## Changes

This allows these methods to be used in the functions which are noexcept without explicit exception handling.

We may have to look at similar changes for log/metrics api separately.

### DIFF
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -14,7 +14,7 @@ jobs:
       uses: actions/checkout@v2
     -
       name: Set up QEMU
-      uses: docker/setup-qemu-action@v1
+      uses: docker/setup-qemu-action@v2
     -
       name: Set up Docker Buildx
       id: buildx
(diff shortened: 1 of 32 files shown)


ThomsonTan:
Just curious whether adding noexcept could break existing API usage in any case.

lalitb:
> Just curious whether adding noexcept could break existing API usage in any case.

I don't think it should break API/ABI compatibility. Though reverse ( removing noexcept) can break things.

reyang:
> > Just curious whether adding noexcept could break existing API usage in any case.
> 
> I don't think it should break API/ABI compatibility. Though reverse ( removing noexcept) can break things.

IIRC noexcept is not included in the mangled name but not sure if it's compiler specific or covered by the standard spec. From semantic standpoint +1 with what @lalitb said.

reyang (approved):

lalitb:
> IIRC noexcept is not included in the mangled name but not sure if it's compiler specific or covered by the standard spec.

Validated it across gcc, clang and msvc and it generates the same mangled name for both these prototypes:
```cpp
void foo1()
void foo2() noexcept
```
msvc: https://gcc.godbolt.org/z/sb7szM8xd
gcc: https://gcc.godbolt.org/z/17fovvd37
clang: https://gcc.godbolt.org/z/Pebcox9xz

ThomsonTan (approved):

PR #1378, opened 2022-05-06
Prometheus exporter meters and instrument name

### LINKED ISSUE #1370: [Metrics] Prometheus exporter reports instrumentation library name instead of instrument name
Greetings,

I am attempting to export to Prometheus from an application based on the Prometheus example.
WITH_METRICS_PREVIEW is undefined.

Running the example binary appears to work, and reports a metric named `prometheus_metric_example`, type counter,
for an instrument named `prometheus_metric_example_counter` in the C++ code.

Running my application,
No matter how instruments or views are named, the data exported to Prometheus always reports the same metric name,
which is the name given as instrumentation library when calling GetMeter().

Should the prometheus export code use `metric_data.instrument_descriptor.name_` instead of `instrumentation_info.instrumentation_library_->GetName()` ?

Also, there is no HELP exported to Prometheus.

The following change helps to go further ...

Regards.

```
diff --git a/exporters/prometheus/src/exporter_utils.cc b/exporters/prometheus/src/exporter_utils.cc
index 39131e2..0b7a26f 100644
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -82,11 +82,17 @@ void PrometheusExporterUtils::SetMetricFamilyByAggregator(
 {
   for (const auto &instrumentation_info : data.instrumentation_info_metric_data_)
   {
+#ifdef NEVER
     auto origin_name    = instrumentation_info.instrumentation_library_->GetName();
     auto sanitized      = SanitizeNames(origin_name);
     metric_family->name = sanitized;
+#endif
     for (const auto &metric_data : instrumentation_info.metric_data_)
     {
+      auto origin_name    = metric_data.instrument_descriptor.name_;
+      auto sanitized      = SanitizeNames(origin_name);
+      metric_family->name = sanitized;
+      metric_family->help = metric_data.instrument_descriptor.description_;
       auto time = metric_data.start_ts.time_since_epoch();
       for (const auto &point_data_attr : metric_data.point_data_attr_)
       {
```

### LINKED ISSUE #1372: [Metrics] Prometheus exporter only works for one Meter with one instrument
Greetings

This is a follow up on:

 [Metrics] Prometheus exporter reports instrumentation library name instead of instrument name #1370 

In method `PrometheusExporterUtils::SetMetricFamilyByAggregator()`, the code contains an outer loop:

`  for (const auto &instrumentation_info : data.instrumentation_info_metric_data_)`

This loop is executed once per Meter, per method `MetricCollector::Collect()`,
so when exporting data for multiple meters, all the data will be written to the same 
  `prometheus_client::MetricFamily *metric_family`
output, which seems incorrect.

Also, even when considering a single Meter,
the next loop:

`    for (const auto &metric_data : instrumentation_info.metric_data_)`

considers all the instruments exported by a Meter.

Again, this loop is executed once per instrument, resulting in all metrics data exported to the same metric_family object,
which seems incorrect.
A prometheus `prometheus_client::MetricFamily` object has only one name and description, so it should map to one otel metric, not N (in my understanding).

Multiple metrics from the same meter will be mixed together in the output.

When testing, indeed, with a Meter with N instruments, the prometheus exporter only exposes 1 instrument (the last one considered).
I did not inspect the measurements values exported, but suspects this is a union of all the measurements seen.

I would like to suggest the following changes, to evaluate and discuss (I don't know the code well enough yet):

- add a method to compute the number of `prometheus_client::MetricFamily` to export
- use it from `PrometheusExporterUtils::TranslateToPrometheus` to size the output correctly
- abandon the `prometheus_client::MetricFamily *metric_family` parameter from `PrometheusExporterUtils::SetMetricFamily()` and `PrometheusExporterUtils::SetMetricFamilyByAggregator()`
- instead, pass the whole `std::vector<prometheus_client::MetricFamily>` output as parameter
- In `PrometheusExporterUtils::SetMetricFamilyByAggregator`, let the code append more metric_family in the output vector as the code sees fit.
- Minor renaming, as "SetMetricFamillyxxx" no longer makes sense without a metric family output parameter.

Regards.

esigo:
Fixes #1372 (issue) and Fixes #1370 (issue)

## Changes


marcalff:
Thanks for the fixes. I will resume Metrics testing once merged.

esigo:
> Thanks for the fixes. I will resume Metrics testing once merged.

@marcalff thanks for your interest and thanks for testing :)

sdk/src/metrics/meter.cc
```diff
@@ -188,10 +188,24 @@ std::unique_ptr<WritableMetricStorage> Meter::RegisterMetricStorage(
...
       instrument_descriptor, *instrumentation_library_,
       [this, &instrument_descriptor, &storages](const View &view) {
-        auto view_instr_desc         = instrument_descriptor;
-        view_instr_desc.name_        = view.GetName();
-        view_instr_desc.description_ = view.GetDescription();
-        auto storage                 = std::shared_ptr<SyncMetricStorage>(new SyncMetricStorage(
+        auto view_instr_desc = instrument_descriptor;
+        if (view.GetName().empty())
+        {
+          view_instr_desc.name_ = instrument_descriptor.name_;
```
lalitb, line 194: nit - do we need this, as line 191 already assigns `instrument_descriptor` ?

sdk/src/metrics/meter.cc
```diff
@@ -188,10 +188,24 @@ std::unique_ptr<WritableMetricStorage> Meter::RegisterMetricStorage(
...
+        }
+        else
+        {
+          view_instr_desc.name_ = view.GetName();
+        }
+        if (view.GetDescription().empty())
+        {
+          view_instr_desc.description_ = instrument_descriptor.description_;
```
lalitb, line 202: same here too.

lalitb (approved):
LGTM, with minor comments. Thanks for fixing it.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -34,28 +34,52 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
...
   {
-    SetMetricFamily(*r, &output[i]);
-    i++;
+    for (const auto &instrumentation_info : r->instrumentation_info_metric_data_)
+    {
+      for (const auto &metric_data : instrumentation_info.metric_data_)
+      {
+        auto origin_name = metric_data.instrument_descriptor.name_;
+        auto sanitized   = SanitizeNames(origin_name);
+        std::puts(sanitized.c_str());
```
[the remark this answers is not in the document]
esigo: it is not, cleaned.

sdk/src/metrics/meter.cc
esigo: oh yeah, thanks fixed.

sdk/src/metrics/meter.cc
esigo: thanks, fixed.

Issue #1291, opened 2022-03-28
Review logger Implementation for the stable log data model.


lalitb:
Log Data Model specs are marked as stable - https://github.com/open-telemetry/opentelemetry-specification/pull/2387. It would be a good time to review the logger api and sdk implementation and validate if the existing log structure can be mapped to this data model. Also, how to continue supporting fields/attributes that are not part of this logger model - eg log-name.

owent:
I have a quick look at [opentelemetry-proto to v0.17.0][1] and the [`main` branch of logging-library-sdk in opentelemetry-specification][2], and I want to confirm and disscuss the changes.

+ `InstrumentationLibrary` is deprecated  and we should use `InstrumentationScope` now.This will also changes the trace and metrics. Should we add a option to switch between legacy `InstrumentationLibrary` and new `InstrumentationScope`?
  + `instrumentation_library_spans` should be replaced by `scope_spans`
  + `instrumentation_library_metrics` should be replaced by `scope_metrics`
  + `instrumentation_library_logs` should be replaced by `scope_logs`
+ `LogProcessor::OnReceive(LogRecordable)` should be renamed to `LogProcessor::Emit(LogData)`
+ Rename `LoggerProvider` to `LogEmitterProvider`
+ Rename `LoggerContext` to `LogEmitter` and move `InstrumentationScope` from `Logger` to `LogEmitter`
+ Rename `LogRecord` to `LogData`

Am I missed anything?

[1]: https://github.com/open-telemetry/opentelemetry-proto/blob/v0.17.0/opentelemetry/proto/logs/v1/logs.proto
[2]: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/logging-library-sdk.md

esigo:
Hi @owent, thanks for looking into this.
I'd vote for not having a switch for supporting legacy.

owent:
Without supporting legacy. There will be break changes to SDK and the output of OTLP exporters.Especially the trace APIs are already stable, is it acceptable?

esigo:
Sure we will have breaking changes here. I'd still vote for not supporting legacy. We are already low on resources, and supporting both legacy and new proto will make maintenance more complicated for us. We can discuss this in the next SIG meeting. I'll update you.
cc @lalitb @ThomsonTan

lalitb:
Thanks, @owent for summarising the changes required.

[...]

PR #1383, opened 2022-05-08
Upgrade proto to v0.17.0, update log data model

owent:
This PR fix part of #1291

## Changes

+ Upgrade proto to v0.17.0
+ Optimize script to dump third_party_release
+ Update the logs APIs to meet current [logs data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md)
+ Using tools/vcpkg in CI scripts


api/include/opentelemetry/logs/logger.h
```diff
@@ -57,7 +56,6 @@ class Logger
    * in order to create a log record.
    */
   virtual void Log(Severity severity,
-                   nostd::string_view name,
```
lalitb, line 60: I may be selfish here for my internal customers, even though we agreed to make it a breaking change, do you think keep `name` as something macro-enabled or deprecated or add separate methods without `name`? If it is too much of a change, for now, we can keep it as it is, and I will raise a separate PR on top of this to add `name` as deprecated/macro-enabled.

exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
```diff
@@ -163,7 +161,7 @@ class Logger : public opentelemetry::logs::Logger
         ActivityIdPtr = &ActivityId;
       }
     }
-    evt[ETW_FIELD_PAYLOAD_NAME]              = std::string(name.data(), name.length());
+    evt[ETW_FIELD_PAYLOAD_NAME]              = std::string(provId.data(), provId.length());
```
lalitb, line 164: Payload name should ideally be the property of each Log, and not the Logger. But I understand we don't have much option without `name` field.

api/include/opentelemetry/logs/logger.h
lalitb: Please ignore this comment, going through the PR I think it would need some management effort to keep both the APIs available.

owent: Fine by me to leave the Logger APIs with name for one or two versions, but the new proto has removed name field of `LogRecord` , so we will just ignore the parameter and just leave some time for users to migrate codes.

exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
owent: What's your suggestion here? I'm not familiar with ETW, and I don't know what should be in ETW data.

api/include/opentelemetry/logs/logger.h
lalitb: Yes, it would be awesome if we can keep the Logger APIs unchanged ( for the `name` parameter ) for some time so as to not break instrumentation library compilation :)

exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
lalitb: If we are not changing the Logger API, do you think we can keep the `name` in the ETW.

owent: I think we can keep `name` temporary ,but we should remove `name` in opentelemetry-api eventually, or it will confuse users.

api/include/opentelemetry/logs/logger.h
owent: APIs with `name` are restored.

lalitb (approved):
thanks for the changes. nicely done.

exporters/otlp/src/otlp_log_recordable.cc
```diff
@@ -7,6 +7,14 @@
...
 #  include "opentelemetry/exporters/otlp/otlp_recordable_utils.h"
 
+#  ifndef likely
+#    ifdef __GNUC__
+#      define likely(x) __builtin_expect(!!(x), 1)
+#    else
+#      define likely(x) !!(x)
+#    endif
+#  endif
+
```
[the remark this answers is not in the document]
owent: Sure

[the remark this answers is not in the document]
owent: > nit - moving it to somewhere more common - maybe `include/opentelemetry/common/macros.h` ?

This is the legacy usage of likely, according to https://en.cppreference.com/w/cpp/language/attributes/likely . The `likely` attribute is declared in the **action** statement, not the **condition** statement like before. 
I also want to discuss about how to use `likely` to support both legacy usage and C++20 usage and keep it simple?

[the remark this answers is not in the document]
lalitb: Do you mean using macro like this : https://github.com/boostorg/outcome/blob/e3556c319d8c2751d6abc2a85fbaab286e1f965e/include/boost/outcome/try.hpp#L173

```cpp
#ifndef BOOST_OUTCOME_TRY_LIKELY_IF
#if(__cplusplus >= 202000L || _HAS_CXX20) && (!defined(__clang__) || __clang_major__ >= 12) && (!defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 9)
#define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__VA_ARGS__) [[likely]]
#elif defined(__clang__) || defined(__GNUC__)
#define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__builtin_expect(!!(__VA_ARGS__), true))
#else
#define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__VA_ARGS__)
#endif
#endif
```

Do you think we will see substantial performance optimization with these branch predictions?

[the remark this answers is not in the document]
owent: > Do you mean using macro like this : https://github.com/boostorg/outcome/blob/e3556c319d8c2751d6abc2a85fbaab286e1f965e/include/boost/outcome/try.hpp#L173
> 
> ```c++
> #ifndef BOOST_OUTCOME_TRY_LIKELY_IF
> #if(__cplusplus >= 202000L || _HAS_CXX20) && (!defined(__clang__) || __clang_major__ >= 12) && (!defined(__GNUC__) || defined(__clang__) || __GNUC__ >= 9)
> #define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__VA_ARGS__) [[likely]]
> #elif defined(__clang__) || defined(__GNUC__)
> #define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__builtin_expect(!!(__VA_ARGS__), true))
> #else
> #define BOOST_OUTCOME_TRY_LIKELY_IF(...) if(__VA_ARGS__)
> #endif
> #endif
> ```
> 
> Do you think we will see substantial performance optimization with these branch predictions?

This macro is limited to `if (condition)` statement.I wonder is there better way to achieve this base on your experience?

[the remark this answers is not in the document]
lalitb: Ah, support both c++20 format and legacy, that would be tricky I guess. Will think if there is an easy way to support it.

sdk/src/logs/logger_provider.cc
```diff
@@ -88,9 +88,20 @@ nostd::shared_ptr<opentelemetry::logs::Logger> LoggerProvider::GetLogger(
   }
   */
 
-  // If no logger with that name exists yet, create it and add it to the map of loggers
-  auto lib = instrumentationlibrary::InstrumentationLibrary::Create(library_name, library_version,
-                                                                    schema_url);
+  // For log sources which define a logger name (e.g. Java Logger Name) the Logger Name SHOULD be
```
ThomsonTan, line 91: nit: put a link to the spec here instead of copy the text?

ThomsonTan (approved):

sdk/src/logs/logger_provider.cc
owent: Done

owent:
> @owent Can you please rebase this to enable merge. Thanks.

Done

PR #1385, opened 2022-05-08
[WIP] protobuf config search mode

### LINKED ISSUE #1384: OpenTelemetry-cpp superbuild

If so, provide a concise description of the problem.
I have a piece of cmake that builds OpenTelemetry-cpp:
``` cmake
ExternalProject_Add(
  opentelemetry
  DEPENDS grpc
  GIT_REPOSITORY https://github.com/open-telemetry/opentelemetry-cpp.git
  GIT_TAG v1.3.0
  GIT_SHALLOW 1
  UPDATE_COMMAND ""
  CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
             -DCMAKE_BUILD_TYPE=Release
             -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE
             -DWITH_ZIPKIN=OFF
             -DWITH_JAEGER=OFF
             -DWITH_EXAMPLES=OFF
             -DCMAKE_INSTALL_PREFIX=/opt/third_party/install
             -DBUILD_TESTING=OFF
             -DWITH_OTLP=ON
             -DBUILD_SHARED_LIBS=ON
             -DProtobuf_DIR=${STAGED_INSTALL_PREFIX}/lib/cmake/protobuf
             -Drotobuf_DIR=${STAGED_INSTALL_PREFIX}/lib/cmake/protobuf
             -DgRPC_DIR=${grpc_DIR}
             -Dabsl_DIR=${STAGED_INSTALL_PREFIX}/lib/cmake/absl
  CMAKE_CACHE_ARGS -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS}
  TEST_AFTER_INSTALL 0
  DOWNLOAD_NO_PROGRESS 1
  LOG_CONFIGURE 1
  LOG_BUILD 0
  LOG_INSTALL 1)
```

Where grpc is another external project (similar to #1382). The build fails using this cmake because we use find_package in module mode which doesn't find grpc/protobuf because of name mismatch. This can be solved if we use config search mode first.

failure:
```
  Found package configuration file:

    build/stage/lib/cmake/grpc/gRPCConfig.cmake

  but it set gRPC_FOUND to FALSE so package "gRPC" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing:
  protobuf::libprotobuf protobuf::libprotoc
```

What do you want to happen instead? What is the expected behavior?
We can use this:
``` cmake
  find_package(Protobuf QUIET CONFIG)
  if((NOT Protobuf_FOUND AND NOT PROTOBUF_FOUND))
    find_package(Protobuf)
  endif()
```

**Additional context**

esigo:
Fixes #1384 (issue)

## Changes


CMakeLists.txt
```diff
@@ -292,10 +292,10 @@ endif()
 
 if(WITH_OTLP)
   set(protobuf_MODULE_COMPATIBLE ON)
-  find_package(Protobuf)
+  find_package(Protobuf PATHS ${EXTERNAL_PROTOBUF_DIR})
```
owent, line 295: Why we need this? External protobuf and grpc can be found by add installation path into [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) and [CMAKE_MODULE_PATH ](https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html#variable:CMAKE_MODULE_PATH),or by setting `Protobuf_DIR/Protobuf_ROOT` and `gRPC_DIR/gRPC_ROOT` ?

> See https://cmake.org/cmake/help/latest/command/find_package.html#id4 for details

esigo: Thanks @owent, I used your [suggestion](https://github.com/open-telemetry/opentelemetry-cpp/issues/1384#issuecomment-1120574543) on the issue, and it works without my changes.
I have another issue now. The files for this library https://github.com/open-telemetry/opentelemetry-cpp/blob/54abc2741b10efef07f786bb5200fd9112b191bb/cmake/opentelemetry-proto.cmake#L191-L203 are not available when calling make, so the first make try fails and if I repeat it the build goes through. Do you have any suggestion to solve this issue?
otherwise I have to close this PR.

owent: Sorry, I don't really get the point about the files are not available when calling make. These files are generated by `add_custom_command(OUTPUT [files]` above.
https://github.com/open-telemetry/opentelemetry-cpp/blob/54abc2741b10efef07f786bb5200fd9112b191bb/cmake/opentelemetry-proto.cmake#L163
This command should run when OUTPUT files are expired or not found and when they are depended to build any target.
If you are writing a standalone Makefile, I think it's required to write a similar TARGET of Makefile to generate these files.

esigo: The problem is that, the files are not there and custome_command is ran on build time not config time. on build time when I use 16 cores to build most of the times the PB files are not generated before the compile of proto library. I think this can be solved by using execute_process.

CMakeLists.txt
```diff
@@ -292,10 +292,10 @@ endif()
...
   if(WITH_OTLP_GRPC OR (NOT DEFINED WITH_OTLP_GRPC AND NOT DEFINED
                                                        CACHE{WITH_OTLP_GRPC}))
-    find_package(gRPC)
+    find_package(gRPC PATHS ${EXTERNAL_GRPC_DIR})
```
ThomsonTan, line 298: Haven't seen where `${EXTERNAL_GRPC_DIR}` is defined?

CMakeLists.txt
owent: Why we need the generated files at config time?If we want a standalone target to just generate these files, but not compile them. I think the better way is adding a custom target by `add_custom_target(generate_opentelemetry_proto SOURCES [source files...])` and run `cmake --build . --target generate_opentelemetry_proto` after configure is done.

esigo: thanks @owent. I'm going to close this PR.

CMakeLists.txt
esigo: this was not defined. I' going to close the PR as the changes are not needed.

PR #1386, opened 2022-05-09
Fix sharing resource in batched exported spans

### LINKED ISSUE #1307: Attempt to create new Tracer changes name and version of existing Tracers
**Describe your environment**
CentOS 8
gcc 10.3.1
opentelemetry-cpp built from main branch on March 25th, configured using this command:
cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP=ON -DWITH_OTLP_GRPC=OFF -DWITH_ZPAGES=ON -DCMAKE_INSTALL_PREFIX=/usr ..
App code is compiled using C++11.

**Steps to reproduce**
Code creates TracerProvider using code from samples, creates two different Tracers and uses them to create some spans:

```cpp
auto tracer_provider = opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider>
    (new opentelemetry::sdk::trace::TracerProvider(std::move(batch_processor), resource, std::move(always_on_sampler)));
opentelemetry::trace::Provider::SetTracerProvider(tracer_provider);

auto tracer1 = opentelemetry::trace::Provider::GetTracerProvider()->GetTracer("Lib1", "1.0");
auto tracer2 = opentelemetry::trace::Provider::GetTracerProvider()->GetTracer("Lib2", "1.1");

auto span1 = tracer1->StartSpan("Span1");
auto span2 = tracer2->StartSpan("Span2");
span1->End();
span2->End();
```

Both spans have `otel.library.name` and `otel.library.version` set to values provided when parent Tracer is created.

All spans have `otel.library.name` and `otel.library.version` set to the same values. It seems that pair with lexicographically smallest name wins - first I created 2 Tracers with names "Ne..." and "NS..." and "Ne..." one won. Then I added 3rd "C..." and this one was selected.

ThomsonTan:
Fixes #1307.

## Changes

Currently the change assumes each span in the batch can come from a different tracer, we may do some optimization on this like adding a hash table to check whether the resource/schema_url from current span are the same from previous one.


owent (approved):
LGTM

exporters/otlp/src/otlp_recordable_utils.cc
```diff
@@ -259,25 +259,19 @@ void OtlpRecordableUtils::PopulateRequest(
...
   }
 
-  auto resource_span       = request->add_resource_spans();
-  auto instrumentation_lib = resource_span->add_instrumentation_library_spans();
-  bool first_pass          = true;
-
   for (auto &recordable : spans)
   {
     auto rec = std::unique_ptr<OtlpRecordable>(static_cast<OtlpRecordable *>(recordable.release()));
+    auto resource_span       = request->add_resource_spans();
```
lalitb, line 265: Can resource_span be outside the loop? Otel-cpp doesn't support multiple TracerProviders, so the resources would be always the same for all the events.

lalitb (approved):
Looks good, with minor comment on resource initialisation.

exporters/otlp/src/otlp_recordable_utils.cc
ThomsonTan: @lalitb, could otel-cpp create multiple `TracerProvider`, and set each of them as the global one in alternative?

lalitb: Yes possible. Have approved it as it was just a nit comment :) should be good to merge.

PR #1388, opened 2022-05-10
Connect async storage with async instruments

### LINKED ISSUE #1387: [Metrics] ASync Instruments not fully implemented.
Greetings,

I am a bit confused about the development status for async instruments in Metrics ...

There are no remaining issues left listed in the "Metrics SDK - Alpha release" milestone, great.

There are 5 unrelated open items in "Metrics SDK - Beta Release", and issue #1338 is about a more complex use case (observing N metrics at once), so this gives the impression that a basic Observable counter is already implemented.

If this is not supported yet, feel free to close this issue as a duplicate.
If supported, then it seems some bits are missing in the implementation:

Using `#undef ENABLE_METRICS_PREVIEW`

The method `Meter::CreateLongObservableCounter()` is available,
and builds a `LongObservableCounter` object, ok.

The constructors for:

- `LongObservableCounter`
- `opentelemetry::metrics::ObservableCounter<long>`
- `Asynchronous<long>`
- `opentelemetry::metrics::AsynchronousInstrument`

keep the instrument name + callback + description + unit,
but never build an instrument descriptor with it.

In the code path of `Meter::CreateLongObservableCounter()`,
the counter returned is not registered with any AsyncStorage,
and nothing keeps a reference to all the observable counters created,
so the ::Collect methods will never invoke callbacks on async instruments.

For synchronous instruments, methods like `Meter::CreateLongCounter()` build an `InstrumentDescriptor` object and call `RegisterMetricStorage()`, so I would expect some similar code, but found none.

Note that the class `AsyncMetricStorage` exists, but is never used in the production code (there is one use only, but from a unit test).

While testing, the callback function given to Meter::CreateLongObservableCounter() is never called,
and async instruments are not exported (tested with the Prometheus exporter)

What am I missing (or not understanding) ?

Regards.

lalitb:
Fixes #1387 

## Changes

While we do support creating async storage separately, it needs to be connected to the async instruments. 
Also added it in the ostream exporter example.


marcalff (commented):
Do we need something similar to RegisterMetricStorage here:

        auto view_instr_desc = instrument_descriptor;
        if (!view.GetName().empty())
        {
          view_instr_desc.name_ = view.GetName();
        }
        if (!view.GetDescription().empty())
        {
          view_instr_desc.description_ = view.GetDescription();
        }

For empty view names and description, the metric exported should use the name and description of the underlying instrument.

ThomsonTan (approved):

PR #1392, opened 2022-05-13
Fix vcpkg package name in doc

ThomsonTan:
Fixes # (issue)

## Changes


reyang (approved):

docs/building-with-vcpkg.md
```diff
@@ -13,7 +13,7 @@ source release of OpenTelemetry C++ SDK.
 The following command can be used to install the public open source release:
 
 ```console
-vcpkg install opentelemetry
+vcpkg install opentelemetry-cpp
```
lalitb, line 16: I think it should be 
`vcpkg install opentelemetry`. 
This documentation talks about the installing the opentelemetry port which comes as part of this repo.

The installation for opentelemetry port available in vcpkg registry is described here - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#using-package-managers

ThomsonTan: @lalitb it makes sense. To install our port in this repo, do we need provide some extra option to vcpkg for our overlay, I think just `vcpkg install opentelemetry` would not work. Thoughts?

lalitb: Yes, right it won't work without the overlay option. I think we can keep changes from this PR.

lalitb (approved):

Issue #1220, opened 2022-02-18
crash in ext::http::client::curl::HttpOperation::Send


ZDzhao:
Enviroment: CentOS Linux release 7.3.1611 (Core)

#0  0x00007ff6ee439a37 in abort () from /lib64/libc.so.6
#1  0x00007ff6ee477f47 in __libc_message () from /lib64/libc.so.6
#2  0x00007ff6ee512d87 in __fortify_fail () from /lib64/libc.so.6
#3  0x00007ff6ee510f40 in __chk_fail () from /lib64/libc.so.6
#4  0x00007ff6ee512cf7 in __fdelt_warn () from /lib64/libc.so.6
#5  0x0000000000d55ed8 in WaitOnSocket (timeout_ms=<optimized out>, for_recv=0, sockfd=3175)
    at bazel-out/k8-opt/bin/external/io_opentelemetry_cpp/ext/_virtual_includes/headers/opentelemetry/ext/http/client/curl/http_operation_curl.h:460
#6  opentelemetry::v1::ext::http::client::curl::HttpOperation::Send (this=0x7ff6b4c02d00)
    at bazel-out/k8-opt/bin/external/io_opentelemetry_cpp/ext/_virtual_includes/headers/opentelemetry/ext/http/client/curl/http_operation_curl.h:214
#7  0x0000000000d5a3f9 in operator() (__closure=0x7ff3e0f4fc88)
    at bazel-out/k8-opt/bin/external/io_opentelemetry_cpp/ext/_virtual_includes/headers/opentelemetry/ext/http/client/curl/http_operation_curl.h:298
#8  __invoke_impl<long int, opentelemetry::v1::ext::http::client::curl::HttpOperation::SendAsync(std::function<void(opentelemetry::v1::ext::http::client::curl::HttpOperation&)>)::<lambda()> > (__f=<optimized out>)
    at /usr/local/gcc-7.5/lib/gcc/x86_64-pc-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/invoke.h:60
#9  __invoke<opentelemetry::v1::ext::http::client::curl::HttpOperation::SendAsync(std::function<void(opentelemetry::v1::ext::http::client::curl::HttpOperation&)>)::<lambda()> > (__fn=<optimized out>) at /usr/local/gcc-7.5/lib/gcc/x86_64-pc-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/invoke.h:96
[...]

PR #1410, opened 2022-05-19
fix: WaitOnSocket select error when sockfd above FD_SETSIZE

taozhijiang:
Fixes select return code is not well handled, and when curl sockfd above FD_SETSIZE (1024), select can not handle it.

Changes:
1. According docs, select return -1 on error, 0 when timeout, and >0 for expected events return. The original cast the int code to bool, so when invalid sock provided, the procedure treat as good, the following procedure crash. So the code carefully check the select return code, and the really events we get.
2. Even though select is POSIX standard and cross platform, but can not handle the case when fd above FD_SETSIZE. In our production environment, we see CURL sock fd far more above it, so when Linux environment possible, we use poll instead of select.

### DIFF
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -489,7 +489,7 @@ protected:
       }
     }
 
-#else 
+#else
 
     struct pollfd fds[1];
     ::memset(fds, 0, sizeof(fds));
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -481,11 +481,11 @@ protected:
     {
       if (for_recv)
       {
-        res = FD_ISSET(sockfd, &infd);
+        res = (0 != FD_ISSET(sockfd, &infd));
       }
       else
       {
-        res = FD_ISSET(sockfd, &outfd);
+        res = (0 != FD_ISSET(sockfd, &outfd));
       }
     }
 
@@ -508,11 +508,11 @@ protected:
     {
       if (for_recv) 
       {
-        res = fds[0].revents & POLLIN;
+        res = (0 != (fds[0].revents & POLLIN));
       } 
       else
       {
-        res = fds[0].revents & POLLOUT;
+        res = (0 != (fds[0].revents & POLLOUT));
       }
     }
 
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -18,6 +18,7 @@
 #  include <winsock2.h>
 #else
 #  include <unistd.h>
+#  include <poll.h>
 #endif
 #include <curl/curl.h>
 
@@ -443,13 +444,19 @@ protected:
    * @param sockfd
    * @param for_recv
    * @param timeout_ms
-   * @return
+   * @return true if expected events occur, false if timeout or error happen
    */
-  static int WaitOnSocket(curl_socket_t sockfd, int for_recv, long timeout_ms)
+  static bool WaitOnSocket(curl_socket_t sockfd, int for_recv, long timeout_ms)
   {
+    bool res = false;
+
+#if defined(_WIN32)
+
+    if (sockfd > FD_SETSIZE)
+      return false;
+
     struct timeval tv;
     fd_set infd, outfd, errfd;
-    int res;
 
     tv.tv_sec  = timeout_ms / 1000;
     tv.tv_usec = (timeout_ms % 1000) * 1000;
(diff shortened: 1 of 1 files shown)


owent:
Could you please do this change base on [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) branch? The `select` call wil be removed after we merge this branch back into `main`.

lalitb:
> Could you please do this change base on [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) branch? The `select` call wil be removed after we merge this branch back into `main`.

@owent - Are these changes relevant once `async-changes` branch is merged to `main`? If not, we Should we can let these changed merged in main branch for now, and later these changes would be removed once `async-changes` branch is merged to `main`. Would leave it to you to decide :)

lalitb:
> Could you please do this change base on [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) branch? The `select` call wil be removed after we merge this branch back into `main`.

@owent - Are these changes relevant once `async-changes` branch is merged to `main`? If not, we Should we can let these changed merged in main branch for now, and later these changes would be removed once `async-changes` branch is merged to `main`. Would leave it to you to decide :)

taozhijiang:
> > Could you please do this change base on [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) branch? The `select` call wil be removed after we merge this branch back into `main`.
> 
> @owent - Are these changes relevant once `async-changes` branch is merged to `main`? If not, we Should we can let these changed merged in main branch for now, and later these changes would be removed once `async-changes` branch is merged to `main`. Would leave it to you to decide :)


This is a definite code defect which may affect the delopyment in production, so I think should be fixed ASAP.
We will try [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) latter in test environment when we are available ;-)

owent:
> relevant

Yes, the implement in `async-changes` branch will use [curl_multi_poll][1] to replace `select`. This problem may still exist because without event loop system, [curl_multi_poll][1] use `select` on some platform, but we will reuse connections for multiple sessions now, and when `max_concurrent_requests / max_requests_per_connection` is less than 1023, we will have 1024 fds at most at the same time.The default value of `max_concurrent_requests` is 64 and default value `max_requests_per_connection` is 8 now, which means will have no more than 9 connections at most by default.

[1]: https://curl.se/libcurl/c/curl_multi_poll.html

owent:
> > > Could you please do this change base on [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) branch? The `select` call wil be removed after we merge this branch back into `main`.
> > 
> > 
> > @owent - Are these changes relevant once `async-changes` branch is merged to `main`? If not, we can let these changes merged in main branch for now, and later these changes would be removed once `async-changes` branch is merged to `main`. Would leave it to you to decide :)
> 
> This is a definite code defect which may affect the delopyment in production, so I think should be fixed ASAP. We will try [async-changes](https://github.com/open-telemetry/opentelemetry-cpp/tree/async-changes) latter in test environment when we are available ;-)

Sure, if you don't mind these changes in this PR may be removed later.
This PR also fixes #1220 .

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -470,7 +477,47 @@ class HttpOperation
     }
 
     /* select() returns the number of signalled sockets or -1 */
-    res = select((int)sockfd + 1, &infd, &outfd, &errfd, &tv);
+    if (select((int)sockfd + 1, &infd, &outfd, &errfd, &tv) > 0)
+    {
+      if (for_recv)
+      {
+        res = FD_ISSET(sockfd, &infd);
```
owent, line 484: Some compiler will warning this(implicit convert from int to bool).We should use `res = 0 != FD_ISSET(sockfd, &infd);` here.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -443,13 +444,19 @@ class HttpOperation
...
+   * @return true if expected events occur, false if timeout or error happen
    */
-  static int WaitOnSocket(curl_socket_t sockfd, int for_recv, long timeout_ms)
+  static bool WaitOnSocket(curl_socket_t sockfd, int for_recv, long timeout_ms)
   {
+    bool res = false;
+
+#if defined(_WIN32)
+
+    if (sockfd > FD_SETSIZE)
```
owent, line 455: Should we retry several times when the `sockfd` is not selected? It usually happens when there is a lot actived fds, and `select` can not get them out once.

owent (changes requested):

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
taozhijiang: Our service is quit busy.
Even though I set the trace sample_ration=0.01, the service startup several minutes latter, the CURLINFO_LASTSOCKET return sock fd above FD_SETSIZE. After this, all the trace message can not be upload and never recovered.
I don't think retry can work.

owent (commented):
LGTM after the format is fixed.

lalitb:
@taozhijiang  - You can fix the format as documented here - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#pull-requests

lalitb (approved):

lalitb:
> LGTM after the format is fixed.

@owent  - Can you remove the requested change if it looks good to merge? Thanks.

owent (approved):

owent:
> > LGTM after the format is fixed.
> 
> @owent - Can you remove the requested change if it looks good to merge. Thanks.

Done

PR #1415, opened 2022-05-23
[BUILD] fix nlohmann_json's (third party) include dir

leodido:
Signed-off-by: Leonardo Di Donato <leodidonato@gmail.com>

Fixes: could not build with nlohmann_json from third_party

## Changes

_Fixes `nlohmann_json` include directory for building it from third party._

### DIFF
--- a/cmake/nlohmann-json.cmake
+++ b/cmake/nlohmann-json.cmake
@@ -26,7 +26,7 @@ ExternalProject_Add(nlohmann_json_download
 )
 
 ExternalProject_Get_Property(nlohmann_json_download INSTALL_DIR)
-SET(NLOHMANN_JSON_INCLUDE_DIR ${INSTALL_DIR}/third_party/src/nlohmann_json_download/single_include)
+SET(NLOHMANN_JSON_INCLUDE_DIR ${INSTALL_DIR}/src/nlohmann_json_download/single_include)
 add_library(nlohmann_json_ INTERFACE)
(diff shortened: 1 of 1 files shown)


cmake/nlohmann-json.cmake
```diff
@@ -26,7 +26,7 @@ ExternalProject_Add(nlohmann_json_download
 )
 
 ExternalProject_Get_Property(nlohmann_json_download INSTALL_DIR)
-SET(NLOHMANN_JSON_INCLUDE_DIR ${INSTALL_DIR}/third_party/src/nlohmann_json_download/single_include)
+SET(NLOHMANN_JSON_INCLUDE_DIR ${INSTALL_DIR}/src/nlohmann_json_download/single_include)
```
ThomsonTan, line 29: Wondering why `third_party` should be removed as it was declared as PREFIX in line 6 which is the download root?

leodido: To my understanding exactly for that reason.

Since PREFIX is specified, all the other directories associated with the external project will be created under it (per CMake docs).

Thus, the install directory (which is used to obtain the include directory) already contains it.

ThomsonTan: I see, thanks for the fix.

ThomsonTan (approved):

owent (approved):
LGTM

PR #1471, opened 2022-06-30
Fix trace sdk builder 1393

### LINKED ISSUE #1393: [Trace SDK] Provide builders to avoid exposing SDK internals
Greetings.

In an application, as a **_user_** of the trace sdk, I need to create a `opentelemetry::exporter::otlp::OtlpGrpcExporter` instance.

Per instructions in `exporters/otlp/README.md`,
the only way to do this seems to call new on the class itself, which imply that:
- the code `#include <opentelemetry/exporters/otlp/otlp_grpc_exporter.h>`
- the user code compiled in the application now depends on the exact memory layout of `OtlpGrpcExporter`.

This is problematic for two reasons:

1. In the specific case of `OtlpGrpcExporter`, the header file `otlp_grpc_exporter.h` contains:
`#include "opentelemetry/proto/collector/trace/v1/trace_service.grpc.pb.h"`

To build my application, I now need to compile this header from protobuf, and I need to install all the dependencies that are generated as well. This is very inconvenient, but still possible. It will slow down adoption of opentelemetry-cpp in any case.

2. In general, the content of `otlp_grpc_exporter.h` is very dependent on the SDK implementation, and is likely to change, causing compatibility issues with the user code. Delivering fixes without breaking the application is now much more difficult, which is the rationale for filing this as a bug and not a feature request.

Now, taking a step back, for applications merely **_using_** the SDK, they do not need to actually know the details of an exporter implementation, they only needs an `opentelemetry::sdk::trace::SpanExporter` to build a TraceProvider.

Applications **_extending_** the SDK will still need to see more details, this is a different topic.

I would suggest to create Builder classes, like:

```
// In namespace opentelemetry::exporter::otlp::
class OtlpGrpcExporterBuilder {
public:
  static std::unique_ptr<opentelemetry::sdk::trace::SpanExporter>
  Build(opentelemetry::exporter::otlp::OtlpGrpcExporterOptions options);
};

```
And implement the builder in a .cc file (in the exporter library), not inline in the header, even for a one-liner.

Note how the builder returns `opentelemetry::sdk::trace::SpanExporter`, not `opentelemetry::exporter::otlp::OtlpHttpExporter`.

This resolves:
- the header file dependencies, as file `otlp_grpc_exporter.h` is no longer needed (options are in a separate header)
- the binary layout dependency on the SDK implementation classes, as class `OtlpHttpExporter` is never seen.

This pattern can be applied to every subclass of SpanExporter, Processor, etc, which are at the top level of the SDK.

This can be applied to Traces, Metrics, and Logs.

Regards.

### LINKED ISSUE #1473: [Examples] Memory ownership issues in InMemorySpanExporter usage example.
In example `multi_processor/main.cc`, the code:

1)

Builds a `InMemorySpanExporter` object

```
  InMemorySpanExporter *memory_span_exporter = new InMemorySpanExporter();
  auto exporter2 = std::unique_ptr<trace_sdk::SpanExporter>(memory_span_exporter);

```
2)

Transfers memory ownership to a `SpanProcessor`

```
  auto processor2 = std::unique_ptr<trace_sdk::SpanProcessor>(
      new trace_sdk::SimpleSpanProcessor(std::move(exporter2)));

```

3)

Use the underlying `InMemorySpanExporter` object directly.

```
  InMemorySpanExporter *memory_span_exporter = initTracer();

  foo_library();
  auto memory_spans = memory_span_exporter->GetData()->GetSpans();
```

This usage of `InMemorySpanExporter` is unsafe, as the exporter pointer is not owned, and the exporter could go away for example when installing a different global tracer provider (if I understand correctly).

Suggested fix:

`initTracer()` should return the underlying `InMemorySpanData` instead, which is safe to do because the data is held by a shared pointer.
Even if the exporter goes away, the data will still be accessible.

marcalff:
(Last edit 2022-07-12)

Fixes #1393
Fixes #1473

## Changes

Implement Trace SDK builders, to avoid exposing SDK details to the application.
This is needed to deploy opentelemetry-cpp as a shared library.

Fixed memory ownership of `InMemorySpanExporter` in examples.

* [X] Changes in public API reviewed (no API change)
* [X] Added design documentation


marcalff:
This PR is now ready for review.

Remaining tasks:
- run again all examples locally
- add doxygen comments
- write Changelog

Reviewers:
- Please start by reading docs/cpp-sdk-factory-design.md
- Please check the Bazel BUILD files in particular (I never used Bazel)

Regards,
-- Marc

lalitb (approved):
Very nicely done. Thanks for the PR. Looks good with few (very) minor comments. Hope it would be fine to merge @owent 's async changes first :)

docs/cpp-sdk-factory-design.md
```diff
@@ -0,0 +1,161 @@
...
+
+The use cases for applications calling the opentelemetry SDK can be
+classified into the following two broad categories:
+
+- SDK users, the application needs to assemble a trace provider / metrics
+  provider / log provider by assembling available SDK elements,
+- SDK implementors, the application needs to extend the opentelemetry SDK
+  by implementing new features.
+
+## SDK users
```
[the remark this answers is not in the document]
marcalff: Thanks @lalitb, and agree.

Revised the wording in general.

examples/batch/main.cc
```diff
@@ -1,28 +1,36 @@
...
+
+#include "opentelemetry/sdk/trace/batch_span_processor_factory.h"
+#include "opentelemetry/sdk/trace/tracer_provider_factory.h"
+
+/* Exporter */
+
 // Using an exporter that simply dumps span data to stdout.
-#include "opentelemetry/exporters/ostream/span_exporter.h"
-#include "opentelemetry/sdk/trace/batch_span_processor.h"
+#include "opentelemetry/exporters/ostream/span_exporter_factory.h"
```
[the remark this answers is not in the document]
marcalff: Fixed, cleaned up includes that got affected by this patch.

I agree, we need to clean up in includes in general by running include-what-you-use,
for the entire repository (in a separate PR).

exporters/memory/CMakeLists.txt
```diff
@@ -1,15 +1,19 @@
-add_library(opentelemetry_exporter_in_memory INTERFACE)
+add_library(opentelemetry_exporter_memory_trace
```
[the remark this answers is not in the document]
marcalff: Before the change, "opentelemetry_exporter_in_memory" was in fact a header only SDK,
no actual library was built and installed, so I don't think there are risks of breaking anything.

With the change, the memory exporter factory needs to be in a *.cc file,
and so we need a (real) library for it.

I used the name opentelemetry_exporter_memory_trace,
to allow for metrics memory exporter libraries later.

marcalff:
> Very nicely done. Thanks for the PR. Looks good with few (very) minor comments. Hope it would be fine to merge @owent 's async changes first :)

Thanks.

The async changes have priority, I will deal with any merge conflict, these should be trivial to resolve anyway.

One remaining point to discuss, that I am not sure of, is whether Factory classes should use:
- std::shared_ptr
- nostd::shared_ptr
because this is part of the application owner / SDK interface.

@lalitb , @owent , @ThomsonTan , @esigo any input ?

owent:
> > Very nicely done. Thanks for the PR. Looks good with few (very) minor comments. Hope it would be fine to merge @owent 's async changes first :)
> 
> Thanks.
> 
> The async changes have priority, I will deal with any merge conflict, these should be trivial to resolve anyway.
> 
> One remaining point to discuss, that I am not sure of, is whether Factory classes should use:
> 
> * std::shared_ptr
> * nostd::shared_ptr
>   because this is part of the application owner / SDK interface.
> 
> @lalitb , @owent , @ThomsonTan , @esigo any input ?

Or `std::unique_ptr` so users can convert it to any implementation of `shared_ptr`?

marcalff:
> Or `std::unique_ptr` so users can convert it to any implementation of `shared_ptr`?

Thanks, good point: a factory method should return a unique_ptr anyway.
Fixed.

lalitb:
@owent Can you please review/approve it once you have time? Thanks :)

owent:
> @owent Can you please review/approve it once you have time? Thanks :)

Sorry for something else in company take my bandwidth. I will review this it in 2 or 3 days.

lalitb:
> > @owent Can you please review/approve it once you have time? Thanks :)
> 
> Sorry for something else in company take my bandwidth. I will review this it in 2 or 3 days.

Thanks. That should be fine :)

owent:
@marcalff Cloud you add a standalone unit test which only include factory headers and do not depend headers of protobuf? I think it's helpful to keep factory heads clean in the future.

exporters/otlp/src/otlp_grpc_exporter_factory.cc
```diff
@@ -0,0 +1,32 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace otlp
+{
+
+std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> OtlpGrpcExporterFactory::Create()
+{
+  OtlpGrpcExporterOptions options;
+  return Create(options);
```
owent, line 20: Should we use `Create(OtlpGrpcExporterOptions{})` to reduce copy here?

exporters/otlp/src/otlp_http_exporter_factory.cc
```diff
@@ -0,0 +1,29 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace otlp
+{
+
+std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> OtlpHttpExporterFactory::Create()
+{
+  OtlpHttpExporterOptions options;
+  return Create(options);
```
owent, line 17: Should we use `Create(OtlpHttpExporterOptions{})` to reduce copy here?

exporters/zipkin/src/zipkin_exporter_factory.cc
```diff
@@ -0,0 +1,31 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace zipkin
+{
+
+std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> ZipkinExporterFactory::Create()
+{
+  ZipkinExporterOptions options;
+  return Create(options);
```
owent, line 19: Just like OTLP exporters, should we use `Create(ZipkinExporterOptions{})` to reduce copy here?

sdk/src/trace/tracer_provider_factory.cc
```diff
@@ -0,0 +1,97 @@
...
+  std::unique_ptr<trace_api::TracerProvider> provider(new trace_sdk::TracerProvider(
+      std::move(processor), resource, std::move(sampler), std::move(id_generator)));
+  return provider;
+}
+
+std::unique_ptr<opentelemetry::trace::TracerProvider> TracerProviderFactory::Create(
+    std::vector<std::unique_ptr<SpanProcessor>> &&processors)
+{
+  auto resource = opentelemetry::sdk::resource::Resource::Create({});
+  return Create(std::move(processors), resource);
```
owent, line 57: Should we use `Create(std::move(processors), opentelemetry::sdk::resource::Resource::Create({}));` or `Create(std::move(processors), std::move(resource));` to reduce copy?

sdk/src/trace/tracer_provider_factory.cc
```diff
@@ -0,0 +1,97 @@
...
+namespace sdk
+{
+namespace trace
+{
+
+std::unique_ptr<trace_api::TracerProvider> TracerProviderFactory::Create(
+    std::unique_ptr<SpanProcessor> processor)
+{
+  auto resource = opentelemetry::sdk::resource::Resource::Create({});
+  return Create(std::move(processor), resource);
```
owent, line 22: Should we use `Create(std::move(processors), opentelemetry::sdk::resource::Resource::Create({}));` or `Create(std::move(processors), std::move(resource));` to reduce copy?

sdk/src/trace/tracer_provider_factory.cc
```diff
@@ -0,0 +1,97 @@
...
+}
+
+std::unique_ptr<trace_api::TracerProvider> TracerProviderFactory::Create(
+    std::unique_ptr<SpanProcessor> processor,
+    opentelemetry::sdk::resource::Resource resource)
+{
+  auto sampler = AlwaysOnSamplerFactory::Create();
+  return Create(std::move(processor), resource, std::move(sampler));
```
owent, line 30: Should we use `Create(std::move(processors), std::move(resource), std::move(sampler));` to reduce copy?

owent (commented):
Could you please add a standalone unit test for OTLP exporters and ES exporters which only include factory headers and do not depend headers of protobuf? I think it's helpful to keep factory heads clean in the future.

exporters/otlp/src/otlp_grpc_exporter_factory.cc
marcalff: Thanks for the comments.
Here the Create() method with an option takes a `const OtlpGrpcExporterOptions &options` parameter,
so there should be no copy.

exporters/otlp/src/otlp_http_exporter_factory.cc
marcalff: Likewise, the Create() method with an option takes a `const &` parameter,
so there should be no copy.

exporters/zipkin/src/zipkin_exporter_factory.cc
marcalff: Likewise, the Create() method with an option takes a `const &` parameter,
so there should be no copy.

sdk/src/trace/tracer_provider_factory.cc
marcalff: Here there is a copy, thanks for pointing this out.

Fixed by using `const opentelemetry::sdk::resource::Resource &resource` in every` Create()` method.

sdk/src/trace/tracer_provider_factory.cc
marcalff: Likewise,

Fixed by using `const opentelemetry::sdk::resource::Resource &resource` in every` Create()` method.

sdk/src/trace/tracer_provider_factory.cc
marcalff: Likewise,

Fixed by using `const opentelemetry::sdk::resource::Resource &resource` in every` Create()` method.

exporters/otlp/test/otlp_grpc_exporter_factory_test.cc
```diff
@@ -0,0 +1,36 @@
...
+  are included by OtlpGrpcExporterFactory,
+  even indirectly, the build will fail.
+*/
+#define PROTOBUF_VERSION 6666666
+
+#include <gtest/gtest.h>
+
+#include "opentelemetry/exporters/otlp/otlp_grpc_exporter_factory.h"
+#include "opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h"
+
```
owent, line 16: Maybe we can add a `static_assert` to ensure we will not break the header dependency in the future?

```cpp
#ifdef GOOGLE_PROTOBUF_VERSION
static_assert(false, "protobuf should not be included here");
#endif
```

owent, line 16: Maybe we can also add a `static_assert` to ensure we will not break the header dependency in the future?

```cpp
#ifdef GOOGLE_PROTOBUF_VERSION
#  error  "protobuf should not be included"
#endif
```

exporters/otlp/test/otlp_http_exporter_factory_test.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "opentelemetry/exporters/otlp/otlp_http_exporter_options.h"
+
+/*
+  Make sure OtlpHttpExporterFactory does not leak nlohmann/json.hpp,
+  even indirectly.
+*/
+#ifdef NLOHMANN_JSON_VERSION_MAJOR
+#  error "OtlpHttpExporterFactory should not expose nlohmann/json.hpp"
+#endif /* NLOHMANN_JSON_VERSION_MAJOR */
+
```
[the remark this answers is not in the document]
marcalff: Fixed as indicated.

The build error message is actually better, thanks.

Not sure what you meant by 'static_assert', so please clarify if something else is needed.

exporters/otlp/test/otlp_grpc_exporter_factory_test.cc
marcalff: Thanks, fixed.

owent (approved):
LGTM

marcalff:
@lalitb, @ThomsonTan  Do not merge yet.

It turns out the library name for the in memory exporter is visible externally after all,
in `cmake/opentelemetry-cpp-config.cmake.in`

`  opentelemetry-cpp::in_memory_span_exporter        - Imported target of opentelemetry-cpp::in_memory_span_exporter`

Reverting un needed changes in `exporter/memory/CMakeLists.txt`, will notify when ready to merge.

exporters/memory/CMakeLists.txt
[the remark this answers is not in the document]
marcalff: Reconsidered, and reverted the library name change.

The name is visible externally in `cmake/opentelemetry-cpp-config.cmake.in`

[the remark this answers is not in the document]
owent: `cmake/opentelemetry-cpp-config.cmake.in` only use the `EXPORT_NAME` attribute, do not use internal target name.
Internal target name change will only break the projects use opentelemery-cpp as a submodule.

docs/cpp-sdk-factory-design.md
```diff
@@ -0,0 +1,166 @@
...
+auto exporter = std::unique_ptr<opentelemetry::sdk::trace::SpanExporter>(
+  new opentelemetry::exporter::otlp::OtlpGrpcExporter(opts));
+```
+
+When the application code is built and linked against static
+opentelemetry-cpp libraries, the result is a binary that is consistent, and
+works.
+
+However, in case of a bug fixed in opentelemetry-cpp itself, the entire
+application must be rebuild (with a fixed version), redistributed,
```
ThomsonTan, line 36: nit
```suggestion
application must be rebuilt (with a fixed version), redistributed,
```

marcalff: Thanks, fixed.

PR #1474, opened 2022-07-01
Fix otlp grpc exporter example in the documentation

lostwong:
The value for `opts.endpoint` for otlp grpc exporter had an extra colon that would cause an exception if used as is.

Fixes # (issue)

## Changes

Minor fix to an example in the opentelemetry-cpp documentation.


lalitb (approved):
Thanks for the fix. This was me with the typo :man_facepalming:

ThomsonTan (approved):

PR #1413, opened 2022-05-20
Merge all commits of `async-changes` into `main`

owent:
Signed-off-by: owent <admin@owent.net>

Fixes #1412 #1243 #1176 #1220

This PR is created by these process below.

```bash
git checkout -b merge_main_into_async-changes open-telemetry/async-changes
git reset --hard open-telemetry/async-changes
git merge open-telemetry/main
# resolve conflicts and then
git commit -m "Merge main into async-changes"
git checkout -b merge_async-changes_into_main open-telemetry/main
git reset --hard open-telemetry/main
git merge --squash merge_main_into_async-changes
git commit -m "Merge all commits of async-changes"
```

Please review these changes, and I will rerun above progress and generate a new commit after some more commints are merged into `main` later. 
@open-telemetry/cpp-approvers 

## Changes

* [SDK] Async Batch Span/Log processor with max async support ([#1306](https://github.com/open-telemetry/opentelemetry-cpp/pull/1306))
* [EXPORTER] OTLP http exporter allow concurrency session ([#1209](https://github.com/open-telemetry/opentelemetry-cpp/pull/1209))
* [EXT] `curl::HttpClient` use `curl_multi_handle` instead of creating a thread
  for every request and it's able to reuse connections now. ([#1317](https://github.com/open-telemetry/opentelemetry-cpp/pull/1317))

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [x] Changes in public API reviewed

PR #1477, opened 2022-07-05
Increase the default size of FD_SETSIZE

### LINKED ISSUE #1476: The default value of FD_SETSIZE is too small for real scenario
**Describe your environment**
Windows 11 10.0.22000.739

**Steps to reproduce**
Run curl_http_test.exe

The test should pass

The test failed

**Additional context**
The test is failed on that the allocated socket handle value is larger than `FD_SETSIZE` which is 64. The value seems too small for real scenario.

ThomsonTan:
Fixes #1476.

## Changes

### DIFF
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -15,6 +15,7 @@
 #include <vector>
 #ifdef _WIN32
 #  include <io.h>
+#  define FD_SETSIZE 1024
 #  include <winsock2.h>
 #else
 #  include <poll.h>


lalitb:
Just wondering if the issue would be still valid once @owent async changes are merged, as that PR removes `select` call usage. Also, there was similar fix done for Linux - #1410 to replace `select` call with `poll`. If there is similar alternative for Windows ?

ThomsonTan:
> Just wondering if the issue would be still valid once @owent async changes are merged, as that PR removes `select` call usage. Also, there was similar fix done for Linux - #1410 to replace `select` call with `poll`. If there is similar alternative for Windows ?

@lalitb thanks for the reminder. Yes, the async change will remove current code with issue, so I think I could close this PR once the async change is merged. Also I think #1410 applies to Windows as well, right?

lalitb:
> Also I think https://github.com/open-telemetry/opentelemetry-cpp/pull/1410 applies to Windows as well, right?

#1410 replaces `select` with `poll` only for POSIX and keeps `select` for Windows. Merging this PR will add more conflict with async PR. If this is not the blocker, we can let @owent to decide how to procced with this PR?

ThomsonTan:
Ok. I'll just leave this PR open and will close it once the async change from @owent gets merged, no need for merging.

owent:
libcurl use `select` on WIndows and `poll` on Linux, just like #1410 . According to https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select, nfds of `select` on Windows is just ignored. I think we can wait for #1413 to be merged.

Issue #1405, opened 2022-05-18
[Metrics SDK] Add InMemory metric exporter


lalitb:
As required by the specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/in-memory.md

This would also be great tool for testing.

marcalff:
Also, please consider the visitor idea, detailed in:

open-telemetry/opentelemetry-cpp#1698

Issue #1428, opened 2022-06-03
Libraries delivered: Reconcile Bazel and CMake build


marcalff:
This is a remaining item identified during review of PR #1420

(1)

Document:
- the name of each library delivered
- the expected dependency graph (DAG)

This documentation is meant to be used by:
- users of `opentelemetry-cpp`,
- maintainers of `opentelemetry-cpp`.

A good place could be `docs/library-distribution.md`

(2)

Fix the `Bazel` build to comply with (1)

(3)

Fix the `CMake` build to comply with (1)

Libraries build with Bazel or CMake should be the same.

yxue:
Do we consider use the tools like https://github.com/google/bazel-to-cmake? The CMakeLists.txt can be generated from the BUILD and WORKSPACE as mentioned.

owent:
I think it's difficult to use [bazel-to-cmake][1] like tools. CMake support environment detection better and has a better compatibility then bazel.(e.g. We support gcc 4.8 and use a old version of gRPC and abseil-cpp with cmake, but we don't suport legacy compilers with bazel now).
Maybe we can consider If there is a tool than can convert cmake script to bazel script, or just write a script to check if is there any target name different between bazel and cmake?

[1]: https://github.com/google/bazel-to-cmake

dbarker:
The CMake libraries/targets delivered are now documented in the INSTALL markdown file by #3220.

PR #1481, opened 2022-07-09
Add the in-memory metrics exporter.

yxue:
Fixes # (issue) [1405](https://github.com/open-telemetry/opentelemetry-cpp/issues/1405)

## Changes

### DIFF
--- a/exporters/memory/include/opentelemetry/exporters/memory/in_memory_metric_exporter.h
+++ b/exporters/memory/include/opentelemetry/exporters/memory/in_memory_metric_exporter.h
@@ -1,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+#ifndef ENABLE_METRICS_PREVIEW
+
+#  include <memory>
+#  include <mutex>
+#  include "opentelemetry/common/spin_lock_mutex.h"
+#  include "opentelemetry/exporters/memory/in_memory_metric_data.h"
+#  include "opentelemetry/sdk/common/circular_buffer.h"
+#  include "opentelemetry/sdk/metrics/data/metric_data.h"
+#  include "opentelemetry/sdk/metrics/instruments.h"
+#  include "opentelemetry/sdk/metrics/metric_exporter.h"
+#  include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace metrics
+{
+const size_t MAX_BUFFER_SIZE = 100;
+
+/**
+ * The in-memory metrics exporter exports metrics data to the memory.
+ *
+ */
+class InMemoryMetricExporter final : public opentelemetry::sdk::metrics::MetricExporter
+{
+public:
+  /**
+   * Create an InMemoryMetricExporter. The constructor takes the buffer size as the input and
+   * initializes the ring buffer with the given size.
+   *
+   */
+  explicit InMemoryMetricExporter(size_t buffer_size = MAX_BUFFER_SIZE)
+      : data_(new InMemoryMetricData(buffer_size))
+  {}
+
+  sdk::common::ExportResult Export(const sdk::metrics::ResourceMetrics &data) noexcept override;
+
+  bool ForceFlush(
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept override;
+
+  bool Shutdown(
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept override;
+
+  /**
+   * @return Returns a shared pointer to this exporters InMemoryMetricData.
+   */
+  inline InMemoryMetricData &GetData() noexcept { return *data_; }
+
+private:
+  bool isShutdown() const noexcept;
+  bool is_shutdown_ = false;
+  mutable opentelemetry::common::SpinLockMutex lock_;
+
+  std::unique_ptr<InMemoryMetricData> data_ = nullptr;
+};
+}  // namespace metrics
+}  // namespace exporter
+OPENTELEMETRY_END_NAMESPACE
+#endif
--- a/exporters/memory/src/in_memory_metric_exporter.cc
+++ b/exporters/memory/src/in_memory_metric_exporter.cc
@@ -1,0 +1,54 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef ENABLE_METRICS_PREVIEW
+
+#  include "opentelemetry/exporters/memory/in_memory_metric_exporter.h"
+#  include "opentelemetry/exporters/otlp/otlp_metric_utils.h"
+#  include "opentelemetry/sdk_config.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace metrics
+{
+
+sdk::common::ExportResult InMemoryMetricExporter::Export(
+    const sdk::metrics::ResourceMetrics &data) noexcept
+{
+  if (isShutdown())
+  {
+    OTEL_INTERNAL_LOG_ERROR("[InMemory Metric] Exporting "
+                            << data.instrumentation_info_metric_data_.size()
+                            << " records(s) failed, exporter is shutdown");
+    return sdk::common::ExportResult::kFailure;
+  }
+
+  std::unique_ptr<proto::metrics::v1::ResourceMetrics> resource_metrics(
+      new proto::metrics::v1::ResourceMetrics);
+  otlp::OtlpMetricUtils::PopulateResourceMetrics(data, resource_metrics.get());
+  data_->Add(std::move(resource_metrics));
+  return sdk::common::ExportResult::kSuccess;
+}
+
+bool InMemoryMetricExporter::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  return true;
+}
+
+bool InMemoryMetricExporter::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
+  is_shutdown_ = true;
+  return true;
+}
+
+bool InMemoryMetricExporter::isShutdown() const noexcept
+{
+  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
+  return is_shutdown_;
+}
+}  // namespace metrics
+}  // namespace exporter
+OPENTELEMETRY_END_NAMESPACE
+#endif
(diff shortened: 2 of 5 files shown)


owent:
Could you please alse add cmake targets in `exporters/memory/CMakeLists.txt` and `cmake/opentelemetry-cpp-config.cmake.in` .

yxue:
> Could you please alse add cmake targets in `exporters/memory/CMakeLists.txt` and `cmake/opentelemetry-cpp-config.cmake.in` .

Sure, I can add that. I feel supporting/maintaining two build systems is a little bit inconvenient. Do we consider using some tools like, https://github.com/google/bazel-to-cmake, so that CMakeLists.txt can be generated from the BUILD and WORKSPACE, or using cmake and bazel (https://github.com/bazelbuild/rules_foreign_cc) supports importing cmake projects.

owent:
> > Could you please alse add cmake targets in `exporters/memory/CMakeLists.txt` and `cmake/opentelemetry-cpp-config.cmake.in` .
> 
> Sure, I can add that. I feel supporting/maintaining two build systems is a little bit inconvenient. Do we consider using some tools like, https://github.com/google/bazel-to-cmake, so that CMakeLists.txt can be generated from the BUILD and WORKSPACE, or using cmake and bazel (https://github.com/bazelbuild/rules_foreign_cc) supports importing cmake projects. 

Good point, I think it require more discussion. In most case, I think cmake users do not want to install another bazel toolset, and bazel users do not want install cmake toolset either.
cmake has many modules and tools to work with legacy build toolchains(pkg-config, makefile, autoconf, android ndk, feature dection by test running or compiling some c/c++ codes, environments and etc.).I'm not family with bazel and I'm not sure if it's easy to let bazel and cmake work together with these modules.

BTW: We had a discussion before and try to find a way to use bazel to build both with legacy gcc 4.8 and the latest compiler here https://github.com/open-telemetry/opentelemetry-cpp/pull/666 , do you has any suggestion about it?

lalitb:
>  I feel supporting/maintaining two build systems is a little bit inconvenient

Yes, I do agree it's inconvenient to maintain both these build systems. The decision to support both was made much before any of the existing contributors joined the team. The SIG work was initiated by the folks from Google and Microsoft, and it was defacto choice to support the build systems used by these companies :). Having said that, most of the existing C++ projects use either(or both) of these build systems, and so it makes sense to keep supporting both the build systems.

ThomsonTan:
The 2 build systems issue were discussed in the past as @lalitb and @owent mentioned. As there are users for both of them, we will try our best to keep both up to date and consistent. For the https://github.com/google/bazel-to-cmake, it has not been updated for a few years, not sure it could work, together with that we've already have full set of cmake files which is not expected to be overwriten.

yxue:
> 



> Thanks for the PR @yxue. This is your first PR here, and it is nicely done :)
> 
> Just thinking whether the InMemoryExporter should have any dependency on the protobuf. For Logs and Traces, the memory/ostream exporters are not dependent on any external library. Is it possible to achieve this for Metrics too - Eg by letting CircularBuffer contain the list of [MetricData ](https://github.com/open-telemetry/opentelemetry-cpp/blob/8691a2d582c2e0c811fbf383b572e1ff882b5b37/sdk/include/opentelemetry/sdk/metrics/data/metric_data.h#L30) or [ResourceMetrics ](https://github.com/open-telemetry/opentelemetry-cpp/blob/8691a2d582c2e0c811fbf383b572e1ff882b5b37/sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h#L27) ?
> 
> We can discuss further if it is something not feasible/or difficult to achieve with the current SDK implementation.

Thanks @lalitb! I was thinking about using the CircularBuffer with the  ResourceMetrics but then we need the special process for the pointers inside the structure.  

Having the MetricData with CircularBuffer might work but some info is lost. 

Another thing is about comparing these structure objects. With protobuf message, the objects can be easily compared. Since we don't define the == for MetricsData and ResourceMetrics, comparing them in the unit test is not trivial.

yxue:
Thanks @lalitb @owent and @ThomsonTan for the info about the building system. 

I am not sure about fixing the gcc 4.8 and grpc compatibility. I do see the building systems are right now used by multiple teams/companies. I also saw there are already some divergence problems between these two systems, e.g., https://github.com/open-telemetry/opentelemetry-cpp/issues/1428. I think we might need some tools to reconcile that if the community is going to keep two building systems.

lalitb:
> Thanks @lalitb! I was thinking about using the CircularBuffer with the ResourceMetrics but then we need the special process for the pointers inside the structure

Yes, I thought so too. Should be fine to keep the dependency unless someone has any concern :)
btw, I didn't realize this PR also has changes for bazel build for metrics proto files and raised a PR for that in #1489. Sorry about that, as you would have to resolve the conflict here :( Would ensure not to merge any further conflicting PRs till this is merged.

yxue:
> > Thanks @lalitb! I was thinking about using the CircularBuffer with the ResourceMetrics but then we need the special process for the pointers inside the structure
> 
> Yes, I thought so too. Should be fine to keep the dependency unless someone has any concern :) btw, I didn't realize this PR also has changes for bazel build for metrics proto files and raised a PR for that in #1489. Sorry about that, as you would have to resolve the conflict here :( Would ensure not to merge any further conflicting PRs till this is merged.

Sure, no worries. The PR was updated. PTAL, thanks!

exporters/memory/test/in_memory_metric_exporter_test.cc
```diff
@@ -0,0 +1,298 @@
...
+// SPDX-License-Identifier: Apache-2.0
+#ifndef ENABLE_METRICS_PREVIEW
+
+#  include "opentelemetry/exporters/memory/in_memory_metric_exporter.h"
+#  include "google/protobuf/text_format.h"
+#  include "opentelemetry/sdk/metrics/data/metric_data.h"
+
+#  include <gmock/gmock.h>
+#  include <gtest/gtest.h>
+#  include <iostream>
```
lalitb, line 11: do we need iostream?

lalitb (approved):
LGTM.

exporters/memory/src/in_memory_metric_exporter.cc
```diff
@@ -0,0 +1,55 @@
...
+  std::unique_ptr<proto::metrics::v1::ResourceMetrics> resource_metrics(
+      new proto::metrics::v1::ResourceMetrics);
+  otlp::OtlpMetricsUtils::PopulateResourceMetrics(data, resource_metrics.get());
+  data_->Add(std::move(resource_metrics));
+  return sdk::common::ExportResult::kSuccess;
+}
+
+bool InMemoryMetricExporter::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
```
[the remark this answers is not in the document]
yxue: done

exporters/memory/include/opentelemetry/exporters/memory/in_memory_data.h
```diff
@@ -0,0 +1,62 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace memory
+{
+/**
+ * A wrapper class holding in memory exporter data
+ */
+template <typename T>
```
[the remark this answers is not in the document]
yxue: sure, https://github.com/open-telemetry/opentelemetry-cpp/issues/1493 created

owent:
Could you please revert changes in `/exporters/otlp` ?It's duplicated with #1487 , I belive this PR is just for memory metrics exporter.

yxue:
> Could you please revert changes in `/exporters/otlp` ?It's duplicated with #1487 , I belive this PR is just for memory metrics exporter.

The change under exporters/otlp is for fixing the missing part of conversion between protobuf and structure. Remove that will fail the unit test.

owent:
> > Could you please revert changes in `/exporters/otlp` ?It's duplicated with #1487 , I belive this PR is just for memory metrics exporter.
> 
> The change under exporters/otlp is for fixing the missing part of conversion between protobuf and structure. Remove that will fail the unit test.

OK. we can solve this conflict later.

owent (changes requested):
LGTM after `in_memory_metric_exporter` is also added into `cmake/opentelemetry-cpp-config.cmake.in` ?

lalitb:
> LGTM after `in_memory_metric_exporter` is also added into `cmake/opentelemetry-cpp-config.cmake.in`.

Good point. @yxue this file contains the rules to enable opentelemetry-cpp included in other projects cmake files using `find_package(opentelemetry-cpp)`. All the targets/libraries created by opentelemetry-cpp should be included in this file.

yxue:
> > LGTM after `in_memory_metric_exporter` is also added into `cmake/opentelemetry-cpp-config.cmake.in`.
> 
> Good point. @yxue this file contains the rules to enable opentelemetry-cpp included in other projects cmake files using `find_package(opentelemetry-cpp)`. All the targets/libraries created by opentelemetry-cpp should be included in this file.

Updated, thanks!

owent (approved):

exporters/memory/CMakeLists.txt
```diff
@@ -21,19 +21,56 @@ install(
...
+  add_library(
+    opentelemetry_exporter_in_memory_metric
+    src/in_memory_metric_exporter.cc ../otlp/src/otlp_metric_utils.cc
+    ../otlp/src/otlp_populate_attribute_utils.cc)
+
+  target_include_directories(
+    opentelemetry_exporter_in_memory_metric
+    PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+           "$<INSTALL_INTERFACE:include>"
+           "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/../otlp/include>")
```
owent, line 41: `target_link_libraries(opentelemetry_exporter_in_memory_metric PUBLIC opentelemetry_otlp_recordable)` would be a better alternative here.But when we building without `WITH_OTLP`, `opentelemetry_otlp_recordable` will be unavailable.
Maybe this module should also be disabled when we build without `WITH_OTLP` now?

exporters/memory/include/opentelemetry/exporters/memory/in_memory_metric_exporter.h
```diff
@@ -0,0 +1,67 @@
...
+  bool ForceFlush(
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept override;
+
+  bool Shutdown(
+      std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept override;
+
+  /**
+   * @return Returns a shared pointer to this exporters InMemoryMetricData.
+   */
+  inline InMemoryMetricData &GetData() noexcept { return *data_; }
```
marcalff, line 55: The comment is outdated, this does not return a shared pointer.

marcalff (commented):
LGTM

ThomsonTan:
@yxue could you please resolve the merge conflicts?

marcalff (changes requested):
As discussed in C++ SIG meeting,
the in memory storage structure should not depend on protobuf:

```
  using InMemoryMetricData =
      opentelemetry::exporter::memory::InMemoryData<proto::metrics::v1::ResourceMetrics>;

```
Some refactoring needed to address this part.

Issue #1493, opened 2022-07-13
Replace the in-memory span data with the templated in-memory data class


yxue:
Replace the in-memory span data with templated in-memory data class introduced in this PR https://github.com/open-telemetry/opentelemetry-cpp/pull/1481

cc @lalitb 

/assign @yxue

PR #1496, opened 2022-07-14
Use template class for in-memory data.

yxue:
Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/1493

## Changes


exporters/memory/include/opentelemetry/exporters/memory/in_memory_span_exporter.h
```diff
@@ -19,14 +21,16 @@ const size_t MAX_BUFFER_SIZE = 100;
  * A in memory exporter that switches a flag once a valid recordable was received
  * and keeps track of all received spans in memory.
  */
+using InMemorySpanData = exporter::memory::InMemoryData<sdk::trace::SpanData>;
```
lalitb, line 24: should we instead have `InMemorySpanData` as a class to inherit from `InMemoryData`? And also have `InMemorySpanData::GetSpans()` method. This way applications using span memory exporter won't need to change.

owent: Agree

yxue: Thanks! Updated the PR to inherit from the InMemoryData. PTAL.

lalitb (approved):
LGTM

exporters/memory/include/opentelemetry/exporters/memory/in_memory_span_exporter_factory.h
```diff
@@ -3,7 +3,7 @@
 
 #pragma once
 
-#include "opentelemetry/exporters/memory/in_memory_span_data.h"
+#include "opentelemetry/exporters/memory/in_memory_span_exporter.h"
```
marcalff, line 6: Why was this include changed ?

The whole point of the factory class is to hide the implementation,
so there should be no dependencies on the in_memory_span_exporter.h header.

marcalff (changes requested):

exporters/memory/include/opentelemetry/exporters/memory/in_memory_span_exporter_factory.h
lalitb: Good catch. I think `InMemorySpanData` class definition should be moved to `in_memory_span_data.h`.

yxue: Thanks! Comments resolved.

owent (approved):
LGTM

marcalff (approved):
LGTM, thanks for the fixes.

PR #1497, opened 2022-07-14
Add automatically semantic conventions from the spec (#873)

### LINKED ISSUE #873: Semantic Conventions - script to automatic generation of semantic conventions c++ code
As part of #868 and #872 , currently we are embedding semantic convention attributes manually into C++ code. This can be automated. 
This ticket is to check the feasibility of automatic generation of C++ code from semantic convention YAML data model using Jinja template engine.

Reference:
YAML data model for semantic conventions: https://github.com/open-telemetry/opentelemetry-specification/tree/main/semantic_conventions
Semantic Conventions generator for markdown/code: https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator

marcalff:
Fixes #873 

## Changes

Generate automatically semantic conventions from the opentelemetry specs.

* [X] Added documentation


owent (approved):
LGTM

lalitb (approved):
LGTM. Thanks for picking this, it was long pending item :)

It would be good to document this script execution somewhere. One of the options could be in [RELEASING.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/RELEASING.md) if we want to recommend generating semantic conventions before every release. Or anywhere is fine as long as it is documented.

marcalff:
Do not merge yet.

Pushing temporary debug code to investigate windows build failure.

diff --git a/examples/http/server.cc b/examples/http/server.cc
index cf7e6bb..4687d1c 100644
--- a/examples/http/server.cc
+++ b/examples/http/server.cc
@@ -3,6 +3,13 @@
 
 ```
#include "server.h"
 #include "opentelemetry/trace/context.h"
+
+// DEBUG: investigating build break
+
+#ifdef DELETE
+#error "Someone broke semantic_conventions"
+#endif
+
 #include "opentelemetry/trace/semantic_conventions.h"
 #include "tracer_common.h"

```

**Comment by marcalff** [comment]
> LGTM. Thanks for picking this, it was long pending item :)
> 
> It would be good to document this script execution somewhere. One of the options could be in [RELEASING.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/RELEASING.md) if we want to recommend generating semantic conventions before every release. Or anywhere is fine as long as it is documented.

Thanks @lalitb 

Added documentation.

There is a very weird build break on windows,
I suspect some define pollution.

Trying to investigate that with CI, and might need some help (I don't have a windows os).

**Comment by marcalff** [comment]
> Do not merge yet.
> 

Resolved and ok to merge.

On windows, some header file causes to define the symbol `DELETE`,
which conflicts with semantic convention _FaasDocumentOperationValues::DELETE_.

Implemented a work around, and CI on windows now passes.

File: examples/http/client.cc [hunk 1]
```diff
@@ -3,9 +3,15 @@
 
 #include "opentelemetry/ext/http/client/http_client_factory.h"
 #include "opentelemetry/ext/http/common/url_parser.h"
-#include "opentelemetry/trace/experimental_semantic_conventions.h"
 #include "tracer_common.h"
 
+#ifdef DELETE
+// Causes a build error with FaasDocumentOperationValues::DELETE
+#  undef DELETE
+#endif
```
lalitb, line 11: Can we include this ifdef/undef in template file, so we won't have to include it other places?

marcalff: Hi @lalitb 

Thanks for the review.

I considered this, but I am reluctant to add workarounds in the generated code for all users and all platforms to see, while the issue root cause is only for one platform (windows) and one header file (still to be identified) that pollutes the build.

If we add such code, it will never be removed.

Instead, I hope we can identify where this define DELETE is coming from, and put the work around closer to the root cause, removing this from examples.

No code in opentelemetry-cpp defines DELETE, so this most likely comes from a third party or from the system headers.

Having XXXFactory::Create() methods (#1393) will, as a side effect, remove a lot of header dependencies, possibly removing the root cause as well, so this work around should be temporary.

Not having windows locally, I can't tell which header is causing this, so anyone jumping in to investigate will be most welcome.

marcalff: Keeping this item unresolved, so we do not forget about it.

@lalitb Please merge #1393 to the repo, so this PR and the need for the windows work around can be re-evaluated after 1393.

ThomsonTan: FYI, `DELETE` is defined as macro with mask value in `winnt.h`.

```c++
#define DELETE                           (0x00010000L)
```

marcalff: @ThomsonTan Thanks for finding the offending header.

@lalitb Updated the PR today to:

- implement a work around only for windows (_WIN32)
- add a compilation message when the code uses undef DELETE, to avoid surprises

Given how many include chain can land in winnt.h, and given how this file can be included indirectly by generated code with gRPC, having the work around in 4 places in examples is the best we can do, and the most robust for now.

Being in examples, this work around also will have value for users, to show how to resolve the build break.

So, proposing to merge the PR as it stands now, with the following code:

```
#ifdef _WIN32
#  ifdef DELETE
// winnt.h defines DELETE
// Causes a build error with FaasDocumentOperationValues::DELETE
#    pragma message(__FILE__ ": removing define on DELETE")
#    undef DELETE
#  endif
#endif

```

Ok to merge.

Please squash all the commit messages when merging.

marcalff:
Revisited the windows work around for DELETE.

Now that the root cause is well identified, and affects one platform only,
implementing the work around -- for that platform -- in the template file is (more) acceptable:

- this is the best solution, for ease of use. Many users are likely to experience build breaks without the work around because `wintnt.h` is hard to avoid for people using this platform, so we might as well provide it for everyone using windows in the generated file semantic_conventions.h
- the examples are now clean of any work around

Also, the generate script now discloses which semantic convention is compiled, with:

- `-Dsemconv=trace`
- `-Dsemconv=resource`

This will make implementing work around for other issues, should that be necessary, easier.

@ThomsonTan Thanks for finding the offending header.

@lalitb Last item about the DELETE work around now resolved (in a better way, in the template)

Ready to merge.

lalitb:
Thanks @marcalff , @ThomsonTan . Merging it.

Issue #1509, opened 2022-07-21
url parser will incorrect parsing url like "http://aaa.com/eee@bbb/x/y"


wangzhen2271:
I report trace to url "http://tracing-analysis-dc-hz.aliyuncs.com/adapt_fdufe5deml@455595430f299c1_fdffe5degl@134f7ad2ffe83/api/otlp/traces".  It always failed,  I trace the code and found the urlparser take a mistake,  the parse result is:  host -- 455595430f299c1_fdffe5degl@134f7ad2ffe83 while it should be -- tracing-analysis-dc-hz.aliyuncs.com. 

I think the "**@**" should be considered just when it in the first segment,  so this should be fixed.

I want to submit a pull request to fix it.

![image](https://user-images.githubusercontent.com/24283291/180213131-a0e8f2c9-cb42-443c-baa8-50a07c03da4f.png)

esigo:
already fixed in #1511.

PR #1511, opened 2022-07-21
fix: urlParser will incorrect parsing url like http://abc.com/xxx@xxx/a/b

wangzhen2271:
Fixes  issue #1509

## Changes

urlParser incorrect url which "@" not in first segment, like "http://abc.com/xxx@xxx".
So I modified that it will judge if "@" is in first segment, that means before the first "/",  otherwise ignore handle it.

### DIFF
--- a/ext/include/opentelemetry/ext/http/common/url_parser.h
+++ b/ext/include/opentelemetry/ext/http/common/url_parser.h
@@ -52,11 +52,16 @@ public:
     }
 
     // credentials
-    pos = url_.find_first_of("@", cpos);
-    if (pos != std::string::npos)
+    size_t pos1 = url_.find_first_of("@", cpos);
+    size_t pos2 = url_.find_first_of("/", cpos);
+    if (pos1 != std::string::npos)
     {
       // TODO - handle credentials
-      cpos = pos + 1;
+      if (pos2 == std::string::npos || pos1 < pos2)
+      {
+        pos  = pos1;
+        cpos = pos1 + 1;
+      }
     }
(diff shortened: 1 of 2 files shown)


ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -52,11 +52,17 @@ class UrlParser
...
+    size_t pos2 = url_.find_first_of("/", cpos);
+    if (pos1 != std::string::npos)
     {
       // TODO - handle credentials
-      cpos = pos + 1;
+      if (pos2 == std::string::npos || pos1 < pos2)
+      {
+        pos = pos1;
+        cpos = pos + 1;
+      }
```
lalitb, line 64: Nit
```cpp
        pos = pos1;
        cpos = pos + 1;
```

to

```
   cpos = pos1 + 1;
```

lalitb (approved):
Thanks for fix. You need to fix the code format as described here - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#how-to-send-pull-requests

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -52,11 +52,17 @@ class UrlParser
...
+
```
ThomsonTan, line 65: nit: remove this empty line.

ThomsonTan (approved):

owent (approved):
LGTM

PR #1512, opened 2022-07-22
Inline print_value() in ostream exporter

ahadnagy:
## Changes

The common_utils.h header is being used in multiple build targets in the ostream exporter. This header defines the print_value() function, and because it's not inlined, it causes the build to fail in case it tries to link to more than one ostream exporter.

Solution: inline function.


lalitb (approved):
Thanks for fix.

esigo (approved):
LGTM
Thanks for the fix :)

PR #1514, opened 2022-07-24
fix compiler warnings

esigo:
Fixes some compiler warnings.

## Changes

Fixes unused success result and unused variable declaration.


sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -148,6 +149,12 @@ class Meter final : public opentelemetry::metrics::Meter
           storage_registry_[instrument_descriptor.name_] = storage;
           return true;
         });
+    if (!success)
```
lalitb, line 152: nit - I think control will never reach here because View API will create the default view if no matching view is found. But no harm in keeping it, as the View API can change in the future.

lalitb (approved):
LGTM

PR #1518, opened 2022-07-26
Don't require applications using jaeger exporter to know about libcurl

### LINKED ISSUE #1517: CMake users using jaeger exporter are forced to know about internal opentelemetry details
**Describe your environment** 
Linux Fedora 36, Opentelemetry built with shared libs, building with CMake

**Steps to reproduce**
Build code that uses the jaeger exporter using a CMake file like:
```
target_link_libraries(qpid-proton-cpp LINK_PRIVATE ...
opentelemetry-cpp::trace opentelemetry-cpp::jaeger_trace_exporter opentelemetry-cpp::ostream_span_exporter)
```
Build with no failure!

```
CMake Warning (dev) at cpp/CMakeLists.txt:122 (add_library):
  Policy CMP0028 is not set: Double colon in target name means ALIAS or
  IMPORTED target.  Run "cmake --help-policy CMP0028" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  Target "qpid-proton-cpp" links to target "CURL::libcurl" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
This warning is for project developers.  Use -Wno-dev to suppress it.
```
...
```
/usr/bin/ld: cannot find -lCURL::libcurl
collect2: error: ld returned 1 exit status
```

**Additional context**
There is a work around for this issue which is to use ```find_package(CURL REQUIRED)``` in the CMake files. However this makes no real sense as using libcurl is an internal detail of the opentelemetry-cpp sdk jaager exporter via the http_curl library and has no reason to be exposed to the user.

I'm also led to think it is an oversight due to it not being documented.

The simple fix that works for me is to make the CURL::libcurl link PRIVATE in ```ext/src/http/client/curl/CMakeLists.txt```

astitcher:
Fixes #1517

## Changes

Changed link of libcurl (CMake CURL::libcurl) to opentelemetry_http_client_curl to be PRIVATE. This fixes the build issue for me on my Linux system using shared libraries, however I've not tested this anywhere else.

### DIFF
--- a/ext/src/http/client/curl/CMakeLists.txt
+++ b/ext/src/http/client/curl/CMakeLists.txt
@@ -8,13 +8,17 @@ if(CURL_FOUND)
                         PROPERTIES EXPORT_NAME http_client_curl)
 
   if(TARGET CURL::libcurl)
-    target_link_libraries(opentelemetry_http_client_curl
-                          PUBLIC opentelemetry_ext CURL::libcurl)
+    target_link_libraries(
+      opentelemetry_http_client_curl
+      PUBLIC opentelemetry_ext
+      PRIVATE CURL::libcurl)
   else()
     target_include_directories(opentelemetry_http_client_curl
                                INTERFACE "${CURL_INCLUDE_DIRS}")
-    target_link_libraries(opentelemetry_http_client_curl
-                          PUBLIC opentelemetry_ext ${CURL_LIBRARIES})
+    target_link_libraries(
+      opentelemetry_http_client_curl
+      PUBLIC opentelemetry_ext
+      PRIVATE ${CURL_LIBRARIES})
   endif()
 
   install(


esigo:
@astitcher thanks for the PR, could you please fix the format as explained [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#how-to-send-pull-requests)?

astitcher:
> @astitcher thanks for the PR, could you please fix the format as explained [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#how-to-send-pull-requests)?

Hmm, I didn't see anywhere there that referenced the cmake formatting spec or the tool to do it - can you be more specific?

esigo:
> > @astitcher thanks for the PR, could you please fix the format as explained [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#how-to-send-pull-requests)?
> 
> Hmm, I didn't see anywhere there that referenced the cmake formatting spec or the tool to do it - can you be more specific?

 This part I mean:
```
git checkout -b feature
# edit files
tools/format.sh
git commit
git push fork feature
```

so running `tools/format.sh` should do the job.

astitcher:
Thanks - looks like I need to install cmake-format...

ThomsonTan:
> Thanks - looks like I need to install cmake-format...

For one file change, you may also try apply the diff mentioned int he build log (see below link) which should fix the format.

https://github.com/open-telemetry/opentelemetry-cpp/runs/7522769448?check_suite_focus=true

esigo (approved):
LGTM
Thanks for the PR :)

ThomsonTan (approved):

owent:
If we build `opentelemetry_http_client_curl` as static library, we still required to let the linker knows and links curl before `opentelemetry_http_client_curl` if any library use it. This PR breaks this dependency, I wonder if is there any better solution to solve this problem?

[1]: https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:LINK_LIBRARY

astitcher:
Are you sure this is a problem? I built opentelemetry-cpp with the jaeger exporter as a static library on my Linux machine and CMake did the right thing (CMake is documented to treat static and shared transitive dependencies differently).
The import module that cmake generated added libcurl and libthrift to the libraries l;inked for the final executable which is the previous behavior requiring the application to add find_package(CURL ...) and find_package(thrift ...).
@owent could you try building/installing with main and seeing if you actually have a problem compared to the previous situation - because for me static libraries are unchanged.

owent:
> Are you sure this is a problem? I built opentelemetry-cpp with the jaeger exporter as a static library on my Linux machine and CMake did the right thing (CMake is documented to treat static and shared transitive dependencies differently). The import module that cmake generated added libcurl and libthrift to the libraries l;inked for the final executable which is the previous behavior requiring the application to add find_package(CURL ...) and find_package(thrift ...). @owent could you try building/installing with main and seeing if you actually have a problem compared to the previous situation - because for me static libraries are unchanged.

You are right, sorry for my mistake and please ignore my reply.

PR #1507, opened 2022-07-20
Upgrade `instrumentation_library_metrics` to `scope_metrics`

owent:
Signed-off-by: owentou <owentou@tencent.com>

Fixes #1492
Fixed #1506 

## Changes

+ Change the name of `instrumentation_info_metric_data_` in `ResourceMetrics` and `instrumentation_info_metrics` in `MetricCollector::Collect`
+ `PopulateResourceMetrics` to `scope_metrics` in OTLP exporters
+ Rename `InstrumentationInfoMetrics` to `ScopeMetrics`
+ Rename `instrumentation_library_` to `scope_`
+ Rename `instrumentation_metrics` to `scope_metrics`
+ Rename `InstrumentationLibrary` to `InstrumentationScope`
+ Rename `GetInstrumentationLibrary` to  `GetInstrumentationScope`
+ Rename `SetInstrumentationLibrary` to  `SetInstrumentationScope`
+ Rename `printInstrumentationLibrary` to `printInstrumentationScope`
+ Rename `instrumentation_library_` to `instrumentation_scope_`
+ Rename `instrumentation_library` to `instrumentation_library`
+ Move instrumentationlibrary/instrumentation_library.h to instrumentationscope/instrumentation_scope.h
+ Add deprecated message for legacy `GetInstrumentationLibrary` and `SetInstrumentationLibrary`

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1523, opened 2022-07-28
fix(metrics): ostream exporter should print out resource attributes

bsarden:
While debugging my `Resource` definition for a metrics exporter, I realized
the `OStreamExporter` did not support printing out `Resource` contents. 

I plan on expanding upon the existing unit(s) for checking the resource
attributes, but wanted to "test the waters" first. 

My goal was to attach git branch/sha information to metrics that are collected
for easier A/B testing, and a `Resource` seemed like a natural place for that. 

For example, using the attributes:
```
service.git.branch
service.git.commit_sha
service.git.commit_sha_short
```

However, there seems to be some debate around [using resources and metrics
with prometheus](https://github.com/open-telemetry/opentelemetry-specification/issues/2640), which is what led me down my debugging path of using the ostream exporter first.

### DIFF
--- a/exporters/ostream/src/metric_exporter.cc
+++ b/exporters/ostream/src/metric_exporter.cc
@@ -4,6 +4,7 @@
 #include <chrono>
 #ifndef ENABLE_METRICS_PREVIEW
 #  include <algorithm>
+#  include <map>
 #  include "opentelemetry/exporters/ostream/common_utils.h"
 #  include "opentelemetry/exporters/ostream/metric_exporter.h"
 #  include "opentelemetry/sdk/metrics/aggregation/default_aggregation.h"
@@ -85,7 +86,7 @@ sdk::common::ExportResult OStreamMetricExporter::Export(
 }
 
 void OStreamMetricExporter::printAttributes(
-    const std::unordered_map<std::string, sdk::common::OwnedAttributeValue> &map,
+    const std::map<std::string, sdk::common::OwnedAttributeValue> &map,
     const std::string prefix)
 {
   for (const auto &kv : map)
@@ -100,7 +101,12 @@ void OStreamMetricExporter::printResources(const opentelemetry::sdk::resource::R
   auto attributes = resources.GetAttributes();
   if (attributes.size())
   {
-    printAttributes(attributes, "\n\t");
+    // Convert unordered_map to map for printing so that iteration
+    // order is guaranteed.
+    std::map<std::string, sdk::common::OwnedAttributeValue> attr_map;
+    for (auto &kv : attributes)
+      attr_map[kv.first] = std::move(kv.second);
+    printAttributes(attr_map, "\n\t");
   }
 }
 
--- a/exporters/ostream/src/metric_exporter.cc
+++ b/exporters/ostream/src/metric_exporter.cc
@@ -79,13 +79,33 @@ sdk::common::ExportResult OStreamMetricExporter::Export(
 
   for (auto &record : data.scope_metric_data_)
   {
-    printInstrumentationInfoMetricData(record);
+    printInstrumentationInfoMetricData(record, data);
   }
   return sdk::common::ExportResult::kSuccess;
 }
 
+void OStreamMetricExporter::printAttributes(
+    const std::unordered_map<std::string, sdk::common::OwnedAttributeValue> &map,
+    const std::string prefix)
+{
+  for (const auto &kv : map)
+  {
+    sout_ << prefix << kv.first << ": ";
+    opentelemetry::exporter::ostream_common::print_value(kv.second, sout_);
+  }
+}
+
+void OStreamMetricExporter::printResources(const opentelemetry::sdk::resource::Resource &resources)
+{
+  auto attributes = resources.GetAttributes();
+  if (attributes.size())
+  {
+    printAttributes(attributes, "\n\t");
+  }
+}
+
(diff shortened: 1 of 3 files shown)


exporters/ostream/src/metric_exporter.cc
```diff
@@ -79,13 +79,33 @@ sdk::common::ExportResult OStreamMetricExporter::Export(
 
   for (auto &record : data.instrumentation_info_metric_data_)
   {
-    printInstrumentationInfoMetricData(record);
+    printInstrumentationInfoMetricData(record, data);
```
lalitb, line 82: Can we call `printResources` here and keep `printInstrumentationInfoMetricData` unchanged. Something like - 

```cpp
for (auto &record : data.instrumentation_info_metric_data_)
  {
    printResource(*data.resource_)
    printInstrumentationInfoMetricData(record);
}
```

lalitb: Also, ostream tests need to be updated too.

bsarden: I updated the tests, but factoring out the `printResource` function would be a lot more work, since `printInstrumentationInfoMetricData` currently requires a lock on `sout_`. Currently, the `trace_sdk` has the concept of a [Recordable interface](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/ostream/src/span_exporter.cc#L47-L96) that is used for encapsulating all recordable data under one abstraction. We'd need something similar on the metrics side to `printRecordable(recordable)` I think.

lalitb: yes agree, I see the issue with proper encapsulation. One minor suggestion, should we rename `printInstrumentationInfoMetricData` method to
`printMetricData(record, resource&) `?

lalitb (approved):
LGTM. Just a side note, there may be conflict resolution needed if #1507 is merged first.

exporters/ostream/src/metric_exporter.cc
bsarden: If it's alright, I'd like to leave the names as is to reduce the footprint of merge conflicts with #1507.

lalitb: yes good point. Thanks.

ThomsonTan (approved):

exporters/ostream/test/ostream_metric_test.cc
```diff
@@ -151,6 +156,11 @@ TEST(OStreamMetricsExporter, ExportHistogramPointData)
       "\n  counts     : [200, 300, 400, 500, ]"
       "\n  attributes\t\t: "
       "\n\ta1: b1"
+      "\n  resources\t:"
+      "\n\tservice.name: unknown_service"
+      "\n\ttelemetry.sdk.version: 1.5.0"
```
lalitb, line 161: Can we use the macro `OPENTELEMETRY_SDK_VERSION` instead of the hardcoding version? This way, we don't need to modify this part of the code with every new release.

bsarden: Good call. Done. There are some strange cmake failures that I'm having difficulties reproducing on my ubuntu 20.04 machine locally, but they are also not marked as "required" on the CI stages...

bsarden-rivos: @lalitb: What are the next steps I need to take here in order to debug? It's not clear to me what the next debugging steps are (or how some of these CI failures are relevant to a largely NFC patch like this one).

bsarden: @lalitb: What are the next steps I need to take here in order to debug? It's not clear to me what the next debugging steps are (or how some of these CI failures are relevant to a largely NFC patch like this one).

lalitb:
@bsarden - I see these two failing tests are running on Windows with OTLP exporter enabled, would it be possible to test your changes on Windows? Else, I can test it sometime this week.

bsarden:
@lalitb Thanks for the quick response! Unfortunately, I don't have access to a windows host for testing ;(.

exporters/ostream/src/metric_exporter.cc
```diff
@@ -79,13 +79,34 @@ sdk::common::ExportResult OStreamMetricExporter::Export(
...
   }
   return sdk::common::ExportResult::kSuccess;
 }
 
+void OStreamMetricExporter::printAttributes(
+    const std::unordered_map<std::string, sdk::common::OwnedAttributeValue> &map,
+    const std::string prefix)
+{
+  for (const auto &kv : map)
```
lalitb, line 91: @bsarden - Looks like this part of the code is causing problem. As it's unordered_map, the order of printing of resource elements is implementation specific ( so different in Windows and Linux). Can we ensure to print them in the given order - maybe move elements to std::map and then print? And also ensure that the same order is used for comparison in unit-test.

lalitb: @bsarden  - Just wondering if you have a plan to fix CI as mentioned above? Else, I can do it on top of your repo if you are fine :)

lalitb: Thanks for fixing it. CI looks good now :)

bsarden-rivos: Thanks for the ping and for reproducing on windows!

bsarden: Thanks for the ping and reproducing the error on a windows box!

PR #1524, opened 2022-07-28
Fix log sdk builder (#1486)

### LINKED ISSUE #1486: [Logs SDK] Provide builders to avoid exposing SDK internals
Same as issue #1393, but for Logs.

Please assign to me.

marcalff:
Fixes #1486

## Changes

Implement builders for log SDK.

* [X] Changes in public API reviewed (no API change)


sdk/src/logs/logger_provider_factory.cc
```diff
@@ -0,0 +1,58 @@
...
+namespace sdk
+{
+namespace logs
+{
+
+std::unique_ptr<opentelemetry::logs::LoggerProvider> LoggerProviderFactory::Create(
+    std::unique_ptr<LogProcessor> &&processor)
+{
+  auto resource = opentelemetry::sdk::resource::Resource::Create({});
+  return Create(std::move(processor), resource);
```
owent, line 20: I think move resource here will have a better performance?

owent (commented):
Just want to discuss, do we need a standalone factor for every processor?Or it's more simple to use just one factory to create different type of process and all return `std::unique_ptr<LogProcessor>`?

exporters/otlp/src/otlp_grpc_log_exporter_factory.cc
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#ifdef ENABLE_LOGS_PREVIEW
+
+// MUST be first (absl)
+#  include "opentelemetry/exporters/otlp/otlp_grpc_log_exporter.h"
+
+#  include "opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h"
+#  include "opentelemetry/exporters/otlp/otlp_grpc_log_exporter_factory.h"
+#  include "opentelemetry/sdk/common/global_log_handler.h"
```
lalitb, line 11: Do we need global_log_handler.h header here?

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
```diff
@@ -0,0 +1,66 @@
...
+   */
+  static std::unique_ptr<opentelemetry::logs::LoggerProvider> Create(
+      std::vector<std::unique_ptr<LogProcessor>> &&processors,
+      const opentelemetry::sdk::resource::Resource &resource);
+
+  /**
+   * Create a LoggerProvider.
+   */
+  static std::unique_ptr<opentelemetry::logs::LoggerProvider> Create(
+      std::shared_ptr<sdk::logs::LoggerContext> context);
```
lalitb, line 60: Similar to TracerContext, should we have provider class for LoggerContext too ?

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
```diff
@@ -0,0 +1,66 @@
...
+};
+
```
lalitb, line 62: Should we also have
```cpp
static std::unique_ptr<opentelemetry::logs::LoggerProvider> Create()
```
As it is possible to add `LogProcessor` instance after creating LoggerProvider?

lalitb:
> Just want to discuss, do we need a standalone factor for every processor?Or it's more simple to use just one factory to create different type of process and all return `std::unique_ptr<LogProcessor>`?

 I don't have strong preference for either approach - both looks good.

sdk/src/logs/logger_provider_factory.cc
marcalff: The resource parameter in Create is a `const &`, there is no copy.

exporters/otlp/src/otlp_grpc_log_exporter_factory.cc
marcalff: Thanks.

No we don't, removed.

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
marcalff: Yes, thanks.

Implemented.

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
marcalff: Create() returns the API abstraction, here` opentelemetry::logs::LoggerProvider`.

This class in the API does not have a way to add log processors,
only the SDK class `opentelemetry::sdk::logs::LoggerProvider` can.

The whole point it to hide the SDK class implementation,
which as a side effect also hides the extra methods the SDK can provide.

By using SDK classes directly, the application owner can:
- put all the log processors in a vector, and create a logger provider
- create a logger provider and add log processors one by one, calling `AddProcessor()`

By using the `Factory::Create()` builder pattern, only the first case is supported.

marcalff:
> Just want to discuss, do we need a standalone factor for every processor?Or it's more simple to use just one factory to create different type of process and all return `std::unique_ptr<LogProcessor>`?

@owent , @lalitb Good question, as it is another possible design.

Assuming something like:

class LogProcessorFactory
{
  static std::unique_ptr<LogProcessor> CreateSimpleLogProcessor(...);
  static std::unique_ptr<LogProcessor> CreateMultiLogProcessor(...);
  static std::unique_ptr<LogProcessor> CreateBatchLogProcessor(...);
};

there is only one class, with builders for each flavor of log processor.

This design works here for log processor, but does not for other cases.

For example applying the same pattern to trace exporters, it leads to:
- a class TraceExporterFactory
- that depends on the Jaeger exporter
- that depends on the OStream exporter
- that depends on the OTLP HTTP exporter
- that depends on the OTLP GRPC exporter

Because of this class, there is no longer a way to link only a few exporters the application owner wants in the final binary.

Instead, every exporter has to be linked in the final application, which is not good.

So, to avoid confusion for the user of this code, the design here sticks to always the same pattern,
which is easy to remember, and also works well with libraries dependencies:

**To get a instance of class XXX, call XXXFactory::Create()**

The XXXFactory class is provided in the same library as class XXX, which makes it possible to use XXX but not YYY.

owent (commented):
LGTM, Could you please also add a standalone unit test so that we can ensure factory headers of OTLP exporters do not include any protobuf header?

lalitb:
> So, to avoid confusion for the user of this code, the design here sticks to always the same pattern,
which is easy to remember, and also works well with libraries dependencies:
To get a instance of class XXX, call XXXFactory::Create()


Agree, the current approach looks good.

lalitb (approved):

marcalff:
> LGTM, Could you please also add a standalone unit test so that we can ensure factory headers of OTLP exporters do not include any protobuf header?

Indeed, I should have known from traces.

Thanks, unit test added.

lalitb:
> LGTM, Could you please also add a standalone unit test so that we can ensure factory headers of OTLP exporters do not include any protobuf header?

@owent  - Can you approve the changes if the recent unit tests look good? Thanks.

owent (approved):
LGTM

Issue #1520, opened 2022-07-26
Opentelemetry API singletons (TraceProvider etc.) do not play well in shared library uses that hide symbol visibility


astitcher:
**Describe your environment** 
We are using the opentelemetry tracing API (and SDK) inside a Linux C++ shared library (.so file) built with the standard gcc toolchain. This library is provides messaging using the AMQP protocol (the library is Qpid Proton C++). We create tracing spans internal to the library to represent the lifecycle of messages sent and received by the library. The contexts associated with the spans are propagated with the messages so that distributed tracing is achieved.

The intention is that any application using our library should be able to create its own tracing spans that are naturally related to the library generated spans (by default using the active span as the parent span). This requires that the TracingProvider accessible to the library and to the application using it is the same TracingProvider.

**Steps to reproduce**
Our library is built with ```-fvisibility=hidden```. This means that any symbol that should be exported from the library needs some extra annotation (```__attribute__((visibility("default")))``` for gcc/clang). We use this to carefully control the visible API/ABI from our library and to be sure that internal details aren't visible outside.

However the way that the singleton inside```Provider``` is implemented in a header file means that it is defined as a static symbol inside the inline member function ```Provider::GetProvider```. This means that there are duplicate symbols for every time the ```provider.h``` header file is included in for example our library and an application that uses it.

If the symbols are all visible to the linker at link and runtime then the symbol will be effectively deduplicated and there will be only one used - to my understanding the application symbol takes precedence, but as long as there is only one it will work correctly.

However when the library symbols are hidden by default as in our library there is no way for the runtime loader to know that there should only be one version of the singleton and what happens is that the library and the application end up with different ```TracerProvider```s.

This is a serious problem for us as we ship our library with the symbols exported explicitly for very good reason and so we can't just use ```Provider::GetTracerProvider``` in the library and application and have it work correctly together.

**What is the expected behavior?**
The way I would expect this to work is that the singletons are not defined in header files and so duplicated in every object file that includes them, but rather the header file only declares ```Provider::GetProvider``` and it is defined in one of the opentelemetry-cpp libraries (the common lib?). I understand that this moves the singleton out of what you think of as the 'API' and into the 'SDK'. But I really think that architecturally singletons are actuall implementation artifacts not API artifiacts.

**Additional context**
I note that this is not an issue at all if you statically link everything as in that case the final link phase takes care of the issue.

[...]

PR #1525, opened 2022-07-28
[POC RFC] header only API for singletons

marcalff:
Fix header only API for singletons (#1520)

## Changes

DRAFT -- implement header only API singletons.


lalitb:
Did a quick test with this PR on Linux, by creating an instrumented library with `-fvisible=hidden` flag and the library is able to use the singleton tracer provider from the application. So changes seem to work on Linux.

Need to check on Windows, but based on what @owent has already tested [here](https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171), it may not work with PE ABI as `__declspec(dllexport)` needs to be held inside the compile unit.

owent:
I have add more examples [here](https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171), and `__declspec (selectany)` of MSVC works well, but `__attribute__ ((selectany))` of GCC on Windows do not work as expected,do I miss anything?

marcalff:
Paper from Ulrich Drepper, see section 2.2 export control, page 17:

https://akkadia.org/drepper/dsohowto.pdf

owent:
> Resources:
> 
> 
> Paper from Ulrich Drepper, see section 2.2 export control, page 17:
> 
> https://akkadia.org/drepper/dsohowto.pdf
> 
> 
> Gnu visibility doc
> 
> https://gcc.gnu.org/wiki/Visibility

I have a quick look at https://akkadia.org/drepper/dsohowto.pdf, I think it's for ELF ABI of Linux, not PE ABI of Windows.
And the `visibility` attribute of gcc document(https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes) also says ***The possible values of visibility_type correspond to the visibility settings in the ELF gABI.*** .

My question is: is there any way to make the singleton unique with GCC on Windows?

lalitb:
@owent - We agreed this PR to be scoped for Linux and Mac. Let us know if you have any comment on that.

owent:
> @owent - We agreed this PR to be scoped for Linux and Mac. Let us know if you have any comment on that.

Fine by me. I think we can document this if we do not support build shared library with gcc and clang on Windows, can we also force set `BUILD_SHARED_LIB` to `OFF` when we use gcc and clang on Windows to reduce misuse?

@lalitb @marcalff

owent (approved):
LGTM, Do we need set `BUILD_SHARED_LIB` to `OFF` with gcc on Windows to avoid misuse?

marcalff:
@lalitb , @esigo , @owent 

I added a singleton_test unit test, and makefiles for both CMake and Bazel.

This test does not build properly on windows platforms, and I can not figure out why.

"component_c" is compiled as a shared library / DDL (expected),
but then when linking the final singleton_test binary, the makefile wants a component_c.lib instead of component_c.ddl,
which is incorrect.

On linux, the same makefile with cmake properly builds static libraries (component_a, component_b) and shared libraries (component_c, d, e and f) as expected.

For bazel, I could not find a proper way to express in the BUILD file how to link a test binary singleton_test with a mix of static (component_a and b) and shared (component_c, d, e and f) libraries.

Any help appreciated, the goal is to have first a unit test that builds everywhere.

Regards,
-- Marc

api/include/opentelemetry/metrics/provider.h
```diff
@@ -29,8 +44,8 @@ class Provider
    */
   static nostd::shared_ptr<MeterProvider> GetMeterProvider() noexcept
   {
-    std::lock_guard<common::SpinLockMutex> guard(lock);
-    nostd::shared_ptr<MeterProvider> result(provider);
+    std::lock_guard<common::SpinLockMutex> guard(s.lock);
```
owent, line 47: I think we should not use a short name here

owent:
> @lalitb , @esigo , @owent
> 
> I added a singleton_test unit test, and makefiles for both CMake and Bazel.
> 
> This test does not build properly on windows platforms, and I can not figure out why.
> 
> "component_c" is compiled as a shared library / DDL (expected), but then when linking the final singleton_test binary, the makefile wants a component_c.lib instead of component_c.ddl, which is incorrect.
> 
> On linux, the same makefile with cmake properly builds static libraries (component_a, component_b) and shared libraries (component_c, d, e and f) as expected.
> 
> For bazel, I could not find a proper way to express in the BUILD file how to link a test binary singleton_test with a mix of static (component_a and b) and shared (component_c, d, e and f) libraries.
> 
> Any help appreciated, the goal is to have first a unit test that builds everywhere.
> 
> Regards, -- Marc

Do we need declare `do_something_in_c` as default visibility or `__declspec(dllexport)/__declspec(dllimport)` ? I find `component_c` is built as a shared library but export nothing.

To my knowledge, MSVC will create both `.lib` and `.dll` for dynamic libraries. `.lib` is for linking and `.dll` is for runtime.

marcalff:
> Do we need declare `do_something_in_c` as default visibility or `__declspec(dllexport)/__declspec(dllimport)` ? I find `component_c` is built as a shared library but export nothing.

Good point, fixing this.

Thanks

marcalff:
This PR contains proof of concept code, for a general fix.

It is now closed, for the benefit of:
- PR #1604, a fix for gcc and clang, ready for merge
- PR #1595, a tentative fix for windows, still a proof of concept.

Work for windows should continue in PR #1595.

PR #1530, opened 2022-07-30
Fix code scanning alert

### LINKED ISSUE #1529: Fix code scanning alert - Failure to use HTTPS URLs
Tracking issue for:
- [ ] https://github.com/open-telemetry/opentelemetry-cpp/security/code-scanning/1

esigo:
Fixes #1529 (issue)

## Changes

uses HTTPS URLs


ext/test/http/curl_http_test.cc
```diff
@@ -329,7 +329,7 @@ TEST_F(BasicCurlHttpTests, SendGetRequestSyncTimeout)
   curl::HttpClientSync http_client;
 
   http_client::Headers m1 = {};
-  auto result             = http_client.Get("http://222.222.222.200:19000/get/", m1);
```
ThomsonTan, line 332: Wondering why is the IP address changed? I think it is intended as some invalid/unreachable address. @lalitb

owent (approved):
LGTM

ext/test/http/curl_http_test.cc
owent: I think this address is invalid either

ThomsonTan: Some local app could listen on https://127.0.0.1:19000?

lalitb: We can also use the address from `Documentation Address Block` (https://datatracker.ietf.org/doc/html/rfc5737#section-4), as they are not allowed to be used on the public internet.
```
192.0.2.0 – 192.0.2.255
198.51.100.0 – 198.51.100.255
233.252.0.0 - 233.252.0.255
```

lalitb (approved):

ThomsonTan (approved):

esigo:
Hi @owent, do you have any idea why the `asan` test [fails](https://github.com/open-telemetry/opentelemetry-cpp/runs/7616756918?check_suite_focus=true#step:5:1022):
```
ext/test/http/curl_http_test.cc:429: Failure
Value of: sessions[i]->IsSessionActive()
  Actual: false
Expected: true
```
I can't reproduce it locally.
thanks

owent (approved):

owent:
> Hi @owent, do you have any idea why the `asan` test [fails](https://github.com/open-telemetry/opentelemetry-cpp/runs/7616756918?check_suite_focus=true#step:5:1022):
> 
> ```
> ext/test/http/curl_http_test.cc:429: Failure
> Value of: sessions[i]->IsSessionActive()
>   Actual: false
> Expected: true
> ```
> 
> I can't reproduce it locally. thanks

I found some implements of libcurl will connect failed immediately and call the callback in the IO thread.And the callback will reset `is_session_active_` to false.Maybe we can add `GetEventHandler::OnEvent()` to set `is_called_` and use `ASSERT_TRUE(sessions[i]->IsSessionActive() || handlers[i]->is_called_);` here.

esigo:
> > Hi @owent, do you have any idea why the `asan` test [fails](https://github.com/open-telemetry/opentelemetry-cpp/runs/7616756918?check_suite_focus=true#step:5:1022):
> > ```
> > ext/test/http/curl_http_test.cc:429: Failure
> > Value of: sessions[i]->IsSessionActive()
> >   Actual: false
> > Expected: true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I can't reproduce it locally. thanks
> 
> Some implements of libcurl will connect failed immediately and call the callback in the IO thread if it try to connect a invalid address. And the callback will reset `is_session_active_` to false.Maybe we can add `GetEventHandler::OnEvent()` to set `is_called_` and use `ASSERT_TRUE(sessions[i]->IsSessionActive() || handlers[i]->is_called_);` here.

Thanks @owent, it fixed :)

PR #1544, opened 2022-08-06
[NFC] Improve scope/instrument names in metrics ostream exporter

bsarden:
Previously, we were overloading the `name` key in the metrics ostream
exporter. This patch qualifies the `name` key with the appropriate prefix
based on the structures they represent.

### DIFF
--- a/exporters/ostream/src/metric_exporter.cc
+++ b/exporters/ostream/src/metric_exporter.cc
@@ -117,14 +117,14 @@ void OStreamMetricExporter::printInstrumentationInfoMetricData(
   // sout_ is shared
   const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
   sout_ << "{";
-  sout_ << "\n  name\t\t: " << info_metric.scope_->GetName()
+  sout_ << "\n  scope name\t: " << info_metric.scope_->GetName()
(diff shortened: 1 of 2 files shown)


exporters/ostream/src/metric_exporter.cc
```diff
@@ -117,14 +117,14 @@ void OStreamMetricExporter::printInstrumentationInfoMetricData(
   // sout_ is shared
   const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
   sout_ << "{";
-  sout_ << "\n  name\t\t: " << info_metric.scope_->GetName()
+  sout_ << "\n  scope name\t: " << info_metric.scope_->GetName()
```
bsarden, line 120: I was unsure if this should really be `scope` or `meter` (since the `meter` is what implicitly creates this scope). WDYT?

lalitb: I think `scope` is the right terminology in compliance with the specs.

lalitb (approved):
LGTM. Thanks for the changes.

exporters/ostream/src/metric_exporter.cc
lalitb: I think `scope` is the right terminology in compliance with the specs.

PR #1554, opened 2022-08-11
METRICS SDK - Calling Observable Instruments callback during metrics collection

### LINKED ISSUE #1550: Nothing calls ObservableRegistry::Observe()
I've been trying to use async instruments, but got slightly confused about the API and the status of the implementation.

The problem I ran into is that the callbacks are not getting evaluated and no data is exported. I see that the API has recently changed in https://github.com/open-telemetry/opentelemetry-cpp/pull/1388 that introduced `ObservableRegistry`. It seems like the `ObservableRegistry::Observe()` method would be the one evaluating all the registered callbacks, but it doesn't get called anywhere. According to the corresponding issue (https://github.com/open-telemetry/opentelemetry-cpp/issues/1338), all the callbacks should be evaluated before/during `collect()`.

Am I missing something here? I don't see any open issues or PRs for async instruments. If this is a bug, I'd glad to help out with a PR after some explanation about where and how this should happen.

Thanks a lot in advance!

lalitb:
Fixes #1550, #1556

## Changes

1. Create ObservableRegistry during Meter creation.
2. Store the reference to ObservableRegistry while Instruments created from that Meter.
3. During the collection (in `Meter::Collect()`), call all the registered callbacks once.

### DIFF
--- a/sdk/test/metrics/meter_test.cc
+++ b/sdk/test/metrics/meter_test.cc
@@ -38,17 +38,6 @@ nostd::shared_ptr<metrics::Meter> InitMeter(MetricReader **metricReaderPtr)
   p->AddMetricReader(std::move(metric_reader));
   auto meter = provider->GetMeter("meter_name");
   return meter;
-  // auto sdk_meter = std::static_pointer_cast<opentelemetry::sdk::metrics::Meter>(meter);
-
-  /*
-    provider->
-    auto context = std::make_shared<opentelemetry::sdk::metrics::MeterContext>();
-    std::unique_ptr<MetricReader> metric_reader(new MockMetricReader());
-    *metricReaderPtr = metric_reader.get();
-    context->AddMetricReader(std::move(metric_reader));
-    std::shared_ptr<Meter> meter(new Meter(context));
-    return meter;
-    */
 }
 }  // namespace
 
--- a/sdk/test/metrics/meter_test.cc
+++ b/sdk/test/metrics/meter_test.cc
@@ -23,9 +23,9 @@ public:
   {
     return AggregationTemporality::kCumulative;
   }
-  virtual bool OnShutDown(std::chrono::microseconds timeout) noexcept override { return true; }
-  virtual bool OnForceFlush(std::chrono::microseconds timeout) noexcept override { return true; }
-  virtual void OnInitialized() noexcept override {}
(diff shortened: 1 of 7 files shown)


esigo:
is this finished?

lalitb:
> is this finished?

Yes changes are fine, as I tested with ostream exporter, I need to add unit test. Will be finishing it today.

lalitb:
> > is this finished?
> 
> Yes changes are fine, as I tested with ostream exporter, I need to add unit test. Will be finishing it today.

I have now added basic tests to validate the async flow, it should be good to review now.

sdk/test/metrics/meter_test.cc
```diff
@@ -0,0 +1,75 @@
...
+void asyc_generate_measurements(opentelemetry::metrics::ObserverResult observer, void *state)
+{
+  auto observer_long =
+      nostd::get<nostd::shared_ptr<opentelemetry::metrics::ObserverResultT<long>>>(observer);
+  observer_long->Observe(10l);
+}
+
+TEST(MeterTest, BasicAsyncTests)
+{
+  MetricReader *metricReaderPtr = nullptr;
```
esigo, line 54: nit: could we use `metric_reader_ptr` istead?

esigo (approved):
LGTM
Thanks for the fix.

sdk/test/metrics/meter_test.cc
lalitb: fixed. thanks for reviewing.

sdk/test/metrics/meter_test.cc
```diff
@@ -0,0 +1,75 @@
...
+class MockMetricReader : public MetricReader
+{
+public:
+  MockMetricReader() = default;
+  AggregationTemporality GetAggregationTemporality(
+      InstrumentType instrument_type) const noexcept override
+  {
+    return AggregationTemporality::kCumulative;
+  }
+  virtual bool OnShutDown(std::chrono::microseconds timeout) noexcept override { return true; }
```
ThomsonTan, line 26: nit, the virtual keywords are not needed.

lalitb: Thanks fixed now

ThomsonTan (approved):

PR #1555, opened 2022-08-11
[Metrics SDK] Fix IWYU error in instruments.h

bsarden:


lalitb (approved):
Thanks for the fix. std::function was added recently, header got missed.

PR #1560, opened 2022-08-11
remove throw check from metrics  with noexcept

### LINKED ISSUE #1559: remove throw check from metrics tests with noexcept
we have exception checks in metrics test which cannot throw anyway. These checks has to be removed so we can use the tests in `bazel.noexcept`.

esigo:
Fixes #1559 (issue)

## Changes


lalitb (approved):
Thanks for the fix.

ThomsonTan (approved):

PR #1581, opened 2022-08-25
Add error log when getting a http error code

### LINKED ISSUE #1578: OTLP HTTP exporter errors are never reported
**Describe your environment**
nothing relevant

**Steps to reproduce**
Configure the http exporter with an URL that will parse correctly but will report errors, e.g. 404 because of an invalid export path

I expected to see something in the log (at least DEBUG) about a failure to export

A log line:
    `[OTLP HTTP Client] DEBUG: Export 1 trace span(s) success`

**Additional context**
n/a

I'm not sure if this should be considered a bug report or a feature request.  But it's very surprising to me that a misconfigured `OtlpHttpClientOptions.url` can result in no errors.  In my case I had the right http address but an incorrect path, and the collector was replying with a 404.  This error showed up nowhere.

I can see that the `otlp_http_client.cc`'s `ResponseHandler::OnResponse` does see the 404, but without `console_debug_` set this isn't even logged.  The `OnResponse` function returns `void` so it doesn't seem to have a way to report this error to the calling framework.

I expected at least an `OTEL_INTERNAL_LOG_DEBUG` from `OnRespnse()` about the error, but ideally it would actually return a status to the framework that was propagated to the return value from `Export()`.

owent:

Fixes #1578 

## Changes

+ Add error log when OTLP HTTP exportor and ES exportor get a http error code

### DIFF
--- a/exporters/elasticsearch/src/es_log_exporter.cc
+++ b/exporters/elasticsearch/src/es_log_exporter.cc
@@ -31,12 +31,29 @@ public:
    */
   ResponseHandler(bool console_debug = false) : console_debug_{console_debug} {}
 
+  std::string BuildResponseLogMessage(http_client::Response &response,
+                                      const std::string &body) noexcept
+  {
+    std::stringstream ss;
+    ss << "Status:" << response.GetStatusCode() << "Header:";
+    response.ForEachHeader([&ss](opentelemetry::nostd::string_view header_name,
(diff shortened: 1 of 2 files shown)


lalitb (approved):
LGTM. Thanks for the fix.

esigo (approved):

PR #1594, opened 2022-09-06
POC fix for GCC/Clang under Linux for header only singletons

### LINKED ISSUE #1520: Opentelemetry API singletons (TraceProvider etc.) do not play well in shared library uses that hide symbol visibility

astitcher:
## Changes

This the minimal change that will work for GCC with Linux. This approach
won't work for Windows. MacOS is unknown but might work.

Fixes #1520

    * This is a very small perhaps trivial change
    * No changes to public API

### DIFF
--- a/api/include/opentelemetry/_metrics/provider.h
+++ b/api/include/opentelemetry/_metrics/provider.h
@@ -7,6 +7,7 @@
 
 #  include "opentelemetry/_metrics/meter_provider.h"
 #  include "opentelemetry/_metrics/noop.h"
+#  include "opentelemetry/common/macros.h"
 #  include "opentelemetry/common/spin_lock_mutex.h"
 #  include "opentelemetry/nostd/shared_ptr.h"
 
@@ -41,13 +42,13 @@ public:
   }
 
 private:
-  static nostd::shared_ptr<MeterProvider> &GetProvider() noexcept
+  OPENTELEMETRY_API_SINGLETON static nostd::shared_ptr<MeterProvider> &GetProvider() noexcept
   {
     static nostd::shared_ptr<MeterProvider> provider(new NoopMeterProvider);
     return provider;
   }
 
-  static common::SpinLockMutex &GetLock() noexcept
+  OPENTELEMETRY_API_SINGLETON static common::SpinLockMutex &GetLock() noexcept
   {
     static common::SpinLockMutex lock;
     return lock;
--- a/api/include/opentelemetry/baggage/baggage.h
+++ b/api/include/opentelemetry/baggage/baggage.h
@@ -6,6 +6,7 @@
 #include <cctype>
 
 #include "opentelemetry/common/kv_properties.h"
+#include "opentelemetry/common/macros.h"
 #include "opentelemetry/nostd/shared_ptr.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/version.h"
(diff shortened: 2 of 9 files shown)


astitcher:
I have created this PR because this is the patch that is going to be used by the Fedora/RHEL rpm packages until this issue is resolved in an official release of opentelemetry-cpp.

I think this is a fine approach for Linux etc. and it is about the minimal change I could find. I'm completely fine with this not being merged though if there is a better approach that might work for Windows/Visual Studio too.

owent:
Is this duplicated of #1595 ?

marcalff:
PR #1595 tests the api fix from this branch, with unit tests implemented by PR #1525 

In general, the fix works for gcc and clang, with the exception of a crash seen when using the STL.

See CI / CMake C++20 test

astitcher:
@marcalff Thank you for #1595. should I close this PR in favor of that one?

marcalff:
> @marcalff Thank you for #1595. should I close this PR in favor of that one?

@astitcher ,

Thanks for also testing on your side.

The change for gcc and clang is the similar in PR#1595 and here, with added unit tests in 1595,
so yes, I don't see the need to keep this PR open.

PR#1595 contains more code for windows, but if this can not be made to work I will cleanup the extra ifdef when merging, so it will look like this PR here if only the gcc/clang fix is merged.

astitcher:
Closing this in favor of #1595

PR #1595, opened 2022-09-06
[POC RFC] header only API for singletons (Windows)

marcalff:
Fix header only API for singletons (#1520)

## Changes

Alternate solution, test annotation on methods instead of members


marcalff:
The first commit fails 18 checks, which is actually **good**.

This commit only contains the unit test, which fails, showing that the unit test logic is working.

api/include/opentelemetry/common/macros.h
```diff
@@ -99,7 +99,8 @@
 #elif defined(__GNUC__)
 #  define OPENTELEMETRY_API_SINGLETON __attribute__((visibility("default")))
 #elif defined(_MSC_VER)
-#  define OPENTELEMETRY_API_SINGLETON
+/* Tentative fix */
+#  define OPENTELEMETRY_API_SINGLETON __declspec(selectany)
```
owent: Just wondering why `__declspec(selectany)` is removed?Doesn't it work with MSVC on Windows?

marcalff: @owent 

`__declspec(selectany)` works when applied to a member.

This PR tested applying it to a method, which failed to compile.

See places where `OPENTELEMETRY_API_SINGLETON` is used in this patch,
which are different compared to PR#1525

owent: Could we also add `__declspec(selectany)` to variables, so we can also support dll on Windows with MSVC?

astitcher:
I think that the only way to make Windows dll work is to have the ```GetXxxx()``` singleton accessor member functions in their own library. At least in my windows testing ```__declspec(selectany)``` with member data didn't work to deduplicate the singleton across the executable and the dll.
One approach to this might be to pull these members out of the class definitions and make them inline in the header files for Linux etc. but ```#ifdef``` the definitions for Windows and include them in a source file there.

marcalff:
Status as of Sept 7:
- fix for gcc and clang works for linux and macos.
- the unit test added, singleton_test, builds, and passes, for linux and macos.
- the unit test builds for windows, links properly with a mix of .LIB and .DLL, and fails at runtime.
- the previous issue (segfault) affecting WITH_STL was due to a makefile issue in CMake, and is fixed.

So, at this point, we can:
- either continue to investigate on windows, to have a full solution
- or merge a partial fix for non windows platforms, to be amended later.

I don't have access to a windows machine, and can not debug, so help is most welcome.

Testing is needed on windows to see if the fix can at least work with .LIB alone.

Testing is also needed to see if gcc/clang on windows can work, and see if changes are needed in the macros.h header for that.

If it turns out there is no solution for windows, then the only option will be to have a windows DLL for the api singletons, which is causing other issues.

owent:
> I think that the only way to make Windows dll work is to have the ```GetXxxx()``` singleton accessor member functions in their own library. At least in my windows testing ```__declspec(selectany)``` with member data didn't work to deduplicate the singleton across the executable and the dll.
> One approach to this might be to pull these members out of the class definitions and make them inline in the header files for Linux etc. but ```#ifdef``` the definitions for Windows and include them in a source file there.

Could you please show the codes about the not working `__declspec(selectany)` above?
I have written some small samples at https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171 , and in those codes `selectany` works well with MSVC, but don't work only with gcc and clang on Windows.

owent:
> See `file api/include/opentelemetry/common/macros.h` for the technical solution
> 
> Status as of Sept 7:
> - fix for gcc and clang works for linux and macos.
> - the unit test added, singleton_test, builds, and passes, for linux and macos.
> - the unit test builds for windows, links properly with a mix of .LIB and .DLL, and fails at runtime.
> - the previous issue (segfault) affecting WITH_STL was due to a makefile issue in CMake, and is fixed.
> 
> So, at this point, we can:
> - either continue to investigate on windows, to have a full solution
> - or merge a partial fix for non windows platforms, to be amended later.
> 
> I don't have access to a windows machine, and can not debug, so help is most welcome.
> 
> Testing is needed on windows to see if the fix can at least work with .LIB alone.
> 
> Testing is also needed to see if gcc/clang on windows can work, and see if changes are needed in the macros.h header for that.
> 
> If it turns out there is no solution for windows, then the only option will be to have a windows DLL for the api singletons, which is causing other issues.
> 

Agree to merge the visibility changes for unix like system first, and we can continue to find a better solution for Windows. Could you please show me the commit id which has problems with mix of .lib and .dll? I can help to test it some times later, the problem of failed CI job I reviewed before may be it doesn't export anything.

astitcher:
>     * either continue to investigate on windows, to have a full solution
> 
>     * or merge a partial fix for non windows platforms, to be amended later.

I would really like to get a solution for non windows into a release so we can create packages based on an actual opentelemetry-cpp release rather than taking one and patching it.

astitcher:
> I don't have access to a windows machine, and can not debug, so help is most welcome.
> 
> Testing is needed on windows to see if the fix can at least work with .LIB alone.

I'll try to find some time to test this with just static linking on Windows and to try to at least build this shared on Windows.

> If it turns out there is no solution for windows, then the only option will be to have a windows DLL for the api singletons, which is causing other issues.

My expectation is that this is unfortunately going to be the option that will we will end up with for Windows DLLs.

owent:
> > See `file api/include/opentelemetry/common/macros.h` for the technical solution
> > Status as of Sept 7:
> > 
> > * fix for gcc and clang works for linux and macos.
> > * the unit test added, singleton_test, builds, and passes, for linux and macos.
> > * the unit test builds for windows, links properly with a mix of .LIB and .DLL, and fails at runtime.
> > * the previous issue (segfault) affecting WITH_STL was due to a makefile issue in CMake, and is fixed.
> > 
> > So, at this point, we can:
> > 
> > * either continue to investigate on windows, to have a full solution
> > * or merge a partial fix for non windows platforms, to be amended later.
> > 
> > I don't have access to a windows machine, and can not debug, so help is most welcome.
> > Testing is needed on windows to see if the fix can at least work with .LIB alone.
> > Testing is also needed to see if gcc/clang on windows can work, and see if changes are needed in the macros.h header for that.
> > If it turns out there is no solution for windows, then the only option will be to have a windows DLL for the api singletons, which is causing other issues.
> 
> Agree to merge the visibility changes for unix like system first, and we can continue to find a better solution for Windows later.
> 
> I can help to test it some times later, the problem may be the dll can not be found by exe. Unlike Linux, the executable on Windows do not have a `RPATH/RUNPATH` like information which contain a search path of dependency shared library. It only search dll in PATH environment and the directory of executable file. I will verify if it's this problem later.

Sorry, but I found `__declspec(selectany)` doesn't work when linking the `.lib` of a shared library on Windows, and I didn't find a way to let cmake link the `.dll` file. I have update https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171 to record this case.
Maybe can we document this and find a better way in the future?

api/test/singleton/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+include(GoogleTest)
+
+add_library(component_a STATIC component_a.cc)
+
```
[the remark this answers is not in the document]
marcalff: Thanks, will fix.

marcalff:
@owent 

> Sorry, but I found `__declspec(selectany)` doesn't work when linking the `.lib` of a shared library on Windows, and I didn't find a way to let cmake link the `.dll` file. I have update [#1520 (comment)](https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171) to record this case.
> Maybe can we document this and find a better way in the future?

To clarify my understanding:
So this does not work for DLL, but because of makefile issues with cmake.
When using the proper link commands manually, the binary works for DLL ?

Does the fix work with simply static libraries (.LIB), for component_a and component_b alone ?

If so, at least we will have a solution for windows static libs only, which is better than none.

Thanks.

owent:
> @owent
> 
> > Sorry, but I found `__declspec(selectany)` doesn't work when linking the `.lib` of a shared library on Windows, and I didn't find a way to let cmake link the `.dll` file. I have update [#1520 (comment)](https://github.com/open-telemetry/opentelemetry-cpp/issues/1520#issuecomment-1198104171) to record this case.
> > Maybe can we document this and find a better way in the future?
> 
> To clarify my understanding: So this does not work for DLL, but because of makefile issues with cmake. When using the proper link commands manually, the binary works for DLL ?
> 
> Does the fix work with simply static libraries (.LIB), for component_a and component_b alone ?
> 
> If so, at least we will have a solution for windows static libs only, which is better than none.
> 
> Thanks.

> When using the proper link commands manually, the binary works for DLL ?
Yes.

> Does the fix work with simply static libraries (.LIB), for component_a and component_b alone ?

Static libraries (.LIB) do not has a standalone heap. I think it works well as before.

lalitb:
> > Does the fix work with simply static libraries (.LIB), for component_a and component_b alone ?
> 
> Static libraries (.LIB) do not has a standalone heap. I think it works well as before.

Yes, static libraries should work fine for all the platforms in the existing setup.

marcalff:
Status:

This PR, 1595, is kept open to continue investigation for windows.

The fix for gcc and clang, which is working, is to be merged in PR #1604.

marcalff:
Added the help wanted tag.

Not a windows expert, so anyone with more knowledge is welcome to investigate,
either using this patch as a base (if it helps) or proposing a different solution.

PR #1604, opened 2022-09-09
Fix header only api singletons (#1520)

marcalff:
This fix is for gcc and clang only.

Fixes # (1520)

## Changes

Add compiler annotations to make singleton methods visible.
This is needed by the linker to ensure uniqueness of singletons.


api/test/singleton/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+include(GoogleTest)
+
+add_library(component_a STATIC component_a.cc)
+target_link_libraries(component_a opentelemetry_api)
+
+add_library(component_b STATIC component_b.cc)
+target_link_libraries(component_b opentelemetry_api)
+
+add_library(component_c SHARED component_c.cc)
```
owent, line 12: Should we disable `component_c` .. `component_f` on Windows?

owent (commented):
Should we disable component_c .. component_f and keep component_a and component_b on Windows?

api/test/singleton/CMakeLists.txt
lalitb: Not sure if I am missing something, but why is this unit test successful for these components on Windows as of now?

owent: > Not sure if I am missing something, but why is this unit test successful for these components on Windows as of now?

Because unit test is disabled on Windows in https://github.com/marcalff/opentelemetry-cpp/blob/9bfbfbe9d5d3ef462d0a4709e50070da49d561e2/api/test/singleton/CMakeLists.txt#L42

lalitb: ok, thanks. Yes, make sense to disable building them for windows then.

marcalff: Disabled the entire unit tests for windows.

api/test/CMakeLists.txt
```diff
@@ -13,3 +13,4 @@ if(WITH_LOGS_PREVIEW)
 endif()
 add_subdirectory(common)
 add_subdirectory(baggage)
+add_subdirectory(singleton)
```
lalitb, line 16: Should we add this subdirectory only for the non-windows platform? This will ensure we don't build the singleton tests for both bazel and cmake.

lalitb (approved):
LGTM

owent (approved):

api/test/CMakeLists.txt
lalitb: Please ignore. My bad, this file is only for cmake. I think it is fine to build the bazel tests for windows if we can't avoid that.

marcalff:
> LGTM, can we merged once conflict is resolved.

@lalitb

Thanks.
Merge conflicts now resolved.

PR #1462, opened 2022-06-21
Upgrade to gRPC 1.46.3 (#1459)

marcalff:
Fixes #1459 

## Changes

Upgrade to gRPC 1.46.3

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

Issue #1459, opened 2022-06-21
[Dependencies] Upgrade to gRPC version 1.46.3


marcalff:
gRPC will require C++14 starting with release 1.47, see the announcement in 1.46:

https://github.com/grpc/grpc/releases/tag/v1.46.0

Upgrade opentelemetry-cpp to the latest gRPC C++ 11 release available, 1.46.3.

This is a required intermediary upgrade step, while awaiting for a decision on if/how to support:
-  C++ 11 gRPC 1.46.3
- C++ 14 gRPC 1.47+
concurrently in the opentelemetry-cpp code base.

marcalff:
gRPC upgraded to 1.48.1 by #1608

Closing

PR #1608, opened 2022-09-10
bump gRPC to v1.48.1 for CMake Linux CI

### LINKED ISSUE #1607: Add CI test with C++14, and latest gRPC release.
As discussed in September 7th, 2022 SIG meeting, we need a CI that tests otlp exporters using the latest gRPC release.
As gRPC no longer support std 11, gRPC has to be compiled with std >14

esigo:
Fixes #1607 (issue)

## Changes

Adds a new CI which tests otlp exporters using the latest gRPC release.


owent (approved):
LGTM

lalitb:
Thanks for the PR. A couple of nit comments 
 - Instead of adding a new CI, should we make the latest gRPC for all cmake builds on ubuntu-latest/ubuntu-20.04/ubuntu-20.10. These virtual machine uses g++-9.3 as the default compiler, which builds in c++-14 mode.
 - Can we try using the latest gRPC for bazel builds too? Not sure if it will work, as it failed earlier in #1462. We can do it in separate PR if it takes time
 - Should we document the gRPC dependency here - https://github.com/open-telemetry/opentelemetry-cpp#supported-development-platforms. We already do it for gcc4.8 for gRPC here.

ThomsonTan:
> Thanks for the PR. A couple of nit comments
> 
> * Instead of adding a new CI, should we use the latest gRPC for all cmake builds on ubuntu-latest/ubuntu-20.04/ubuntu-20.10. These virtual machine uses g++-9.3 as the default compiler, which builds in c++-14 mode.
> * Can we try using the latest gRPC for bazel builds too? Not sure if it will work, as it failed earlier in [Upgrade to gRPC 1.46.3 (#1459) #1462](https://github.com/open-telemetry/opentelemetry-cpp/pull/1462). We can do it in separate PR if it takes time
> * Should we document the gRPC dependency here - [open-telemetry/opentelemetry-cpp#supported-development-platforms](https://github.com/open-telemetry/opentelemetry-cpp#supported-development-platforms). We already do it for gcc4.8 for gRPC here.

One more item is for Windows build, could we also make the same update there?

lalitb:
> One more item is for Windows build, could we also make the same update there?

That can be challenging it seems. We use vcpkg to install windows dependencies as [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/951768a8b1c7d03b8888e33aabd3138271c69ec7/tools/setup-buildtools.cmd#L70) and latest gRPC version there is v1.41.0.

https://vcpkg.io/en/packages.html

esigo:
@lalitb Windows CMake CI doesn't use gRPC. 
We use C++14 for building gRPC itself only so I believe we don't need to change our documentation. I would keep this PR as it is.

I raised another PR to bump vcpkg #1633. I created another issue for bazel #1634.

esigo:
meanwhile gRPC 1.49.1 [released](https://github.com/grpc/grpc/releases/tag/v1.49.1).

lalitb (approved):

PR #1618, opened 2022-09-15
[ETW Exporter] - Add Trace flags in SpanContext

lalitb:
## Changes

ETW exporter doesn't preserve trace flags while creating span. The trace flags contain the `IsSampled` flag based on the sampling decision. This result in incorrect sampling logic execution. 

Also added code to copy the tracer_id from parent span to child span. 

The code is taken from the sdk implementation, and a unit test is added to validate it.


lalitb:
@ThomsonTan Can you please help review it? Thanks.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -433,12 +410,42 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+      if (spanContext.IsValid())
       {
-        parentContext = span_context;
+        parentContext = spanContext;
       }
     }
+    auto traceId = parentContext.IsValid() ? parentContext.trace_id() : traceId_;
+
+    // Sampling based on attributes is not supported for now, so passing empty below.
+    std::map<std::string, int> empty_attributes = {{}};
```
esigo, line 422: cosmetic: It seems this file has mixed camel case and snake case.
It would be good if you could unify at least the parts you are changing.

lalitb: Thanks good comment. I gave it a thought, and using snake-case now, to make it consistent with the existing ETW exporter naming style. We will eventually move this component out of the core library.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -433,12 +410,42 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+        std::unique_ptr<opentelemetry::trace::SpanContext>(new opentelemetry::trace::SpanContext(
+            traceId, GetIdGenerator(tracerProvider_).GenerateSpanId(), traceFlags, false,
+            sampling_result.trace_state
+                ? sampling_result.trace_state
+                : parentContext.IsValid() ? parentContext.trace_state()
+                                          : opentelemetry::trace::TraceState::GetDefault()));
+
+    if (sampling_result.decision == sdk::trace::Decision::DROP)
+    {
+      auto noopSpan = nostd::shared_ptr<trace::Span>{
```
ThomsonTan, line 445: Is it need to create the noopSpan with the new `spanContext`?

lalitb: yes, it is required. While noopSpan is not recorded/sampled, it would be used as a parent to decide whether the child of this noopSpan should be recorded or not. And for that to happen, the parent should have a valid span context.

ThomsonTan: So int he drop case, only the `traceFlags` matter in the `spanContext`? If so, could we create a static `spanContext` for that? Just want to see if the `spanContext` allocated could be avoided.

lalitb: We don't know what part of `spancontext` can be used in the sampling decisions, as the application can come with its own custom sampler, which can possibly use trace_id from parent context to decide what to do with child span.

lalitb: Also the sdk [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/trace/sdk.md#sdk-span-creation) states this

``
Generate a new span ID for the Span, independently of the sampling decision. This is done so other components (such as logs or exception handling) can rely on a unique span ID, even if the Span is a non-recording instance.
``

ThomsonTan: Thanks for the details which makes sense.

ThomsonTan (approved):

PR #1619, opened 2022-09-18
[gRPC]: Fix oob access of string_view in example

Tradias:
The GrpcClient/ServerCarrier in the gRPC examples assume that the provided string_views are null-terminated, 
which is not necessarily the case, e.g. when they are provided by the HttpTraceContext:

https://github.com/open-telemetry/opentelemetry-cpp/blob/82a81157a98c7120ab90e06295df0ff5ff328e56/api/include/opentelemetry/trace/propagation/http_trace_context.h#L89-L102

### DIFF
--- a/examples/grpc/tracer_common.h
+++ b/examples/grpc/tracer_common.h
@@ -38,7 +38,7 @@ public:
                    opentelemetry::nostd::string_view value) noexcept override
   {
     std::cout << " Client ::: Adding " << key << " " << value << "\n";
-    context_->AddMetadata(key.data(), value.data());
+    context_->AddMetadata(std::string(key), std::string(value));
   }
 
   ClientContext *context_;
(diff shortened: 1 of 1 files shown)


examples/grpc/tracer_common.h
```diff
@@ -38,7 +38,7 @@ class GrpcClientCarrier : public opentelemetry::context::propagation::TextMapCar
                    opentelemetry::nostd::string_view value) noexcept override
   {
     std::cout << " Client ::: Adding " << key << " " << value << "\n";
-    context_->AddMetadata(key.data(), value.data());
+    context_->AddMetadata(std::string(key), std::string(value));
```
owent, line 41: I think it's better to use `static_cast<std::string>(key), static_cast<std::string>(value)` here. Not all implementations of `string_view` can be used as constructor parameter of `std::string`, but most of them implement `operator std::string()`.

Tradias: For user-defined types the two are identical :)

https://godbolt.org/z/zfsz6baGr

owent (approved):
LGTM

lalitb (approved):

PR #1620, opened 2022-09-19
Fix more build warnings (#1616)

### LINKED ISSUE #1616: [BUILD] More build warnings
Fix build warnings reported by `-Werror=overloaded-virtual` in gcc

For example:

```
In file included from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/context/context_value.h:13,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/context/context.h:7,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/trace/tracer.h:6,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/detail/tracer_handle.h:6,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/tracer.h:10,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/factory.h:9,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/detail/dynamic_load_unix.h:12,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/dynamic_load.h:11,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/test/plugin/dynamic_load_test.cc:4:
/home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/trace/span.h:63:16: error: ‘virtual void opentelemetry::v1::trace::Span::AddEvent(opentelemetry::v1::nostd::string_view, const opentelemetry::v1::common::KeyValueIterable&)’ was hidden [-Werror=overloaded-virtual]
   63 |   virtual void AddEvent(nostd::string_view name,
      |                ^~~~~~~~
In file included from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/factory.h:9,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/detail/dynamic_load_unix.h:12,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/dynamic_load.h:11,
                 from /home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/test/plugin/dynamic_load_test.cc:4:
/home/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/include/opentelemetry/plugin/tracer.h:37:8: note:   by ‘opentelemetry::v1::plugin::Span::AddEvent’
   37 |   void AddEvent(nostd::string_view name,
      |        ^~~~~~~~
cc1plus: all warnings being treated as errors
```

marcalff:
Fixes #1616 

## Changes

Fixed the following warnings reported by gcc:

-Wall
-Werror
-Wextra-semi
-Werror=unused-parameter
-Werror=undef
-Werror=overloaded-virtual
-Werror=ignored-qualifiers

In particular, gcc gets confused by the trace::Span::AddEvent() methods,
and reports an error with -Werror=overloaded-virtual.

Likewise for logs::Logger::Log()

Fixed the following warnings reported by clang:

[-Werror,-Wimplicit-exception-spec-mismatch]

[-Werror,-Winconsistent-missing-override]

[-Werror,-Wpessimizing-move]

[-Werror,-Wimplicit-const-int-float-conversion]

[-Werror,-Wdelete-non-abstract-non-virtual-dtor]

### DIFF
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -18,7 +18,7 @@ class Token
 public:
   bool operator==(const Context &other) const noexcept { return context_ == other; }
 
-  ~Token();
+  ~Token() noexcept;
 
 private:
   friend class RuntimeContextStorage;
--- a/api/include/opentelemetry/logs/noop.h
+++ b/api/include/opentelemetry/logs/noop.h
@@ -49,6 +49,16 @@ public:
            trace::TraceFlags /* trace_flags */,
            common::SystemTimestamp /* timestamp */) noexcept override
   {}
+
+  void Log(Severity /* severity */,
+           nostd::string_view /* name */,
+           nostd::string_view /* body */,
+           const common::KeyValueIterable & /* attributes */,
+           trace::TraceId /* trace_id */,
+           trace::SpanId /* span_id */,
+           trace::TraceFlags /* trace_flags */,
+           common::SystemTimestamp /* timestamp */) noexcept override
+  {}
 };
 
 /**
--- a/api/include/opentelemetry/metrics/async_instruments.h
+++ b/api/include/opentelemetry/metrics/async_instruments.h
@@ -15,8 +15,8 @@ using ObservableCallbackPtr = void (*)(ObserverResult, void *);
 class ObservableInstrument
 {
 public:
-  ObservableInstrument() {}
-  virtual ~ObservableInstrument() {}
+  ObservableInstrument()          = default;
+  virtual ~ObservableInstrument() = default;
 
   /**
    * Sets up a function that will be called whenever a metric collection is initiated.
--- a/api/include/opentelemetry/metrics/async_instruments.h
+++ b/api/include/opentelemetry/metrics/async_instruments.h
@@ -15,6 +15,9 @@ using ObservableCallbackPtr = void (*)(ObserverResult, void *);
 class ObservableInstrument
 {
 public:
+  ObservableInstrument() {}
+  virtual ~ObservableInstrument() {}
+
   /**
    * Sets up a function that will be called whenever a metric collection is initiated.
    */
--- a/api/include/opentelemetry/metrics/noop.h
+++ b/api/include/opentelemetry/metrics/noop.h
@@ -58,7 +58,7 @@ public:
                     nostd::string_view /* description */,
                     nostd::string_view /* unit */) noexcept
   {}
-  ~NoopUpDownCounter() {}
+  ~NoopUpDownCounter() = default;
   void Add(T /* value */) noexcept override {}
   void Add(T /* value */, const opentelemetry::context::Context & /* context */) noexcept override
   {}
--- a/api/include/opentelemetry/metrics/noop.h
+++ b/api/include/opentelemetry/metrics/noop.h
@@ -58,6 +58,7 @@ public:
                     nostd::string_view /* description */,
                     nostd::string_view /* unit */) noexcept
   {}
+  ~NoopUpDownCounter() {}
   void Add(T /* value */) noexcept override {}
   void Add(T /* value */, const opentelemetry::context::Context & /* context */) noexcept override
   {}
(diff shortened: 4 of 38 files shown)


esigo:
could you please change ctor and dtors you added?
```cpp
  ctor() {}
  virtual ~dtor() {}
```

to 
```cpp
  ctor() = default;
  virtual ~dtor() = default;
```

marcalff:
> could you please change ctor and dtors you added?
> 
> ```c++
>   ctor() {}
>   virtual ~dtor() {}
> ```
> 
> to
> 
> ```c++
>   ctor() = default;
>   virtual ~dtor() = default;
> ```

Done.

esigo (approved):
LGTM
Thanks for the fix :)

lalitb:
> In particular, gcc gets confused by the trace::Span::AddEvent() methods,
and reports an error with -Werror=overloaded-virtual.

Do we know if this is some bug in gcc, as it makes us redefine all these methods in the derived class.

marcalff:
> > In particular, gcc gets confused by the trace::Span::AddEvent() methods,
> > and reports an error with -Werror=overloaded-virtual.
> 
> Do we know if this is some bug in gcc, as it makes us redefine all these methods in the derived class.

For AddEvent():
- there is a method with 2 parameters in the base class that, inlined, calls a method with 3 parameters in the base class
- there are sub classes, reimplementing the method with 3 parameters.

Testing with gcc 10, gcc reports that:
- a method with 3 parameters in a sub class
- hides a method with 2 parameters in the base class

This is surprising, because the prototype is different, so in theory there should be no virtual method hiding another.

After looking at this for a long time, I still can not decide if:
- this is a bug in gcc, which does not sees the prototype as different,
- this is an exceptionally clever interpretation in gcc, which is able to follow inline virtual methods, and finds two paths of execution that can go from AddEvent(2 param) in the base class to AddEvent(3 param) in a sub class, complaining about the ambiguity.

In any case, even if this is a gcc bug, the code should be clear enough so that correct binary is produced by the compiler. At the end of the day, what is executed in production is the binary, not the source code, and we need to avoid surprises there.

The fix implements AddEvent(2 param) in all sub classes, to follow the same pattern already used for every other AddEvent() methods, which will also help maintenance.

From a performance point of view also, I think:
- delegating from AddEvent(2 parameter) to AddEvent(3 parameters) in the base class, making a call to std::chrono::system_clock::now()
- to later land in a no-op implementation of AddEvent(3 parameters) in a sub class
is unnecessary overhead.

There is no point to call optimistically std::chrono::system_clock::now() in case it might be needed.
Instead, re implementing all methods in sub classes allows each sub class to do only the strict necessary.

lalitb:
Thanks, @marcalff for the detailed explanation, does making these hidden api methods non-virtual fix the issue? Though I agree with the performance issue you mentioned with multiple function calls and the unnecessary ts calculation, just trying way to avoid definitions of the same method at multiple places.

marcalff:
> Thanks, @marcalff for the detailed explanation, does making these hidden api methods non-virtual fix the issue? Though I agree with the performance issue you mentioned with multiple function calls and the unnecessary ts calculation, just trying way to avoid definitions of the same method at multiple places.

Hi @lalitb 

Changing a virtual method to a non virtual is a binary incompatible change I think (the virtual table is different).
It could fix the issue a different way and avoid repetition, but still at the cost of overhead when the timestamp is not needed.

The current fix looks like the best solution so far.

lalitb:
> Changing a virtual method to a non virtual is a binary incompatible change I think (the virtual table is different).

Agree.

lalitb (approved):
LGTM. Thanks for fixing these warnings.

PR #1622, opened 2022-09-20
Fixes span creation benchmark issue #1612.

### LINKED ISSUE #1612: [Benchmark] Test BM_SpanCreation() and BM_NoopSpanCreation() are misleading, sampler is ignored
In file:

`sdk/test/trace/sampler_benchmark.cc`

tests `BM_SpanCreation()` and `BM_NoopSpanCreation()` are identical.

This is because helper `BenchmarkSpanCreation()` ignores the sampler parameter passed.

Fix: use the sampler, so that the tests measure what is intended.

marcalff:
Fixes #1612 

## Changes

The sampler parameter was ignored in benchmark span creation test.

Now BM_NoopSpanCreation() uses the always off sampler, measuring the Noop creation code.


esigo (approved):
LGMT
Thanks for the fix :)

owent (approved):
LGTM

PR #1625, opened 2022-09-21
[Semantic Conventions] Align on the specification 1.13.0

### LINKED ISSUE #1624: [Semantic Conventions] Align on semantic conventions from specification 1.13
opentelemetry-specification released version 1.13

Align semantic conventions on the 1.13 spec.

marcalff:
Fixes #1624 

## Changes

Upgrade semantic conventions to opentelemetry-specification v1.13.0

BREAKING: rename `net.peer.ip` to `net.sock.peer.addr`, `net.host.ip` to `net.sock.host.addr`,
`net.peer.name` to `net.sock.peer.name` for socket-level instrumentation.

See the opentelemetry-specification v.1.13.0 release notes,
section Semantic Conventions, for full details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Increment the:
   [#1606](https://github.com/open-telemetry/opentelemetry-cpp/pull/1606)
 * [METRICS EXPORTER] Add `OtlpGrpcClient` [#1606](https://github.com/open-telemetry/opentelemetry-cpp/pull/1606)
 * [BUILD] Fix header only api singletons [#1604](https://github.com/open-telemetry/opentelemetry-cpp/pull/1604)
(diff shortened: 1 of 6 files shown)


ThomsonTan:
@lalitb could you please take a look?

I think we want to include this in the coming patch release. Should this be marked as breaking change, or probably no need as semantic convention itself is not in stable stage?

lalitb (approved):

lalitb:
> @lalitb could you please take a look?
> 
> I think we want to include this in the coming patch release. Should this be marked as breaking change, or probably no need as semantic convention itself is not in stable stage?

Yes, I think we can mark it as a breaking change, as there may be code changes required for instrumentation libraries. We don't need a major version upgrade as you rightly mentioned it as experimental.

PR #1640, opened 2022-09-29
[ETW Exporter]  - Associate trace_id with the transaction, NOT the Tracer.

lalitb:
Fixes # (issue)

## Changes

As of now, In ETW exporter, trace_id is bound to a `Tracer`. And all the spans created through that tracer get the same trace_id, even if they belong to a different transaction. This is incorrect, as trace_id should be associated with the transaction and not the Tracer.

Fixed the issue by 
 - removing traceid_ member variable from Tracer class, 
 - and generate trace_id for a span if the parent context is invalid.

TraceIdRatioBasedSampling is not working correctly because of this issue.

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -31,7 +31,7 @@ Both these dependencies are listed here:
       License`
   - Uses Abseil C++ Library for `absl::variant` as default `nostd::variant` if
     `WITH_ABSEIL` cmake option or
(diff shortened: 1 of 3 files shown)


lalitb:
closing, as the alternate solution is to create a tracer for every transaction, which ensures the unique trace_id.

PR #1641, opened 2022-09-30
Fix #1588 - Observable Gauge does not reflect updated values, and send the old value always

### LINKED ISSUE #1588: Observable Gauge does not reflect updated values, and send the old value always
**Describe your environment** 
https://github.com/open-telemetry/opentelemetry-cpp; main branch

**Steps to reproduce**
Modify https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/common/metrics_foo_library/foo_library.cc:71 to change observable counter to observable gauge. and accordingly modify the view in https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/metrics_simple/metrics_ostream.cc :58 to display Gauge value. and run the sample test

What did you expect to see?
the Gauge value updates as per the updated value sent in the callback.

What did you see instead?
The value always remain the same, "1" is the initial value, and then it changes to 2,3,4,...
2{
  scope name    : ostream_metric_example
  schema url    :
  version       : 1.2.0
  start time    : Thu Sep  1 06:54:33 2022
  end time      : Thu Sep  1 06:54:34 2022
  instrument name       : ostream_metric_example_observable_counter
  description   :
  unit          :
  type     : LastValuePointData
  timestamp     : 0
  valid     : true
  value     : **1**
  attributes            :
  resources     :
        service.name: unknown_service
        telemetry.sdk.language: cpp
        telemetry.sdk.name: opentelemetry
        telemetry.sdk.version: 1.6.0
}
3{
  scope name    : ostream_metric_example
  schema url    :
  version       : 1.2.0
  start time    : Thu Sep  1 06:54:34 2022
  end time      : Thu Sep  1 06:54:35 2022
  instrument name       : ostream_metric_example_observable_counter
  description   :
  unit          :
  type     : LastValuePointData
  timestamp     : 0
  valid     : true
  value     : **1**
  attributes            :
  resources     :
        service.name: unknown_service
        telemetry.sdk.language: cpp
        telemetry.sdk.name: opentelemetry
        telemetry.sdk.version: 1.6.0
}

**Additional context**

lalitb:
Fixes #1588 

## Changes

Fixed the logic to convert monotonic increasing async counter values to the delta. Also fixed the ostream metric example to generate monotonic increasing values for the async counter.

More details - 
While the observable instrument always records monotonically increasing values, it is possible for exporter/reader to request for either cumulative or delta of it. The AsyncMetricStorage::Record<T>() method does conversion from recorded value to delta, and further passes it to TemporalMetricStorage during collection. TemporalMetricStorage further uses it to maintain the unreported values for each collector.

There was a problem in the conversion logic of recorded value -> delta in AsyncMetricStorage::Record<T>(). This code is fixed now. And also added a test case to validate it.

### DIFF
--- a/sdk/test/metrics/async_metric_storage_test.cc
+++ b/sdk/test/metrics/async_metric_storage_test.cc
@@ -107,12 +107,12 @@ TEST_P(WritableMetricStorageTestFixture, TestAggregation)
   opentelemetry::sdk::metrics::AsyncMetricStorage storage(
       instr_desc, AggregationType::kSum, default_attributes_processor.get(),
       std::shared_ptr<opentelemetry::sdk::metrics::AggregationConfig>{});
-  long get_count                                                                  = 20l;
-  long put_count                                                                  = 10l;
-  size_t attribute_count                                                          = 2;
-  std::unordered_map<MetricAttributes, long, AttributeHashGenerator> measurements = {
-      {{{"RequestType", "GET"}}, get_count}, {{{"RequestType", "PUT"}}, put_count}};
-  storage.RecordLong(measurements,
+  long get_count1                                                                  = 20l;
+  long put_count1                                                                  = 10l;
+  size_t attribute_count                                                           = 2;
+  std::unordered_map<MetricAttributes, long, AttributeHashGenerator> measurements1 = {
+      {{{"RequestType", "GET"}}, get_count1}, {{{"RequestType", "PUT"}}, put_count1}};
+  storage.RecordLong(measurements1,
                      opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now()));
 
   storage.Collect(collector.get(), collectors, sdk_start_ts, collection_ts,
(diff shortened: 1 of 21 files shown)


sdk/test/metrics/async_metric_storage_test.cc
```diff
@@ -123,20 +123,58 @@ TEST_P(WritableMetricStorageTestFixture, TestAggregation)
...
+                       data_attr.attributes.find("RequestType")->second) == "PUT")
+          {
+            if (temporality == AggregationTemporality::kCumulative)
+            {
+              EXPECT_EQ(opentelemetry::nostd::get<long>(data.value_), put_count2);
+            }
+            else
+            {
+              EXPECT_EQ(opentelemetry::nostd::get<long>(data.value_),
+                        put_count2 - put_count1);  // 50 - 30
```
esigo, line 172: nit: redundant comment

esigo (approved):
LGTM
Thanks for the fix :)

sdk/test/metrics/async_metric_storage_test.cc
lalitb: thanks removed.

tauseef-ah:
@lalitb I was testing these changes. They work fine for ObservableCounter, but Observable Gauge still send the old value. Can you pls check for Observable Gauge as well

reyang:
> @tauseef-ah - Can you test if by setting AggergationTemporality as `kDelta` (default is `kCumulative` for ostream exporter), as it gives the desired result.

I guess aggregation temporality doesn't apply to Gauge since it is non-additive?

lalitb:
> I guess aggregation temporality doesn't apply to Gauge since it is non-additive?

lalitb:
@tauseef-ah  This has been fixed in #1651 if you want to test in the latest commit.

PR #1657, opened 2022-10-04
Add user agent for OTLP http/grpc client

### LINKED ISSUE #1656: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

owent:

Fixes #1656 

## Changes

Add user agent for OTLP http/grpc client


esigo (approved):
LGTM
Thanks for the fix :)

codeboten (commented):
Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter?

marcalff (approved):
LGTM

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -38,6 +38,8 @@ constexpr char kDefaultMetricsPath[] = "/v1/metrics";
 // The HTTP header "Content-Type"
 constexpr char kHttpJsonContentType[]   = "application/json";
 constexpr char kHttpBinaryContentType[] = "application/x-protobuf";
+// The HTTP header "User-Agent"
+constexpr char kHttpUserAgent[] = "OTel OTLP Exporter Cpp/" OPENTELEMETRY_VERSION;
```
lalitb, line 42: nit - should we instead use OPENTELEMETRY_SDK_VERSION? While both refer to same sem-ver value for now, in future if they deviate sdk version would be more relevant.

lalitb:
> Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter or does this address both http and grpc OTLP exporters?

This is for http OTLP exporter. Separate PR would be done for gRPC.

lalitb (approved):
LGTM. Thanks for adding this.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
owent: Done

lalitb:
> > Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter or does this address both http and grpc OTLP exporters?
> 
> This is for http OTLP exporter. Separate PR would be done for gRPC.

BTW, I hope gRPC-cpp allows overriding default user-agent using `ClientContext::AddMetaData()`. If not, it would be difficult to add support, as the other approach using Interceptors at client side is still experimental ( https://grpc.github.io/grpc/cpp/classgrpc_1_1experimental_1_1_interceptor.html).

owent:
> > Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter or does this address both http and grpc OTLP exporters?
> 
> This is for http OTLP exporter. Separate PR would be done for gRPC.

I'm not familiar with gRPC. Is gRPC use `User-Agent`

> > > Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter or does this address both http and grpc OTLP exporters?
> > 
> > 
> > This is for http OTLP exporter. Separate PR would be done for gRPC.
> 
> BTW, I hope gRPC-cpp allows overriding default user-agent using `ClientContext::AddMetaData()`. If not, it would be difficult to add support, as the other approach using Interceptors at client side is still experimental ( https://grpc.github.io/grpc/cpp/classgrpc_1_1experimental_1_1_interceptor.html).

gRPC user agent is added now. We could use `grpc::ChannelArguments::SetUserAgentPrefix` to prepend otel user agent before the default value(which is "grpc-c++/GRPC VERSION").

owent:
> Thanks for the quick turn around on the fix! Will there be a separate PR to add the user agent to the grpc exporter or does this address both http and grpc OTLP exporters?

Do you think should we use `OTel-OTLP-Exporter-Cpp/1.2.3` instead of `OTel OTLP Exporter Cpp/1.2.3`.
According to https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3 user agent use `token`, which is defined at https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6 and do not contains spaces.

lalitb:
> Should we use OTel-OTLP-Exporter-Cpp/1.2.3 instead of OTel OTLP Exporter Cpp/1.2.3 ?

Good point. I think we should rename, as produce-token can't contain spaces.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
```diff
@@ -21,6 +22,11 @@ namespace exporter
 namespace otlp
 {
 
+inline const std::string GetOtlpDefaultUserAgent()
+{
+  return "OTel OTLP Exporter Cpp/" OPENTELEMETRY_SDK_VERSION;
```
ThomsonTan, line 27: Avoiding including space, like "OtlpExporterCpp/1.2.3"? "OTel OTLP" seems a little redundant as "OTel" is the "O" in "OTLP".

ThomsonTan: Saw the comment https://github.com/open-telemetry/opentelemetry-specification/pull/2684/#discussion_r975520619. The current format looks good.

ThomsonTan (approved):

lalitb:
> > Should we use OTel-OTLP-Exporter-Cpp/1.2.3 instead of OTel OTLP Exporter Cpp/1.2.3 ?
> 
> Good point. I think we should rename, as produce-token can't contain spaces.

@codeboten  - What do you think of the format without spaces, to be in-compliant with the rfc?

lalitb:
As discussed in the SIG meeting, we will stick to the user-agent format as given in the [specs ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent), and change it later if required. Merging the PR based on this.

PR #1666, opened 2022-10-07
Fix #1663 Threading issue between Meter::RegisterSyncMetricStorage and Meter::Collect

### LINKED ISSUE #1663: [Metrics SDK] Threading issue between Meter::RegisterSyncMetricStorage and Meter::Collect
`Meter::Collect` is usually called in a background thread (via `PeriodicExportingMetricReader`)

`Meter::RegisterSyncMetricStorage` and `Meter::RegisterAsyncMetricStorage` are called for each new instrument.

These methods are not thread safe, and can cause SEGFAULT under certain conditions.

lalitb:
Fixes #1663

## Changes

Add a lock guard to prevent race conditions arising due to 
 - Concurrent creation of sync/async instruments. 
 - Creation of sync/async instruments while Metric collection is ongoing.
 - Concurrent collection of metrics for same meters.

These operations should ensure that sync/sync storage structures and callback registry are atomically updated.
Also added unit-test to stress meter using multiple threads.


lalitb:
This is ready for review now.

sdk/test/metrics/meter_test.cc
```diff
@@ -70,6 +71,55 @@ TEST(MeterTest, BasicAsyncTests)
...
+  std::vector<nostd::shared_ptr<opentelemetry::metrics::ObservableInstrument>>
+      observable_instruments;
+  while (numIterations--)
+  {
+    for (size_t i = 0; i < MAX_THREADS; i++)
+    {
+      if (threadCount++ < MAX_THREADS)
+      {
+        auto t = std::thread([&]() {
+          std::this_thread::yield();
```
esigo, line 96: I'm not sure why we need this.

sdk/test/metrics/meter_test.cc
```diff
@@ -70,6 +71,55 @@ TEST(MeterTest, BasicAsyncTests)
...
+            observable_instruments.push_back(std::move(observable_instrument));
+            do_collect.store(true);
+          }
+          if (do_collect.exchange(false))
+          {
+            metric_reader_ptr->Collect([](ResourceMetrics &metric_data) { return true; });
+            do_sync_create.store(true);
+          }
+        });
+        t.detach();
```
esigo, line 117: I might be picky here. I understand that we don't care what happens to the threads here. However, can we make sure at least some of the threads finished?

esigo (commented):
Thanks for the fix.
It would be great if you could make sure at least some of the threads finish in the unit test.
For example each thread can increment a counter when they start and decrement it when they finish.

sdk/test/metrics/meter_test.cc
```diff
@@ -70,6 +71,55 @@ TEST(MeterTest, BasicAsyncTests)
...
+          if (do_sync_create.exchange(false))
+          {
+            std::string instrument_name = "test_couter_" + std::to_string(numIterations);
+            meter->CreateLongCounter(instrument_name, "", "");
+            do_async_create.store(true);
+          }
+          if (do_async_create.exchange(false))
+          {
+            auto observable_instrument =
+                meter->CreateLongObservableGauge("test_gauge" + std::to_string(numIterations));
```
esigo, line 106: nit: to be consistent?
```suggestion
                meter->CreateLongObservableGauge("test_gauge_" + std::to_string(numIterations));
```

sdk/test/metrics/meter_test.cc
lalitb: Good point. I think there is no harm in letting the threads join in the end before exiting the test. I have modified the code accordingly,  and also removed the sleep as an alternate wait.

sdk/test/metrics/meter_test.cc
lalitb: Thanks fixed.

sdk/test/metrics/meter_test.cc
lalitb: In general, a good idea to context switch for the waiting threads to execute before invoking operations in the new thread.

sdk/src/metrics/meter.cc
```diff
@@ -293,6 +295,7 @@ std::unique_ptr<AsyncWritableMetricStorage> Meter::RegisterAsyncMetricStorage(
 std::vector<MetricData> Meter::Collect(CollectorHandle *collector,
                                        opentelemetry::common::SystemTimestamp collect_ts) noexcept
 {
+  std::lock_guard<opentelemetry::common::SpinLockMutex> guard(storage_lock_);
```
ThomsonTan, line 298: Move this to below context lock? So the error handling and call to `Observe` is not necessary to be guarded by this lock?

lalitb: Yes, good point.  Have moved it below the context lock.

esigo (approved):

ThomsonTan (approved):

PR #1675, opened 2022-10-12
Fix:1674, Add Monotonic Property to Sum Aggregation, and unit tests for Up Down Counter

### LINKED ISSUE #1674: [Metrics SDK] - Add monotonicity property to Sum Aggregation
**Describe your environment** 

As Sum Aggregation is default mapping for both Counter, and UpDownCounter, it is important to maintain the monotonic property for this aggregation to differentiate the type of value stored. This is required to
 - validate the measurement value being recorded.
 - Select the correct metric family at the exporter. Eg Prometheus exporter should use `gauge` for UpDownCounter, and `counter` for Counter.

lalitb:
Fixes #1674 

## Changes

- Add monotonic properly for Sum Aggregation
- Select correct metric family for UpDownCounter in prometheus exporter
- Add unit test for UpDownCounter.
- 


sdk/src/metrics/aggregation/sum_aggregation.cc
```diff
@@ -26,6 +28,12 @@ LongSumAggregation::LongSumAggregation(const SumPointData &data) : point_data_{d
...
 void LongSumAggregation::Aggregate(long value, const PointAttributes & /* attributes */) noexcept
 {
+  if (point_data_.is_monotonic_ && value < 0)
+  {
+    OTEL_INTERNAL_LOG_WARN(" Negative value ignored for Monotonic increasing measurement. Value"
+                           << value);
+    return;
+  }
   const std::lock_guard<opentelemetry::common::SpinLockMutex> locked(lock_);
   point_data_.value_ = nostd::get<long>(point_data_.value_) + value;
```
ThomsonTan, line 38: Does overflow need to be handled here for the sum operation?

lalitb: With #1686 we use int64_t to store the sum. The number limit of signed 64-bit is enormous, while it can still eventually overflow if the system generating measurement keeps running for long, without resetting. Unless there are real-world examples for the overflow scenario, I think we can let go for now the overflow scenario ( and so avoid extra validation for every measurement added). Even in the case of such real-world examples, the exporter can be configured with Delta temporality (if the backend system supports it), and so it would bypass adding the values. Otherwise, we can rely on the backend system to detect any such overflow and handle it accordingly.

lalitb:
Have fixed the CI errors. This should be good to review now.

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
```diff
@@ -0,0 +1,327 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <memory>
```
esigo, line 4: this could go below `ENABLE_METRICS_PREVIEW`

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
```diff
@@ -0,0 +1,327 @@
...
+#ifndef ENABLE_METRICS_PREVIEW
```
esigo, line 5: not a blocker, bazel is missing for this test.

esigo (approved):
LGTM
just had a few minor comments
Thanks for the PR :)

sdk/src/metrics/aggregation/sum_aggregation.cc
```diff
@@ -26,6 +28,12 @@ LongSumAggregation::LongSumAggregation(const SumPointData &data) : point_data_{d
 
 void LongSumAggregation::Aggregate(long value, const PointAttributes & /* attributes */) noexcept
 {
+  if (point_data_.is_monotonic_ && value < 0)
+  {
+    OTEL_INTERNAL_LOG_WARN(" Negative value ignored for Monotonic increasing measurement. Value"
+                           << value);
```
marcalff, line 34: Ok for now.

In the long term, to be of any use, the warning message will need to add some context, like the name of the meter / metric involved that causes the warning.

marcalff (approved):
LGTM

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -159,12 +173,16 @@ metric_sdk::AggregationType PrometheusExporterUtils::getAggregationType(
...
 prometheus_client::MetricType PrometheusExporterUtils::TranslateType(
-    metric_sdk::AggregationType kind)
+    metric_sdk::AggregationType kind,
+    bool is_monotonic)
 {
   switch (kind)
   {
     case metric_sdk::AggregationType::kSum:
-      return prometheus_client::MetricType::Counter;
+      if (!is_monotonic)
```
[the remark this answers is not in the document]
lalitb: Thanks, added/

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
lalitb: Added

sdk/src/metrics/aggregation/sum_aggregation.cc
lalitb: Thanks. added slightly more context here.

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
```diff
@@ -0,0 +1,327 @@
...
+#  include "opentelemetry/sdk/metrics/instruments.h"
+#  include "opentelemetry/sdk/metrics/state/sync_metric_storage.h"
+#  include "opentelemetry/sdk/metrics/view/attributes_processor.h"
+
+#  include <gtest/gtest.h>
+#  include <map>
+
+using namespace opentelemetry::sdk::metrics;
+using namespace opentelemetry::common;
+using M         = std::map<std::string, std::string>;
```
ThomsonTan, line 18: Is using `M` as namespace alias from some convention?

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
```diff
@@ -0,0 +1,327 @@
...
+  std::map<std::string, std::string> attributes_get = {{"RequestType", "GET"}};
+  std::map<std::string, std::string> attributes_put = {{"RequestType", "PUT"}};
+
+  std::unique_ptr<DefaultAttributesProcessor> default_attributes_processor{
+      new DefaultAttributesProcessor{}};
+  opentelemetry::sdk::metrics::SyncMetricStorage storage(
+      instr_desc, AggregationType::kSum, default_attributes_processor.get(),
+      ExemplarReservoir::GetNoExemplarReservoir(), nullptr);
+
+  storage.RecordLong(10l, KeyValueIterableView<std::map<std::string, std::string>>(attributes_get),
```
ThomsonTan, line 58: define the number `10l` as a variable as it is shared between this and the following line?

ThomsonTan (approved):

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
lalitb: No, I don't think there is any convention. What do you recommend ?

sdk/test/metrics/sync_metric_storage_up_down_counter_test.cc
lalitb: Thanks, done

PR #1673, opened 2022-10-11
Add timeout support to MeterContext::ForceFlush

ThomsonTan:
Fixes #1672 1672

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1650, opened 2022-10-03
Add CMake OTELCPP_MAINTAINER_MODE

marcalff:
Fixes #1648 

### CMake OTELCPP_MAINTAINER_MODE

Add a new CMake option, OTELCPP_MAINTAINER_MODE.

Option OTELCPP_MAINTAINER_MODE=ON is supported for:

- GCC
- CLANG
- MSVC

and causes the build to fail on warnings (-Wall -Werror).

This option is useful for maintainers, when testing a build locally.

### New CI builds

Add new builds in Github [CI]

- GCC (maintainer mode)
- CLANG (maintainer mode)
- MSVC (maintainer mode)

These build use OTELCPP_MAINTAINER_MODE=ON,
to enforce the build is warnings free for all compilers.

* [X] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [X] Changes in public API reviewed (no change)

PR #1682, opened 2022-10-14
Fix:1676 Segfault when short export period is used for metrics 

### LINKED ISSUE #1676: [Metrics SDK] Segfault when short export period is used for metrics
We're experiencing segfaults when using async instruments with a relatively short (e.g. `OTEL_METRICS_EXPORT_INTERVAL_MILLIS=100 OTEL_METRICS_EXPORT_TIMEOUT_MILLIS=50`) export period.

For me, this looks like a race condition issue and strongly resembles https://github.com/open-telemetry/opentelemetry-cpp/issues/1663 , but it's still present after bumping `opentelemetry-cpp` to the last commit.

I was able to catch this with GDB, here's the stack trace:
[(gdb)_backtrace_full.txt](https://github.com/open-telemetry/opentelemetry-cpp/files/9766970/gdb._backtrace_full.txt)

If necessary, I'll try to put together a self-contained reproducer, but I'm hoping that it's something simple. :)

lalitb:
Fixes #1676 

## Changes

MeterContext::ForceFlush() is incorrectly setting shutdown latch. Later when MeterContext::Shutdown() is invoked, the latch is already set, and shutdown is not done.

The PR would have merge conflict with #1673.

@ahadnagy  - If you would like to test the changes, I got the segfault with with your code during MeterProvider shutdown, and this fixes that. I am hoping it was the same crash you observed :)


esigo (approved):
LGTM
Thanks for the fix :)
let's wait for #1650 to merge first.

ahadnagy:
@lalitb I had a chance to try this today, and looks like it's solved the issue! Thanks a lot for fixing this so quickly!

ahadnagy (approved):
@lalitb I had a chance to try this today, and looks like it's solved the issue! Thanks a lot for fixing it so quickly!

lalitb:
> @lalitb I had a chance to try this today, and looks like it's solved the issue! Thanks a lot for fixing it so quickly!

Thanks for confirming. Also calling MeterProvider->Shutdown() is now optional, as it would be called as part of MeterProvider destruction. Have added that change in this PR.

PR #1710, opened 2022-10-25
[Metrics SDK] Move Metrics Exemplar processing behind feature flag

lalitb:
Exemplar are still experimental, so exemplar processing is moved behind feature flag macro - `ENABLE_METRICS_EXEMPLAR_PREVIEW`. The exemplar would be processed in hot path only if the flag is enabled. The flag needs to be explicitly enabled in build script to use exemplars.

This allows us 
 - to give a message to users that this functionality could change in future.
 - and achieve (slight) performance gain during measurement recording, even while using the non-sampling exemplar.

The CI tests still continue to test the full metrics functionality with feature flag enabled for both cmake and bazel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,6 +192,7 @@ option(WITH_EXAMPLES "Whether to build examples" ON)
 option(WITH_METRICS_PREVIEW "Whether to build metrics preview" OFF)
 option(WITH_LOGS_PREVIEW "Whether to build logs preview" OFF)
 option(WITH_ASYNC_EXPORT_PREVIEW "Whether enable async export" OFF)
+# Exemplar specs status is experimental, so behind feature flag by default
 option(WITH_METRICS_EXEMPLAR_PREVIEW
        "Whethere to enable exemplar within metrics" OFF)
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -192,6 +192,8 @@ option(WITH_EXAMPLES "Whether to build examples" ON)
 option(WITH_METRICS_PREVIEW "Whether to build metrics preview" OFF)
 option(WITH_LOGS_PREVIEW "Whether to build logs preview" OFF)
 option(WITH_ASYNC_EXPORT_PREVIEW "Whether enable async export" OFF)
+option(WITH_METRICS_EXEMPLAR_PREVIEW
+       "Whethere to enable exemplar within metrics" OFF)
```
marcalff, line 196: Typo: Whether

marcalff (approved):
Approved, see minor typo.

esigo (approved):
LGTM
Thanks for the PR :)

CMakeLists.txt
ThomsonTan: And mentioned the spec status is experimental, so user will be aware of that instead of considering it as implementation gap?

ThomsonTan (approved):

PR #1713, opened 2022-10-26
Fix: 1712 -  Validate Instrument meta data (name, unit, description)

### LINKED ISSUE #1712: [Metrics SDK] - Validate Instrument meta data (name, unit, description)
Metrics Specs specifies the Instrument naming rules - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule

Valid Instruments should be created only if the `name`, `unit` and `description` are compliant with these naming rules.

Also, adding it after GA would be a breaking change specifically for the non-compliant instruments created before enabling this validation. Once validation is enabled, it won't be possible to send the metrics for these non-compliant instruments.
So scoping this for GA.

lalitb:
Fixes #1712

## Changes

- Create valid instrument if the meta data is validated
- Else, create noop-instrument.


sdk/src/metrics/meter.cc
```diff
@@ -39,6 +40,13 @@ nostd::unique_ptr<metrics::Counter<uint64_t>> Meter::CreateUInt64Counter(
     nostd::string_view description,
     nostd::string_view unit) noexcept
 {
+  if (!ValidateInstrument(name, description, unit))
+  {
+    OTEL_INTERNAL_LOG_ERROR("Meter::CreateUInt64Counter - failed. Invalid parameters."
+                            << name << " " << description << " " << unit);
+    return nostd::unique_ptr<metrics::Counter<uint64_t>>(
+        new metrics::NoopCounter<uint64_t>(name, description, unit));
```
marcalff, line 48: The calling instrumented application will not notice the error if returning a noop counter.

Not crashing now with a noop is not helping.
Instead, bugs will be filed later against opentelemetry-cpp about instrument that produce no data.

How about returning a nullptr, forcing the calling application to do something about it ?

lalitb: Yes, good point. I too thought about it, but also felt that making some critical application crash for some issue in telemetry configuration is not a good idea.  We can also be more explicit in error message stating that measurements won't be recorded.

This is also inline with the general error handling [guidelines ](https://github.com/jack-berg/opentelemetry-specification/blob/6307e44b88dd8f654134946c765cd22dc01c8455/specification/error-handling.md#error-handling-and-performance) - 
``
Whenever API call returns values that is expected to be non-null value - in case of error in processing logic - SDK MUST return a "no-op" or any other "default" object that was (ideally) pre-allocated and readily available. This way API call sites will not crash on attempts to access methods and properties of a null objects.
``

api/include/opentelemetry/common/macros.h
```diff
@@ -90,6 +90,14 @@
 #  define OPENTELEMETRY_DEPRECATED_MESSAGE(msg)
 #endif
 
+// Regex support
+#if (__GNUC__ == 4 && (__GNUC_MINOR__ == 8 || __GNUC_MINOR__ == 9))
```
ThomsonTan, line 94: nit: __GNUC_MINOR__ >= 8?

sdk/src/metrics/instrument_metadata_validator.cc
```diff
@@ -0,0 +1,78 @@
...
+const std::string kInstrumentNamePattern = "[a-zA-Z][-_.a-zA-Z0-9]{0,62}";
+//
+const std::string kInstrumentUnitPattern = "[\x01-\x7F]{0,63}";
+// instrument-unit = It can have a maximum length of 63 ASCII chars
+
+InstrumentMetaDataValidator::InstrumentMetaDataValidator()
+#  if HAVE_WORKING_REGEX
+    : name_reg_key_{kInstrumentNamePattern}, unit_reg_key_
+{
+  kInstrumentUnitPattern
```
ThomsonTan, line 28: Was this line formatted by clang-format? The format looks a little weird.

sdk/src/metrics/instrument_metadata_validator.cc
```diff
@@ -0,0 +1,78 @@
...
+}
+#  endif
+{}
```
ThomsonTan, line 31: Should this be put under `#else` block?

api/include/opentelemetry/common/macros.h
lalitb: The regex support is partially implemented for gcc 4.8 and gcc 4.9 only. So the existing condition looks good to me.

sdk/src/metrics/instrument_metadata_validator.cc
lalitb: yeah clang-format behaved weirdly here. I have disable the clang-format for this block of code. Looks better now.
Thanks for noticing.

sdk/src/metrics/instrument_metadata_validator.cc
lalitb: no it's fine actually. The part of code was somewhat confusing with clang-format. It should be more clear now.

sdk/src/metrics/instrument_metadata_validator.cc
```diff
@@ -0,0 +1,78 @@
...
+{
+#  if HAVE_WORKING_REGEX
+  return std::regex_match(name.data(), name_reg_key_);
+#  else
+  // size atmost 63 chars
+  if (name.size() > kMaxSize)
+  {
+    return false;
+  }
+  // first char should be alphanumeric
```
ThomsonTan, line 43: First char should be alpha, not alphanumeric?

api/include/opentelemetry/common/macros.h
lalitb: But yeah, even  GNUC_MINOR >= 8 should work fine. But let it be explicit I think :)

sdk/src/metrics/instrument_metadata_validator.cc
lalitb: yeah good point. fixed the comment.

sdk/src/metrics/meter.cc
marcalff: Thanks for the details. Ok with noop then.

sdk/src/metrics/meter.cc
```diff
@@ -52,6 +61,14 @@ nostd::unique_ptr<metrics::Counter<double>> Meter::CreateDoubleCounter(
     nostd::string_view description,
     nostd::string_view unit) noexcept
 {
+  if (!ValidateInstrument(name, description, unit))
+  {
+    OTEL_INTERNAL_LOG_ERROR("Meter::CreateDoubleCounter - failed. Invalid parameters."
```
ThomsonTan, line 66: nit: probably use macro `__func__` to avoid hard code function name in string liternal.

sdk/test/metrics/instrument_metadata_validator_test.cc
```diff
@@ -0,0 +1,75 @@
...
+      "d1234_-sDSDs.sdsd344",                 // combination of all valid characters
+      CreateVeryLargeString(5) + "ABCERTYG",  // total 63 charactes
+  };
+  for (auto const &str : valid_names)
+  {
+    EXPECT_TRUE(validator.ValidateName(str));
+  }
+}
+
+TEST(InstrumentMetadataValidator, TestUnit)
```
ThomsonTan, line 49: Missing unit test on description?

ThomsonTan (approved):

sdk/test/metrics/instrument_metadata_validator_test.cc
lalitb: We don't have any validation checks for the description, so no tests added for it.

sdk/src/metrics/meter.cc
lalitb: yes, probably we can modify `OTEL_INTERNAL_*` macros to automatically add class and method name to the logs.

PR #1725, opened 2022-10-31
Set `CMAKE_EXPORT_COMPILE_COMMANDS` in `CMakeLists.txt`

JoeLoser:
It is useful to generate a `compile_commands.json` to allow for downstream tooling like `clangd` and friends to work well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,10 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.5")
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+endif()
+
 # Don't use customized cmake modules if vcpkg is used to resolve dependence.
 if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules/")


CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
lalitb, line 18: Should this be enabled only for CMake v3.5 or greater?

JoeLoser: Valid point — just added a version check since the CMake variable wasn't introduced until 3.5.

JoeLoser:
@lalitb FYI I'm still waiting for employer approval regarding the CLA. This was more along the lines of something useful for me when navigating the codebase, so figured I'd upstream it.

CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.5")
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
owent, line 19: Do we need add this by default?I think IDEs should add this option automatically when them need it.(When we use vscode and set C/C++ configurationProvider to `ms-vscode.cmake-tools`, this option will be set by `cmake-tools` plugin.)

And most time when building a package or run in CI, we do not need it.

JoeLoser: I'm not opposed to putting this behind a CMake flag, but turning it "ON" by default seems the right thing to do from a developer perspective. For example, consider vim or emacs users (personally, I use `neovim`) — in those cases, I'm not aware of any launcher script or anything that sets this in the user profile side of things to propagate to the CMake configure process.

I agree that this flag isn't useful in CI on its own currently; it's mainly targeting local developers. Is there any noticeable "slowdowns" on CI to make this something we should hide behind a flag?

owent: I didn't test the cost on otel-cpp yet, but I think it's not a huge project so it may  not cost too much. What's your think about it @lalitb ?

esigo: I wonder why you don't just use `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=1`, when configuring?

JoeLoser: I'm OK setting `-DCMAKE_EXPORT_COMPILE_COMMANDS` when configuring `CMake`.  While this is generally useful to more developers, it doesn't seem there is much interest in having it be set at the top-level, so I'll keep it to my own configuration then. Going to close this PR as such — thanks for the discussion.

PR #1730, opened 2022-11-02
[Build] Bring your own dependency: opentelemetry-proto

### LINKED ISSUE #1719: [BUILD] Bring your own dependency: opentelemetry-proto
Assume a CMake superproject, which uses opentelemetry-cpp in a sub directory.

The superproject builds is done in a controlled environment, such that downloading code from elsewhere is not allowed.

Any code used during the superproject build, which includes all code used by dependencies such as opentelemetry-cpp, like opentelemetry-proto, should be provided up front.

When opentelemetry-cpp builds, CMake currently attempts to:

- either use git submodules to find opentelemetry-proto
- or download opentelemetry-proto from github directly.

Please provide a way, in the opentelemetry-cpp CMakeLists.txt, to provide an explicit dependency for opentelemetry-proto.

For example, in the superproject:
`OTELCPP_PROTO_PATH=/path/to/opentelemetry-proto`
would cause opentelemetry-cpp to build with the code provided.

marcalff:
Fixes #1719

## Changes

Add option OTELCPP_PROTO_PATH in CMake.


owent (approved):
LGTM.
Is it better to set `OTELCPP_PROTO_PATH` when using internal proto by `ExternalProject_Add`?

cmake/opentelemetry-proto.cmake
```diff
@@ -1,27 +1,62 @@
...
+# 3 - Download opentelemetry-proto from github
+#
+# Code from the required version is used,
+# unless a specific release tag is provided
+# in variable ${opentelemetry-proto}
+#
+
+if(OTELCPP_PROTO_PATH)
+  message(STATUS "opentelemetry-proto dependency satisfied by: external path")
+  set(PROTO_PATH ${OTELCPP_PROTO_PATH})
```
lalitb, line 30: should this also validate the presence of common.proto , trace.proto and logs.proto, and error-out if missig?

lalitb (approved):
LGTM. Thanks.

cmake/opentelemetry-proto.cmake
marcalff: Good point, looking into this.

marcalff:
> LGTM. Is it better to set `OTELCPP_PROTO_PATH` when using internal proto by `ExternalProject_Add`?

Thanks @owent 

I don't understand the suggestion itself.

Do you mean to set OTELCPP_PROTO_PATH as well after:

```
    ExternalProject_Add(
      opentelemetry-proto
      GIT_REPOSITORY https://github.com/open-telemetry/opentelemetry-proto.git
      GIT_TAG "${opentelemetry-proto}"
      UPDATE_COMMAND ""
      BUILD_COMMAND ""
      INSTALL_COMMAND ""
      CONFIGURE_COMMAND ""
      TEST_AFTER_INSTALL 0
      DOWNLOAD_NO_PROGRESS 1
      LOG_CONFIGURE 1
      LOG_BUILD 1
      LOG_INSTALL 1)
    ExternalProject_Get_Property(opentelemetry-proto INSTALL_DIR)
    set(PROTO_PATH "${INSTALL_DIR}/src/opentelemetry-proto")
```

If so, this would mean the makefile alter choices made by the user, making it unclear if this variable is an input to the makefile, or an output from the makefile (when using submodules and/or a github download), of both, causing confusion in my opinion.

Please clarify.

cmake/opentelemetry-proto.cmake
marcalff: Fixed:

Checked common.proto to detect, at configuration time (cmake), an invalid path, instead of having weird errors during the build.

Testing every proto file is probably overkill. If someone really modifies the content of opentelemetry-proto, things will break at compile time anyway, the same way it will break with a broken tag in git submodules, or a broken tag in github pull.

lalitb:
> > LGTM. Is it better to set `OTELCPP_PROTO_PATH` when using internal proto by `ExternalProject_Add`?
> 
> Thanks @owent
> 
> I don't understand the suggestion itself.
> 
> Do you mean to set OTELCPP_PROTO_PATH as well after:
> 
> ```
>     ExternalProject_Add(
>       opentelemetry-proto
>       GIT_REPOSITORY https://github.com/open-telemetry/opentelemetry-proto.git
>       GIT_TAG "${opentelemetry-proto}"
>       UPDATE_COMMAND ""
>       BUILD_COMMAND ""
>       INSTALL_COMMAND ""
>       CONFIGURE_COMMAND ""
>       TEST_AFTER_INSTALL 0
>       DOWNLOAD_NO_PROGRESS 1
>       LOG_CONFIGURE 1
>       LOG_BUILD 1
>       LOG_INSTALL 1)
>     ExternalProject_Get_Property(opentelemetry-proto INSTALL_DIR)
>     set(PROTO_PATH "${INSTALL_DIR}/src/opentelemetry-proto")
> ```
> 
> If so, this would mean the makefile alter choices made by the user, making it unclear if this variable is an input to the makefile, or an output from the makefile (when using submodules and/or a github download), of both, causing confusion in my opinion.
> 
> Please clarify.

@owent Do you have any further comments, or can this be merged. thanks.

PR #1731, opened 2022-11-02
Remove misleading comments about ABI compatibility for nostd::span

### LINKED ISSUE #1705: libstdc++ span not ABI compatible with nostd::span since GCC 11
I was debugging some segfaults that occurred when I was building the OTel libraries with `WITH_STL=ON`, statically linking them to an application. I wasn't using an explicit `CMAKE_CXX_STANDARD` for OTel which caused it to build with C++20, while the application is built with C++17. This caused a segfault in the `InMemorySpanExporter` as that was using a different implementation of `nostd::span` because its included through a header-only file.

The solution was to just explicitly set the same `CMAKE_CXX_STANDARD` for the OTel build so the same `nostd::span` implementation is selected, but I did notice the following.

There is a comment here hinting that it is desirable to have ABI compatibility with `libstdc++`'s implementation of `std::span`:
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/nostd/span.h#L243-L247

However, that implementation has since GCC 11 swapped around the `extent` and `data` fields:
https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.3.0/libstdc++-v3/include/std/span#L392-L393

This was causing my segfaults.

Of course my problem was fixed by properly setting the `CMAKE_CXX_STANDARD` for OTel the same as my application, but I just wanted to mention that here in case this has any implications I am not aware of.

marcalff:
Fixes #1705 

## Changes

Remove misleading comments about ABI compatibility for nostd::span

### DIFF
--- a/api/include/opentelemetry/nostd/span.h
+++ b/api/include/opentelemetry/nostd/span.h
@@ -240,9 +240,6 @@ public:
   T *end() const noexcept { return data_ + extent_; }
 
 private:
-  // Note: matches libstdc++'s layout for std::span
-  // See
-  // https://github.com/gcc-mirror/gcc/blob/a60701e05b3878000ff9fdde1aecbc472b9dec5a/libstdc%2B%2B-v3/include/std/span#L402-L403
   size_t extent_;
   T *data_;
 };


lalitb (approved):

api/include/opentelemetry/nostd/span.h
```diff
@@ -240,9 +240,6 @@ class span<T, dynamic_extent>
   T *end() const noexcept { return data_ + extent_; }
 
 private:
-  // Note: matches libstdc++'s layout for std::span
```
ThomsonTan, line 243: But it seems the definition here does provide the ABI compatibility as libc++, do you mean this is unnecessary?

marcalff: It was compatible with one version of libc++ at one point, but the library implementation changed other time.

We can't guarantee ABI compatibility with a moving target, so we should not claim or hint is is compatible.

owent (approved):

PR #1761, opened 2022-11-10
[SEMANTIC CONVENTIONS] Upgrade to version 1.15.0

### LINKED ISSUE #1758: [SEMANTIC CONVENTIONS] Upgrade to version 1.15.0
opentelemetry-specification 1.15.0 is now released.

It contains changes in semantic conventions.

Run code generation again to align on spec 1.15.0.

marcalff:
Fixes #1758

## Changes

Upgrade semantic conventions to opentelemetry-specification 1.15.0
Change naming to use camel case, to avoid collisions with macros.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ Breaking changes:
     to camel case, like `SemanticConventions::kServiceName`.
     This is necessary to avoid collisions with macros in general,
     which breaks the build on some platforms.
-  * Semantic conventions are still flagged as unstable,
+  * Semantic conventions are flagged as experimental,
     which is why this change is done in this release.
 
 ## [1.7.0] 2022-10-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Increment the:
 * [METRICS SDK] - Remove old metrics from Github CI
   [#1733](https://github.com/open-telemetry/opentelemetry-cpp/pull/1733)
 * [BUILD] Add CMake OTELCPP_PROTO_PATH [#1730](https://github.com/open-telemetry/opentelemetry-cpp/pull/1730)
+* [SEMANTIC CONVENTIONS] Upgrade to version 1.15.0
+  [#1761](https://github.com/open-telemetry/opentelemetry-cpp/pull/1761)
 
 Deprecation notes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,18 @@ Deprecation notes:
     Note that function `opentelemetry::utils::hashCode`,
     declared in the ETW exporter, is not affected by this deprecation.
 
+Breaking changes:
+
+* [SEMANTIC CONVENTIONS] Upgrade to version 1.15.0
+  [#1761](https://github.com/open-telemetry/opentelemetry-cpp/pull/1761)
+  * Naming of semantic conventions has changed from uppercase constants,
+    like `SemanticConventions::SERVICE_NAME`,
+    to camel case, like `SemanticConventions::kServiceName`.
+    This is necessary to avoid collisions with macros in general,
+    which breaks the build on some platforms.
+  * Semantic conventions are still flagged as unstable,
+    which is why this change is done in this release.
+
 ## [1.7.0] 2022-10-28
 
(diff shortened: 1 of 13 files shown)


lalitb (approved):
Thanks for the upgrade.

marcalff:
@lalitb 

Please note: I had to implement a work around for ERROR in windows, similar to DELETE:

```
#ifdef _WIN32
#  ifdef ERROR
#    pragma message(__FILE__ ": removing define on ERROR")
#    undef ERROR
#  endif
#endif
```

Build on windows fails without it.

ThomsonTan:
> @lalitb
> 
> Please note: I had to implement a work around for ERROR in windows, similar to DELETE:
> 
> ```
> #ifdef _WIN32
> #  ifdef ERROR
> #    pragma message(__FILE__ ": removing define on ERROR")
> #    undef ERROR
> #  endif
> #endif
> ```
> 
> Build on windows fails without it.

Wondering why this is needed now.

lalitb:
> > @lalitb
> > Please note: I had to implement a work around for ERROR in windows, similar to DELETE:
> > ```
> > #ifdef _WIN32
> > #  ifdef ERROR
> > #    pragma message(__FILE__ ": removing define on ERROR")
> > #    undef ERROR
> > #  endif
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Build on windows fails without it.
> 
> Wondering why this is needed now.

I think this line is having conflict with ERROR macro

`static constexpr const char *ERROR = "ERROR";`

ThomsonTan:
> > > @lalitb
> > > Please note: I had to implement a work around for ERROR in windows, similar to DELETE:
> > > ```
> > > #ifdef _WIN32
> > > #  ifdef ERROR
> > > #    pragma message(__FILE__ ": removing define on ERROR")
> > > #    undef ERROR
> > > #  endif
> > > #endif
> > > ```
> > > Build on windows fails without it.
> > 
> > 
> > Wondering why this is needed now.
> 
> I think this line is having conflict with ERROR macro
> 
> `static constexpr const char *ERROR = "ERROR";`

Can we rename the new variable here?

lalitb:
> Can we rename the new variable here?

Yeah good point, if it is possible to do in the jinja template.

marcalff:
Some comments about ERROR:

- I don't know what the ERROR macro is on Windows, and what it is for. So, no idea if `undef ERROR` is intrusive or not for the application code. 
- It is confirmed, from the build logs, that there is an ERROR macro somewhere, see bazel windows.
- Could someone point out which header file on windows is defining an ERROR macro, will be needed in the comments.

If we rename it:

- Any proposal about the name ?
- Should we have some consistency about alternate names, like `SEMCONV_<foo>` to replace `<foo>` on name collisions ? It is likely that more collisions will happen again, with new semantic conventions.
- Should the alternate name, say `SEMCONV_ERROR`, be used only on the offending platform only or on all platforms ?
- Should we change `undef DELETE` to `SEMCONV_DELETE` then ?

My preference at this point is:
- change `DELETE` to `SEMCONV_DELETE`
- change `ERROR` to `SEMCONV_ERROR`
- apply for all platform, so user code that builds on multiple platforms uses the same symbol in all cases.

Requesting re-review to settle this.

lalitb:
@marcalff Should we instead change the variable name to follow the C++ style guide (https://google.github.io/styleguide/cppguide.html#Constant_Names) i.e, `k<camelCase>`. I see the [to_camel_case](https://github.com/open-telemetry/build-tools/blob/48180eafbe34ae7df3ba1fee3526e510f4892f6b/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L152) method defined in templating code, not sure if we can use it and prefix `k` before it.

Also semantic conventions specs are still experimental, so we can use that to our benefit :)

marcalff:
> @marcalff Should we instead change the variable name to follow the C++ style guide (https://google.github.io/styleguide/cppguide.html#Constant_Names) i.e, `k<camelCase>`. I see the [to_camel_case](https://github.com/open-telemetry/build-tools/blob/48180eafbe34ae7df3ba1fee3526e510f4892f6b/semantic-conventions/src/opentelemetry/semconv/templating/code.py#L152) method defined in templating code, not sure if we can use it and prefix `k` before it.
> 
> Also semantic conventions specs are still experimental, so we can use that to our benefit :)

Thanks @lalitb, this is a much better solution.

Implemented.

lalitb (commented):
LGTM. Thanks :)

Issue #1688, opened 2022-10-17
[Logs API] - Logger Changes


lalitb:
Specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#logger-operations

- The API specs defines two interfaces to emit logs and events.
- As per the specs, the interface to emit the logs is ONLY for the Appenders, and not the end-user/instrumentation library. The specs intentionally leaves the general-purpose logging API for the instrumentation library as unspecified. Which means, we may continue using the existing APIs as it is, and add two new interfaces for emitting events and logs.
- event_name would be copied to Logs attributes before exporting to the destination, ( as part of SDK implementation)

Something like   -

```cpp

  virtual nostd::unique_ptr<LogRecord> Logger::CreateLogRecord() = 0;

  virtual void Logger::EmitEvent(nostd::string event_name, nostd::unique_ptr<LogRecord> log_record) = 0;
 
  virtual void Logger::EmitLog(nostd::unique_ptr<LogRecord> log_record) = 0;

```

- In above code design, `LogRecord` is thought as analogous to `Span`, and the setters are defined through the Recordable interface implementation by exporter.

- The existing Logger::Log() method can be renamed as Logger::EmitLog() to be consistent with the newly introduced Logger::EmitLog(log_record) method. This will ensure there won't be much code changes required by instrumentation library.

Issue #1687, opened 2022-10-17
[Logs API] LoggerProvider changes


lalitb:
Changes required in LoggerProvider::GetLogger()` method as per the logs API specs:
 - `Logger name` is not the part of logs data model, and can be removed.
 - `event_domain` (optional, required if you want to emit events) - These should be added to instrumentation scope attributes through SDK ( with key as `event_domain`)
 - `include_trace_context` (optional, true by default)
 - `attributes` - These should be added to instrumentation scope attributes through SDK. 

The GetLogger() method would look something like - 

```cpp
  virtual nostd::shared_ptr<Logger> LoggerProvider::GetLogger(nostd::string_view library_name,
                                              nostd::string_view library_version = "",
                                              nostd::string_view schema_url      = "",
                                              nostd::string_view event_domain,
                                              bool include_trace_context = true,
                                               const common::KeyValueIterable &attributes) = 0;
```

owent:
Please don't stale this issue, I will continue finish left works when I have time.
Sorry, I only work on this in spare time and can not finish all changes in a short period of time.

lalitb:
Thanks for informing @owent. Have un-stalled all these issues. In general, feel free to remove the `stale` tag for any such issues  - this is justified as you are the approver for this repo :)

PR #1766, opened 2022-11-13
New `LogRecord` and `Recordable` implementations.

### LINKED ISSUE #1689: [Logs API] - New LogRecord abstract class.
Specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#logrecord

In order to avoid storing any data as part of API implementation, `LogRecord` can be pure virtual class with setters which would be implemented by SDK. These setters would directly serialize the data using `Recordable` implementation in format expected by exporter.

Something like -
```cpp
 class LogRecord
 {
    public:
        virtual void SetAttribute(nostd::string_view key,
                            const common::AttributeValue &value) noexcept = 0;
        virtual void SetTimeStamp(common::SystemTimestamp timestamp) noexcept = 0;
         ----
   

 };
```

### LINKED ISSUE #1691: [Logs SDK] Rename and update existing LogRecord class, modify existing Recordable interface
- To avoid conflict with the new LogRecord class, rename existing sdk::logs::LogRecord class to some better name (say sdk::logs::LogRecordData).
- Add setter for observed_timestamp in sdk:logs::Reordable and sdk::logs::LogRecordData.
- Similar to Span's Recordable, add a getter method for LogRecordData in Recordable class. This can be eventually implemented by all it's derived implementation (not in scope for now).
```cpp
  /**
   * Get the LogRecordData object for this Recordable.
   *
   * @return LogRecordData*
   */

  virtual explicit operator LogRecordData *() const { return nullptr; }

```

This would let us avoid having `ReadWriteLogRecord` class. This method can be potentially used inside `LogRecordProcessor` to filter the logs based on it's contents (not in scope for now).

owent:
Signed-off-by: WenTao Ou <admin@owent.net>

Fixes #1691 
Fixes #1689 

## Changes

**There are a lot break changes in this PR.But it's easy to migrate if users do not implement their own Logger and Exporter.**

+ Removed deprecated `Logger::Log` APIs.
  + All APIs with the old `name` field are removed now.The `name` field is ognored from v1.4.0.
+ New `LogRecord`, `ReadableLogRecord`, `ReadWriteLogRecord`
  + https://github.com/open-telemetry/opentelemetry-specification/blob/v1.15.0/specification/logs/api.md#logrecord
  + https://github.com/open-telemetry/opentelemetry-specification/blob/v1.15.0/specification/logs/sdk.md#additional-logrecord-interfaces
+ New `logs::Recordable` to collect data from the all in one `LogRecord`.
  + #1691 
+ Add a simple `EmitEvent` function which will be use to implement #1687 and #1688 in the future.
  + https://github.com/open-telemetry/opentelemetry-specification/blob/v1.15.0/specification/logs/api.md#emit-event
  + It's not finished yet, and will be continued in another PR later.
+ Replace `std::unique_ptr<T>` with `nostd::unique_ptr<T>` in logs.


sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
```diff
@@ -0,0 +1,190 @@
...
+   * Set instrumentation_scope for this log.
+   * @param instrumentation_scope the instrumentation scope to set
+   */
+  void SetInstrumentationScope(const opentelemetry::sdk::instrumentationscope::InstrumentationScope
+                                   &instrumentation_scope) noexcept;
+
+private:
+  // Default values are set by the respective data structures' constructors for all fields,
+  // except the severity field, which must be set manually (an enum with no default value).
+  opentelemetry::logs::Severity severity_                 = opentelemetry::logs::Severity::kInvalid;
```
lalitb, line 168: So, these LogRecord fields (severity, body, attributes) would be copied/stored in the SDK before exporting them, unlike earlier when they were directly serialized in the exporter's format ?

owent: Yes, according to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#logrecord , [LogRecords](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-logrecord) MUST be able to set these following fields.
So I think we can only store them into ReadWriteLogRecord, and we can just store `string_view` and integers here to reduce copy.

lalitb: Theonly advantage I see with having ReadableLogRecord, and ReadWriteLogRecord is that we can now easily define the getter methods for logRecord as required by specs. But I am not seeing the use-case of these getter methods. For traces too, we deviated from specs by not have read/write spans and directly uses recordable to serialize the data as a tradeoff for performance.

owent: `ReadableLogRecord` and `ReadWriteLogRecord` is defined in SDK, while `LogRecord` is defined in API. They are just like `Span` in API and SDK, and `logs::Recordable` is just like the `trace::Recordable`.
I believe which cost more is store data into `proto::v1::LogRecord` in OTLP and convert data into json in ETW and ES exporters.To reduce the cost, I try to only keep string_view and AttributeValue in LogRecord classes, these are all just a reference to the real data without lifetime(There is a except, attributes are store in a `unordered_map<std::string, AttributeValue>` now, maybe we can use a reference to `KeyValueIterable` instead of using `unordered_map`, but it may make it difficult to set attribute one by one and add `event.name` when emit a record with event), The getter methods are used to pick datas from LogRecord and set them into `Recordable` now.

owent: I think it easier to add or remove fields if log content changes in the future with `LogRecord` in API.

lalitb: Storing non-owning objects can have undesirable issues if these objects get out of scope before they get serialized by the exporter. I am still not seeing the benefit of storing them for implementing getter methods unless there is a use case where these getters would be used as public API.

owent: The exporter should create it's own `Recordable` object from `ReadableLogRecord` or `ReadWriteLogRecord` and the `Recordable` of exporter should own objects, just like the old implementation.The mainly different between old and new implementation is the old implementation set data fields one by one and the new implementation collect all data into `ReadWriteLogRecord` the lifetime and data copy times will not change, just like we talked before in #1691 .
`ReadableLogRecord` is not necessary but I think it's better to follow [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.15.0/specification/logs/sdk.md#readablelogrecord). Do you mean we can delete `ReadWriteLogRecord` and just keep `ReadWriteLogRecord` ?

lalitb:
> All APIs with the old name field are removed now.

Do you think we can keep the existing API intact (maybe renaming to `Logger::EmitLog(..., ..., ...)`)  as mentioned in #1688. As per the logging API specs, the newly introduced interface to emit the log-records is only for Appenders. The specs doesn't enforce it for instrumentation library which makes sense because most of the languages (except C++) have logging api as part of language specs. The existing API can continue using direct `Recordable` interface to send the logs to exporter, without creating LogRecord object. Sometime, instrumentation library author would prefer to write a single line of code to emit logs.

And then the new interface can build the `Recordable` through `ReadWriteLogRecord` as you are doing right now in this PR.

owent:
> intact



> > All APIs with the old name field are removed now.
> 
> Do you think we can keep the existing API intact (maybe renaming to `Logger::EmitLog(..., ..., ...)`) as mentioned in #1688. As per the logging API specs, the newly introduced interface to emit the log-records is only for Appenders. The specs doesn't enforce it for instrumentation library which makes sense because most of the languages (except C++) have logging api as part of language specs. The existing API can continue using direct `Recordable` interface to send the logs to exporter, without creating LogRecord object. Sometime, instrumentation library author would prefer to write a single line of code to emit logs.
> 
> And then the new interface can build the `Recordable` through `ReadWriteLogRecord` as you are doing right now in this PR.

Sorry, I'm poor in English and perhaps I didn't make it clear. The `name` field means All `Logger::Log` with `name` parameter, this argument is ignored, for example `OPENTELEMETRY_DEPRECATED_MESSAGE("name will be removed in the future") void Log(Severity severity, nostd::string_view name, nostd::string_view message) noexcept`. We have discussed at #1383 before, and I think it's long enough to leave time for users to migrate now.

For users, we can still use the old `Logger::Log` APIs to emit logs, if we allow processor or exporters to use `Recordable` interface to send the logs without `LogRecord`, we can make `Recordable` derive from `LogRecord`. But if we changes or add more fields into `LogRecord` , we have to modify all `Recordable` implementations. Do you think it will be more difficult to maintain and upgrade for users?

sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
lalitb: > does it means these getters would be able to be used as public API?

Yes, these getters should be available as public API.

lalitb:
> > intact
> 
> > > All APIs with the old name field are removed now.
> > 
> > 
> > Do you think we can keep the existing API intact (maybe renaming to `Logger::EmitLog(..., ..., ...)`) as mentioned in #1688. As per the logging API specs, the newly introduced interface to emit the log-records is only for Appenders. The specs doesn't enforce it for instrumentation library which makes sense because most of the languages (except C++) have logging api as part of language specs. The existing API can continue using direct `Recordable` interface to send the logs to exporter, without creating LogRecord object. Sometime, instrumentation library author would prefer to write a single line of code to emit logs.
> > And then the new interface can build the `Recordable` through `ReadWriteLogRecord` as you are doing right now in this PR.
> 
> Sorry, I'm poor in English and perhaps I didn't make it clear. The `name` field means All `Logger::Log` with `name` parameter, this argument is ignored, for example `OPENTELEMETRY_DEPRECATED_MESSAGE("name will be removed in the future") void Log(Severity severity, nostd::string_view name, nostd::string_view message) noexcept`. We have discussed at #1383 before, and I think it's long enough to leave time for users to migrate now.
> 
> For users, we can still use the old `Logger::Log` APIs to emit logs, if we allow processor or exporters to use `Recordable` interface to send the logs without `LogRecord`, we can make `Recordable` derive from `LogRecord`. But if we changes or add more fields into `LogRecord` , we have to modify all `Recordable` implementations. Do you think it will be more difficult to maintain and upgrade for users?

Thanks for explanation. So to summarise 
 - We can remove all the deprecated `Logger::Log()` methods ( i.e. those containing `name` ), as we have given enough time for users to migrate.
 -  Rest of the existing methods can be (probably ?) renamed to `Logger::EmitLog()`, and they will keep using `Recordable` interface (which is derived from `LogRecord` class).

This looks fine to me. Log Data model is already stable, I don't think LogRecord would change any further, definitely not after log API is also made stable. So, there shouldn't be upgrade issues with users.

owent:
> > > intact
> > 
> > 
> > > > All APIs with the old name field are removed now.
> > > 
> > > 
> > > Do you think we can keep the existing API intact (maybe renaming to `Logger::EmitLog(..., ..., ...)`) as mentioned in #1688. As per the logging API specs, the newly introduced interface to emit the log-records is only for Appenders. The specs doesn't enforce it for instrumentation library which makes sense because most of the languages (except C++) have logging api as part of language specs. The existing API can continue using direct `Recordable` interface to send the logs to exporter, without creating LogRecord object. Sometime, instrumentation library author would prefer to write a single line of code to emit logs.
> > > And then the new interface can build the `Recordable` through `ReadWriteLogRecord` as you are doing right now in this PR.
> > 
> > 
> > Sorry, I'm poor in English and perhaps I didn't make it clear. The `name` field means All `Logger::Log` with `name` parameter, this argument is ignored, for example `OPENTELEMETRY_DEPRECATED_MESSAGE("name will be removed in the future") void Log(Severity severity, nostd::string_view name, nostd::string_view message) noexcept`. We have discussed at #1383 before, and I think it's long enough to leave time for users to migrate now.
> > For users, we can still use the old `Logger::Log` APIs to emit logs, if we allow processor or exporters to use `Recordable` interface to send the logs without `LogRecord`, we can make `Recordable` derive from `LogRecord`. But if we changes or add more fields into `LogRecord` , we have to modify all `Recordable` implementations. Do you think it will be more difficult to maintain and upgrade for users?
> 
> Thanks for explanation. So to summarise
> 
> * We can remove all the deprecated `Logger::Log()` methods ( i.e. those containing `name` ), as we have given enough time for users to migrate.
> * Rest of the existing methods can be (probably ?) renamed to `Logger::EmitLog()`, and they will keep using `Recordable` interface (which is derived from `LogRecord` class).
> 
> This looks fine to me. Log Data model is already stable, I don't think LogRecord would change any further, definitely not after log API is also made stable. So, there shouldn't be upgrade issues with users.

Thanks, I will push another commit later to change the relationship between `Recordable` and `LogRecord`.

I also have a idea about the new `Logger::EmitLog()` APIs, how about using variable template when implement `Logger::EmitLog()`, and call specify `LogRecord::SetXXX` by type of argument?When we get `trace::TraceId` , call `LogRecord::SetTraceId()`,call `LogRecord::SetAttribute()` when we get any type that can be converted to `common::KeyValueIterable` and so on. So that we don't need to implement so many overload `Logger::EmitLog()` functions.

lalitb:
> I also have a idea about the new Logger::EmitLog() APIs, how about using variable template when implement Logger::EmitLog(), and call specify LogRecord::SetXXX by type of argument?When we get trace::TraceId , call LogRecord::SetTraceId(),call LogRecord::SetAttribute() when we get any type that can be converted to common::KeyValueIterable and so on. So that we don't need to implement so many overload Logger::EmitLog() functions.

owent:
> > I also have a idea about the new Logger::EmitLog() APIs, how about using variable template when implement Logger::EmitLog(), and call specify LogRecord::SetXXX by type of argument?When we get trace::TraceId , call LogRecord::SetTraceId(),call LogRecord::SetAttribute() when we get any type that can be converted to common::KeyValueIterable and so on. So that we don't need to implement so many overload Logger::EmitLog() functions.
> 
> Aren't variable templates part of C++14 ?

No, it's C++11. See https://en.cppreference.com/w/cpp/compiler_support/11

lalitb:
> > > I also have a idea about the new Logger::EmitLog() APIs, how about using variable template when implement Logger::EmitLog(), and call specify LogRecord::SetXXX by type of argument?When we get trace::TraceId , call LogRecord::SetTraceId(),call LogRecord::SetAttribute() when we get any type that can be converted to common::KeyValueIterable and so on. So that we don't need to implement so many overload Logger::EmitLog() functions.
> > 
> > 
> > Aren't variable templates part of C++14 ?
> 
> No, it's C++11. See https://en.cppreference.com/w/cpp/compiler_support/11

Oh variadic templates :) Not sure if that would look good at public API surface. I would prefer overloaded methods, but would leave it to you.

owent:
@lalitb Could you please take some time to review this PR again when you have time?The `Recordable` is derived from `LogRecord` now.
I think I can rename the other `Logger::Log` into `Logger::EmitLog` in another PR and maybe we can also keep the old `Log` APIs for serveral versions to let users to migrate.

api/include/opentelemetry/logs/logger.h
```diff
@@ -38,8 +41,78 @@ class Logger
...
+   */
+  virtual nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept = 0;
+
+  /**
+   * Emit a Log Record object with event
+   *
+   * @param event_name event.name
+   * @param log_record
+   */
+  virtual void EmitEvent(nostd::string_view event_name,
```
lalitb, line 56: Not for immediate attention, but just to update that Log/Event API specs has changed again :( - https://github.com/open-telemetry/opentelemetry-specification/pull/2941/

We can accommodate these changes separately in another PR. Will create a task for that.

api/include/opentelemetry/logs/logger.h
```diff
@@ -51,45 +124,75 @@ class Logger
...
   virtual void Log(Severity severity,
                    nostd::string_view body,
                    const common::KeyValueIterable &attributes,
                    trace::TraceId trace_id,
                    trace::SpanId span_id,
                    trace::TraceFlags trace_flags,
-                   common::SystemTimestamp timestamp) noexcept = 0;
+                   common::SystemTimestamp timestamp) noexcept
+  {
+    nostd::unique_ptr<LogRecord> log_record = CreateLogRecord();
```
lalitb, line 135: Please confirm if the existing methods wouldn't use LogRecord, but directly use Recordable interface ?

api/include/opentelemetry/logs/logger.h
owent: OK, I removed `EmitEvent` and will add it back into `EventLogger` later when implementing #1687 . We can just modify #1687 to follow the changes.

api/include/opentelemetry/logs/logger.h
owent: `Recordable` derive from `LogRecord`. It's a break change in API here, the existing old `UserLogger::Log` must be replaced by `UserLogger::CreateLogRecord` and `UserLogger::EmitLogRecord`, or it won't compile.(Just like the changes in `etw::Logger`).
And the existing exporters or processes use the old `Recordable` still work after add a new override `Recordable::SetObservedTimestamp` function.

lalitb: Sorry for the delay. Let me go through all the changes in next couple of days, and would complete the review by then.

owent: Thanks, please let me know if there is any misunderstanding about specification.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
```diff
@@ -41,7 +41,7 @@ class BatchLogRecordProcessor : public LogRecordProcessor
    * equal to max_queue_size
    */
   explicit BatchLogRecordProcessor(
-      std::unique_ptr<LogRecordExporter> &&exporter,
+      nostd::unique_ptr<LogRecordExporter> &&exporter,
```
lalitb, line 44: Why do we need to switch to `nostd::unique_ptr` ?

sdk/include/opentelemetry/sdk/logs/logger.h
```diff
@@ -30,46 +31,20 @@ class Logger final : public opentelemetry::logs::Logger
   explicit Logger(
       opentelemetry::nostd::string_view name,
       std::shared_ptr<LoggerContext> context,
-      std::unique_ptr<instrumentationscope::InstrumentationScope> instrumentation_scope =
+      nostd::unique_ptr<instrumentationscope::InstrumentationScope> instrumentation_scope =
```
lalitb, line 34: Why do we need to switch to `nostd::unique_ptr` here, and in general in most of other places in SDK ?

api/include/opentelemetry/logs/logger.h
lalitb: No, I think the changes are much aligned to the specs, and look better. I would have some minor comments only.

sdk/include/opentelemetry/sdk/logs/logger.h
owent: Isn't it has better ABI compatibility? I think we should better use `nostd::unique_ptr` everywhere.

lalitb: We don't claim ABI compatibility at SDK level. And so have been using `std::` classes throughout SDK ( in both traces and metrics). Only API interfaces implemented by SDK use `nostd::` classes.

owent: Thanks for the explaination, I restore `nostd::unique_ptr` to `std::unique_ptr` in sdk and exporters now.

sdk/src/logs/logger.cc
```diff
@@ -30,88 +30,31 @@ const nostd::string_view Logger::GetName() noexcept
...
-  else if (span_context.trace_flags().IsSampled())
-  {
-    recordable->SetTraceFlags(span_context.trace_flags());
-  }
+  // TODO: Sampler (should include check for minSeverity)
 
-  // Send the log record to the processor
-  processor.OnEmit(std::move(recordable));
+  // Send the log recordable to the processor
+  processor.OnEmit(std::unique_ptr<Recordable>(static_cast<Recordable *>(log_record.release())));
```
lalitb, line 57: Oh I didn't realize this would need conversion from nostd::unique_ptr to std::unique_ptr.

lalitb (commented):
Have some minor comments. In general looks in good shape. Need to review otlp exporter part to finish the review, will be doing that tomorrow.

api/include/opentelemetry/logs/logger.h
```diff
@@ -51,45 +86,75 @@ class Logger
...
-   * in order to create a log record.
-   */
   virtual void Log(Severity severity,
                    nostd::string_view body,
                    const common::KeyValueIterable &attributes,
                    trace::TraceId trace_id,
                    trace::SpanId span_id,
                    trace::TraceFlags trace_flags,
-                   common::SystemTimestamp timestamp) noexcept = 0;
+                   common::SystemTimestamp timestamp) noexcept
```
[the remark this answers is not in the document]
owent: Should `observerd_timestamp` always be set to the time when log record is created?

api/include/opentelemetry/logs/logger.h
```diff
@@ -51,45 +86,75 @@ class Logger
...
-                   OPENTELEMETRY_MAYBE_UNUSED nostd::string_view name,
                    nostd::string_view body,
                    const common::KeyValueIterable &attributes,
-                   trace::TraceId trace_id,
-                   trace::SpanId span_id,
-                   trace::TraceFlags trace_flags,
                    common::SystemTimestamp timestamp) noexcept
   {
-    Log(severity, body, attributes, trace_id, span_id, trace_flags, timestamp);
+    nostd::unique_ptr<LogRecord> log_record = CreateLogRecord();
```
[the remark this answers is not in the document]
owent: According to specification, log should be emitted by `Logger::EmitLogRecord`, not `Logger::Log`.

sdk/src/logs/logger.cc
owent: Because API create `nostd::unique_ptr` but sdk and exporters use `std::unique_ptr`, as we talked before at https://github.com/open-telemetry/opentelemetry-cpp/pull/1766#discussion_r1037982249_

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -6,9 +6,12 @@
 
 #  include <memory>
 #  include <vector>
+
 #  include "opentelemetry/nostd/span.h"
+#  include "opentelemetry/nostd/unique_ptr.h"
 #  include "opentelemetry/sdk/common/exporter_utils.h"
 #  include "opentelemetry/sdk/logs/processor.h"
+#  include "opentelemetry/sdk/logs/readable_log_record.h"
```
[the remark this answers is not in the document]
owent: Thanks and removed.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -25,10 +28,8 @@ class LogRecordExporter
   virtual ~LogRecordExporter() = default;
 
   /**
-   * Create a log recordable. This object will be used to record log data and
-   * will subsequently be passed to LogRecordExporter::Export. Vendors can implement
-   * custom recordables or use the default LogRecord recordable provided by the
-   * SDK.
+   * Create a log recordable from readable log record object. This object will be used to record log
```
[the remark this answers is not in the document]
owent: Thanks and sorry I forget to update these comments.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -23,8 +26,8 @@ class LogRecordProcessor
   virtual ~LogRecordProcessor() = default;
 
   /**
-   * Create a log recordable. This requests a new log recordable from the
-   * associated exporter.
+   * Create a log recordable from readable log record object
```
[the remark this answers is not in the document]
owent: Thanks and sorry I forget to update these comments.

sdk/src/logs/logger.cc
```diff
@@ -3,7 +3,7 @@
 
 #ifdef ENABLE_LOGS_PREVIEW
 #  include "opentelemetry/sdk/logs/logger.h"
-#  include "opentelemetry/sdk/logs/log_record.h"
+#  include "opentelemetry/sdk/logs/read_write_log_record.h"
```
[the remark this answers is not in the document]
owent: Thanks and removed.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
lalitb: Yes, I think we can set it internally as current timestamp instead of making it part of Logger::Log() API.

exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
```diff
@@ -100,53 +185,38 @@ class Logger : public opentelemetry::logs::Logger
...
-  void Log(opentelemetry::logs::Severity severity,
-           nostd::string_view body,
-           const common::KeyValueIterable &attributes,
-           opentelemetry::trace::TraceId trace_id,
-           opentelemetry::trace::SpanId span_id,
-           opentelemetry::trace::TraceFlags trace_flags,
-           common::SystemTimestamp timestamp) noexcept override
+  nostd::unique_ptr<opentelemetry::logs::LogRecord> CreateLogRecord() noexcept
   {
+    return nostd::unique_ptr<opentelemetry::logs::LogRecord>(new LogRecord());
```
lalitb, line 190: This will create a intermediate storage for LogRecord object even when we invoke Logger:Log(). We can keep it as of now, I will try to see it later if we can optimize it.

sdk/src/logs/readable_log_record.cc
```diff
@@ -0,0 +1,51 @@
...
+    return opentelemetry::logs::SeverityNumToText[0];
+  }
+
+  return opentelemetry::logs::SeverityNumToText[severity_index];
+}
+
+const opentelemetry::sdk::instrumentationscope::InstrumentationScope &
+ReadableLogRecord::GetDefaultInstrumentationScope() noexcept
+{
+  // FIXME: Use shared default instrumentation scope?
```
lalitb, line 31: nit - is this comment valid, as we seems to be using shared static instrumentation scope below.

sdk/src/logs/read_write_log_record.cc
```diff
@@ -0,0 +1,177 @@
...
+
+const std::unordered_map<std::string, opentelemetry::common::AttributeValue>
+    &ReadWriteLogRecord::GetAttributes() const noexcept
+{
+  return attributes_map_;
+}
+
+const opentelemetry::sdk::resource::Resource &ReadWriteLogRecord::GetResource() const noexcept
+{
+  if (nullptr != resource_)
```
lalitb, line 145: nit - use OPENTELEMETRY_LIKELY_IF ?

lalitb (approved):
LGTM. with few minor comments. 
We can wait for couple of days if there are more reviews before merging.

Thanks for the work, it is nicely done :)

sdk/src/logs/readable_log_record.cc
owent: The shared default instrumentation scope I mean is a shared instrumentation scope between all of the components of SDK, includes trace and metrics, not just for logs.

sdk/src/logs/read_write_log_record.cc
owent: Yes and thanks, I will change it tomorrow.

PR #1780, opened 2022-11-16
[ETW Exporter] Tail based sampling support

lalitb:
## Changes

Adding limited support for tail based sampling. Similar to head-sampling, TailSampler::ShouldSample() method needs to be overridden with sampling logic, which get's executed before the span is exporter.

```cpp
class TailSampler
{
public:
  // convert to etw span if required for getters on span.
  // auto etw_span = static_cast<const opentelemetry::exporter::etw::Span*>(&span);
  // Decision based on
  //      Span::GetStatus()
  //      Span::GetProperties()
  //      Span::GetContext()
  virtual opentelemetry::sdk::trace::SamplingResult ShouldSample(
      const opentelemetry::trace::Span &span) noexcept = 0;
};
```

The inherited TailSampler class object needs to be passed as argument to TracerProvider constructor. The default is `AlwaysOnTailSampler`

Also, added use of macro `OPENTELEMETRY_NOT_USE_RTTI` to disable using RTTI in ETW exporter even if it is enabled in compiler. This is because the ETW exporter modified the attributes passed as argument if RTTI is enabled to reuse them for creating ETW event structure, which is sometime not desirable.


exporters/etw/test/etw_tracer_test.cc
```diff
@@ -459,6 +468,26 @@ TEST(ETWTracer, AlwayOffSampler)
...
+  std::unique_ptr<TailSampler> always_off_tail{new AlwaysOffTailSampler()};
+  exporter::etw::TracerProvider tp
+    ({
+      {"enableTraceId", true},
+      {"enableSpanId", true},
+      {"enableActivityId", true},
+      {"enableRelatedActivityId", true},
+      {"enableAutoParent", true}
+     },
+     std::move(always_on),
```
ThomsonTan, line 485: Could we start a span and test it is indeed dropped?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -235,9 +236,17 @@ class Tracer : public opentelemetry::trace::Tracer,
...
                        const opentelemetry::trace::EndSpanOptions & = {})
   {
-    const auto &cfg = GetConfiguration(tracerProvider_);
+    const auto &cfg          = GetConfiguration(tracerProvider_);
+    const auto &tail_sampler = GetTailSampler(tracerProvider_);
     const opentelemetry::trace::Span &spanBase =
         reinterpret_cast<const opentelemetry::trace::Span &>(span);
+
+    //  Sample span based on the decision of tail based sampler
+    auto sampling_result = const_cast<TailSampler *>(&tail_sampler)->ShouldSample(spanBase);
```
ThomsonTan, line 245: Instead of making a virtual call to AlwaysOn sampler by default, perhaps check if there is a non-default sampler set, then make the virtual call? In this way, it doesn't introduce extra cost (except the null check) when tail based sampler is not used.

lalitb: Good comment. Actually we do have virtual method calls for head-sampler, id-generator both in ETW exporter, and the otel-sdk. There is no reported performance reported with them. Seemingly modern compilers can optimize these repeated calls by caching the address, and avoid virtual indirection. Perhaps we can look back to this if there are reported overhead internally and/or by community.

exporters/etw/test/etw_tracer_test.cc
lalitb: Yes I was thinking to mock the actual ETW transport layer and do the test. In current design, this is not easy to do, perhaps can be taken separately if it is fine?

ThomsonTan (approved):

PR #1793, opened 2022-11-23
[EXPORTER] Add OTLP HTTP SSL support

### LINKED ISSUE #389: Enable http(s) ssl verification for curl based http_client implementation
Currently, the curl based http client implementation (https://github.com/open-telemetry/opentelemetry-cpp/pull/385) doesn't support ssl( client and server) verification. This needs to be implemented.

### LINKED ISSUE #1402: [Trace SDK] OTLP Trace Exporter secure connection configuration options
Greetings,

In the spec, for the OpenTelemetry Protocol Exporter configurations options:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specifying-headers-via-environment-variables

The spec says:

> The following configuration options MUST be available to configure the OTLP exporter. Each configuration option MUST be overridable by a signal specific option.

(...)

(1) Certificate File:

Env vars: `OTEL_EXPORTER_OTLP_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE` ...

(2) Client key file:

Env vars: `OTEL_EXPORTER_OTLP_CLIENT_KEY` `OTEL_EXPORTER_OTLP_TRACES_CLIENT_KEY` ...

(3) Client certificate file:

Env vars: `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE` `OTEL_EXPORTER_OTLP_TRACES_CLIENT_CERTIFICATE` ...

Currently, opentelemetry-cpp supports the following environment variables:
-  `OTEL_EXPORTER_OTLP_CERTIFICATE`
- `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE`
which covers (1), per the spec.
- `OTEL_EXPORTER_OTLP_CERTIFICATE_STRING`
- `OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE_STRING`
which covers (1), in addition to the spec (great).

Implementation for (2) and (3) appear to be missing.

Also, secure connection options are only used for the `OtlpGrpcExporter`

Should the `OtlpHttpExporter` honor the same variables as well ?

Currently there are no security options in `OtlpHttpExporterOptions`.

My understanding from the spec is that secure options apply to the OTLP exporter in general: that is, to both OTLP/gRPC and OTLP/HTTP end points.

Regards.

### LINKED ISSUE #1756: TLS settings for OtlpHttpExporter
I am checking how to enable TLS encryption for connection between `OtlpHttpExporter` and Otel Collector. I found that exporter should supports https endpoint address, but it does not allow to configure TLS options. As a minimum I have to specify custom root CA certificate file. I also would like to enforce TLS 1.2+ and selected ciphers. Please add such configuration options for exporter.

I found that grpc exporter allows to specify path to CA certificates using `ssl_credentials_cacert_path` option or `OTEL_EXPORTER_OTLP_CERTIFICATE` env var. However it also does now allow to specify minimum TLS version and ciphers list: https://github.com/open-telemetry/opentelemetry-cpp/tree/main/exporters/otlp

I suspect that this may require some changes to OTEL specification too.

marcalff:
Fixes #389 Enable http(s) ssl verification for curl based http_client implementation
Fixes #1402  [Trace SDK] OTLP Trace Exporter secure connection configuration options
Fixes #1756 TLS settings for OtlpHttpExporter

## Changes in Exporters environment variables

Added all the helpers required to parse experimental TLS environment variables:

  * GetOtlpDefaultTracesSslTlsMinVersion();
  * GetOtlpDefaultMetricsSslTlsMinVersion();
  * GetOtlpDefaultLogsSslTlsMinVersion();
  * GetOtlpDefaultTracesSslTlsMaxVersion();
  * GetOtlpDefaultMetricsSslTlsMaxVersion();
  * GetOtlpDefaultLogsSslTlsMaxVersion();

// For TLS 1.0, 1.1, 1.2
  * GetOtlpDefaultTracesSslTlsCipher();
  * GetOtlpDefaultMetricsSslTlsCipher();
  * GetOtlpDefaultLogsSslTlsCipher();

// For TLS 1.3
  * GetOtlpDefaultTracesSslTlsCipherSuite();
  * GetOtlpDefaultMetricsSslTlsCipherSuite();
  * GetOtlpDefaultLogsSslTlsCipherSuite();

## Changes in Exporters options

* In OtlpHttpExporterOptions, added the following members:
  * ssl_insecure_skip_verify
  * ssl_ca_cert_path
  * ssl_ca_cert_string
  * ssl_client_key_path
  * ssl_client_key_string
  * ssl_client_cert_path
  * ssl_client_cert_string

* Added a feature flag `ENABLE_OTLP_HTTP_SSL`, because this is an SDK ABI change.
  * There is an existing spec for SSL and mTLS, only the implementation is missing 

* In OtlpHttpExporterOptions, added the following members:
  * ssl_min_tls
  * ssl_max_tls
  * ssl_cipher
  * ssl_cipher_suite

* Added a sub feature flag `ENABLE_OTLP_HTTP_SSL_TLS`, because this is an SDK ABI change.
  * There are no spec for TLS version environment variables, **and there will be no spec**, because there is a global freeze on new configuration variables.
  * Using per language (OTEL_CPP_) environment variables.

* Likewise for metrics in OtlpHttpMetricExporterOptions

* Likewise for logs in OtlpHttpLogRecordExporterOptions

## Changes in Http client

* Fixed every call to CURL, to check for errors
* Improved error reporting for CURL, to include detailed error messages
* Implemented added CURL calls to support SSL and TLS

## Functional tests

* Implemented a test client to cover various connections configurations
* Tested against no server
* Tested against an opentelemetry-collector in http mode (without SSL)
* Tested against an opentelemetry-collector in https mode (with SSL)
* Integrated tests in CI

## Changes in documentation

* Added step by step instructions to generate certificate and keys, to use when testing with OLTP HTTP SSL.

* [X] Documentation added
* [X] Functional tests have been added


owent:
There are several compatibility problems when using options of libcurl.
[CURLOPT_ISSUERCERT_BLOB](https://curl.se/libcurl/c/CURLOPT_ISSUERCERT_BLOB.html) is available from 7.71.0
[CURLOPT_SSLKEY_BLOB](https://curl.se/libcurl/c/CURLOPT_SSLKEY_BLOB.html) is available from 7.71.0
[CURLOPT_SSLCERT_BLOB](https://curl.se/libcurl/c/CURLOPT_SSLCERT_BLOB.html) is available from 7.71.0

And also, there are some SSL options for ALPN, Custom CA, password for certification files, TLS cipher/TLS 1.3 cipher and certification files just for proxy. Some of them also just work on a high version.

Can we adapt these options by `LIBCURL_VERSION_NUM`? So that we can still use HTTP Client on a old system with the curl in package manager?(For example: Cent OS 7/curl 7.29.0)

lalitb:
Just a thought, for CURL version 7.70.0 or lesser, if CA certificate for validation is provided as string, we should enforce the SSL handshake to fails, probably by overriding the default system path to read the certificates from:

(pseducode)
```cpp
#if LIBCURL_VERSION_NUM >=0x071f00  and defined OTEL_VERIFY_HTTPS_CERTIFICATE
if (input_ssl_ca_cert_string.size()) {
       curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L);
       curl_easy_setopt(curl, CURLOPT_CAINFO, "<inaccessible_file>");
       curl_easy_setopy(curl, CURLOPT_CANPATH, "<inaccessible_path>");
}
#endif 
```

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -410,9 +412,99 @@ CURLcode HttpOperation::Setup()
...
+      return CURLE_UNKNOWN_OPTION;
+#endif
+    }
+
+    /* 4 - ENFORCE VERIFICATION */
+
+    curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_USE_SSL, (long)CURLUSESSL_ALL);
+
+    // FIXME: Need to get a real CERT, not self signed, for testing:
+    curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_SSL_VERIFYPEER, 1L);
```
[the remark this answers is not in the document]
marcalff: Thanks @owent

On one hand, we need an option to relax the code to use `CURLOPT_SSL_VERIFYPEER = 0L` to run unit tests with a self signed certificate.

On the other hand, disabling `CURLOPT_SSL_VERIFYPEER` defeats the purpose of `SSL`, and is not secure, so this option should never leak into production.

My concern is to find a way to satisfy the former for testing, without creating a hole with the later.

If we have an option for that, it should be abundantly clear that it is for testing only, not to be used in production. Maybe it can be available in debug releases only. I will investigate.

marcalff:
> Can we adapt these options by `LIBCURL_VERSION_NUM`? So that we can still use HTTP Client on a old system with the curl in package manager?(For example: Cent OS 7/curl 7.29.0)

Thanks @owent 

Implemented `LIBCURL_VERSION_NUM` checks.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -410,9 +476,145 @@ CURLcode HttpOperation::Setup()
...
+    if (version_range != 0)
+    {
+      curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_SSLVERSION, version_range);
+    }
+
+    /* 5 - CIPHER_LIST */
+
+    if (!ssl_options_.ssl_cipher_list.empty())
+    {
+      // TODO: Unclear when to use CURLOPT_SSL_CIPHER_LIST versus CURLOPT_TLS13_CIPHERS
```
sirzooro, line 593: Looks that OpenSSL uses separate cipher list for TLS1.3, so people who wants both 1.2 and 1.3 have to provide both ciphers list. This would require separate options in OTLP libs too.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -410,9 +476,145 @@ CURLcode HttpOperation::Setup()
...
+      return CURLE_UNKNOWN_OPTION;
+#endif
+    }
+
+    /* 2 - CLIENT KEY */
+
+    if (!ssl_options_.ssl_client_key_path.empty())
+    {
+      const char *path = ssl_options_.ssl_client_key_path.c_str();
+      curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_SSLKEY, path);
```
sirzooro, line 510: Private key may be protected by password, so you also need one more option for `CURLOPT_KEYPASSWD`

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -410,9 +476,145 @@ CURLcode HttpOperation::Setup()
...
-  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_SSL_VERIFYPEER, 0);  // 1L
-  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_SSL_VERIFYHOST, 0);  // 2L
+  if (ssl_options_.use_ssl)
+  {
+    /* 1 - CA CERT */
+
+    if (!ssl_options_.ssl_ca_cert_path.empty())
+    {
+      const char *path = ssl_options_.ssl_ca_cert_path.c_str();
+      curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_CAINFO, path);
```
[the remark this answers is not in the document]
marcalff: @sirzooro 

Thanks for the suggestion.
Implemented, and revisited existing CURLOPT calls while at it.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: @sirzooro 

Exposing a separate 1.2 and 1.3 cipher list option in the configuration is not likely to be accepted in the opentelemetry-specification.

I am experimenting with a possible mitigation, to use both when both 1.2 and 1.3 are needed.
If it works, all the better. If not, this may be a remaining limitation to resolve later.

Cases for only 1.2 or only 1.3 are trivial, and will use only the same cipher_list variable.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: @sirzooro 

Good point.

However:

- currently, the existing CLIENT_KEY variables in the spec assumes a non password protected key,
  there are no provisions to store a password
- in my understanding, the whole point of the password is to protect the private key itself.
  If we store the password in an environment variable next to the private key, it won't be any more secure
- this can be re-evaluated once there is a secured way to store secrets in the configuration

So, not planning to support passwords for now.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -398,61 +401,506 @@ void HttpOperation::Cleanup()
...
     return CURLE_FAILED_INIT;
   }
 
-  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_VERBOSE, 0);
+  CURLcode rc;
+
+  curl_error_message_[0] = '\0';
+  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_ERRORBUFFER, curl_error_message_);
+
+  rc = SetCurlOption(CURLOPT_VERBOSE, 0);
```
owent, line 523: Should it be `SetCurlOption(CURLOPT_VERBOSE, 0L)` ?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -398,61 +401,506 @@ void HttpOperation::Cleanup()
...
   // abort if slower than 4kb/sec during 30 seconds
-  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_LOW_SPEED_TIME, 30L);
-  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_LOW_SPEED_LIMIT, 4096);
+  rc = SetCurlOption(CURLOPT_LOW_SPEED_TIME, 30L);
+  if (rc != CURLE_OK)
+  {
+    return rc;
+  }
+
+  rc = SetCurlOption(CURLOPT_LOW_SPEED_LIMIT, 4096);
```
owent, line 825: Should it be `SetCurlOption(CURLOPT_LOW_SPEED_LIMIT, 4096L)` ?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -398,61 +401,506 @@ void HttpOperation::Cleanup()
...
+
+  return message;
+}
+
+template <class T>
+CURLcode HttpOperation::SetCurlOption(CURLoption option, T value)
+{
+  CURLcode rc;
+
+  rc = curl_easy_setopt(curl_resource_.easy_handle, option, value);
```
owent, line 499: `curl_easy_setopt` seems always take `long/const char*/void*` as value. Is it better to check the type of T or convert numeric and boolean value to long here?
I think it may be UB when any option need a long type but we pass a int here(Just like `CURLOPT_VERBOSE` and `CURLOPT_LOW_SPEED_LIMIT` below), the type size of int and long may be different.

owent:
Just a suggestion, can we use [libcurl-share](https://curl.se/libcurl/c/libcurl-share.html) to shared TLS sessions between curl handles?I think it will improve the performance.Or can we add this in another issue and PR?

ext/src/http/client/curl/http_operation_curl.cc
sirzooro: > * in my understanding, the whole point of the password is to protect the private key itself.
>   If we store the password in an environment variable next to the private key, it won't be any more secure

Environment variables are quite secure, at least on Linux - only process owner and root can see them via /proc filesystem. Process command line is open to everyone, so various applications usually allows to pass credentials via environment variable and/or file too.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Thanks @owent 

Fixed by using strongly typed helpers, `SetCurlLongOption` / `SetCurlStrOption` / etc.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Fixed to SetCurlLongOption().

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Fixed to SetCurlLongOption().

marcalff:
> Just a suggestion, can we use [libcurl-share](https://curl.se/libcurl/c/libcurl-share.html) to shared TLS sessions between curl handles?I think it will improve the performance.Or can we add this in another issue and PR?

Interesting, I was not aware of this possibility.

Best to address in a separate PR.

sdk/src/common/env_variables.cc
```diff
@@ -0,0 +1,157 @@
...
+{
+  std::string raw_value;
+  bool exists = GetRawEnvironmentVariable(env_var_name, raw_value);
+  if (!exists)
+  {
+    value = false;
+    return false;
+  }
+
+  if (strcasecmp(raw_value.c_str(), "true") == 0)
```
owent, line 67: I'm not sure but should it be `_stricmp` in MSVC?

marcalff: > I'm not sure but should it be `_stricmp` in MSVC?

It is, see this code at the start of env_variables.cc:

```
#ifdef _MSC_VER
#  include <string.h>
#  define strcasecmp _stricmp
#else
#  include <strings.h>
#endif
```

owent: Thanks, sorry I forgot this.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -251,6 +288,9 @@ class OtlpHttpClient
   // The configuration options associated with this HTTP client.
   const OtlpHttpClientOptions options_;
 
+  /** SSL options. */
+  ext::http::client::HttpSslOptions ssl_options_;
```
owent, line 292: Is is better to let be a member of `OtlpHttpClientOptions`, many fields are duplicated.

sdk/src/common/env_variables.cc
```diff
@@ -0,0 +1,157 @@
...
+
+bool GetRawEnvironmentVariable(const char *env_var_name, std::string &value)
+{
+#if !defined(NO_GETENV)
+  const char *endpoint_from_env = nullptr;
+#  if defined(_MSC_VER)
+  // avoid calling std::getenv which is deprecated in MSVC.
+  size_t required_size = 0;
+  getenv_s(&required_size, nullptr, 0, env_var_name);
+  std::unique_ptr<char> endpoint_buffer;
```
owent, line 32: Should it be `std::unique_ptr<char[]>`?

exporters/otlp/src/otlp_environment.cc
```diff
@@ -0,0 +1,964 @@
...
+  constexpr char kSignalEnv[]  = "OTEL_EXPORTER_OTLP_TRACES_TIMEOUT";
+  constexpr char kGenericEnv[] = "OTEL_EXPORTER_OTLP_TIMEOUT";
+
+  std::chrono::system_clock::duration value;
+  bool exists;
+
+  exists = GetDurationDualEnvVar(kSignalEnv, kGenericEnv, value);
+  if (exists)
+  {
+    return value;
```
owent, line 846: Can we check the value and make sure this value to be greater than 0?It's used by `wait_for` in OtlpHttpClient and there may be some problems when we wait for zero.

marcalff:
Found and reported #1955 while testing SSL+TLS in ASYNC mode

functional/otlp/func_http_main.cc
```diff
@@ -170,6 +170,8 @@ void payload()
   auto span = tracer->StartSpan(k_span_name);
   span->SetAttribute(k_attr_test_name, opt_test_name);
   span->End();
+
+  tracer->ForceFlushWithMicroseconds(1000000);
```
owent, line 174: `Tracer::ForceFlushWithMicroseconds` do nothing and it do not wait data exporting in async mode now.
Maybe we need another way to waiting in async mode.

sdk/src/common/env_variables.cc
marcalff: Indeed, thanks.

This code comes from the main branch, so it should be fixed there also, to avoid keeping the fix in this PR for too long.

Please see PR #1971

exporters/otlp/src/otlp_environment.cc
marcalff: Agreed, and fixed.

marcalff:
Ported all the environment variable code in a smaller PR: #1982, to merge this part independently.

marcalff:
@owent @lalitb @ThomsonTan @esigo @sirzooro 

This PR is now ready for review.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: > Looks that OpenSSL uses separate cipher list for TLS1.3, so people who wants both 1.2 and 1.3 have to provide both ciphers list. This would require separate options in OTLP libs too.

Resolved, with separate options for TLS <= 1.2 and TLS 1.3.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
marcalff: Indeed, thanks.

Fixed.

tools/setup-protobuf.sh
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 pushd /tmp
-curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v21.12/protoc-21.12-linux-x86_64.zip
-unzip -o protoc-21.12-linux-x86_64.zip -d /usr/local/
+curl -OL https://github.com/protocolbuffers/protobuf/releases/download/v3.14.0/protoc-3.14.0-linux-x86_64.zip
+unzip -o protoc-3.14.0-linux-x86_64.zip -d /usr/local/
```
owent: Why do we downgrade protobuf version? should we use the same version as it in `ci/install_protobuf.sh` ?

marcalff: @owent 

This commit is actually a revert, to leave the `tools/setup-protobuf.sh` file unchanged compared to the main branch.

`workflows/ci.yml` uses `./ci/install_protobuf.sh`

owent (approved):
LGTM and thanks.

sirzooro:
LGTM too.

api/CMakeLists.txt
```diff
@@ -98,6 +98,16 @@ if(WITH_ASYNC_EXPORT_PREVIEW)
   target_compile_definitions(opentelemetry_api INTERFACE ENABLE_ASYNC_EXPORT)
 endif()
 
+# A better place should be in sdk, not api
+if(WITH_OTLP_HTTP_SSL_PREVIEW)
+  target_compile_definitions(opentelemetry_api INTERFACE ENABLE_OTLP_HTTP_SSL)
```
lalitb, line 103: Is this experimental feature, as it is behind macro ?

api/CMakeLists.txt
```diff
@@ -98,6 +98,16 @@ if(WITH_ASYNC_EXPORT_PREVIEW)
...
+
+  if(WITH_OTLP_HTTP_SSL_TLS_PREVIEW)
+    target_compile_definitions(opentelemetry_api
+                               INTERFACE ENABLE_OTLP_HTTP_SSL_TLS)
```
lalitb, line 107: nit - should we keep the macro name also prefixed with `_PREVIEW`, as per the existing release guidelines - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/Versioning.md#release-policy.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter_options.h
```diff
@@ -57,6 +56,30 @@ struct OtlpHttpExporterOptions
...
 #endif
+
+#ifdef ENABLE_OTLP_HTTP_SSL
+  bool ssl_insecure_skip_verify{false};
+
+  std::string ssl_ca_cert_path   = GetOtlpDefaultTracesSslCertificatePath();
+  std::string ssl_ca_cert_string = GetOtlpDefaultTracesSslCertificateString();
+
+  std::string ssl_client_key_path   = GetOtlpDefaultTracesSslClientKeyPath();
+  std::string ssl_client_key_string = GetOtlpDefaultTracesSslClientKeyString();
```
lalitb, line 67: Instead of having both CA and client certificate, can we just have client certificate file/path. The user can decide whether to send single client certificate, or send the client certificate chain (if there are intermediate certificates) in the single file/path.

api/CMakeLists.txt
marcalff: Agreed.

Fixed both feature flags to have a _PREVIEW suffix.

api/CMakeLists.txt
marcalff: This is new code, so behind a macro, following the pattern for `WITH_ASYNC_EXPORT_PREVIEW` or `WITH_METRICS_EXEMPLAR_PREVIEW` when a new feature is added in a GA signal.

This helps to provide stability and mitigate risks, with optional adoption.

Once the code is merged, published in a new release, and after some time to address reported bugs if any, I hope the code can be made mainstream, by removing the feature flags ... this will be a separate PR and a separate discussion.

The good news is that the flags removal will only affect the SDK, not the API.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter_options.h
lalitb: Ok, just to be clear, we are using - 

 OTEL_EXPORTER_OTLP_CERTIFICATE -> To set the trusted CA certificate(s) which would be used to authenticate the collector during the SSL handshake.
OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE - To set the client certificate(or certificate chain) which would be used to authenticate otlp-http client during the SSL handshake.

It looks good if this is the case. Sorry for the confusion, will go through the full PR :)

marcalff: @lalitb Correct - auth is done both ways (mTLS)

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -49,6 +49,9 @@ struct OtlpHttpClientOptions
 {
   std::string url;
 
+  /** SSL options. */
+  ext::http::client::HttpSslOptions ssl_options;
```
lalitb, line 53: this can be under macro ENABLE_OTLP_HTTP_SSL_PREVIEW ?

lalitb (approved):
Changes are done well. Have few comments, but nothing blocker. Thanks for the PR :)

ext/include/opentelemetry/ext/http/client/http_client.h
```diff
@@ -110,13 +110,127 @@ struct cmp_ic
...
+  */
+  std::string ssl_cipher{};
+
+  /**
+    TLS Cipher suite.
+    This is for TLS 1.3.
+    The list is delimited by colons (":").
+    Cipher names depends on the underlying CURL implementation.
+  */
+  std::string ssl_cipher_suite{};
```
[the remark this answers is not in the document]
marcalff: Thanks for the comments.

Working on a fix, but delayed by other items, hoping to address this soon.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
marcalff: Fixed

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -967,6 +967,7 @@ OtlpHttpClient::createSession(
     request->AddHeader(header.first, header.second);
   }
   request->SetUri(http_uri_);
+  request->SetSslOptions(options_.ssl_options);
```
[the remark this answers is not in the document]
marcalff: Fixed.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -215,10 +221,52 @@ class HttpClientSync : public opentelemetry::ext::http::client::HttpClientSync
   opentelemetry::ext::http::client::Result Get(
       const nostd::string_view &url,
       const opentelemetry::ext::http::client::Headers &headers) noexcept override
+  {
+    HttpSslOptions no_ssl;
```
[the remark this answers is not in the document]
marcalff: Fixed

ext/include/opentelemetry/ext/http/client/http_client.h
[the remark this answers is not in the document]
marcalff: @lalitb Thanks for the review.

For cases when SSL may or may not be used (OTLP HTTP), I don't expect the runtime overhead to be an issue.

For cases when SSL is never used with HTTP, in particular for exporters that do not support SSL (like Zipkin), I agree there should be a way to opt out of SSL entirely, if someone does not use the OTLP HTTP exporter.

For TLS, it is also desirable to have a way to out out.

As a result, implemented the following changes:

* Flag ENABLE_OTLP_HTTP_SSL_PREVIEW is still used, but the scope is only for the OTLP HTTP exporter
* Flag ENABLE_OTLP_HTTP_SSL_TLS_PREVIEW is still used, but the scope is only for the OTLP HTTP exporter
* Flag ENABLE_HTTP_SSL_PREVIEW is added, the scope is the SSL support in the HTTP client
* Flag ENABLE_HTTP_SSL_TLS_PREVIEW is added, the scope is the SSL TLS support in the HTTP client 

This makes it possible to use ENABLE_HTTP_SSL_PREVIEW for another http client, should it become necessary, independently of the otlp http exporter, and likewise for ENABLE_HTTP_SSL_TLS_PREVIEW.

The code has a lot if ifdef/endif because of this, which is tedious, but the overall structure is regular and should be still maintainable.

Tested the following build locally, in maintainer mode:

* SSL + TLS
* SSL only
* no SSL

marcalff:
All comments to date are addressed.

Given the size of the patch, waiting for @ThomsonTan and/or @esigo to comment, and will only merge after an ok-to-merge flag.

esigo (approved):
LGTM
Thanks for the great work :)

marcalff:
Thanks. Planning to merge after the SIG meeting

PR #1799, opened 2022-11-23
Fix CI benchmark

### LINKED ISSUE #1800: [CI] Benchmarks broken by ubuntu-latest upgrade
After the upgrade of ubuntu-latest to ubuntu-22.04,
the benchmark workflow is broken (build failure in gtest 1.10 with gcc 11).

marcalff:
Fixes #1800

## Changes

Fix `OpenTelemetry-cpp benchmarks / Run OpenTelemetry-cpp benchmarks` in CI

The ubuntu-latest upgrade from 20.04 to 22.04 broke benchmarks.
Restoring ubuntu-20.04

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   benchmark:
     name: Run OpenTelemetry-cpp benchmarks
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
@@ -45,7 +45,7 @@ jobs:
       matrix:
         components: ["api", "sdk", "exporters"]
     name: Store benchmark result
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
       - uses: actions/download-artifact@master


marcalff:
@open-telemetry/cpp-approvers 

This is a blind fix, it can not be tested on PR push,
because the benchmark workflow runs on main only.

Most likely it will fix the break,
and if not it will be trivial to revert anyway.

ThomsonTan:
> @open-telemetry/cpp-approvers
> 
> This is a blind fix, it can not be tested on PR push, because the benchmark workflow runs on main only.
> 
> Most likely it will fix the break, and if not it will be trivial to revert anyway.

Can we run benchmark on PR build too?

lalitb:
> Can we run benchmark on PR build too?

It was removed from PR build purposefully, in order to not have heavy load on servers when benchmark test is running.

esigo:
> > @open-telemetry/cpp-approvers
> > This is a blind fix, it can not be tested on PR push, because the benchmark workflow runs on main only.
> > Most likely it will fix the break, and if not it will be trivial to revert anyway.
> 
> Can we run benchmark on PR build too?

That would pollute our benchmark graphs.

marcalff:
To clarify, this fixes:
`OpenTelemetry-cpp benchmarks / Run OpenTelemetry-cpp benchmarks`
not to be confused with
`CI / Benchmark`

esigo (approved):

ThomsonTan (approved):

PR #1810, opened 2022-11-25
[BUILD] Move cmake CI job builds into container

### LINKED ISSUE #1809: Move cmake CI job builds into container
To avoid surprises when GitHub rolls out new images, we should move our builds into our own containers:

`THRIFT=ON ci/run_docker.sh ./ci/do_ci.sh cmake.test`

This being tested here https://github.com/esigo/opentelemetry-cpp/pull/24/files

esigo:
Fixes #1809 (issue)

## Changes


owent (approved):

lalitb:
I don't have any concern in general building in controlled environment using docker container as long as there is no build time increase. But can we see if we can build both bazel and cmake builds through docker, and ensure bazel cache is reused.  It would defeat the purpose if half of the CI is using docker build.

esigo:
> Building in controlled environment using docker container is good move as long as there is no build time increase. And comparing the build time with previous builds, it looks similar, so should be good. Can we see if we can have both bazel and cmake builds through docker, and ensure bazel cache is reused. It would defeat the purpose if only half of the CI is using docker build :)

Thanks for the feedback. We can have the same for bazel also with cache. #1838 should cover bazel builds as well. We can close this PR and continue on #1838. Please note that we don't have to use image from artifact, we could still build the image for each CI as needed. The script will check if the image is available otherwise will build one before running the build.

lalitb:
> Thanks for the feedback. We can have the same for bazel also with cache. #1838 should cover bazel builds as well. We can close this PR and continue on #1838. Please note that we don't have to use image from artifact, we could still build the image for each CI as needed. The script will check if the image is available otherwise will build one before running the CI build.

Thanks @esigo . Can you please close this PR, so that we can track both cmake and bazel changes through #1838 .

PR #1811, opened 2022-11-26
[BUILD] move client::nosend under test_common

### LINKED ISSUE #1226: Remove gmock/gtest depepdency from ext::http::client::nosend
As discussed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/1185#discussion_r808715341), it would be good to remove gmock/gtest dependency from [ext::http::client::nosend](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/src/http/client/nosend/CMakeLists.txt). 

One option could be - We do plan to modify [EventHandler::OnEvent](https://github.com/open-telemetry/opentelemetry-cpp/blob/ef3a623a3bedd3fded167bade625d7fa392d2b88/ext/include/opentelemetry/ext/http/client/http_client.h#L203) to return the client implementation data at various phases of http(s) request handling if we can use that to return the Request object and validate for tests.

esigo:
Fixes #1226 (issue)

## Changes

As `client::nosend` is meant to be used only for unit tests, this PR moves it under `test_common`. We could use this directory for any other common code needed for testing.

This directory is compiled only when `BUILD_TESTING` is set to `ON`.


owent (approved):
LGTM

lalitb:
How about moving this to `exporters/test_common` or `ext/test_common`, just to keep the root of repo clean, with only required directories ?

PR #1820, opened 2022-11-29
Fix Prometheus target name

### LINKED ISSUE #1819: Installed target name of prometheus export changes in v1.8.0
**Describe your environment**

Build System: CMake

The installed target name of prometheus export changes from `opentelemetry-cpp::prometheus_exporter` to `opentelemetry-cpp::opentelemetry_exporter_prometheus` when using cmake.Should it keep the old name?

The old codes link `opentelemetry-cpp::prometheus_exporter` can not compile.

lalitb:
Fixes #1819

## Changes

Revert back the target name for Prometheus in otel-cpp cmake config.

This is incorrect in the latest release, probably we have to plan next release bit sooner, let's discuss in community meeting. 

### DIFF
--- a/exporters/prometheus/CMakeLists.txt
+++ b/exporters/prometheus/CMakeLists.txt
@@ -19,6 +19,9 @@ endif()
 
 add_library(opentelemetry_exporter_prometheus src/exporter.cc src/collector.cc
                                               src/exporter_utils.cc)
+
+set_target_properties(opentelemetry_exporter_prometheus
+                      PROPERTIES EXPORT_NAME exporter_prometheus)
 target_include_directories(
   opentelemetry_exporter_prometheus
   PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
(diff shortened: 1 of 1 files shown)


owent (approved):
LGTM

ThomsonTan:
What's the impact of incorrect target name? And wondering should we delay pushing 1.8.0 to package managers?

lalitb:
> What's the impact of incorrect target name? And wondering should we delay pushing 1.8.0 to package managers?

For vcpkg, it should be fine to push it with WITH_PROMETHEUS=OFF, and can be enabled in next release ?

ThomsonTan:
> > What's the impact of incorrect target name? And wondering should we delay pushing 1.8.0 to package managers?
> 
> For vcpkg, it should be fine to push it with WITH_PROMETHEUS=OFF, and can be enabled in next release. There is no breaking change for vcpkg as we never pushed metrics in that, so it sounds like a breaking change there.

We cannot just set `WITH_PROMETHEUS=OFF` in vcpkg as we produce a sub-package with prometheus like below the user can install.
> opentelemetry-cpp[prometheus]             Whether to include the Prometheus Client in the SDK

lalitb:
> We cannot just set WITH_PROMETHEUS=OFF in vcpkg as we produce a sub-package with prometheus like below the user can install.

This is old Prometheus sub-package, which is anyway has different target name ( with `deprecated` keyword) ?
Can we disable this package generation for v1..8.0, and enable for new Prometheus in v1.9.0 ?

PR #1829, opened 2022-12-01
Moved otlp_grpc_utils.cc to opentelemetry_exporter_otlp_grpc_client.

### LINKED ISSUE #1826: [Build] Build break when using WITH_OTLP_HTTP=ON, WITH_OTLP_GRPC=OFF
When compiling with the OTLP HTTP exporter but *not* the OTLP GRPC exporter,
the build fails:

```
In file included from .../opentelemetry-cpp/opentelemetry-cpp-1.8.0/exporters/otlp/src/otlp_grpc_utils.cc:4:
.../opentelemetry-cpp/opentelemetry-cpp-1.8.0/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_utils.h:6:10: fatal error: grpcpp/grpcpp.h: No such file or directory
    6 | #include <grpcpp/grpcpp.h>
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
```

Suggested fix for `exporter/otlp/CMakeLists.txt`:

- do not add `otlp_grpc_utils.cc` in the `opentelemetry_otlp_recordable` library
- add `otlp_grpc_utils.cc` in the `opentelemetry_exporter_otlp_grpc` library instead ? (to investigate)

lalitb:
Fixes #1826

## Changes

Moved otlp_grpc_utils.cc to opentelemetry_exporter_otlp_grpc_client.

### DIFF
--- a/exporters/otlp/BUILD
+++ b/exporters/otlp/BUILD
@@ -96,7 +96,6 @@ cc_library(
     deps = [
         ":otlp_recordable",
         ":otlp_grpc_client",
-        ":otlp_grpc_utils",
         "//ext:headers",
         "//sdk/src/trace",
 
--- a/exporters/otlp/BUILD
+++ b/exporters/otlp/BUILD
@@ -47,31 +47,16 @@ cc_library(
 )
 
 cc_library(
-    name = "otlp_grpc_utils",
-    srcs = [
-        "src/otlp_grpc_utils.cc",
-    ],
-    hdrs = [
-        "include/opentelemetry/exporters/otlp/otlp_grpc_utils.h",
-    ],
-    strip_include_prefix = "include",
-    tags = ["otlp"],
-    deps = [
-        "//api",
-        "//sdk:headers",
-        "@com_github_grpc_grpc//:grpc++",
-    ],
-)
-
-cc_library(
     name = "otlp_grpc_client",
     srcs = [
         "src/otlp_grpc_client.cc",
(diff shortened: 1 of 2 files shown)


esigo:
Shall we make this consistent with our bazel?
https://github.com/open-telemetry/opentelemetry-cpp/blob/82d8cb49cdc2104ef5db1f73cf362a6c7a8a76a2/exporters/otlp/BUILD#L50-L64

lalitb:
> 

Yes, good point to be consistent with bazel. Will do so

lalitb:
Or should we have single library otlp_grpc_client for both bazel and cmake, as in the latest changes in this PR - I don't see reason to have two different libraries for gRPC stuff.

esigo:
> Or should we have single library otlp_grpc_client for both bazel and cmake, as in the latest changes in this PR - I don't see reason to have two different libraries for gRPC stuff.

Yeah agree. doesn't make sense to have two libs.

ThomsonTan:
> > Or should we have single library otlp_grpc_client for both bazel and cmake, as in the latest changes in this PR - I don't see reason to have two different libraries for gRPC stuff.
> 
> Yeah agree. doesn't make sense to have two libs.

+1. Put all grpc specific code in otlp_grpc_client makes sense for both bazel and cmake.

marcalff (approved):
Thanks for the fix

esigo (approved):
LGTM
Thanks for the fix :)

PR #1838, opened 2022-12-03
Fix OpenTelemetry-cpp dependencies image CI

esigo:
Fixes OpenTelemetry-cpp dependencies image CI

## Changes

Also adds a composite which could be used in our CI. This reduces `cmake.exporter.otprotocol.test` job time to half.
The composite can be used as:

[.github/actions/setup/action.yaml](https://github.com/open-telemetry/opentelemetry-cpp/pull/1838/files#diff-5243a283b8b14b6692858f044bd84aa8a274a52804df8a29eca0cccb06382af5) is a composite that can be reused in any CI job. This composite downloads the tar file of the build image and loads it in Docker. Here is an example of using the composite:

```yaml
cmake_otlp:
  name: CMake test (with otlp-exporter)
  runs-on: ubuntu-20.04
  steps:
  - uses: actions/checkout@v3
    with:
      submodules: 'recursive'
  - uses: ./.github/actions/setup
    with:
      github_token: ${{secrets.GITHUB_TOKEN}}
      pull_sha: ${{github.event.pull_request.head.sha}}
  - name: cmake test
    run: |
      docker images
      bash ci/ci_run_docker.sh ci/do_ci.sh cmake.exporter.otprotocol.test
```

ci/Dockerfile.ubuntu installs dependencies based on the arguments passed in build time, this includes bazel, abseil, thrift, gRPC.

ci/ci_do_ci.sh runs any command passed as argument inside a container. This script will build the image if the image is not available. for example:
`bash ci/ci_run_docker.sh -tag -i` builds an image with gRPC, thrift and abseil installed.
or  `bash ci/ci_run_docker.sh -ag ci/do_ci.sh cmake.exporter.otprotocol.test` will run `cmake.exporter.otprotocol.test` test of CI inside the container.


.github/actions/setup/action.yaml
```diff
@@ -0,0 +1,34 @@
+name: "Setup"
+description: "Downloads and loads build image"
+
+inputs:
+  image_name:
+    description: “Build image name”
```
ThomsonTan, line 6: The double quotes here seems non-ascii chars, or at least different from the one in line 2?

esigo:
closing as was not needed.

PR #1846, opened 2022-12-07
Enable setting Span endtime for ETW exporter

lalitb:
## Changes

While it is possible to set the end-time while ending span as per the spec, the ETW exporter use the endtime passed as argument in Span::End(). This PR fixes it.


exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -899,7 +899,15 @@ class Span : public opentelemetry::trace::Span
    */
   void End(const opentelemetry::trace::EndSpanOptions &options = {}) noexcept override
   {
-    end_time_ = std::chrono::system_clock::now();
+    if (options.end_steady_time == opentelemetry::common::SteadyTimestamp())
```
ThomsonTan, line 902: Update this to compare with 0 explicitly. And add TODO to choose a different value to represent "not set" status, as 0 is a valid value, and the user might set endtime to it.

ThomsonTan (approved):

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
lalitb: Thanks fixed this.

PR #1860, opened 2022-12-14
Collect and Export metric data before PeriodicMetricReader shutdown.

### LINKED ISSUE #1858: Collect in PeriodicExportingMetricReader ForceFlush before exporting

Consider the scenario where some executable emitting metrics through the `PeriodicExportingMetricReader` is about to exit, but desires to collect and export the last value of an `ObservableGauge` one more time without waiting for the periodic collect and export.

When using the `PeriodicExportingMetricReader`, it would be nice if `ForceFlush()` collects before exporting.

I have a metric that could be a `Counter`, but it gets updated very often and emits too many metric points. Therefore I really want to just collect and export it periodically using an `ObservableGauge`. I could work around it by using a `Counter` and periodically updating that myself, forcing one final update and flush upon exit. However, it seems the behavior I desire is common among other language SDKs as well, so I think it may be desired by others for `ForceFlush` to follow the behavior seen elsewhere.

**Additional context**

I hope I am not missing anything, but if I understand correctly, in the current implementation, an export is performed but no metrics are collected before exporting.
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/export/periodic_exporting_metric_reader.cc#L85-L88

Other language SDKs appear to collect before exporting on ForceFlush:

Java: https://javadoc.io/static/io.opentelemetry/opentelemetry-sdk-metrics/1.21.0/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.html#forceFlush()
Python (see force_flush at the bottom, the superclass force_flush collects first): https://opentelemetry-python.readthedocs.io/en/latest/_modules/opentelemetry/sdk/metrics/_internal/export.html#PeriodicExportingMetricReader.force_flush

lalitb:
Fixes #1858 

## Changes
As discussed in this issue,  `PeriodicMetricReader` should ensure the metric collection and export is done once before shutdown, for all the data points be uploaded. This is also supported in [Java ](https://github.com/open-telemetry/opentelemetry-java/blob/f6deb4c1b7b9ed250f3f7cda992a012b6e738af2/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/export/PeriodicMetricReader.java#L83), and discussed in specs - https://github.com/open-telemetry/opentelemetry-specification/issues/2983

The shutdown sequence is:
`MeterProvider::Shutdown()`  -> `MeterContext::Shutdown()` -> `MetricCollector()` -> `Shutdown()` -> `MetricReader::Shutdown()` -> `PeriodicMetricReader::OnShutdown()`
     -> `[[[Perform Collect and Shutdown ]]]` ->` MetricExporter::Shutdown()`

This ensures that the MericExporter instance remains valid while MetricReader is doing collect-and-export during it's shutdown. Once the collect-and-export is completed, the MetricExporter is shutdown.


esigo (approved):
LGTM
Thanks for the PR :)

PR #1871, opened 2022-12-15
Log a warning when the BatchSpanProcessor queue is full.

### LINKED ISSUE #1870: Log a warning when BatchSpanProcessor queue is full and span is dropped
Yes, the problem is that when the `BatchSpanProcessor` queue is full, spans are silently dropped. This may result in incomplete traces downstream. As lost spans can have many reasons, it is hard to pinpoint the problem if it originates from the BSP queue being full.

It would be nice if a log message is emitted to notify the user a span got dropped because the queue was full. This will allow them to understand the problem and mitigate it quickly by tweaking their BSP queue size.

Making the queue blocking, but this is not allowed by the specification.

johanpel:
Fixes #1870 

## Changes

Adds a warning when a span is dropped in case the `BatchSpanProcessor` queue is full.

* [n/a] Changes in public API reviewed

### DIFF
--- a/sdk/test/trace/batch_span_processor_test.cc
+++ b/sdk/test/trace/batch_span_processor_test.cc
@@ -190,10 +190,10 @@ struct MockLogHandler : public sdk::common::internal_log::LogHandler
   using Message = std::pair<sdk::common::internal_log::LogLevel, std::string>;
 
   void Handle(sdk::common::internal_log::LogLevel level,
-              const char *file,
-              int line,
+              const char * /*file*/,
+              int /*line*/,
               const char *msg,
-              const sdk::common::AttributeMap &attributes) noexcept
+              const sdk::common::AttributeMap & /*attributes*/) noexcept
   {
     messages.emplace_back(level, msg);
   }
--- a/sdk/test/trace/batch_span_processor_test.cc
+++ b/sdk/test/trace/batch_span_processor_test.cc
@@ -193,7 +193,7 @@ struct MockLogHandler : public sdk::common::internal_log::LogHandler
               const char * /*file*/,
               int /*line*/,
               const char *msg,
-              const sdk::common::AttributeMap & /*attributes*/) noexcept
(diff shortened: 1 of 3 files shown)


sdk/test/trace/batch_span_processor_test.cc
```diff
@@ -211,6 +234,18 @@ TEST_F(BatchSpanProcessorTestPeer, TestManySpansLoss)
 
   // Span should be exported by now
   EXPECT_GE(max_queue_size, spans_received->size());
+
+  // At least one log message warning about the dropped span should have been emitted by now.
+  auto &vec = dynamic_cast<MockLogHandler *>(log_handler.get())->messages;
```
owent, line 239: Should we use `static_cast` here?Sometimes we will compile otel-cpp without RTTI.(With `-fno-rtti` or `/GR-`)

johanpel: Thanks for pointing this out, CI was also failing because of this.

johanpel:
> Thanks @johanpel for the PR.
> 
> FYI, there is a delay in running CI, because we have to manually allow it after a new commit, due to github rules for first time contributors.
> 
> I see the CLA is signed already, which is great.
> 
> Once this PR is merged, the process will be much easier ;-)

Makes sense. I've added a few commits to hopefully fix the CI issues. Most were related to the timing around the `BatchSpanProcessor`. With Valgrind/ASAN/TSAN the `BatchSpanProcessor` still seems to be able to keep the queue non-full in the `BatchSpanProcessorTestPeer/TestManySpansLoss` test. 

I've reduced the queue size. Hope that gives more consistent results.

johanpel:
It was still a bit too flaky, added a commit to only check logs if anything is actually dropped.

owent (approved):
LGTM

marcalff (approved):
LGTM, thanks for the fix.

PR #1873, opened 2022-12-19
Migrate from @bazel_tools//platforms to https://github.com/bazelbuild/platforms

davidmandle:
## Changes

This MR migrates [platforms](https://bazel.build/extending/platforms) constraints from the deprecated `@bazel_tools//platforms` target to instead depend on the `https://github.com/bazelbuild/platforms` repository, per https://github.com/bazelbuild/bazel/issues/8622

This enables compatibility with the newly released [Bazel 6.0.0](https://github.com/bazelbuild/bazel/releases).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Increment the:
   cmake when using C++20 [#1852](https://github.com/open-telemetry/opentelemetry-cpp/pull/1852)
 * [SEMANTIC CONVENTIONS] Upgrade to version 1.16.0
   [#1854](https://github.com/open-telemetry/opentelemetry-cpp/pull/1854)
+* [BUILD] Migrate from @bazel_tools//platforms to [Bazel Platforms](https://github.com/bazelbuild/platforms)
+  to enable Bazel 6.0.0 compatibility [#1873](https://github.com/open-telemetry/opentelemetry-cpp/pull/1873)
 
 ## [1.8.1] 2022-12-04
 
--- a/bazel/BUILD
+++ b/bazel/BUILD
@@ -2,5 +2,5 @@ package(default_visibility = ["//:__subpackages__"])
 
 config_setting(
     name = "windows",
-    constraint_values = ["@bazel_tools//platforms:windows"],
+    constraint_values = ["@platforms//os:windows"],
 )
--- a/bazel/curl.BUILD
+++ b/bazel/curl.BUILD
@@ -7,7 +7,7 @@ package(features = ["no_copts_tokenization"])
 config_setting(
     name = "windows",
     constraint_values = [
-        "@bazel_tools//platforms:windows",
+        "@platforms//os:windows",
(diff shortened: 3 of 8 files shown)


davidmandle:
> Could this break some older version of bazel?

I don't know for sure, but I do know that it continues to build successfully with the Bazel version specified in `.bazelversion` (4.2.0), and that other large projects have already successfully undergone this migration (e.g. `grpc`: https://github.com/grpc/grpc/pull/29842).

davidmandle:
Let me know if there's any actions I can take to address concerns about Bazel backwards compatibility. For what its worth, the `grpc` library, which has undergone this migration, tests over its set of [supported Bazel versions ](https://github.com/grpc/grpc/blob/99a7f2995e66f9d6f5c3403dc7178238251fb2b5/bazel/supported_versions.txt).

lalitb:
I don’t see any issue if this migration enable supporting bazel v6.0.0, is also working with v4.2.0. As of now we haven’t documented minimal supported bazel version - perhaps we can start doing that once we bump up bazel version to 6.0.0. 

Also not related to this PR, but we need to think about continuing g++4.8 support, as that support is restricted with lower version of gRPC and bazel. 

Approving this, but need one more approval to merge, and most of the approvers/maintainers are not active during holiday season. Thanks for the PR.

lalitb (approved):

owent (approved):
LGTM

owent (changes requested):
It seems also need upgrade gtest to 1.11.0 or upper and upgrade gRPC to 1.49.0 or upper.
Or the dependencies still use `@bazel_tools//platforms`, which can not compile with latest bazel.

lalitb:
> It seems also need upgrade gtest to 1.11.0 or upper and upgrade gRPC to 1.49.0 or upper.

Oh didn't realize that. Thanks for pointing out. Is it better to upgrade gRPC now, or once we bump the bazel version?

davidmandle:
Thanks @owent, I've upgraded gtest to 1.12.1 and gRPC to 1.49.2. Let me know if there's any other actions I should take.

lalitb:
> Thanks @owent, I've upgraded gtest to 1.12.1 and gRPC to 1.49.2. Let me know if there's any other actions I should take.

CMake is still using older versions of these libraries, and also [third_part_releases](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release) file should have these updated versions.

owent (approved):
LGTM now.And the gRPC version in [ci/setup_grpc.sh](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/setup_grpc.sh) and [ci/do_ci.sh](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/do_ci.sh) can also be updated.

davidmandle:
Thanks. I believe now I've updated the versions throughout.

PR #1884, opened 2022-12-22
Update logs sdk and api to follow specification

### LINKED ISSUE #1687: [Logs API] LoggerProvider changes

### LINKED ISSUE #1688: [Logs API] - Logger Changes

### LINKED ISSUE #1690: [Logs SDK] Modify InstrumentationScope class
Instrumentation class needs to be modified to store key-value attributes. This is required for storing the event_name and attributes passed through LoggerProvider::GetLogger() method.

This is also consistent with the OTLP proto specs- 

https://github.com/open-telemetry/opentelemetry-proto/blob/1e69bf2e789665a56eed3a3e9cf4fd51ff784ab6/opentelemetry/proto/common/v1/common.proto#L71

### LINKED ISSUE #1692: [Logs SDK] LoggerProvider, LoggerContext changes.
- LoggerProvider
  - LoggerProvider::GetLogger() : As per the new prototype. 
       -   Add `event_domain` to the list of attributes passed as argument, and eventually pass them to InstrumentationScope.
       -    Pass `include_trace_context` flag to Logger object during it's creation.
  - AddProcessor/RemoveProcessor - Pass LogRecordProcessor instead of LogProcessor. Should we change the name of methods too ?

- LoggerContext
   - AddProcessor/RemoveProcessor - Pass LogRecordProcessor instead of LogProcessor. Should we change the name of methods too ?

### LINKED ISSUE #1693: [Logs SDK] - LogRecord , Logger changes
- Logger 
   -  Constructor takes `include_trace_context` flag to use it further within spancontext fetch logic.
   - CreateLogRecord() - This will return new instance of sdk::LogRecord object.
   - EmitLogRecord() - Would invoke LogRecordProcessor::OnEmit method to pass logrecord to processor.
   - SpanContext fetch logic - 
        -  If valid `SpanContext` is passed through `EmitLogRecord()`, or valid `TraceId`/`SpanId` is passed using existing `Log()` method, use that irrespective of the value of `include_trace_context` flag.
        -  Else, if  `include_trace_context`  is set, use the `SpanContext` from currently active span.

- LogRecord
   -  The only way to create LogRecord object is through `Logger::CreateLogRecord()`. It's constructor will  take Logger object as argument. 
   - Implement the setters calling corresponding setters from Recordable. The Recordable object would be obtained through the Logger object passed in constructor above.

owent:
Fixes #1692
Fixes #1693
Fixes #1687
Fixes #1688 

## Changes

* [x] Add `include_trace_context` and scope attribute for logs API and SDK.
* [x] Add type trait to set field of LogRecord.
* [x] Add `EventLoggerProvider` and `LoggerProvider`


lalitb:
Sorry for delay on this. Will provide my comments in this week.

owent:
> Sorry for delay on this. Will provide my comments in this week.

Thanks, should we mark the old `Logger::Log()` APIs as deprecated?The new `Logger::EmitLogRecord()` and `EventLogger::EmitEvent()` accept the SpanID, TraceID, Attributes and etc in any order.

lalitb:
@owent - Do you think this issue - https://github.com/open-telemetry/opentelemetry-specification/issues/3086 - will bring any changes in the Event API/SDK implementation ?

lalitb:
> should we mark the old Logger::Log() APIs as deprecated?The new Logger::EmitLogRecord() and EventLogger::EmitEvent() accept the SpanID, TraceID, Attributes and etc in any order.

I thought Logger::Log() API provides single line convenience API to instrumentation libraries for emitting logs, and we should keep them as it is.

owent:
> @owent - Do you think this issue - [open-telemetry/opentelemetry-specification#3086](https://github.com/open-telemetry/opentelemetry-specification/issues/3086) - will bring any changes in the Event API/SDK implementation ?

Not yet. This PR implement the main branch of https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md by now.And the EventLogger is a standalone component.
Maybe some APIs of EventLogger may change in the future, I can continue modify them then.

owent:
> > should we mark the old Logger::Log() APIs as deprecated?The new Logger::EmitLogRecord() and EventLogger::EmitEvent() accept the SpanID, TraceID, Attributes and etc in any order.
> 
> I thought Logger::Log() API provides single line convenience API to instrumentation libraries for emitting logs, and we should keep them as it is.

OK, we can keep `Logger::Log()`, just wondering should we also modify the parameters so SpanID, TraceID, Attributes and etc and be passed in any order, just like `Logger::EmitLogRecord()`?

api/include/opentelemetry/logs/event_logger.h
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * Emit a event Log Record object with arguments
+   *
+   * @param event_name Event name
+   * @param log_record Log record
+   */
+  template <class... ArgumentType>
+  void EmitEvent(Severity severity, nostd::string_view event_name, ArgumentType &&... args)
```
lalitb, line 95: Why do we need Severity as argument ?

api/include/opentelemetry/logs/event_logger.h
```diff
@@ -0,0 +1,115 @@
...
+   * Emit a event Log Record object with arguments
+   *
+   * @param event_name Event name
+   * @param log_record Log record
+   * @tparam args Arguments
+   */
+  template <class... ArgumentType>
+  void EmitEvent(nostd::string_view event_name,
+                 nostd::unique_ptr<LogRecord> &&log_record,
+                 ArgumentType &&... args)
```
lalitb, line 74: We are passing log_record, and parameter packs both ? Isn't event_name and log_record object enough.

api/include/opentelemetry/logs/event_logger.h
```diff
@@ -0,0 +1,115 @@
...
+#  include "opentelemetry/nostd/shared_ptr.h"
+#  include "opentelemetry/nostd/span.h"
+#  include "opentelemetry/nostd/string_view.h"
+#  include "opentelemetry/nostd/type_traits.h"
+#  include "opentelemetry/nostd/unique_ptr.h"
+#  include "opentelemetry/trace/span_context.h"
+#  include "opentelemetry/trace/span_id.h"
+#  include "opentelemetry/trace/trace_flags.h"
+#  include "opentelemetry/trace/trace_id.h"
+#  include "opentelemetry/version.h"
```
lalitb, line 29: I don't see all these headers used in this file ? Are they may be used with expansion of parameter pack argument ?

api/include/opentelemetry/logs/event_logger.h
owent: This is used to create `LogRecord`, just like `EmitLogRecord`.

api/include/opentelemetry/logs/event_logger.h
owent: You are right, I remove it now.

api/include/opentelemetry/logs/event_logger.h
owent: Thanks, these headers are removed.

api/include/opentelemetry/logs/event_logger.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /* Returns the delegate logger of this event logger */
+  virtual nostd::shared_ptr<Logger> GetDelegateLogger() noexcept = 0;
+
+  /**
+   * Create a Log Record object
+   *
+   * @return nostd::unique_ptr<LogRecord>
+   */
+  virtual nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept = 0;
```
lalitb, line 43: Do we need this method for EventLogger, or instead it can be obtained as:
`auto log_record = event_logger->GetDelegateLogger()->CreateLogRecord()`

api/test/logs/provider_test.cc
```diff
@@ -17,20 +19,26 @@ namespace nostd = opentelemetry::nostd;
...
 class TestProvider : public LoggerProvider
 {
-  nostd::shared_ptr<Logger> GetLogger(nostd::string_view /* logger_name */,
-                                      nostd::string_view /* options */,
-                                      nostd::string_view /* library_name */,
-                                      nostd::string_view /* library_version */,
-                                      nostd::string_view /* schema_url */) override
+  nostd::shared_ptr<Logger> GetLogger(
+      nostd::string_view /* logger_name */,
+      nostd::string_view /* options */,
```
lalitb, line 24: should we remove `options`? I don't think we need to deprecate it as Logs API is still experimental

api/include/opentelemetry/logs/logger.h
```diff
@@ -35,6 +36,48 @@ namespace logs
...
 {
 public:
+  /**
+   * Utility function to help to make a attribute view from initializer_list
+   *
+   * @param attributes
+   * @return nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>
+   */
+  inline static nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>
+  MakeAttributes(std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>>
```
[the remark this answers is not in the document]
owent: initializer_list can not be detected by universial reference(`ArgumentType &&`).Is there any better suggestion here?

api/include/opentelemetry/logs/event_logger.h
owent: Yes,`Severity` cam be also added into `logger_type_traits.h`, I can finish it one or two days later. We are in holiday these days.

api/include/opentelemetry/logs/event_logger.h
owent: Yes, it's duplicate here. I will remove it in one or two days.

api/test/logs/provider_test.cc
owent: I'm not sure why `options` is added here.Maybe we can alose remove it from all Logger's APIs.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
lalitb: What if we remove the variadic template argument in from EmitLogRecord() and EmitEvent(), and have a method in LogRecord to add attributes?

`LogRecord::SetAttribute(nostd::string_view key,
                            const opentelemetry::common::AttributeValue &value)
LogRecord::SetAttributes(const common::KeyValueIterable &attributes)`

api/include/opentelemetry/logs/event_logger.h
lalitb: No problem @owent, Happy Lunar New Year. Enjoy your holidays.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
owent: `LogRecord` already  has `SetAttribute`. I think it's easier to use for users if they do not need to emit log in just one call. Just like `Logger::Log`, users don't need `CreateLogRecord()` and then `SetAttribute()`.
`MakeAttributes` can be used to set attributes easier.

```cpp
  logger->EmitLogRecord(Severity::kFatal, "Logging an initializer list",
                        Logger::MakeAttributes({{"key1", "value 1"}, {"key2", 2}}));
```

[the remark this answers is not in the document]
lalitb: Thanks for explanation. For logger API, we will now have two different methods doing similar thing:

```cpp
logger->Log(Severity::kFatal, "Logging an initializer list", {{"key1", "value 1"}, {"key2", 2}})
```

And
```cpp
logger->EmitLogRecord(Severity::kFatal, "Logging an initializer list",
                        Logger::MakeAttributes({{"key1", "value 1"}, {"key2", 2}}));
```

And the existing (first) one is much simpler. Or are you thinking to remove the Logger::Log() method after adding the new method ?

api/test/logs/provider_test.cc
lalitb: Yes, I think it would be good to remove it from all Logger's API.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
owent: I have the same question above( https://github.com/open-telemetry/opentelemetry-cpp/issues/1884#issuecomment-1371741041_ ). I think we can discuss if we should keep `Logger::Log`.

api/test/logs/provider_test.cc
owent: `options` is removed.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
lalitb: I also don't see advantage having both the methods. We can get rid of `Logger::Log`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,25 +98,43 @@ class Logger
...
+        detail::LogRecordSetterTrait<typename std::decay<ArgumentType>::type>::template Set(
+            log_record.get(), std::forward<ArgumentType>(args))...);
 
     EmitLogRecord(std::move(log_record));
   }
 
+  /**
+   * Emit a Log Record object with arguments
+   *
+   * @param log_record
```
lalitb, line 124: Nit - we can remove log_record, and perhaps mention the allowed parameters (in any order).

api/include/opentelemetry/logs/logger.h
```diff
@@ -55,25 +98,43 @@ class Logger
   virtual void EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record) noexcept = 0;
 
   /**
-   * Emit a Log Record object with custom span context
+   * Emit a Log Record object with arguments
    *
-   * @param log_record
+   * @param log_record Log record
+   * @tparam args Arguments
```
lalitb, line 104: Nit - should we mention the allowed parameters (in any order), as this will go into the API documentation.

api/include/opentelemetry/logs/logger.h
owent: Done.

api/include/opentelemetry/logs/logger.h
owent: Thanks and comments are added.

api/include/opentelemetry/logs/logger.h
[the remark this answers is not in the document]
owent: Thanks, `Logger::Log` is removed now.

api/include/opentelemetry/logs/logger.h
```diff
@@ -35,578 +36,277 @@ namespace logs
...
-    this->Log(Severity::kInfo, attributes);
+    this->EmitLogRecord(Severity::kDebug, std::forward<ArgumentType>(args)...);
   }
 
   /**
    * Writes a log with a severity of info.
-   * @param message The message of the log
-   * @param attributes The attributes of the log as an initializer list
+   * @tparam args Arguments which can be used to set data of log record by type.
+   *  Severity                                -> severity, severity_text
```
lalitb, line 221: Should we not mention Severity as argument here, and also in Warn(), Error() and Fatal() ?

lalitb (approved):
LGTM Thanks.

api/include/opentelemetry/logs/logger.h
owent: My mistake and thanks.I also add a type trait to prevent to pass `Severity` as args.

lalitb:
Thanks @owent  for fixing all the comments. This PR was discussed in today's community meeting. As we are using variadic templates/pack-paraments, and other meta-programming concepts in the code, there was interest for other community members to have a look into the PR. The suggestion was to keep this PR open for review till next community meeting (i.e, 30th Jan) for sufficient time for others to review and comment if required :)

owent:
> Thanks @owent for fixing all the comments. This PR was discussed in today's community meeting. As we are using variadic templates/pack-paraments, and other meta-programming concepts in the code, there was interest for other community members to have a look into the PR. The suggestion was to keep this PR open for review till next community meeting (i.e, 30th Jan) for sufficient time for others to review and comment if required :)

Thanks.

api/include/opentelemetry/logs/event_logger.h
```diff
@@ -0,0 +1,94 @@
...
+    {
+      return;
+    }
+
+    IgnoreTraitResult(
+        detail::LogRecordSetterTrait<typename std::decay<ArgumentType>::type>::template Set(
+            log_record.get(), std::forward<ArgumentType>(args))...);
+
+    EmitEvent(event_name, std::move(log_record));
+  }
```
marcalff, line 84: @owent 

So, with the use of variadic templates, type traits, lamda expressions in the implementation, etc ... there is a considerable complexity here.

Because this is in the public API, users trying to figure out how to use this might have some questions and will need a clear documentation in `opentelemetry.io` with examples, and I include myself in "users" here.

This also increases complexity for maintenance, because if this breaks, understanding all the details and providing a fix will take more time, compared to more trivial C++ code.

I assume that the complexity here also saves from some repetitive code that would be also hard to maintain otherwise, so there are plus sides as well.

Overall, the complexity argument does not make the code invalid, so I view it as:

* subjective (people with different experiences with these patterns will have different views)
* and not objective (a given area in the code is technically broken and needs a fix)

To confirm, ok with me to have this code merged to main, knowing fully that I would need to improve my own variadic template skills as a pre requisite if I ever have to make changes here.

LGTM for this part, did not check everything else yet.

marcalff (commented):
Ok with the coding patterns used.

api/include/opentelemetry/logs/event_logger.h
owent: Thanks for your feedback. Here is my thinking about this complexity.There is a little different between trace and log APIs, with trace, we always need a variable to hold the lifetime of span(Span or Scope) and decide when to call `End()` by RAII, so we can set data field of `Span` one by by one after we create the span.
But in the use case of logs, it's easier for users(I'm also a user of both log and trace.) to use by call just one function, this function is `Logger::Log()` before and `Logger::EmitLogRecord()` now.There are also many data fields we need set to `LogRecord`, log body, trace information, attributes, timestamp and so on.There are three ways to set attributes before(By `const T&`, `std::initializer_list<std::pair<nostd::string_view, common::AttributeValue>>`, `const common::KeyValueIterable &`), and user also can decide whether to set other data fields, log body or trace information. Just consider log body, trace context and attributes, there will be more than `2*2*3+2*2=16` possibilities.Actually, we had more than 10 overloads of `Logger::Log()` before.If we have more options for logs, then the overloads will increase exponentially(I'm trying to add TraceContext now, according to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#logrecord).It will also be hard to maintain these overloads, especially in the future we may add more fields.

I think we can disscuss more deeply if there is any better idea about how to solve this problem.

*The approach of this PR try to use `template <class ValueType> struct LogRecordSetterTrait;` to implement these overloads by type, and also ignore the argument order.We can use `std::decay<ArgumentType>::type` to remove `const`, `reference` and `violate` of `ArgumentType` and use it as the template argument of `template <class ValueType>`, so we can implement any overload action by just write one template specialization for `LogRecordSetterTrait`. And the `template <class ArgumentType> inline static LogRecord *Set(LogRecord *log_record, ArgumentType &&arg) noexcept` use the universal reference so it can accept both left value reference and right value with or without `const` or `violate`.*

esigo (approved):
LGTM
Thanks for the great work :)
I agree that, maintaining this will be challenging but I see no better option.

api/include/opentelemetry/logs/event_logger.h
marcalff: Thanks @owent for the details.

This seems the best solution then ... and great work.

marcalff (approved):

sdk/test/logs/logger_provider_sdk_test.cc
```diff
@@ -68,15 +71,30 @@ TEST(LoggerProviderSDK, LoggerProviderLoggerArguments)
...
-  // Check GetLogger(logger_name, args)
-  std::array<nostd::string_view, 1> sv{"string"};
-  nostd::span<nostd::string_view> args{sv};
-  auto logger2     = lp->GetLogger("logger2", args, "opentelelemtry_library", "", schema_url);
+  auto logger2     = lp->GetLogger("logger2", "opentelelemtry_library", "", schema_url);
   auto sdk_logger1 = static_cast<opentelemetry::sdk::logs::Logger *>(logger1.get());
   auto sdk_logger2 = static_cast<opentelemetry::sdk::logs::Logger *>(logger2.get());
   ASSERT_EQ(sdk_logger2->GetInstrumentationScope(), sdk_logger1->GetInstrumentationScope());
+
+  auto logger3 = lp->GetLogger("logger3", "opentelelemtry_library", "", schema_url, true,
```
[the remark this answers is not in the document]
owent: This can be done when anyone implementing #1690 later, it need scope attribute and I think I can finish it in another PR.

PR #1887, opened 2022-12-22
[BUILD] Fixing CMake to build GTest on Windows

bachittle:
## Changes

Trying to compile with CMake on Windows. It breaks because of the Required clause. Removing it gets it working and installs required vcpkg files. 

some improvements to consider:
- this vcpkg run is very slow
- googletest is already added in third_party (default command the docs tell you to run is --recursive, which gets googletest)

* [ ] `CHANGELOG.md` updated for non-trivial changes (change is trivial)
* [ ] Unit tests have been added (comment if this is needed)
* [ ] Changes in public API reviewed (no changes to public API)


bachittle:
well it seems that even with this change the code still doesn't compile, it can't seem to find the vcpkg libraries afterwards. I might just experiment with including GTest manually like how it is done for linux (just including .so files instead of .a files)

ThomsonTan:
> well it seems that even with this change the code still doesn't compile, it can't seem to find the vcpkg libraries afterwards. I might just experiment with including GTest manually like how it is done for linux (just including .so files instead of .a files)

Wondering whether the vcpkg packages were installed correctly to <repo_root>\tools\vcpkg? The vcpkg is added to current CMake search path as below.

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CMakeLists.txt#L217

bachittle:
@ThomsonTan the vcpkg root was the issue, it would find it on first install but not on consecutive runs. Now I got my code to compile, but the test cases fail. Most likely due to GMOCK_LIB dependency. I couldn't find a vcpkg dependency of something that is similar to libgmock-dev that allowed me to get it to work on linux...

lalitb:
> @ThomsonTan the vcpkg root was the issue, it would find it on first install but not on consecutive runs. Now I got my code to compile, but the test cases fail. Most likely due to GMOCK_LIB dependency. I couldn't find a vcpkg dependency of something that is similar to libgmock-dev that allowed me to get it to work on linux...

What is the failure error in tests ? gmock should be installed as part of gtest installation using vcpkg.

bachittle:
It works now. Only one failed test, which is also what showed up in my linux tests. Setting the toolchain file so late in the build process seems to not have an effect on finding proper packages, so I just set it manually under cmake-gui. Not ideal, but if there are any better solutions let me know!

test_common/src/http/client/nosend/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if(${BUILD_TESTING})
   endif()
 
   target_link_libraries(
-    opentelemetry_http_client_nosend ${GTEST_BOTH_LIBRARIES} ${GMOCK_LIB}
+    opentelemetry_http_client_nosend ${GTEST_BOTH_LIBRARIES}
```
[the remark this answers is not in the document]
lalitb: yeah, GTEST_BOTH_LIBRARIES includes both gtest and gmock as set here - https://github.com/open-telemetry/opentelemetry-cpp/blob/a343da043e1351c3cc3003bfce94724345ee22f1/CMakeLists.txt#L464

lalitb:
@bachittle  - Can you fix the CLA, so that it can be reviewed and merged?

bachittle:
It says that EasyCLA check passed, but now there are other CICD issues...

CMakeLists.txt
```diff
@@ -236,9 +236,12 @@ function(install_windows_deps)
   set(ENV{ARCH} ${ARCH})
   execute_process(
     COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/tools/setup-buildtools.cmd)
-  set(CMAKE_TOOLCHAIN_FILE
-      ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
-      PARENT_SCOPE)
+  # set(CMAKE_TOOLCHAIN_FILE
+  #     ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
+  #     PARENT_SCOPE)
+  message("Make sure that vcpkg.cmake is set as the CMAKE_TOOLCHAIN_FILE at the START of the cmake build process!
```
owent, line 242: Can we print this message only when we do not set toolchain file?

```cmake
if(NOT CMAKE_TOOLCHAIN_FILE)
  message(WARNING "Make sure that vcpkg.cmake is set as the CMAKE_TOOLCHAIN_FILE at the START of the cmake build process!
Can be command-line arg (cmake -DCMAKE_TOOLCHAIN_FILE=...) or set in your editor of choice.")
endif()
```

bachittle: The issue is the cmakefile tries to set the CMAKE_TOOLCHAIN_FILE in the cache at the start. This does not work unless its at the very start of the build process.

image

So its set, but it wasn't detecting my vcpkg unless I set it manually using `cmake -DCMAKE_TOOLCHAIN_FILE=... `
It may have to do with the submodules thinking the source directory is different, but gtest was not getting detected even after running install_windows_deps()

owent: If we use `cmake -DCMAKE_TOOLCHAIN_FILE=...` to set toolchain, we should also replace `NOT DEFINED CMAKE_TOOLCHAIN_FILE` by (NOT DEFINED CMAKE_TOOLCHAIN_FILE AND NOT DEFINED CACHE{CMAKE_TOOLCHAIN_FILE}).

bachittle: It can also just be removed entirely since it doesn't work

owent: Sorry for my mistake, `DEFINED <VAR NAME>` can also be used to check cached variables.
LGTM after format problem is solved.

tools/setup-buildtools.cmd
```diff
@@ -58,6 +58,7 @@ if %ERRORLEVEL% == 1 (
   popd
 ) else (
   echo Using existing vcpkg installation...
+  set "VCPKG_ROOT=%~dp0\vcpkg"
```
ThomsonTan, line 61: Should we extract `VCPKG_ROOT` from the full path of detected vcpkg.exe?

CMakeLists.txt
ThomsonTan: @bachittle I am still not clear how CMakeCache.txt affects the validity of `CMAKE_TOOLCHAIN_FILE` here. Could you please elaborate a little more? Thanks.
> The issue is the cmakefile tries to set the CMAKE_TOOLCHAIN_FILE in the cache at the start. This does not work unless its at the very start of the build process.

lalitb (approved):

CMakeLists.txt
bachittle: so I've been doing some digging and I think the issue is that you have to set the CMAKE_TOOLCHAIN_FILE variable very early on in the build process as it indicates to cmake that other top level includes need to be set for things to work properly. When this is set later on it can't find gtest, so maybe some hidden higher-level include is being skipped. Either that or I am doing something incorrectly...

ThomsonTan: Wondering what is your full CMake command line.

ThomsonTan: Asking because I'd like to check if OTLP is not enabled in your case. For the `WITH_OTLP=ON` case, `${CMAKE_TOOLCHAIN_FILE}` is included explicitly in below line before run `find_package(...)` again, so there is no need to set the variable `CMAKE_TOOLCHAIN_FILE` at very early time. Probably we should do the same for the call to `find_package(GTest)` which seems the issue here.

https://github.com/open-telemetry/opentelemetry-cpp/blob/c4c8ee053f68d052f2287dd15dffcf42cc9e98a8/CMakeLists.txt#L321

bachittle:
hmm the ci format is still failing, any reason as to why?

bachittle:
so I ran the formatter manually (./tools/format.sh) and it seems to be working without issue. it seems that the diff is showing that the error occurs in buildifier, not even cmake-format, and is just saying that the file is changed (because it is) so doesn't seem like an error that should be throwing.

CMakeLists.txt
ThomsonTan: @bachittle ping on my comment here.

owent (approved):

CMakeLists.txt
bachittle: sorry I didn't see this, will take a look

bachittle: image
these are my cmake flags that are set using cmake-gui. notice the WITH_OTLP is not set, so I am just building the tests.

ThomsonTan: Thanks @bachittle. This does confirm that OTLP is not enabled. Instead of commenting out the set of `CMAKE_TOOLCHAIN_FILE`, can we add a line like below in the call after `find_package(GTest REQUIRED)` in the same file?

https://github.com/open-telemetry/opentelemetry-cpp/blob/504644380a9482258604ad644531c643937b216c/CMakeLists.txt#L331

bachittle: I tried this and it still failed, check latest commit: https://github.com/open-telemetry/opentelemetry-cpp/pull/1887/commits/2bd872bc1cbdab770936c61a8fbc492d65b89575. Let me know if I should try putting this somewhere else.

marcalff:
@bachittle 

Please fix the remaining CI failure about cmake-format.

This can be done by:

* (1) installing and running cmake-format,
* (2) or, copy and apply the patch given by the format failure in the build logs,
* (3) or even manually change the code just by reading the diff from the logs, which is what I personally do most of the times.

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c3c8b1e..b7cfe62 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,8 +247,8 @@ function(install_windows_deps)
   execute_process(
     COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/tools/setup-buildtools.cmd)
   set(CMAKE_TOOLCHAIN_FILE
-  ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
-  CACHE FILEPATH "")
+      ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
+      CACHE FILEPATH "")
   message(
     "Make sure that vcpkg.cmake is set as the CMAKE_TOOLCHAIN_FILE at the START of the cmake build process!
     Can be command-line arg (cmake -DCMAKE_TOOLCHAIN_FILE=...) or set in your editor of choice."
```

In this case, there are just two lines to adjust.

CMakeLists.txt
```diff
@@ -247,8 +247,13 @@ function(install_windows_deps)
   execute_process(
     COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/tools/setup-buildtools.cmd)
   set(CMAKE_TOOLCHAIN_FILE
-      ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
-      PARENT_SCOPE)
+    ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
+    CACHE FILEPATH "")
```
marcalff, line 251:
```suggestion
      ${CMAKE_CURRENT_SOURCE_DIR}/tools/vcpkg/scripts/buildsystems/vcpkg.cmake
      CACHE FILEPATH "")
```

marcalff (commented):
Almost there, cmake-format needs 2 more spaces.

CMakeLists.txt
bachittle: I should really just install cmake-format on my system, I normally don't use it in my other projects just clang-format :P

ThomsonTan (approved):

PR #1891, opened 2022-12-29
mac shlib dependencies fixes

ays7:
cmake tweaks to get shlibs to link on Mac with OTLP ON

### DIFF
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -17,7 +17,8 @@ target_include_directories(
 set(OPENTELEMETRY_OTLP_TARGETS opentelemetry_otlp_recordable)
 target_link_libraries(
   opentelemetry_otlp_recordable
-  PUBLIC opentelemetry_common opentelemetry_trace opentelemetry_resources opentelemetry_proto)
+  PUBLIC opentelemetry_common opentelemetry_trace opentelemetry_resources
+         opentelemetry_proto)
 
 if(WITH_LOGS_PREVIEW)
   target_link_libraries(opentelemetry_otlp_recordable PUBLIC opentelemetry_logs)
@@ -33,10 +34,11 @@ if(WITH_OTLP_GRPC)
                         PROPERTIES EXPORT_NAME otlp_grpc_client)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_client
-    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
+    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_ext
+           opentelemetry_proto)
 
   target_link_libraries(opentelemetry_exporter_otlp_grpc_client
-                        PUBLIC gRPC::grpc++)
+                        PRIVATE gRPC::grpc++)
   get_target_property(GRPC_INCLUDE_DIRECTORY gRPC::grpc++
                       INTERFACE_INCLUDE_DIRECTORIES)
   if(GRPC_INCLUDE_DIRECTORY)
@@ -60,8 +62,8 @@ if(WITH_OTLP_GRPC)
 
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc
-    PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+    PUBLIC opentelemetry_otlp_recordable opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS opentelemetry_exporter_otlp_grpc)
 
@@ -75,8 +77,8 @@ if(WITH_OTLP_GRPC)
 
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_log
-    PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+    PUBLIC opentelemetry_otlp_recordable opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS opentelemetry_exporter_otlp_grpc_log)
 
@@ -89,8 +91,8 @@ if(WITH_OTLP_GRPC)
 
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_metrics
-    PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+    PUBLIC opentelemetry_otlp_recordable opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS
        opentelemetry_exporter_otlp_grpc_metrics)
@@ -102,8 +104,8 @@ if(WITH_OTLP_HTTP)
                         PROPERTIES EXPORT_NAME otlp_http_client)
   target_link_libraries(
     opentelemetry_exporter_otlp_http_client
-    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_proto opentelemetry_http_client_curl
-           nlohmann_json::nlohmann_json)
+    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_proto
+           opentelemetry_http_client_curl nlohmann_json::nlohmann_json)
   if(nlohmann_json_clone)
     add_dependencies(opentelemetry_exporter_otlp_http_client
                      nlohmann_json::nlohmann_json)
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -17,7 +17,7 @@ target_include_directories(
 set(OPENTELEMETRY_OTLP_TARGETS opentelemetry_otlp_recordable)
 target_link_libraries(
   opentelemetry_otlp_recordable
-  PUBLIC opentelemetry_trace opentelemetry_resources opentelemetry_proto)
+  PUBLIC opentelemetry_common opentelemetry_trace opentelemetry_resources opentelemetry_proto)
 
 if(WITH_LOGS_PREVIEW)
   target_link_libraries(opentelemetry_otlp_recordable PUBLIC opentelemetry_logs)
@@ -33,7 +33,7 @@ if(WITH_OTLP_GRPC)
                         PROPERTIES EXPORT_NAME otlp_grpc_client)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_client
-    PUBLIC opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
+    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
 
   target_link_libraries(opentelemetry_exporter_otlp_grpc_client
                         PRIVATE gRPC::grpc++)
@@ -61,7 +61,8 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc
     PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+           opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS opentelemetry_exporter_otlp_grpc)
 
@@ -90,7 +91,8 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_metrics
     PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+           opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS
        opentelemetry_exporter_otlp_grpc_metrics)
@@ -102,7 +104,7 @@ if(WITH_OTLP_HTTP)
                         PROPERTIES EXPORT_NAME otlp_http_client)
   target_link_libraries(
     opentelemetry_exporter_otlp_http_client
-    PUBLIC opentelemetry_sdk opentelemetry_proto opentelemetry_http_client_curl
+    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_proto opentelemetry_http_client_curl
            nlohmann_json::nlohmann_json)
   if(nlohmann_json_clone)
     add_dependencies(opentelemetry_exporter_otlp_http_client
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -36,7 +36,7 @@ if(WITH_OTLP_GRPC)
     PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
 
   target_link_libraries(opentelemetry_exporter_otlp_grpc_client
-                        PRIVATE gRPC::grpc++)
+                        PUBLIC gRPC::grpc++)
   get_target_property(GRPC_INCLUDE_DIRECTORY gRPC::grpc++
                       INTERFACE_INCLUDE_DIRECTORIES)
   if(GRPC_INCLUDE_DIRECTORY)
@@ -61,8 +61,7 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc
     PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client
-    PRIVATE gRPC::grpc++)
+           opentelemetry_exporter_otlp_grpc_client)
 
   list(APPEND OPENTELEMETRY_OTLP_TARGETS opentelemetry_exporter_otlp_grpc)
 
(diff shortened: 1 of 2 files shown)


exporters/otlp/CMakeLists.txt
```diff
@@ -87,7 +88,8 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_metrics
     PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+           opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
```
owent, line 92: This exporters must not link gRPC::grpc++, there are symbols mixing/overlapping problem.(#1603 )

exporters/otlp/CMakeLists.txt
```diff
@@ -58,7 +58,8 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc
     PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+           opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
```
owent, line 62: This exporters must not link gRPC::grpc++, there are symbols mixing/overlapping problem.(#1603 )

owent (changes requested):
Only one target (`opentelemetry_exporter_otlp_grpc_client`) can link `gRPC::grpc++`, or there may be symbols mixing/overlapping problem.

exporters/otlp/CMakeLists.txt
ays7: @owent okey-dokey - left only `opentelemetry_exporter_otlp_grpc_client` with explicit link against `gRPC::grpc++`. That link can't be kept private though

exporters/otlp/CMakeLists.txt
```diff
@@ -30,10 +30,10 @@ if(WITH_OTLP_GRPC)
                         PROPERTIES EXPORT_NAME otlp_grpc_client)
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_client
-    PUBLIC opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
+    PUBLIC opentelemetry_common opentelemetry_sdk opentelemetry_ext opentelemetry_proto)
 
   target_link_libraries(opentelemetry_exporter_otlp_grpc_client
-                        PRIVATE gRPC::grpc++)
+                        PUBLIC gRPC::grpc++)
```
owent, line 36: Why change this to PUBLIC link here?This should be PRIVATE link or all targets depend opentelemetry_exporter_otlp_grpc_client will link `gRPC::grpc++`, which will cause the problem in #1603 .

owent (changes requested):

exporters/otlp/CMakeLists.txt
ays7: Sorry, struggling to reproduce #1603...
Before spending much time on it, let me guess: Is this about linking against static gRPC libs by a chance?

owent: > Sorry, struggling to reproduce #1603... Before spending much time on it, let me guess: Is this about linking against static gRPC libs by a chance?

This BUG can be repreduce when compiling gRPC as static library and compiling otel-cpp as dynamic library on Linux(With ELF ABI).
In this situation, the global variable will shared the same address between dynamic libraries but the initialization guards is private, which will construct these global variables more than onece.

lalitb: @ays7  - Do you have further comments on above discussion ?

ays7: Yes, I will try to look into static gRPC linking soon -- marked PR as WIP for now

ays7:
@owent I was able to reproduce #1603 with static gRPC libraries, however the fact that multiple opentelemetry libraries were linked against gRPC turns out to be a red-herring, the real culprit is OTEL's grpc tracer harness was defined in both `libopentelemetry_exporter_otlp_grpc` & `libopentelemetry_exporter_otlp_grpc_log`:
```
[root@b8ef01aee8b5 otlp]# nm libopentelemetry_exporter_otlp_grpc_log.so | c++filt | grep grpc_tracer_init
0000000000269570 T grpc_tracer_init(char const*)
00000000002693c0 T grpc_tracer_init()
0000000000188b03 t grpc_tracer_init() [clone .cold.4]
[root@b8ef01aee8b5 otlp]# nm libopentelemetry_exporter_otlp_grpc.so | c++filt | grep grpc_tracer_init
00000000002696f0 T grpc_tracer_init(char const*)
0000000000269540 T grpc_tracer_init()
0000000000188c83 t grpc_tracer_init() [clone .cold.4]
```
That was later resolved with move of common code into `libopentelemetry_exporter_otlp_grpc_client` 
```
[root@b8ef01aee8b5 otlp]# nm libopentelemetry_exporter_otlp_grpc.so | c++filt | grep grpc_tracer_init
[root@b8ef01aee8b5 otlp]# nm libopentelemetry_exporter_otlp_grpc_log.so | c++filt | grep grpc_tracer_init
[root@b8ef01aee8b5 otlp]# nm libopentelemetry_exporter_otlp_grpc_client.so | c++filt | grep grpc_tracer_init
0000000000245b70 T grpc_tracer_init(char const*)
00000000002459c0 T grpc_tracer_init()
000000000017626f t grpc_tracer_init() [clone .cold.4]
```
That said, you right, gRPC library link[s] better be kept private

exporters/otlp/CMakeLists.txt
```diff
@@ -75,8 +77,8 @@ if(WITH_OTLP_GRPC)
 
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_log
-    PUBLIC opentelemetry_otlp_recordable
-           opentelemetry_exporter_otlp_grpc_client)
+    PUBLIC opentelemetry_otlp_recordable opentelemetry_exporter_otlp_grpc_client
+    PRIVATE gRPC::grpc++)
```
owent, line 81: Why do we link `gRPC::grpc++` into `opentelemetry_exporter_otlp_grpc`, `opentelemetry_exporter_otlp_grpc_log` and `opentelemetry_exporter_otlp_grpc_metrics`? If we build otel-cpp as static libraries, they will link it just by link `opentelemetry_exporter_otlp_grpc_client`, and if we build otel-cpp as dynamic libraries, I think the global variables of gRPC are both in `opentelemetry_exporter_otlp_grpc_client`,  `opentelemetry_exporter_otlp_grpc`, `opentelemetry_exporter_otlp_grpc_log` and `opentelemetry_exporter_otlp_grpc_metrics`, the symbol overlap problem still exists.

ays7: 1- each one of `_otlp_grpc_` libraries makes direct use of gRPC APIs and gRPC has to be linked in for both gRPC header to be found [think: install gRPC in /foo/bar] and to satisfy shared library dependencies on mac in particular. This can be accomplished either via PUBLIC dependency inherited from another library or via explicit PRIVATE dependency. You kinda disagreed with a single common PUBLIC dependency and after reading https://cmake.org/pipermail/cmake/2016-May/063400.html I tend to agree that keeping this dependency private makes more sense

2- Cmake has smarts to actually link in static library [even PRIVATE] only if it was not linked into any dependencies. As you can see below `opentelemetry_exporter_otlp_grpc_metrics`, which depends on `opentelemetry_exporter_otlp_grpc_client` has much smaller set of linked libraries and specifically missing /usr/lib/libgrpc++.a:
_make VERBOSE=1 opentelemetry_exporter_otlp_grpc_client
...
/usr/bin/c++ -fPIC -O2 -g -DNDEBUG -shared -Wl,-soname,libopentelemetry_exporter_otlp_grpc_client.so -o libopentelemetry_exporter_otlp_grpc_client.so CMakeFiles/opentelemetry_exporter_otlp_grpc_client.dir/src/otlp_grpc_client.cc.o CMakeFiles/opentelemetry_exporter_otlp_grpc_client.dir/src/otlp_grpc_utils.cc.o  ../../libopentelemetry_proto.a **/usr/lib/libgrpc++.a** /usr/lib64/libabsl_bad_any_cast_impl.a /usr/lib64/libprotobuf.so /usr/lib/libgrpc.a /usr/lib64/libz.so /usr/lib/libcares.a /usr/lib/libaddress_sorting.a -pthread -lre2 /usr/lib/libupb.a /usr/lib64/libabsl_raw_hash_set.a /usr/lib64/libabsl_hashtablez_sampler.a /usr/lib64/libabsl_hash.a /usr/lib64/libabsl_city.a /usr/lib64/libabsl_low_level_hash.a /usr/lib64/libabsl_statusor.a /usr/lib64/libabsl_bad_variant_access.a /usr/lib/libgpr.a -ldl -lm -lrt /usr/lib64/libabsl_status.a /usr/lib64/libabsl_strerror.a /usr/lib64/libabsl_random_distributions.a /usr/lib64/libabsl_random_seed_sequences.a /usr/lib64/libabsl_random_internal_pool_urbg.a /usr/lib64/libabsl_random_internal_randen.a /usr/lib64/libabsl_random_internal_randen_hwaes.a /usr/lib64/libabsl_random_internal_randen_hwaes_impl.a /usr/lib64/libabsl_random_internal_randen_slow.a /usr/lib64/libabsl_random_internal_platform.a /usr/lib64/libabsl_random_internal_seed_material.a /usr/lib64/libabsl_random_seed_gen_exception.a /usr/lib64/libabsl_cord.a /usr/lib64/libabsl_bad_optional_access.a /usr/lib64/libabsl_cordz_info.a /usr/lib64/libabsl_cord_internal.a /usr/lib64/libabsl_cordz_functions.a /usr/lib64/libabsl_exponential_biased.a /usr/lib64/libabsl_cordz_handle.a /usr/lib64/libabsl_str_format_internal.a /usr/lib64/libabsl_synchronization.a /usr/lib64/libabsl_stacktrace.a /usr/lib64/libabsl_symbolize.a /usr/lib64/libabsl_debugging_internal.a /usr/lib64/libabsl_demangle_internal.a /usr/lib64/libabsl_graphcycles_internal.a /usr/lib64/libabsl_malloc_internal.a /usr/lib64/libabsl_time.a /usr/lib64/libabsl_strings.a /usr/lib64/libabsl_throw_delegate.a /usr/lib64/libabsl_int128.a /usr/lib64/libabsl_strings_internal.a /usr/lib64/libabsl_base.a -lpthread /usr/lib64/libabsl_raw_logging_internal.a /usr/lib64/libabsl_log_severity.a /usr/lib64/libabsl_spinlock_wait.a -lrt /usr/lib64/libabsl_civil_time.a /usr/lib64/libabsl_time_zone.a /usr/lib64/libssl.so /usr/lib64/libcrypto.so_

_make VERBOSE=1 opentelemetry_exporter_otlp_grpc_metrics
...
/usr/bin/c++ -fPIC -O2 -g -DNDEBUG -shared -Wl,-soname,libopentelemetry_exporter_otlp_grpc_metrics.so -o libopentelemetry_exporter_otlp_grpc_metrics.so CMakeFiles/opentelemetry_exporter_otlp_grpc_metrics.dir/src/otlp_grpc_metric_exporter.cc.o CMakeFiles/opentelemetry_exporter_otlp_grpc_metrics.dir/src/otlp_grpc_metric_exporter_factory.cc.o  -Wl,-rpath,/build/exporters/otlp:/build/sdk/src/trace:/build/sdk/src/logs:/build/sdk/src/resource:/build/sdk/src/metrics:/build/sdk/src/common: libopentelemetry_otlp_recordable.so **libopentelemetry_exporter_otlp_grpc_client.so** ../../sdk/src/trace/libopentelemetry_trace.so ../../sdk/src/logs/libopentelemetry_logs.so ../../sdk/src/resource/libopentelemetry_resources.so ../../sdk/src/metrics/libopentelemetry_metrics.so ../../sdk/src/common/libopentelemetry_common.so ../../libopentelemetry_proto.a /usr/lib64/libprotobuf.so /usr/lib64/libabsl_bad_variant_access.a /usr/lib64/libabsl_bad_any_cast_impl.a /usr/lib64/libabsl_city.a /usr/lib64/libabsl_base.a -lpthread /usr/lib64/libabsl_raw_logging_internal.a /usr/lib64/libabsl_log_severity.a /usr/lib64/libabsl_spinlock_wait.a -lrt_

lalitb: @ays7 - Can you reply to @owent  concerns here.

ays7: @owent `gRPC::grpc++` must be linked either directly or via shared dependency into all libs that make direct gRPC API calls to satisfy Mac shared library linker -- Mac shared libraries are linked like executables on linux -- all referenced symbols must be resolved to somewhere.
One way to achieve this is by exposing gPRC as a public dependency of `opentelemetry_exporter_otlp_grpc_client`, which is a common dependency of the 3 libs you mentioned -- that was my original suggestion.

For the static linking: here is my understanding: when linking shared libs on linux, linker first attempts to resolve referenced symbols to one of linked in shared libraries & if that fails to one of linked in static libraries -- in a later case code from the static library is copied into the output.

With that in mind, in the weird scenario of static gRPC/dynamic OTEL libraries mix, what happens is - bulk of gPRC library code is embedded into `opentelemetry_exporter_otlp_grpc_client` and all other otlp libs [which are dependent on `_client`] reference gRPC code from the `_client` .so  and not explicitly linked static gRPC library.

There is a potential issue if, say, both `opentelemetry_exporter_otlp_grpc_log` & `opentelemetry_exporter_otlp_grpc_metrics` call certain function that is not called inside `opentelemetry_exporter_otlp_grpc_client`  --  in that case both `_log` & `_metrics` may get a copy of some gRPC lib code. Fortunately there are no gRPC symbols like that at this point, but maybe it is indeed safer to declare gRPC lib as common PUBLIC dependency as I originally suggested?

owent: We can not link static gRPC into more than one dynamic libraries and executable. There will be more than one copy of some global variables in gRPC and some of them will be initialized more than once and some will not be initialized for ever.So the solution is just link gRPC into one dynamic library, that is `opentelemetry_exporter_otlp_grpc_client`, other libraries should only use the API from `opentelemetry_exporter_otlp_grpc_client` and can not use gRPC APIs directly.
Here is a simple sample to reproduce the problem in gRPC.

### `a.h`

```cpp
struct foo {
  int bar;
  foo();
  ~foo();

  void print(const char*);

  static foo _;
};

void print_static_global(const char*);
```

### `a.cpp`

```cpp
#include "a.h"

#include <iostream>
#include <thread>
#include <chrono>
#include <memory>

struct bar {
  foo* ptr = nullptr;
};

foo foo::_;
static bar s_;

foo::foo(): bar(127) {
    s_.ptr = this;
    std::cout<< "construct "<< this<< std::endl;
}

foo::~foo() {
    std::cout<< "destroy "<< this<< std::endl;
}

void foo::print(const char* prefix) {
    std::cout<< prefix<< "-foo: "<< this<< ": "<< bar<< std::endl;
}

void print_static_global(const char* prefix) {
    foo::_.print(prefix);
    std::cout<< prefix<< "-piblic API bar: "<< s_.ptr<< std::endl;
}

std::shared_ptr<std::thread> g_t(new std::thread([]() {
  std::this_thread::sleep_for(std::chrono::seconds{1});
  std::cout<< "internal API bar: "<< s_.ptr<< std::endl;
}), [](std::thread* thd) {
  thd->join();
  delete thd;
});
```

### `b.cpp`

```cpp
#include "a.h"

void dll_func_b() {
    print_static_global("b");
}
```

### `c.cpp`

```cpp
#include "a.h"

void dll_func_b();

int main() {
  print_static_global("c");
  dll_func_b();
  return 0;
}
```

### Compile and run

```bash
[owent@VM-144-59-centos test]$ clang++ a.cpp -o libtest_a.a -c -fPIC -pthread 
[owent@VM-144-59-centos test]$ clang++ b.cpp -o libtest_b.so -shared -fPIC -L$PWD -ltest_a -pthread
[owent@VM-144-59-centos test]$ clang++ c.cpp -fPIC -L$PWD -ltest_b -ltest_a '-Wl,-rpath=$ORIGIN' -pthread
[owent@VM-144-59-centos test]$ ./a.out 
[owent@VM-144-59-centos test]$ ./a.out 
construct 0x55af97755338
construct 0x55af97755338
c-foo: 0x55af97755338: 127
c-piblic API bar: 0x55af97755338
b-foo: 0x55af97755338: 127
b-piblic API bar: 0x55af97755338
internal API bar: (nil)
internal API bar: 0x55af97755338
destroy 0x55af97755338
destroy 0x55af97755338
[owent@VM-144-59-centos test]$
```

> `$ORIGIN` may be replaced by `@loader_path` on macOS.


As you see, the global variable `0x55af97755338` is constructed twice and destroyed twice,And the internal API use different `static bar s_;`. I didn't test it on macOS but i guess it has the same problem.It's more complicated in gRPC and it will make gRPC do not work in some dynamic libraries.

Actually, if we set gRPC as private dependency of `opentelemetry_exporter_otlp_grpc_client`, and when we build `opentelemetry_exporter_otlp_grpc_client` into a static library. All the targets will links gRPC automatically by just link  `opentelemetry_exporter_otlp_grpc_client`.

ays7: @owent Thanks for the test - I will play with it soon.
In a mean time here is another idea: anything particularly wrong with combining all `_otlp_grpc_` libraries into a single `opentelemetry_exporter_otlp_grpc` library?

owent: > @owent Thanks for the test - I will play with it soon. In a mean time here is another idea: anything particularly wrong with combining all `_otlp_grpc_` libraries into a single `opentelemetry_exporter_otlp_grpc` library?

I prefer spilt modules one by one personally, the build system may built all libraries but most users will use only part of them.

*It will cost more memory and CPU if we pack all codes into one library, we have a project which will cost about 8+GB memory to link every executable in this project and it's easy OOM when we link them concurrency.We have done many jobs to split dependencies into small libraries to solve this problem.*

> And one more thing: if static gRPC library can only be linked into one DLL then any application that use gRPC [either directly or indirectly via, say, third party dependency] will be in conflict with `otlp_grpc` libraries. Which logically means that use of static gRPC libraries is a bad idea. Right?

Yes, in my understanding, it's safer to use static or dynamic libraries for all dependencies, not mixed. There were issues report this problem when only use otlp-cpp, so I think the only thing we can do is keep it right when users only use and link otlp-cpp’s SDK.

lalitb: @ays7  - Do you have any questions on above comments?

ays7: @lalitb Sorry, I don't think I see a way out: static linking gRPC is a bad idea one way or another. IMHO.
Looks like @owent is exploring some alternatives with PR #2005  -- perhaps this PR can be abandoned [as long as #2005 fixes mac build too]

esigo:
closing as discussed above.

PR #1905, opened 2023-01-07
sdk::resource::Resource::Merge should be const

### LINKED ISSUE #1898: opentelemetry::sdk::resource::Resource::Merge should be const method
This method https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/resource/resource.cc#L20 does not modify the object itself, so it should be `const`

Not a blocker, just had to rewrap it in my version (Windows, dll) and had to create shim for it, so ended up wondering why is Merge taking non-const "this" Resource, and returning one, then checked the source code and saw no actual changes are done (I could've missed something, but it's easy to follow).

esigo:
Fixes #1898 (issue)

## Changes
changes `sdk::resource::Resource::Merge` to const.

### DIFF
--- a/sdk/include/opentelemetry/sdk/resource/resource.h
+++ b/sdk/include/opentelemetry/sdk/resource/resource.h
@@ -40,7 +40,7 @@ public:
    * @returns the newly merged Resource.
    */
 
-  Resource Merge(const Resource &other) noexcept;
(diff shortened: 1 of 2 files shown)


lalitb (approved):
LGTM. Thanks for the fix.

marcalff (approved):
LGTM

PR #1909, opened 2023-01-09
Opentracing shim

### LINKED ISSUE #78: Implement OpenTracing shim
Refer to https://github.com/open-telemetry/opentelemetry-python/tree/v1.10.0/shim/opentelemetry-opentracing-shim.

chusitoo:
Fixes #78, answers to #1180

## Changes

* Implement Opentracing shim as described in https://opentelemetry.io/docs/reference/specification/compatibility/opentracing/
* Unit tests are a combination of legacy Opentracing tests as well as new tests mimicking the Java and JS counterparts
* Add CMake support (turned off by default) for externally linked Opentracing library or use local submodule
* Add basic build support with Bazel. Windows is unsupported due to Opentracing lacking build support for the platform.


owent:
Is it better to move this into a standalone directory?

opentracing-shim/include/opentelemetry/opentracingshim/propagation.h
```diff
@@ -0,0 +1,96 @@
...
+template<typename T, nostd::enable_if_t<std::is_base_of<opentracing::TextMapWriter, T>::value, bool> = true>
+class CarrierWriterShim : public opentelemetry::context::propagation::TextMapCarrier
+{
+public:
+  CarrierWriterShim(const T& writer) : writer_(writer) {}
+
+  // returns the value associated with the passed key.
+  virtual nostd::string_view Get(nostd::string_view key) const noexcept override
+  {
+    return "";
```
ThomsonTan, line 28: Is returning empty value expected here?

CMakeLists.txt
```diff
@@ -497,6 +499,29 @@ include_directories(api/include)
 
 add_subdirectory(api)
 
+if(WITH_OPENTRACING)
```
ThomsonTan, line 502: Perhaps add a CI job for this.

chusitoo: Sure, I think it would fit well in the cmake.maintainer.test, would that be acceptable?

opentracing-shim/include/opentelemetry/opentracingshim/propagation.h
chusitoo: Yes, this carrier wraps around an Opentracing TextMapWriter, which is essentially an interface with a single Set method, so it should never require a Get method in the same situation. I'll add a comment for clarity.

CMakeLists.txt
```diff
@@ -482,10 +482,13 @@ if(WITH_OPENTRACING)
     set(OPENTRACING_DIR "third_party/opentracing-cpp")
     message("Trying to use local ${OPENTRACING_DIR} from submodule")
     if(EXISTS "${PROJECT_SOURCE_DIR}/${OPENTRACING_DIR}/.git")
+      set(SAVED_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+      set(CMAKE_CXX_FLAGS OFF)
```
owent: Why do we need set `CMAKE_CXX_FLAGS` to `OFF` here?
If we use `-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0` , I think it's also need be passed into all targets.

CMakeLists.txt
```diff
@@ -474,6 +476,32 @@ include_directories(api/include)
...
 add_subdirectory(api)
 
+if(WITH_OPENTRACING)
+  find_package(OpenTracing CONFIG QUIET)
+  if(NOT OpenTracing_FOUND)
+    set(OPENTRACING_DIR "third_party/opentracing-cpp")
+    message("Trying to use local ${OPENTRACING_DIR} from submodule")
+    if(EXISTS "${PROJECT_SOURCE_DIR}/${OPENTRACING_DIR}/.git")
+      set(SAVED_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+      set(CMAKE_CXX_FLAGS "")
```
owent, line 486: Why do we need set `CMAKE_CXX_FLAGS` to `""` here?
If we use `-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0` , I think it's also need be passed into all targets.

chusitoo: The main issue I encountered with using the passed down CXX flags is that the _third-party/opentracing-cpp_ build will fail due to `-Werror=redundant-move`, as observed in this run: https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/3915878675/jobs/6694688717

I also considered disabling BUILD_MOCKTRACER which would effectively get rid of this compilation error and is arguably a much cleanerapproach, but I find it is more of a band aid than a fix, since any other flags that get added in the future might still interfere with the core opentracing-cpp build.

CMakeLists.txt
chusitoo: Done

CMakeLists.txt
owent: I think it's better to disable this warning when include opentracing-cpp headers or add `-Wno-error=redundant-move` when include opentracing-cpp's headers.
Just reset `CMAKE_CXX_FLAGS` may break the ABI.

chusitoo: Done. Thanks for the advise.

bazel/repository.bzl
```diff
@@ -179,6 +179,17 @@ def opentelemetry_cpp_deps():
...
     )
 
+    # Opentracing
+    maybe(
+        http_archive,
+        name = "com_github_opentracing",
+        sha256 = "5b170042da4d1c4c231df6594da120875429d5231e9baa5179822ee8d1054ac3",
+        strip_prefix = "opentracing-cpp-1.6.0",
+        urls = [
+            "https://github.com/opentracing/opentracing-cpp/archive/refs/tags/v1.6.0.tar.gz",
```
lalitb, line 189: also add this dependency in [third_party_release](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release) file, and [dependencies.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/dependencies.md) with license info.

ci/do_ci.sh
```diff
@@ -282,15 +292,17 @@ elif [[ "$1" == "bazel.legacy.test" ]]; then
   exit 0
 elif [[ "$1" == "bazel.noexcept" ]]; then
   # there are some exceptions and error handling code from the Prometheus and Jaeger Clients
+  # as well as Opentracing shim (due to some third party code in its Opentracing dependency)
   # that make this test always fail. ignore Prometheus and Jaeger exporters in the noexcept here.
-  bazel $BAZEL_STARTUP_OPTIONS build --copt=-fno-exceptions --build_tag_filters=-jaeger $BAZEL_OPTIONS_ASYNC -- //... -//exporters/prometheus/... -//exporters/jaeger/... -//examples/prometheus/... -//sdk/test/metrics:attributes_hashmap_test
-  bazel $BAZEL_STARTUP_OPTIONS test --copt=-fno-exceptions --build_tag_filters=-jaeger $BAZEL_TEST_OPTIONS_ASYNC -- //... -//exporters/prometheus/... -//exporters/jaeger/... -//examples/prometheus/... -//sdk/test/metrics:attributes_hashmap_test
+  bazel $BAZEL_STARTUP_OPTIONS build --copt=-fno-exceptions --build_tag_filters=-jaeger $BAZEL_OPTIONS_ASYNC -- //... -//exporters/prometheus/... -//exporters/jaeger/... -//examples/prometheus/... -//sdk/test/metrics:attributes_hashmap_test -//opentracing-shim/...
```
lalitb, line 297: nit - Just wondering if we can use `dynamic_cast` if OPENTELEMETRY_RTTI_ENABLED is enabled, and use `static_cast` otherwise ?

lalitb (commented):
Changes in general look good to me in general, but that's me with very limited experience with Opentracing. Have few nit cosmetic comments.

.github/workflows/ci.yml
```diff
@@ -131,6 +131,20 @@ jobs:
         sudo ./ci/install_abseil.sh
         ./ci/do_ci.sh cmake.abseil.test
 
+  cmake_opentracing_shim_test:
+    name: CMake test (with opentracing-shim)
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
chusitoo: Sure, should not be a problem, will update to latest Ubuntu.

ci/do_ci.sh
chusitoo: One inconvenience with static_cast is that I do not know if the downcasting succeeds, so I cannot simply assume that I have a SpanContextShim as this would be undefined behavior. My idea is using a 64bit tag to disambiguate instances of the shim class which will make it a bit safer to call `context()` and `baggage()`, both of which are methods specific to shim only.

opentracing-shim/test/shim_utils_test.cc
```diff
@@ -0,0 +1,280 @@
...
+  ASSERT_EQ(value.index(), common::AttributeType::kTypeInt64);
+  ASSERT_EQ(nostd::get<int64_t>(value), 42l);
+
+  value = shim::utils::attributeFromValue(55ul);
+  ASSERT_EQ(value.index(), common::AttributeType::kTypeUInt64);
+  ASSERT_EQ(nostd::get<uint64_t>(value), 55ul);
+
+  value = shim::utils::attributeFromValue(std::string{"a string"});
+  ASSERT_EQ(value.index(), common::AttributeType::kTypeString);
+  ASSERT_EQ(nostd::get<nostd::string_view>(value), nostd::string_view{});
```
gregoryfranklin, line 77: `std::string{"a string"}` should equal `nostd::string_view{"a string"}` rather than `nostd::string_view{}`?

chusitoo: Agree, an overlook on my side. I had temporarily removed this mapping since string_view doesn't own the underlying data from the passed in string and would be flagged as undefined behavior by the address sanitizer check if passed by const reference.

marcalff:
The copyright check in CI is new, please fix:

```
Missing copyright in ./opentracing-shim/BUILD
Missing license in ./opentracing-shim/BUILD
Missing copyright in ./opentracing-shim/CMakeLists.txt
Missing license in ./opentracing-shim/CMakeLists.txt
```

marcalff (commented):
@chusitoo 

If I understand correctly, the Opentracing shim introduces a dependency on the [opentracing-cpp](https://github.com/opentracing/opentracing-cpp) repository, which is a deprecated library, per:

https://github.com/opentracing/specification/issues/163

It is a bit unusual to add dependencies like that, but it makes sense in this case, so ok.

Not knowing enough of Opentracing, it is a bit hard to review the code for correctness for every details.

That said, the overall code looks ok.

The contributed code is protected by a `WITH_OPENTRACING` option, and is separate from the opentelemetry-cpp code which is otherwise unchanged, so there is very little technical risks with this PR, in my opinion.

`SpanContextShim::extractFrom()` has a hack with `kUniqueTag`, but after looking at it I don't have a better proposal to offer, so ok with it.

Please fix the prometheus-cpp spurious change, and I will approve this PR.

Great work.

marcalff (approved):

lalitb (approved):
Sorry for responding late on it. Changes in general look good, and clean approach  - the shim confined to separate directory, and behind feature flag.
Most of the existing contributors doesn't have expertise on Opentracing, and bandwidth to manage this component, so @chusitoo it would be appreciated if you can own any related issues/changes going ahead :)

chusitoo:
> Sorry for responding late on it. Changes in general look good, and clean approach - the shim confined to separate directory, and behind feature flag. Most of the existing contributors doesn't have expertise on Opentracing, and bandwidth to manage this component, so @chusitoo it would be appreciated if you can own any related issues/changes going ahead :)

@lalitb Thanks for the review. Not a problem, I am not an expert in OpenTracing myself but I can support any related issues or requests related to the shim.

PR #1916, opened 2023-01-10
Cleanup CMake makefiles for CURL usage

### LINKED ISSUE #1913: [BUILD] CMake calls find_package(CURL) in multiple places
Per the CMake makefiles:

```
./examples/http/CMakeLists.txt:find_package(CURL)
./exporters/otlp/CMakeLists.txt:  find_package(CURL REQUIRED)
./exporters/zipkin/CMakeLists.txt:find_package(CURL REQUIRED)
./ext/src/http/client/curl/CMakeLists.txt:find_package(CURL)
./ext/test/http/CMakeLists.txt:find_package(CURL)
./ext/test/w3c_tracecontext_test/CMakeLists.txt:find_package(CURL)
./CMakeLists.txt:    find_package(CURL)
```

CMake should not call find_package(CURL) so many times from nested directories.

Beside, the current logic is questionable:

```
find_package(CURL)
if(NOT CURL_FOUND)
  message(WARNING "Skipping example_w3c_tracecontext_test: CURL not found")
else()
```

If `CURL` is not found and `BUILD_W3CTRACECONTEXT_TEST` is set, the build should fail, not pass and do nothing.

Instead:

- Evaluate once **IF** CURL is needed, similar to `USE_NLOHMANN_JSON`
- Call find_package(CURL) once from the top level CMakeLists.txt
- Assume CURL_FOUND and related variables are already evaluated, in sub directories.

Implementing this cleanup will help to add the CURL dependency from different sources (git submodule, external project, alternate location).

marcalff:
Fixes #1913 

## Changes

General CMake cleanup related to CURL:

- Removed multiple `find_package(CURL)`, used `find_package(CURL)` from only one place in the top level makefile,
- Removed `if(CURL_FOUND)` tests, which are semantically incorrect. A given library should be compiled when there is a reason to build it, not because CURL happens to be available. This also make the build more robust in a super build, when unrelated packages can use `CURL` and set `CURL_FOUND` for unrelated reasons,
- Fixed the W3C trace test build, which did not honor the build option,
- Fixed the `WITH_OTLP_HTTP` option to be predictable,
- Added a build option for HTTL client/server examples, there was previously no way to control the build of these artifacts.

See the CHANGELOG Important changes section for more details.


CMakeLists.txt
```diff
@@ -195,6 +202,9 @@ option(
   OFF)
 option(WITH_EXAMPLES "Whether to build examples" ON)
 
+# This requires CURL, OFF by default.
+option(WITH_EXAMPLES_HTTP "Whether to build http client/server examples" OFF)
```
owent, line 206: Do you think it's better to set `WITH_EXAMPLES_HTTP` to `ON` when `WITH_HTTP_CLIENT_CURL` is `ON` or CURL is found?

marcalff: @owent 

Thanks for the comments.

The dependency is:

- WITH_EXAMPLES_HTTP
  - requires WITH_HTTP_CLIENT_CURL
    - requires CURL

so the configuration is evaluated from the top (what the user wants) to the bottom (what low level packages are needed to satisfy that).

What you suggest is to take the opposite route (bottom up), and to build some parts (WITH_EXAMPLES_HTTP) even when the user did not asked for it in the makefile, just because it is possible (CURL present)

The bottom up approach leads to unpredictable behavior:

- if another part uses CURL for something else, why would suddenly the http examples be built ?
- WITH_ZIPKIN will build http client/server examples as a side effect ?

Taking the bottom up approach, unpredictable items could be compiled, then packaged, and then installed, even when nobody asked for them.

I think the top down logic is more robust.

owent: Thanks for your explanation. In my understanding, if we set `WITH_EXAMPLES` to `ON` by default, I will espect all examples will be built.
That's fine by me to exclude http examples by default, I just thought it may lead to some misunderstanding.

owent (approved):
LGTM

ci/do_ci.sh
```diff
@@ -193,6 +194,7 @@ elif [[ "$1" == "cmake.exporter.otprotocol.test" ]]; then
   rm -rf *
   cmake -DCMAKE_BUILD_TYPE=Debug  \
         -DWITH_OTLP=ON \
+        -DWITH_OTLP_HTTP=ON \
```
owent, line 197: Do we need also add `-DWITH_EXAMPLES_HTTP=ON` in at least one CI job? Or we can not ensure the http example can be compiled success.

owent (commented):
Do we need also add `-DWITH_EXAMPLES_HTTP=ON` in at least one CI job? Or we can not ensure the http example can be compiled success.

marcalff:
> Do we need also add `-DWITH_EXAMPLES_HTTP=ON` in at least one CI job? Or we can not ensure the http example can be compiled success.

Good point, looking into this.

marcalff:
> > Do we need also add `-DWITH_EXAMPLES_HTTP=ON` in at least one CI job? Or we can not ensure the http example can be compiled success.
> 
> Good point, looking into this.

Resolved, tests build in maintainer mode.

CMakeLists.txt
```diff
@@ -314,22 +324,49 @@ if(WITH_OTLP)
...
+else()
+  set(WITH_HTTP_CLIENT_CURL OFF)
+endif()
+
+#
+# Do we need CURL ?
+#
+
+if(WITH_HTTP_CLIENT_CURL)
+  set(USE_CURL ON)
```
lalitb, line 356: Do we need `USE_CURL`, or just `WITH_HTTP_CLIENT_CURL` is enough ?

CMakeLists.txt
```diff
@@ -148,6 +151,10 @@ if(WITH_STL)
 endif()
 
 option(WITH_OTLP "Whether to include the OpenTelemetry Protocol in the SDK" OFF)
+
+option(WITH_OTLP_HTTP "Whether to include the OTLP http exporter in the SDK"
+       OFF)
```
lalitb, line 156: So to confirm, `WITH_OTLP_HTTP` is now a valid CMake option for command-line, but `WITH_OTLP_GRPC` is not. And this needs to be fixed as separate bug?

CMakeLists.txt
```diff
@@ -195,6 +202,9 @@ option(
   OFF)
 option(WITH_EXAMPLES "Whether to build examples" ON)
 
+# This requires CURL, OFF by default.
+option(WITH_EXAMPLES_HTTP "Whether to build http client/server examples" OFF)
+
```
lalitb, line 207: If `WITH_EXAMPLES` is off, `WITH_EXAMPLES_HTTP` will not be entertained even if CURL is installed ?

ext/src/CMakeLists.txt
```diff
@@ -2,4 +2,6 @@ if(WITH_ZPAGES)
   add_subdirectory(zpages)
 endif()
 
-add_subdirectory(http/client/curl)
+if(WITH_HTTP_CLIENT_CURL)
+  add_subdirectory(http/client/curl)
+endif()
```
lalitb, line 7: Not related to this PR, but while you are on it, can you add the copyright header here too :)

examples/CMakeLists.txt
```diff
@@ -24,4 +24,7 @@ add_subdirectory(batch)
 add_subdirectory(metrics_simple)
 add_subdirectory(multithreaded)
 add_subdirectory(multi_processor)
-add_subdirectory(http)
+
+if(WITH_EXAMPLES_HTTP)
+  add_subdirectory(http)
+endif()
```
lalitb, line 30: Not related to this PR, but while you are on it, can you add the copyright header here too :)

exporters/zipkin/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 include_directories(include)
-find_package(CURL REQUIRED)
```
lalitb, line 16: Not related to this PR, but while you are on it, can you modify the copyright header here to be consistent with other places.

lalitb (approved):
LGTM with nit comments :)

esigo (approved):
LGTM
Thanks for the fix :)

CMakeLists.txt
marcalff: The two are identical because only WITH_HTTP_CLIENT_CURL requires CURL. Simplified the makefile.

CMakeLists.txt
marcalff: Thanks.

Implemented a check to make sure WITH_EXAMPLE=ON when WITH_EXAMPLES_HTTP=ON.

examples/CMakeLists.txt
marcalff: Done.

exporters/zipkin/CMakeLists.txt
marcalff: Done.

ext/src/CMakeLists.txt
marcalff: Done.

marcalff:
@owent @lalitb @esigo 

Thanks all for the review.
All comments addressed, now ready to merge.

PR #1918, opened 2023-01-11
Make macros.h available for all source files via version.h

ThomsonTan:
## Changes

As the macros defined in `macros.h` is general for all sources, including it in `version.h` to make it available for all sources (no need for separate `#include "opentelemetry/common/macros.h"` statement.

### DIFF
--- a/api/include/opentelemetry/common/macros.h
+++ b/api/include/opentelemetry/common/macros.h
@@ -3,8 +3,6 @@
 
 #pragma once
 
-#include <cstdint>
-
 #if !defined(OPENTELEMETRY_LIKELY_IF) && defined(__cplusplus)
 // GCC 9 has likely attribute but do not support declare it at the beginning of statement
 #  if defined(__has_cpp_attribute) && (defined(__clang__) || !defined(__GNUC__) || __GNUC__ > 9)
@@ -90,10 +88,9 @@
 
 // Regex support
 #if (__GNUC__ == 4 && (__GNUC_MINOR__ == 8 || __GNUC_MINOR__ == 9))
-#  define HAVE_WORKING_REGEX 0
+#  define OPENTELEMETRY_HAVE_WORKING_REGEX 0
 #else
-#  include <regex>
-#  define HAVE_WORKING_REGEX 1
+#  define OPENTELEMETRY_HAVE_WORKING_REGEX 1
 #endif
 
 /* clang-format off */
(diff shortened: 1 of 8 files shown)


marcalff (commented):
Making this change will expose:

- the `cstdint` header file
- the `regex` headerfile
- the `HAVE_WORKING_REGEX` symbol, which is not using an `OPENTELEMETRY_` prefix and therefore pollutes the build

for any user of any api header file.

I would prefer to apply the include-what-you-use principles, and only include macro.h in the few places where it is needed.

The API header files are stable by now, so there is no real maintenance burden and breaks caused by missing includes on macro.h

What is the motivation or rationale for this change ?

ThomsonTan:
> Making this change will expose:
> 
> * the `cstdint` header file
> * the `regex` headerfile
> * the `HAVE_WORKING_REGEX` symbol, which is not using an `OPENTELEMETRY_` prefix and therefore pollutes the build
> 
> for any user of any api header file.
> 
> I would prefer to apply the include-what-you-use principles, and only include macro.h in the few places where it is needed.
> 
> The API header files are stable by now, so there is no real maintenance burden and breaks caused by missing includes on macro.h
> 
> What is the motivation or rationale for this change ?

Perhaps we should remove cstdint from macro here? `HAVE_WORKING_REGEX` is like any definition in `config.h` which is fine and convenient to be included for every source file under the repo?

The intention was the make the all the generic macros in a common header file for all source files. I am adding some declspec macro for building DLL, so I'd like to avoid including it for every header (there could be many of them).

ThomsonTan:
Perhaps move `#include <regex>` to the source file which need it instead of putting into this generic header file?

owent (approved):
LGTM

marcalff (approved):
LGTM with the following changes, please consider to:

- rename HAVE_WORKING_REGEX to OPENTELEMETRY_HAVE_WORKING_REGEX, everywhere (including in the SDK)
- only set OPENTELEMETRY_HAVE_WORKING_REGEX based on the GCC version, but do not include regex.h when setting the flag
- include (conditionally) regex.h when using the flag
- remove the code duplication from trace_state.h

With this, all uses should look like:

```
#if OPENTELEMETRY_HAVE_WORKING_REGEX
#  include <regex>
#endif
```

I know this is an existing issue, but cleaning it up before exposing macro.h is needed, to avoid causing side effects on user code.

No strong opinion on whether to put the code to set  OPENTELEMETRY_HAVE_WORKING_REGEX in macro.h or config.h

Thanks for the fix.

marcalff:
> Perhaps move `#include <regex>` to the source file which need it instead of putting into this generic header file?

Good idea, see my latest review comments.

PR #1929, opened 2023-01-18
Fix warning for misconfiguration of PeriodicExportingMetricReader

jwbuurlage:
## Changes

The periodically exporting metric reader requires the timeout field to be strictly less than the interval field. However, the warning that is logged out gives the opposite requirement. The existing logic in the code is correct.


lalitb (approved):
Thanks for the fix.

marcalff (approved):
LGTM, thanks for the fix.

PR #1932, opened 2023-01-19
Build OpenTelemetry C++ SDK and exporter into DLL

### LINKED ISSUE #1105: Support building DLL for windows
opentelemetry-cpp uses bazel and cmake to build the required targets. These targets are by default created as static libraries. 
To create shared libraries / DLLs, we can just add `-DBUILD_SHARED_LIBS=ON`  to cmake, and `linkstatic=False` flag in cc_library/cc_binary rule in bazel.
This works fine for building shared libraries(.so) on Unix-like systems but doesn't work by default for DLL on Windows. This needs code changes to define export/import declarations for all exporter API and classes. As each DLL on Windows will have its own heap for memory usage and symbols. 

More discussions from the slack channel here (Need to be part of cloud-native.slack.com Org to view the discussions ):
https://cloud-native.slack.com/archives/C01N3AT62SJ/p1626285244072700
https://cloud-native.slack.com/archives/C01N3AT62SJ/p1637665049041200

While there is no immediate plan to support this requirement, unless there is enough traction on this issue, and contributors are ready to add the support.

ThomsonTan:
Fixes #1105 

## Changes


lalitb:
Thanks @ThomsonTan . 

@meastp, @owent   - It would be really helpful whenever you have time to review this.

ThomsonTan:
Yes, it should resolve the issues in #1105. I still need some minor change on this draft, but feel free to take a look as the major change has been verified.


> Thanks @ThomsonTan . Just to be clear, is this going to resolve the #1105 ?
> 
> @meastp, @owent - It would be really helpful whenever you have time to review this.

marcalff (commented):
In:

```
  class OPENTELEMETRY_API TracerProviderFactory
```

It feels strange to see "API" in the name, for a class from the SDK.

IMPORT / EXPORT is orthogonal to api/sdk/etc

Rename to `OPENTELEMETRY_EXPORT`, which will apply to any api / sdk / exporter classes ?

ThomsonTan:
> In:
> 
> ```
>   class OPENTELEMETRY_API TracerProviderFactory
> ```
> 
> It feels strange to see "API" in the name, for a class from the SDK.
> 
> IMPORT / EXPORT is orthogonal to api/sdk/etc
> 
> Rename to `OPENTELEMETRY_EXPORT`, which will apply to any api / sdk / exporter classes ?

I actually named the macro as like `OPENTELEMETRY_EXPORT_IMPORT`, but then changed to `OPENTELEMETRY_API` because all the classes/functions marked as by this implies that they are exported to user to reference/invoke. So `OPENTELEMETRY_API` here all the OpenTelemetry APIs and a few selected SDK method we want to expose to user to initialize SDK. But we can discuss the name here.

lalitb:
> > In:
> > ```
> >   class OPENTELEMETRY_API TracerProviderFactory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It feels strange to see "API" in the name, for a class from the SDK.
> > IMPORT / EXPORT is orthogonal to api/sdk/etc
> > Rename to `OPENTELEMETRY_EXPORT`, which will apply to any api / sdk / exporter classes ?
> 
> I actually named the macro as like `OPENTELEMETRY_EXPORT_IMPORT`, but then changed to `OPENTELEMETRY_API` because all the classes/functions marked as by this implies that they are exported to user to reference/invoke. So `OPENTELEMETRY_API` here all the OpenTelemetry APIs and a few selected SDK method we want to expose to user to initialize SDK. But we can discuss the name here.

Don't have strong opinion on naming, but OPENTELEMETRY_EXTERN or OPENTELEMETRY_WIN_EXTERN can be considered too.

sdk/include/opentelemetry/sdk/trace/simple_processor_factory.h
```diff
@@ -17,7 +17,7 @@ namespace trace
 /**
  * Factory class for SimpleSpanProcessor.
  */
-class SimpleSpanProcessorFactory
+class OPENTELEMETRY_API SimpleSpanProcessorFactory
```
lalitb, line 20: Should we export BatcSpanProcessorFactory, TracerContextFactory and SpanExporter classes too from tracer ?

ThomsonTan:
> 

Looked at some Chromium code, 

> > > In:
> > > ```
> > >   class OPENTELEMETRY_API TracerProviderFactory
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > It feels strange to see "API" in the name, for a class from the SDK.
> > > IMPORT / EXPORT is orthogonal to api/sdk/etc
> > > Rename to `OPENTELEMETRY_EXPORT`, which will apply to any api / sdk / exporter classes ?
> > 
> > 
> > I actually named the macro as like `OPENTELEMETRY_EXPORT_IMPORT`, but then changed to `OPENTELEMETRY_API` because all the classes/functions marked as by this implies that they are exported to user to reference/invoke. So `OPENTELEMETRY_API` here all the OpenTelemetry APIs and a few selected SDK method we want to expose to user to initialize SDK. But we can discuss the name here.
> 
> Don't have strong opinion on naming, but OPENTELEMETRY_EXTERN or OPENTELEMETRY_WIN_EXTERN can be considered too, if we want to avoid API and EXPORT keywords :)

Looked some Chromium code and pattern, seems naming it as *_EXPORT makes sense as suggested by @marcalff. See https://chromium.googlesource.com/chromium/src.git/+/master/base/base_export.h

CMakeLists.txt
```diff
@@ -513,6 +513,13 @@ endif()
 
 include(CMakePackageConfigHelpers)
 
+if(DEFINED OPENTELEMETRY_BUILD_DLL)
```
owent, line 516: Why not use `BUILD_SHARED_LIBS`?

ThomsonTan: `BUILD_SHARED_LIBS` will build each `add_library` as a DLL which produces a lot of DLL. I'd like to consolidate all the DLLs into 1 with only OpenTelemetry API and selected functions exported.

api/include/opentelemetry/common/macros.h
```diff
@@ -175,3 +175,24 @@ point.
 #  define OPENTELEMETRY_API_SINGLETON
 
 #endif
+
+//
+// The if/elif order matters here. If both OPENTELEMETRY_BUILD_IMPORT_DLL and
+// OPENTELEMETRY_BUILD_EXPORT_DLL are defined, the former takes precedence.
+//
+#if defined(_MSC_VER) && defined(OPENTELEMETRY_BUILD_IMPORT_DLL)
```
owent, line 183: I think we can declare `OPENTELEMETRY_EXPORT` just like `DLL_PUBLIC` in https://gcc.gnu.org/wiki/Visibility#How_to_use_the_new_C.2B-.2B-_visibility_support for better compatibility.
We meet some users add `-fvisibility=hidden` to `CXXFLAGS` environment variable before and it will prevent otel-cpp API to be shared between modules.
And also, add `__attribute__ ((visibility ("default")))` into declaration will also make RTTI codes public, or the virtual base classes can not be used by external codes when compiling otel-cpp with `-fvisibility=hidden`.

CMakeLists.txt
owent: Thanks for the explaination.

api/include/opentelemetry/common/macros.h
ThomsonTan: Is this for supporting DLL with cygwin/gcc? I thought about this, but as there is some extra work needed to make it work like exporting the unreferenced symbols explicitly. As it is not been tested, probably I could leave a TODO item here?

owent: That's great.

meastp:
@latlib I'm on leave, so haven't looked at the changes in detail, but this is fantastic. Great work @ThomsonTan! :) 

When I did my "fork" I had problems with some of the examples/tests. I copied the shared dlls manually, which made it a bit cumbersome, but I think cmake has some function to fix this. 
Anyway (as you know, of course), it is important to check the examples, otlp with both http and grpc (I had some threading trouble, but hopefully this PR/newer library versions have fixed this. If the problems i encountered is still present, they will be visible just by running tests and/or examples, so I wouldn't worry much about it. :)

Again, this makes me very happy - thanks! well done :)

api/include/opentelemetry/common/macros.h
ThomsonTan: Added TODO to `macros.h`?

api/include/opentelemetry/logs/logger_provider.h
```diff
@@ -44,4 +45,29 @@ class LoggerProvider
...
+public:
+  void operator()(OPENTELEMETRY_NAMESPACE::logs::LoggerProvider *logger_provider)
+  {
+    delete logger_provider;
+  }
+};
+
+}  // namespace std
+
+#  endif  // defined(OPENTELEMETRY_EXPORT)
```
marcalff, line 71: `default_delete<OPENTELEMETRY_NAMESPACE::logs::LoggerProvider>` is only implemented on windows, when `OPENTELEMETRY_EXPORT` is defined.

Assuming the default implementation used by unique_ptr / shared_ptr does the same anyway to delete an object, should be define `default_delete<OPENTELEMETRY_NAMESPACE::logs::LoggerProvider>` on all platforms, to have less differences between windows and non windows code ?

docs/build-as-dll.md
```diff
@@ -0,0 +1,8 @@
+# Design of building OpenTelemetry C++ SDK and exporters as DLL
+
+OpenTelemetry C++ can be built into single DLL named opentelemetry_cpp.dll on
+Windows, with CMake flag `OPENTELEMETRY_BUILD_DLL` set to `ON`.
+
+To instrument application based on the DLL, please define macro
+`OPENTELEMETRY_BUILD_IMPORT_DLL` before including any OpenTelemetry API header
+files in the application source files.
```
marcalff, line 8: See previous comment, defining _IMPORT_DLL can probably be avoided.

CMakeLists.txt
```diff
@@ -513,6 +513,13 @@ endif()
...
+  if(NOT MSVC)
+    message(WARNING "Build DLL is supposed to work with MSVC")
+  endif()
+  add_definitions(-DOPENTELEMETRY_BUILD_EXPORT_DLL)
```
marcalff, line 520: This has a global scope, for the entire built tree.

When OPENTELEMETRY_BUILD_DLL is set,
the content of the library (api, sdk, exporters, ext) should be build with _EXPORT_DDL,
but everything else that is not part of the library (examples, tests) should not see _EXPORT_DDL.

It seems this is the root cause for need of a _IMPORT_DDL flag.

examples/common/foo_library/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if(DEFINED OPENTELEMETRY_BUILD_DLL)
+  add_definitions(-DOPENTELEMETRY_BUILD_IMPORT_DLL)
+endif()
```
marcalff, line 3: This is correcting global scope pollution caused by the top level CMakeList.txt,
see earlier comments.

examples are not part of the opentelemetry-cpp library, so compiling examples should never see OPENTELEMETRY_BUILD_DDL defined.

examples/otlp/CMakeLists.txt
```diff
@@ -1,43 +1,69 @@
...
+
+  target_link_libraries(example_otlp_grpc ${CMAKE_THREAD_LIBS_INIT}
+                        common_foo_library)
+
+  if(DEFINED OPENTELEMETRY_BUILD_DLL)
+    target_link_libraries(example_otlp_grpc opentelemetry_cpp)
+  else()
+    target_link_libraries(example_otlp_grpc opentelemetry_trace
+                          opentelemetry_exporter_otlp_grpc)
+  endif()
```
marcalff, line 23: Two questions here:

- do we really want to put all the sdk and exporter code (which ones, it depends on WITH_JAEGER etc) into the same opentelemetry_cpp library, or should be have one opentelemetry_sdk library and libraries for each exporters ?
- maintaining different libraries dependencies based on platform and compiling options is going to be fragile, and easy to break, causing burden. Any way to improve this ?

ext/src/dll/opentelemetry_cpp.src
```diff
@@ -0,0 +1,18 @@
+// clang-format off
+LIBRARY opentelemetry_cpp
+EXPORTS
+    ?Create@OStreamSpanExporterFactory@trace@exporter@v1@opentelemetry@@SA?AV?$unique_ptr@VSpanExporter@trace@sdk@v1@opentelemetry@@U?$default_delete@VSpanExporter@trace@sdk@v1@opentelemetry@@@std@@@std@@XZ
```
marcalff, line 4: This file is checked in in git, so it is considered source code.

I assume it was generated somehow.

For unix users sending PR to the opentelemetry-cpp repository, how are users without access to a windows machine expected to come up with the mangled symbol syntax used here for exports ?

sdk/test/trace/simple_processor_test.cc
```diff
@@ -19,7 +19,7 @@ TEST(SimpleProcessor, ToInMemorySpanExporter)
 {
   std::unique_ptr<InMemorySpanExporter> exporter(new InMemorySpanExporter());
   std::shared_ptr<InMemorySpanData> span_data = exporter->GetData();
-  SimpleSpanProcessor processor(std::move(exporter));
+  SimpleSpanProcessor processor(std::unique_ptr<SpanExporter>{exporter.release()});
```
marcalff, line 22: How about:

```
InMemorySpanExporter *exporter_ptr = new InMemorySpanExporter();
std::shared_ptr<InMemorySpanData> span_data = exporter_ptr->GetData();
std::unique_ptr<SpanExporter> exporter(exporter_ptr);
```

For exporter, std::unique_ptr<> should always use the abstraction (SpanExporter), not a specific sub class.

This will remove the need to release + create a unique pointer again to go from a sub class to the abstraction.

marcalff (changes requested):
Please see some comments, questions, as well as suggestions for code cleanup.

api/include/opentelemetry/common/macros.h
```diff
@@ -175,3 +175,27 @@ point.
 #  define OPENTELEMETRY_API_SINGLETON
 
 #endif
+
+//
+// The if/elif order matters here. If both OPENTELEMETRY_BUILD_IMPORT_DLL and
+// OPENTELEMETRY_BUILD_EXPORT_DLL are defined, the former takes precedence.
```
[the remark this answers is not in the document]
ThomsonTan: Use 2 macros internally will make it more clear, as there are 3 compilation status for each file, export, import, and static. Use separate variable makes the usage clear. And also these are internal variable and not exposed to users.

api/include/opentelemetry/logs/logger_provider.h
ThomsonTan: The customized deletor is for separating the `delete` into DLL, other platforms without this issue don't need this customization which could incur non-necessary perf overhead. I don't assume the default is implemented like this so apply this to all platforms.

docs/build-as-dll.md
ThomsonTan: As there are 3 status, I think define 2 internal variables could make it more clear:)

CMakeLists.txt
ThomsonTan: Yes, define `OPENTELEMETRY_BUILD_EXPORT_DLL` is global scope is intended than pushing it down into discrete libraries which should be compiled as DLL. So in cases of adding future component, it doesn't need to add this flag explicitly for DLL build (which could be error prone).

examples/common/foo_library/CMakeLists.txt
ThomsonTan: If example sees `OPENTELEMETRY_BUILD_DLL`, it knows it is going to reference opentelemetry as DLL. `OPENTELEMETRY_BUILD_DLL` implies opentelemetry should be in DLL form instead of compiling any current code as DLL.

examples/otlp/CMakeLists.txt
ThomsonTan: 1. I prefer building SDK/exporter into one binary, so we only need export API and some necessary factory functions from SDK. If building SDK/exporter into separate DLLs, we need export more SDK internal functions for exporter which is fragile and we don't guarantee any stability there.
2. What kind of fragility do you mean? Any concrete example on this? The libraries and dependencies should be in one package and are expected to be updated as a whole.

ext/src/dll/opentelemetry_cpp.src
ThomsonTan: This is manually edited as these are just decorated name. I was considering adding the undecorated symbol name to this file for better readability.

This file is dedicated for Windows. Linux related PR should not need to touch this unless changing API/Factory method which should be very rare? But the fallback is the build break message will show the unresolved decorated symbol(s) which might be added here.

sdk/test/trace/simple_processor_test.cc
ThomsonTan: Yes, this also works and will simply the test code, will apply this change, thanks!

api/include/opentelemetry/logs/logger_provider.h
```diff
@@ -44,4 +45,29 @@ class LoggerProvider
...
+namespace std
+{
+
+//
+// Partial specialization of default_delete used by unique_ptr.
+// This makes the delete of the type in unique_ptr happening in the DLL where it
+// is allocated.
+//
+template <>
+class OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::logs::LoggerProvider>
```
lalitb, line 60: The default_delete is the destruction policy used by unique_ptr, however the LoggerProvider is of type nostd::shared_ptr. How will this work with shared_ptr.

ThomsonTan: Yes, this the shared_ptr could be a problem here. Thanks for the catch, @marcalff and @lalitb . While I am looking more into it, is there any recommendation on preventing the memory free pattern?

api/include/opentelemetry/logs/logger_provider.h
marcalff: Ok, thanks for the clarification.

examples/otlp/CMakeLists.txt
marcalff: > 1. I prefer building SDK/exporter into one binary, so we only need export API and some necessary factory functions from SDK. If building SDK/exporter into separate DLLs, we need export more SDK internal functions for exporter which is fragile and we don't guarantee any stability there.

Assume two applications.

Application A uses the OTLP GRPC exporter and PROMETHEUS, and the build uses WITH_OTLP_HTTP=NO, WITH_ZIPKIN=NO.

Application B uses the OTLP HTTP exporter and ZIPKIN, and uses WITH_OTLP_GRPC=NO, WITH_PROMETHEUS=NO

How can both applications A and B be deployed on the same machine, when they both use a DLL of the same name, opentelemetry-cpp, and expect different content in it ?

marcalff: >     2. What kind of fragility do you mean? Any concrete example on this? The libraries and dependencies should be in one package and are expected to be updated as a whole.

Assume the example code needs an extra dependency on a example_helpers library.

Because this library is not in opentelemetry-cpp, both target_link_library() calls need to be updated.

Now, the example code also needs opentelemetry_resource.

Because this library is in opentelemetry-cpp, only the second call needs to be updated.

Granted, all incorrect dependencies can and should be caught in CI with a build break, but having twice the amount of target_link_library() in the source code makes it harder to maintain.

ext/src/dll/opentelemetry_cpp.src
marcalff: Thanks for the clarifications.

Please add the undecorated names in comments, it will help to search the source code (for example, when searching for "OStreamSpanExporterFactory::Create").

Changing exported apis should be rare, but adding new ones will be more frequent.

api/include/opentelemetry/logs/logger_provider.h
marcalff: According to [cpp_reference](https://en.cppreference.com/w/cpp/memory/unique_ptr), for unique pointers, the default deleter is:

```
class Deleter = std::default_delete<T>
```

For shared_ptr, the [doc](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr) does not provide a default, and gives constraints for the expected behavior of the deleter:

```
template< class Y, class Deleter >
shared_ptr( Y* ptr, Deleter d );
```

If I understand correctly, it means we will have to implement Deleter classes for every shared pointer used,
and declare shared_ptr as:

```
shared_ptr<TracerProvider, TracerProviderDeleter>
```

In this case, implementing the deleter twice (once as std::default_delete<> for unique_ptr, once for shared_ptr, is an issue. We may use TracerProviderDeleter as well for unique_ptr and never rely on the default.

Also, in this case, we need typedefs to make the code more readable:

typedef shared_ptr<TracerProvider, TracerProviderDeleter> TracerProviderSharedPtr;
typedef unique_ptr<TracerProvider, TracerProviderDeleter> TracerProviderUniquePtr;

or maybe helper types:

TracerProvider::shared_ptr
TracerProvider::unique_ptr

If this turns out to be required, this change is going to break existing code and abi.

ext/src/dll/opentelemetry_cpp.src
ThomsonTan: Ok. Added the undecorated names to the comment.

api/include/opentelemetry/logs/logger_provider.h
ThomsonTan: Here is the change for this.
```diff
  explicit shared_ptr(pointer ptr)
  {
-    std::shared_ptr<T> ptr_(ptr);
+    std::shared_ptr<T> ptr_(ptr, std::default_delete<T>());
```

examples/otlp/CMakeLists.txt
ThomsonTan: > > 1. I prefer building SDK/exporter into one binary, so we only need export API and some necessary factory functions from SDK. If building SDK/exporter into separate DLLs, we need export more SDK internal functions for exporter which is fragile and we don't guarantee any stability there.
> 
> Assume two applications.
> 
> Application A uses the OTLP GRPC exporter and PROMETHEUS, and the build uses WITH_OTLP_HTTP=NO, WITH_ZIPKIN=NO.
> 
> Application B uses the OTLP HTTP exporter and ZIPKIN, and uses WITH_OTLP_GRPC=NO, WITH_PROMETHEUS=NO
> 
> How can both applications A and B be deployed on the same machine, when they both use a DLL of the same name, opentelemetry-cpp, and expect different content in it ?

The DLL is considered as common library, so it should container the features which can be used by client. Like here, DLL should include OTLP GRPC/HTTP, ZIPKIN, PROMETHEUS then application A and B can share one DLL. For application with has customized needs like some_unknown_exporter, it is still recommended to use static link instead of DLL here.

ThomsonTan: > > ```
> > 2. What kind of fragility do you mean? Any concrete example on this? The libraries and dependencies should be in one package and are expected to be updated as a whole.
> > ```
> 
> Assume the example code needs an extra dependency on a example_helpers library.
> 
> Because this library is not in opentelemetry-cpp, both target_link_library() calls need to be updated.
> 
> Now, the example code also needs opentelemetry_resource.
> 
> Because this library is in opentelemetry-cpp, only the second call needs to be updated.
> 
> Granted, all incorrect dependencies can and should be caught in CI with a build break, but having twice the amount of target_link_library() in the source code makes it harder to maintain.

For example_helper, it should not add to both target_link_libraries because it is common for both DLL and non-DLL build, so it can be added to target_link_libraries unconditionally.

For the case like opentelemetry-resource, it may need to be added to the second call of target_link_libraries if it is not referenced by any other libraries in the DLL. Currently this is necessary as the DLL dependencies are managed separately. Agree this introduces complexity to maintenance, but assume maintaining the extra dependencies for DLL should be acceptable?

api/include/opentelemetry/logs/logger_provider.h
marcalff: This will fix only nostd::shared_ptr<T> when building without C++ 11 std::shared_ptr, WITH_STL=OFF.

What about the build where nostd::shared_ptr is an alias for std::shared_ptr, WITH_STL=ON ?

This code will not be used, and the deleter for std::shared_ptr<T> will be the same as before.

.github/workflows/ci.yml
```diff
@@ -507,6 +507,23 @@ jobs:
...
+        submodules: 'recursive'
+    - name: setup
+      run: |
+        ./ci/setup_windows_cmake.ps1
+        ./ci/setup_windows_ci_environment.ps1
+        ./ci/install_windows_protobuf.ps1
+    - name: run cmake test (DLL build)
+      run: ./ci/do_ci.ps1 cmake.dll.test
+    - name: run otprotocol test (DLL build)
+      run: ./ci/do_ci.ps1 cmake.exporter.otprotocol.dll.test
```
marcalff, line 525: Please define both:

* a DDL build WITH_STL=ON
* a DDL build WITH_STL=OFF

marcalff (changes requested):

api/include/opentelemetry/logs/logger_provider.h
ThomsonTan: This is a good point. To fix `WITH_STIL=ON`, we may need implement the 5th constructor of `shared_ptr` for `nostd::shared_ptr`, then passing the deleter in the caller (like in [GetLogger](https://github.com/open-telemetry/opentelemetry-cpp/blob/1ae2fe538ceaaf3829c2972a05e043cf43676000/sdk/src/logs/logger_provider.cc#L110)). 

I'd prefer emitting an error for DLL build when `WITH_STL=ON` and fix this later.

ThomsonTan: As the major purpose of providing nostd class is for ABI compatibility, I am wondering whether we recommend DLL user to build with `WITH_STIL=ON`, or we should not recommend that because that doesn't provide ABI compatibility.

ThomsonTan: Added error out for DLL build and `WITH_STL=ON`.

.github/workflows/ci.yml
ThomsonTan: > Please define both:
> 
> * a DDL build WITH_STL=ON
> * a DDL build WITH_STL=OFF

The 1st definition will be unnecessary if it is not supported for not (error out).

examples/otlp/CMakeLists.txt
marcalff: > The DLL is considered as common library, so it should container the features which can be used by client. Like here, DLL should include OTLP GRPC/HTTP, ZIPKIN, PROMETHEUS then application A and B can share one DLL. For application with has customized needs like some_unknown_exporter, it is still recommended to use static link instead of DLL here.

To me this is a critical problem, maybe I did not expressed it well enough.

How can 2 vendors ship applications A and B independently ?

Will vendor A see an unwanted exporter in the DLL binary, loaded in application A process, just because vendor B needs it ?

marcalff:
All,

Following comments in each comment thread is getting hard,
so summarizing here the most critical points found so far.

I have more comments on less critical parts,
saving them for later to not divert the discussion.

## Technical solution

First, I am not convinced the following code works.

```
//
// Partial specialization of default_delete used by unique_ptr or shared_ptr.
// This makes the delete of the type in unique_ptr/shared_ptr happening in the
// DLL in which it is allocated.
//
template <>
struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter>
{
public:
  void operator()(OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter *span_exporter)
  {
    delete span_exporter;
  }
};
```

1)

default_delete<T> is used for std::unique_ptr
It is _not_ used by std::shared_ptr<T>,
as found recently looking at cppreference.

std::shared_ptr<T> uses a _delete expression_, as in `delete ptr` instead.

In my understanding, `std::shared_ptr<T>` will never call default_delete<T>,
making the fix incomplete.

Changing nostd::shared_ptr<T> is not going to address std::shared_ptr<T>

2)

this is an inline template, exposed in header files,
and compiled with the application code compiling unit.

Regardless of what OPENTELEMETRY_EXPORT expands to, this is still inlined
code that will invoke `delete span_exporter`, from the application binary,
and not from the opentelemetry dll.

I don't understand what is achieved here, compared to the default
std::unique_ptr<T> code, which invokes default_delete<T>, also calling
`delete ptr` from the application space.

I would understand the OPENTELEMETRY_EXPORT part if this was not an inline
definition but only a declaration, with the definition in the DLL, but this
is not the case.

I admit I am not familiar with DLL IMPORT/EXPORT,
so it could be I missed something here.

In that case, please explain, with a lot of details, why and HOW this code works.

Also, a description of the technical issue to resolve would help here:
compared to linux shared libraries, what is it with windows DLL that
introduce additional or different constraints ?

## Verification

The fact that unit tests runs in CI is inconclusive, and does not prove
the fix works.

Do the tests actually expose the broken behavior in the first place,
that is, destroying memory from the application space (the unit test)
when removing the last reference of a unique_ptr or shared_ptr that was
allocated in the opentelemetry space ?

If the last reference to an object is removed from inside the
opentelemetry DLL (see CleanupTracer() in examples), the original bug is not exposed,
and the code change can not be tested.

Take this test for example:

```
TEST(OStreamSpanExporter, Shutdown)
{
  auto exporter  = std::unique_ptr<trace_sdk::SpanExporter>(new exportertrace::OStreamSpanExporter);
  auto processor = std::shared_ptr<trace_sdk::SpanProcessor>(
      new trace_sdk::SimpleSpanProcessor(std::move(exporter)));

  auto recordable = processor->MakeRecordable();
  recordable->SetName("Test Span");

  // Capture the output of cout
  const auto captured = WithOStreamCapture(std::cout, [&]() {
    EXPECT_TRUE(processor->Shutdown());
    processor->OnEnd(std::move(recordable));
  });
  std::string err_message =
      "[Ostream Trace Exporter] Exporting 1 span(s) failed, exporter is shutdown";
  EXPECT_TRUE(captured.find(err_message) != std::string::npos);
}

```

`exporter` and `processor` are allocated locally, from the unit test, and not
allocated from the opentelemetry-cpp library.

In fact, this test links with opentelemetry_exporter_ostream_span directly,
not with the DLL at all.

The only place that links with the DLL is examples, and examples are not
executed in CI.

## Technical risk

Currently, the build can use WITH_STL=ON or WITH_STL=OFF

Ultimately, the DDL should exist for both builds.

Given how unique_ptr and shared_ptr are involved,
starting to implement a solution for nostd only is a major red flag in my
opinion: once this works for nostd, there is no guarantee that the solution
can be transposed to std, especially is code changes that deviate from the
std api are implemented in nostd.

It will be less risk to implement a DDL for WITH_STL=ON first, and then
adapt the code to support WITH_STL=OFF second.

## Deployment

As discussed previously with packaging in general,
I don't see how a single DDL can be shipped for the SDK and all exporters.

We don't need to cut the SDK into smaller parts (per signal), so a single
DDL for the entire SDK should be ok.

For exporters, each exporter should have its own DDL in my opinion.

If somehow a single DDL is desirable, please describe how is it supposed to be used, I just don't get it.

CMakeLists.txt
```diff
@@ -521,6 +521,19 @@ endif()
...
 include(CMakePackageConfigHelpers)
 
+if(DEFINED OPENTELEMETRY_BUILD_DLL)
+  if(NOT MSVC)
+    message(WARNING "Build DLL is supposed to work with MSVC!")
+  endif()
+  if(WITH_STL)
+    message(
+      STATUS_FATAL
+      "Build DLL with C++ STL is not recommended because ABI compatibility could be broken!"
```
lalitb, line 531: This error message is actually confusing. Should we just say "Build DLL with C++ STL is not supported!", in case we plan to not support it due to technical challenges.

api/include/opentelemetry/trace/tracer_provider.h
```diff
@@ -29,3 +29,27 @@ class TracerProvider
...
+namespace std
+{
+
+//
+// Partial specialization of default_delete used by unique_ptr or shared_ptr.
+// This makes the delete of the type in unique_ptr/shared_ptr happening in the
+// DLL in which it is allocated.
+//
+template <>
+struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::trace::TracerProvider>
```
lalitb, line 44: Should we have this definition only specific for `DLL` scenario ?

sdk/include/opentelemetry/sdk/trace/exporter.h
```diff
@@ -53,3 +53,27 @@ class SpanExporter
...
+
+namespace std
+{
+
+//
+// Partial specialization of default_delete used by unique_ptr or shared_ptr.
+// This makes the delete of the type in unique_ptr/shared_ptr happening in the
+// DLL in which it is allocated.
+//
+template <>
```
lalitb, line 67: should we have this definition only specifically for DLL scenario ?

CMakeLists.txt
ThomsonTan: > All,
> 
> Following comments in each comment thread is getting hard, so summarizing here the most critical points found so far.
> 
> I have more comments on less critical parts, saving them for later to not divert the discussion.
> 
> ## Technical solution
> First, I am not convinced the following code works.
> 
> ```
> //
> // Partial specialization of default_delete used by unique_ptr or shared_ptr.
> // This makes the delete of the type in unique_ptr/shared_ptr happening in the
> // DLL in which it is allocated.
> //
> template <>
> struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter>
> {
> public:
>   void operator()(OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter *span_exporter)
>   {
>     delete span_exporter;
>   }
> };
> ```
> 
> 
> default_delete is used for std::unique_ptr It is _not_ used by std::shared_ptr, as found recently looking at cppreference.
> 
> std::shared_ptr uses a _delete expression_, as in `delete ptr` instead.
> 
> In my understanding, `std::shared_ptr<T>` will never call default_delete, making the fix incomplete.
> 
> Changing nostd::shared_ptr is not going to address std::shared_ptr
> 
> 
> this is an inline template, exposed in header files, and compiled with the application code compiling unit.
> 
> Regardless of what OPENTELEMETRY_EXPORT expands to, this is still inlined code that will invoke `delete span_exporter`, from the application binary, and not from the opentelemetry dll.
> 
> I don't understand what is achieved here, compared to the default std::unique_ptr code, which invokes default_delete, also calling `delete ptr` from the application space.
> 
> I would understand the OPENTELEMETRY_EXPORT part if this was not an inline definition but only a declaration, with the definition in the DLL, but this is not the case.
> 
> I admit I am not familiar with DLL IMPORT/EXPORT, so it could be I missed something here.
> 
> In that case, please explain, with a lot of details, why and HOW this code works.
> 
> Also, a description of the technical issue to resolve would help here: compared to linux shared libraries, what is it with windows DLL that introduce additional or different constraints ?
> 
> ## Verification
> The fact that unit tests runs in CI is inconclusive, and does not prove the fix works.
> 
> Do the tests actually expose the broken behavior in the first place, that is, destroying memory from the application space (the unit test) when removing the last reference of a unique_ptr or shared_ptr that was allocated in the opentelemetry space ?
> 
> If the last reference to an object is removed from inside the opentelemetry DLL (see CleanupTracer() in examples), the original bug is not exposed, and the code change can not be tested.
> 
> Take this test for example:
> 
> ```
> TEST(OStreamSpanExporter, Shutdown)
> {
>   auto exporter  = std::unique_ptr<trace_sdk::SpanExporter>(new exportertrace::OStreamSpanExporter);
>   auto processor = std::shared_ptr<trace_sdk::SpanProcessor>(
>       new trace_sdk::SimpleSpanProcessor(std::move(exporter)));
> 
>   auto recordable = processor->MakeRecordable();
>   recordable->SetName("Test Span");
> 
>   // Capture the output of cout
>   const auto captured = WithOStreamCapture(std::cout, [&]() {
>     EXPECT_TRUE(processor->Shutdown());
>     processor->OnEnd(std::move(recordable));
>   });
>   std::string err_message =
>       "[Ostream Trace Exporter] Exporting 1 span(s) failed, exporter is shutdown";
>   EXPECT_TRUE(captured.find(err_message) != std::string::npos);
> }
> ```
> 
> `exporter` and `processor` are allocated locally, from the unit test, and not allocated from the opentelemetry-cpp library.
> 
> In fact, this test links with opentelemetry_exporter_ostream_span directly, not with the DLL at all.
> 
> The only place that links with the DLL is examples, and examples are not executed in CI.
> 
> ## Technical risk
> Currently, the build can use WITH_STL=ON or WITH_STL=OFF
> 
> Ultimately, the DDL should exist for both builds.
> 
> Given how unique_ptr and shared_ptr are involved, starting to implement a solution for nostd only is a major red flag in my opinion: once this works for nostd, there is no guarantee that the solution can be transposed to std, especially is code changes that deviate from the std api are implemented in nostd.
> 
> It will be less risk to implement a DDL for WITH_STL=ON first, and then adapt the code to support WITH_STL=OFF second.
> 
> ## Deployment
> As discussed previously with packaging in general, I don't see how a single DDL can be shipped for the SDK and all exporters.
> 
> We don't need to cut the SDK into smaller parts (per signal), so a single DDL for the entire SDK should be ok.
> 
> For exporters, each exporter should have its own DDL in my opinion.
> 
> If somehow a single DDL is desirable, please describe how is it supposed to be used, I just don't get it.

1. Yes, `std::shared_ptr<>` issue is open, but as we don't support `WITH_STL` with DLL build, this is fine.

I thought about fixing `std::shared_ptr<>`. For `WITH_STL`, we can still construct it with a second explicit parameter of deleter for `std::shared_ptr<>`, but the issue is the user's code can drive either `TracerProvider` and `LoggerProvider` and implements their own GetLogger/GetTracer which returns `std::shared_ptr<>` without deleter. So instead of fixing that for SDK but leave holes to the user, I'd prefer we just disable the scenario from the beginning.

Also, not support `WITH_STL` with DLL is mainly not for the reason that it is hard for us/maintainer to implement it, the rationale is it is risky for user to bring their own STL to the binary which will be passed to our DLL, because that STL could be in different version our the one compiled into DLL, then causing inconsistency between binary layout.

2.  `OPENTELEMETRY_EXPORT` is expanded to `__declspec(dllimport)` in the application code, so the referenced symbol like the `operator()` function will be marked as import (symbol name changed after decoratoin) which only resolves to the one marked as `__declspec(dllexport)` in the DLL. It will not resolve to the local definition and could get unresolved link error if it is not exported in DLL.

## Verification
Agree it is lacking testing for now. Not all tests can be converted to using the DLL, like the one you listed as `OStreamSpanExporter::Shutdown` which tests the internal SDK behavior, and should be considered as external user/component of the DLL. For example, SDK user is not expected to new `SpanExporter` or its derived classes directly, the factory class like `OStreamSpanExporterFactory` should be used instead. So the examples are in perfect position of doing this migration which demos the real user scenario. But yes, examples are not tests. let me create some DLL specific tests which only relies on the API and exposed SDK functions. Let me know if there is any concern on this.

## Technical risk
As mentioned in 1. above, don't support WITH_STL in DLL build gives us long term benefit which less fragile on ABI compatibility. But I am also open to hear any customer concern like WITH_STL is required even if they call into our DLL. And anyway, the DLL build will be in preview for some time and we can still make change to it after this PR.

## Deployment
For separating SDK and exporters into different DLLs, we discussed this internally. But this requires us to expose more functions in SDK for exporter use, and dealing with more potential inconsistent memory alloc/free between SDK and exporters. Further more, a more stable interface between SDK and exporters will be needed or a customer provided exporter DLL can be broken easily with SDK DLL upgrade. So we prefer to build SDK and exporters in 1 DLL. But this can be change if there is valid request.

The single DLL includes SDK and selected exporter which the customer can call into. If the user needs an exporter which is not in the existing DLL, they need build it as static or request the DLL owner to add it.
> If somehow a single DDL is desirable, please describe how is it supposed to be used, I just don't get it.

ThomsonTan:
> All,
> 
> Following comments in each comment thread is getting hard, so summarizing here the most critical points found so far.
> 
> I have more comments on less critical parts, saving them for later to not divert the discussion.
> 
> ## Technical solution
> First, I am not convinced the following code works.
> 
> ```
> //
> // Partial specialization of default_delete used by unique_ptr or shared_ptr.
> // This makes the delete of the type in unique_ptr/shared_ptr happening in the
> // DLL in which it is allocated.
> //
> template <>
> struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter>
> {
> public:
>   void operator()(OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter *span_exporter)
>   {
>     delete span_exporter;
>   }
> };
> ```
> 
> 
> default_delete is used for std::unique_ptr It is _not_ used by std::shared_ptr, as found recently looking at cppreference.
> 
> std::shared_ptr uses a _delete expression_, as in `delete ptr` instead.
> 
> In my understanding, `std::shared_ptr<T>` will never call default_delete, making the fix incomplete.
> 
> Changing nostd::shared_ptr is not going to address std::shared_ptr
> 
> 
> this is an inline template, exposed in header files, and compiled with the application code compiling unit.
> 
> Regardless of what OPENTELEMETRY_EXPORT expands to, this is still inlined code that will invoke `delete span_exporter`, from the application binary, and not from the opentelemetry dll.
> 
> I don't understand what is achieved here, compared to the default std::unique_ptr code, which invokes default_delete, also calling `delete ptr` from the application space.
> 
> I would understand the OPENTELEMETRY_EXPORT part if this was not an inline definition but only a declaration, with the definition in the DLL, but this is not the case.
> 
> I admit I am not familiar with DLL IMPORT/EXPORT, so it could be I missed something here.
> 
> In that case, please explain, with a lot of details, why and HOW this code works.
> 
> Also, a description of the technical issue to resolve would help here: compared to linux shared libraries, what is it with windows DLL that introduce additional or different constraints ?
> 
> ## Verification
> The fact that unit tests runs in CI is inconclusive, and does not prove the fix works.
> 
> Do the tests actually expose the broken behavior in the first place, that is, destroying memory from the application space (the unit test) when removing the last reference of a unique_ptr or shared_ptr that was allocated in the opentelemetry space ?
> 
> If the last reference to an object is removed from inside the opentelemetry DLL (see CleanupTracer() in examples), the original bug is not exposed, and the code change can not be tested.
> 
> Take this test for example:
> 
> ```
> TEST(OStreamSpanExporter, Shutdown)
> {
>   auto exporter  = std::unique_ptr<trace_sdk::SpanExporter>(new exportertrace::OStreamSpanExporter);
>   auto processor = std::shared_ptr<trace_sdk::SpanProcessor>(
>       new trace_sdk::SimpleSpanProcessor(std::move(exporter)));
> 
>   auto recordable = processor->MakeRecordable();
>   recordable->SetName("Test Span");
> 
>   // Capture the output of cout
>   const auto captured = WithOStreamCapture(std::cout, [&]() {
>     EXPECT_TRUE(processor->Shutdown());
>     processor->OnEnd(std::move(recordable));
>   });
>   std::string err_message =
>       "[Ostream Trace Exporter] Exporting 1 span(s) failed, exporter is shutdown";
>   EXPECT_TRUE(captured.find(err_message) != std::string::npos);
> }
> ```
> 
> `exporter` and `processor` are allocated locally, from the unit test, and not allocated from the opentelemetry-cpp library.
> 
> In fact, this test links with opentelemetry_exporter_ostream_span directly, not with the DLL at all.
> 
> The only place that links with the DLL is examples, and examples are not executed in CI.
> 
> ## Technical risk
> Currently, the build can use WITH_STL=ON or WITH_STL=OFF
> 
> Ultimately, the DDL should exist for both builds.
> 
> Given how unique_ptr and shared_ptr are involved, starting to implement a solution for nostd only is a major red flag in my opinion: once this works for nostd, there is no guarantee that the solution can be transposed to std, especially is code changes that deviate from the std api are implemented in nostd.
> 
> It will be less risk to implement a DDL for WITH_STL=ON first, and then adapt the code to support WITH_STL=OFF second.
> 
> ## Deployment
> As discussed previously with packaging in general, I don't see how a single DDL can be shipped for the SDK and all exporters.
> 
> We don't need to cut the SDK into smaller parts (per signal), so a single DDL for the entire SDK should be ok.
> 
> For exporters, each exporter should have its own DDL in my opinion.
> 
> If somehow a single DDL is desirable, please describe how is it supposed to be used, I just don't get it.

1. Yes, `std::shared_ptr<>` issue is open, but as we don't support `WITH_STL` with DLL build, this is fine.

I thought about fixing `std::shared_ptr<>`. For `WITH_STL`, we can still construct it with a second explicit parameter of deleter for `std::shared_ptr<>`, but the issue is the user's code can drive either `TracerProvider` and `LoggerProvider` and implements their own GetLogger/GetTracer which returns `std::shared_ptr<>` without deleter. So instead of fixing that for SDK but leave holes to the user, I'd prefer we just disable the scenario from the beginning.

Also, not support `WITH_STL` with DLL is mainly not for the reason that it is hard for us/maintainer to implement it, the rationale is it is risky for user to bring their own STL to the binary which will be passed to our DLL, because that STL could be in different version our the one compiled into DLL, then causing inconsistency between binary layout.

2.  `OPENTELEMETRY_EXPORT` is expanded to `__declspec(dllimport)` in the application code, so the referenced symbol like the `operator()` function will be marked as import (symbol name changed after decoratoin) which only resolves to the one marked as `__declspec(dllexport)` in the DLL. It will not resolve to the local definition and could get unresolved link error if it is not exported in DLL.

## Verification
Agree it is lacking testing for now. Not all tests can be converted to using the DLL, like the one you listed as `OStreamSpanExporter::Shutdown` which tests the internal SDK behavior, and should be considered as external user/component of the DLL. For example, SDK user is not expected to new `SpanExporter` or its derived classes directly, the factory class like `OStreamSpanExporterFactory` should be used instead. So the examples are in perfect position of doing this migration which demos the real user scenario. But yes, examples are not tests. let me create some DLL specific tests which only relies on the API and exposed SDK functions. Let me know if there is any concern on this.

## Technical risk
As mentioned in 1. above, don't support WITH_STL in DLL build gives us long term benefit which less fragile on ABI compatibility. But I am also open to hear any customer concern like WITH_STL is required even if they call into our DLL. And anyway, the DLL build will be in preview for some time and we can still make change to it after this PR.

## Deployment
For separating SDK and exporters into different DLLs, we discussed this internally. But this requires us to expose more functions in SDK for exporter use, and dealing with more potential inconsistent memory alloc/free between SDK and exporters. Further more, a more stable interface between SDK and exporters will be needed or a customer provided exporter DLL can be broken easily with SDK DLL upgrade. So we prefer to build SDK and exporters in 1 DLL. But this can be change if there is valid request.

The single DLL includes SDK and selected exporter which the customer can call into. If the user needs an exporter which is not in the existing DLL, they need build it as static or request the DLL owner to add it.
> If somehow a single DDL is desirable, please describe how is it supposed to be used, I just don't get it.

sdk/include/opentelemetry/sdk/trace/exporter.h
ThomsonTan: In line 57, there is a check on macro `OPENTELEMETRY_EXPORT`, so this is only defined for DLL scenario. Or anything I missed?

examples/otlp/CMakeLists.txt
ThomsonTan: DLL linking requires kind of dev package with [import library](https://learn.microsoft.com/en-us/cpp/build/reference/building-an-import-library-and-export-file?view=msvc-170), so A and B links to one DLL, they should be aware that what exporters they could get.

Or if you mean A might load opentelemetry_cpp.dll released with B which is unexpected, app A should either build its own opentelemetry_cpp.dll and release it with A (in the same folder). Any app tries to install opentelemery_cpp.dll to system location should be aware that this may break others and this is a known [DLL hell](https://en.wikipedia.org/wiki/DLL_Hell) issue it it happens.

marcalff:
@ThomsonTan 

Thanks for the details.

About this:

> OPENTELEMETRY_EXPORT is expanded to __declspec(dllimport) in the application code, so the referenced symbol like the operator() function will be marked as import (symbol name changed after decoratoin) which only resolves to the one marked as __declspec(dllexport) in the DLL. It will not resolve to the local definition and could get unresolved link error if it is not exported in DLL.

If 

```
template <>
struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter>
{
public:
  void operator()(OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter *span_exporter)
  {
    delete span_exporter;
  }
};
```

does not resolve to the local definition, why provide a local definition inline then ?

Should it be instead:

```
template <>
struct OPENTELEMETRY_EXPORT default_delete<OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter>
{
public:
  void operator()(OPENTELEMETRY_NAMESPACE::sdk::trace::SpanExporter *span_exporter);
};
```

For the second part of the comment, "and could get unresolved link error if it is not exported in DLL", doesn't this prevent libraries to ship instrumented code, compiled against the API only, that can be linked into a final application that is _not_ linked with opentelemetry-cpp ?

In my understanding, instrumented code must work both with, and without, a SDK.

examples/otlp/CMakeLists.txt
marcalff: > DLL linking requires kind of dev package with [import library](https://learn.microsoft.com/en-us/cpp/build/reference/building-an-import-library-and-export-file?view=msvc-170), so A and B links to one DLL, they should be aware that what exporters they could get.
> 
> Or if you mean A might load opentelemetry_cpp.dll released with B which is unexpected, app A should either build its own opentelemetry_cpp.dll and release it with A (in the same folder). Any app tries to install opentelemery_cpp.dll to system location should be aware that this may break others and this is a known [DLL hell](https://en.wikipedia.org/wiki/DLL_Hell) issue it it happens.

If each application A and B need to keep a private copy of the opentelemetry-cpp DLL, customized for they own use (choice of exporters), what is the whole point of having DLL then ? Applications A and B could be statically linked instead.

The scenario I am concerned about is DLL stomping, from the link you provided, with a DDL installed in a common location.

ThomsonTan:
For the inline definition of `default_delete<>::operator()`, I actually started with the way you mentioned and put the definition in a separate .cc file which works, but I found out that putting it inline also works because the local copy in the application is not used. This way is looks simpler and no need to find an appropriate .cc file to host the definition.

The DLL currently includes API, SDK and exporters, it is not meant for SDK only. So yes, even compile with API needs this DLL which exported all the API and hosts all the global variables.
> "and could get unresolved link error if it is not exported in DLL", doesn't this prevent libraries to ship instrumented code, compiled against the API only, that can be linked into a final application that is not linked with opentelemetry-cpp ?

examples/otlp/CMakeLists.txt
ThomsonTan: Install application local DLL is for edge case, and still even a single application could have multiple components(DLL) which can share a single copy of the DLL built by the app.

Install a DLL to a common location should be well considered because it will then be considered as kind of system components and should fulfil most user requirements. Like if Windows ship opentelemetry_cpp.dll into c:\windows\system32\, then the users can link their app to it (if the required exporter is included there then linking will succeed). If the link failed because missing exporters, the user has to built their own DLL or use static linking.

CMakeLists.txt
ThomsonTan: With customizing std::default_delete removed, this was updated as a warning message.

api/include/opentelemetry/trace/tracer_provider.h
ThomsonTan: This is removed now.

ThomsonTan:
@marcalff 

I removed all the customized `default_delete`. Run our sample (debug build) should should catch the memory allocation issue (included in the cmake.dll.test CI task)

I looked more on the memory allocation/free between DLLs/EXE, it seems specializing `default_delete` is a little over conservative, and the default `shared_ptr` and `unique_ptr` should be working. The point is that the object returned has virtual destructor defined which causes "scalar deleting destructor" being generated and saved in the virtual table of the object. The EXE and DLL have its own copy if virtual table separately, and create `new` object in the DLL/EXE will put the virtual table in current module into the object. So at `delete` time, the (virtual) destructor in the original module which `new`ed the object will be called, which is the exact behavior we expected. So it works.

(There is one case `default_delete` still doesn't work which is for [TracerContext](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/trace/tracer_context.h) because there is no virtual dtor defined there. In this case, the object returned from [TracerContextFactory](https://github.com/open-telemetry/opentelemetry-cpp/blob/32af3529efc50bd42d558be06b9206ee0dadb1e0/sdk/include/opentelemetry/sdk/trace/tracer_context_factory.h#L26) could be freed in the caller module incorrectly. I added default virtual dtor for `class TracerContext`. I think the cost is negligible but I could make it only for the DLL build if preferred.

With the removal of `default_deleter`, the issue for DLL+WITH_STL is also no longer an issue.

For validation, it is also more complicated to create a failed case. With VS2015 and later, [UCRT](https://devblogs.microsoft.com/cppblog/introducing-the-universal-crt/) replaces CRT, and UCRT no longer support creating dedicated heap for a module (DLL/EXE, with both dynamic and static linking), the [Process Heap](https://learn.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-getprocessheap) is always used. The implies that of both DLL and EXE use UCRT, it is Ok to alloc memory in one of them and free it in another one because both modules reference the same heap. I tried override `operator new(...)`, `operator delete(...)`  and redirect the request to a new heap. This seems working, but I don't think we want to introduce such complexity for the edge (might cause distraction).

marcalff:
> @marcalff
> 
> I removed all the customized `default_delete`. Run our sample (debug build) should should catch the memory allocation issue (included in the cmake.dll.test CI task)
> 

Thanks.

I will look into the patch in more details this week.

Sounds very good from the description.

CHANGELOG.md
```diff
@@ -64,6 +64,8 @@ Increment the:
   [#1916](https://github.com/open-telemetry/opentelemetry-cpp/pull/1916)
 * [SEMANTIC CONVENTIONS] Upgrade to version 1.17.0
   [#1927](https://github.com/open-telemetry/opentelemetry-cpp/pull/1927)
+* [BUILD] Build OpenTelemetry SDK and exporters into DLL
+  [#1932](https://github.com/open-telemetry/opentelemetry-cpp/pull/1932)
```
marcalff, line 68: Release 1.8.2 has shipped, the changelog entry needs to move to the unreleased section

docs/build-as-dll.md
```diff
@@ -0,0 +1,8 @@
+# Design of building OpenTelemetry C++ SDK and exporters as DLL
+
+OpenTelemetry C++ can be built into single DLL named opentelemetry_cpp.dll on
+Windows, with CMake flag `OPENTELEMETRY_BUILD_DLL` set to `ON`.
+
+To instrument a application based on the DLL, please define macro
+`OPENTELEMETRY_BUILD_IMPORT_DLL` before including any OpenTelemetry API header
+files in the application source files.
```
marcalff, line 8: I am convinced this part can be simplified.

Once OPENTELEMETRY_BUILD_DDL is defined, the build uses a DDL.

The user code should not have to define OPENTELEMETRY_BUILD_IMPORT_DDL.

Instead, the makefiles in opentelemetry-cpp should know when a file that belong to the DDL is compiled (EXPORT) or when a file that does not belong to the DDL is compiled (IMPORT).

marcalff (commented):
Thanks for all the changes, this patch fixes my most important concerns.

Having a single DDL for opentelemetry-cpp and all the exporters is still an area of discussion:

* should not be the long term goal in my opinion,
* ok to have this single DDL as an intermediate step, assuming we document this is experimental and subject to change.

marcalff (commented):
Thanks for this round of fixes, most critical concerns are now addressed.

marcalff (commented):
(struggling with github UI, restart review to remove requested changes flag)

docs/build-as-dll.md
ThomsonTan: The user mode can also built with the exiting static library which is default, so we need a new macro to guard referencing the DLL, so we don't break any existing users' instrumentation code.
> The user code should not have to define OPENTELEMETRY_BUILD_IMPORT_DLL.

ext/src/dll/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+  endif()
+endif()
+
+add_custom_command(
+  OUTPUT ${OPENTELEMETRY_EXPORT_DEF}
+  COMMAND
+    ${CMAKE_CXX_COMPILER}
+    "-D$<JOIN:$<TARGET_PROPERTY:opentelemetry_cpp,COMPILE_DEFINITIONS>,;-D>" /EP
+    ${CMAKE_CURRENT_SOURCE_DIR}/opentelemetry_cpp.src >
+    ${CMAKE_CURRENT_BINARY_DIR}/opentelemetry_cpp.def
```
lalitb, line 45: nit - can we use `${OPENTELEMETRY_EXPORT_DEF}` here instead of `${CMAKE_CURRENT_BINARY_DIR}/opentelemetry_cpp.def` ?

ext/src/dll/opentelemetry_cpp.src
```diff
@@ -0,0 +1,32 @@
+// clang-format off
```
lalitb, line 1: - If this is checked-in, we can add the copyright
- How is this file different from the def file, as there are two different extensions used in the PR - def and src ?
- Do we need def file explicitly when we are already specifying `__declspec(dllexport)` symbols in the code ?

lalitb:
> Having a single DDL for opentelemetry-cpp and all the exporters is still an area of discussion:
> 
> * should not be the long term goal in my opinion,
> * ok to have this single DDL as an intermediate step, assuming we document this is experimental and subject to change.

Agreed, with single DLL for opentelemetry-cpp, there may be further requests for single shared/static library for other platforms too :). As of now, building OpenTelemetry C++ as shared target(s) for Windows user is something missing, and this looks to be a promising first step towards it. Also building multiple DDL targets ( specifically for API ) would require us to either move away from header-only API support, or create source-api from header during build -  and this needs to be discussed further.
We can discuss this in the community meeting, my vote would be to provide this feature as experimental, and clearly document the constraints.

docs/build-as-dll.md
```diff
@@ -0,0 +1,11 @@
...
+
+**NOTE**: the current build produces a single DLL which includes API, SDK and
+exporters. This **MAY** change in future.
+
+OpenTelemetry C++ can be built into single DLL named opentelemetry_cpp.dll on
+Windows, with CMake flag `OPENTELEMETRY_BUILD_DLL` set to `ON`.
+
+To instrument an application based on the DLL, please define macro
+`OPENTELEMETRY_BUILD_IMPORT_DLL` before including any OpenTelemetry API header
+files in the application source files.
```
lalitb, line 11: Do you this we should update the documentation here too - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#building-shared-libs-for-windows ?

lalitb:
> (struggling with github UI, restart review to remove requested changes flag)

Have removed the flag, hope it works.

ext/src/dll/opentelemetry_cpp.src
ThomsonTan: * Added copyright.
* .src is preprocessed into .def.
* Adding to .src/.def file is needed if there is no direct reference to the class/function. `__declspec(dllexport)` doesn't guarantee the class/function will be linked to the DLL if it is not invoked.

ThomsonTan:
@owent could you please help review this PR when you have time? Thanks.

docs/build-as-dll.md
ThomsonTan: Thanks for the catching. Updated https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#building-shared-libs-for-windows.

ci/do_ci.ps1
```diff
@@ -116,6 +143,28 @@ switch ($action) {
       exit $exit
     }
   }
+  "cmake.exporter.otprotocol.dll.test" {
```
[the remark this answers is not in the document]
ThomsonTan: These 2 tests try to mirror `cmake.test` and `cmake.exporter.otprotocol.test`. Or do you mean the latter is a subset of the first one? I assumed no because OTLP is not turned on by default in the first test. Let me know if this is wrong.

I think it is a good to add test for `WITH_STL=ON`. Will see how to add that.

docs/build-as-dll.md
marcalff: Trying one more time with code, to make my suggestion more clear.

The choice between import and export is driven by which file is compiled, not by the user code.

In CMakeList.txt:

```
if(DEFINED OPENTELEMETRY_BUILD_DLL)
  if(NOT MSVC)
    message(WARNING "Build DLL is supposed to work with MSVC!")
  endif()
  if(WITH_STL)
    message(
      WARNING "Build DLL with C++ STL could cause binary incompatibility!")
  endif()
  add_definitions(-DOPENTELEMETRY_BUILD_DLL) // <-------------------------
endif()
```

In macro.h:

```
#if defined(_MSC_VER) && defined(OPENTELEMETRY_BUILD_DLL)

#ifdef OPENTELEMETRY_BUILD_EXPORT_DLL // <-------------------------------------
#  define OPENTELEMETRY_EXPORT __declspec(dllexport)
#else
#  define OPENTELEMETRY_EXPORT __declspec(dllimport)
#endif

#else

//
// build OpenTelemetry as static library or not on Windows.
//
#  define OPENTELEMETRY_EXPORT

#endif
```

And then, when compiling every file that is part of the DLL:

File `sdk/src/trace/CMakeLists.txt`

```
add_definitions(-DOPENTELEMETRY_BUILD_EXPORT_DLL) // <------------------------------
```

This totally avoids user to add `OPENTELEMETRY_BUILD_IMPORT_DLL`.

marcalff (approved):
This is buried in comments, so repeating here to avoid missing it.

Please consider this suggestion:

The choice between import and export is driven by which file is compiled, not by the user code.

In CMakeList.txt:

```
if(DEFINED OPENTELEMETRY_BUILD_DLL)
  if(NOT MSVC)
    message(WARNING "Build DLL is supposed to work with MSVC!")
  endif()
  if(WITH_STL)
    message(
      WARNING "Build DLL with C++ STL could cause binary incompatibility!")
  endif()
  add_definitions(-DOPENTELEMETRY_BUILD_DLL) // <-------------------------
endif()
```

In macro.h:

```
#if defined(_MSC_VER) && defined(OPENTELEMETRY_BUILD_DLL)

#ifdef OPENTELEMETRY_BUILD_EXPORT_DLL // <-------------------------------------
#  define OPENTELEMETRY_EXPORT __declspec(dllexport)
#else
#  define OPENTELEMETRY_EXPORT __declspec(dllimport)
#endif

#else

//
// build OpenTelemetry as static library or not on Windows.
//
#  define OPENTELEMETRY_EXPORT

#endif
```

And then, when compiling every file that is part of the DLL:

File `sdk/src/trace/CMakeLists.txt`

```
add_definitions(-DOPENTELEMETRY_BUILD_EXPORT_DLL) // <------------------------------
```

This totally avoids user to add `OPENTELEMETRY_BUILD_IMPORT_DLL`.

ThomsonTan:
Currently there are 3 macros, `OPENTELEMETRY_BUILD_DLL` which is a CMake option, `OPENTELEMETRY_BUILD_EXPORT_DLL` and `OPENTELEMETRY_BUILD_IMPORT_DLL` which are preprocess macros for C++ source files.

The suggestion seems about merging the CMake and one of the C++ macro definition which indeed can eliminate one def. So the point is that `OPENTELEMETRY_BUILD_DLL` alone implies importing OpenTelemetry DLL, and `OPENTELEMETRY_BUILD_DLL` with `OPENTELEMETRY_BUILD_EXPORT_DLL` implies building SDK itself as DLL.

My concern is that this extra implicit assumption makes the intention less clear to the user. For example, ask the user to define C++ macro `OPENTELEMETRY_BUILD_DLL` could cause confusion as the actual behavior is importing OpenTelemetry from the published DLL/LIB (nothing is forced to be built into DLL in this case).  In the other way, changing `OPENTELEMETRY_BUILD_DLL` to `OPENTELEMETRY_BUILD_IMPORT_DLL` could cause similar confusion to SDK developers. 
> This totally avoids user to add OPENTELEMETRY_BUILD_IMPORT_DLL.

lalitb (approved):

api/include/opentelemetry/common/macros.h
```diff
@@ -175,3 +175,28 @@ point.
...
+
+/* clang-format on */
+//
+// The if/elif order matters here. If both OPENTELEMETRY_BUILD_IMPORT_DLL and
+// OPENTELEMETRY_BUILD_EXPORT_DLL are defined, the former takes precedence.
+//
+// TODO: consider define OPENTELEMETRY_EXPORT for cygwin/gcc, see below link.
+// https://gcc.gnu.org/wiki/Visibility#How_to_use_the_new_C.2B-.2B-_visibility_support
+//
+#if defined(_MSC_VER) && defined(OPENTELEMETRY_BUILD_IMPORT_DLL)
```
owent, line 187: I think we can use `target_compile_definitions(opentelemetry_api INTERFACE OPENTELEMETRY_BUILD_IMPORT_DLL)` to export this definition into cmake CONFIG file, and check `OPENTELEMETRY_BUILD_EXPORT_DLL` first.And then users do not need to declare `OPENTELEMETRY_BUILD_IMPORT_DLL` themself when using cmake.

api/include/opentelemetry/logs/logger_provider.h
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
+
 #ifdef ENABLE_LOGS_PREVIEW
 
 #  include "opentelemetry/common/key_value_iterable.h"
```
owent, line 8: Should we export `LoggerProvider` ? It has static variables also.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_factory.h
```diff
@@ -17,7 +17,7 @@ namespace otlp
 /**
  * Factory class for OtlpGrpcLogRecordExporter.
  */
-class OtlpGrpcLogRecordExporterFactory
+class OPENTELEMETRY_EXPORT OtlpGrpcLogRecordExporterFactory
```
owent, line 20: Should we also export factory of metric exporters?

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
```diff
@@ -24,7 +24,7 @@ namespace logs
 /**
  * Factory class for LoggerProvider.
  */
-class LoggerProviderFactory
+class OPENTELEMETRY_EXPORT LoggerProviderFactory
```
owent, line 27: Do we need also export log processor factories?

ext/src/dll/opentelemetry_cpp.src
owent: Is there any document about how to generate this file or can we write some script to generate it automatically?It will be very helpful for developers who is not familiar with this(Such as me).The experimental APIs may still changes frequently.

owent (commented):
Just curious, is it better to export all classes and functions when building a dll?

api/include/opentelemetry/common/macros.h
ThomsonTan: Do you mean when CMake flag `OPENTELEMETRY_BUILD_DLL` is defined, add `OPENTELEMETRY_BUILD_IMPORT_DLL` to the `opentelelemtry_api` target to user to link? If so, then yes, the user doesn't need define `OPENTELEMETRY_BUILD_EXPORT_DLL` as C++ macro, but this only applies to CMake user. The current approach gives user API header files and .dll, then the user can referent it with any build system, CMake bazel or others. Just define a C++ macro `OPENTELEMETRY_BUILD_EXPORT_DLL` before including our header files will work.

But I think we can make some improvement for CMake user with `find_package(...)`. Can we do this in later PR?

api/include/opentelemetry/logs/logger_provider.h
ThomsonTan: Thanks for the catch, `LoggerProvider` should be exported. This PR makes trace and logs working, metrics will be added later based on the mechanism in this PR.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_factory.h
ThomsonTan: Yes, metrics will be in later PR.

sdk/include/opentelemetry/sdk/logs/logger_provider_factory.h
ThomsonTan: Yes, the all the log processor factories need to be exported. The `SimpleLogRecordProcessorFactory` class was in the `openteleemtry_cpp.def` file which makes the `__declspec(export)` unnecessary. But I added the export declaration to all the log record factories. Thanks for catching.

ThomsonTan:
> Just curious, is it better to export all classes and functions when building a dll?

@owent what does it mean for exporting all classes? Event the classes internal to API/SDK/Exporters? If so, this could make the output DLL very fragile as any change on the internal classes could break the binary to be loaded to the app which linked to the older DLL, even there is not API change in this repo.

ext/src/dll/opentelemetry_cpp.src
ThomsonTan: I am not sure whether there is a easy way to automatically generate this. Currently I am doing is extracting the unresolved decorated symbol from the error message as below, and check the undecorated symbol (also in the error message below) and decide whether it makes sense to export this symbol, and add it to the list finally. For example, the unresolved symbol below is https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/logs/batch_log_record_processor_factory.cc#L15. I tweaked the http_log example to call `BatchLogRecordProcessorFactory::Create`. Hope this helps.

> http_log_main.cc.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: static class std::unique_ptr<class opentelemetry::v1::sdk::logs::LogRecordProcessor,struct std::default_delete<class opentelemetry::v1::sdk::logs::LogRecordProcessor> > __cdecl opentelemetry::v1::sdk::logs::BatchLogRecordProcessorFactory::Create(class std::unique_ptr<class opentelemetry::v1::sdk::logs::LogRecordExporter,struct std::default_delete<class opentelemetry::v1::sdk::logs::LogRecordExporter> > &&,struct opentelemetry::v1::sdk::logs::BatchLogRecordProcessorOptions const &)" (__imp_?Create@BatchLogRecordProcessorFactory@logs@sdk@v1@opentelemetry@@SA?AV?$unique_ptr@VLogRecordProcessor@logs@sdk@v1@opentelemetry@@U?$default_delete@VLogRecordProcessor@logs@sdk@v1@opentelemetry@@@std@@@std@@$$QEAV?$unique_ptr@VLogRecordExporter@logs@sdk@v1@opentelemetry@@U?$default_delete@VLogRecordExporter@logs@sdk@v1@opentelemetry@@@std@@@7@AEBUBatchLogRecordProcessorOptions@2345@@Z) referenced in function "void __cdecl `anonymous namespace'::InitLogger(void)" (?InitLogger@?A0x55fc834a@@YAXXZ)

lalitb: @ThomsonTan Can we add some comment in this file to explain
  - when to add symbols in this file.
  - how to get these decorated symbols e.g., if there is tool (say objdump/ dependency walker) to get the mangled symbols from object file, and put it here. 
  
Not blocker, we can add these a separate PR, but would be helpful for other developers.

api/include/opentelemetry/common/macros.h
owent: > Do you mean when CMake flag `OPENTELEMETRY_BUILD_DLL` is defined, add `OPENTELEMETRY_BUILD_IMPORT_DLL` to the `opentelelemtry_api` target to user to link? If so, then yes, the user doesn't need define `OPENTELEMETRY_BUILD_EXPORT_DLL` as C++ macro, but this only applies to CMake user. The current approach gives user API header files and .dll, then the user can referent it with any build system, CMake bazel or others. Just define a C++ macro `OPENTELEMETRY_BUILD_EXPORT_DLL` before including our header files will work.
> 
> But I think we can make some improvement for CMake user with `find_package(...)`. Can we do this in later PR?

Yes, that's what I mean.We can do it in another PR.

owent:
> > Just curious, is it better to export all classes and functions when building a dll?
> 
> @owent what does it mean for exporting all classes? Event the classes internal to API/SDK/Exporters? If so, this could make the output DLL very fragile as any change on the internal classes could break the binary to be loaded to the app which linked to the older DLL, even there is not API change in this repo.

Thanks for your explanation.I have another question, we install all headers when we run cmake --install now.In my understanding, is it better to select a subset of headers to install(all exported classes and their dependencies). Maybe we can also improve it in another PR.

owent (approved):
LGTM

ThomsonTan:
> is it better to select a subset of headers to install(all exported classes and their dependencies)

Filed https://github.com/open-telemetry/opentelemetry-cpp/issues/1980 to track this.

Issue #1950, opened 2023-02-01
Alpine Packages


svrnm:
While playing around with the OpenTelemetry demo applications I figured out that it would be nice to have the ability to load the OpenTelemetry C++ libraries pre-compiled from a package repository like grpc, re2, etc

So instead of 

```
RUN apk add git cmake make g++ grpc-dev re2-dev protobuf-dev c-ares-dev 
RUN git clone opentelemetry-cpp ... build ... make
```

Wouldn't it be nice to do the following instead?

```
RUN apk add git cmake make g++ grpc-dev re2-dev protobuf-dev c-ares-dev opentelemetry-cpp-dev
```

I played around with that a little bit and now I have the following:

https://github.com/svrnm/otel-abuild

I consider to create a merge request with https://gitlab.alpinelinux.org/alpine/aports/, but I wanted to check with you first what you think about that and maybe if there's some feedback on what I created.

Thanks

marcalff:
Thanks @svrnm 

From a quick look:

* add license information on every file ?
* The jaeger exporter is [deprecated](https://github.com/open-telemetry/opentelemetry-cpp/issues/1892), and the build uses WITH_JAEGER=OFF, great. There should be no `opentelemetry-cpp-exporter-jaeger` package then.
* For the packages names, maybe use only the `opentelemetry-cpp-*` namespace, not `opentelemetry-*` (as seen in `RUN apk add ... opentelemetry-dev`)

For the license, I assume it will be:

```
# Copyright The OpenTelemetry Authors
# SPDX-License-Identifier: Apache-2.0
```

I have not tried the packages yet, will investigate further.

svrnm:
> add license information on every file ?

The only file that's really relevant is the APKBUILD file, everything else is just there to speed up the build. If this get's added to the official alpine repositories they are no longer needed.

> The jaeger exporter is https://github.com/open-telemetry/opentelemetry-cpp/issues/1892, and the build uses WITH_JAEGER=OFF, great. There should be no opentelemetry-cpp-exporter-jaeger package then.

I forgot to take this off the list of packages that are generated. I just upgraded to 1.8.2 and saw the deprecation notes, will remove it.

> For the packages names, maybe use only the opentelemetry-cpp-* namespace, not opentelemetry-* (as seen in RUN apk add ... opentelemetry-dev)

Also a typo, thanks!

lalitb:
[...]

PR #1957, opened 2023-02-06
Add alpine packages to INSTALL.md

svrnm:
As discussed in #1950 a minor update to INSTALL.md to also mention the alpine packages.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -320,7 +320,7 @@ with `vcpkg install opentelemetry-cpp` and follow the then displayed
 descriptions. Please see the vcpkg project for any issues regarding the
 packaging.
 
-If you are using [alpine linx](https://www.alpinelinux.org/) you can install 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -320,6 +320,10 @@ with `vcpkg install opentelemetry-cpp` and follow the then displayed
 descriptions. Please see the vcpkg project for any issues regarding the
 packaging.
 
+If you are using [alpine linx](https://www.alpinelinux.org/) you can install
```
marcalff, line 323: There is a trailing space, making the format CI fail.

Also, `alpine linux` for the link name ?

marcalff, line 323:
```suggestion
If you are using [alpine linux](https://www.alpinelinux.org/) you can install
```

marcalff (approved):
LGTM, thanks the the fix, and for the packages.

esigo (approved):
LGTM
Thanks for the package :)

PR #1958, opened 2023-02-06
Update Bazel build support for jaeger exporter with an explicit depen…

bjschnei:
…dency on boost(Linux only).

## Changes

Previously boost libraries had to be installed on the build machine.  This  change allows Bazel to pull it in.


lalitb:
@bjschnei Changes look good, but Jaeger exporter is already deprecated, and no new changes are accepted - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/DEPRECATED.md#opentelemetry-cpp-exporter. Jaeger natively supports OTLP format, and so OTLP exporter should be used to export spans.

lalitb:
Close as discussed above.

PR #1963, opened 2023-02-06
Update copyright for the remaining repo

ThomsonTan:
This added copyright header for all files based on below 3 PRs.

https://github.com/open-telemetry/opentelemetry-cpp/pull/1960
https://github.com/open-telemetry/opentelemetry-cpp/pull/1961
https://github.com/open-telemetry/opentelemetry-cpp/pull/1962

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #1965, opened 2023-02-07
Enforce copyright check in CI

marcalff:
Enforce copyright check in CI.

## Changes

Added:

* a Copyright check in CI,
* a `tools/check_copyright.sh` script,
* file `.copyright-ignored` to allow for copyright exceptions.

Checklist:

* [X] Changes in public API reviewed (no changes)

### DIFF
--- a/.copyright-ignore
+++ b/.copyright-ignore
@@ -1,18 +1,22 @@
 # Copyright 2023, The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-## Ignore the following directories
+# Ignore the following directories
 
 ./.git/*
 ./.github/*
 ./third_party/*
 ./tools/vcpkg/*
+./tools/ports/*
 
 # Third party code
+
 ./api/include/opentelemetry/nostd/internal/absl/*
 ./exporters/jaeger/thrift-gen/*
+./exporters/etw/include/opentelemetry/exporters/etw/TraceLoggingDynamic.h
 
 # Doc
+
 ./docs/*
 
 ## Ignore the following files patterns
@@ -23,9 +27,13 @@
 *.log
 *.patch
 *.json
+*.nuspec
 
-## Ignore the following files
+# Ignore the following files
 
+./.bazelignore
+./.bazelversion
 ./LICENSE
 ./exporters/etw/include/opentelemetry/exporters/etw/LICENSE
+./docker/.gitignore
 
--- a/.copyright-ignore
+++ b/.copyright-ignore
@@ -1,0 +1,34 @@
+# Copyright 2023, The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+## Ignore the following directories
+
+./.git/*
+
+./third_party/*
+
+./tools/vcpkg/*
+
+./api/include/opentelemetry/nostd/internal/absl/*
+
+./exporters/jaeger/thrift-gen/*
+
+## Ignore the following files patterns
+
+# Markdown (unsure how to add comments in markdown)
+
+*.md
+
+# Images
+
+*.png
+
+# Log files
+
+*.log
+
+## Ignore the following files
+
+./LICENSE
+./exporters/etw/include/opentelemetry/exporters/etw/LICENSE
+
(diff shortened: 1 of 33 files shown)


marcalff:
With the initial commit, the Copyright CI fails, as expected, because all files do not have a proper copyright.

This is an opportunity to check the logs in the failed CI.

After PR #1963 is merged, the list of issues should shrink, and list only remaining items.

marcalff:
Ready for review.

tools/check_copyright.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+# For all files in /tmp/all_checked
+# - verify there is copyright
+# - verify there is a license
+# and append to /tmp/all_missing
+#
+# Valid copyright strings are:
+# - Copyright The OpenTelemetry Authors
+# - Copyright YEAR, The OpenTelemetry Authors
+# - Copyright YEAR, YEAR, The OpenTelemetry Authors
```
lalitb, line 55: Is the copyright string with YEAR special case for trace_id_ratio.cc, or there are other files too ?

lalitb (approved):
LGTM. nicely done, thanks.

tools/check_copyright.sh
marcalff: As usual with these cases, upfront warning : I am not a lawyer, nor do I represent one, etc.

Historically, the copyright in opentelemetry-cpp uses the first form, without year, for the most part.

In my experience however, I have always seen copyright applied with either:

* the year it is claimed (second form)
* the first year it is claimed, followed by the most recent year it is updated (third form).

For example, code created in 2020 will say "2020".

When later changed in 2021, it will say "2020, 2021".

When later changed in 2022, it will say "2020, 2022", and not use "2020, 2021, 2022".

Note that the Apache V2 license itself reads:

```
   APPENDIX: How to apply the Apache License to your work.

      To apply the Apache License to your work, attach the following
      boilerplate notice, with the fields enclosed by brackets "[]"
      replaced with your own identifying information. (Don't include
      the brackets!)  The text should be enclosed in the appropriate
      comment syntax for the file format. We also recommend that a
      file or class name and description of purpose be included on the
      same "printed page" as the copyright notice for easier
      identification within third-party archives.

   Copyright [yyyy] [name of copyright owner]
```

So, we are supposed to add a YEAR, for example:

`Copyright 2023, The OpenTelemetry Authors`

About file trace_id_ratio.cc, it contains copyright to OpenCensus as well as opentelemetry-cpp, so I wanted to preserve the dates for extra clarity.

My preference would be to use forms "YEAR" and "YEAR, YEAR" form everywhere, it is not specific to a file.

marcalff: Decision from today's SIG C++ meeting is to keep only `Copyright The OpenTelemetry Authors`.

Will adjust the patch accordingly.

.copyright-ignore
```diff
@@ -0,0 +1,45 @@
+# Copyright 2023, The OpenTelemetry Authors
```
ThomsonTan, line 1: Is year(2023) needed here?

.copyright-ignore
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Ignore the following directories
+
+./.git/*
+./.github/*
+./third_party/*
```
ThomsonTan, line 8: nit: under third_party, there is BUILD file which is still for this repo and needs check. Could we specify some files are checked explicitly even matching the ignore pattern? But it is Ok to not have this.

ThomsonTan (approved):

tools/check_copyright.sh
marcalff: Fixed

.copyright-ignore
marcalff: Fixed by using this instead:

```
./third_party/benchmark/*
./third_party/boost/*
./third_party/googletest/*
./third_party/ms-gsl/*
./third_party/nlohmann-json/*
./third_party/opentelemetry-proto/*
./third_party/prometheus-cpp/*
```

.copyright-ignore
marcalff: Fixed.

PR #1991, opened 2023-02-18
[ETW Exporter]Support serialize span/log attributes into JSON

### LINKED ISSUE #1990: [ETW Exporter] Support serializing span/log attributes to a single column in JSON
Currently each span/log attribute is output as a separate column in ETW exporter, this make it possible for schema explosion if there are a lot of attributes and make it hard to process in the backend. We should provider an option for the user to serialize all the attributes into a single column in JSON.

ThomsonTan:
Fixes #1990 

## Changes

As ETW exporter is header only, this PR provide a C++ macro `ENABLE_ENV_PROPERTIES`, when defined, does serialization on all span/log attributes to a single attribute with JSON representation of all the attributes.


exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -250,8 +250,41 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+        }
+      }
+      if (has_customer_attribute)
+      {
+        env_properties_env[ETW_FIELD_ENV_PROPERTIES] = env_properties_json.dump();
+        evt = std::move(env_properties_env);
+      }
+    }
+
+#endif  // defined(ENABLE_ENV_PROPERTIES)
```
lalitb, line 286: Do we also need to handle Span::SetAttribute() method ?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -250,8 +250,41 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+    {
+      bool has_customer_attribute = false;
+      nlohmann::json env_properties_json = nlohmann::json::object();
+      for (auto &kv : evt)
+      {
+        nostd::string_view key = kv.first.data();
+
+        // don't serialize fields propagated from span data.
+        if (key == ETW_FIELD_NAME || key == ETW_FIELD_SPAN_ID || key == ETW_FIELD_TRACE_ID ||
+            key == ETW_FIELD_SPAN_PARENTID)
```
lalitb, line 269: what should be the behavior if key is `env_properties` - extract the value from it, and append it to existing `env_properties` ? Current behavior looks good to if we document that.

ThomsonTan: Yes, `env_properties` set by user will be nested. I think this is expected and has parity with Linux, because `Span::SetAttribute` on Linux doesn't parse the attribute name with special handling.

Are you aware of any appropriate document (link) for this?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
ThomsonTan: I think no based on the current implementation, because all the attributes will be processed at `EndSpan` time.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
lalitb: No, I think it should be fine to be consistent with fluentd.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
lalitb: ok yeah good point. Thanks.

lalitb (approved):
LGTM Thanks for the fix.

PR #1992, opened 2023-02-19
Add example for Logs ostream exporter

### LINKED ISSUE #1883: [Logs SDK] Add example for Logs ostream exporter
We have:
Simple Span ostream exporter example - https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples/simple
Simple Metrics ostream exporter example - https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples/metrics_simple

Similar example code needs to be added for Logs Ostream Exporter too.

MrSparc:
Fixes #1883

## Changes

Add example for Logs ostream exporter 


examples/logs_simple/BUILD
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    tags = [
+        "examples",
+        "ostream",
+    ],
+    deps = [
+        "//api",
+        "//examples/common/logs_foo_library:common_logs_foo_library",
+        "//exporters/ostream:ostream_log_record_exporter",
+        "//exporters/ostream:ostream_span_exporter",
```
lalitb, line 17: Should we use this PR only for logs, as there is already [span ](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/examples/simple)example.

MrSparc: @lalitb in this example is initializing a trace provider because the example is reusing logs_foo_library that inject in the log the reference about the current  `trace_id `and `span_id `: https://github.com/open-telemetry/opentelemetry-cpp/blob/95f6c66de253e46d09b232facc840b311a4a354d/examples/common/logs_foo_library/foo_library.cc#L35
I think that's relevant for the simple log example about how to relate logs with spans and follows the existing approach done in OTLP examples for logs.

But feel free to ignore my PR If this is not the intended example.

lalitb: You are right. The PR looks good to me, thanks for the contribution.

marcalff (approved):
LGTM, thanks for the contribution

lalitb (approved):

PR #2011, opened 2023-03-02
Rename the global SDK version variables to avoid naming clash

ThomsonTan:
## Changes

Avoid using the general macro like name as `MAJOR_VERSION`. It could clash with the macro definition form other header files even it is under namespace. Add the prefix `OPENTELEMETRY_SDK_` to make the name more specific.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ Increment the:
 
 ## [Unreleased]
 
+* [SDK] Rename the global SDK version variables to avoid naming clash
+  [#2011](https://github.com/open-telemetry/opentelemetry-cpp/pull/2011)
 * [SDK]Add attributes for InstrumentationScope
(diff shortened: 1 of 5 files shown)


lalitb:
Can we also update the pre-commit script here - https://github.com/open-telemetry/opentelemetry-cpp/blob/3ef579bca397817358c06a55761b17091d2af8bd/buildscripts/pre-commit#L64

ThomsonTan:
> Can we also update the pre-commit script here -
> 
> https://github.com/open-telemetry/opentelemetry-cpp/blob/3ef579bca397817358c06a55761b17091d2af8bd/buildscripts/pre-commit#L64

Thanks for the catch, updated.

lalitb:
Looks good to me. Just need to be aware if some external libraries are using this macro to detect otel-version, it would break them.

owent (approved):

PR #2014, opened 2023-03-03
Provide version major/minor/patch macros

### LINKED ISSUE #2012: Provide major/minor/patch version macros

In Apache Arrow Flight C++, we'd like to support a wide range of versions of opentelemetry-cpp. Sometimes there are breaking changes in the API, such as in #1761. It's cool that there are breaking changes (I appreciate the pace at which the project is evolving), except that in our project we don't have a way to check the version and easily adapt our code to support versions both before and after the change.

It's common in many C++ projects to provide macros for:

 * `OPENTELEMETRY_VERSION_MAJOR`
 * `OPENTELEMETRY_VERSION_MINOR`
 * `OPENTELEMETRY_VERSION_PATCH`

Then, downstream libraries can write code like:

```cpp
#if !(OPENTELEMETRY_VERSION_MAJOR <= 2 && OPENTELEMETRY_VERSION_MINOR < 5)
// Handle old API
#else
// Handle new API
#endif
```

Would you be open to providing these?

We could use CMake [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html) feature to detect what APIs are available at compile time, but that's quite burdensome to maintain.

marcalff:
Fixes #2012

## Changes

Added the following macros:

* OPENTELEMETRY_VERSION_MAJOR
* OPENTELEMETRY_VERSION_MINOR
* OPENTELEMETRY_VERSION_PATCH

This is helpful for user code, to support multiple versions in case of incompatible changes.


esigo (approved):
LGTM
Thanks for the PR :)

api/include/opentelemetry/version.h
```diff
@@ -8,6 +8,10 @@
 
 #define OPENTELEMETRY_ABI_VERSION_NO 1
 #define OPENTELEMETRY_VERSION "1.8.2"
```
ThomsonTan, line 10: Use # operator for preprocessor to define OPENTELEMETRY_VERSION with the new VERSION macros?

lalitb (approved):

api/include/opentelemetry/version.h
marcalff: Technically possible, but the resulting code, with `OPENTELEMETRY_CONCAT()` and `OPENTELEMETRY_STRINGIFY()`, will be extremely obfuscated.

There is no risk that the string version goes out of sync with major/minor/patch versions, there is a unit test to catch this.

ThomsonTan (approved):

PR #2022, opened 2023-03-07
Add `OPENTELEMETRY_INSTALL` to allow user to skip install targets.

### LINKED ISSUE #2015: Add CMake option to skip export and install

If so, provide a concise description of the problem.

My current project uses "FetchContent" to introduce the GRPC library and skip the Install step.
I found that this library will default "Install(Export)", which causes me to export GRPC targets too.
I want to skip "export" and "install" in this library to avoid a target name clash.

What do you want to happen instead? What is the expected behavior?0

I would like to have the option to skip it, just like "gRPC_INSTALL_default" in the GRPC library.

Otherwise, I have to rename a lot of the target names of the existing project.

**Additional context**

I'm not sure if this is a good idea, do you have any other recommendations?

owent:
Fixes #2015 

## Changes

+ Add `OPENTELEMETRY_INSTALL` to allow user to skip install targets.


marcalff (approved):
LGTM, thanks for the fix.

CMakeLists.txt
```diff
@@ -118,6 +118,14 @@ option(WITH_GSL
 
 option(WITH_ABSEIL "Whether to use Abseil for C++latest features" OFF)
 
+if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  set(OPENTELEMETRY_INSTALL_default ON)
+else()
+  set(OPENTELEMETRY_INSTALL_default OFF)
```
dbolduc, line 124: (Maybe this is too late now, but) Do you think this should default to ON in all cases for backwards compatibility?

Do you think the new feature should be listed in INSTALL.md?

https://github.com/open-telemetry/opentelemetry-cpp/blob/f7026762506fd41608da323ddb170c491451393c/INSTALL.md?plain=1#L75-L86

lalitb: > (Maybe this is too late now, but) Do you think this should default to ON in all cases for backwards compatibility?

True, somehow missed this during review, and I broke one of my libraries with this change :(

PR #2025, opened 2023-03-07
Fix compile error with GCC 11.3

tyler92:
Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/2026

## Changes

There was a compile error with GCC 11.3 and the following cmake options:

```
-DBUILD_TESTING=OFF -DWITH_EXAMPLES_HTTP=ON -DWITH_PROMETHEUS=ON -DWITH_STL=ON
```

GCC output:

```
In file included from opentelemetry-cpp/examples/http/client.cc:7:
opentelemetry-cpp/examples/http/tracer_common.h:29:26: error: expected ‘)’ before ‘&’ token
   29 |   HttpTextMapCarrier<T>(T &headers) : headers_(headers) {}
```

See https://godbolt.org/z/Pv3fqq7zx


marcalff (approved):
LGTM, thanks for the fix.

lalitb (approved):

PR #2031, opened 2023-03-08
[REMOVAL] Remove the jaeger exporter

### LINKED ISSUE #1938: [REMOVAL] Remove the Jaeger Exporter
This removal follows the deprecation announced in:

[DEPRECATION] Deprecate jaeger exporter #1892

marcalff:
Fixes #1938

## Changes

* Remove the jaeger exporter
* Remove usage of thrift
* Remove usage of boost


esigo (approved):
LGTM
Thanks for the PR :)

CMakeLists.txt
```diff
@@ -175,8 +175,6 @@ option(WITH_ELASTICSEARCH
 
 option(WITH_ZPAGES "Whether to include the Zpages Server in the SDK" OFF)
 
-option(WITH_JAEGER "DEPRECATED - Whether to include the Jaeger exporter" OFF)
```
ThomsonTan, line 178: Could we keep the option but only allow the default value? I mean if we delete the option, the existing cmake invocation like `cmake -DWITH_JAEGER=OFF` will have warning triggered. The same applies to `cmake -DWITH_JAEGER=ON`. Better to show an explicit error message the user tries to enable Jaeger?

lalitb:
Thanks for the PR. Changes look good. Was trying to see if the specs says anything about removing support of Jaeger . There are couple of reference - 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#jaeger-exporter

```
Jaeger exporter support will be removed from OpenTelemetry in July 2023.
```

And changelog - https://github.com/open-telemetry/opentelemetry-specification/blob/b674b49a9be32bf38a58865e9a04803c0f9349fb/CHANGELOG.md?plain=1#LL213C3-L213C70

```
- Deprecate jaeger exporter, scheduled for spec removal in July 2023.
```

Will it make sense to merge this PR when the Jaeger exporter reference is removed from specs ?

marcalff:
> 
> ```
> - Deprecate jaeger exporter, scheduled for spec removal in July 2023.
> ```
> 
> Will it make sense to merge this PR when the Jaeger exporter reference is removed from specs ?

Looks like I missed the date from the spec when writing the C++ deprecation PR, which is why the planned removal date for C++ was set to April, 2023 instead.

Ok to wait for July 2023 then.

CMakeLists.txt
marcalff: > Could we keep the option but only allow the default value? I mean if we delete the option, the existing cmake invocation like `cmake -DWITH_JAEGER=OFF` will have warning triggered. The same applies to `cmake -DWITH_JAEGER=ON`. Better to show an explicit error message the user tries to enable Jaeger?

Thanks for the early comments.

The initial deprecation was announced with release 1.8.2 published on 2023-01-31.
It was advertised at great length, including with pinned issues, so users should be well aware of this by now.

With this removal, the following commands:

```
cmake -DWITH_JAEGER=ON
cmake -DWITH_JAEGER=OFF
```

will both produce a warning, with CMake indicating that an unknown option is used:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    WITH_JAEGER
```

Using `-DWITH_JAEGER=OFF` produces a warning only, and will not cause the build to fail.

The fact that the `WITH_JAEGER` option is no longer available is now documented in the CHANGELOG, in the important changes section, for the next release.

I don't think the CMakeList.txt should still contain logic about obsolete options, as it makes _removing_ options impossible, this is mitigated by the important changes in the CHANGELOG instead.

marcalff:
Now ready for review.

marcalff:
TODO, waiting on spec changes.

The Jaeger Propagator will not be removed from the spec:
* https://github.com/open-telemetry/opentelemetry-specification/pull/3567

so it needs to be preserved in opentelemetry-cpp as well.

marcalff:
do not merge yet, minor rework on Jaeger Propagator needed.

Will clear the do not merge flag once resolved.

marcalff:
Jaeger Propagator is no longer removed, per spec.

Updated DEPRECATED.md to indicate the propagator stays deprecated in opentelemetry-cpp (not removed)

ThomsonTan (approved):
Thanks for driving this change, @marcalff

lalitb (approved):
LGTM.  Nicely done, as usual :)

PR #1953, opened 2023-02-03
Convert Prometheus Exporter to Pull MetricReader

lalitb:
Fixes #1941

## Changes
Changes are as explained here - https://github.com/open-telemetry/opentelemetry-cpp/issues/1941#issuecomment-1413236782

Prometheus Exporter now implements `MetricReader` class, instead of `PushMetricExporter`. And the metrics is collected only when the actual http pull request is received from Prometheus  agent to fetch the metrics. Earlier metrics were fetched periodically by PeriodicMetricReader and stored into intermediate storage, from where the Prometheus agent used to pull the data. This resulted in duplicate entries of data in the intermediate storage.

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [x] Changes in public API reviewed

PR #2034, opened 2023-03-09
Fix view names in Prometheus example

tyler92:
As far as I can see View should have the same name as Instrument

> The name of the View (optional). If not provided, the **Instrument name** MUST be used by default. This will be used as the name of the [metrics stream](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#events--data-stream--timeseries).

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view

### DIFF
--- a/examples/prometheus/README.md
+++ b/examples/prometheus/README.md
@@ -59,7 +59,7 @@ Start the application and keep it running. Now we should be able to see the
 metrics at [http://localhost:9464/metrics](http://localhost:9464/metrics) from a
 web browser:
 
-![Browser UI](https://user-images.githubusercontent.com/71217171/168492500-12bd1c99-33ab-4515-a294-17bc349b5d13.png)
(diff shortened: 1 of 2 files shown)


tyler92:
I am a newcomer in OpenTelemetry so please let me know if I am wrong and if there is no error in the examples. Without this change I don't see "_counter" on the Prometheus page.

lalitb (approved):
Changes look good. Thanks for the contribution.

tyler92:
If the change is good, I will change also README with pictures. And revome WIP after that. Right now I wanted to make sure that the fix is correct

tyler92:
Why don't I see a picture like in README? I see lines instead of histograms. Where is a mistake?

image

lalitb:
@tyler92   In case you are seeing a separate line for histogram bucket, the output is correct. The existing screenshot in README.md is from the previous implementation (prior to design changes in #1953), and is no longer valid output.

PR #2055, opened 2023-03-15
Remove unused 'alerting' section from prometheus.yml in examples

tyler92:
Alerting is not used in this example and port 9464 can't be reused for this purpose


lalitb (approved):
Thanks.

esigo (approved):

PR #2060, opened 2023-03-15
[EXPORTER] GRPC endpoint scheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE

### LINKED ISSUE #2058: OTLP exporter assumes it must use SSL because GetOtlpDefaultIsSslEnable() is returning a wrong value
**Describe your environment** 

- opentelemetry-cpp v1.8.3
- OTLP gRPC
- No SSL
- No OTLP env vars, just default settings 

**Steps to reproduce**
I'm facing an issue with SSL after upgrade from v1.8.2 to v1.8.3 with my gRPC telemetry.
The error is:
```
E0315 15:58:51.193959723  145649 ssl_transport_security.cc:1501] Handshake failed with fatal error SSL_ERROR_SSL: error:1408F10B:SSL routines:ssl3_get_record:wrong version number.
```

So debugging and comparing the behavior of both versions, I noticed that in v1.8.2 `OtlpGrpcExporterOptions` constructor is setting as default `use_ssl_credentials: false` but in the new **v1.8.3**  this default constructor is setting to `use_ssl_credentials: true`

This looks like a behaviour change introduced by this refactoring handling Boolean environment variables  https://github.com/open-telemetry/opentelemetry-cpp/pull/1982

Digging into the code to look the initialization of `use_ssl_credentials` in the constructor:
https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h#L19-L25

This method  `GetOtlpDefaultIsSslEnable()` to initialize `use_ssl_credentials` is based on `GetOtlpDefaultTracesIsInsecure()`:

https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h#L60-L64

The issue is in the default return value of the following methods: 

- `GetOtlpDefaultTracesIsInsecure()` 
  https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/src/otlp_environment.cc#L251
- `GetOtlpDefaultMetricsIsInsecure()` 
  https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/src/otlp_environment.cc#L295
- `GetOtlpDefaultLogsIsInsecure()`
  https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/src/otlp_environment.cc#L312

All these methods are returning by default  `false`  if no env vars are defined.
So a `false` return means, IS NOT insecure, then IS secure :exploding_head: 

I don't know the gRPC transpost must be by default secure or insecure, but if the default is secure then this behaviour was enforced in this version.

According to the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md): 
_Insecure: Whether to enable client transport security for the exporter's gRPC connection. This option only applies to OTLP/gRPC when an endpoint is provided without the http or https scheme_

So, if the gRPC endpoint scheme returns 'http' or 'https' as the following default, I think this value should overrided:
https://github.com/open-telemetry/opentelemetry-cpp/blob/15ab9ed9adb8150ff1f487a7e31303d6c12e863e/exporters/otlp/src/otlp_environment.cc#L78-L82

Upgrade from v1.8.2 to v1.8.3 without issues.

Upgrade to v1.8.3 wrongly assumes that gRPC connection are using SSL and fails because no certificates are defined.

**Additional context**
As workaround I need to set the env var `OTEL_EXPORTER_OTLP_INSECURE=true` in order to do not use SSL

marcalff:
Fixes #2058

## Changes

The logic to decide whether or not an OLTP GRPC exporter uses SSL has changed, to comply with the specification.

In particular, the endpoint scheme ("http" or "https") takes precedence.

* [X] Changes in public API reviewed (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,12 @@ Increment the:
 
 * [RESOURCE SDK] Fix schema URL precedence bug in `Resource::Merge`.
   [#2036](https://github.com/open-telemetry/opentelemetry-cpp/pull/2036)
-* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
+* [EXPORTER] GRPC endpoint scheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
   [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
 
 Important changes:
 
-* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
+* [EXPORTER] GRPC endpoint scheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
   [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
   * The logic to decide whether or not an OTLP GRPC exporter uses SSL has
     changed to comply with the specification:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,31 @@ Increment the:
 
 * [RESOURCE SDK] Fix schema URL precedence bug in `Resource::Merge`.
   [#2036](https://github.com/open-telemetry/opentelemetry-cpp/pull/2036)
+* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
+  [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
+
+Important changes:
+
+* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
+  [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
+  * The logic to decide whether or not an OTLP GRPC exporter uses SSL has
+    changed to comply with the specification:
+    * Before this change, the following settings were evaluated, in order:
+      * OTEL_EXPORTER_OTLP_TRACES_INSECURE (starting with 1.8.3)
+      * OTEL_EXPORTER_OTLP_INSECURE (starting with 1.8.3)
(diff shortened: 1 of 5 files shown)


compoundradius:
I tested this fix, and it works properly in my environment.

CHANGELOG.md
```diff
@@ -17,6 +17,31 @@ Increment the:
 
 * [RESOURCE SDK] Fix schema URL precedence bug in `Resource::Merge`.
   [#2036](https://github.com/open-telemetry/opentelemetry-cpp/pull/2036)
+* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
+  [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
+
+Important changes:
+
+* [EXPORTER] GRPC endpoint sheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
```
lalitb, line 25: nit - scheme

lalitb (approved):
LGTM. Thanks for the quick fix.

CHANGELOG.md
MrSparc: Probably is also a good idea include in the OTLP exporter doc  about `use_ssl_credentials` in https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/README.md with the same info about sheme precedence   described in the changelog with env vars.

marcalff: > nit - scheme

Thanks, fixed.

PR #2062, opened 2023-03-16
Fix default value of `OPENTELEMETRY_INSTALL_default`.

### LINKED ISSUE #2061: `OPENTELEMETRY_INSTALL_default` should be `ON` when oetl-cpp is on top level.
We should still install otel-cpp by default when otel-cpp is on the top level. Just like `gRPC_INSTALL_default`.

owent:
Fixes #2061 

## Changes

The default value of `OPENTELEMETRY_INSTALL_default` should be `ON` when otel-cpp in on top level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Increment the:
   [#2036](https://github.com/open-telemetry/opentelemetry-cpp/pull/2036)
 * [EXPORTER] GRPC endpoint scheme should take precedence over OTEL_EXPORTER_OTLP_TRACES_INSECURE
   [#2060](https://github.com/open-telemetry/opentelemetry-cpp/pull/2060)
+* [BUILD] Restore detfault value of `OPENTELEMETRY_INSTALL` to `ON` when it's on
+  top level.[#2062](https://github.com/open-telemetry/opentelemetry-cpp/pull/2062)
 
 Important changes:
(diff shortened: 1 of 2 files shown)


marcalff (commented):
What about simplifying to:

```
option(OPENTELEMETRY_INSTALL "Whether to install opentelemetry targets" ON)
```

and let users override the value to avoid installation if they want to, based on their own logic when used in super projects.

owent:
> What about simplifying to:
> 
> ```
> option(OPENTELEMETRY_INSTALL "Whether to install opentelemetry targets" ON)
> ```
> 
> and let users override the value to avoid installation if they want to, based on their own logic when used in super projects.

In my opnion, it's easier for users to enable the installation when it's the top level, because most libraries will include it too by default.
Both ways are fine to me.

marcalff (approved):
Both solutions:

* patch as proposed
* simple always ON option by default

looks good to me.

Waiting on input from other reviewers, ok to merge in both cases.

lalitb (approved):

lalitb:
@owent  - Can you resolve the CHANGELOG.md conflict, will enable to merge this PR.

owent:
> @owent - Can you resolve the CHANGELOG.md conflict, will enable to merge this PR.

Done nad thanks.

PR #2064, opened 2023-03-16
[Metrics SDK] Implement Forceflush for Periodic Metric Reader

### LINKED ISSUE #2056: Error from Metricreader::Collect at shutdown when using PeriodicExportingMetricReader
**Describe your environment** 
macOS Ventura, otel-cpp 1.8.3

**Steps to reproduce**
Run the `metrics_simple` example

Example should run without error printouts

During shutdown, the error `MetricReader::Collect invoked while Shutdown in progress!` is printed.

**Additional context**
I found this while implementing metrics in a project I work on, but I then discovered that the metrics_simple example does the same.

I’m using a `PeriodicExportingMetricReader`. An lldb run confirms that `PeriodicExportingMetricReader::OnShutDown` is being called, but the log message is emitted during the `join()` call there on the thread. A backtrace shows that `OnShutDown` is called from `MetricReader::Shutdown()`, but that doesn’t happen until after `MetricReader::shutdown_` is already set to true. If `PeriodicExportingMetricReader::OnShutDown` does one more call to `Collect` during the join (which I wouldn’t be surprised if it does), then it’s going to always output that error.

lalitb:
Fixes #2056

## Changes

As of now, Periodic Metric reader tries to perform one last collection and export during Shutdown. And this fetch fails with warning that the Shutdown is already in progress.

The correct approach should be NOT to do this fetch during shutdown, instead implement PeriodicMetricReader::ForceFlush() method to perform the fetch whenever invoked by application. And then application can just invoke this method before doing the shutdown.

WIP as need to add the unit-tests for this.

### DIFF
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -66,11 +66,7 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
         is_force_wakeup_background_worker_.store(false, std::memory_order_release);
         return true;
       }
-      if (IsShutdown())
-      {
-        return true;
-      }
-      return false;
+      return IsShutdown();
     });
   } while (IsShutdown() != true);
 }
@@ -151,7 +147,7 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
     timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
   }
 
-  // If it will be already signaled, we must wait util notified.
+  // If it will be already signaled, we must wait until notified.
   // We use a spin lock here
   if (false == is_force_flush_pending_.exchange(false, std::memory_order_acq_rel))
   {
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -165,7 +165,7 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
       }
     }
   }
-  
+
   is_force_flush_notified_.store(false, std::memory_order_release);
   if (result)
   {
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -169,8 +169,8 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
   is_force_flush_notified_.store(false, std::memory_order_release);
   if (result)
   {
-
-    result = exporter_->ForceFlush(timeout);
+    result = exporter_->ForceFlush(
+        std::chrono::duration_cast<std::chrono::microseconds>(timeout_steady));
   }
   return result;
 }
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -165,17 +165,7 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
       }
     }
   }
-
   is_force_flush_notified_.store(false, std::memory_order_release);
-
-  if (timeout == std::chrono::steady_clock::duration::zero())
-  {
-  }
-  if (timeout_steady <= std::chrono::steady_clock::duration::zero())
-  {
-    // forceflush timeout, exporter force-flush won't be triggered.
-    result = false;
-  }
 
   if (result)
   {
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -91,6 +91,7 @@ bool PeriodicExportingMetricReader::CollectAndExportOnce()
       return true;
     });
   });
+
   std::future_status status;
   do
   {
@@ -129,9 +130,9 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
     }
     return is_force_flush_notified_.load(std::memory_order_acquire);
   };
+
   timeout = opentelemetry::common::DurationUtil::AdjustWaitForTimeout(
       timeout, std::chrono::microseconds::zero());
-
   std::chrono::steady_clock::duration timeout_steady =
       std::chrono::duration_cast<std::chrono::steady_clock::duration>(timeout);
   if (timeout_steady <= std::chrono::steady_clock::duration::zero())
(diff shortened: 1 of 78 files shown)


lalitb:
This is ready for review now.

lalitb:
@owent - Once you have time, can you shortly look into the `PeriodicMetricReader::ForceFlush()`, as the method is inspired from the batch span/log processor method, and you know that part of code better :)

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -96,12 +102,77 @@ bool PeriodicExportingMetricReader::CollectAndExportOnce()
...
+        std::this_thread::yield();
+      }
+    }
+  }
+
+  is_force_flush_notified_.store(false, std::memory_order_release);
+  if (result)
+  {
+
+    result = exporter_->ForceFlush(timeout);
```
owent, line 173: Is it better to use `timeout_steady` if it's greater than zero here? We have already wait some time before and the left timeout should be less.

lalitb: Good point. Have changed it to use `timeout_steady` now.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -169,8 +169,8 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
   is_force_flush_notified_.store(false, std::memory_order_release);
   if (result)
   {
-
-    result = exporter_->ForceFlush(timeout);
+    result = exporter_->ForceFlush(
+        std::chrono::duration_cast<std::chrono::microseconds>(timeout_steady));
```
owent, line 173: I think we should check and ensure the `timeout_steady` to be greater than zero, some exporters use zero and negative value as indefinitely waiting.

lalitb: What's your suggestion - 
 - if the original `timeout` as argument to period-reader::forceflush is zero, we can pass that directly to the exporter::forceflush, irrespective of the value of `timeout_steady`.
 - else if the timeout_steady is zero or less, we can return error, and don't call exporter::forceflush
 - else call exporter::forceflush with non-zero timeout_steady

owent: Looks good.

lalitb: Thanks, have updated the code as per above logic, let me know if that looks fine :)

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -167,10 +167,36 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
   }
 
   is_force_flush_notified_.store(false, std::memory_order_release);
+
+  if (timeout == std::chrono::steady_clock::duration::zero())
```
owent, line 171: Do we need remove this if branch?It does nothing.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -167,10 +167,36 @@ bool PeriodicExportingMetricReader::OnForceFlush(std::chrono::microseconds timeo
...
+  {
+  }
+  if (timeout_steady <= std::chrono::steady_clock::duration::zero())
+  {
+    // forceflush timeout, exporter force-flush won't be triggered.
+    result = false;
```
owent, line 177: Do we need reset result to false here?I think result will keep false and break the loop above when `timeout_steady <= std::chrono::steady_clock::duration::zero()`.
And if result is true and `timeout_steady <= std::chrono::steady_clock::duration::zero()`, result will also be set to false below.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
lalitb: yes removed thanks.

owent (approved):
LGTM now, thanks.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -60,14 +60,19 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
     auto end            = std::chrono::steady_clock::now();
     auto export_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
     auto remaining_wait_interval_ms = export_interval_millis_ - export_time_ms;
-    cv_.wait_for(lk, remaining_wait_interval_ms);
+    cv_.wait_for(lk, remaining_wait_interval_ms, [this]() {
+      if (is_force_wakeup_background_worker_.load(std::memory_order_acquire))
+      {
+        is_force_wakeup_background_worker_.store(false, std::memory_order_release);
```
ThomsonTan, line 66: Can store fail then false should be returned?

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -60,14 +60,19 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
...
+        return true;
+      }
+      if (IsShutdown())
```
ThomsonTan, line 69: nit: `return IsShutdown();`.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -96,12 +102,93 @@ bool PeriodicExportingMetricReader::CollectAndExportOnce()
...
+  {
+    // When is_force_flush_notified_.store(true) and force_flush_cv_.notify_all() is called
+    // between is_force_flush_pending_.load() and force_flush_cv_.wait(). We must not wait
+    // for ever
+    std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
+    result = force_flush_cv_.wait_for(lk_cv, export_interval_millis_, break_condition);
+    timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
+  }
+
+  // If it will be already signaled, we must wait util notified.
```
ThomsonTan, line 154:
```suggestion
  // If it will be already signaled, we must wait until notified.
```

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
lalitb: fixed. thanks

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
lalitb: fixed. thanks.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
lalitb: I don't know if it can ever fail. It's atomic assignment operation so should eventually be successful.  Also, this method returns `void`, not sure how to get any failure status from this method.
Let me know if I am missing something :)

PR #2080, opened 2023-03-30
Make Windows build environment parallel

### LINKED ISSUE #2077: Some of CI build targets are not compiled in parallel (Windows)
**Describe your environment** For example, when run ./ci/do_ci.ps1 cmake.maintainer.test on my local machine, it takes too much time to build from scratch.

**Steps to reproduce**
run ./ci/do_ci.ps1 cmake.maintainer.test

Target should be built with parallel option of make command if possible

Some of targets in CI pipeline does not have any parallel options which takes too much time to build from scratch.

**Additional context**

cngzhnp:
Fixes #2077

## Changes

### DIFF
--- a/ci/do_ci.ps1
+++ b/ci/do_ci.ps1
@@ -40,7 +40,7 @@ switch ($action) {
     if ($exit -ne 0) {
       exit $exit
     }
-    cmake --build .
+    cmake --build . -j
     $exit = $LASTEXITCODE
     if ($exit -ne 0) {
       exit $exit
@@ -61,7 +61,7 @@ switch ($action) {
     if ($exit -ne 0) {
       exit $exit
     }
-    cmake --build .
+    cmake --build . -j
     $exit = $LASTEXITCODE
     if ($exit -ne 0) {
       exit $exit
@@ -90,7 +90,7 @@ switch ($action) {
     if ($exit -ne 0) {
       exit $exit
     }
-    cmake --build .
+    cmake --build . -j
     $exit = $LASTEXITCODE
     if ($exit -ne 0) {
       exit $exit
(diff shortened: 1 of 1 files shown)


ci/do_ci.ps1
```diff
@@ -40,7 +40,7 @@ switch ($action) {
     if ($exit -ne 0) {
       exit $exit
     }
-    cmake --build .
+    cmake --build . -j
```
lalitb, line 43: Just curious, Will CMake automatically use Ninja as build system, or we need to configure it explicitly?
The build time is reduced from 14 min to 10 min for Windows, so there seems to be some level of  parallelism.

cngzhnp: With my solution (getting number of physical cores from Powershell command), I assume that it takes less time than before, what do you think?

lalitb (approved):

ci/do_ci.ps1
cngzhnp: Yeah, sorry for inconvenience. After discussion on issue, I decided to use Ninja but it does not work at all. So, I will update the PR title and do you want me to change commit message either?

esigo:
Are we sure we have enough RAM on the runners to use all the cores?

cngzhnp:
> Are we sure we have enough RAM on the runners to use all the cores?

What do you mean by that? I do not understand the relation between RAM and parallel build. If you say that, one core must be isolated for other jobs on the system, I can get it but RAM sorry I do not. Could you please elaborate more?

esigo:
parallel builds require more RAM for build. The more core you use for build the more RAM you should have. Normal GHA runners can't do more than 3 parallel jobs. I'm not sure what setup we have here.

ThomsonTan (approved):

PR #2086, opened 2023-04-07
Use sdk_start_ts for MetricData start_ts for instruments having cumulative aggregation temporality.

davidmandle:
## Changes

The [OpenTelemetry data model for metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#metric-points) states that "with cumulative aggregation temporality ... we expect to report the full sum since 'start'", and the accompanying diagram illustrates the same `start_time` present in each report.

However, prior to this change, cumulative metrics were being reported with a start time equal to the last report time. This broke compatibility with services expecting the documented behavior. (For example, without this change, Counter metrics submitted to the Google Cloud Monitoring API via the googlecloudexporter are not represented correctly). 


lalitb (approved):
Thanks for the fix.

davidmandle:
Thanks

marcalff (approved):
LGTM, thanks for the fix.

PR #2094, opened 2023-04-13
Add user facing Logging API and Benchmarks

### LINKED ISSUE #2093: Add user facing logging API for OpenTelemetry C++

The OpenTelemetry Spec currently defines [bridge-api](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md) which is not intended to be called by application developers. OpenTelemetry C++ needs create its own user facing C++ for application developer to instrument their code.

Design a new set of logging API to support common scenario based on the bridge API.

Existing API has drawbacks to be used in production for large volume instrumentation.

**Additional context**

ThomsonTan:
Fixes #2093 

## Changes

This PR adds a new set of user facing logging API for application developer to instrument their code. This new set of API is designed for production environment with large volume of code instrumented while maintaining high throughput.

Here are the main design considerations:
* Add optional event Id to the log inside `Logger` object because categorization is very important for large volume of logs data. It helps both in doing runtime filtering in SDK/exporters and in backend grouping/querying.
* Added `Enabled` predicate on `Logger` to check if log should be sent to SDK for processing. It is assumed that most instrumented logs will not be turned on so it is fast-path for such code with minimum runtime performance overhead.
* Multi-thread safety is built into the design and the same logger object can be used in multiple threads/cores.
* Benchmark is added to measure the runtime cost of API calls. It shows the fast path check only takes a few machine cycles (~5).

This is only for API change, and SDK change will be added later.


api/include/opentelemetry/logs/logger.h
```diff
@@ -238,18 +239,238 @@ class Logger
...
+
+  /**
+   * Log an event
+   *
+   * @severity severity of the log
+   * @event_id event identifier of the log
+   * @format an utf-8 string following https://messagetemplates.org/
+   * @attributes key value pairs of the log
+   */
+  virtual void Log(Severity severity,
```
owent, line 285: Do we need these overloaded functions? We can set the data in the type trait, so we don't have to write a lot of overloaded functions for all parameter variation combinations.

api/include/opentelemetry/logs/severity.h
```diff
@@ -16,8 +16,8 @@ namespace logs
  */
 enum class Severity : uint8_t
 {
-  kInvalid,
-  kTrace,
+  kInvalid = 0,
+  kTrace   = 1,
```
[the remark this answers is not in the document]
ThomsonTan: I am fine on each way. For the current change, `kInvlid` is assigned 0 because it is not given in the OTel spec, and only created for us, so makes it explicit. for `kTrace`, it is the first severity defined in OTel spec with value. It doesn't break the spec if we change the value of `kInvalid` later.

api/include/opentelemetry/common/macros.h
```diff
@@ -7,9 +7,7 @@
 // GCC 9 has likely attribute but do not support declare it at the beginning of statement
 #  if defined(__has_cpp_attribute) && (defined(__clang__) || !defined(__GNUC__) || __GNUC__ > 9)
 #    if __has_cpp_attribute(likely)
-#      define OPENTELEMETRY_LIKELY_IF(...) \
-        if (__VA_ARGS__)                   \
-        [[likely]]
+#      define OPENTELEMETRY_LIKELY_IF(...) if (__VA_ARGS__) [[likely]]
```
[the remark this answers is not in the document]
ThomsonTan: I agree the suggestion totally make sense and I was considering rewrote it. But since it is not introduced here and used somewhere else, perhaps make the change in a different PR? Then that PR will be straightforward and could be merged before the current PR.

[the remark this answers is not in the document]
ThomsonTan: One more thing to add. The macro tries to prefer the likely attribute, `if (cond) [likely]` which cannot be defined like the same as `likely/unlikely` macros. Perhaps this is the reason to define it with if condition check.

api/include/opentelemetry/logs/logger.h
ThomsonTan: @owent, as we are defining user facing API, I'd prefer make the API signature explicit. Type traits could save us all the variant of overloads, but still not all parameter variation combinations makes sense, and we need add more code to the type traits for such validation. Compare with this, probably overload is more straightforward, and doesn't incur runtime overhead (overload resolution happens at compile time). Please let me know what you think about this.

lalitb: Explicit arguments with defined order makes sense for user facing API to me if we have limited number of arguments.

api/test/logs/logger_test.cc
```diff
@@ -79,36 +87,65 @@ TEST(Logger, LogMethodOverloads)
                 opentelemetry::common::MakeAttributes({{"key1", "value 1"}, {"key2", 2}}));
   logger->Trace(m);
   logger->Trace(opentelemetry::common::MakeAttributes({{"key1", "value 1"}, {"key2", 2}}));
+  logger->Trace(trace_event_id, "Test log message",
+                opentelemetry::common::MakeAttributes({{"key1", "value 1"}, {"key2", 2}}));
```
lalitb, line 91: Just to add, if we are not using the variadic arguments in user-facing API, we can also add few more overloading to simplify the API by removing the use of `opentelemetry::common::MakeAttributes`. So something below should work:

```
logger->Trace(trace_event_id, "Test log message", {{"key1", "value 1"}, {"key2", 2}});
```

api/include/opentelemetry/common/macros.h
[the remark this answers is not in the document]
owent: > One more thing to add. The macro tries to prefer the likely attribute, `if (cond) [likely]` which cannot be defined like the same as `likely/unlikely` macros. Perhaps this is the reason to define it with if condition check.

Yes, it's.

api/test/logs/logger_test.cc
owent: There are at least 1 type of event id, 1 type of trace infomation, 1 type of timestamp, 1 type of body, 3 types of attributes, 6 different severities and a `Log` function with severity parameter by noww.
Every field is optional, then we have to write `2*2*2*2*4*7=448` functions for every paramenters.
Do you think this will also confuse users and also make us hard to maintain if we add more special attributes in the future?

lalitb: Yeah good point. This will indeed increase number of overloads.

api/include/opentelemetry/common/macros.h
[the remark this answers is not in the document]
marcalff: Maybe just:

```
#define OPENTELEMETRY_LIKELY [[likely]]
```

so this can be used in if, but also in switch case, since this attribute is per branch (not expression as I assumed)

To address separately anyway, setting this issue resolved.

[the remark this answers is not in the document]
owent: > Maybe just:
> 
> ```
> #define OPENTELEMETRY_LIKELY [[likely]]
> ```
> 
> so this can be used in if, but also in switch case, since this attribute is per branch (not expression as I assumed)
> 
> To address separately anyway, setting this issue resolved.

How can we adapt the likely/unlikely in GCC/Clang through this approach? We hope to maintain consistency in the usage of both C++ attributes and traditional compiler built-in macros.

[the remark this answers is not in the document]
ThomsonTan: BTW, created a new issue to track this.

https://github.com/open-telemetry/opentelemetry-cpp/issues/2098

marcalff (commented):
There is a point in the design that I do not understand.

Method Logger::SetMinimumSeverity() is protected, so it can not be invoked by the instrumented application, this is not part of the api.

It sounds like minimum_severity_, a class member, is part of the implementation already.

Why making this code visible in the API, instead of implementing the minimum_severity_ member in the SDK, in a sub class of Logger ?

Instead of using atomic macros in the API, the SDK could use std::atomic instead, and this will avoid exposing all the per platform and compiler code in the API itself.

Please clarify the intent here.

ThomsonTan:
> There is a point in the design that I do not understand.
> 
> Method Logger::SetMinimumSeverity() is protected, so it can not be invoked by the instrumented application, this is not part of the api.
> 
> It sounds like minimum_severity_, a class member, is part of the implementation already.
> 
> Why making this code visible in the API, instead of implementing the minimum_severity_ member in the SDK, in a sub class of Logger ?
> 
> Instead of using atomic macros in the API, the SDK could use std::atomic instead, and this will avoid exposing all the per platform and compiler code in the API itself.
> 
> Please clarify the intent here.

Yes, `SetMinimumSeverity()` is not part of the user-facing API. It is intended to SDK or any one which implements this API to use internally.

And you are right that `minimum_severity_` is part of implementation instead of API, but it is added here for getting the best performance on the common path (no virtual function call is needed if the logger is not enabled which I assume is the common case for most components). So in brief, providing implementation of the minimum severity check in this abstract logger is for eliminating the virtual dispatch for common cases/scenarios so the runtime cost of instrumentation will be minimum.

It is also correct that if `minimum_severity_` is moved to SDK, the atomic support will be much easier, but calling into SDK causes performance overhead which I want to avoid here.

Please let me know if anything I missed here.

api/include/opentelemetry/logs/logger.h
```diff
@@ -246,10 +247,225 @@ class Logger
...
+
+  inline void Fatal(nostd::string_view message) noexcept { this->Log(Severity::kFatal, message); }
+
+  //
+  // End of OpenTelemetry C++ user-facing Log API.
+  //
+
+protected:
+  // TODO: discuss with community about naming for internal methods.
+  virtual bool EnabledImplementation(Severity /*severity*/,
```
reyang, line 442: Instead of having a function which child classes can derive, I wonder if it makes sense to use a function pointer (similar to the severity level field which can be set by the implementations).

api/test/logs/logger_benchmark.cc
```diff
@@ -0,0 +1,260 @@
...
+  auto logger = lp->GetLogger("UnstructuredLogWithEnabledReturnFalse");
+
+  for (auto _ : state)
+  {
+    state.PauseTiming();
+
+    MultiThreadRunner(state, [&logger]() {
+      for (int64_t i = 0; i < kMaxIterations; i++)
+      {
+        if (logger->Enabled(Severity::kTrace, 0x12345678), false)
```
marcalff, line 240: Does this really build ? How ?

api/include/opentelemetry/logs/logger.h
```diff
@@ -246,10 +247,225 @@ class Logger
...
+
+  // Convenient wrappers based on virtual Log() virtual methods.
+  // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitynumber
+
+  inline void Trace(const EventId &event_id,
+                    nostd::string_view format,
+                    const common::KeyValueIterable &attributes) noexcept
+  {
+    this->Log(Severity::kTrace, event_id, format, attributes);
+  }
```
marcalff, line 315: I am confused about the intended usage for this helper:

Is the caller supposed to check Enabled for each call, as in:

```
if (logger->Enabled(xxx)) {
  logger->Trace(xxx);
}
```

or is Trace() responsible to check the severity flag itself, as in:

```
logger->Trace(xxx);
```

The benchmarks shows both usage, is this code writing un wanted traces in the second case ?
The check for Enabled() seems missing from the code path.

Beside, if the whole point of Logger::minimum_severity_ is to check the flags in the API before making a virtual function call, then why making calls here to virtual functions Log() and EmitLogRecord() ?

Suggested changes:

* Call Enabled() in every user facing api.
* Do not call a user facing api like Log() from another user facing api like Trace(), call EmitLogRecord directly, so the Enabled() flags are not checked multiple times.

For example:

```
inline void Trace(xxx) {
  if (Enabled(xxx)) {
    EmitLogRecord(xxx);
  }
}
```

marcalff (commented):
More comments / questions

marcalff:
> Yes, SetMinimumSeverity() is not part of the user-facing API. It is intended to SDK or any one which implements this API to use internally.

Thanks for the clarifications.

I can see how multiple threads may use the same Logger instance, if GetLogger() returns a shared object used concurrently.

What is missing here is details about when SetMinimumSeverity() is called.

It is really the case that SetMinimumSeverity() can be called concurrently on a logger already created ? How ?

Could Logger::minimum_severity_ be set in the Logger constructor instead, assuming it is constant for a given logger instance.

This will remove all the atomic code entirely.

api/include/opentelemetry/logs/event_id.h
```diff
@@ -0,0 +1,35 @@
...
+ * EventId class which acts the Id of the event with an optional name.
+ */
+class EventId
+{
+public:
+  EventId(int64_t id, nostd::string_view name) noexcept
+  {
+    id_   = id;
+    name_ = nostd::unique_ptr<char[]>{new char[name.length() + 1]};
+    std::copy(name.begin(), name.end(), name_.get());
```
lalitb, line 25: This goes against the standard practice of not doing any memory copy at the API/SDK surface. The recordable interface ensures that we directly serialize the data at exporter, without creating such intermediate copies. One solution can be to get rid of EventId class, and pass the event-id and event-name directly to Log methods. Also we only use PODs, or `nostd` types as API arguments. Even though we can debate that passing reference/pointer to a user created class mayn't break the ABI, in general good to avoid if we can.

api/test/logs/logger_benchmark.cc
ThomsonTan: Thanks for the catch. It builds but the trailing `, false` should be removed. The variation comes from an iteration of experiment change, but I just validated that the benchmark result is consistent after removing the extra experiment code.

api/include/opentelemetry/logs/logger.h
ThomsonTan: Thanks for the suggestion @marcalff . 

> Call Enabled() in every user facing api.
`Enabled()` should be called by the user's code instead of from our all our wrappers which is intentional. The idea is to avoid calculating any parameters which is only for logging. In the simple case, `common::MakeAttributes({"key1", “value1}, {"key2", "value2"}}` takes significant extra time than the `Enabled()` check. And there could be many other heavy query result which passed as parameters to query. Calling `Enabled()` in user's code can eliminate such heavy query from the first place.

```c++
logger->Trace("some message template", common::MakeAttributes({"some_value_takes_time_calculate", long_query()}));
``` 

If we are not going to call `Enabled()` in user facing API, then inline the call of `DoLogRecord` seems unnecessary for all the wrappers then? I don't have strong preference as the name implied they are just wrappers. But I can also revisit this part in the SDK support because changing the implementation will not break the user facing API.

Please let me know if anything is missed here.

api/include/opentelemetry/logs/event_id.h
ThomsonTan: > This goes against the standard practice of not doing any memory copy at the API/SDK surface. 

No memory will be allocated in our logging APIs. The `EventId` class is supposed to be created by the user with sort of global lifetime, it so copies the string instead of using any `string_view` on existing `string` because it then requires the user to manage the original string object explicitly. The allocation and free of such object is not in the scope of logging API as it only takes const reference on such object.

ThomsonTan:
> > Yes, SetMinimumSeverity() is not part of the user-facing API. It is intended to SDK or any one which implements this API to use internally.
> 
> Thanks for the clarifications.
> 
> I can see how multiple threads may use the same Logger instance, if GetLogger() returns a shared object used concurrently.
> 
> What is missing here is details about when SetMinimumSeverity() is called.
> 
> It is really the case that SetMinimumSeverity() can be called concurrently on a logger already created ? How ?
> 
> Could Logger::minimum_severity_ be set in the Logger constructor instead, assuming it is constant for a given logger instance.
> 
> This will remove all the atomic code entirely.

` Logger::minimum_severity_` will only be available for SDK to config (SDK can choose how to present this config to the user. I am working on this). It is not supposed to be a constant after creation of logger object and SDK can configure it as runtime, then atomic r/w will be required for this.

api/include/opentelemetry/logs/logger.h
ThomsonTan: Assume this is for internal implementation so doesn't block the PR.

The virtual function here looks a restricted function pointer which is stored in the virtual table. If we store the pointer here like `class Logger`, it seems more flexible like not SDK but any other library can provide a function to do real `Enabled()` work. It might also be more risky if the function pointer is compromised then it can invoke anything as code. The virtual function is safer in this regard as it cannot be patched at runtime, and it is usually in the protected path.

reyang: I actually consider the patch as a bless than a curse. (and I don't see why virtual functions cannot be patched)

api/include/opentelemetry/logs/severity.h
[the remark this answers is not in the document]
reyang: +1 on being explicit

api/include/opentelemetry/logs/logger.h
ThomsonTan: Virtual function tables are in read only area of memory in modern operation system which cannot be patched normally.

But I think the idea still applies to provider a mechanism for any related part to register its own `Enabled()` check to the API `Logger`. Opened https://github.com/open-telemetry/opentelemetry-cpp/issues/2105 to track this because it is **NOT** API breaking change.

api/include/opentelemetry/logs/severity.h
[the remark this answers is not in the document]
ThomsonTan: Assigned explicit number value to each severity level.

owent:
Just wondering, do you think we can add a new class to contains these functions?The name `event` seems different from the [event in specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md) .

api/include/opentelemetry/logs/logger.h
owent: Just want to have a discuss.
We have a log macro like this before in our project.

```cpp
#define WLOGDEBUG(logger, ...) \
  if(logger.check_level(log_level::DEBUG)) {
    logger->log(log_level::DEBUG, __VA_ARGS__);
  }
```

And when we use `WLOGDEBUG(logger, "Debug message: {}", protobuf_message.DebugString());` and if the debug log is not enabled, `protobuf_message.DebugString()` will not be called and it's a high cost function.

Is there any way we can achieve a similar effect?

ThomsonTan:
> Just wondering, do you think we can add a new class to contains these functions? The name `event` seems different from the [event in specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/event-api.md?rgh-link-date=2023-04-20T02%3A44%3A56Z) .

The OTel event API is still in experiment, and like the log bridge API, I think it is also not user-facing API. And as per the below spec issue, the log bridge API is reaching stable state which we can build our user-facing API upon.

https://github.com/open-telemetry/opentelemetry-specification/issues/2911

With layering of the architecture, our user-facing API is in the upper layer, so ideally we should consider the requirement of users to write logging code, even we've already review the current bridge API spec and made sure it is sufficient to implement our API. The bridge API is in the middle layer supporting the user facing API, so even in future the event API reach stable, we can just update our SDK like to send logs with event Id to that event API, which will help user logging user to upgrade seamlessly.

Finally, I think provide 2 sets of similar API, for logs and events, to user will cause more confuse than the benefit it gives, and it hinders user to switch between logs and events if do so which adds unnecessary complexity (like if something starts as log, as evolving happens, existing it to log makes sense, but the 2 API and separate class approach will require the user to acquire a new `EventLogger` and call some totally different methods. But with our current approach, the only change will be adding `event_id` to the parameter list.

Let me know if I missed anything here.

api/include/opentelemetry/logs/logger.h
ThomsonTan: Perhaps we can define similar macro to help the user reduce boilerplate code, but it can avoid the logging cost. Based on the logging statement you shared, the the variable `protobuf_message` might be only need for logging, and it could do heavy initialization (like map or load string resources). This happens in the definition of `protobuf_message` which cannot be be optimized here.

```c++
  WLOGDEBUG(logger, "Debug message: {}", protobuf_message.DebugString());
```

While providing macros could help, the pointer in our API is make `Enabled` check explicit to the user, so they are aware of this and encouraged to do the optimization to archive high performance logging.

api/include/opentelemetry/logs/event_id.h
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+public:
+  int64_t id_;
+  nostd::unique_ptr<char[]> name_;
```
owent, line 30: Why not use `std::string name_;` here? The usage of this variable above seems just like std::string, and do not add a tail `\0` ?

ThomsonTan: `EventId` can be read from both user and SDK code, which requires it to be ABI compatible. `std::string` cannot guarantee such ABI compatibility requirement across the boundary between user code/SDK so cannot be used. I'd expect there is `nostd::string` but we don't have it, then switch to primitive `char` pointer wrapped in `nostd::unique_ptr`. Let me know if this addresses the concern.

ThomsonTan: Yes, that makes sense. Set the last element of `name_` as 0.

api/include/opentelemetry/logs/logger.h
marcalff: About the name, `EnabledImplementation`, discussion is deferred to the SDK patch if I understand correctly.

Using a function callback in a class member instead of a virtual method is asking for too much trouble in my opinion:
* what if it changes, should the pointer itself be looked up using atomics, similar to the severity ?
* Even after an atomic read, is the pointer still valid anyway (for all we know, it can be in a shared library), so is it safe to invoke ?

In short, the current proposal (virtual method) looks good to me, apart from the name to discuss later.

api/include/opentelemetry/logs/logger.h
```diff
@@ -246,10 +247,225 @@ class Logger
...
+                   const common::KeyValueIterable &attributes) noexcept
+  {
+    this->EmitLogRecord(severity, event_id, format, attributes);
+  }
+
+  virtual void Log(Severity severity,
+                   int64_t event_id,
+                   nostd::string_view format,
+                   const common::KeyValueIterable &attributes) noexcept
+  {
```
marcalff, line 291: Do we need each of these helpers to be virtual functions ?

A simple inline function seem sufficient.

api/include/opentelemetry/logs/logger.h
marcalff: Let's continue the discussion about Enabled() in the next patch, for the SDK part.

The fact that the caller MAY use Enabled() to optimize paths that evaluate complex arguments is good.

Still we can't force that the caller MUST use Enabled() all the time, so the implementation will need to check it again at some point.

marcalff (approved):
Thanks for the clarifications.

Looks ok for now, so approving the PR.

Some items here in fact depends on the intended implementation from the SDK,
so I expect we will be able to revisit particular points later, when not affecting the usage of logger but only the implementation from the logger subclass.

owent (approved):
LGTM and thanks.

lalitb (approved):

malkia:
I was reading above, but I still don't get why std::atomic wasn't used here.

PR #2097, opened 2023-04-15
Allow build shared `opentelemetry_proto` on non-Windows platform.

### LINKED ISSUE #2096: Allow to build proto as dynamic on non-Windows platform.
This is from slack channel <https://cloud-native.slack.com/archives/C01N3AT62SJ/p1681345822148859>

### LINKED ISSUE #2139: [BUILD] OTLP GRPC exporter fails to build with non default GRPC install
Using the OTLP GRPC exporter.

GRPC is installed in a non default location, for example `/opt/grpc-1.46.4`

In CMake,

`gRPC_DIR` is set to `/opt/grpc-1.46.4/lib/cmake/grpc`

During the build:

```
malff@localhost.localdomain:build-otelcpp-grpc> make
Consolidate compiler generated dependencies of target opentelemetry_proto
[  0%] Building CXX object CMakeFiles/opentelemetry_proto.dir/generated/third_party/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.grpc.pb.cc.o
In file included from /home/malff/CODE/MARC_GITHUB/build-otelcpp-grpc/generated/third_party/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.grpc.pb.cc:6:
/home/malff/CODE/MARC_GITHUB/build-otelcpp-grpc/generated/third_party/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.grpc.pb.h:25:10: fatal error: 'grpcpp/impl/codegen/async_generic_service.h' file not found
#include <grpcpp/impl/codegen/async_generic_service.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

C++ code is properly generated from protobuf files.

The generated C++ code fails to build, because GRPC headers can not be found, as they are in a non default location, not added in the makefiles.

Suggested fix:

```
diff --git a/cmake/opentelemetry-proto.cmake b/cmake/opentelemetry-proto.cmake
index 1aa1ba03..7addca8b 100644
--- a/cmake/opentelemetry-proto.cmake
+++ b/cmake/opentelemetry-proto.cmake
@@ -285,6 +285,10 @@ else() # cmake 3.8 or lower
   target_link_libraries(opentelemetry_proto INTERFACE ${Protobuf_LIBRARIES})
 endif()
 
+if(WITH_OTLP_GRPC)
+  target_link_libraries(opentelemetry_proto PUBLIC gRPC::grpc++)
+endif()
+
 if(BUILD_SHARED_LIBS)
   set_property(TARGET opentelemetry_proto PROPERTY POSITION_INDEPENDENT_CODE ON)
 endif()
```

owent:
Fixes #2096 
Fixes #2139 

## Changes

+ Allow build shared `opentelemetry_proto` on non-Windows platform.
+ Add `opentelemetry_proto_grpc` which contains generated grpc files (Which depends gRPC) to make OTLP Http exporter do not depends gRPC when otel-cpp is built as dynamic libraries.

*Sorry for push a wrong repo by mistake before, I have moved the `fix_2096` branch into my fork.*


cmake/opentelemetry-proto.cmake
```diff
@@ -282,9 +292,21 @@ if(TARGET protobuf::libprotobuf)
...
   target_include_directories(opentelemetry_proto
                              PUBLIC ${Protobuf_INCLUDE_DIRS})
-  target_link_libraries(opentelemetry_proto INTERFACE ${Protobuf_LIBRARIES})
+  target_link_libraries(opentelemetry_proto PUBLIC ${Protobuf_LIBRARIES})
+endif()
+
+if(WITH_OTLP_GRPC)
+  if(WITH_ABSEIL)
+    find_package(absl CONFIG)
+    if(TARGET absl::synchronization)
```
ThomsonTan, line 301: Is the target `absl::synchronization` optional, or error out if it doesn't exist?

cmake/opentelemetry-proto.cmake
```diff
@@ -229,31 +229,41 @@ endif()
...
+unset(OTELCPP_PROTO_TARGET_OPTIONS)
+if(CMAKE_SYSTEM_NAME MATCHES "Windows|MinGW|WindowsStore")
+  list(APPEND OTELCPP_PROTO_TARGET_OPTIONS STATIC)
+elseif(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
+  list(APPEND OTELCPP_PROTO_TARGET_OPTIONS SHARED)
+else()
+  list(APPEND OTELCPP_PROTO_TARGET_OPTIONS STATIC)
+endif()
+
+list(APPEND OTELCPP_PROTO_TARGET_OPTIONS)
```
lalitb, line 241: Not sure if I am missing something, but how are we using `OTELCPP_PROTO_TARGET_OPTIONS`

lalitb:
@owent As I understand, this PR adds two new shared libs - opentelemetry_proto and opentelemetry_proto_grpc. And these shared libraries would be used by otlp_recordable, otlp_http_client and otlp_grpc_client. How much saving in file size are we seeing with these changes ? As this could be a breaking change if users are hardcoding the required otel libraries in their build, we should ensure there is sufficient saving with the change.

cmake/opentelemetry-proto.cmake
owent: Sorry, it's a mistake after merging.

owent:
> @owent As I understand, this PR adds two new shared libs - `opentelemetry_proto` and `opentelemetry_proto_grpc`. And these shared libraries would be used by `otlp_recordable`, `otlp_http_client` and `otlp_grpc_client` targets. How much reduction in total otel lib size we are seeing with this change ? As this could be a breaking change if users are hardcoding the required otel libraries in their build, we should have a sufficient total size reduction to reason for a breaking change.

Spliting `opentelemetry_proto` into `opentelemetry_proto` and `opentelemetry_proto_grpc` is used to avoid to depend on gRPC when users only use OTLP HTTP exporters. The generated gRPC sources in `opentelemetry_proto_grpc` depend gRPC.
If users use both OTLP HTTP and gRPC exporters, I think the total size will not decreased(at lease in my tests it do not decrease), but we can allow users use only one of them to reduce the final package size.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Increment the:
   [#2123](https://github.com/open-telemetry/opentelemetry-cpp/pull/2123)
 * [BUILD] Build break with old curl, macro CURL_VERSION_BITS unknown
   [#2102](https://github.com/open-telemetry/opentelemetry-cpp/pull/2102)
+* [BUILD] Allow build shared opentelemetry_proto on non-Windows platform.
```
lalitb, line 22: Do you think we should mentioned this PR potential breaking change here in changelog, as we have are introducing a new library?

cmake/opentelemetry-proto.cmake
lalitb: Just curious the motive of this change - was there any build error with WITH_ABSEIL?

CHANGELOG.md
owent: > Do you think we should mentioned this PR potential breaking change here in changelog, as we have are introducing a new library?
Sure, the changelog is added now.

> Just curious the motive of this change - was there any build error with WITH_ABSEIL?

Yes, the related CI job is here https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/4707416429/jobs/8349260349 .

lalitb (approved):
LGTM. Thanks

marcalff (approved):
LGTM, thanks for the fix for 2139 as well.

PR #2102, opened 2023-04-19
[BUILD] Build break with old curl, macro CURL_VERSION_BITS unknown

### LINKED ISSUE #2101: [BUILD] Build break with old curl, macro CURL_VERSION_BITS unknown
When compiling code that depends on CURL, like the OTLP HTTP exporter, the build fails for old curl.

The issue is that file:
`ext/src/http/client/curl/http_operation_curl.cc`
contains uses of the CURL macro `CURL_VERSION_BITS`.

This macro was introduced only in CURL 7.43.0, so building with older curl breaks.

Fix:
* Keep using CURL_VERSION_BITS for readability
* define the macro if missing.

Curl commit that introduced this macro:

```
commit 39b9bf60d13ff7cb62a6a031c210d27a32113c4f
Author: Daniel Stenberg <daniel@haxx.se>
Date:   Thu May 14 23:33:27 2015 +0200

    curlver: introducing new version number (checking) macros

```

marcalff:
Fixes #2101

## Changes

Provide macro `CURL_VERSION_BITS`, when building old (< 7.43.0) CURL versions.

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -8,6 +8,11 @@
 #include "opentelemetry/ext/http/client/curl/http_client_curl.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
 
+// CURL_VERSION_BITS was added in CURL 7.43.0
+#ifndef CURL_VERSION_BITS
+#  define CURL_VERSION_BITS(x, y, z) ((x) << 16 | (y) << 8 | (z))
+#endif
+
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace ext
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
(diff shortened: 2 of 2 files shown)


owent (approved):
LGTM

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -8,6 +8,11 @@
 #include "opentelemetry/ext/http/client/curl/http_client_curl.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
 
+// CURL_VERSION_BITS was added in CURL 7.43.0
```
ThomsonTan, line 11: Check the version of CURL before define CURL_VERSION_BITS?

lalitb (approved):

ext/src/http/client/curl/http_operation_curl.cc
marcalff: The code uses ifndef, and only adds the macro if it is not provided already by the curl headers included earlier.

I think this is the most robust form, testing the curl version is more error prone.

ThomsonTan (approved):

Issue #1429, opened 2022-06-03
Move implementation from SDK header files to SDK *.cc


marcalff:
This is a remaining item identified during review of PR #1420

For all the static variables found in SDK header files,
consider (to discuss) to either:
- move the code to the *.cc files instead,
- remove the static keyword,
- add comments to explain why use of static is safe.

List of variables identified:

File `opentelemetry/sdk/common/empty_attributes.h`

```
  static const std::array<std::pair<std::string, int>, 0> array{};
  static const opentelemetry::common::KeyValueIterableView<
      std::array<std::pair<std::string, int>, 0>>
      kEmptyAttributes(array);
```

File `opentelemetry/sdk/metrics/aggregation/drop_aggregation.h`

```
    static DropPointData point_data;
```

File `opentelemetry/sdk/metrics/exemplar/always_sample_filter.h`

```
    static nostd::shared_ptr<ExemplarFilter> alwaysSampleFilter{new AlwaysSampleFilter{}};
```

File `opentelemetry/sdk/metrics/view/view_registry.h`

```
      static View view("otel-default-view");
```

File `opentelemetry/sdk/resource/experimental_semantic_conventions.h`

```
static const std::unordered_map<uint32_t, const char *> attribute_ids = { ... };
```

This one (`attribute_ids`) probably needs more investigation before doing any changes (strings resolved at compile time or runtime ?).

File `opentelemetry/sdk/trace/multi_recordable.h`

```
    static std::unique_ptr<Recordable> empty(nullptr);
```

File `opentelemetry/sdk/trace/span_data.h`

```
      static opentelemetry::sdk::resource::Resource resource =

      static std::unique_ptr<opentelemetry::sdk::instrumentationlibrary::InstrumentationLibrary>
          instrumentation_library =
```

Code cleanup, low priority.

marcalff:
As discussed in community meeting, expanding the scope from "moving static variables to *.cc" to "moving all code to *.cc".

PR #2109, opened 2023-04-21
Fix: #2100 Building shared libraries with version suffix, along with the symbolic link

### LINKED ISSUE #2100: [Build] Building shared libraries with version suffix, along with the symbolic link
A recommended practice for the shared libraries deployed in the unix-like systems is have the libraries suffixed with version, along with the symbolic links pointing to them.
Something like - 

```
libopentelemetry_trace.so  -> libopentelemetry_trace.so.1 -> libopentelemetry_trace.so.1.9.0
```

It would be good to update the CMake build to generate them. This could be a build breaking change - as libopentelemetry_trace.so would now become the sym link, and not the actual file. so this could be introduced under separate cmake option.

This issue is specifically for CMake. The bazel build change can be introduced with separate issue.

lalitb:
Fixes #2100

## Changes

Versioning policy is described here for debian - https://www.debian.org/doc/debian-policy/ch-sharedlibs.html. In general same is followed across other *unix distributoins. 
- Added CMake option OTELCPP_VERSIONED_LIBS to enable generating versioned shared libs.
- Currently using OPENTELEMETRY_ABI_VERSION_NO as SOVERSION, we can also use MAJOR version. But as per the recommendation, SOVERSION should change for any ABI breaking change.
- The output is as below:

```console
$ ls -ltr /usr/local/lib/libopentelemetry_*common*so*
-rw-r--r-- 1 root root 281640 Apr 20 21:59 /usr/local/lib/libopentelemetry_common.so.1.9.0
lrwxrwxrwx 1 root root     32 Apr 20 22:01 /usr/local/lib/libopentelemetry_common.so.1 -> libopentelemetry_common.so.1.9.0
lrwxrwxrwx 1 root root     28 Apr 20 22:01 /usr/local/lib/libopentelemetry_common.so -> libopentelemetry_common.so.1

$ objdump -p /usr/local/lib/libopentelemetry_common.so.1.9.0 | grep SONAME
  SONAME               libopentelemetry_common.so.1
$
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,7 +274,7 @@ function(set_target_version target_name)
   if(OTELCPP_VERSIONED_LIBS)
     set_target_properties(
       ${target_name} PROPERTIES VERSION ${OPENTELEMETRY_VERSION}
-                                SOVERSION ${OPENTELEMETRY_VERSION})
+                                SOVERSION ${OPENTELEMETRY_ABI_VERSION_NO})
   endif()
 endfunction()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -191,6 +191,10 @@ option(WITH_OPENTRACING "Whether to include the Opentracing shim" OFF)
 option(OTELCPP_VERSIONED_LIBS "Whether to generate the versioned shared libs"
        OFF)
 
+if(OTELCPP_VERSIONED_LIBS AND NOT BUILD_SHARED_LIBS)
+  message(FATAL_ERROR "OTELCPP_VERSIONED_LIBS=ON requires BUILD_SHARED_LIBS=ON")
+endif()
+
 set(OTELCPP_PROTO_PATH
     ""
     CACHE PATH "Path to opentelemetry-proto")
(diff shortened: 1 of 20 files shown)


CMakeLists.txt
```diff
@@ -188,6 +188,9 @@ option(OTELCPP_MAINTAINER_MODE "Build in maintainer mode (-Wall -Werror)" OFF)
 
 option(WITH_OPENTRACING "Whether to include the Opentracing shim" OFF)
 
+option(OTELCPP_VERSIONED_LIBS "Whether to generate the versioned shared libs"
+       OFF)
+
```
marcalff, line 193: This option only applies for builds using shared libraries.

Suggestion for the "Verify options dependencies" section in CMakeLists.txt below.

```
if(OTELCPP_VERSIONED_LIBS AND NOT BUILD_SHARED_LIBS)
  message(FATAL_ERROR "OTELCPP_VERSIONED_LIBS=ON requires BUILD_SHARED_LIBS=ON")
endif()
```

marcalff (commented):
Using OPENTELEMETRY_ABI_VERSION_NO as SOVERSION looks like the correct choice.

See minor comment on dependent options.

CMakeLists.txt
lalitb: Good point. Have added this check.

marcalff (approved):
LGTM.

To clarify, I don't think opentelemetry-cpp is ready to use shared libraries _yet_.

We need to make sure the proper symbol is part of the proper library, see issue #1429.

This is however a different issue, which should not prevent this part to go in the main branch.

Ok to merge.

ThomsonTan (approved):

PR #2114, opened 2023-04-21
[CI] Upgrade GoogleTest version from 1.12.1 to 1.13.0

cngzhnp:
Upgraded GoogleTest to `1.13.0` + `C++14`

## Changes

Revised the versions used for GoogleTest:

* By default, upgraded to `1.13.0`, which requires `C++14`
* Still using `1.12.1` (unchanged) for CI builds that requires `C++11`
* Still using `1.10.0` (unchanged) for CI builds using `GCC 4.8`


ci/setup_cmake.sh
```diff
@@ -12,8 +12,16 @@ if [ "x$CMAKE_VERSION" = "x" ]; then
   export CMAKE_VERSION=3.15.2
 fi
 
+# This variable not set on CI pipeline for only legacy environment(GCC 4.8).
+# With 1.13.0 version, C++14 must be set which does not supported by legacy environment anymore.
+# Also with this version, release version path needs to be adapted.
 if [ "x$GOOGLETEST_VERSION" = "x" ]; then
-  export GOOGLETEST_VERSION=1.12.1
+  export GOOGLETEST_VERSION=1.10.0
```
lalitb, line 19: Why is default version changed to 1.10.0, can it remain as 1.12.1 ?

lalitb:
Can you update the description for the changes. As the latest version doesn't support C++11, how are we handling now CI tests for C++11 ?

ci/setup_cmake.sh
cngzhnp: If GoogleTest version is not set on CI Pipeline, it was set as 1.10.0 by default. It could not be 1.12.1 version because GCC 4.8 oldish environment to support C++14.

CMakeLists.txt
```diff
@@ -535,6 +535,7 @@ list(APPEND CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}")
 
 include(CTest)
 if(BUILD_TESTING)
+  set(CMAKE_CXX_STANDARD 20)
```
lalitb, line 538: @cngzhnp  - I think this will break the builds for users who are building library with tests using gcc4.8 ( or any C++11 only compiler) ?

cngzhnp: Yes, you are right. However, it is like CI/CD pipeline design problem for opentelemetry-cpp. There are a lot of scripts which setup different versions of third party libraries or submodules etc. All of them are not in one page.

From my perspective, if we still want to support GCC 4.8 version with C++11 standard, we need a better CMake/CI-CD/Script design for this. Otherwise, like me people might only focus on passing the CI tests but not the real usecases.

.github/workflows/ci.yml
```diff
@@ -277,7 +283,6 @@ jobs:
       env:
         CC: /usr/bin/gcc-4.8
         CXX: /usr/bin/g++-4.8
-        GOOGLETEST_VERSION: "1.10.0"
```
ThomsonTan, line 280: Why is this version requirement for GCC 4.8 removed?

cngzhnp: It is not removed, it was set in the build script. As I wrote above comments, "If GoogleTest version is not set on CI Pipeline, it was set as 1.10.0 by default. It could not be 1.12.1 version because GCC 4.8 oldish environment to support C++14."

CHANGELOG.md
```diff
@@ -58,6 +58,8 @@ Increment the:
   [#2124](https://github.com/open-telemetry/opentelemetry-cpp/pull/2124)
 * [API] Add user facing Logging API and Benchmarks
   [#2094](https://github.com/open-telemetry/opentelemetry-cpp/pull/2094)
+* [DEPRECATION] Drop C++11 support
+  [#2146](https://github.com/open-telemetry/opentelemetry-cpp/pull/2146)
```
marcalff, line 62: These lines are not added by this PR, and it should not show here.

There is something wrong the way merged from the main branch to this PR branch are done.

.github/workflows/ci.yml
marcalff: The logic is not intuitive, and convoluted in my opinion.

Instead of:

* setting a very old version by default in setup_cmake
* overriding with a new version in ci.yml in every places expect a few old builds

I suggest the opposite:

* Set by default the most up to date version in setup_cmake (1.13.0)
* Override the version in ci.yml for builds where we do want, on purpose, an older version (gcc 4.8, and the new C++ build to come in PR #2152)

ci/do_ci.sh
```diff
@@ -196,6 +196,7 @@ elif [[ "$1" == "cmake.c++20.stl.test" ]]; then
         -DWITH_LOGS_PREVIEW=ON \
         -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
         -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+        -DCMAKE_CXX_STANDARD=20 \
```
marcalff, line 199: CMAKE_CXX_STANDARD=20 is valid for cmake.c++20.stl.test, ok.

For all the builds that now uses Gtest 1.13.0, which require C++14, ci/do_ci.sh most likely needs to be updated to add CMAKE_CXX_STANDARD=14 in relevant builds.

CMakeLists.txt
marcalff: This sets C++20 for all builds using tests, which breaks gcc4.8.

See related comment, to set C++14 is relevant ci/do_ci.sh targets instead.

marcalff (commented):
Thanks for the PR.

See comments for action items.

marcalff (commented):
@cngzhnp 

Now that the SDK header file cleanup is merged (thanks for the PR),
it is now time to look again at GTest.

First, please note I did a main -> PR 2114 merge to get up to date code.

As a result, be sure to pull before making any new changes.

Secondly, there is now a new build in CI:

CI / CMake gcc 12 (maintainer mode, C++11, async) (pull_request) 

It currently fails, so please fix it by using GTest version 1.12.1 and C++11.

The ci/setup_cmake.sh file needs to be fixed as follows:

When GOOGLETEST_VERSION is empty, the default value should be the most up to date version, so 1.13.0

Then, when proceeding to download GTest, the script must check the version number to build the proper GOOGLETEST_VERSION_PATH / GOOGLETEST_FOLDER_PATH.

This must work for 1.10.0, 1.12.1 and 1.13.0, and is independent of whether a value was provided by the caller or not.

Last, please cleanup the extra entry in the CHANGELOG.

marcalff:
Ping @cngzhnp

Did you see the previous comments ?

cngzhnp:
Hello @marcalff, sorry for the late answer. Yeah, I saw that but I was busy with other stuff. I planned to work this task on this week.

cngzhnp:
@marcalff Could you please verify the changes after your comments? Also, could you please take a look that why CodeQL build is failed, I am not sure that fail is directly related to my changes?

marcalff:
@cngzhnp Thanks for the cleanup.

Remaining points below.

## CodeQL

The `CodeQL / CodeQL-Build` build fails on:

```
  [  0%] Building CXX object api/test/core/CMakeFiles/version_test.dir/version_test.cc.o
  cd /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/_lgtm_build_dir/api/test/core && /usr/bin/c++  -DENABLE_TEST -I/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/api/include  -fpermissive -O3 -DNDEBUG   -Wno-error=deprecated-declarations -std=gnu++11 -o CMakeFiles/version_test.dir/version_test.cc.o -c /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/api/test/core/version_test.cc
  In file included from /usr/include/gtest/gtest-message.h:57,
                   from /usr/include/gtest/gtest-assertion-result.h:46,
                   from /usr/include/gtest/gtest.h:64,
                   from /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/api/test/core/version_test.cc:6:
  /usr/include/gtest/internal/gtest-port.h:270:2: error: #error C++ versions less than C++14 are not supported.
    270 | #error C++ versions less than C++14 are not supported.
        |  ^~~~~
```

To fix this and keep the build in C++11, add:

```
GOOGLETEST_VERSION: 1.12.1
```

in file codeql-analysis.yml

## setup-cmake.sh

Per previous comments, this code:

```
# This variable not set on CI pipeline for only legacy environment(GCC 4.8).
# With 1.13.0 version, C++14 must be set which does not supported by legacy environment anymore.
# Also with this version, release version path needs to be adapted.
if [ -z "${GOOGLETEST_VERSION}" ]; then
  # By default, GoogleTest version set the following version.
  export GOOGLETEST_VERSION=1.13.0
  GOOGLETEST_VERSION_PATH="v${GOOGLETEST_VERSION}"
  GOOGLETEST_FOLDER_PATH="googletest-${GOOGLETEST_VERSION}"
else
  GOOGLETEST_VERSION_PATH="release-${GOOGLETEST_VERSION}"
  GOOGLETEST_FOLDER_PATH="googletest-release-${GOOGLETEST_VERSION}"

fi
```

should decouple setting the default release with the release format.

Something like:

```
if [ -z "${GOOGLETEST_VERSION}" ]; then
  # Version by default. Requires C++14
  export GOOGLETEST_VERSION=1.13.0
fi
```

followed by something like:

```
if [  << GOOGLETEST_VERSION greater or equal to 1.13.0 >> ]; then
  GOOGLETEST_VERSION_PATH="v${GOOGLETEST_VERSION}"
  GOOGLETEST_FOLDER_PATH="googletest-${GOOGLETEST_VERSION}"
else
  GOOGLETEST_VERSION_PATH="release-${GOOGLETEST_VERSION}"
  GOOGLETEST_FOLDER_PATH="googletest-release-${GOOGLETEST_VERSION}"
fi
```

Expectation is that all the following use case works

```
unset GOOGLETEST_VERSION
setup_cmake
```

installs 1.13.0

```
GOOGLETEST_VERSION=1.13.0
setup_cmake
```

installs 1.13.0

```
GOOGLETEST_VERSION=1.12.1
setup_cmake
```

installs 1.12.1

```
GOOGLETEST_VERSION=1.10.0
setup_cmake
```

installs 1.10.0

cngzhnp:
@marcalff Could you please check & verify what I changed so for? Now, I fixed the CodeQL and split up GoogleTest version and its path.

marcalff:
> @marcalff Could you please check & verify what I changed so for? Now, I fixed the CodeQL and split up GoogleTest version and its path.

Thanks.

How about this instead:

```
OLD_VERSION_REGEXP="^1\.([0-9]|10|11|12)(\..*)?$"

if [[ ${GOOGLETEST_VERSION} =~ ${OLD_VERSION_REGEXP} ]]; then
  # Old package name
else
  # new (starting with 1.13.0) package name
fi
```

.github/workflows/ci.yml
marcalff: Marking as resolved, the version 1.10.0 is back in the CI script

marcalff (approved):
The code LGTM.

Please revise:

* the PR description
* code comments

for cleanup.

@owent 

This change may cause merge collisions with your work, when setting CMAKE_CXX_STANDARD. I assume these will be trivial to fix, so not an issue.

@lalitb @esigo @owent 

Please take a second look.

owent (approved):
LGTM and thanks.

owent:
> The code LGTM.
> 
> Please revise:
> 
> * the PR description
> * code comments
> 
> for cleanup.
> 
> @owent
> 
> This change may cause merge collisions with your work, when setting CMAKE_CXX_STANDARD. I assume these will be trivial to fix, so not an issue.
> 
> @lalitb @esigo @owent
> 
> Please take a second look.

It's fine by me , we also use gtest 13 and a high `CMAKE_CXX_STANDARD` in some of our products.While for the users still use gcc 4.8 and C++11, them can still use gtest 1.10.0 and it works well now.

marcalff:
Set merge temporarily on hold, to avoid conflicts with the release of opentelemetry-cpp 1.10.0.

CMakeLists.txt
```diff
@@ -208,8 +208,8 @@ set(OTELCPP_PROTO_PATH
 if(WIN32)
   if(BUILD_TESTING)
     if(MSVC)
-      # GTest bug: https://github.com/google/googletest/issues/860
-      add_compile_options(/wd4275)
+      # Warning as error: warning STL4036: <ciso646> is removed in C++20
+      add_compile_options(/wd4996)
```
ThomsonTan, line 212: Could this option be applied to test code only instead of making it a global option?

ci/do_ci.sh
```diff
@@ -89,6 +89,7 @@ if [[ "$1" == "cmake.test" ]]; then
   cd "${BUILD_DIR}"
   rm -rf *
   cmake -DCMAKE_BUILD_TYPE=Debug  \
+        -DCMAKE_CXX_STANDARD=14 \
```
lalitb, line 92: This property would be redundant once we switch to C++11 (#2147), as we would update the property [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/ca8c23481ae10c1569858cee19fa076c27b8f261/CMakeLists.txt#L139). Need to make a note of removing from here once that is done.

lalitb (approved):
LGTM. Thanks for the changes.

ci/do_ci.sh
marcalff: Added a note in #2147, so this part is resolved (nothing to change in this PR).

CMakeLists.txt
marcalff: As discussed in C++ SIG meeting, 2023-07-12, this is a minor item that will be addressed in a separate PR.

PR #2120, opened 2023-04-25
[EXPORTER] OTLP GRPC mTLS support

### LINKED ISSUE #1785: [EXPORTER] OTLP GRPC mTLS support
Wanted to setup mtls connection between opentelemetry collector and opentelemetry-cpp client. 

But was receiving error like 
```sh
E1116 16:33:08.044460991 1642002 ssl_transport_security.cc:555] Corruption detected.
E1116 16:33:08.044507949 1642002 ssl_transport_security.cc:531] error:10000412:SSL routines:OPENSSL_internal:SSLV3_ALERT_BAD_CERTIFICATE
E1116 16:33:08.044518471 1642002 secure_endpoint.cc:306]     Decryption error: TSI_DATA_CORRUPTED
E1116 16:33:08.048551622 1642002 ssl_transport_security.cc:555] Corruption detected.
E1116 16:33:08.048570967 1642002 ssl_transport_security.cc:531] error:10000412:SSL routines:OPENSSL_internal:SSLV3_ALERT_BAD_CERTIFICATE
E1116 16:33:08.048579525 1642002 secure_endpoint.cc:306]     Decryption error: TSI_DATA_CORRUPTED
```

Upon research found out there is no mtls support from opentelemetry-cpp client side to enable connection with otel collector.

**Steps to reproduce Issue**
otel config should be: 
```sh
receivers:
  otlp:
    protocols:
      grpc:
        tls:
          client_ca_file: "path to client.pem"
          cert_file: "path to server.crt"
          key_file: "path to server.key"
processors:
  batch:

exporters:
  logging:
    logLevel: debug

service:
  pipelines:
    logs:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]

    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]

    metrics:
      receivers: [otlp]
      processors: [batch]
      exporters: [logging]
```
Run Client as following: 
```sh
<path_to_dir>/opentelemetry-cpp/build/examples/otlp$ ./example_otlp_grpc localhost:4317 <path_to_ca_cert>
```

Please enable mTLS support on opentelemetry-cpp client side.

kylepl:
Fixes #1785

## Changes

### DIFF
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include "absl/types/optional.h"
+#include "absl/types/variant.h"
 #include "opentelemetry/exporters/otlp/otlp_environment.h"
 
 #include <memory>
@@ -28,6 +30,19 @@ struct OtlpGrpcExporterOptions
   // ssl_credentials_cacert_as_string in-memory string representation of .pem file to be used for
   // SSL encryption.
   std::string ssl_credentials_cacert_as_string = GetOtlpDefaultSslCertificateString();
+
+#ifdef ENABLE_OTLP_GRPC_MTLS_PREVIEW
+  // At most one of ssl_client_key_* should be non-empty. If use_ssl_credentials, they will
+  // be read to allow for mTLS.
+  std::string ssl_client_key_path   = GetOtlpDefaultTracesSslClientKeyPath();
+  std::string ssl_client_key_string = GetOtlpDefaultTracesSslClientKeyString();
+
+  // At most one of ssl_client_cert_* should be non-empty. If use_ssl_credentials, they will
+  // be read to allow for mTLS.
+  std::string ssl_client_cert_path   = GetOtlpDefaultTracesSslClientCertificatePath();
+  std::string ssl_client_cert_string = GetOtlpDefaultTracesSslClientCertificateString();
+#endif
+
   // Timeout for grpc deadline
   std::chrono::system_clock::duration timeout = GetOtlpDefaultTimeout();
   // Additional HTTP headers
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
@@ -29,7 +29,7 @@ struct OtlpGrpcExporterOptions
   // SSL encryption.
   std::string ssl_credentials_cacert_as_string = GetOtlpDefaultSslCertificateString();
 
-#ifdef ENABLE_OTLP_GRPC_MTLS_PREVIEW
+#ifdef ENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW
   // At most one of ssl_client_key_* should be non-empty. If use_ssl_credentials, they will
   // be read to allow for mTLS.
   std::string ssl_client_key_path   = GetOtlpDefaultTracesSslClientKeyPath();
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
@@ -3,8 +3,6 @@
 
 #pragma once
 
-#include "absl/types/optional.h"
-#include "absl/types/variant.h"
 #include "opentelemetry/exporters/otlp/otlp_environment.h"
 
 #include <memory>
--- a/exporters/otlp/src/otlp_grpc_client.cc
+++ b/exporters/otlp/src/otlp_grpc_client.cc
@@ -69,17 +69,17 @@ std::shared_ptr<grpc::Channel> OtlpGrpcClient::MakeChannel(const OtlpGrpcExporte
   grpc_arguments.SetUserAgentPrefix(options.user_agent);
 
   if (options.use_ssl_credentials) {
-    grpc::SslCredentialsOptions ssl_opts = {
-      .pem_root_certs = GetFileContentsOrInMemoryContents(
-          options.ssl_credentials_cacert_path,
-          options.ssl_credentials_cacert_as_string),
-#if ENABLE_OTLP_GRPC_MTLS_PREVIEW
-      .pem_private_key = GetFileContentsOrInMemoryContents(
-          options.ssl_client_key_path,
-          options.ssl_client_key_string),
-      .pem_cert_chain = GetFileContentsOrInMemoryContents(
-          options.ssl_client_cert_path,
-          options.ssl_client_cert_string)
+    grpc::SslCredentialsOptions ssl_opts;
+    ssl_opts.pem_root_certs = GetFileContentsOrInMemoryContents(
+        options.ssl_credentials_cacert_path,
+        options.ssl_credentials_cacert_as_string);
+#if ENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW
+    ssl_opts.pem_private_key = GetFileContentsOrInMemoryContents(
+      options.ssl_client_key_path,
+      options.ssl_client_key_string);
+    ssl_opts.pem_cert_chain = GetFileContentsOrInMemoryContents(
+      options.ssl_client_cert_path,
+      options.ssl_client_cert_string);
 #endif
     };
     channel =
--- a/exporters/otlp/src/otlp_grpc_client.cc
+++ b/exporters/otlp/src/otlp_grpc_client.cc
@@ -36,9 +36,11 @@ static std::string GetFileContents(const char *fpath)
 }
 
 // If the file path is non-empty, returns the contents of the file. Otherwise returns contents.
-static std::string GetFileContentsOrInMemoryContents(
-  const std::string& file_path, const std::string& contents) {
-  if (!file_path.empty()) {
+static std::string GetFileContentsOrInMemoryContents(const std::string &file_path,
+                                                     const std::string &contents)
+{
+  if (!file_path.empty())
+  {
     return GetFileContents(file_path.c_str());
   }
   return contents;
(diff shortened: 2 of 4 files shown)


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
```diff
@@ -6,13 +6,44 @@
 #include "opentelemetry/exporters/otlp/otlp_environment.h"
 
 #include <memory>
+// TODO: These requires C++17, is that OK? It seems like we need to support C++0x?
```
owent, line 9: We need support c++11. You can use absl::variant and absl::optional for alternative.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
```diff
@@ -6,13 +6,44 @@
...
 namespace otlp
 {
 
+// This type allows the variants to be default-constructed to neither type.
+// This avoids needing deeper template if std::optional<std::variant<>>.
+struct Unset {};
+
+// TODO: Check if this strong-typing is ok. It seems nice, and allows variant typing.
+struct FilePath {
+  std::string value;
```
owent, line 25: I think we'd better use `nostd::string_view` here.

owent:
The trace is already stable. I think we need discuss this approch because it break the API and also we need it to support environment variables in [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md).

lalitb:
@kylepl Thanks for working on this. You can also go through #1793, as this provides support for client-side cert/key for OTLP HTTP exporter.

kylepl:
Thanks for the comments!

Based on my understanding of [Stable](https://opentelemetry.io/docs/reference/specification/versioning-and-stability/#stable), compilation breakages are not allowed except for major version changes (reasonable). Is it reasonable to still make things compatible, but mark `ssl_credentials_cacert_path` and `ssl_credentials_cacert_as_string`  as deprecated? It's not quite obvious to me what needs to ABI compatible (i.e. I don't know what gets shipped by itself).

Supporting the environment variables sounds good to me, can do so.

Also, a related thing related to what I'm interested in - is gRPC actually has _experimental_ support refreshing certificates every X seconds from disk. If that seems reasonable, while it could be a separate PR, given it may effect the API, it may be worth discussing now.

And of course feel free to give me your opinions on style and such, once we agree on the high-level, I'll take another pass on the code:
* If it is reasonable to have the ssl credentials held in a sub-struct (to group them together), and thus somehow mark `ssl_credentials_cacert_path` and `ssl_credentials_cacert_as_string` - while still supporting them.
* If supporting auto-refresh of certificates seems reasonable (it is an experimental gRPC feature, we could add a field called "experimental_enable_auto_refresh" to signify that.

lalitb:
@kylepl Why do we want to deprecate the existing members - `ssl_credentials_cacert_path` and `ssl_credentials_cacert_as_string`. These are for server-side auth, and can remain there. Can't we just add the new members to OtlpGrpcExporterOptions for client-side auth ? Sorry if I am missing something here.

kylepl:
@lalitb Just to be precise, and is likely  what you mean, `ssl_credentials_cacert_{path|as_string}` are specified by the client as the root certificates to validate the server.

There are few pieces to the change:
* I think it is useful to group related options together into sub-structs to show what is related, and to avoid an unwieldy top-level struct. Thus the thinking was to group client-side cert, key and root certs (for authenticating the server). This is what gRPC does too.
* I personally like more type-safe objects - hence using a single field as either a file or string contents. This avoids the need to just have a comment that only one is respected - there is no way to specify both.

But, there is a cost of changing what is already there even if we did agree on what to do if this was entirely new code. Thus my proposal is to support the new format and the old format, with a deprecation warning - but this is where I defer to everyone on what makes sense.

So if you want the two existing options to remain where they are, and potentially just have `SSLOptions` but without the root certs, that's also fine, just not quite as ideal IMO long-term - but everything is trade-offs. :)

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
kylepl: Just note that the existing options in OtlpGrpcExporterOptions all use std::string, hence why I used that. If this should be different, so be it - just checking before changing it.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
kylepl: Done.

kylepl:
I've switched to using absl::variant/optional, thus the compilation should be fixed for earlier versions.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h
```diff
@@ -17,7 +16,7 @@ namespace otlp
 {
 
 // This type allows the variants to be default-constructed to neither type.
-// This avoids needing deeper template if std::optional<std::variant<>>.
+// This avoids needing deeper template if absl::optional<absl::variant<>>.
 struct Unset {};
```
kylepl, line 20: Note to self: This is what nostd::monostate is for, remove this,

marcalff:
@kylepl

First, thanks for picking this up, support for mTLS in the gRPC exporter is
long overdue.

Please find here some comments and guidance.

# struct OtlpGrpcExporterOptions

The ca_cert is already implemented with ssl_credentials_cacert_path and
ssl_credentials_cacert_as_string, so these should not change unfortunately.

For the client cert and key, please use a flat structure, with different
members for the path or string options.

Using a std::variant is technically correct, and more type safe, but it
causes additional complexity to use the interface itself, and the pattern is
different from the HTTP exporter.

Please follow the pattern from struct OtlpHttpExporterOptions, for
consistency.

In the long term, configurations using environment variables is going to be
replaced by configuration using a yaml file, with a different structure, per
OTEP 225:
* https://github.com/open-telemetry/oteps/pull/225
* https://github.com/open-telemetry/oteps/blob/main/text/0225-configuration.md

When this change is implemented for all opentelemetry repositories, the
options struct in opentelemetry-cpp will probably be revisited then.

For now, consistency with the existing HTTP pattern will avoid too many intermediate changes.

# Environment variables

struct OtlpGrpcExporterOptions should by default read environment variables,
so that configuration options are honored.

Please call:
* GetOtlpDefaultTracesSslClientKeyPath();
* GetOtlpDefaultTracesSslClientKeyString();
* GetOtlpDefaultTracesSslClientCertificatePath();
* GetOtlpDefaultTracesSslClientCertificateString();

Likewise, see struct OtlpHttpExporterOptions.

# CMake

This is a new feature, so it needs a CMake option, for optional adoption.

Something like:
* CMake option WITH_OTLP_GRPC_SSL_MTLS_PREVIEW
* define ENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW

See existing options for the OTLP HTTP exporter.

In the implementation:
* it is ok to require a recent gRPC version if mTLS support is required
* it is not ok to require a recent gRPC version if mTLS support is not
  required.

In other words, the code should still build with old gRPC libraries,
with the understanding that mTLS will not be available then.

See build failures in CI related to gcc 4.8 with a very old gRPC library, it
should still build.

C++11 support, and support for old libraries, will go away over time, but for
now these old builds should still be functional for existing features.

# Unit tests

For unit tests for the option struct itself, see file
otlp_grpc_exporter_test.cc.

Examples of units tests for the HTTP exporter can be found in file
otlp_http_exporter_test.cc.

# Functional tests

See functional/otlp for existing OTLP HTTP functional tests.

Either in the same PR, or in subsequent work, functional tests talking to an
opentelemetry collector using the OTLP GRPC exporter can be implemented the
same way.

marcalff:
@kylepl 

Please confirm if you are still working on this.

The following [link](https://github.com/open-telemetry/opentelemetry-cpp/pull/2120#issuecomment-1557441442) contains comments and guidance about how to implement this change, in case you missed it.

Thanks.

kylepl:
Apologies for the long delay. I've taken another pass - doing everything except for the tests - which I can do once this LGTY.

kylepl:
There were a number of failures - I think because of the unnecessary variant/optional include. I've removed them, so hopefully the bazel versions pass now.

For formatting, is there a script I should run to do the auto-formatting?

Also, I wasn't sure if the thumbs-up meant review seemed reasonable, but I'll wait for an explicit comment before doing the tests. :)

kylepl:
And I've put in one more fix for Windows not liking initializer lists for structs.

Should just be the formatting errors to resolve (once I know what to run for it).

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -58,17 +68,20 @@ std::shared_ptr<grpc::Channel> OtlpGrpcClient::MakeChannel(const OtlpGrpcExporte
...
+        options.ssl_credentials_cacert_as_string);
+#if ENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW
+    ssl_opts.pem_private_key = GetFileContentsOrInMemoryContents(
+      options.ssl_client_key_path,
+      options.ssl_client_key_string);
+    ssl_opts.pem_cert_chain = GetFileContentsOrInMemoryContents(
+      options.ssl_client_cert_path,
+      options.ssl_client_cert_string);
+#endif
+    };
```
marcalff, line 84: It seems this line breaks the build.

Please make sure the code builds locally with grpc before pushing, it will save a few round trips.

marcalff (commented):
@kylepl 

Thanks for the PR, this is looking much better already.

Right now, the build fails (to fix), and does not use the new MTLS_PREVIEW option anyway.

Please look for:

```
        -DWITH_OTLP_HTTP_SSL_PREVIEW=ON \
        -DWITH_OTLP_HTTP_SSL_TLS_PREVIEW=ON \
```

in file ci/do_ci.sh,

and add both `WITH_OTLP_GRPC` and `WITH_OTLP_GRPC_SSL_MTLS_PREVIEW` at the same places, to have some CI coverage of the new code.

marcalff:
To fix format errors, run clang-format version 10.

See docs/using-clang-format.md

exporters/otlp/src/otlp_grpc_client.cc
kylepl: Apologies for that, I tested a later patch, but not exactly this - missing it.

kylepl:
I've done clang-format on the .cc/.h files, and enabled this path in the CI.

kylepl:
I've done another pass of fixing format errors.

There are also now failures about gRPC inputs not being included, I think because more options must be specified - but I'm not sure what exactly needs to be copied over (seems like a lot?). e.g.

```
  grpc_cpp_plugin=`which grpc_cpp_plugin`
  proto_make_file="CMakeFiles/opentelemetry_proto.dir/build.make"
  sed -i "s~gRPC_CPP_PLUGIN_EXECUTABLE-NOTFOUND~$grpc_cpp_plugin~" ${proto_make_file} #fixme
```

marcalff:
@kylepl 

> I've done another pass of fixing format errors.
> 
> There are also now failures about gRPC inputs not being included, I think because more options must be specified - but I'm not sure what exactly needs to be copied over (seems like a lot?). e.g.


Indeed, a lot more setup is needed beside just adding `WITH_OTLP_GRPC`.

This was bad guidance from me, sorry about that.

Let's revert this changes to `do_ci.sh`, and instead, add `WITH_OTLP_GRPC_SSL_MTLS_PREVIEW` to only `cmake.exporter.otprotocol.test` then.

This should be simpler as the `GRPC` setup is already done for this build.

owent (approved):
LGTM and thanks.

lalitb (approved):
Looks good. Thanks for the contribution.
Perhaps as separate PR, we can add the functional tests for validating OTLP gRPC with client certificate chain  + Server ca certificate. 
Ref- https://github.com/open-telemetry/opentelemetry-cpp/tree/main/functional/otlp

marcalff (approved):
Approved, thanks for the contribution.

@kylepl 

If you want to continue working in this area, feel free to pick
* https://github.com/open-telemetry/opentelemetry-cpp/issues/2208

PR #2123, opened 2023-04-27
SDK support for the new OTel log API

### LINKED ISSUE #2122: Support new user-facing log API in SDK
As #2094 is merged, SDK should also be updated to support the new API.

ThomsonTan:
Fixes #2122 

## Changes


api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -48,9 +48,22 @@ template <>
...
+  inline static LogRecord *Set(LogRecord *log_record, ArgumentType &&arg) noexcept
+  {
+    log_record->SetEventId(arg.id_, nostd::string_view{arg.name_.get()});
+
+    return log_record;
+  }
+};
+
+template <>
+struct LogRecordSetterTrait<int64_t>
```
owent, line 60: Is it conflicted with Body?it can also be converted from `int64_t` implicitly.

ThomsonTan: Yes, thanks for the catch. Seems the template part of filling logging record requires each parameter to be in different type, and the `any` type on body make it conflict with other parameters. Do you have any suggestion for this problem?

owent: I have no idea for this problem by now, and I only avoided to specialize templates for these types before.
Maybe we can add a unit test to check the mistake.

ThomsonTan: I removed the passing of int64_t to SDK, instead, a temporary EventId struct will be constructed which should fix the issue.

For unittest, do you mean to cover all the variant types of `body` field?

api/include/opentelemetry/logs/event_id.h
```diff
@@ -18,14 +18,16 @@ namespace logs
 class EventId
 {
 public:
-  EventId(int64_t id, nostd::string_view name) noexcept
+  EventId(int64_t id, nostd::string_view name) noexcept : id_{id}
```
lalitb, line 21: not required to initialize `id_` inside constructor body if it is initialized using initializer list.

lalitb (approved):
LGTM. Thanks

api/include/opentelemetry/logs/event_id.h
ThomsonTan: Thanks for the catch.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_log_recordable.h
```diff
@@ -67,6 +67,16 @@ class OtlpLogRecordable final : public opentelemetry::sdk::logs::Recordable
    */
   void SetBody(const opentelemetry::common::AttributeValue &message) noexcept override;
 
+  /**
+   * @brief Set the Event Id for this log.
+   * @param id the event Id to set
+   * @param name  the event name to set
+   */
+  void SetEventId(int64_t /* id */, nostd::string_view /* name */) noexcept override
```
owent, line 75: Did I miss anything? I can find any specification about event id in OTLP, should we add it just like [semantic convention for event attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/semantic_conventions/events.md)?

api/include/opentelemetry/logs/logger_type_traits.h
owent: Yes, use EmitLogRecord to cover all the variant types of body field.
Maybe we can add it in another PR later.

sdk/include/opentelemetry/sdk/logs/multi_recordable.h
```diff
@@ -55,6 +55,14 @@ class MultiRecordable final : public Recordable
    */
   void SetBody(const opentelemetry::common::AttributeValue &message) noexcept override;
 
+  /**
+   * Set the Event Id
+   * @param id the event id to set
+   * @param name  the event name to set
+   */
+  // TODO: implement set event id
+  void SetEventId(int64_t /* id */, nostd::string_view /* name */) noexcept override {}
```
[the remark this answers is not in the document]
ThomsonTan: Thanks. Yes, the `multi_recordable` method should just forward the call to each individual recordable. Updated with this.

api/include/opentelemetry/logs/logger_type_traits.h
ThomsonTan: FYI, I added an integration test to cover if the EventId passed to the user-facing API is passed correctly to the SDK.

https://github.com/ThomsonTan/opentelemetry-cpp/blob/otel_cpp_log_sdk_impl/exporters/ostream/test/ostream_log_test.cc#L465

ThomsonTan:
@owent please let me know if there is any comments I missed.

PR #2124, opened 2023-04-27
Fix inclusion header files and use forward declaration for API

cngzhnp:
Fixes #2043 but only API side.

## Changes


cngzhnp:
@marcalff Could you please review these changes only related to API folder?

api/include/opentelemetry/plugin/hook.h
```diff
@@ -41,6 +40,11 @@
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace plugin
 {
+
+struct LoaderInfo;
+class Factory;
```
lalitb, line 45: Is this required ?

api/include/opentelemetry/trace/noop.h
```diff
@@ -102,15 +102,14 @@ class OPENTELEMETRY_EXPORT NoopTracer final : public Tracer,
 /**
  * No-op implementation of a TracerProvider.
  */
-class OPENTELEMETRY_EXPORT NoopTracerProvider final : public opentelemetry::trace::TracerProvider
+class OPENTELEMETRY_EXPORT NoopTracerProvider final : public trace_api::TracerProvider
```
lalitb, line 105: should we instead use `trace::TracerProvider`, and remove `trace_api` namespace alias from the header file. It's not recommended to have namespace alias in header files (though this is existing code)

lalitb:
Changes look good in general, and nicely done. I think this would be the breaking change for applications if they are taking dependency on these unnecessary header inclusions in our API. These applications need to explicitly include these removed header files / forward-declared classes.

lalitb (approved):
Approving the changes with nit comments. I believe @marcalff has already approved these api changes as indicated here - https://github.com/open-telemetry/opentelemetry-cpp/pull/2043#pullrequestreview-1395717324. Still have concern on this breaking the existing application code, so would be good to discuss when/how to merge/release in next community meeting.

api/include/opentelemetry/plugin/hook.h
cngzhnp: No need to put this class as a forward declaration, probably even stands on my early trials. Will be removed.

api/include/opentelemetry/trace/noop.h
cngzhnp: Yes, you are right. I'd change it as you wish.

lalitb: Looks like the `trace_api` alias defined here is been used in other places too. We can keep it here in that case.

cngzhnp: Reverted back only `trace_api` alias. However, I will take it as a note, maybe it would be removed after same changes are applied to SDK too.

api/test/plugin/dynamic_load_test.cc
```diff
@@ -2,6 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "opentelemetry/plugin/dynamic_load.h"
+#ifdef _WIN32
```
ThomsonTan, line 5: Why is this needed as the `#ifdef _WIN32` is included in `opentelemetry/plugin/dynamic_load.h` which has already been included above?

cngzhnp: In `dynamic_load.h` file, there is no preprocessor definition anymore, I removed it in this pull request. Because there is only one function declaration which is not related to operating system itself. 

However, only this source file is related to operating system itself with dynamic_load and which has to be included like that. Otherwise, there will be a compiler error on test codes.

ThomsonTan: Why was the including of such `dynamic_load_<os>.h` moved out of `dynamic_load.h`? With change, the user code like the test code here need to include both files separately. It looks to me that including the os header in `dynamic_load.h` is more convenient.

cngzhnp: Ok, I will revert it back.

ThomsonTan (approved):

owent (approved):

PR #2126, opened 2023-04-29
Fix if JSON library already added another CMake target

### LINKED ISSUE #1879: Nlohmann library can cause conflict on CMake targets during build
**Describe your environment** 

Currently, in our project we are using `Nlohmann::JSON` library for another purpose which is added as a submodule. On other hand, we are also using `opentelemetry-cpp` library which brings with its own `Nlohmann::JSON` library which is a problematic. 

**Steps to reproduce**
During our build process, `Nlohmann::JSON` library that we used version is already parsed and `nlohmann_json` target is created by CMake. Afterwards, the next submodule is opentelemetry-cpp library and if it is built with Zipkin option (or could be another exporter which uses Nlohmann) which causes CMake build error like:

```
add_library cannot create target "nlohmann_json" because another target
  with the same name already exists.  The existing target is an interface
  library created in source directory
  "x/y/z/3rdParty/json".  See
  documentation for policy CMP0002 for more details.
```

My expectation is, even though 3rdParty dependency collision, it should be built up smoothly.

If project is used its own `Nlohmann::JSON` library, it is not possible to build.

**Additional context**
As a solution, in CMakeLists.txt after line 358, such a case must be included:

```cmake
...
  if(nlohmann_json_FOUND)
    message("Using external nlohmann::json")
  elseif(TARGET nlohmann_json) 
    message("nlohmann_json was already added as a CMake target!")
```
This will prevent during CMake process, if `Nlohmann::JSON` target is already found in the CMake target list, no need to do anything.

cngzhnp:
Fixes #1879

## Changes

### DIFF
--- a/cmake/nlohmann-json.cmake
+++ b/cmake/nlohmann-json.cmake
@@ -33,6 +33,8 @@ find_package(nlohmann_json QUIET)
 set(nlohmann_json_clone FALSE)
 if(nlohmann_json_FOUND)
   message(STATUS "nlohmann::json dependency satisfied by: package")
+elseif(TARGET nlohmann_json)
+  message(STATUS "nlohmann::json is already added as a CMake target!")
 elseif(EXISTS ${PROJECT_SOURCE_DIR}/.git
        AND EXISTS
            ${PROJECT_SOURCE_DIR}/third_party/nlohmann-json/CMakeLists.txt)


cmake/nlohmann-json.cmake
```diff
@@ -33,6 +33,8 @@ find_package(nlohmann_json QUIET)
 set(nlohmann_json_clone FALSE)
 if(nlohmann_json_FOUND)
   message(STATUS "nlohmann::json dependency satisfied by: package")
+elseif(TARGET nlohmann_json)
```
ThomsonTan, line 36: Wondering the scenario that `nlohmann_json_FOUND` is not set, but `nlohann_json` target is available.

cngzhnp: `nlohmann_json_FOUND` true only when the library installed on the system like package, not as a submodule or third party usage in another project. 

In my use case, we need this library for another purpose as a third party, so CMake acts that `nlohmann_json` is a CMake target which is the correct in project perspective. If we compile `opentelemetry-cpp + my project + my nlohmann_json` together, `opentelemetry-cpp` will use my `nlohmann_json` library because it is already known as a CMake target by CMake.

Otherwise, it throws an error which I already explained in the bug description.

owent: nlohmann_json may be imported by `add_subdirectory`.

> Wondering the scenario that `nlohmann_json_FOUND` is not set, but `nlohann_json` target is available.

owent (approved):
LGTM  and thanks.

lalitb (approved):

ThomsonTan (approved):

PR #2143, opened 2023-05-19
Missed include

### LINKED ISSUE #2142: Compilation failure with Gcc on Fedora 38
**Describe your environment** 
Fedora 38

**Steps to reproduce**
Opentelemetry-cpp build fails

Successful build

```
In file included from /home/andrew/Work/opentelemetry-cpp/ext/test/http/url_parser_test.cc:4:
/home/andrew/Work/opentelemetry-cpp/ext/include/opentelemetry/ext/http/common/url_parser.h:30:3: error: ‘uint16_t’ does not name a type
   30 |   uint16_t port_;
      |   ^~~~~~~~
/home/andrew/Work/opentelemetry-cpp/ext/include/opentelemetry/ext/http/common/url_parser.h:9:1: note: ‘uint16_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
```

**Additional context**
Looks to me that cstdint was transitively included before and now with an updated gcc library is no longer being so. On inspection it does look like the mentioned header file is clearly incorrect.

astitcher:
Fixes #2142

## Changes
Added a needed include to avoid compilation error on Fedora 38


lalitb (approved):
LGTM. Thanks for the fix.

esigo (approved):
LGTM
Thanks for the fix :)

PR #2156, opened 2023-05-24
[DOC] Small fix for Histogram documentation.

lalitb:
Fixes # (issue)

## Changes

The comment is not consistent with [implementation ](https://github.com/open-telemetry/opentelemetry-cpp/blob/61e8741a70cde8f273b3de49a772f05a12566e80/sdk/src/metrics/sync_instruments.cc#L185 )and [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#record). The Histogram API only records non-negative value.


api/include/opentelemetry/metrics/sync_instruments.h
```diff
@@ -97,14 +97,14 @@ class Histogram : public SynchronousInstrument
   /**
    * Records a value.
    *
-   * @param value The increment amount. May be positive, negative or zero.
+   * @param value The increment amount. MUST be non-negative.
```
marcalff, line 100: The comment also says "The increment amount", which sounds like a copy and paste from counters, and does not apply to histograms.

In my understanding:

For an histogram, the metric value should be passed, and will decide which bucket stat to increment.

There is no such thing as giving an "increment amount".

lalitb: Good point. Have updated this now.

marcalff (approved):
LGTM, thanks for the fix.

esigo (approved):
LGTM
Thanks :)

Issue #1359, opened 2022-04-28
 undefined reference to `opentelemetry::v1::sdk::common


ccoqueiro:
Hi Everyone,

A few days ago I'm trying to compile a simple hello world application(cpp), after the saga of the compilation of opentelemetry cpp SDK, I'm having the following error when running make:

/usr/bin/ld: /opt/otelcpp/lib/libopentelemetry_trace.a(random_id_generator.cc.o): in function `opentelemetry::v1::sdk::trace::RandomIdGenerator::GenerateSpanId()':
random_id_generator.cc:(.text+0x41): undefined reference to `opentelemetry::v1::sdk::common::Random::GenerateRandomBuffer(opentelemetry::v1::nostd::span<unsigned char, 18446744073709551615ul>)'
/usr/bin/ld: /opt/otelcpp/lib/libopentelemetry_trace.a(random_id_generator.cc.o): in function `opentelemetry::v1::sdk::trace::RandomIdGenerator::GenerateTraceId()':
random_id_generator.cc:(.text+0xc7): undefined reference to `opentelemetry::v1::sdk::common::Random::GenerateRandomBuffer(opentelemetry::v1::nostd::span<unsigned char, 18446744073709551615ul>)'
/usr/bin/ld: /opt/otelcpp/lib/libopentelemetry_exporter_jaeger_trace.a(THttpTransport.cc.o): in function `opentelemetry::v1::exporter::jaeger::THttpTransport::THttpTransport(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::multimap<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, opentelemetry::v1::ext::http::client::cmp_ic, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >)':
THttpTransport.cc:(.text+0xb6): undefined reference to `opentelemetry::v1::ext::http::client::HttpClientFactory::CreateSync()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/hello-otel.dir/build.make:113: hello-otel] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/hello-otel.dir/all] Error 2
[...]

Issue #705, opened 2021-04-27
Error linking OTEL library


jefjos:
**Describe your environment** Describe any aspect of your environment relevant to the problem, including your platform, build system, version numbers of installed dependencies, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main branch.
Platform - WSL
Buildsystem - CMake v3.19

Code

main.cpp

```
#include <iostream>

int main(int argc, char *argv[])
{
   std::cout << "Hello CMake!" << std::endl;
   return 0;
}
```
CMakeLists.txt

```
# Set the minimum version of CMake that can be used
# To find the cmake version run
# $ cmake --version
cmake_minimum_required (VERSION 3.19)
cmake_policy(SET CMP0110 NEW)

# Set the project name
project (hello_cmake)

# Add an executable
add_executable(hello_cmake main.cpp)

find_package(opentelemetry-cpp REQUIRED)

```
target_include_directories(hello_cmake PUBLIC ${OPENTELEMETRY_CPP_INCLUDE_DIRS})
target_link_libraries(hello_cmake PUBLIC ${OPENTELEMETRY_CPP_LIBRARIES})

**Steps to reproduce**
Describe exactly how to reproduce the error. Include a code sample if applicable.
I have attached a sample CMake code

cmake -G "Ninja"  -DCMAKE_BUILD_TYPE:STRING="Debug" -DCMAKE_INSTALL_PREFIX:PATH="hello-cmake/out/install/WSL-GCC-Debug"  hello-cmake/CMakeLists.txt

**What is the expected behavior?**
CMake Generation should happen without any warnings or errors

**What is the actual behavior?**
```
-- Configuring done
CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0111 is not set: An imported target missing its location property
  fails during generation.  Run "cmake --help-policy CMP0111" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  IMPORTED_LOCATION not set for imported target "opentelemetry-cpp::version"
  configuration "Debug".
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0111 is not set: An imported target missing its location property
  fails during generation.  Run "cmake --help-policy CMP0111" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

[...]

PR #2163, opened 2023-05-27
Support protobuf 3.22 or upper

### LINKED ISSUE #2095: [BUILD] Compile WITH_OTLP failed
**Describe your environment** 
Linux CentOS7 3.10.0-327.el7.x86_64 #1 SMP Thu Nov 19 22:10:57 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
gcc 12.2.0

**Steps to reproduce**
1) git clone master
2) cmake3 -DBUILD_TESTING=OFF -DCMAKE_CXX_STANDARD=17 -DWITH_OTLP=ON ..

-- The C compiler identification is GNU 12.2.0
-- The CXX compiler identification is GNU 12.2.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc - works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Building for architecture ARCH=x64
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
CMake Warning at /usr/share/cmake3/Modules/FindProtobuf.cmake:500 (message):
  Protobuf compiler version 22.2 doesn't match library version 4.22.2
Call Stack (most recent call first):
  CMakeLists.txt:333 (find_package)

-- Found Protobuf: /usr/local/lib64/libprotobuf.a;-lpthread (found version "4.22.2") 
-- PROTOBUF_PROTOC_EXECUTABLE=/usr/local/bin/protoc
-- opentelemetry-proto dependency satisfied by: git submodule
-- Performing Test check_cxx_compiler_flag_-Wno-type-limits
-- Performing Test check_cxx_compiler_flag_-Wno-type-limits - Success
-- Performing Test check_cxx_compiler_flag_-Wno-deprecated-declarations
-- Performing Test check_cxx_compiler_flag_-Wno-deprecated-declarations - Success
-- Performing Test check_cxx_compiler_flag_-Wno-unused-parameter
-- Performing Test check_cxx_compiler_flag_-Wno-unused-parameter - Success
-- gRPC_CPP_PLUGIN_EXECUTABLE=/usr/local/bin/grpc_cpp_plugin
Building with nostd types...
PATH:/root/opentelemetry-cpp/examples/grpc/protos:/root/opentelemetry-cpp/examples/grpc/protos/messages.proto
-- Configuring done
-- Generating done
-- Build files have been written to: /root/opentelemetry-cpp/build

3) make

success

failed

**Additional context**
[error.txt](https://github.com/open-telemetry/opentelemetry-cpp/files/11220281/error.txt)

owent:
Fixes #2095 

## Changes

Protobuf depends on abseil-cpp 20230125 from v22/v4.22.

It removed `stringpiece.h`, `google::protobuf::Base64Escape` and use `absl::Base64Escape` for alternative.

~~We have encountered multiple problems related to the usage of CMake packages in the past (#705 #1359). As a solution, I have added a patch script `cmake/patch-imported-config.cmake` that allows users to use prebuilt packages with different CONFIG settings when building otel-cpp.~~

BTW: gRPC v1.55.0 also depend the new protobuf and abseil-cpp, so I also test the new gRPC version in the new CI job.

+ Support protobuf 3.22 or upper
+ Add a CI job (`cmake_modern_protobuf_grpc_test`) to test building with new protobuf and gRPC.
+ ~~Patch imported target in cmake CONFIG packages to enable users to use prebuilt packages with different CONFIG settings.~~
+ Remove `--experimental_allow_proto3_optional` when running protoc when we use protobuf 3.16 or upper. (It's not experimental any more)


esigo:
@owent could you please add more detailed information in the PR description. thanks :)

owent:
> @owent could you please add more detailed information in the PR description. thanks :)

Done.

ci/setup_grpc.sh
```diff
@@ -7,24 +7,41 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 old_grpc_version='v1.33.2'
 new_grpc_version='v1.49.2'
+moderd_grpc_version='v1.55.0'
```
lalitb, line 10: nit - did we mean `modern` ?

.github/workflows/ci.yml
```diff
@@ -387,6 +387,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: 'recursive'
+    - name: setup
+      env:
+        PROTOBUF_VERSION: '23.2'
+        ABSEIL_CPP_VERSION: '20230125.3'
+        CXX_STANDARD: '17'
```
lalitb, line 401: Why do we need C++17 standard for build ?

.github/workflows/ci.yml
```diff
@@ -387,6 +387,31 @@ jobs:
...
+      run: |
+        sudo ./ci/setup_cmake.sh
+        sudo ./ci/setup_ci_environment.sh
+        sudo -E ./ci/install_abseil.sh
+        sudo -E ./ci/install_protobuf.sh
+    - name: run otlp exporter tests
+      env:
+        WITH_ABSEIL: 'ON'
```
lalitb, line 409: Should we also have a test without WITH_ABSEIL to ensure that ABI compatible build works fine ?

.github/workflows/ci.yml
owent: The new versions of abseil-cpp, protobuf and gRPC requires C++14, and this version of abseil-cpp seems has some compatibility problem(See https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/5099430537/jobs/9167125807 for details). So I keep the same setting with gRPC here.

.github/workflows/ci.yml
owent: The latest versions of protobuf and gRPC require the new abseil-cpp, so we must enable it when using these versions.

ci/setup_grpc.sh
owent: Thanks and fixed.

.github/workflows/ci.yml
lalitb: But `WITH_ABSEIL` cmake option will enable `HAVE_ABSEIL` macro, and so enforce to use `abseil::variant` at API surface. Do you mean with latest versions of gRPC and protobuf, we can no longer use `nostd::variant` at API surface ?

CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
+# Prefer cmake CONFIG to auto resolve dependencies.
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
```
ThomsonTan, line 22: This flag seems new in CMake 3.15, but we support 3.1. Is this required?

.github/workflows/ci.yml
lalitb: Thanks for confirming. This looks concerning to me.

owent: Yes, it'a problem.Do you have any suggestion?

lalitb: Not for now. I would like to test the build with newer protobuf to fully understand the issue with our internal nostd::variant, and if we can do any fix in that class. Will try to test during this week. If not, good to discuss this in the next community meeting before deciding.
Thanks once again for providing all the details.

owent: Can we raise another issue to track this ABI compliance problem later? This problem existed before with abseil.

lalitb: Can we wait for Monday SIG meeting, also would like to test this once over weekend to have better understanding. Don't want to hold this PR any way as we need to support latest protobuf.

owent: Thanks.

ci/install_protobuf.sh
```diff
@@ -31,12 +31,41 @@ set -e
...
+fi
+
+if [[ ${PROTOBUF_VERSION/.*/} -ge 22 ]]; then
+  CPP_PROTOBUF_BUILD_OPTIONS=(${CPP_PROTOBUF_BUILD_OPTIONS[@]} "-Dprotobuf_MODULE_COMPATIBLE=ON")
+fi
+
+if [[ ${PROTOBUF_VERSION/.*/} -ge 22 ]]; then
+  export CPP_PROTOBUF_VERSION="${PROTOBUF_VERSION}"
+  CPP_PROTOBUF_PACKAGE_NAME="protobuf-${CPP_PROTOBUF_VERSION}"
+  CPP_PROTOBUF_BUILD_OPTIONS=(${CPP_PROTOBUF_BUILD_OPTIONS[@]} "-Dprotobuf_ABSL_PROVIDER=package")
```
marcalff, line 52: Please move this line to the previous test on PROTOBUF_VERSION,
to keep build options together with -Dprotobuf_MODULE_COMPATIBLE=ON

ci/install_protobuf.sh
```diff
@@ -31,12 +31,41 @@ set -e
...
+  "-Dprotobuf_MODULE_COMPATIBLE=ON"
+)
+
+if [ ! -z "${CXX_STANDARD}" ]; then
+    CPP_PROTOBUF_BUILD_OPTIONS=(${CPP_PROTOBUF_BUILD_OPTIONS[@]} "-DCMAKE_CXX_STANDARD=${CXX_STANDARD}")
+fi
+
+if [[ ${PROTOBUF_VERSION/.*/} -ge 22 ]]; then
+  CPP_PROTOBUF_BUILD_OPTIONS=(${CPP_PROTOBUF_BUILD_OPTIONS[@]} "-Dprotobuf_MODULE_COMPATIBLE=ON")
+fi
```
marcalff, line 47: Please add a comment to explain _why_ we need this.

Something like:
- protobuf version 22.xxx does ???
- protobuf version 3.yyy does ??? instead

ci/install_protobuf.sh
```diff
@@ -31,12 +31,41 @@ set -e
...
+
+if [[ ${PROTOBUF_VERSION/.*/} -ge 22 ]]; then
+  export CPP_PROTOBUF_VERSION="${PROTOBUF_VERSION}"
+  CPP_PROTOBUF_PACKAGE_NAME="protobuf-${CPP_PROTOBUF_VERSION}"
```
marcalff, line 51: Please add a comment to explain _why_ we need this.

Something like:
- protobuf 22.xxx ships packages named ???
- protobuf 3.yyy ships packages named ???

Also, please clarify if we should pick:

https://github.com/protocolbuffers/protobuf/releases/download/v22.3/protobuf-22.3.tar.gz

or

https://github.com/protocolbuffers/protobuf/archive/refs/tags/v22.3.tar.gz

If there is a difference, detail why.

If this is the same, can the following URL pattern work for both 22.xxx and 3.yyy ?

https://github.com/protocolbuffers/protobuf/archive/refs/tags/v22.3.tar.gz

I admit I am totally confused here by the protobuf version names, hence comments explaining what probuf is doing will help.

ci/setup_grpc.sh
```diff
@@ -7,24 +7,41 @@ set -e
...
+fi
 install_grpc_version=${new_grpc_version}
 install_dir='/usr/local/'
 build_shared_libs=''
+build_internal_abseil_cpp=1
+GRPC_BUILD_OPTIONS=()
 usage() { echo "Usage: $0 [-v <gcc-version>] [-i <install_dir>"] 1>&2; exit 1;}
 
-while getopts ":v:i:r:s:TH" o; do
+while getopts ":v:i:mp:r:s:TH" o; do
```
marcalff, line 23: Please update the usage() helper to show all options.

ci/setup_grpc.sh
```diff
@@ -59,22 +77,26 @@ git clone --depth=1 -b ${install_grpc_version} https://github.com/grpc/grpc
...
+    ABSEIL_CPP_BUILD_OPTIONS=(
+        -DCMAKE_BUILD_TYPE=Release
+        -DCMAKE_POSITION_INDEPENDENT_CODE=TRUE
+        -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR
+    )
+    if [ ! -z "$build_shared_libs" ]; then
+        ABSEIL_CPP_BUILD_OPTIONS=(${ABSEIL_CPP_BUILD_OPTIONS[@]} "-DBUILD_SHARED_LIBS=$build_shared_libs")
+    fi
+    cmake  ${ABSEIL_CPP_BUILD_OPTIONS[@]} ..
+    cmake --build . -j${nproc} --target install && popd
```
marcalff, line 94: Could all the logic related to abseil be moved to install_abseil.sh instead ?

marcalff (commented):
Thanks for all the changes.

This fix is extensive, touching a lot of parts in makefiles, and depends on how third party packages are organized, and their own dependencies ...

Obviously, you have a very good understanding of the constraints and technical details.

A reader can see what the CI and install scripts are doing, and this alone requires to reverse engineer the code already to understand what is done.

The critical part which is missing is to know why something is done.

Something like:

protobuf 22 does so-and-so with cmake packages, so as a result we need to use this-and-that to consume it

will go a very long way to grasp what is going on here, and will allow others to actually maintain this.

Please share this knowledge in comments.

lalitb:
Have tested the changes, and see relevant. There are 2 points I would like to discuss with this PR
- The internal nostd components are no longer usable for latest version of gRPC and onwards.
- We can't use C++11/14 to build the SDK. The minimum supported version would be  C++17.

ci/setup_grpc.sh
owent: I'm not sure, but some CI jobs use `setup_grpc.sh` but without `install_abseil.sh`. This installation existed before, I only add a option to allow gRPC find abseil-cpp from a existed installation path.
Maybe we can optimize abseil-cpp installation in another PR?

ci/setup_grpc.sh
owent: Thanks, and all options are added into usage message.

ci/install_protobuf.sh
owent: `protobuf_MODULE_COMPATIBLE=ON` is not required now, I remove it now.

ci/install_protobuf.sh
owent: It seems 3.XX.YY/4.XX.YY is alias for XX.YY. And there will be no 3.XX.YY after 22.
And the released source packages are moved into the tag of XX.YY and not 4.XX.YY after 22.
Also, there is only protobuf-22.3.tar.gz, and there is no protobuf-cpp-22.3.tar.gz any more.

marcalff:
Looking again at abseil, I don't think we need C++17 with abseil.

Per the abseil doc, C++14 is supported.

Per abseil code, ABSL_USES_STD_STRING_VIEW controls whether std::string_view is used when available in C++17, or whether a replacement is used for C++11/C++14.

If abseil was really broken for std::string_view in C++14, there would be several bug reports and complaints about it, but there are none.

What is more likely here, is that somehow, the CI build does not pass the C++14 standard flags all the way to the compiler, when building code, so that code ends up being compiled assuming C++17, therefore assuming std::string_view is available, causing breaks.

To clarify this, the CI scripts should be fixed to:
* always define a `CXX_STANDARD` from `ci.yaml`
* always use `sudo -E` instead of `sudo`, to propagate the compiler choice (CXX) and C++ mode used
* always use the `CXX_STANDARD` passed in to build a proper `CMAKE_CXX_STANDARD` option, when invoking cmake from various setup_xxx.sh helpers.

Once all this is done, it is likely the code will also build in C++14 mode, with the same `CXX` compiler and the same `C++14` mode used across all build scripts.

owent:
> Looking again at abseil, I don't think we need C++17 with abseil.
> 
> Per the abseil doc, C++14 is supported.
> 
> Per abseil code, ABSL_USES_STD_STRING_VIEW controls whether std::string_view is used when available in C++17, or whether a replacement is used for C++11/C++14.
> 
> If abseil was really broken for std::string_view in C++14, there would be several bug reports and complaints about it, but there are none.
> 
> What is more likely here, is that somehow, the CI build does not pass the C++14 standard flags all the way to the compiler, when building code, so that code ends up being compiled assuming C++17, therefore assuming std::string_view is available, causing breaks.
> 
> To clarify this, the CI scripts should be fixed to:
> 
> * always define a `CXX_STANDARD` from `ci.yaml`
> * always use `sudo -E` instead of `sudo`, to propagate the compiler choice (CXX) and C++ mode used
> * always use the `CXX_STANDARD` passed in to build a proper `CMAKE_CXX_STANDARD` option, when invoking cmake from various setup_xxx.sh helpers.
> 
> Once all this is done, it is likely the code will also build in C++14 mode, with the same `CXX` compiler and the same `C++14` mode used across all build scripts.

`CXX_STANDARD` is changed to 14 now.

marcalff:
Adding CVE tag: this fix is now required to upgrade to gRPC >= 1.53.0 to fix:
* https://github.com/open-telemetry/opentelemetry-cpp/issues/2231

marcalff:
@owent 

Please merge main into this PR to resolve conflicts, so we get up-to-date code.

Thanks.

owent:
> @owent
> 
> Please merge main into this PR to resolve conflicts, so we get up-to-date code.
> 
> Thanks.

Done

esigo:
Discussed in the weekly 17.07.2023. @lalitb and @marcalff will review. the issue with not being able to build with internal abseil will be handled in a separate issue.

marcalff (approved):
@owent 

Sorry the review took so long.

Overall comments below

1)

In `otlp_http_client.cc`, there is now an optional dependency on abseil,
for base 64 encoding.

Ok.

2)

In the OTLP exporter CMakeLists, there is an optional dependency on abseil.

Ok.

3)

In the OTLP exporter BUILD, for bazel, the dependency on abseil is required.

This is not optional in bazel, even when the C++ code might not even use abseil.

It would be better to have an optional dependency in the bazel build,
but I am not even sure if bazel supports this to begin with,
and know even less how to implement this ... so no solution to offer here.

Given that the C++ code in otlp_http_client.cc still include bazel optionally,
I guess having a mandatory-but-unused dependency on `absl/strings` from the BUILD file will be ok.

4)

About the changes in `cmake/*.cmake`,
I don't understand CMake well enough to make valuable comments in this area.

5)

Changes in the CI build itself looks ok to me, and beside, these only affects the CI build,
not the end user when using a release, so I think there is little technical risk here.

Having explicit `CXX_STANDARD` lines in `ci.yml` is much better, as the CI code is now clearer, and easier to maintain.

6)

As discussed previously,

when the build uses WITH_ABSEIL because of a dependency,
a side effect is also that the ABI is changed to also use WITH_ABSEIL,
at least in my understanding (to investigate further).

We may need to decouple:

* usage of WITH_ABSEIL in the implementation (protobuf)
* usage of WITH_ABSEIL in the ABI

This will be investigated in a separate issue and PR.

---

LGTM overall.

Thanks for the fix, approved.

CHANGELOG.md
```diff
@@ -170,7 +173,7 @@ Breaking changes:
   [#2118](https://github.com/open-telemetry/opentelemetry-cpp/pull/2118)
 * [CI] Upgrade Google Benchmark version from 1.6.0 to 1.7.1
   [#2116](https://github.com/open-telemetry/opentelemetry-cpp/pull/2116)
-* [CI] Upgrade Nlohmann JSON library version from 3.10.5 to 3.11.2
+* [CI] Upgrade Nlohmann JSON library version from 3.10.5 to 3.11.
```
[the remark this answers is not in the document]
owent: Thanks, this is a mistake.I have restored this change.

owent:
> @owent
> 
> Sorry the review took so long.
> 
> Overall comments below
> 
> 
> In `otlp_http_client.cc`, there is now an optional dependency on abseil, for base 64 encoding.
> 
> Ok.
> 
> 
> In the OTLP exporter CMakeLists, there is an optional dependency on abseil.
> 
> Ok.
> 
> 
> In the OTLP exporter BUILD, for bazel, the dependency on abseil is required.
> 
> This is not optional in bazel, even when the C++ code might not even use abseil.
> 
> It would be better to have an optional dependency in the bazel build, but I am not even sure if bazel supports this to begin with, and know even less how to implement this ... so no solution to offer here.
> 
> Given that the C++ code in otlp_http_client.cc still include bazel optionally, I guess having a mandatory-but-unused dependency on `absl/strings` from the BUILD file will be ok.
> 
> 
> About the changes in `cmake/*.cmake`, I don't understand CMake well enough to make valuable comments in this area.
> 
> 
> Changes in the CI build itself looks ok to me, and beside, these only affects the CI build, not the end user when using a release, so I think there is little technical risk here.
> 
> Having explicit `CXX_STANDARD` lines in `ci.yml` is much better, as the CI code is now clearer, and easier to maintain.
> 
> 
> As discussed previously,
> 
> when the build uses WITH_ABSEIL because of a dependency, a side effect is also that the ABI is changed to also use WITH_ABSEIL, at least in my understanding (to investigate further).
> 
> We may need to decouple:
> 
> * usage of WITH_ABSEIL in the implementation (protobuf)
> * usage of WITH_ABSEIL in the ABI
> 
> This will be investigated in a separate issue and PR.
> 
> LGTM overall.
> 
> Thanks for the fix, approved.

3. I not a professional of bazel, but can we use `select()` to switch the implementations of abseil-cpp?
4. The changes in `cmake/*.cmake` optimize some code reuse for protocols and remove the useless options of protoc and the more important change is fix the linking problems when the configure types of third party packages are different from building otel-cpp.
5. I'm not sure if it's easy to decouple the usage of WITH_ABSEIL in the implementation of OTLP exporters. But we can hide the symbols which use protobuf and do not export them to users.

ci/install_abseil.sh
```diff
@@ -3,19 +3,27 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-set -e
+set -ex
```
ThomsonTan, line 6: Do we need the `-x` option seems which seems for debugging?

CMakeLists.txt
```diff
@@ -337,9 +342,21 @@ if(WITH_PROMETHEUS)
...
+  find_package(absl CONFIG REQUIRED)
+endif()
+
 if(WITH_OTLP_GRPC OR WITH_OTLP_HTTP)
-  set(protobuf_MODULE_COMPATIBLE ON)
   find_package(Protobuf)
+  if(Protobuf_VERSION AND Protobuf_VERSION VERSION_GREATER_EQUAL "3.22.0")
+    if(NOT WITH_ABSEIL)
+      message(
+        FATAL_ERROR
```
ThomsonTan, line 354: Seems with new protobuf, the main SDK will no longer buildable with `WITH_ABSEIL=0` so this will be a required dependency. Can we separate the abseil dependency to protobuf related library only?

lalitb:
Sorry for the late reply to this. The changes for protobuf upgrade look good. We do have an issue with ABSEIL dependency for the API, but let's track it separately, as it is essential to upgrade with vulnerability in the older protobuf version.
The changes in cmake/*cmake are hard to understand for me too, and seems they are also not directly related to the protobuf upgrade. 

@owent Do you think we can separate out the cmake/*cmake related changes from this PR, and approve this solely for the protobuf upgrade? We can then target to merge it within this week if you can do these changes soon. Or else, if any of the approvers/maintainers are comfortable with the cmake changes here, I am fine with that too :) Sorry once again for lots of delays and asking for more changes.

CMakeLists.txt
owent: protobuf has a internal reference to abseil and will use it when building without `-Dprotobuf_ABSL_PROVIDER=package`, just like gRPC.I think it's better to keep the same version for protobuf, gRPC and otel-cpp.

owent:
@lalitb I create #2251 to track the problems fixed by `cmake/*` , and I will push the codes later after this PR is merged.

lalitb:
> @lalitb I create #2251 to track the problems fixed by `cmake/*` , and I will push the codes later after this PR is merged.

Thanks @owent. Will be reviewing it tomorrow. Had a glance, and the changes look good to me in general.

lalitb (approved):
The changes look good. Thanks for working on this. Couple of concerns in general to investigate separately:
 - Need to have a Github to investigate building "otel-cpp + OTLP exporter" with the latest gRPC/protobuf without the WITH_ABSEIL option. The aim is to enable the API interface of otel-cpp to function with the internal `nostd` components independently, without being reliant on the abseil dependency of protobuf.
 - upgrade the gRPC version for the bazel build.

.github/workflows/ci.yml
```diff
@@ -378,6 +385,31 @@ jobs:
...
+      with:
+        submodules: 'recursive'
+    - name: setup
+      env:
+        PROTOBUF_VERSION: '23.3'
+        ABSEIL_CPP_VERSION: '20230125.3'
+        CXX_STANDARD: '14'
+      run: |
+        sudo ./ci/setup_cmake.sh
+        sudo ./ci/setup_ci_environment.sh
```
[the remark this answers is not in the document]
owent: These two script do not use these environment variables. Do we still add `-E`?

> Need to have a Github to investigate building "otel-cpp + OTLP exporter" with the latest gRPC/protobuf without the WITH_ABSEIL option. The aim is to enable the API interface of otel-cpp to function with the internal nostd components independently, without being reliant on the abseil dependency of protobuf.

Done.

> upgrade the gRPC version for the bazel build.

This PR do not upgrade the default version of gRPC and protobuf, just add supporting for the new versions. Do you think we can keep the default versions of gRPC and protobuf unchanged to check the compatibility with the old versions?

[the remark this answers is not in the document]
lalitb: > This PR do not upgrade the default version of gRPC and protobuf, just add supporting for the new versions. Do you think we can keep the default versions of gRPC and protobuf unchanged to check the compatibility with the old versions?

Yeah good point. Should be fine. thanks.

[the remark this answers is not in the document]
lalitb: > These two script do not use these environment variables. Do we still add -E?

 Just thought of following a practice if in future we change these scripts to use env variable. But no strong opinion, we can change it later when situation arises.

ThomsonTan:
There seems some syntax error for modern protobuf without abseil. Is this expected, or modern protobuf does require abseil?

https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/5753570940/job/15599220018?pr=2163

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -65,6 +50,146 @@ namespace otlp
...
+{
+  if (nullptr == dest)
+  {
+    return;
+  }
+
+  Base64EscapeInternal(*dest, reinterpret_cast<const unsigned char *>(src.data()), src.size(),
+                       kBase64CharsBasic, '=');
+}
+
```
lalitb, line 193: Just curious, why to implement this and not use from abseil/google. Maintaining this code would be issue in future ?

lalitb, line 193: Ok I think we need these methods for building without WITH_ABSEIL and latest protobuf. I hope we have enough confidence with these methods, and tested well. And these are good to use even when `absl::Base64Escape` or `google::protobuf::Base64Escape` are available ? Can we add some unit-tests to test possible scenarios. Thanks.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -65,6 +50,146 @@ namespace otlp
...
+  if (slen == 0)
+  {
+    *olen = 0;
+    return (0);
+  }
+
+  n = (slen + 2) / 3;
+
+  if (n > (BASE64_SIZE_T_MAX - 1) / 4)
```
lalitb: where is BASE64_SIZE_T_MAX  defined ?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -65,6 +50,146 @@ namespace otlp
...
+  }
+  return ret;
+}
+
+// Base64Escape()
+//
+// Encodes a `src` string into a base64-encoded 'dest' string with padding
+// characters. This function conforms with RFC 4648 section 4 (base64) and RFC
+// 2045. See also CalculateBase64EscapedLen().
+static void Base64Escape(nostd::string_view src, std::string *dest)
```
lalitb, line 183: should `dest` by reference ?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -63,6 +51,146 @@ namespace otlp
 namespace
 {
 
+using Base64EscapeChars                              = const unsigned char[64];
```
lalitb, line 54: should these methods be moved to https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/common/string_util.h ?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -63,6 +51,146 @@ namespace otlp
...
+// Encodes a `src` string into a base64-encoded 'dest' string with padding
+// characters. This function conforms with RFC 4648 section 4 (base64) and RFC
+// 2045. See also CalculateBase64EscapedLen().
+static void Base64Escape(nostd::string_view src, std::string *dest)
+{
+  if (nullptr == dest)
+  {
+    return;
+  }
+
```
lalitb, line 189: early exit if `src` is empty ?

exporters/otlp/src/otlp_http_client.cc
owent: Sure. These codes are from out internal repo. I will also copy the unit tests here too.

exporters/otlp/src/otlp_http_client.cc
owent: There are a lot of codes here, do you think it's better to move it into sdk?

lalitb: Yeah good point, not required. We can move it later if need to share with other components. Thanks.

exporters/otlp/src/otlp_http_client.cc
owent: I think it's better to keep the same parameters as it in abseil.

exporters/otlp/src/otlp_http_client.cc
owent: base64 algorithm is moved into sdk now.

owent:
> * Need to have a Github task to investigate building "otel-cpp + OTLP exporter" with the latest gRPC/protobuf without the WITH_ABSEIL option. The aim is to enable the API interface of otel-cpp to function with the internal `nostd` components independently, without being reliant on the abseil dependency of protobuf.
> * upgrade the gRPC version for the bazel build.

I just tested this CI job and find protobuf can not work when both external and internal abseil-cpp exists.Because they share the same header paths. Maybe we can move headers os internal abseil-cpp into another path so we can keep ABI compatibility and use the internal and external abseil at the same time?

@lalitb This PR is ready to be reviewed again.

owent:
> There seems some syntax error for modern protobuf without abseil. Is this expected, or modern protobuf does require abseil?
> 
> https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/5753570940/job/15599220018?pr=2163

I just tested this CI job and find protobuf can not work when both external and internal abseil-cpp exists.Because they share the same header paths. Maybe we can move headers os internal abseil-cpp into another path so we can keep ABI compatibility and use the internal and external abseil at the same time?

@lalitb This PR is ready to be reviewed again and thanks,

exporters/otlp/src/otlp_http_client.cc
lalitb: Really sorry if my comment above was not clear. I did realize that this will bloat the sdk, and we should move it only when we need to share it with other components.

lalitb:
> Maybe we can move headers os internal abseil-cpp into another path, do not inline namespace otel_v1 and replace all usage absl:: in otel-cpp by a macro (OTABSL_NAMESPACE for example) and then we can define OTABSL_NAMESPACE to absl when we have HAVE_ABSEIL, or define it to absl::otel_v1. By this way we can keep ABI compatibility and use the internal and external abseil at the same time without break the ABI.

Thanks for your analysis @owent. This would be helpful.

owent:
@lalitb @marcalff Could you please review this PR again? I have moved base64 codes into sdk and there are a lot of changes, so I removed `od-to-merge` label. I think we can add it back after the new changes are reviewed.

lalitb:
> @lalitb @marcalff Could you please review this PR again? I have moved base64 codes into sdk and there are a lot of changes, so I removed `od-to-merge` label. I think we can add it back after the new changes are reviewed.

@owent  - Seems you missed my comment here - https://github.com/open-telemetry/opentelemetry-cpp/pull/2163#discussion_r1284985513 . Actually, I didn't ask for moving base64 from otlp to sdk :)

owent:
> > @lalitb @marcalff Could you please review this PR again? I have moved base64 codes into sdk and there are a lot of changes, so I removed `od-to-merge` label. I think we can add it back after the new changes are reviewed.
> 
> @owent - Seems you missed my comment here - [#2163 (comment)](https://github.com/open-telemetry/opentelemetry-cpp/pull/2163#discussion_r1284985513) . Actually, I didn't ask for moving base64 from otlp to sdk in my subsequent comments. But if you want, we can keep it in sdk for now, and have a separate PR to move back to otlp. Rest of the PR looks good.

Thanks. I can move it back to otlp exporter later in another PR.

PR #2166, opened 2023-05-31
[EXPORTER] Support empty arrays in `OtlpRecordable` attributes

mbrobbel:
## Changes

This PR changes behavior for empty array attribute values in `OtlpRecordable`. Empty array attribute values did not set the  `AnyValue` field to `ArrayValue`.

However, following the spec it seems we should forward empty arrays.

https://github.com/open-telemetry/opentelemetry-specification/blob/ce2c5946faa9c961ea01d89c7aed83621865b67b/specification/common/README.md?plain=1#L37-L39

Invoking `mutable_array_value()` before looping over the values sets the `AnyValue` variant to `ArrayValue` regardless of the number of items in the iterator.

### DIFF
--- a/exporters/otlp/src/otlp_populate_attribute_utils.cc
+++ b/exporters/otlp/src/otlp_populate_attribute_utils.cc
@@ -67,58 +67,66 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
   }
   else if (nostd::holds_alternative<nostd::span<const uint8_t>>(value))
   {
+    auto array_value = proto_value->mutable_array_value();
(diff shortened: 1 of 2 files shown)


marcalff (commented):
Very clean PR, thanks.

Please expand the unit tests to cover every data types beside `std::vector<int64_t>`.

This will help to catch bugs if somehow the fix for other alternatives gets lost, with bad merges.

owent (approved):
LGTM and thanks.

lalitb (approved):

ThomsonTan (approved):

PR #2178, opened 2023-06-05
[EXPORTER] Fixed HTTP CURL for 32bits platforms

### LINKED ISSUE #2175: set option <30120> failed: <A libcurl function was given a bad argument>
Hey all, 

There is an error message `ext\src\http\client\curl\http_operation_curl.cc:548 CURL, set option <30120> failed: <A libcurl function was given a bad argument>` while running the `example_otlp_http` example with Win32 on my Windows 10. I have already debug it and found that after add these follow codes, it works well.

```
    #ifdef WIN32
    rc = SetCurlLongOption(CURLOPT_POSTFIELDSIZE, req_size);
    #else
    rc = SetCurlLongOption(CURLOPT_POSTFIELDSIZE_LARGE, req_size);
    #endif
```

It should be a 32-bit compatibility problem.

marcalff:
Fixes #2175

## Changes

Fixed the HTTP CURL implementation to use the proper data types, when using options with type `curl_off_t`.

This fixes 32 bits builds, like WIN32.

### DIFF
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -269,6 +269,8 @@ private:
 
   CURLcode SetCurlLongOption(CURLoption option, long value);
 
+  CURLcode SetCurlOffOption(CURLoption option, curl_off_t value);
+
   const char *GetCurlErrorMessage(CURLcode code);
 
   std::atomic<bool> is_aborted_;   // Set to 'true' when async callback is aborted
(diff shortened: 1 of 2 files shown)


owent (approved):
LGTM and thanks.

lalitb (approved):

ThomsonTan (approved):

Korevishal:
How to use this change? is it available with 1.9.1 release ?
I believe I have to sync the main and build the Openetelemetry binaries, do I need to call SetCurlOffOption explicitly from my client ? if yes what values would I need to pass.
I am also seeing this issue on 32 bit windows

PR #2182, opened 2023-06-06
[SDK] Header files cleanup, use forward declarations

cngzhnp:
Fixes #2043

## Changes

Clean up header files in SDK folder to use as much as possible forward declarations.


lalitb:
@cngzhnp Just curious, how are you finding out the indirect header includes ?

marcalff (approved):
Reviewed all 172 files changed.

LGTM, thanks for the cleanup.

lalitb:
> @cngzhnp Just curious, how are you finding out these indirect header includes ?

@cngzhnp Could you please reply, this will help review in better way :)

cngzhnp:
> > @cngzhnp Just curious, how are you finding out these indirect header includes ?
> 
> @cngzhnp Could you please reply, this will help review in better way :)

Sorry for the late response, actually I was doing manually which means it is not perfect. However I could share my following these steps like below:

- Just pass through each header file and find the incomplete types.
- Only forward declaration is enough and remove unnecessary header files.
- If dependent header file is already included in the header file, remove it from implementation file.

This could be done by using IWYU as well btw.

examples/jaeger/main.cc
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "opentelemetry/exporters/jaeger/jaeger_exporter_factory.h"
+#include "opentelemetry/sdk/trace/processor.h"
```
lalitb, line 5: We need to add "sdk/trace/exporter.h" too here ?

examples/otlp/grpc_log_main.cc
```diff
@@ -6,7 +6,9 @@
 #  include "opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_factory.h"
 #  include "opentelemetry/logs/provider.h"
 #  include "opentelemetry/sdk/logs/logger_provider_factory.h"
+#  include "opentelemetry/sdk/logs/processor.h"
```
lalitb, line 9: We need to add "opentelemetry/sdk/logs/exporter.h" too ?

examples/otlp/grpc_log_main.cc
```diff
@@ -6,7 +6,9 @@
...
 #  include "opentelemetry/sdk/logs/simple_log_record_processor_factory.h"
+#  include "opentelemetry/sdk/trace/processor.h"
```
lalitb, line 11: We need to add "opentelemetry/sdk/trace/exporter.h" too ?

examples/jaeger/main.cc
cngzhnp: I guess, I did not catch yet due to "order of refactoring header files" :).  Currently, I did it for only API and SDK folders but not for exporters folder. Probably, in that time, I will just remove **"sdk/trace/exporter.h"** file from **opentelemetry/exporters/jaeger/jaeger_exporter_factory.h** file and it will cause an compilation error in this file due to complete type problem.

examples/otlp/grpc_log_main.cc
cngzhnp: Same reason for that I mentioned above.

examples/otlp/grpc_log_main.cc
cngzhnp: Same reason for that I mentioned above.

examples/jaeger/main.cc
marcalff: @lalitb @cngzhnp 

The scope of this cleanup is mostly the content of sdk, with minor adjustments as needed to other parts (exporter, examples) when needed.

Lalit comments are on the examples area which will be cleaned up later, sure, but the comments are still valid.

@cngzhnp Please implement comments as indicated, so this review can complete, and the patch be merged.

This will avoid the extra burden, for everyone, of keeping these comments aside, remember them, and either verify the fix later or raise comments again when the content of examples is fixed as well.

4 lines to fix in a patch touching 172 files is very minor, so looking forward to a quick resolution on this, so that @lalitb can approve the overall change, allowing the merge to main soon.

Thanks both the the massive cleanup and review.

esigo (approved):
LGTM
thanks :)
nit: Jaeger exporter will be removed. no need to have changes there.

sdk/include/opentelemetry/sdk/logs/logger_provider.h
```diff
@@ -6,16 +6,13 @@
...
 #  include "opentelemetry/logs/logger_provider.h"
-#  include "opentelemetry/logs/noop.h"
 #  include "opentelemetry/nostd/shared_ptr.h"
-#  include "opentelemetry/sdk/common/atomic_shared_ptr.h"
-#  include "opentelemetry/sdk/logs/logger.h"
-#  include "opentelemetry/sdk/logs/logger_context.h"
-#  include "opentelemetry/sdk/logs/processor.h"
+#  include "opentelemetry/nostd/string_view.h"
+#  include "opentelemetry/sdk/resource/resource.h"
+#  include "opentelemetry/version.h"
```
lalitb, line 15: This file has reference to `KeyValueIterable`, - seems it is indirectly included.  Not a blocker, we can fix this separately.

sdk/include/opentelemetry/sdk/logs/multi_recordable.h
```diff
@@ -20,6 +19,8 @@ namespace sdk
 {
 namespace logs
 {
+class LogRecordProcessor;
+
```
lalitb, line 23: This file has reference to `opentelemetry::common::SystemTimestamp` - seems to be indirectly included. Not a blocker, but can be fixed separately.

lalitb (approved):

sdk/include/opentelemetry/sdk/logs/multi_recordable.h
marcalff: @lalitb

Good find.

I think we should automate this with include-what-you-use, because doing manual reviews checking all dependencies gets too tedious and error prone.

PR #2193, opened 2023-06-12
Fix OTLP HTTP exporting in sync mode.

### LINKED ISSUE #2191: OTLP http exporter may cause crash in sync mode
**Describe your environment** 
OS: Linux
Compiler: GCC 9

**Steps to reproduce**
Use otlp http exporter and make export step timeout.

**Additional context**
Sync mode Export function will create a stack varibale `session_result`, and will keep it's reference in the result_callback lambda function. However, the callback may be called after the function returned. 
https://github.com/open-telemetry/opentelemetry-cpp/blob/bd83aa367c3d48b23334ca0352ea06105e4925ac/exporters/otlp/src/otlp_http_client.cc#L712-L731

If the timeout is reached, the wait_for function will fail and the sync export process will end without calling the callback function. 
https://github.com/open-telemetry/opentelemetry-cpp/blob/bd83aa367c3d48b23334ca0352ea06105e4925ac/exporters/otlp/src/otlp_http_client.cc#L769-L785

Then, when the session finishes with an error finally(send failed with timeout in my case), it will call the callback and access a invalid stack address. In my case, the batch span processor thread is trying to `PopulateRequest` other spans, and `scope_spans` variable was destroyed by the callback in other thread which cause the program crash.

owent:
Fixes #2191 

## Changes

+ Use shared_ptr to hold the lifetime when using OTLP HTTP exporting in sync mode.


marcalff (commented):
Thanks for the fix.

From the bug report,

the lambda capture `[&session_result]` causes a capture of a reference to a local variable, causing the bug.

From the fix,

the lambda capture `[session_result]` now captures a shared pointer, presumably incrementing the reference counter upon capture.

From existing code,

the callback function in OtlpHttpClient::Export can be given to a session, with

```
auto session = createSession(message, std::move(result_callback));
```

and then the session, which owns the callback, is stored in `running_sessions_` by:

```
addSession(std::move(opentelemetry::nostd::get<HttpSessionData>(session)));
```

So, when some code paths return from ::Export() on timeout, the callback is kept elsewhere already, and local variables (in the old code) are destroyed.


Please:

1) Confirm my understanding is correct

2) Detail where / when is the destructor called for the lambda capture `[session_result]`, that would decrement the last reference to the session_result, and destroy the object created by make_shared in the fix ... I don't understand this part.

owent:
> Thanks for the fix.
> 
> From the bug report,
> 
> the lambda capture `[&session_result]` causes a capture of a reference to a local variable, causing the bug.
> 
> From the fix,
> 
> the lambda capture `[session_result]` now captures a shared pointer, presumably incrementing the reference counter upon capture.
> 
> From existing code,
> 
> the callback function in OtlpHttpClient::Export can be given to a session, with
> 
> ```
> auto session = createSession(message, std::move(result_callback));
> ```
> 
> and then the session, which owns the callback, is stored in `running_sessions_` by:
> 
> ```
> addSession(std::move(opentelemetry::nostd::get<HttpSessionData>(session)));
> ```
> 
> So, when some code paths return from ::Export() on timeout, the callback is kept elsewhere already, and local variables (in the old code) are destroyed.
> 
> Please:
> 
>     1. Confirm my understanding is correct
> 
>     2. Detail where / when is the destructor called for the lambda capture `[session_result]`, that would decrement the last reference to the session_result, and destroy the object created by make_shared in the fix ... I don't understand this part.

1. Yes, that's right.
2. The callback is stored in `HttpSessionData` which is holded by `OtlpHttpClient::running_sessions_`, `OtlpHttpClient::gc_sessions_` .We remove session data from `running_sessions_` only when we get a event or response from http client.And when wait for response in sync mode, `session_waker_.wait_for` may be timeout before we get any event from it.

marcalff (approved):
LGTM, thanks for the fix and clarifications.

lalitb (approved):
Looks good. Thanks for the fix. This part of code has lots of edge scenarios, would be good to extend the unit-tests to cover them sometime later.

ThomsonTan:
Seems this current fix requires a heap allocation, even this is sync Export. And the heap allocation is only for the edge case (timeout). Wondering is it possible to avoid the heap allocation on ExportResult here?

owent:
> Seems this current fix requires a heap allocation, even this is sync Export. And the heap allocation is only for the edge case (timeout). Wondering is it possible to avoid the heap allocation on ExportResult here?

By now, I have no idea without break the behaviour of return value.

Issue #1983, opened 2023-02-13
Prepare release, include fixes with abi breaking changes


marcalff:
Prepare future releases with breaking changes.

Due to expected changes in the spec, there will be breaking changes in the API/ABI.

Due to changes in the implementation, there are also known changes in the SDK interface, which can break applications when using shared libraries.

It is desirable to group many fixes that causes breaking changes at once, so listing here all the known issues that impact API/ABI.

It is also desirable to define the overall strategy to introduce breaking changes.

API/ABI:

* #2033 (API + SDK)
* #2032 (API + SDK)
* #1922 (API)
* #1416 (API)
* #828 (API) -> Providing alternate Span creation API to return nostd::unique_ptr.
* #2155
* #1753
* #2132
* #2279

SDK:

* https://github.com/open-telemetry/opentelemetry-cpp/issues/1845 (remove exporter options struct inheritance)
* https://github.com/open-telemetry/opentelemetry-cpp/issues/1402 (new members in exporter options struct)

lalitb:
Adding few more issues in the list -

* https://github.com/open-telemetry/opentelemetry-cpp/issues/1922 (API)
* https://github.com/open-telemetry/opentelemetry-cpp/issues/1416 (API)
* https://github.com/open-telemetry/opentelemetry-cpp/issues/828 (API) -> Providing alternate Span creation API to return nostd::unique_ptr.

marcalff:
On hold until the spec change is settled:

https://github.com/open-telemetry/opentelemetry-specification/pull/3186

If there is an ABI change, we nay need to expose both old and new ABI at the same time, by using something similar to

```
#define OPENTELEMETRY_BEGIN_OLD_NAMESPACE(VER) \
  namespace opentelemetry { inline namespace VER {

#define OPENTELEMETRY_END_OLD_NAMESPACE \
  }}

#define OPENTELEMETRY_OLD_NAMESPACE(VER) opentelemetry :: VER
```

and then expose old abi (for binary compatibility) and new abi, as follows:

```
OPENTELEMETRY_BEGIN_OLD_NAMESPACE(v1)
... declaration for class Span in v1 goes here, code unchanged ...
OPENTELEMETRY_END_OLD_NAMESPACE

#define OPENTELEMETRY_ABI_VERSION_NO 2

OPENTELEMETRY_BEGIN_NAMESPACE
... declaration for class Span in v2 goes here, with added methods ...
[...]

PR #2205, opened 2023-06-27
[API] Remove Meters

### LINKED ISSUE #2184: [API] Remove Meters
Related to:
* https://github.com/open-telemetry/opentelemetry-specification/issues/2232
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4919 (see methods to remove an instrument and a meter)

Context

Using opentelemetry-cpp 1.9.1
Using metrics, with asynchronous instruments.
Using callbacks located in a shared library.

Use case:
- Create a Meter
- Create an asynchounous instrument
- add a callback to the instrument

Measurements are taken invoking the callback, and are exported.

Then:
- remove the callback from the instrument
- release the last instrument reference
- release the last meter reference

From this point, no new measurements are called.

The OTLP HTTP exporter however, still exports data about the metrics no longer measured, as seen in the opentelemetry-collector logs in the endpoint.

In my understanding, this is due to the aggregations used:
- sum aggregation of async counters continues, reporting a constant value
- probably likewise for gauges and the last value (not tested)
- etc

Please provide:
- in the API/SDK, an operation to remove a synchronous / asynchronous instrument from a Meter,
- in the API/SDK, an operation to remove a Meter from a MeterProvider.

so that exporters no longer report anything about defunct Meter and Metrics.

This is blocking: in practice, using Metrics with shared libraries is not possible currently with opentelemetry-cpp.

marcalff:
Fixes #2184

## Changes

Changes:

* `CMake`: Add an experimental option, `WITH_REMOVE_METER_PREVIEW`.
* `API`:  expose the new method `MeterProvider::RemoveMeter()`.
* `SDK`: implementation for `MeterProvider::RemoveMeter()`.
* Cleanup misc usage of the `opentelemetry::v1` namespace, to make the implementation independent of the ABI version exposed.


lalitb:
- Just curious why is this issue blocker for the shared libraries. Or do we mean it is a blocker for scenarios when applications dynamically load (dlopen()) and unload (dlclose()) shared libraries, and the Meter should be removed once the shared library is unloaded?
 - Do we need to increment the ABI version for this change, as the new API is still not part of the specs (and mayn't be eventually)?  Instead, as of now, we can introduce it under macro say - OPENTELEMETRY_REMOVE_METERS. Once this is part of the specification, we can increment ABI version for this.

 - Looks like this PR is also introducing maintaining multiple ABI versions - maintaining all older versions would add complexity to the code, and need some discussion (refer - https://github.com/open-telemetry/opentelemetry-cpp/issues/1983#issuecomment-1440755915)

marcalff:
@lalitb 

> Thanks for the PR. Few questions before reviewing it :)
> 
>     * Just curious why is this issue blocker for the shared libraries. Or do we mean it is a blocker for scenarios when applications dynamically load (dlopen()) and unload (dlclose()) shared libraries, and the Meter should be removed once the shared library is unloaded?
> 

This is a blocker for scenarios when applications use `dlopen()` and `dlclose()` during normal operations (i.e., not just on the application startup and shutdown, but during the application workload).

Disclosure:

I happen to be instrumenting such an application, and in our case, the issue is so severe that we just can not use opentelemetry for metrics, **at all**, if this is not resolved.

The proper label can be either "blocking", or "priority:p0", no strong opinion here, as long as it gets addressed, hence the PR.

>     * Do we need to increment the ABI version for this change, as the new API is still not part of the specs (and mayn't be eventually)?  Instead, as of now, we can introduce it under macro say - OPENTELEMETRY_REMOVE_METERS. Once this is part of the specification, we can increment ABI version for this.
> 

This is a two-fold question:
* `RemoveMeter()` is not part of `ABI v1`, so when this gets added to the ABI, it must be under `ABI v2` in any case, as v1 can not change.
* This API is not (yet) part of the spec either, which raise the question if it will ever be there, and if so, if the spec will use the prototype proposed, or something similar but different. I agree that we need a feature flag to clearly identify apis that are not in the spec, to convey the notion that the api is experimental and subject to change, independently of the ABI version used.


>     * Looks like this PR is also introducing maintaining multiple ABI versions - maintaining all older versions would add complexity to the code, and need some discussion (refer - [Prepare release, include fixes with abi breaking changes #1983 (comment)](https://github.com/open-telemetry/opentelemetry-cpp/issues/1983#issuecomment-1440755915)). As of now, should be keep it separate from this PR ?

I will split this PR to keep only the fix itself, and move the ABI changes with #1983, for clarity, and to merge this PR sooner.

The discussion on ABI still needs to happen, as this is blocking many fixes, but sure, let's do that separately.

marcalff:
Now ready for review.

sdk/src/metrics/meter_context.cc
```diff
@@ -79,6 +80,26 @@ void MeterContext::AddMeter(std::shared_ptr<Meter> meter)
...
+  for (auto &meter : meters_)
+  {
+    auto scope = meter->GetInstrumentationScope();
+    if (!scope->equal(name, version, schema_url))
+    {
+      filtered_meters.push_back(meter);
+    }
+  }
+
+  meters_.swap(filtered_meters);
```
lalitb, line 100: nit - should we log a warning if the meter to be removed is not found?

lalitb (approved):
LGTM. Good to add a unit test. Also, should we enable this macro in the maintainers CI?

esigo (approved):
LGTM
Thanks :)

marcalff:
> LGTM. Good to add a unit test. Also, should we enable this macro in the maintainers CI?

Thanks for the review.

Added unit test, and coverage in maintainers CI.

sdk/src/metrics/meter_context.cc
marcalff: Thanks.

Printed an INFO message when removing a meter instead, this is better and provides more debug help if needed.

PR #2213, opened 2023-07-01
[EXPORTER] Prometheus: Add unit to names, convert to word

### LINKED ISSUE #1840: Prometheus exporter: Add unit to metric names, and convert unit to words

From the OTel prometheus [compatibility spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-metadata-1): 

> The Unit of an OTLP metric point MUST be added as [OpenMetrics UNIT metadata](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily). Additionally, the unit MUST be added as a suffix to the metric name, and SHOULD be converted to [base units](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#units-and-base-units) recommended by OpenMetrics when possible. The unit suffix comes before any type-specific suffixes.

For now, I would hold off trying to convert to base units (e.g. milliseconds to seconds), but would still encourage converting from abbreviations to full words for the unit (e.g. `ms` to `milliseconds`).

For example, a metric with name "foo", type=monotonic sum, unit=ms should have the name: `foo_milliseconds_total`.

The current behavior is that the name of the metric would be `foo_total`.

### LINKED ISSUE #2317: Prometheus exporter adds underscore suffix for metrics without unit
**Steps to reproduce**

If a metric does not have a unit set, the prometheus exporter will add an extra `_` suffix to the metric.  I have only second-hand knowledge of the issue, and don't have reproduction steps.

If the unit is empty, the name should not have an underscore or unit suffix added.

The metric `grpc.client.attempt.started` without a unit turns into:

```
# TYPE grpc_client_attempt_started_ counter
grpc_client_attempt_started_{grpc_method="grpc.testing.TestService/UnaryCall",grpc_target="dns:///10.52.1.55:14285"} 10 1694819876572
```

I believe this is because of this line: https://github.com/open-telemetry/opentelemetry-cpp/blob/0563a71ce3ea2c580f535996aeb8ce9246f2d703/exporters/prometheus/src/exporter_utils.cc#L45

We append an underscore regardless of if the unit is empty or not.

esigo:
Fixes #1840 Fixes #2317 (issue)

## Changes


exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +167,156 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+std::string PrometheusExporterUtils::ConvertRateExpressedToPrometheusUnit(
+    const std::string &rate_expressed_unit)
+{
+  if (rate_expressed_unit.find("/") == std::string::npos)
+  {
+    return rate_expressed_unit;
+  }
+
+  std::vector<std::string> rate_entities;
+  size_t pos = rate_expressed_unit.find("/");
```
marcalff, line 263: nit

do not execute the find twice, save the result in `pos` and test for `pos == std::string::npos` at the start of this method.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -55,7 +51,11 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
               nostd::get<sdk::metrics::SumPointData>(point_data_attr.point_data).is_monotonic_;
         }
         const prometheus_client::MetricType type = TranslateType(kind, is_monotonic);
-        metric_family.type                       = type;
+        prometheus_client::MetricFamily metric_family;
+        metric_family.type = type;
+        metric_family.name = MapToPrometheusName(metric_data.instrument_descriptor.name_,
+                                                 metric_data.instrument_descriptor.unit_, type);
+        metric_family.help = metric_data.instrument_descriptor.description_;
```
marcalff, line 58: This code moved inside the loop on data points.

The metric name, type, help, etc should be constant for all data points, so can this be moved outside ?

Even if `type` is needed and not available elsewhere, peek at the first data point before entering the loop on data points ?

marcalff (commented):
I don't understand the inner loops of
PrometheusExporterUtils::TranslateToPrometheus(),
and the change there, please clarify.

It seems the name mapping is done for every data point of a metric, instead of once per metric.

exporters/prometheus/src/exporter_utils.cc
dashpole: +1.  As long as it is safe to assume all data points have the same type, peeking at the first seems like the right thing to do.  I took that approach in https://github.com/open-telemetry/opentelemetry-cpp/pull/2288/files#diff-e92aebc994eed5c134b09ecde972d9db39a9bfc46edc6d93b0b5a4c98e4ff27fR46.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -114,8 +114,8 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
                 "invalid SumPointData type");
           }
         }
+        output.emplace_back(metric_family);
```
dashpole, line 117: Does the prometheus C++ library merge metric families?  We definitely don't want one metric family for each data point.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +167,156 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+                                                  {"TiBy", "tibibytes"},
+                                                  {"KBy", "kilobytes"},
+                                                  {"MBy", "megabytes"},
+                                                  {"GBy", "gigabytes"},
+                                                  {"TBy", "terabytes"},
+                                                  {"B", "bytes"},
+                                                  {"KB", "kilobytes"},
+                                                  {"MB", "megabytes"},
+                                                  {"GB", "gigabytes"},
+                                                  {"TB", "terabytes"},
```
dashpole, line 214: We've removed B and $ from mappings elsewhere, as B means `bel` in UCUM.  See https://github.com/open-telemetry/opentelemetry-java/pull/5719

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +167,156 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+
+  // Append prometheus unit if not null or empty.
+  if (!prometheus_equivalent_unit.empty() &&
+      sanitized_name.find(prometheus_equivalent_unit) == std::string::npos)
+  {
+    sanitized_name += "_" + prometheus_equivalent_unit;
+  }
+
+  // Special case - counter
+  if (prometheus_type == prometheus_client::MetricType::Counter &&
```
dashpole, line 304: Counters MUST end in _total, so we need to be less tolerant here.  We can only skip the _total suffix if the name **ends** in _total.  Since we need unit suffixes to come before type suffixes, one approach would be to trim _total suffixes from the name before appending the unit suffix, and finally append the final _total suffix.

dashpole (commented):
Apologies for duplicating work, and thanks for working on this!  It looks like this fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/2287 as well, so i'll close my PR.

exporters/prometheus/src/exporter_utils.cc
esigo: thanks, removed.

exporters/prometheus/src/exporter_utils.cc
esigo: thanks, will not add total unless if it has already.

exporters/prometheus/src/exporter_utils.cc
esigo: thanks, creating new family only if it doesn't exist aleady.

exporters/prometheus/src/exporter_utils.cc
esigo: thanks, moved back to where it was.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +175,159 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+
+  std::string converted_metric_unit_name = RemoveUnitPortionInBraces(raw_metric_unit_name);
+  converted_metric_unit_name = ConvertRateExpressedToPrometheusUnit(converted_metric_unit_name);
+
+  return CleanUpString(GetPrometheusUnit(converted_metric_unit_name));
+}
+
+std::string PrometheusExporterUtils::GetPrometheusUnit(const std::string &unit_abbreviation)
+{
+  static std::map<std::string, std::string> units{// Time
```
owent, line 200: Is is better to use unordered_map and string_view here?

esigo: thanks, done.

owent: What I mean is use hash map to find unit conversion here and below, not just the return type of `TranslateToPrometheus`.

string_view may not be necessary here, because all string value is small. In some old implementation it will use COW and the new implementation will use small string optimization to reduce malloc.

punya:
I believe this would also fix #2317.

marcalff:
Hi @esigo 

Please merge with a recent main to resolve conflicts, so this PR can be reviewed and merged.

Thanks.

esigo:
> Hi @esigo 
> 
> Please merge with a recent main to resolve conflicts, so this PR can be reviewed and merged.
> 
> Thanks.
> 

Hi @marcalff thanks for the reminder. I finish this over the weekend.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +175,159 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+      sanitized_name.find(prometheus_equivalent_unit) == std::string::npos)
+  {
+    sanitized_name += "_" + prometheus_equivalent_unit;
+  }
+
+  // Special case - counter
+  if (prometheus_type == prometheus_client::MetricType::Counter)
+  {
+    auto t_pos           = sanitized_name.rfind("_total");
+    bool ends_with_total = t_pos == sanitized_name.size() - 6;
```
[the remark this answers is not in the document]
esigo: @dashpole the output will be `foo_bar_total_seconds_total`.

exporters/prometheus/src/exporter_utils.cc
esigo: @owent could you please check my last commit?

exporters/prometheus/src/exporter_utils.cc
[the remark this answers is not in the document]
dashpole: Ideally, you would trim _total from counters before appending the unit, and then always add _total after the unit for counters.

exporters/prometheus/src/collector.cc
```diff
@@ -39,7 +39,7 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   reader_->Collect([&result](sdk::metrics::ResourceMetrics &metric_data) {
     auto prometheus_metric_data = PrometheusExporterUtils::TranslateToPrometheus(metric_data);
     for (auto &data : prometheus_metric_data)
-      result.emplace_back(data);
+      result.emplace_back(data.second);
```
owent, line 42: Is it better to use `std::move(data.second)` here? `::prometheus::MetricFamily` may contains a lot of data, and copy these data may have performance problem.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -27,35 +29,42 @@ namespace metrics
...
  */
-std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateToPrometheus(
-    const sdk::metrics::ResourceMetrics &data)
+std::unordered_map<std::string, prometheus_client::MetricFamily>
+PrometheusExporterUtils::TranslateToPrometheus(const sdk::metrics::ResourceMetrics &data)
 {
 
   // initialize output vector
-  std::vector<prometheus_client::MetricFamily> output;
+  std::unordered_map<std::string, prometheus_client::MetricFamily> output;
```
owent, line 37: Can we use `output.reserve()` first to reduce rehash?

dashpole:
@esigo Is there any way I can help with this?  I would love to get this over the finish line soon.

esigo:
> @esigo Is there any way I can help with this? I would love to get this over the finish line soon.

@dashpole thanks for the reminder. Becasue of the conflict I need to redo the PR. I finish this over the weekend.

exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
```diff
@@ -45,6 +56,79 @@ class PrometheusExporterUtils
...
+                                         const std::string &unit,
+                                         ::prometheus::MetricType prometheus_type);
+
+  /**
+   * A utility function that returns the equivalent Prometheus name for the provided OTLP metric
+   * unit.
+   *
+   * @param raw_metric_unitName The raw metric unit for which Prometheus metric unit needs to be
+   *     computed.
+   * @return the computed Prometheus metric unit equivalent of the OTLP metric un
```
dashpole, line 69:
```suggestion
   * @return the computed Prometheus metric unit equivalent of the OTLP metric unit
```

dashpole (approved):
Thanks!

lalitb (approved):
Thanks for the fix.

marcalff (approved):

PR #2215, opened 2023-07-04
[BUILD] Add `OPENTELEMETRY_CPP_FOUND` into cmake CONFIG file.

### LINKED ISSUE #2212: OpenTelemetry's cmake file claims to publish OPENTELEMETRY_CPP_FOUND/etc. but really publishes OPENTELEMETRY-CPP_FOUND - with a dash.
**Describe your environment** 
Users of this library will look at the CMAKE file cmake/opentelemetry-cpp-config.cmake.in (as did [opentelemetry-cpp-contrib](https://github.com/open-telemetry/opentelemetry-cpp-contrib!) and try to use that target, but if you print CMAKE variables, you'll see that the _FOUND variable actually has the dash, which makes sense if you look at the find_package_handle_standard_args call at the bottom of [this file](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/opentelemetry-cpp-config.cmake.in).

**Steps to reproduce**
Print the CMAKE variables after running `find_package(opentelemetry-cpp CONFIG QUIET)`, and you'll see: `-- opentelemetry-cpp_FOUND=1` not `-- opentelemetry_cpp_FOUND=1`.

The documented CMAKE variables (which have an underscore rather than a dash) should match the actual CMAKE variables (with are with dashes).

They do not match.

**Additional context**
This leads to other dependent repos failing to include opentelemetry-cpp correctly.

owent:
Fixes #2212 

## Changes

+ Add `OPENTELEMETRY_CPP_FOUND` in cmake CONFIG file.


esigo (approved):
LGTM
Thanks :)

ThomsonTan (approved):

PR #2224, opened 2023-07-07
[API] Add InstrumentationScope attributes in MeterProvider::GetMeter()

### LINKED ISSUE #2033: [Metrics API/SDK] Add InstrumentationScope attributes in MeterProvider::GetMeter()
As mentioned in the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#get-a-meter), it should be possible to optionally pass the InstrumentationScope attributes while creating/fetching a Meter through MeterProvider.

```
Get a Meter

This API MUST accept the following parameters:
 * name (required)
 * version (optional)
 * [since 1.4.0] schema_url (optional)
 * [since 1.13.0] attributes (optional)
```

With #2004, the [InstrumentationScope ](https://github.com/open-telemetry/opentelemetry-cpp/blob/7db57f9d2b034848a901861ca3ba0e45dff6deae/sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h#L42) class already support storing the attributes. 

The relevant metric exporters -  OTLP, OStream, Prometheus(?)  should be modified to support sending/dumping these attributes.

Also this would be an API/ABI breaking change (modifying/overloading the virtual method in the API), so need to be planned accordingly.

marcalff:
Fixes #2033

## Changes

API:

* Implemented a new `GetMeter()` api with 4 parameters, only available in `ABI VERSION 2`
* Renamed parameters to be consistent with the spec.

CI:

* Added a maintainer build to test `ABI VERSION 2`.


marcalff:
Proof of concept, ready for review and discussion.

This code assumes that #2222 is already merged in main, and is an example of how to deliver fixes with breaking changes.

lalitb:
Thank you for the POC for supporting multiple ABI versions. I don't see any issue with this implementation to support the idea. However, I have concerns about the long-term management of these versions. Given that there may be ABI breaking changes that affect various files and components in the repository, we are using ifdef macros throughout these areas. As time goes on, the code may become increasingly challenging to handle.

The question we need to address is whether it is truly necessary to maintain backward compatibility for ABI compliance. At present, when we talk about ABI compliance for the otel-api, it encompasses the following aspects:
 - using nostd constructs at the API surface for instrumented libraries that are built with the same otel ABI version but potentially with different compilers or versions. 
 - ensure to increment the ABI version whenever there is a breaking change in otel-api. This ensures that any ABI-only, non-API breaking changes will result in a link error if someone attempts to link libraries that are instrumented with different otel ABI versions. And so, this will prevent runtime errors from occurring.

esigo (commented):
Thanks Mark for the great work as always. changes in this PR and #2222 look good.
However, I have similar concern as @lalitb. I think it will be very challenging to maintain backward compatibility. Even if we choose to follow this path, we need to decide how far we are going to be backward compatible. Let's discuss in the next SIG meeting :)

marcalff:
Thanks for looking at the proposal.

Some comments below, organized by topics.

# Maintenance complexity

Assume we implement 10 breaking changes, for 10 different unrelated reasons.

The proposal is **not** to introduce a feature flag for each change,
which would lead to power(2, 10) build configurations to test.

Instead, the proposal is to define only one additional configuration,
ABI v2, that contains these 10 changes, on top of the existing ABI v1.

This point alone makes a huge difference (and, keep in mind that if we do not
adopt the proposal, chances are that we will end up with independent feature flags anyway,
which is much worse).

By the time ABI v2 is introduced,
builds with ABI v1 and ABI v2 are necessary in CI.

Now, as long as ABI v2 is still in experimental status,
there is no strong reason to test all the possible feature combinations
in the experimental ABI v2, the following is sufficient:
- multiple CI builds to test ABI v1 combinations (as of today)
- a single CI build, with all features enabled, to test ABI v2

A single build in CI is not an issue, and should not be a burden.

By the time ABI v2 is changed from experimental to stable,
it may now contain 15 breaking changes.

*IF* we want to maintain concurrently ABI v1 and ABI v2 in stable status for
some time, the CI build will then need:
- multiple CI builds to test ABI v1 combinations (as of today)
- multiple CI builds to test ABI v2 combinations (as of today)

in effect increasing the CI load.

The next logical step is to declare ABI v1 deprecated,
and encourage migration from ABI v1 to ABI v2.

In this phase, the CI build can be simplified to:
- a single CI build, with all features enabled, to test deprecated ABI v1
- multiple CI builds to test ABI v2 combinations (as of today for ABI v1)

In conclusion, and assuming we repeat the pattern, the complexity at any
time is:
- light (one build with all options) CI for deprecated ABI N-1
- full (testing all feature options) CI for stable ABI N
- light (one build with all options) CI for experimental ABI N+1

Today we have:
- full (testing all feature options) CI for frozen and immutable ABI v1

In conclusion, yes, this increases complexity by having more combinations
to support, but to a very limited extend that should be manageable.

# Maintain several ABI

A valid discussion point is to decide if we do or do not need to maintain
multiple ABI versions at the same time.

To evaluate this, let's first consider a typical usage scenario,
then see how each solution works.

## Use case

Assume that a vendor delivers a library, instrumented for opentelemetry-cpp,
and delivers a binary package.

All versions are for illustration.

(1)

Vendor LIB distributes a package compiled using opentelemetry-cpp 1.10.0.

(2)

Then, a different party, vendor APP, uses the package to build an application,
using the opentelemetry-cpp library available at time of build.

Vendor APP:
- links with the package produced by LIB in (1)
- configures the opentelemetry-cpp SDK
- uses opentelemetry-cpp 1.15.0

## Changing the ABI

In this scenario:
- ABI v1 as shipped in version 1.10.0 has one binary signature
- ABI v1 as shipped in version 1.15.0 has a different binary signature

and the APP crashes.

We all know this is to avoid, which is why no breaking changes are introduced
in stable ABI.

This solution is not viable, the point of this section is to repeat why.

## Supporting only one ABI at a time

Assume a breaking change is implemented in opentelemetry-cpp 1.12.0,
which changes the ABI from v1 to v2.

In this scenario:
- LIB is compiled against opentelemetry-cpp 1.10.0, using ABI v1
- APP is compiled against opentelemetry-cpp 1.15.0, using ABI v2

The code links properly, and does not crash.

At runtime however, LIB is not instrumented,
as all instrumentation calls lands in the noop code shipped by default in ABI
v1.

Setting a global provider singleton in the APP, using in ABI v2,
does not allow LIB, using ABI v1, to see the SDK.

My opinion on this is that this is not viable.

We can't ask applications to link with an old library, like APP using
opentelemetry-cpp 1.10.0 or 1.12.0 in this case.

We can't tell people producing a LIB binary package for a library instrumented for
opentelemetry that their instrumentation may just be ignored,
by no fault of their own, at any point in time ... and we can not ask them to
release a different package again just because a newer ABI was defined.


## Supporting multiple ABI

In this case:

When compiling with opentelemetry-cpp 1.10.0, only one ABI is available,
so that LIB is using ABI v1.

When compiling with opentelemetry-cpp 1.15.0, two ABI are available, ABI v1
and ABI v2.

The application, APP, decides to compile against opentelemetry-cpp 1.15.0
using ABI v1.

The application links, does not crash, AND provides instrumentation in LIB.

By the time opentelemetry-cpp 1.20.0 is released, maybe ABI v1 will be
deprecated, so that LIB shipping a newer package for a newer library version
will then decide to use ABI v2 instead.

The point here is that deciding to use one ABI versus another is a choice made
deliberately by the package and/or application at build time, not by opentelemetry-cpp.

marcalff:
Fix blocked by:
* #2222

api/include/opentelemetry/metrics/meter_provider.h
```diff
@@ -20,15 +23,56 @@ class MeterProvider
...
+  {
+    nostd::span<const std::pair<nostd::string_view, common::AttributeValue>> attributes_span{
+        attributes.begin(), attributes.end()};
+
+    common::KeyValueIterableView<
+        nostd::span<const std::pair<nostd::string_view, common::AttributeValue>>>
+        iterable_attributes{attributes_span};
+
+    return GetMeter(name, version, schema_url, &iterable_attributes);
+  }
```
lalitb, line 73: Not sure, but good to check if we need more overloads as done for [StartSpan](https://github.com/open-telemetry/opentelemetry-cpp/blob/61e7429d6362bd2dc166ca81f0cb28963eac0ee8/api/include/opentelemetry/trace/tracer.h#L50C1-L57C4) - to pass the instance of custom classes which are compliant to `is_key_value_iterable` trait?

api/include/opentelemetry/metrics/meter_provider.h
```diff
@@ -20,15 +23,111 @@ class MeterProvider
...
+   * Gets or creates a named Meter instance (ABI).
+   *
+   * @since ABI_VERSION 2
+   *
+   * @param[in] name Meter instrumentation scope
+   * @param[in] version Instrumentation scope version, optional
+   * @param[in] schema_url Instrumentation scope schema URL, optional
+   * @param[in] attributes Instrumentation scope attributes, optional
+   */
+  virtual nostd::shared_ptr<Meter> DoGetMeter(
```
lalitb, line 39: Is this intended to be public method?

marcalff: Still investigation options here:

* Naming this GetMeter() causes too much confusion with overloaded prototypes, so that {{"foo", 1}} does not use std::initializer but attempts to use this one instead, causing a break.
* If we keep GetMeter() only on the API, and have DoGetMeter() as virtual method, it can be made protected instead

marcalff: I am finding out that having too many overloaded prototypes, with optional parameters, present both at the API level and at the SDK level utterly confuses the compiler, when calling a method on the SDK class.

This is also confusing for humans, because nobody knows if GetMeter(prototype X) and GetMeter(prototype Y) are part of the API or the ABI.

To discuss, currently leaning towards:

* GetMeter() only on API, delegates to virtual DoGetMeter(ABI prototype)
* Only DoGetMeter() is present on the SDK class
* GetMeter() is public
* DoGetMeter() is protected

The point is to avoid a public GetMeter() method in the SDK that hides other GetMeter() from the API.

Given the complexity of:
* https://en.cppreference.com/w/cpp/language/overload_resolution

I think we should avoid pushing too much complexity with everything having just one name into this.

lalitb: Not sure if I understand the problem very well with GetMeter overloads, but can't we add the overloads as done for GetLogger here - https://github.com/open-telemetry/opentelemetry-cpp/blob/a4961c421c9687ea5b750c8df3374cd842b1d5a8/api/include/opentelemetry/logs/logger_provider.h#L39C1-L68C1. What is actually breaking with this ?

marcalff: I found out the root cause of the overload issues.

Fixed with this in the SDK:

```
  /*
    Make sure GetMeter() helpers from the API are seen in overload resolution.
  */
  using opentelemetry::metrics::MeterProvider::GetMeter;
```

api/include/opentelemetry/metrics/meter_provider.h
marcalff: Fixed

api/include/opentelemetry/metrics/meter_provider.h
```diff
@@ -20,19 +23,113 @@ class MeterProvider
...
    */
-  virtual nostd::shared_ptr<Meter> GetMeter(nostd::string_view library_name,
-                                            nostd::string_view library_version = "",
-                                            nostd::string_view schema_url      = "") noexcept = 0;
+  virtual nostd::shared_ptr<Meter> GetMeter(nostd::string_view name,
+                                            nostd::string_view version    = "",
+                                            nostd::string_view schema_url = "") noexcept = 0;
+#endif
+
 #ifdef ENABLE_REMOVE_METER_PREVIEW
```
lalitb, line 129: Not related to this PR, but I guess this is a candidate to be moved to ABI v2.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -118,6 +118,19 @@ class AttributeMap : public std::unordered_map<std::string, OwnedAttributeValue>
...
 
+  // Construct attribute map and populate with optional attributes
+  AttributeMap(const opentelemetry::common::KeyValueIterable *attributes) : AttributeMap()
+  {
+    if (attributes != nullptr)
+    {
+      attributes->ForEachKeyValue(
+          [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
+            SetAttribute(key, value);
+            return true;
```
lalitb, line 129: nit - some code duplication, which can be moved to a separate utils method. But don't see it really required.

lalitb (approved):
LGTM. nicely done.

api/include/opentelemetry/metrics/meter_provider.h
marcalff: Indeed, I will fix this is a separate PR to use ABI v2 instead.

marcalff:
@owent @esigo @ThomsonTan 

Do you want to have a second look, or ok to merge now ?

owent (approved):
LGTM and thanks.

PR #2255, opened 2023-08-04
[SDK] Cardinality limits for metrics streams (Sync Instruments + Delta Temporality)

lalitb:
## Changes

#2347 


lalitb:
This is ready for review now.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -66,6 +78,11 @@ class AttributesHashMap
       return it->second.second.get();
     }
 
+    if (hash_map_.size() >= attributes_limit_)
```
marcalff, line 81: Per the spec:

"
The SDK MUST create an Aggregator with the overflow attribute set prior to reaching the cardinality limit and use it to aggregate events for which the correct Aggregator could not be created. The maximum number of distinct, non-overflow attributes is one less than the limit, as a result.
"

So, with a limit of 10, there should be room for 9 + 1 aggregations.

The unit tests show a size of 11 instead of 10.

How about:

```
if (hash_map_.size() + 1 >= attributes_limit_)
```

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -23,6 +23,15 @@ namespace metrics
 
 using opentelemetry::sdk::common::OrderedAttributeMap;
 
+constexpr uint16_t kAggregationCardinalityLimit = 2000;
```
marcalff, line 26: Use size_t instead of uint16_t, like the rest of the code ?

marcalff (commented):
Early review comments.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -23,6 +23,15 @@ namespace metrics
...
+const std::string kAggregationCardinalityLimitOverflowError =
+    "Maximum data points for metric stream exceeded. Entry added to overflow";
+const std::string kAttributesLimitOverflowKey   = "otel.metrics.overflow";
+const std::string kAttributesLimitOverflowValue = "true";
```
marcalff, line 30: Per the spec:

"
An overflow attribute set is defined, containing a single attribute otel.metric.overflow having (boolean) value true, which is used to report a synthetic aggregation of the metric events that could not be independently aggregated because of the limit.
"

The overflow value should be a boolean true, not a string "true".

marcalff (commented):
Overflow value

lalitb:
Thanks for the comments @marcalff. They are valid, have fixed them.

marcalff (approved):
LGTM, thanks for the fix.

esigo (approved):
LGTM
Thanks :)

PR #2258, opened 2023-08-08
[WIP]check if the variant holds_alternative SumPointData to prevent std::bad_variant_access

Sohit1212:
Fixes # (issue)
std::bad_variant_access on using otlp_http_metrics_exporter

## Changes
Add condition to check if variant hold alternative


exporters/otlp/src/otlp_metric_utils.cc
```diff
@@ -60,21 +60,25 @@ void OtlpMetricUtils::ConvertSumMetric(const metric_sdk::MetricData &metric_data
     proto::metrics::v1::NumberDataPoint *proto_sum_point_data = sum->add_data_points();
     proto_sum_point_data->set_start_time_unix_nano(start_ts);
     proto_sum_point_data->set_time_unix_nano(ts);
-    auto sum_data = nostd::get<sdk::metrics::SumPointData>(point_data_with_attributes.point_data);
-
-    if ((nostd::holds_alternative<int64_t>(sum_data.value_)))
+    if ((nostd::holds_alternative<sdk::metrics::SumPointData>(
+            point_data_with_attributes.point_data)))
```
lalitb, line 64: Just curious, when are you getting this error. As `ConvertSumMetrics()` is called for `metric_sdk::AggregationType::kSum` aggregation, and is always supposed to contain `SumPointData`. If you are getting this error, there may be problem somewhere else.

lalitb:
@Sohit1212  - Can you please reply to the comment above? Or the example code to reproduce the issue.

marcalff:
We need some clarifications about the exact original issue, the root cause, and how to reproduce the issue, before making such changes.

The risk here is that the bad variant access can happen due to a totally different reason, for example when mixing different variant implementations (there are many) in the build, so the code change may not address the real problem.

Closing the PR for now.

Feel free to re-open when more details are available.

PR #2259, opened 2023-08-08
sdk: include batch_span_processor_options to fix build

mfrw:
We should include `batch_span_processor_options` in the sdk as the `otel_ngx_module` is unable to compile.
I think, instead of adding an include in that project, we should add it in the sdk.

```bash
../otel-cpp-contrib/opentelemetry-cpp-contrib-866351ce2422a9d953ded4447bcdea0e587d6a53/instrumentation/nginx/src/otel_ngx_module.cpp: In function 'std::unique_ptr<opentelemetry::v1::sdk::trace::SpanProcessor> CreateProcessor(const OtelNgxAgentConfig*, std::unique_ptr<opentelemetry::v1::sdk::trace::SpanExporter>)':
../otel-cpp-contrib/opentelemetry-cpp-contrib-866351ce2422a9d953ded4447bcdea0e587d6a53/instrumentation/nginx/src/otel_ngx_module.cpp:1039:41: error: aggregate 'opentelemetry::v1::sdk::trace::BatchSpanProcessorOptions opts' has incomplete type and cannot be defined
 1039 |     sdktrace::BatchSpanProcessorOptions opts;
```

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
+++ b/sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
@@ -10,6 +10,7 @@
 #include <thread>
 
 #include "opentelemetry/sdk/common/circular_buffer.h"
+#include "opentelemetry/sdk/trace/batch_span_processor_options.h"
 #include "opentelemetry/sdk/trace/processor.h"
 #include "opentelemetry/version.h"
 


mfrw:
> Is this a problem of [opentelemetry-cpp-contrib](https://github.com/open-telemetry/opentelemetry-cpp-contrib)?

I am not sure if I should say it's a problem;
But as a downstream user I would expect that all required headers are included IMHO.
In this specific case, the options struct is only forward declared, I am not an expert; so I thought maybe this should fix the problem.
Please excuse my ignorance if this may not be the right solution.

Thanks

lalitb:
> But as a downstream user I would expect that all required headers are included IMHO.

General principal is -
 - Header file should include ONLY what is use ( and forward-declare where-ever possible)
 - Source file should include the header files for anything it uses directly, 
 
So if your downstream source need to initialize `batch_span_processor_options`, it should include `batch_span_processor_options.h`. Another scenario, could be source file which has method which accepts reference to BatchSpanProcessor, and eventually calls `ForceFlush` on it. This source file only needs to include `batch_span_processor.h`, and not `batch_span_processor_options.h`

mfrw:
> This source file only needs to include `batch_span_processor.h`, and not `batch_span_processor_options.h`

Yes, this is what my expectation is as well.

lalitb:
> Yes, this is what my expectation is as well.

I mean this PR is not necessary with above reasoning.

mfrw:
> > Yes, this is what my expectation is as well.
> 
> I mean this PR is not necessary with above reasoning.

Sorry my bad. I did not understand it properly.

PR #2263, opened 2023-08-15
[BUILD] Support to use different cmake package CONFIG of dependencies.

### LINKED ISSUE #2251: Support to use different cmake package CONFIG of dependencies.

Some issues (#705, #1359) was created before, and as talked in #2163, create a separate issue here.

Patch imported target in cmake CONFIG packages to enable users to use prebuilt packages with different CONFIG settings.

owent:
Fixes #2251

## Changes

+ Patch imported targets to support to use different cmake package CONFIG of dependencies.


cmake/tools.cmake
```diff
@@ -117,3 +117,73 @@ function(project_build_tools_get_imported_location OUTPUT_VAR_NAME TARGET_NAME)
         PARENT_SCOPE)
   endif()
 endfunction()
+
+function(project_build_tools_patch_default_imported_config)
+  set(PATCH_VARS
+      IMPORTED_IMPLIB
+      IMPORTED_LIBNAME
```
ThomsonTan, line 124: Is there a way to know the list of patched variables here is comprehensive?

owent: I find them in document here https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-targets .

marcalff (commented):
Thanks for the change.

Please add some comment to describe what feature it actually provides (or bug it fixes),
with an example of when it is useful, and how to use it.

Please assume the reader has no or little knowledge of CMake, and make the example as explicit as possible.

cmake/tools.cmake
```diff
@@ -117,3 +117,83 @@ function(project_build_tools_get_imported_location OUTPUT_VAR_NAME TARGET_NAME)
         PARENT_SCOPE)
   endif()
 endfunction()
+
+#[[
+If we build third party packages with a different CONFIG setting from building
+otel-cpp, cmake may not find a suitable file in imported targets (#705, #1359)
+when linking. But on some platforms, different CONFIG settings can be used when
+ABI these CONFIG settings have the same ABI. For example, on Linux, we can build
```
marcalff, line 125:
```suggestion
these CONFIG settings have the same ABI. For example, on Linux, we can build
```

cmake/tools.cmake
```diff
@@ -117,3 +117,83 @@ function(project_build_tools_get_imported_location OUTPUT_VAR_NAME TARGET_NAME)
...
+gRPC and protobuf with -DCMAKE_BUILD_TYPE=Release, but build otel-cpp with
+-DCMAKE_BUILD_TYPE=Debug and links these libraries together.
+The properties of imported targets can be find here:
```
marcalff, line 128:
```suggestion
The properties of imported targets can be found here:
```

marcalff (approved):
Approved, thanks for the comments.

See two suggestions.

owent:
> Approved, thanks for the comments.
> 
> See two suggestions.

Thanks, the spell errors are fixed.

marcalff:
Added the please-review label.

This PR looks ok to me, but given this affects makefiles with CMake in general, maybe @lalitb @ThomsonTan or @esigo can look at it also.

cmake/tools.cmake
```diff
@@ -117,3 +117,83 @@ function(project_build_tools_get_imported_location OUTPUT_VAR_NAME TARGET_NAME)
...
   endif()
 endfunction()
+
+#[[
+If we build third party packages with a different CONFIG setting from building
+otel-cpp, cmake may not find a suitable file in imported targets (#705, #1359)
+when linking. But on some platforms, different CONFIG settings can be used when
+these CONFIG settings have the same ABI. For example, on Linux, we can build
+gRPC and protobuf with -DCMAKE_BUILD_TYPE=Release, but build otel-cpp with
+-DCMAKE_BUILD_TYPE=Debug and links these libraries together.
```
esigo, line 127: nit:
```suggestion
-DCMAKE_BUILD_TYPE=Debug and link these libraries together.
```

esigo (approved):
LGMT
Thanks for the PR :)

cmake/tools.cmake
owent: Thanks and fixed.

Issue #1940, opened 2023-01-30
Linker error when building otlp_grpc_exporter_test


svrnm:
**Describe your environment** 

- Alpine Linux v3.16 on Aarch64

**Steps to reproduce**
- Commands I am running:
    ```
        sudo apk add cmake grpc-dev re2-dev protobuf-dev c-ares-dev curl-dev nlohmann-json thrift-dev boost-dev
        sudo apk add gtest-dev benchmark-dev
        git clone https://github.com/open-telemetry/opentelemetry-cpp.git
        cd opentelemetry-cpp
        cmake -B build \
                -DCMAKE_INSTALL_PREFIX=/usr \
                -DWITH_PROMETHEUS=OFF \
                -DBUILD_SHARED_LIBS=ON \
                -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
                -DCMAKE_BUILD_TYPE=Release \
                -DBUILD_TESTING=ON \
                -DWITH_EXAMPLES=OFF \
                -DWITH_OTLP=ON \
                -DWITH_OTLP_HTTP=ON \
                -DWITH_JAEGER=ON \
                -DWITH_ZIPKIN=ON
        cmake --build build
    ```

**What is the expected behavior?**
No error

**What is the actual behavior?**
Linker error:

```
[ 83%] Linking CXX executable otlp_grpc_exporter_test
/usr/lib/gcc/aarch64-alpine-linux-musl/11.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: CMakeFiles/otlp_grpc_exporter_test.dir/test/otlp_grpc_exporter_test.cc.o: undefined reference to symbol '_ZN4grpc6Status2OKE'
/usr/lib/gcc/aarch64-alpine-linux-musl/11.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: /usr/lib/libgrpc++.so.1.46: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [exporters/otlp/CMakeFiles/otlp_grpc_exporter_test.dir/build.make:109: exporters/otlp/otlp_grpc_exporter_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:4940: exporters/otlp/CMakeFiles/otlp_grpc_exporter_test.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

**Additional context**
[...]

PR #2268, opened 2023-08-16
added public link of `opentelemetry_proto_grpc` against [shared] gRPC lib

ays7:
only if gRPC library is shared

Fixes #1891

Changes implemented as alternative to PR #1891 do not resolve issues with mac builds against shared gRPC library.
Furthermore, code generated by recent protobuf-grpc compiler contain references into gRPC library, hence   `opentelemetry_proto_grpc` should be linked against gRPC library.

Since it is believed that STATIC gRPC library should only be [privately] linked into a single library, I suppose there are no issues with linking SHARED gRPC library wherever it is needed.


lalitb:
Thanks for the fix. I think the fix is in line with what we discussed here - https://github.com/open-telemetry/opentelemetry-cpp/issues/1940#issuecomment-1649185424

owent (approved):

lalitb (approved):

PR #2269, opened 2023-08-18
Support `pkg-config`

### LINKED ISSUE #1850: Consider supporting `pkg-config`
Supporting [`pkg-config`](https://people.freedesktop.org/~dbn/pkg-config-guide.html) means installing `*.pc` files for the different targets. This is a convenience for developers, which will enable things like [building with `make`](https://www.mrpt.org/tutorials/programming/first-steps/compiling_custom_applications_in_linux_with_a_makefile_and_pkg-config/) (or anything else that consumes `pkg-config` output).

dbolduc:
Fixes #1850 

Generates and installs pkgconfig (`*.pc`) files for the following targets:
- `opentelemetry_api`
- `opentelemetry_logs`
- `opentelemetry_metrics`
- `opentelemetry_resources`
- `opentelemetry_trace`
- `opentelemetry_version`

The API is necessary, because we need to capture cflags, such as `-DWITH_ABSEIL`. I did not think installing an `opentelemetry_sdk.pc` was necessary or helpful.

## Testing

I added a test to validate the installed package configs. This does not guarantee that their contents are 100% correct.

I am new to using the CI. I am yolo'ing the changes to `.github/workflows/ci.yml`. I ran the steps locally, though, and they were successful:
```sh
ci/run_docker.sh

# Then inside docker...
ci/setup_cmake.sh
ci/setup_ci_environment.sh
ci/install_abseil.sh
ci/do_ci.sh cmake.install.test
ci/verify_packages.sh
```

My test installs `opentelemetry-cpp` built `-DWITH_ABSEIL`. The dimensions we care about are: [with/without abseil, with/without gsl] (as far as the install logic for `opentelemetry_api` goes). But I did not want to write all these tests, so I just checked the abseil case.

### Testing while developing

I did some more rigorous testing while developing. The changes are in [this commit](https://github.com/dbolduc/opentelemetry-cpp/commit/17cef61ce9ccfeac960144d5d5415442581f2cac).

Basically it has an executable that uses OTel types. This executable is built using a Makefile, through the use of `pkg-config`. With it, I was able to confirm that the installed libraries link, the include directories work, the compile definitions work, etc.

I tested this with/without abseil for all of the added targets listed above.


dbolduc:
Sorry about iterating in the CI. It looks like the "Bazel tsan config" build is failing across the board, so I think this PR is ready for review.

owent:
Do you think we can implement a function replace `add_library` in otlp-cpp and call `opentelemetry_add_pkgconfig` in it. Then we can add all cmake targets automately?

dbolduc:
> Can we replace the `add_library` function in otlp-cpp with a new function which can call `add_library` and `opentelemetry_add_pkgconfig`.And then it could automatically add all cmake targets and ensure that the package name in pkgconfig matches the cmake package name?

`opentelemetry_add_pkgconfig` inspects properties of the library target: https://github.com/open-telemetry/opentelemetry-cpp/blob/b275af1d271734019054671fa6624fa3d9fe3a0e/cmake/pkgconfig.cmake#L35

... so I think we are already ensuring that the package name in pkgconfig matches the cmake package name.

This also means that we can only call `opentelemetry_add_pkgconfig` after all of the `target_include_directories(target, ...)` and `target_compile_definitions(target, ...)` have been set. I think this means we can not combine the `add_library` and `opentelemetry_add_pkgconfig` functions.

> otlp-cpp

Sorry if I am misunderstanding your question. I am not so sure what this is. I did not touch any of the OTLP components. We can always address them in a follow up PR.

owent:
> This also means that we can only call opentelemetry_add_pkgconfig after all of the target_include_directories(target, ...) and target_compile_definitions(target, ...) have been set. I think this means we can not combine the add_library and opentelemetry_add_pkgconfig functions.

Thanks, I think it's fine.

> Sorry if I am misunderstanding your question. I am not so sure what this is. I did not touch any of the OTLP components. We can always address them in a follow up PR.
Sorry for my mistake. I mean `otel-cpp`, not OTLP.

owent (approved):
LGTM and thanks.

dbolduc:
Any chance this feature makes it into the v1.11.0 release?

lalitb (approved):
Nicely done. Thanks for contributing.

@ThomsonTan  - Do you think it can be included in the v1.11.0?

dbolduc:
> As this is a new feature, I think it will be good to stay in main for a release cycle.

Ack. Thanks for merging.

PR #2288, opened 2023-09-01
Add _total suffix to prometheus counters

### LINKED ISSUE #2287: Prometheus exporter does not add _total suffixes to counters

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#sums: 

> If the metric name for monotonic Sum metric points does not end in a suffix of _total a suffix of _total MUST be added by default, otherwise the name MUST remain unchanged. Exporters SHOULD provide a configuration option to disable the addition of _total suffixes.

Based on the unit test here: 

https://github.com/open-telemetry/opentelemetry-cpp/blob/f52ec6caded074437658643566bed0d83152a78a/exporters/prometheus/test/exporter_utils_test.cc#L107-L119

The counter metric does not have a `_total` suffix appended.

dashpole:
Fixes #2287

## Changes

Append _total suffix to Prometheus counters.  This is done by:

* Determine type of a metric based on the first data point.  This assumes all data points within a metric translated have the same type.
* Trim _total suffixes from counter metric names before appending the unit (_total must come after the unit suffix).
* Add _total suffix to counters after the unit suffix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [EXPORTER] Add _total suffixes to Prometheus counters
+  [#2288](https://github.com/open-telemetry/opentelemetry-cpp/pull/2288)
+
 ## [1.11.0] 2023-08-21
 
 * [BUILD] Fix more cases for symbol name for 32-bit win32 DLL build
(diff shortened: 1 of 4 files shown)


dashpole:
Duplicate of https://github.com/open-telemetry/opentelemetry-cpp/pull/2213

PR #2293, opened 2023-09-01
prometheus exporter: Add otel_scope_name and otel_scope_version labels

dashpole:
Part of https://github.com/open-telemetry/opentelemetry-cpp/issues/1841

## Changes

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1:

> Prometheus exporters MUST add the scope name as the otel_scope_name label and the scope version as the otel_scope_version label on all metric points by default, based on the scope the original data point was nested in.

This adds the otel_scope_name and otel_scope_version labels to all metrics, if they are non-empty.

### DIFF
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -164,14 +164,16 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
             auto last_value_point_data =
                 nostd::get<sdk::metrics::LastValuePointData>(point_data_attr.point_data);
             std::vector<metric_sdk::ValueType> values{last_value_point_data.value_};
-            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type, &metric_family);
+            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type,
+                    &metric_family);
           }
           else if (nostd::holds_alternative<sdk::metrics::SumPointData>(point_data_attr.point_data))
           {
             auto sum_point_data =
                 nostd::get<sdk::metrics::SumPointData>(point_data_attr.point_data);
             std::vector<metric_sdk::ValueType> values{sum_point_data.value_};
-            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type, &metric_family);
+            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type,
+                    &metric_family);
           }
           else
           {
@@ -187,7 +189,8 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
             auto sum_point_data =
                 nostd::get<sdk::metrics::SumPointData>(point_data_attr.point_data);
             std::vector<metric_sdk::ValueType> values{sum_point_data.value_};
-            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type, &metric_family);
+            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type,
+                    &metric_family);
           }
           else
           {
@@ -296,7 +299,8 @@ void PrometheusExporterUtils::SetData(
 /**
  * Set labels to metric data
  */
-void PrometheusExporterUtils::SetMetricBasic(prometheus_client::ClientMetric &metric,
+void PrometheusExporterUtils::SetMetricBasic(
+    prometheus_client::ClientMetric &metric,
     const metric_sdk::PointAttributes &labels,
     const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope)
 {
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -22,6 +22,10 @@ namespace metrics
 {
 namespace
 {
+
+static constexpr const char *kScopeNameKey    = "otel_scope_name";
+static constexpr const char *kScopeVersionKey = "otel_scope_version";
+
 /**
  * Sanitize the given metric name by replacing invalid characters with _,
  * ensuring that multiple consecutive _ characters are collapsed to a single _.
@@ -150,7 +154,7 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
             sum = nostd::get<int64_t>(histogram_point_data.sum_);
           }
           SetData(std::vector<double>{sum, (double)histogram_point_data.count_}, boundaries, counts,
-                  point_data_attr.attributes, &metric_family);
+                  point_data_attr.attributes, instrumentation_info.scope_, &metric_family);
         }
         else if (type == prometheus_client::MetricType::Gauge)
         {
@@ -160,14 +164,14 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
             auto last_value_point_data =
                 nostd::get<sdk::metrics::LastValuePointData>(point_data_attr.point_data);
             std::vector<metric_sdk::ValueType> values{last_value_point_data.value_};
-            SetData(values, point_data_attr.attributes, type, &metric_family);
+            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type, &metric_family);
           }
           else if (nostd::holds_alternative<sdk::metrics::SumPointData>(point_data_attr.point_data))
           {
             auto sum_point_data =
                 nostd::get<sdk::metrics::SumPointData>(point_data_attr.point_data);
             std::vector<metric_sdk::ValueType> values{sum_point_data.value_};
-            SetData(values, point_data_attr.attributes, type, &metric_family);
+            SetData(values, point_data_attr.attributes, instrumentation_info.scope_, type, &metric_family);
           }
           else
           {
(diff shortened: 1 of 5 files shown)


dashpole:
```
[ RUN      ] PrometheusExporterUtils.TranslateToPrometheusIntegerCounter
unknown file: Failure
C++ exception with description "basic_string::_M_create" thrown in the test body.
```

Any pointers on what this might be?

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -242,40 +247,57 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
     metric.label.resize(labels.size());
-    size_t i = 0;
     for (auto const &label : labels)
     {
       auto sanitized          = SanitizeNames(label.first);
       metric.label[i].name    = sanitized;
       metric.label[i++].value = AttributeValueToString(label.second);
     }
   }
+  auto scope_name = scope->GetName();
```
dbolduc, line 287: 1. We have to make sure scope is not null before using it. I'd add:

```cc
if (!scope) return;
```

---

The problem you are seeing in the tests is that the `ScopeMetrics::scope_` pointer is uninitialized.... and I am sure it is assumed to be `nullptr` somewhere. Looks like #2301 will fix this. (and it will also give you some merge conflicts): https://github.com/open-telemetry/opentelemetry-cpp/pull/2301/files#diff-e774f313b47bc09b00e03c7d294a3f09fa7c04205b7bb24734f1722030363613

dbolduc (commented):
**EDIT:** I was wrong about this. There is an implicit conversion to std::string. I am not sure what will happen in the case where OTel uses the standard library though.

---

Also, maybe related, this part of the implementation is error prone:

https://github.com/open-telemetry/opentelemetry-cpp/blob/cdbc90abf3b7170f898439fd34126469c42686c6/sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h#L140

This assumes that the `string_view`s are null-terminated, which might not be the case. I can send a PR to fix it.

exporters/prometheus/src/exporter_utils.cc
dashpole: I added the suggested line.

dbolduc (commented):
If you want to see the CI pass, you could steal the changes from #2301

https://github.com/open-telemetry/opentelemetry-cpp/blob/049f7e8a19bf0a0ff3e6a748f6556f2fac7bb180/sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h#L34

Basically make this
```cc
 const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope_ = nullptr;
```

exporters/prometheus/src/exporter_utils.cc
dashpole: Even after adding the check for !scope, I am still getting `C++ exception with description "basic_string::_M_create" thrown in the test body.` on this line.  Any suggestions?

dbolduc: I was able to check out your PR and run the tests successfully a week ago with these edits. It is too bad that I deleted the branch.

When I try now, it does not work. I do not know if that is from rebasing or changes to this PR.

The `scope_` is still nonsense. It is a raw pointer. Something has to own the object. My guess is that it is being deleted before we try to use it.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -242,40 +251,63 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
+    const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope)
 {
   metric.timestamp_ms = time.count() / 1000000;
 
-  // auto label_pairs = ParseLabel(labels);
   if (!labels.empty())
   {
-    metric.label.resize(labels.size());
-    size_t i = 0;
+    metric.label.reserve(labels.size() + 2);
```
dbolduc, line 282: This should go outside the labels check. In fact, we should not even check if labels is empty. It would just skip the loop.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -242,40 +251,63 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
     for (auto const &label : labels)
     {
-      auto sanitized          = SanitizeNames(label.first);
-      metric.label[i].name    = sanitized;
-      metric.label[i++].value = AttributeValueToString(label.second);
+      prometheus_client::ClientMetric::Label sanitizedLabel;
+      sanitizedLabel.name  = SanitizeNames(label.first);
+      sanitizedLabel.value = AttributeValueToString(label.second);
+      metric.label.push_back(std::move(sanitizedLabel));
```
dbolduc, line 288: Here and below consider:

```cc
metric.label.emplace_back();
metric.label.back().name = ...;
metric.label.back().value = ...;
```

Although I think I saw a helper for this in #2301

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -242,40 +251,63 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
   }
+  if (!scope)
+  {
+    return;
+  }
+  auto scope_name = scope->GetName();
+  if (!scope_name.empty())
+  {
+    prometheus_client::ClientMetric::Label label;
+    label.name  = std::move(kScopeNameKey);
```
dbolduc, line 299: do not `std::move()` the `constexpr`s

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -242,40 +251,63 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
+    label.value = std::move(scope_name);
+    metric.label.push_back(std::move(label));
+  }
+  auto scope_version = scope->GetVersion();
+  if (!scope_version.empty())
+  {
+    prometheus_client::ClientMetric::Label label;
+    label.name  = std::move(kScopeNameKey);
```
dbolduc, line 307: s/Name/Version/

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
```diff
@@ -31,8 +31,21 @@ namespace metrics
...
+  inline ScopeMetrics() {}
+  inline ScopeMetrics(const ScopeMetrics &) = default;
+  inline ScopeMetrics(ScopeMetrics &&)      = default;
+
+  inline ScopeMetrics &operator=(const ScopeMetrics &) = default;
+
+  inline ScopeMetrics &operator=(ScopeMetrics &&) = default;
 };
 
 struct ResourceMetrics
```
dbolduc, line 51: It didn't help me, but you might want to add the changes for `resource_` as well.

dbolduc (commented):
I tried for an hour to get this to work. I am giving up. Here are some code nits for you:

punya:
#2322 merges cleanly with this PR and fixes the failing tests :tada:

exporters/prometheus/src/exporter_utils.cc
dashpole: done

exporters/prometheus/src/exporter_utils.cc
dashpole: done

exporters/prometheus/src/exporter_utils.cc
dashpole: fixed by https://github.com/open-telemetry/opentelemetry-cpp/pull/2322

exporters/prometheus/src/exporter_utils.cc
dashpole: done

exporters/prometheus/src/exporter_utils.cc
dashpole: done

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
```diff
@@ -31,7 +31,7 @@ namespace metrics
  */
 struct ScopeMetrics
 {
-  const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope_;
+  const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope_ = nullptr;
```
punya, line 34: I think this change is the reason for the CI failures. Now that this field has an initial value, `ScopeMetrics` is no longer considered an "aggregate type", which [breaks](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/6275544529/job/17043256873?pr=2293#step:6:726) the way it's used in the OTLP tests:

```
ostream_metric_test.cc:58:70: error: no matching function for call to ‘std::vector<opentelemetry::v1::sdk::metrics::ScopeMetrics>::vector(<brace-enclosed initializer list>)’
       {scope.get(), std::vector<metric_sdk::MetricData>{metric_data}}};D
```

dashpole: I removed this, and that fixed the tests.

marcalff:
Please merge with a recent main, to resolve the conflicts.

dashpole:
@marcalff done

lalitb (approved):

lalitb:
Depending on which PR is merged first - #2293 or #2301, the other PR will need to address merge conflicts. One argument could be to merge smaller and less complex PR (this?) first, so easy to resolve conflicts in other one.

dashpole:
I'm happy to wait and rebase my PR.

marcalff (approved):
LGTM, thanks for the fix.

PR #2301, opened 2023-09-08
Export resource for prometheus

### LINKED ISSUE #1842: Prometheus exporter emits target_info for Resource
From the spec for [handling resource in prometheus exporters](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1):

> In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to a single [target_info metric](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems); otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metric families. The target_info metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one target_info metric exposed on an SDK Prometheus endpoint.

The Prometheus exporter [currently](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/prometheus/src/exporter_utils.cc#L42) doesn't use resource attributes for anything.

owent:
Fixes #1842 

## Changes

+ Prometheus exporter emits resource attributes
+ Fix incorrect address point in unit test of prometheus utils.

Follow <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1>

### DIFF
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -63,15 +63,6 @@ private:
    */
   static bool ShouldIgnoreResourceAttribute(const std::string &name);
 
-  /**
-   * Some attributes should be ignored when converting metric attributes to
-   * prometheus labels.
-   *
-   * @param name attribute name
-   * @return true if the attribute should be ignored, false otherwise.
-   */
-  static bool ShouldIgnoreMetricAttribute(const std::string &name);
-
   static opentelemetry::sdk::metrics::AggregationType getAggregationType(
       const opentelemetry::sdk::metrics::PointType &point_type);
 
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -63,6 +63,15 @@ private:
    */
   static bool ShouldIgnoreResourceAttribute(const std::string &name);
 
+  /**
+   * Some attributes should be ignored when converting metric attributes to
+   * prometheus labels.
+   *
+   * @param name attribute name
+   * @return true if the attribute should be ignored, false otherwise.
+   */
+  static bool ShouldIgnoreMetricAttribute(const std::string &name);
+
   static opentelemetry::sdk::metrics::AggregationType getAggregationType(
       const opentelemetry::sdk::metrics::PointType &point_type);
 
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -41,6 +41,24 @@ private:
    */
   static std::string SanitizeNames(std::string name);
 
+  /**
+   * Some attributes should be ignored when converting resource attributes to
+   * prometheus labels.
+   *
+   * @param name resource attribnuet name
+   * @return true if the attribute should be ignored, false otherwise.
+   */
+  static bool ShouldIgnoreResourceAttribute(const std::string &name);
+
+  /**
+   * Some attributes should be renamed when converting resource attributes to
+   * prometheus labels.
+   *
+   * @param name resource attribnuet name
+   * @return const std::string&
+   */
+  static const std::string &GetPrometheusAttributeName(const std::string &name);
+
   static opentelemetry::sdk::metrics::AggregationType getAggregationType(
       const opentelemetry::sdk::metrics::PointType &point_type);
 
@@ -59,7 +77,8 @@ private:
                       const opentelemetry::sdk::metrics::PointAttributes &labels,
                       ::prometheus::MetricType type,
                       std::chrono::nanoseconds time,
-                      ::prometheus::MetricFamily *metric_family);
+                      ::prometheus::MetricFamily *metric_family,
+                      const opentelemetry::sdk::resource::Resource *resource);
 
   /**
    * Set metric data for:
@@ -71,14 +90,16 @@ private:
                       const std::vector<uint64_t> &counts,
                       const opentelemetry::sdk::metrics::PointAttributes &labels,
                       std::chrono::nanoseconds time,
-                      ::prometheus::MetricFamily *metric_family);
+                      ::prometheus::MetricFamily *metric_family,
+                      const opentelemetry::sdk::resource::Resource *resource);
 
   /**
    * Set time and labels to metric data
    */
   static void SetMetricBasic(::prometheus::ClientMetric &metric,
                              std::chrono::nanoseconds time,
-                             const opentelemetry::sdk::metrics::PointAttributes &labels);
+                             const opentelemetry::sdk::metrics::PointAttributes &labels,
+                             const opentelemetry::sdk::resource::Resource *resource);
 
   /**
    * Convert attribute value to string
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -45,15 +45,6 @@ private:
                                  std::string value,
                                  std::vector<::prometheus::ClientMetric::Label> *labels);
 
-  /**
-   * Some attributes should be ignored when converting resource attributes to
-   * prometheus labels.
-   *
-   * @param name resource attribute name
-   * @return true if the attribute should be ignored, false otherwise.
-   */
-  static bool ShouldIgnoreResourceAttribute(const std::string &name);
-
   static opentelemetry::sdk::metrics::AggregationType getAggregationType(
       const opentelemetry::sdk::metrics::PointType &point_type);
 
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -58,6 +58,7 @@ private:
    * Add a target_info metric to collect resource attributes
    */
   static void SetTarget(const sdk::metrics::ResourceMetrics &data,
+                        const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope,
                         std::vector<::prometheus::MetricFamily> *output);
 
   /**
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -26,10 +26,12 @@ public:
    * to Prometheus metrics data collection
    *
    * @param records a collection of metrics in OpenTelemetry
+   * @param populate_target_info whether to populate target_info
    * @return a collection of translated metrics that is acceptable by Prometheus
    */
   static std::vector<::prometheus::MetricFamily> TranslateToPrometheus(
-      const sdk::metrics::ResourceMetrics &data);
+      const sdk::metrics::ResourceMetrics &data,
+      bool populate_target_info = true);
 
 private:
   /**
@@ -47,8 +49,6 @@ private:
    * @param name label name
    * @param value label value
    * @param labels target labels
-   *
-   * @return true if the attribute should be ignored, false otherwise.
    */
   static void AddPrometheusLabel(std::string name,
                                  std::string value,
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -63,15 +63,6 @@ private:
    */
   static bool ShouldIgnoreResourceAttribute(const std::string &name);
 
-  /**
-   * Some attributes should be renamed when converting resource attributes to
-   * prometheus labels.
-   *
-   * @param name resource attribute name
-   * @return const std::string&
-   */
-  static const std::string &GetPrometheusAttributeName(const std::string &name);
-
   static opentelemetry::sdk::metrics::AggregationType getAggregationType(
       const opentelemetry::sdk::metrics::PointType &point_type);
 
--- a/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
+++ b/exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
@@ -42,10 +42,23 @@ private:
   static std::string SanitizeNames(std::string name);
 
   /**
+   * Append key-value pair to prometheus labels.
+   *
+   * @param name label name
+   * @param value label value
+   * @param labels target labels
+   *
+   * @return true if the attribute should be ignored, false otherwise.
+   */
+  static void AddPrometheusLabel(std::string name,
+                                 std::string value,
+                                 std::vector<::prometheus::ClientMetric::Label> *labels);
+
+  /**
    * Some attributes should be ignored when converting resource attributes to
    * prometheus labels.
    *
-   * @param name resource attribnuet name
+   * @param name resource attribute name
    * @return true if the attribute should be ignored, false otherwise.
    */
   static bool ShouldIgnoreResourceAttribute(const std::string &name);
@@ -54,7 +67,7 @@ private:
    * Some attributes should be renamed when converting resource attributes to
    * prometheus labels.
    *
-   * @param name resource attribnuet name
+   * @param name resource attribute name
    * @return const std::string&
    */
   static const std::string &GetPrometheusAttributeName(const std::string &name);
@@ -67,6 +80,12 @@ private:
    */
   static ::prometheus::MetricType TranslateType(opentelemetry::sdk::metrics::AggregationType kind,
                                                 bool is_monotonic = true);
+
+  /**
+   * Add a target_info metric to collect resource attributes
+   */
+  static void SetTarget(const sdk::metrics::ResourceMetrics &data,
+                        std::vector<::prometheus::MetricFamily> *output);
 
   /**
    * Set metric data for:
--- a/exporters/prometheus/src/collector.cc
+++ b/exporters/prometheus/src/collector.cc
@@ -17,7 +17,10 @@ namespace metrics
  * This constructor initializes the collection for metrics to export
  * in this class with default capacity
  */
-PrometheusCollector::PrometheusCollector(sdk::metrics::MetricReader *reader) : reader_(reader) {}
+PrometheusCollector::PrometheusCollector(sdk::metrics::MetricReader *reader,
+                                         bool populate_target_info)
+    : reader_(reader), populate_target_info_(populate_target_info)
+{}
 
 /**
  * Collects all metrics data from metricsToCollect collection.
@@ -36,7 +39,17 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   collection_lock_.lock();
 
   std::vector<prometheus_client::MetricFamily> result;
-  reader_->Collect([&result](sdk::metrics::ResourceMetrics &metric_data) {
+
+  // If populate_target_info_ is true, populate target_info once.
+  bool populate_target_info = false;
+  if (populate_target_info_)
+  {
+    populate_target_info = true;
+
+    const_cast<PrometheusCollector *>(this)->populate_target_info_ = false;
+  }
+
+  reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
     auto prometheus_metric_data = PrometheusExporterUtils::TranslateToPrometheus(metric_data);
     for (auto &data : prometheus_metric_data)
       result.emplace_back(data);
--- a/exporters/prometheus/src/collector.cc
+++ b/exporters/prometheus/src/collector.cc
@@ -40,10 +40,9 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
 
   std::vector<prometheus_client::MetricFamily> result;
 
-  bool populate_target_info = populate_target_info_;
-  reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
+  reader_->Collect([&result, this](sdk::metrics::ResourceMetrics &metric_data) {
     auto prometheus_metric_data =
-        PrometheusExporterUtils::TranslateToPrometheus(metric_data, populate_target_info);
+        PrometheusExporterUtils::TranslateToPrometheus(metric_data, this->populate_target_info_);
     for (auto &data : prometheus_metric_data)
       result.emplace_back(data);
     return true;
--- a/exporters/prometheus/src/collector.cc
+++ b/exporters/prometheus/src/collector.cc
@@ -39,16 +39,8 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   collection_lock_.lock();
 
   std::vector<prometheus_client::MetricFamily> result;
-
-  // If populate_target_info_ is true, populate target_info once.
-  bool populate_target_info = false;
-  if (populate_target_info_)
-  {
-    populate_target_info = true;
-
-    const_cast<PrometheusCollector *>(this)->populate_target_info_ = false;
-  }
-
+  
+  bool populate_target_info = populate_target_info_;
   reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
     auto prometheus_metric_data =
         PrometheusExporterUtils::TranslateToPrometheus(metric_data, populate_target_info);
--- a/exporters/prometheus/src/collector.cc
+++ b/exporters/prometheus/src/collector.cc
@@ -39,7 +39,7 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   collection_lock_.lock();
 
   std::vector<prometheus_client::MetricFamily> result;
-  
+
   bool populate_target_info = populate_target_info_;
   reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
     auto prometheus_metric_data =
--- a/exporters/prometheus/src/collector.cc
+++ b/exporters/prometheus/src/collector.cc
@@ -50,7 +50,8 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   }
 
   reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
-    auto prometheus_metric_data = PrometheusExporterUtils::TranslateToPrometheus(metric_data);
+    auto prometheus_metric_data =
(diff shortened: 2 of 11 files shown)


sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
```diff
@@ -31,14 +32,40 @@ namespace metrics
...
+      : scope_{std::move(scope)}, metric_data_{std::move(metric)}
+  {}
+
+  inline ScopeMetrics()                     = default;
+  inline ScopeMetrics(const ScopeMetrics &) = default;
+  inline ScopeMetrics(ScopeMetrics &&)      = default;
+
+  inline ScopeMetrics &operator=(const ScopeMetrics &) = default;
+
+  inline ScopeMetrics &operator=(ScopeMetrics &&) = default;
```
dbolduc, line 49: just curious, why is this stuff necessary?

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
```diff
@@ -31,14 +32,40 @@ namespace metrics
  */
 struct ScopeMetrics
 {
-  const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope_;
+  const opentelemetry::sdk::instrumentationscope::InstrumentationScope *scope_ = nullptr;
```
dbolduc, line 35: Consider sending separate PRs for the SDK fix and the Prometheus exporter changes.

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
owent: Some old versions of GCC have BUGs with default member initializer.Please see the details here: https://github.com/grpc/grpc/issues/28147

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
owent: Prometheus exporter do not use `resource` in `ResourceMetrics` before and I'm not sure if there will be any problems when prometheus exporter use it without initialize it to a valid value.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -260,20 +294,79 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
+    {
+      if (ShouldIgnoreResourceAttribute(label.first))
+      {
+        continue;
+      }
+
+      prometheus_client::ClientMetric::Label prometheus_label;
+      prometheus_label.name  = SanitizeNames(GetPrometheusAttributeName(label.first));
+      prometheus_label.value = AttributeValueToString(label.second);
+      metric.label.emplace_back(std::move(prometheus_label));
```
lalitb, line 357: Lots of repetitive code here. Can we move it to function like:
```cpp
void AddPrometheusLabel(
    const std::string &value,
    std::vector<prometheus_client::ClientMetric::Label> &labels)
{
  prometheus_client::ClientMetric::Label prometheus_label;
  prometheus_label.name = "job";
  prometheus_label.value = value;
  labels.emplace_back(std::move(prometheus_label));
}
```

lalitb:
Going through this link: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1, it states:

``
In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to a single [target_info metric](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems); otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metric families. The target_info metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one target_info metric exposed on an SDK Prometheus endpoint.
``

It says that resource attributes should be converted into a separate 'target_info` metrics, or else dropped. I don't see we are doing this here.

sdk/include/opentelemetry/sdk/metrics/export/metric_producer.h
dbolduc: Ack, thanks.

FTR, I am not going to review this PR. I stumbled upon the PR randomly and left questions, because I was curious. Thanks for answering them!

owent:
> Going through this link: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1, it states:
> 
> `In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to a single [target_info metric](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems); otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metric families. The target_info metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one target_info metric exposed on an SDK Prometheus endpoint.`
> 
> It states that resource attributes should be converted into a separate `target_info` metrics, or else dropped. I don't see we are doing it here.

Sorry, I had some misunderstanding before. Could you please reviews the codes again and please let me know if there is any problem.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -220,6 +266,59 @@ prometheus_client::MetricType PrometheusExporterUtils::TranslateType(
...
+  metric_family.type = prometheus_client::MetricType::Info;
+  metric_family.metric.emplace_back();
+
+  prometheus_client::ClientMetric &metric = metric_family.metric.back();
+  metric.info.value                       = 1.0;
+
+  std::chrono::nanoseconds time;
+  for (const auto &instrumentation_info : data.scope_metric_data_)
+  {
+    for (const auto &metric_data : instrumentation_info.metric_data_)
```
lalitb, line 297: would this be more effective then to have nested loops:
```cpp
if (!data.scope_metric_data_.empty() && !data.scope_metric_data_.begin()->metric_data_.empty()) {
  // .. get time.
}
```

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -220,6 +266,59 @@ prometheus_client::MetricType PrometheusExporterUtils::TranslateType(
...
+  }
+
+  prometheus_client::MetricFamily metric_family;
+  metric_family.name = "target_info";
+  metric_family.help = "Target metadata";
+  metric_family.type = prometheus_client::MetricType::Info;
+  metric_family.metric.emplace_back();
+
+  prometheus_client::ClientMetric &metric = metric_family.metric.back();
+  metric.info.value                       = 1.0;
```
lalitb, line 292: where is this value `1.0` coming from ?

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -260,22 +361,118 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
+      }
+      if (service_instance_id_it != resource->GetAttributes().end())
+      {
+        has_instance_label = true;
+        AddPrometheusLabel(kPrometheusInstance,
+                           AttributeValueToString(service_instance_id_it->second), &metric.label);
+      }
+    }
+  }
+
```
lalitb, line 470: Can you please revisit this function - looks like we are somewhat overengineering stuff here. The `job` and `instance` labels are supposed to be handled in the Prometheus exporter implemented for Collector, and it is already done here - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go.

exporters/prometheus/src/exporter_utils.cc
owent: > If info-typed metric families are not yet supported by the language Prometheus client library, a gauge-typed metric family named target_info with a constant value of 1 MUST be used instead.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1

> An OpenMetrics Info can be thought of as a special-case of the OpenMetrics Gauge which has a value of 1, and whose labels generally stays constant over the life of the process. It is converted to a Non-Monotonic Sum, rather than a Gauge, because the value of 1 is intended to be viewed as a count, which should be summed together when aggregating away labels.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#info

These documents do not describe which value should be used as info-typed metric.But I think we can use the same value to keep compatibility with the situations above.

BTW: The example in https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems also set value to 1.

exporters/prometheus/include/opentelemetry/exporters/prometheus/exporter_utils.h
```diff
@@ -41,6 +41,28 @@ class PrometheusExporterUtils
...
   static std::string SanitizeNames(std::string name);
 
+  /**
+   * Append key-value pair to prometheus labels.
+   *
+   * @param name label name
+   * @param value label value
+   * @param labels target labels
+   *
+   * @return true if the attribute should be ignored, false otherwise.
```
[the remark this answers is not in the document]
owent: Thanks and fixed.

exporters/prometheus/src/exporter_utils.cc
owent: I found no description about how to export labels when users already set `job` and `instance` as attributes. In my opinion, we can keep compatibility with old codes which already set `job` and `instance` as attributes of metrics data or resource.
Or do you think we can keep one or two versions to let users use old ways?

lalitb: > I found no description about how to export labels when users already set job and instance as attributes.

Yes, as the specs don't specify any mapping for `job` and `instance` set within attributes/resources, we shouldn't be adding them. And let Prometheus backend add this data itself, as it has the endpoint URL it is scrapping to, as specified here - https://prometheus.io/docs/concepts/jobs_instances/. Also, needn't ignore any of the resources/attributes while copying them to target_info/labels.

The specs define a mapping for the Collector's Prometheus exporter, as this exporter could be fetching data from multiple targets, so each metric data could have a different `job` and `instance` value depending on the target.

>  Or do you think we can keep one or two versions to let users use old ways?

We can keep the existing version i.e., sanitize and copy all the metrics attributes to Prometheus labels? No need to ignore resources or attributes during conversions.

I am talking on the basis of reading specs - do you see any issue with that from an end-user perspective?

owent-api: I mean if a metric has resource attribute `service.instance.id` and attribute `instance`. If we do not ignore one, there will be two `instance` in prometheus metric.The implementation in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go use `map[string]string` to set labels from metric attributes and then from resource attributes, which means if when both `service.instance.id` and `instance` are set, it will use `service.instance.id`. But in prometheus-cpp, it use `std::vector<Label>`, not a `map, if we do not ignore one, there will be two `instance` labels.
I just update the code to keep the same behaviour as it in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go . The `job` and `[service.namespace/]service.name` have the same problem.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -220,6 +271,51 @@ prometheus_client::MetricType PrometheusExporterUtils::TranslateType(
...
+  {
+    return;
+  }
+  if ((*data.scope_metric_data_.begin()).metric_data_.empty())
+  {
+    return;
+  }
+
+  prometheus_client::MetricFamily metric_family;
+  metric_family.name = "target_info";
```
dashpole, line 291: this should just be `target`.  See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems. I'll update the prometheus specification to fix this.
```suggestion
  metric_family.name = "target";
```

https://github.com/open-telemetry/opentelemetry-specification/issues/3700

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -167,6 +193,31 @@ std::string PrometheusExporterUtils::SanitizeNames(std::string name)
...
+{
+  prometheus_client::ClientMetric::Label prometheus_label;
+  prometheus_label.name  = std::move(name);
+  prometheus_label.value = std::move(value);
+  labels->emplace_back(std::move(prometheus_label));
+}
+
+bool PrometheusExporterUtils::ShouldIgnoreResourceAttribute(const std::string &name)
+{
+  static std::unordered_set<std::string> ignores{
```
dashpole, line 209: Why are you ignoring these resource attributes?

Edit: ah, I see your reasoning below.  I think you can remove this filtering.

exporters/prometheus/src/exporter_utils.cc
dashpole: SDK exporters (not collector) should simply map all resource attributes to labels on the target_info metric with no filtering or translation.  I realize there can possibly be collisions if someone is using prometheus exporters in the collector, but that is best solved by receiver config (honor_labels) or processor (e.g. resource) configuration.

exporters/prometheus/src/exporter_utils.cc
owent-api: According to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes , some attributes MUST NOT be added as metric attributes.
And ignoring these attributes is also used to avoid duplicate `job` and `instance`.

dashpole: This is the relevant section for SDK exporters:

> In SDK Prometheus (pull) exporters, resource attributes SHOULD be converted to a single [target_info metric](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#supporting-target-metadata-in-both-push-based-and-pull-based-systems); otherwise, they MUST be dropped, and MUST NOT be attached as labels to other metric families. The target_info metric MUST be an info-typed metric whose labels MUST include the resource attributes, and MUST NOT include any other labels. There MUST be at most one target_info metric exposed on an SDK Prometheus endpoint.

Where do you see that some attributes MUST NOT be added as metric attributes?

exporters/prometheus/src/collector.cc
```diff
@@ -36,8 +39,19 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
...
-  reader_->Collect([&result](sdk::metrics::ResourceMetrics &metric_data) {
-    auto prometheus_metric_data = PrometheusExporterUtils::TranslateToPrometheus(metric_data);
+
+  // If populate_target_info_ is true, populate target_info once.
+  bool populate_target_info = false;
+  if (populate_target_info_)
+  {
+    populate_target_info = true;
+
+    const_cast<PrometheusCollector *>(this)->populate_target_info_ = false;
```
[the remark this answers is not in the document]
owent-api: > There MUST be at most one target_info metric exposed on an SDK Prometheus endpoint.
According to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes-1 , the target info metric should not change and it must be exposed only once.Do I missed anything?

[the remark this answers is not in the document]
dashpole: This means once per-scrape/collection.  I.e. there should never be a target_info with two streams:

```
# TYPE target info
# HELP target Target metadata
target_info{env="prod",hostname="myhost",datacenter="sdc",region="europe",owner="frontend"} 1
target_info{env="staging",hostname="myhost",datacenter="sdc",region="europe",owner="frontend"} 1
```

You should have the target info metric on every scrape.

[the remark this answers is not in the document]
lalitb: I guess this was modified for the comment [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/2301#discussion_r1329652826).  Probably I was not very clear - the ask was to compute and cache the `target_info` metric internally during first scrape(as resources are immutable), and add it during every scrape/collection.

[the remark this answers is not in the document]
dashpole: +1.  Caching is great.

exporters/prometheus/src/exporter_utils.cc
owent-api: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go#L302
This collector also remove these attributes. Should we keep it the same behaviour as the collector?

dashpole: We should not.  The collector and SDK exporters intentionally have different behavior.  That behavior (converting service.name, service.namespace, and service.instance.id to job and instance) is undoing what is done in the prometheus receiver.  Since this (and other SDK) exporter does not receive metrics from other prometheus sources, it shouldn't include that logic.

exporters/prometheus/src/exporter_utils.cc
owent-api: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#resource-attributes

<blockquote>
<p dir="auto">The following attributes SHOULD be associated with scraped metrics as resource
attributes, and MUST NOT be added as metric attributes:</p>


OTLP Resource Attribute | Description
-- | --
server.address | The <host> portion of the target's URL that was scraped
server.port | The <port> portion of the target's URL that was scraped
url.scheme | http or https


</blockquote>

Thanks, I may have some misunderstanding before. I remove the ignores in metric attributes, the rest ignores in resource attributes are jus used to avoid duplicate `job` and `instance`. And kee the same behaviour as https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go#L302 .

dashpole: Ah, that is in the Prometheus -> OTLP section.  This exporter is OTLP -> Prometheus.

owent: I remove the ignores in metric attributes, the rest ignores in resource attributes are just used to avoid duplicate `job` and `instance` labels. The behaviour is just like [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go#L302) now .

dashpole: Please do not ignore any labels.  The otel-collector-contrib exporter is _not_ a good comparison point, as it follows a different specification.  I would recommend comparing to the [otel-go](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/prometheus/exporter.go#L165) implementation, the [otel-java](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/prometheus/src/main/java/io/opentelemetry/exporter/prometheus/Serializer.java#L246) implementation, or the [otel-rust](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/src/lib.rs#L300) implementation.  Those languages are all fully complaint with the prometheus specification for SDKs.

owent: This PR do not ignore any label now, The ignore list is used to avoid duplicated adding `instance` and `job` label.[otel-go](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/prometheus/exporter.go#L294) also use `map[string][]string` to merge keys and avoid duplicate, but prometheus-cpp use `vector`(just like array list), which can not ensure `instance` and `job` only be added once.

exporters/prometheus/src/collector.cc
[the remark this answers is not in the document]
dashpole: The referenced line disables sending target_info only if we encounter an error when creating the target_info metric.  If we failed to create the target_info metric, it will be nil, so we shouldn't try and send it. In the normal, non-error case, otel-go caches the target_info metric (as c.target_info), and [adds it on every scrape](https://github.com/open-telemetry/opentelemetry-go/blob/9bbefc6cc31995c7bdd37c69eb153fc2201c3562/exporters/prometheus/exporter.go#L178).

You are welcome to follow the same pattern of disabling target_info if creating the metric fails, but please do not only add target_info on the first scrape.

exporters/prometheus/src/exporter_utils.cc
dashpole: You do not need to worry about duplicate job or instance labels, since this exporter does not add `target` or `instance` labels.  If a user did add those labels as resource attributes, they would be overwritten by the prometheus server at scrape time unless the prometheus server is configured with `honor_labels=true`. By not filtering it here, we give the prometheus operator the decision of whether to keep them or not.

Duplicate label keys are handled by changes from https://github.com/open-telemetry/opentelemetry-cpp/pull/2326, so you do not need to worry about that in this PR.  This exporter takes a different (and more efficient) approach compared to otel-go, which is similar to what otel-java does: https://github.com/open-telemetry/opentelemetry-java/pull/5717.  It relies on attributes being sorted to avoid creating a map[string][]string.

Please, please do not ignore any resource attributes.

owent: OK, all ignores are removed now.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -252,6 +289,43 @@ prometheus_client::MetricType PrometheusExporterUtils::TranslateType(
...
+    return;
+  }
+  if (data.scope_metric_data_.empty())
+  {
+    return;
+  }
+  if ((*data.scope_metric_data_.begin()).metric_data_.empty())
+  {
+    return;
+  }
```
dashpole, line 306: Why do we skip target_info if scope_metrics is empty?  We should still write target_info even if we don't have any other metrics.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -324,6 +401,51 @@ void PrometheusExporterUtils::SetMetricBasic(prometheus_client::ClientMetric &me
...
+    {
+      AddPrometheusLabel(kPrometheusInstance,
+                         AttributeValueToString(service_instance_id_it->second), &metric.label);
+    }
+    else
+    {
+      // Add a empty instance label if it's not exist
+      AddPrometheusLabel(kPrometheusInstance, "", &metric.label);
+    }
+  }
```
dashpole, line 448: Please remove this.  SDK exporters should not convert service.* resource attributes to job + instance.  That conversion is done by the prometheus receiver in the otel collector, rather than SDKs.

owent: Thanks and done.

exporters/prometheus/src/exporter_utils.cc
owent: Thanks and done.

exporters/prometheus/src/collector.cc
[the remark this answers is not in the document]
lalitb: @owent - It seems that sending target_info with every scrape should be corrected, right? I'd like to revisit the PR once this is addressed, so checking.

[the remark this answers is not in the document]
owent: > @owent - It seems that sending target_info with every scrape should be corrected, right? I'd like to revisit the PR once this is addressed, so checking.

Fixed, sorry for forgot it.

dashpole (approved):
We can open a follow-up issue to cache target_info so it doesn't need to be recomputed each scrape.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -147,10 +170,10 @@ std::vector<prometheus_client::MetricFamily> PrometheusExporterUtils::TranslateT
           }
           else
           {
-            sum = nostd::get<int64_t>(histogram_point_data.sum_);
+            sum = static_cast<double>(nostd::get<int64_t>(histogram_point_data.sum_));
```
lalitb, line 173: Not related to this PR, but there is possibility of precision loss while casting from `int64_t` to `double` for larger values ( > 2^52) of `sum`. Good to have a issue to track this - to check for any such loss and log a warning.

lalitb (approved):
Nicely done. Thanks for the PR.

lalitb:
@owent Can you please resolve the conflict, probably one last change to merge this :)

exporters/prometheus/src/exporter_utils.cc
owent: Thanks, I add this static cast because it will trigger a warning with clang 16 and `-Werror` will make it a error.
BTW:Do you think think it may have epsilon problem in the code `metric->histogram.sample_count = static_cast<std::uint64_t>(values[1]);` in `PrometheusExporterUtils::SetValue` ?

owent:
I'm on holiday these days. sorry for the delay.

exporters/prometheus/src/collector.cc
```diff
@@ -36,8 +39,11 @@ std::vector<prometheus_client::MetricFamily> PrometheusCollector::Collect() cons
   collection_lock_.lock();
 
   std::vector<prometheus_client::MetricFamily> result;
-  reader_->Collect([&result](sdk::metrics::ResourceMetrics &metric_data) {
-    auto prometheus_metric_data = PrometheusExporterUtils::TranslateToPrometheus(metric_data);
+
+  bool populate_target_info = populate_target_info_;
+  reader_->Collect([&result, populate_target_info](sdk::metrics::ResourceMetrics &metric_data) {
+    auto prometheus_metric_data =
+        PrometheusExporterUtils::TranslateToPrometheus(metric_data, populate_target_info);
```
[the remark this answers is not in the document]
owent: Thanks and done.

exporters/prometheus/src/exporter_utils.cc
```diff
@@ -313,7 +381,7 @@ void PrometheusExporterUtils::SetMetricBasic(
   // Note that attribute keys are sorted, but sanitized keys can be out-of-order.
   // We could sort the sanitized keys again, but this seems too expensive to do
   // in this hot code path. Instead, we ignore out-of-order keys and emit a warning.
-  metric.label.reserve(labels.size() + 2);
+  metric.label.reserve(labels.size());
```
marcalff, line 384: Why this ?

Space is still needed to add a scope_name and scope_version.

marcalff (commented):
Small question on label reserve(size) or size +2.

exporters/prometheus/src/exporter_utils.cc
dashpole: Ah, I probably caused some confusion... The previous iteration of this PR added `job` and `instance` labels, which meant I asked that +2 to be removed in https://github.com/open-telemetry/opentelemetry-cpp/pull/2301#discussion_r1342680813.  But now, after rebasing and adding scope name and scope verison, the +2 needs to stay.

marcalff: Ok, makes sense. Resolved then, Owent just fixed it.

marcalff (approved):
LGTM, thanks for the fix.

PR #2304, opened 2023-09-11
[BUILD] Need fine-grained HAVE_CPP_STDLIB

### LINKED ISSUE #1071: Need fine-grained HAVE_CPP_STDLIB

The `std::shared_ptr` & `std::unique_ptr` are well-supported by wide-spread compilers in c++11, while `std::string_view` & `std::variant` are unavailable until c++17. We currently use a coarse-grained switch `HAVE_CPP_STDLIB` to control all these things. As a result, we cannot enable any of them.

Using our private implemented `nonstd::shared_ptr` would lead to interoperability issue with `std::shared_ptr`, which is quite annoying.

I'd like to split `HAVE_CPP_STDLIB` into more fine-grained switches:

* OPENTELEMETRY_HAVE_CXX11
  * OPENTELEMETRY_HAVE_STD_SHARED_PTR
  * OPENTELEMETRY_HAVE_STD_UNIQUE_PTR
* OPENTELEMETRY_HAVE_CXX17
  * OPENTELEMETRY_HAVE_STD_STRING_VIEW
  * OPENTELEMETRY_HAVE_STD_VARIANT
* OPENTELEMETRY_HAVE_CXX20
  * OPENTELEMETRY_HAVE_STD_SPAN

Didn't find a better approach.

marcalff:
Fixes #1071

## Changes

Fixed the following issues with unit tests:

* In `nostd/span_test.cc`:
  * the unit tests were expecting specific behaviors when contructing a span with a static extent, and providing the wrong element count.
  * According to item `(2)` in https://en.cppreference.com/w/cpp/container/span/span, this is undefined behavior.
  * Resolved by removing tests that expected undefined behavior, as the result is implementation dependent anyway (the STL implementation was not tested because of possible differences).

* In `url_parser_test`:
  * The makefile with CMake did not depend on opentelemetry_api.
  * As a result, option WITH_STL was not honored per CMake settings, it was always unset
  * Fixed to always honor WITH_STL by adding the dependency on opentelemetry_api

Fixed namespace pollution in the `API` header files:

* `#define HAVE_SPAN` is replaced by `#define OPENTELEMETRY_HAVE_SPAN`
* `#define HAVE_CPP_STDLIB` is replaced by `#define OPENTELEMETRY_STL_VERSION=<version>`

Improved the granularity of the CMake WITH_STL option:

* Before, `WITH_STL` was either a "all" (C++20, or C++17 with extra code) or nothing choice.
* After, `WITH_STL` can gradually pick witch parts of the STL to use, with `OFF`, `CXX11`, `CXX14`, `CXX17`, `CXX20` or `CXX23` options.

Fixed compiling options pollution in `CMake`:

* Before, the makefile could decide to set `CMAKE_CXX_STANDARD` to specific values, possibly interfering with existing applications makefiles and compiling options.
* After, `CMAKE_CXX_STANDARD` is no longer overwritten. The `CMAKE_CXX_STANDARD` setting, if any, or the `stdc++` compiler options, if any, provided by the caller are honored.


api/CMakeLists.txt
```diff
@@ -51,11 +51,33 @@ if(WITH_ABSEIL)
...
+  message(STATUS "Building WITH_STL=CXX14")
+  target_compile_definitions(opentelemetry_api
+                             INTERFACE OPENTELEMETRY_STL_VERSION=2014)
+elseif(WITH_STL STREQUAL "CXX17")
+  message(STATUS "Building WITH_STL=CXX17")
+  target_compile_definitions(opentelemetry_api
+                             INTERFACE OPENTELEMETRY_STL_VERSION=2017)
+elseif(WITH_STL STREQUAL "CXX20")
+  message(STATUS "Building WITH_STL=CXX20")
+  target_compile_definitions(opentelemetry_api
```
ThomsonTan, line 72: Check for C++ 23 as well?

ext/test/http/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ endif()
 set(URL_PARSER_FILENAME url_parser_test)
 add_executable(${URL_PARSER_FILENAME} ${URL_PARSER_FILENAME}.cc)
 target_link_libraries(${URL_PARSER_FILENAME} ${GTEST_BOTH_LIBRARIES}
-                      ${CMAKE_THREAD_LIBS_INIT})
+                      ${CMAKE_THREAD_LIBS_INIT} opentelemetry_api)
```
ThomsonTan, line 28: Is this required?

api/CMakeLists.txt
```diff
@@ -51,11 +51,33 @@ if(WITH_ABSEIL)
        "absl_bits" "absl_city")
 endif()
 
-if(WITH_STL)
-  message("Building with standard library types...")
-  target_compile_definitions(opentelemetry_api INTERFACE HAVE_CPP_STDLIB)
+if(WITH_STL STREQUAL "OFF")
+  message(STATUS "Building WITH_STL=OFF")
+  target_compile_definitions(opentelemetry_api
+                             INTERFACE OPENTELEMETRY_STL_VERSION=0)
```
ThomsonTan, line 57: Probably just leave `OPENTELEMETRY_STL_VERSION` as undefined here?

ext/test/http/CMakeLists.txt
owent: I think it's required. It will all the public definitions of `opentelemetry_api` to `${URL_PARSER_FILENAME}`.

api/CMakeLists.txt
marcalff: Added CXX23, it will behave as CXX20 since we do not use C++23 specific features at the moment.

ext/test/http/CMakeLists.txt
marcalff: It is required for unit test url_parser_test to honor WITH_STL, see the PR comments for details.

api/CMakeLists.txt
marcalff: Fixed.

owent (approved):

api/CMakeLists.txt
```diff
@@ -51,11 +51,36 @@ if(WITH_ABSEIL)
        "absl_bits" "absl_city")
 endif()
 
-if(WITH_STL)
-  message("Building with standard library types...")
-  target_compile_definitions(opentelemetry_api INTERFACE HAVE_CPP_STDLIB)
+if(WITH_STL STREQUAL "OFF")
+  message(STATUS "Building WITH_STL=OFF")
+elseif(WITH_STL STREQUAL "CXX11")
+  message(STATUS "Building WITH_STL=CXX1")
```
lalitb, line 57:
```suggestion
  message(STATUS "Building WITH_STL=CXX11")
```

lalitb (approved):

PR #2322, opened 2023-09-21
Fix lifetime issues in test utils

punya:
The MetricsData instance returned from these helpers outlived the  
Resource and InstrumentationScope instances that it pointed to.


punya:
@dashpole and @quentinmit this was informed by our discussion earlier today. Once it's merged we can use this approach in #2293 as well.

lalitb (approved):
Thanks for the fix.

ThomsonTan (approved):

PR #2326, opened 2023-09-23
[exporters/prometheus] Handle attribute key collisions from sanitation

### LINKED ISSUE #2290: Prometheus exporter: Handle attribute key collisions from sanitization

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-attributes,

> OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). String Attribute values are converted directly to Metric Attributes, and non-string Attribute values MUST be converted to string attributes following the [attribute specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute). Prometheus metric label keys are required to match the following regex: `[a-zA-Z_]([a-zA-Z0-9_])*`. Metrics from OpenTelemetry with unsupported Attribute names MUST replace invalid characters with the _ character. Multiple consecutive _ characters MUST be replaced with a single _ character. This may cause ambiguity in scenarios where multiple similar-named attributes share invalid characters at the same location. **In such unlikely cases, if multiple key-value pairs are converted to have the same Prometheus key, the values MUST be concatenated together, separated by ;, and ordered by the lexicographical order of the original keys.**

One solution is to group colliding keys, and then join them with `;`, like the go exporter does:
https://github.com/open-telemetry/opentelemetry-go/blob/02616a25c68e674a04df01a1b7ba185a2db8cd5a/exporters/prometheus/exporter.go#L284-L305

Another solution is to sort the keys first, and then check whether it collides with the key from the previous iteration, like i'm working on in java: https://github.com/open-telemetry/opentelemetry-java/pull/5717

punya:
Fixes #2290

## Changes

*   [x] `CHANGELOG.md` updated for non-trivial changes
*   [x] Unit tests have been added
*   [x] Changes in public API reviewed

### DIFF
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -258,16 +258,37 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
 void PrometheusExporterUtils::SetMetricBasic(prometheus_client::ClientMetric &metric,
                                              const metric_sdk::PointAttributes &labels)
 {
-  // auto label_pairs = ParseLabel(labels);
-  if (!labels.empty())
+  if (labels.empty())
   {
-    metric.label.resize(labels.size());
-    size_t i = 0;
-    for (auto const &label : labels)
+    return;
+  }
+
+  // Concatenate values for keys that collide after sanitation.
+  // Note that attribute keys are sorted, but sanitized keys can be out-of-order.
+  // We could sort the sanitized keys again, but this seems too expensive to do
+  // in this hot code path. Instead, we ignore out-of-order keys and emit a warning.
+  metric.label.reserve(labels.size());
+  std::string previous_key;
+  for (auto const &label : labels)
+  {
+    auto sanitized = SanitizeNames(label.first);
+    int comparison = previous_key.compare(sanitized);
+    if (metric.label.empty() || comparison < 0)  // new key
     {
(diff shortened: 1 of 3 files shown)


exporters/prometheus/src/exporter_utils.cc
```diff
@@ -258,16 +258,37 @@ void PrometheusExporterUtils::SetData(std::vector<T> values,
...
+  if (labels.empty())
   {
-    metric.label.resize(labels.size());
-    size_t i = 0;
-    for (auto const &label : labels)
+    return;
+  }
+
+  // Concatenate values for keys that collide after sanitation.
+  // Note that attribute keys are sorted, but sanitized keys can be out-of-order.
```
lalitb, line 267: The SDK doesn't sort the attribute keys. Unless the instrumentation library is recording them in sorted order, I am not sure if below logic would work.

punya: The type is `OrderedAttributeMap` which is an alias for `std::vector<std::string, ...>`. I see that a similar unordered type exists, but that's not the one used here.

lalitb (approved):
LGTM. Thanks.

marcalff (approved):
LGTM, thanks for the fix.

PR #2330, opened 2023-09-25
[exporters/prometheus] Sanitize labels according to Prometheus spec

### LINKED ISSUE #2289: Prometheus label keys must not include colons

From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-attributes,

> OpenTelemetry Metric Attributes MUST be converted to [Prometheus labels](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). String Attribute values are converted directly to Metric Attributes, and non-string Attribute values MUST be converted to string attributes following the [attribute specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute). Prometheus metric label keys are required to match the following regex: `[a-zA-Z_]([a-zA-Z0-9_])*`.

Metric names are allowed to include colons, but label keys are not.

SetMetricBasic translates attributes to prometheus labels:

https://github.com/open-telemetry/opentelemetry-cpp/blob/f52ec6caded074437658643566bed0d83152a78a/exporters/prometheus/src/exporter_utils.cc#L261-L276

It uses the same sanitization function used for metric names:

https://github.com/open-telemetry/opentelemetry-cpp/blob/f52ec6caded074437658643566bed0d83152a78a/exporters/prometheus/src/exporter_utils.cc#L131-L139

...which allows colons as a valid character.

punya:
Fixes #2289

*   [x] `CHANGELOG.md` updated for non-trivial changes
*   [x] Unit tests have been added
*   [x] Changes in public API reviewed

### DIFF
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -20,6 +20,85 @@ namespace exporter
 {
 namespace metrics
 {
+namespace
+{
+/**
+ * Sanitize the given metric name by replacing invalid characters with _,
+ * ensuring that multiple consecutive _ characters are collapsed to a single _.
+ *
+ * @param valid a callable with the signature `(int pos, char ch) -> bool` that
+ *        returns whether `ch` is valid at position `pos` in the string
+ * @param name the string to sanitize
+ */
+template <typename T>
+inline std::string Sanitize(std::string name, const T &valid)
+{
+  static_assert(std::is_convertible<T, std::function<bool(int, char)>>::value,
+                "valid should be a callable with the signature "
+                "(int, char) -> bool");
+
+  constexpr const auto replacement     = '_';
+  constexpr const auto replacement_dup = '=';
+
+  bool has_dup = false;
+  for (int i = 0; i < (int)name.size(); ++i)
+  {
+    if (valid(i, name[i]) && name[i] != replacement)
+    {
+      continue;
+    }
+    if (i > 0 && (name[i - 1] == replacement || name[i - 1] == replacement_dup))
+    {
+      has_dup = true;
+      name[i] = replacement_dup;
+    }
+    else
+    {
+      name[i] = replacement;
+    }
+  }
+  if (has_dup)
+  {
+    auto end = std::remove(name.begin(), name.end(), replacement_dup);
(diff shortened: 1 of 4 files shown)


exporters/prometheus/src/exporter_utils.cc
```diff
@@ -20,6 +20,69 @@ namespace exporter
...
+ * Sanitize the given metric label key according to Prometheus rule.
+ * Prometheus metric label keys are required to match the following regex:
+ * [a-zA-Z_]([a-zA-Z0-9_])*
+ */
+std::string SanitizeLabel(std::string label_key)
+{
+  return Sanitize(label_key, [](int i, char c) {
+    return (c >= 'a' && c <= 'z') ||  //
+           (c >= 'A' && c <= 'Z') ||  //
+           (c >= '0' && c <= '9' && i > 0);
```
lalitb, line 66: nit - the comment mentions underscore `_` in regex match, while it is missing in above conditional check.

punya: Thanks, that's a great catch. I didn't look very closely because I was adapting the existing implementation, which had the same bug.

punya: On second thought - the code is okay but for a subtle reason. It classifies `_` as invalid, then replaces it with `_` as part of sanitation.

I've made the change anyway for the sake of clarity.

lalitb: I felt the change is required otherwise valid string like `_ _ a` will be converted to `_ = a` and eventually to ` _ a` which is not expected :)

punya: But that *is* the desired behavior (collapse multiple consecutive underscores to a single underscore). @lalitb maybe I'm misunderstanding your comment.

lalitb: Yes, sorry for the confusion. I didn't read that part of the specs. Should be good.

lalitb (approved):

marcalff (approved):
LGTM, thanks for the fix.

PR #2335, opened 2023-09-27
[Exporter] Group spans by resource and instrumentation scope in OTLP export requests

mbrobbel:
This PR changes the implementation of `OtlpRecordableUtils::PopulateRequest` to group spans by their resource and instrumentation scope.

Instead of duplicating the resource and instrumentation scope for every span this now collects spans with the same resources and instrumentation scopes, reducing the size of the resulting `proto::collector::trace::v1::ExportTraceServiceRequest`.

### DIFF
--- a/exporters/otlp/src/otlp_recordable_utils.cc
+++ b/exporters/otlp/src/otlp_recordable_utils.cc
@@ -77,21 +77,27 @@ void OtlpRecordableUtils::PopulateRequest(
   {
     // Add the resource
     auto resource_spans = request->add_resource_spans();
-    proto::resource::v1::Resource resource_proto;
-    OtlpPopulateAttributeUtils::PopulateAttribute(&resource_proto, *input_resource_spans.first);
-    *resource_spans->mutable_resource() = resource_proto;
-    resource_spans->set_schema_url(input_resource_spans.first->GetSchemaURL());
+    if (input_resource_spans.first)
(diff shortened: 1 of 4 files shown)


exporters/otlp/src/otlp_recordable_utils.cc
```diff
@@ -56,19 +56,54 @@ void OtlpRecordableUtils::PopulateRequest(
...
+        *scope_spans->mutable_scope() = instrumentation_scope_proto;
+        scope_spans->set_schema_url(input_scope_spans.first->GetSchemaURL());
+      }
+
+      // Add all spans to this scope spans
+      for (auto &input_span : input_scope_spans.second)
+      {
+        *scope_spans->add_spans() = std::move(input_span->span());
+      }
+    }
```
lalitb, line 106: nit - perhaps we can remove `OtlpRecordable::ProtoResource()`, `OtlpRecordable::GetProtoInstrumentationScope()` and `OtlpRecordable::GetInstrumentationLibrarySchemaURL()` now that we don't use it to populate request object.

lalitb (approved):
LGTM. Nicely done. Perhaps we can do similar changes for logs too. Will raise a issue for that, so someone can pick it.

marcalff (approved):
LGTM, and very clean fix.

PR #2344, opened 2023-10-07
[BUILD] DLL export interface for Metrics

### LINKED ISSUE #2313: DLL export interface for Metrics
### Discussed in https://github.com/open-telemetry/opentelemetry-cpp/discussions/2311

<div type='discussions-op-text'>

<sup>Originally posted by **TomSvk88** September 14, 2023</sup>
Hi,

I can see from the code itself that Traces and Logs classes as exported as dll interface but Metrics classes are not. Is there any particular reason for that or is it just it was not added yet?
Thank you for any response.

Tomas</div>

perhapsmaple:
Fixes #2313 

## Changes

Add Metrics classes to exported DLL interface.


examples/common/metrics_foo_library/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+if(DEFINED OPENTELEMETRY_BUILD_DLL)
```
owent, line 4: This is not a problem of this PR, but I think we should use `if(OPENTELEMETRY_BUILD_DLL)` instead of `if(DEFINED OPENTELEMETRY_BUILD_DLL)`. And use `target_compile_definitions(opentelemetry_api INTERFACE OPENTELEMETRY_BUILD_IMPORT_DLL)` instead of `add_definitions(-DOPENTELEMETRY_BUILD_IMPORT_DLL)`, so we can exporting this definition to cmake package.

owent (approved):

ext/src/dll/CMakeLists.txt
```diff
@@ -22,6 +22,22 @@ if(WITH_OTLP_HTTP)
                         PRIVATE opentelemetry_exporter_otlp_http)
 endif()
 
+target_link_libraries(
+  opentelemetry_cpp PRIVATE opentelemetry_metrics
+                            opentelemetry_exporter_ostream_metrics)
+
+if(WITH_OTLP_GRPC)
+  add_compile_definitions(WITH_OTLP_GRPC)
```
ThomsonTan, line 30: Is this duplicated with line 14 above?

perhapsmaple: Yes, an oversight from my end. Fixed in the latest commit

lalitb:
The current CI is only executing traces example as here - https://github.com/open-telemetry/opentelemetry-cpp/blob/e918960e06b45e899cd6e1f361acdb578d968334/ci/do_ci.ps1#L77

Should we also similarly add executing the metrics example, to ensure that the example is using with DLL.

perhapsmaple:
Hi, I have added both the metrics and logs tests to the cmake.dll.test workflow. Let me know if any other changes are required. Thanks

lalitb (approved):
Thanks. Good to have @ThomsonTan approval before merging, as he wrote the initial DLL code.

ThomsonTan (approved):
Thanks for the contribution.

meastp:
@perhapsmaple thanks a lot for working on this! we discovered this today while trying to get metrics to work - we are however having some build issues. Could you have a look, perhaps you can help us solve it? :) https://github.com/open-telemetry/opentelemetry-cpp/issues/2202#issuecomment-1782639249

Issue #2297, opened 2023-09-06
[CI] Add a C++ static code analyser in the build


marcalff:
This is needed for security audits on the code.

parent issue: #2282

lalitb:
@sakshi-1505 - Thanks for the interest on this. I think @marcalff is travelling this week, so mayn't respond. It's assigned to you along with @marcalff. Feel free to come up with draft PR or suggestions to discuss further. @marcalff has some good ideas to enable this, hopefully he can guide you over.

marcalff:
@sakshi-1505 Thanks for volunteering, and welcome.

First step is to investigate the C++ tooling available, and decide which tool to use.

Second step will be to implement CI scripts to use the tool we decide to pick.

sakshi-1505:
We took a look into multiple static check tools for cpp, and we had two top contenders: `cppcheck` & `clang static analyser`.

Here's the brief comparison b/w both:

I did a comparison b/w clang static analyzer & cppcheck. Few of the difference which I could figure out were:
- The major difference b/w them is cppcheck uses flow sensitive static checks & clang analyser uses path sensitive. Due to the same no. of false positives are quite less in clang analyser.
- Now due to the above reason, clang analyser is quite slow in comparision to cppcheck.
- There are not enough support/community around clang analyser as there are around cppcheck.

I also did take a bit of GPT help here & here's what GPT says:


Cppcheck is a good choice for developers who are looking for a free and easy-to-use static check tool. It is also a good choice for developers who are new to static code analysis. Cppcheck can detect a wide range of bugs and security vulnerabilities, but it may produce some false positives.

Clang Static Analyzer is a good choice for developers who need a static check tool that is highly accurate and can handle large and complex codebases. However, it is important to note that Clang Static Analyzer can be slow and difficult to use.
I checked out both & I can integrate either of them in our CI pipeline, although since there is gh-action for cppcheck which will make it's integration more faster.

chusitoo:
Just a thought on this; would it be reasonable to create a story to add static analysis checks to the build pipeline of the contrib repo, as well? 
The repo appears to have been relatively quiet lately so maybe the timing is OK to consider making the necessary cleanup and enforcing it in the pipeline.

PR #2346, opened 2023-10-08
feat: add static checker for cpp with cppcheck

sakshi-1505:
Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/2297

## Changes

Added staticcheck analysis


sakshi-1505:
@lalitb https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/6446820822/job/17505022142?pr=2346 Looks like we will need to blacklist the `third_party` folder or the run will continue like this. Let me push a fix

sakshi-1505:
@lalitb The CI worked & was quite fast all 300+ files in just 4 min! Do we want to exit 0 at the failures for cppcheck?

sakshi-1505:
@lalitb I have made the action to return error whenever any rule is violated for staticcheck. This may cause check failure in CI, which the team might want to look at.

lalitb:
> @lalitb I have made the action to return error whenever any rule is violated for staticcheck. This may cause check failure in CI, which the team might want to look at.

Yeah the CI should fail in case of any static check violations in the code.

lalitb:
If there are only few similar pattern of errors, e.g “ The one definition rule is violated,” error as seen in logs, can we suppress cppcheck rules for those error, and let this CI pass. This will enable merging this PR, and then track those errors separately.

marcalff:
All the test code should be ignored.

I think we need a way to list, say in a separate file, all the directories to include/exclude from cppcheck.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,33 @@ on:
...
+        cd ..
+        git clone https://github.com/danmar/cppcheck.git --branch 2.12.x
+        cd cppcheck
+        mkdir build
+        cd build
+        cmake ..
+        cmake --build .
+        chmod +x bin/cppcheck
+        cd ../../opentelemetry-cpp
+        ../cppcheck/build/bin/cppcheck -ithird_party/ --error-exitcode=1 .
```
lalitb, line 36: Should we have this CI similar to that for `format` : https://github.com/open-telemetry/opentelemetry-cpp/blob/11d5d9e0d8fd8ba876c8994714cc2647479b6574/.github/workflows/ci.yml#L678-L686

- Create script to install a particular version of cppcheck passed as argument.
- And invoke the cppcheck through `./do_ci.sh cppcheck`

This will also allow developers to invoke static analysis locally.

sakshi-1505:
> All the test code should be ignored.
> 
> I think we need a way to list, say in a separate file, all the directories to include/exclude from cppcheck.

Acknowledged, let me see if there's a way

sakshi-1505:
Sorry folks for delay here, I wasn't able to find any way by which I could actually use a file for giving the exclude paths. What I can do is create an exclude_path & wrap the cppcheck execution in a bash script which may use that excludePaths, will this work @lalitb @marcalff ?

lalitb:
> What I can do is create an exclude_path & wrap the cppcheck execution in a bash script which may use that excludePaths, will this work

If you plan to create the bash script as suggested [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/2346#discussion_r1349868069), the script can read all the paths to exclude from a file, and create cppcheck arguments as  `-i<path1> -i <path2>`.

marcalff:
Hi @sakshi-1505 

Please find below some guidance to resolve current issues.

1)

The CI fails because:

```
Missing copyright in ./ci/exclude_cppcheck.txt
Missing license in ./ci/exclude_cppcheck.txt
Total number of failed checks: 2
```

To fix this, add copyright and license as comments.

And yes, it means this file must support comments: a simple `grep -v '#'` can be used to filter them out.

2)

Please rename this configuration file to `.cppcheck` at the top level directory, for consistency with other tools (.clang-format, etc)

3)

As @lalitb already indicated, the scripts should separate clearly:

* building cppcheck itself
* running cppcheck

For the CI workflow, it should use:

* env: CPPCHECK_VERSION = xxx
* sudo -E ./ci/install_cppcheck.sh

and then

* ./ci/do_ci.sh cppcheck

A developer locally will have cppcheck already installed, and will run only the second part.

4)

About the cppcheck file format, I think it is simpler to provide full cppcheck options already instead of building them.

For example:

```
# Exclude third parties
-i ./third_party

# Exclude unit tests
-i ./api/test
-i ./sdk/test

etc
```

.github/workflows/ci.yml
```diff
@@ -7,6 +7,33 @@ on:
...
+        CC: /usr/bin/gcc-10
+        CXX: /usr/bin/g++-10
+      run: |
+        sudo -E ./ci/setup_cmake.sh
+        sudo -E ./ci/setup_ci_environment.sh
+        sudo apt-get -y install cmake
+    - name: build and run cppcheck
+      run: |
+        cd ..
+        git clone https://github.com/danmar/cppcheck.git --branch 2.12.x
```
ThomsonTan, line 28: Use the `checkout` action to checkout the cppcheck repo?

.github/workflows/ci.yml
ThomsonTan: And should `test_common` also be ignored here?

.github/workflows/ci.yml
```diff
@@ -7,6 +7,33 @@ on:
...
+      with:
+        submodules: 'recursive'
+    - name: setup
+      env:
+        CC: /usr/bin/gcc-10
+        CXX: /usr/bin/g++-10
+      run: |
+        sudo -E ./ci/setup_cmake.sh
+        sudo -E ./ci/setup_ci_environment.sh
+        sudo apt-get -y install cmake
```
ThomsonTan, line 24: Should cmake install already handled by `./ci/setup_cmake.sh` 2 lines above?

sakshi-1505:
Yes I will get this updated by tonight.

marcalff:
@sakshi-1505 

This PR has been opened for 3 months now.

We (opentelemetry-cpp maintainers) have offered comments, suggestions and guidance to move forward.

The harsh reality as of time of writing (January 2024) is that not only nothing has changed (I understand this part, we all get busy), there is also (from my perspective) a bigger problem of lack of communication, with complete silence and no reply to comments, instead of engaging into a technical discussion to make progress.

I am now closing this PR.

Thanks for your interest in opentelemetry-cpp.

PR #2354, opened 2023-10-09
[BUILD] error: read-only reference ‘value’ used as ‘asm’ output

### LINKED ISSUE #2343: Build error: read-only reference ‘value’ used as ‘asm’ output
**Describe your environment** 

* SLES 15 SP 1
* g++ 7.4.1
* Kernel 4.12.14-195-default
* googletest-1.14.0 manually downloaded built and installed under /opt/gtest
* Google benchmark 1.8.3 manually downloaded built and installed under /opt/benchmark
* CMake 3.27.1 manually downloaded  and installed under 
* main branch, commit 0803e6a8f26458df4c987c5fcc19823ca530cc41 .

**Steps to reproduce**

* Used CMake to configure
* Used CMake to build

Build should have concluded successfully

During build, got the error:
```
In file included from /path/to/src/opentelemetry-cpp/sdk/test/trace/sampler_benchmark.cc:17:0:
/opt/benchmark/include/benchmark/benchmark.h: In instantiation of ‘typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type benchmark::DoNotOptimize(Tp&&) [with Tp = opentelemetry::v1::sdk::trace::ParentBasedSampler; typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type = void]’:
/path/to/src/opentelemetry-cpp/sdk/test/trace/sampler_benchmark.cc:49:85:   required from here
/opt/benchmark/include/benchmark/benchmark.h:540:46: error: read-only reference ‘value’ used as ‘asm’ output
   asm volatile("" : "+m"(value) : : "memory");
                                              ^
/opt/benchmark/include/benchmark/benchmark.h: In instantiation of ‘typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type benchmark::DoNotOptimize(Tp&&) [with Tp = opentelemetry::v1::sdk::trace::TraceIdRatioBasedSampler; typename std::enable_if<((! std::is_trivially_copyable<_Tp>::value) || (sizeof (Tp) > sizeof (Tp*)))>::type = void]’:
/path/to/src/opentelemetry-cpp/sdk/test/trace/sampler_benchmark.cc:58:60:   required from here
/opt/benchmark/include/benchmark/benchmark.h:540:46: error: read-only reference ‘value’ used as ‘asm’ output
```
**Additional context**

N/A

marcalff:
Fixes #2343

## Changes

The benchmark code fails to build with GCC, due to an upstream bug.

As a work around, comment the two failing benchmarks.

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
@@ -52,8 +52,8 @@ public:
   nostd::string_view GetDescription() const noexcept override;
 
 private:
-  std::shared_ptr<Sampler> delegate_sampler_;
-  std::string description_;
+  const std::shared_ptr<Sampler> delegate_sampler_;
+  const std::string description_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
@@ -52,8 +52,8 @@ public:
   nostd::string_view GetDescription() const noexcept override;
 
 private:
-  const std::shared_ptr<Sampler> delegate_sampler_;
-  const std::string description_;
+  std::shared_ptr<Sampler> delegate_sampler_;
+  std::string description_;
 };
 }  // namespace trace
 }  // namespace sdk
--- a/sdk/include/opentelemetry/sdk/trace/samplers/trace_id_ratio.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/trace_id_ratio.h
@@ -64,7 +64,7 @@ public:
 
 private:
   std::string description_;
-  uint64_t threshold_;
+  const uint64_t threshold_;
 };
 }  // namespace trace
 }  // namespace sdk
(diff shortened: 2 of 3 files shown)


lalitb:
I thought commenting the faling test was the better approach :) Not sure if we should remove const-ness of member variables for upstream bug. This would in general prevent any optimizations from compiler, and also type safety.

marcalff:
> I thought commenting out the failing test for time being was the better approach :) Not super sure if we should ditch the const on those member variables because of a bug somewhere else This would in general prevent any optimizations from compiler, and also type safety.

The `const` itself is highly suspicious.

Members like:

* `const char *ptr;`
* `const & int;`

make sense, but I have never seen a member using:

* `const std::string;`

If the member is really `const`, how is it properly destroyed by the destructor for example ?

I think there is confusion in this code, if an object is `const`, the entire instance should be const (as in `const ParentBasedSampler *sampler`), using a non const ParentBasedSampler object which has a const member makes no sense (to me).

marcalff:
Taking this for example:

```
const std::string description_;
```

The __method__  GetDescription() is const, so it does not change the object, this part is expected.

Making the __member__ const is rather strange.

marcalff:
Reverted to first solution, comment out code that fails to build.

lalitb (approved):
Thanks for troubleshooting, and fixing it.

PR #2356, opened 2023-10-10
Fix behavior of opentracing-shim when added as subdirectory of a larger project

### LINKED ISSUE #2355: opentracing-shim target includes are missing when opentelemetry-cpp is added as a subdirectory to a larger project

Rocky Linux 8, cmake version 3.27.7

**Steps to reproduce**
1. Create a cmake project (`/myproject`) that adds opentelemetry-cpp as a subdirectory, either with add_subdirectory or FetchContent_MakeAvailable
2. Add opentelemetry_opentracing_shim as a link to a local target in your project
3. Try to build your local target

I expect that the opentelemetry_opentracing_shim include paths point to the location of the third_party/opentracing-cpp header files

The include paths are relative to my local project's root, not the opentelemetry-cpp root. That is, I see paths like `/myproject/third_party/opentracing-cpp/include` instead of `/myproject/opentelemetry-cpp/third_party/opentracing-cpp/include`. This causes errors about missing headers.

**Additional context**
The root cause appears to be in `opentracing-shim/CMakeLists.txt`. It is added some include dirs to the shim target, but they are relative to CMAKE_BINARY_DIR/CMAKE_SORUCE_DIR. This works when opentelemetry-cpp is the top-level project, but in this case `myproject` is at the top-level, so these paths are no longer correct. I think making these paths relative to PROJECT_BINARY_DIR/PROJECT_SOURCE_DIR would be better since the paths would always be constructed relative to the actual opentelemetry-cpp directory.

andremarianiello:
Fixes #2355

## Changes

Change the opentracing-shim include paths to be relative to opentelemetry-cpp directory, rather than the root cmake project directory. Those two directories will be different if opentelemetry-cpp is added to a larger project.

I also switched `include_directories` to `target_include_directories` as my understanding is that the latter is current best-practice. This necessitated gating the includes with a BUILD_INTERFACE generator expression. However, I don't think this change is required for this fix. I just wanted to improve it a little bit.

### DIFF
--- a/opentracing-shim/CMakeLists.txt
+++ b/opentracing-shim/CMakeLists.txt
@@ -14,10 +14,13 @@ target_include_directories(
                         "$<INSTALL_INTERFACE:include>")
 
 if(OPENTRACING_DIR)
-  include_directories(
-    "${CMAKE_BINARY_DIR}/${OPENTRACING_DIR}/include"
-    "${CMAKE_SOURCE_DIR}/${OPENTRACING_DIR}/include"
-    "${CMAKE_SOURCE_DIR}/${OPENTRACING_DIR}/3rd_party/include")
+  target_include_directories(
+    ${this_target}
+    PUBLIC
+      "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/${OPENTRACING_DIR}/include>"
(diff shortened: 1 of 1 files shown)


lalitb (approved):

marcalff:
@owent Could you take a look ? CMake related

PR #2362, opened 2023-10-12
Fix protoc searching with non-imported protobuf::protoc target.

### LINKED ISSUE #2360: cmake fails to correctly use non-imported protobuf::protoc target

Rocky Linux 9, cmake 3.27.7, protobuf 21.12

**Steps to reproduce**
Create a local project (`myproject`). Add Protobuf and opentelemetry-cpp as dependencies using FetchContent_Declare, with OVERRIDE_FIND_PACKAGE, then add opentelemetry-cpp to the project via `find_package`/`FetchContent_MakeAvailable`/`add_subdirectory`. Then configure and build

```
cmake_minimum_required(VERSION 3.24)

project(myproject)

include(FetchContent)

set(protobuf_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
set(protobuf_BUILD_SHARED_LIBS ON CACHE INTERNAL "" FORCE)
FetchContent_Declare(
  Protobuf
  GIT_REPOSITORY "https://github.com/protocolbuffers/protobuf.git"
  GIT_TAG "v21.12"
  OVERRIDE_FIND_PACKAGE
  )

set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
set(WITH_EXAMPLES OFF CACHE BOOL "" FORCE)
set(WITH_OTLP_HTTP ON CACHE BOOL "" FORCE) # triggers the inclusion of protobuf
FetchContent_Declare(
  opentelemetry-cpp
  GIT_REPOSITORY "https://github.com/open-telemetry/opentelemetry-cpp.git"
  GIT_TAG "main"
  )

FetchContent_MakeAvailable(opentelemetry-cpp)
```

I would expect a full build to succeed, and I would expect opentelemetry-proto to build using the protoc built from the Declared protobuf package.

We fail to build opentelemetry-proto protos
```
[ 27%] [Run]: "" "--proto_path=/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto" "--cpp_out=/root/tmp/build/generated/third_party/opentelemetry-proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/common/v1/common.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/resource/v1/resource.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/trace/v1/trace.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/logs/v1/logs.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/metrics/v1/metrics.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/collector/logs/v1/logs_service.proto" "/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto"
/bin/sh: line 1: --proto_path=/root/tmp/build/_deps/opentelemetry-cpp-src/third_party/opentelemetry-proto: No such file or directory
```

Notice the first argument is `""` which is supposed to be the path to protoc.

**Additional context**
The top-level CMakeLists.txt puts the path to protoc in PROTOBUF_PROTOC_EXECUTABLE via calls to `find_package(Protobuf)`. However, when `find_package` is overridden (for any reason - OVERRIDE_FIND_PACKAGE) is just one way) then PROTOBUF_PROTOC_EXECUTABLE is not set. There is code to extract an executable path from an imported `protobuf::protoc` target, BUT this only works if it is actually an imported target. If that happens to be an ALIAS target, like would be the case when overriding `find_package`, then `PROTOBUF_PROTOC_EXECUTABLE` remains unset, causing this problem.

**Workaround**
You can *attempt* to work around this problem by defining `PROTOBUF_PROTOC_EXECUTABLE` yourself. In my example you could use
```
set(PROTOBUF_PROTOC_EXECUTABLE ${CMAKE_BINARY_DIR}_deps/protobuf-build/protoc)
```
though
```
FetchContent_MakeAvailable(Protobuf)
set(PROTOBUF_PROTOC_EXECUTABLE ${protobuf_BINARY_DIR}/protoc)
```
is probably more correct. Neither way solve the whole problem though. The protoc command will be correct, but protoc itself might not yet exists. This is because passing the path to opentelemetry-proto this way doesn't impose a dependency on protoc.  You can workaround this problem by adding
```
add_dependencies(opentelemetry_proto protobuf::protoc)
```
but IMO none of these workarounds should be required.

owent:
Fixes #2360 

## Changes

+ Fix protoc searching with non-imported protobuf::protoc target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -430,6 +430,10 @@ if(WITH_OTLP_GRPC OR WITH_OTLP_HTTP)
   if(TARGET protobuf::protoc)
     project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
                                               protobuf::protoc)
+    # If protobuf::protoc is not a imported target, then we use the target directly for fallback
+    if(NOT PROTOBUF_PROTOC_EXECUTABLE)
+      set(PROTOBUF_PROTOC_EXECUTABLE protobuf::protoc)
+    endif()
   elseif(Protobuf_PROTOC_EXECUTABLE)
     # Some versions of FindProtobuf.cmake uses mixed case instead of uppercase
     set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -430,7 +430,8 @@ if(WITH_OTLP_GRPC OR WITH_OTLP_HTTP)
   if(TARGET protobuf::protoc)
     project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
                                               protobuf::protoc)
-    # If protobuf::protoc is not a imported target, then we use the target directly for fallback
+    # If protobuf::protoc is not a imported target, then we use the target
+    # directly for fallback
     if(NOT PROTOBUF_PROTOC_EXECUTABLE)
       set(PROTOBUF_PROTOC_EXECUTABLE protobuf::protoc)
     endif()


andremarianiello:
I filed #2360 and this fix worked for me. I would like to see it merged when possible. Thanks for the fix!

marcalff (approved):
LGTM, and fix tested by reporter.

Thanks for the fix.

Issue #2081, opened 2023-03-31
When using CreateInt64ObservableCounter I get "Error during observe.The metric storage is invalid"


ashley-b:
**Describe your environment** 
opentelemetry-cpp: 1.8.1
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
Fedora release 37

**Steps to reproduce**
Using a simple OStreamMetricExporter and observable counter meter I find if the counter fall out of scope before running it errors in debug or segfaults in release.
I assume I am using it incorrectly, but I would assume when the counter goes out of scope it should clean up it's resource correctly.
```
int main()
{
    create_exporter();

    auto meterProvider = metrics_api::Provider::GetMeterProvider();
    auto meter = meterProvider->GetMeter("test", "");
    {
        auto counter = meter->CreateInt64ObservableCounter("counter");
        counter->AddCallback(&get_counter, nullptr);
    }
    std::this_thread::sleep_for(std::chrono::seconds(10));
    return 0;
}
```

**What is the expected behavior?**
Not to see a Error message or segfault

**What is the actual behavior?**
In debug build
```
[Error] File: opentelemetry-cpp/1.8.1/sdk/src/metrics/state/observable_registry.cc:55[ObservableRegistry::Observe] - Error during observe.The metric storage is invalid
```
In Release build it segfaults

**Additional notes**
The life times and ownership of opentelemetry-cpp object is not clear from the documentation

lalitb:
@ashley-b This is documented here - https://opentelemetry.io/docs/instrumentation/cpp/manual/#create-observable-counter

``
Create a Observable Counter Instrument from the Meter, and add a callback. The callback would be used to record the measurement during metrics collection. Ensure to keep the Instrument object active for the lifetime of collection.
``

This is requirement for the asynchronous instruments.

marcalff:
@lalitb 

After `AddCallback()` is called, the `ObservableRegistry` contains a back pointer to the instrument.

I think that what is missing here, is cleanup in the `ObservableInstrument` destructor to remove back pointers from the registry, possibly printing a warning to complain about the missing call to `RemoveCallback`.

This will make the code more robust, and avoid crashes in cases of misuse.

The proper usage pattern is not changed, but the likelihood of users following the proper pattern will be much improved.

lalitb:
[...]

PR #2376, opened 2023-10-18
[SDK] Metrics ObservableRegistry Cleanup

### LINKED ISSUE #2373: Error during observe.The metric storage is invalid on removing callbacks
OpenTelemetry v1.9.1 via vcpkg
OpenSUSE 15.5
**Steps to reproduce**

as per opentelemetry specifications,

```
For callback functions registered after an asynchronous instrument is created, the API is required to support a mechanism for unregistration. For example, the object returned from register_callback can support an unregister() method directly
```

for c++, there is AddCallback() and RemoveCallback()

here is how I've created a async gauge, this is done in a new thread whose only job is to check the addresses and update values based on testFetcher.

```
auto meter_provider = metrics_api::Provider::GetMeterProvider();
instrument_holder.emplace_back(meter->CreateDoubleObservableGauge(gauge_name, description, unit));
instrument_holder.back()->AddCallback(MeasurementFetcher::testFetcher, value);
```

When I want to end all metric callbacks, I'm facing issues with the handling of the lifecycle of threads and the instrument objects.
My process - 
1 - remove callbacks from each ` ObservableCallbackPtr ` in ` instrument_holder `
2 - erase all items ` instrument_holder ` 
3 - join all threads created, which were stored in an object called ` threads `

```
std::cout<<"end callback for size "<< instrument_holder.size()<<std::endl;
for(int i=0; i<instrument_holder.size();i++)
{
	std::cout<<"inside end callback for index "<< i<<std::endl;
	instrument_holder.back()->RemoveCallback(MeasurementFetcher::testFetcher,nullptr);
}
instrument_holder.erase(instrument_holder.begin(),instrument_holder.end());
for(auto &a: threads)
{
	a.join();
}
threads.erase(threads.begin(),threads.end());
```

Should not get any errors while exporting.

I get the following error 
```
[Error] File: /home/kmh/vcpkg/buildtrees/opentelemetry-cpp/src/v1.9.1-711a32e9e2.clean/sdk/src/metrics/state/observable_registry.cc:56 [ObservableRegistry::Observe] - Error during observe.The metric storage is invalid
```

**Additional context**

from issue #2081, this seems to be an issue regarding the destructor.
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/async_instruments.cc doesn't have any defined destructor.

What is the current behaviour on the object going out of scope, what does the destructor do? Does the backpointer still remain?

I'm not that well versed in C++, so please let me know if I'm missing something very obvious. How can I get rid of this error?

perhapsmaple:
Fixes #2373, #2081 

## Changes

Added a function called CleanupCallbacks to ObservableRegistry that is called from the destructor of ObservableInstrument to remove back pointers from the registry. 

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/async_instruments.h
+++ b/sdk/include/opentelemetry/sdk/metrics/async_instruments.h
@@ -25,6 +25,7 @@ public:
   ObservableInstrument(InstrumentDescriptor instrument_descriptor,
                        std::unique_ptr<AsyncWritableMetricStorage> storage,
                        std::shared_ptr<ObservableRegistry> observable_registry);
+  ~ObservableInstrument() override;
 
   void AddCallback(opentelemetry::metrics::ObservableCallbackPtr callback,
                    void *state) noexcept override;
--- a/sdk/include/opentelemetry/sdk/metrics/state/observable_registry.h
+++ b/sdk/include/opentelemetry/sdk/metrics/state/observable_registry.h
@@ -35,6 +35,8 @@ public:
                       void *state,
                       opentelemetry::metrics::ObservableInstrument *instrument);
 
+  void CleanupCallback(opentelemetry::metrics::ObservableInstrument *instrument);
+
   void Observe(opentelemetry::common::SystemTimestamp collection_ts);
 
 private:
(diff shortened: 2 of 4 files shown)


lalitb (approved):
Thanks.

marcalff (approved):
LGTM, thanks for the fix.

PR #2379, opened 2023-10-19
[SDK] Creating DoubleUpDownCounter with no matching view

### LINKED ISSUE #2377: Warning with DoubleUpDownCounter and negative values in 1.12.0
**Describe your environment** 
otel-cpp: v1.12.0
Platform: macOS (but happens on others as well)

**Steps to reproduce**
- Create a `DoubleUpDownCounter`
- Add a negative value to it

The counter's value is decremented by the value.

```
[Warning] File: /Users/tim/Desktop/projects/opentelemetry-cpp/sdk/src/metrics/aggregation/sum_aggregation.cc:90  DoubleSumAggregation::Aggregate Negative value ignored for Monotonic increasing measurement. Value-374.1
```

**Additional context**
Attached is a patch for the metrics_simple example that exhibits the problem.

[updown.patch](https://github.com/open-telemetry/opentelemetry-cpp/files/13031352/updown.patch)

lalitb:
Fixes #2377

## Changes

In case of a matching view, the SDK picks the Aggregation with correct monotonic for DoubleUpDownCounter as [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/231ca4ade18c8cf691c9520d174b518feae78f35/sdk/include/opentelemetry/sdk/metrics/aggregation/default_aggregation.h#L106). The problem was in the logic when there is no matching view, which is fixed now. Also added test for this scenario.

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/aggregation/default_aggregation.h
+++ b/sdk/include/opentelemetry/sdk/metrics/aggregation/default_aggregation.h
@@ -35,7 +35,8 @@ public:
       case AggregationType::kSum:
         return (instrument_descriptor.value_type_ == InstrumentValueType::kLong)
                    ? std::move(std::unique_ptr<Aggregation>(new LongSumAggregation(is_monotonic)))
-                   : std::move(std::unique_ptr<Aggregation>(new DoubleSumAggregation(true)));
+                   : std::move(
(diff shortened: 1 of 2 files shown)


marcalff (approved):
LGTM, thanks for the fix

PR #2407, opened 2023-11-15
[EXPORTER] Async exporting for otlp grpc

### LINKED ISSUE #2046: [Logs SDK] Log Appender for spdlog
[spdlog](https://log4cpp.sourceforge.net/#git) is one of the commonly used logging framework for C++, and it would be good to have the logging appender for the framework using logging bridge API.

More details:

Bridge API specs - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md
Bridge API implementation -  https://github.com/open-telemetry/opentelemetry-cpp/tree/main/api/include/opentelemetry/logs
Implementing sinks in spdlog - https://github.com/gabime/spdlog/wiki/4.-Sinks#implementing-your-own-sink

The implementation should be part of the [opentelemetry-cpp-contrib](https://github.com/open-telemetry/opentelemetry-cpp-contrib) repo under the `instrumentation` directory.

### LINKED ISSUE #2406: Async exporting for OTLP gRPC exporter

If we exporting too many logs a a high delay system, OTLP gRPC exporter is easy to drop datas.
OTLP HTTP exporter support async exporting to solve this problem for a long time.

**Additional context**

+ https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlpgrpc-concurrent-requests
+ https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-concurrent-requests

owent:
Fixes #2406

## Changes

+ Add arena for OTLP exporters.
+ Add async exporting for OTLP gRPC exporter, and move data lifetime with RPC calls.


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
```diff
@@ -25,12 +29,21 @@ namespace otlp
...
 /**
  * The OTLP gRPC client contains utility functions of gRPC.
  */
 class OtlpGrpcClient
 {
 public:
+#ifdef ENABLE_ASYNC_EXPORT
+  OtlpGrpcClient();
+  ~OtlpGrpcClient();
+#endif
```
marcalff, line 45: nit - always declare the constructor and destructor, and use the ifdef in the implementation

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -23,6 +31,73 @@ namespace exporter
 namespace otlp
 {
 
+#ifdef ENABLE_ASYNC_EXPORT
+
+namespace
+{
+struct OPENTELEMETRY_LOCAL_SYMBOL OtlpGrpcAsyncCallDataBase
```
marcalff, line 38: I don't understand the use of OPENTELEMETRY_LOCAL_SYMBOL on a type.

I would expect this to apply to global variables, if any, not on a type definition.

Please clarify (or remove).

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -23,6 +31,73 @@ namespace exporter
...
+  std::unique_ptr<google::protobuf::Arena> arena;
+  grpc::Status grpc_status;
+  std::unique_ptr<grpc::ClientContext> grpc_context;
+
+  opentelemetry::sdk::common::ExportResult export_result =
+      opentelemetry::sdk::common::ExportResult::kFailure;
+  GrpcAsyncCallback grpc_async_callback = nullptr;
+
+  OtlpGrpcAsyncCallDataBase() {}
+  virtual ~OtlpGrpcAsyncCallDataBase() {}
```
marcalff, line 51: A struct with a virtual ? Either use a class, or remove the virtual destructor.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -23,6 +31,73 @@ namespace exporter
...
+};
+
+template <class GrpcRequestType, class GrpcResponseType>
+struct OPENTELEMETRY_LOCAL_SYMBOL OtlpGrpcAsyncCallData : public OtlpGrpcAsyncCallDataBase
```
marcalff, line 55: Likewise, OPENTELEMETRY_LOCAL_SYMBOL seems out of place.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -23,6 +31,73 @@ namespace exporter
...
+  ResponseType *response = nullptr;
+
+  std::function<bool(opentelemetry::sdk::common::ExportResult,
+                     std::unique_ptr<google::protobuf::Arena> &&,
+                     const RequestType &,
+                     ResponseType *)>
+      result_callback;
+
+  OtlpGrpcAsyncCallData() {}
+  virtual ~OtlpGrpcAsyncCallData() {}
```
marcalff, line 70: Likewise, make this a class not a struct.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -46,16 +121,144 @@ static std::string GetFileContentsOrInMemoryContents(const std::string &file_pat
...
+  };
+
+  ++async_data->start_request_counter;
+  ++async_data->running_requests;
+#  if defined(GRPC_CPP_VERSION_MAJOR) && \
+      (GRPC_CPP_VERSION_MAJOR * 1000 + GRPC_CPP_VERSION_MINOR) >= 1039
+  stub->async()
+#  else
+  stub->experimental_async()
+#  endif
```
marcalff, line 199: What is the motivation for supporting GRPC < 1.39 ?

This seems very old, can we document that GRPC async requires a recent GRPC instead ?

If somehow this causes compile time breaks with OTLP HTTP async, how about to use a different compile time flag, like ENABLE_GRPC_ASYNC_EXPORT, to decouple HTTP async from GRPC async.

This will allow the OTLP HTTP implementation to become mainline sooner too, if we decide to do so.

exporters/otlp/src/otlp_grpc_exporter.cc
```diff
@@ -52,37 +60,85 @@ sdk::common::ExportResult OtlpGrpcExporter::Export(
...
+    return client_->DelegateAsyncExport(
+        options_, trace_service_stub_.get(), std::move(context), std::move(arena),
+        std::move(*request),
+        [](opentelemetry::sdk::common::ExportResult result,
+           std::unique_ptr<google::protobuf::Arena> &&,
+           const proto::collector::trace::v1::ExportTraceServiceRequest &request,
+           proto::collector::trace::v1::ExportTraceServiceResponse *) {
+          if (result != opentelemetry::sdk::common::ExportResult::kSuccess)
+          {
+            OTEL_INTERNAL_LOG_ERROR("[OTLP HTTP Client] ERROR: Export "
```
marcalff, line 93: OTLP TRACE GRPC Exporter

exporters/otlp/src/otlp_grpc_exporter.cc
```diff
@@ -52,37 +60,85 @@ sdk::common::ExportResult OtlpGrpcExporter::Export(
...
+                                    << request.resource_spans_size()
+                                    << " trace span(s) error: " << static_cast<int>(result));
+          }
+          else
+          {
+            OTEL_INTERNAL_LOG_DEBUG("[OTLP HTTP Client] Export " << request.resource_spans_size()
```
marcalff, line 99: OTLP TRACE GRPC Exporter

exporters/otlp/src/otlp_grpc_exporter.cc
```diff
@@ -52,37 +60,85 @@ sdk::common::ExportResult OtlpGrpcExporter::Export(
...
+      return sdk::common::ExportResult::kFailure;
+    }
+#ifdef ENABLE_ASYNC_EXPORT
   }
+#endif
   return sdk::common::ExportResult::kSuccess;
 }
 
-bool OtlpGrpcExporter::ForceFlush(std::chrono::microseconds /* timeout */) noexcept
+bool OtlpGrpcExporter::ForceFlush(std::chrono::microseconds timeout) noexcept
```
marcalff, line 124: Probably needs some MAYBE_UNUSED to avoid compile warnings without async.
Same with every force flush / shutdown method.

exporters/otlp/src/otlp_grpc_log_record_exporter.cc
```diff
@@ -65,35 +73,85 @@ opentelemetry::sdk::common::ExportResult OtlpGrpcLogRecordExporter::Export(
...
+    return client_->DelegateAsyncExport(
+        options_, log_service_stub_.get(), std::move(context), std::move(arena),
+        std::move(*request),
+        [](opentelemetry::sdk::common::ExportResult result,
+           std::unique_ptr<google::protobuf::Arena> &&,
+           const proto::collector::logs::v1::ExportLogsServiceRequest &request,
+           proto::collector::logs::v1::ExportLogsServiceResponse *) {
+          if (result != opentelemetry::sdk::common::ExportResult::kSuccess)
+          {
+            OTEL_INTERNAL_LOG_ERROR("[OTLP HTTP Client] ERROR: Export "
```
marcalff, line 106: OTLP LOG GRPC Exporter

exporters/otlp/src/otlp_grpc_log_record_exporter.cc
```diff
@@ -65,35 +73,85 @@ opentelemetry::sdk::common::ExportResult OtlpGrpcLogRecordExporter::Export(
...
+                                    << request.resource_logs_size()
+                                    << " log(s) error: " << static_cast<int>(result));
+          }
+          else
+          {
+            OTEL_INTERNAL_LOG_DEBUG("[OTLP HTTP Client] Export " << request.resource_logs_size()
```
marcalff, line 112: OTLP LOG GRPC Exporter

exporters/otlp/src/otlp_grpc_metric_exporter.cc
```diff
@@ -60,35 +66,85 @@ opentelemetry::sdk::common::ExportResult OtlpGrpcMetricExporter::Export(
...
+    return client_->DelegateAsyncExport(
+        options_, metrics_service_stub_.get(), std::move(context), std::move(arena),
+        std::move(*request),
+        [](opentelemetry::sdk::common::ExportResult result,
+           std::unique_ptr<google::protobuf::Arena> &&,
+           const proto::collector::metrics::v1::ExportMetricsServiceRequest &request,
+           proto::collector::metrics::v1::ExportMetricsServiceResponse *) {
+          if (result != opentelemetry::sdk::common::ExportResult::kSuccess)
+          {
+            OTEL_INTERNAL_LOG_ERROR("[OTLP HTTP Client] ERROR: Export "
```
marcalff, line 99: OTLP METRIC GRPC Exporter

exporters/otlp/src/otlp_grpc_metric_exporter.cc
```diff
@@ -60,35 +66,85 @@ opentelemetry::sdk::common::ExportResult OtlpGrpcMetricExporter::Export(
...
+                                    << request.resource_metrics_size()
+                                    << " metric(s) error: " << static_cast<int>(result));
+          }
+          else
+          {
+            OTEL_INTERNAL_LOG_DEBUG("[OTLP HTTP Client] Export " << request.resource_metrics_size()
```
marcalff, line 105: OTLP METRIC GRPC Exporter

Please adjust all the logging for the HTTP exporters as well, they use "OTLP HTTP Client" instead of "OTLP TRACE/METRIC/LOGS HTTP Exporter".

marcalff (commented):
Thanks for the fix.

It will take me several review passes to go over this change,
please find a first batch of comments for now.

exporters/otlp/src/otlp_grpc_client.cc
owent: This will hide vtable and functions when compiling with `-fvisibility=default`.

exporters/otlp/src/otlp_grpc_metric_exporter.cc
owent: Thanks and one.

exporters/otlp/src/otlp_grpc_exporter.cc
owent: Thanks and one.

exporters/otlp/src/otlp_grpc_client.cc
owent: Thanks and one.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
owent: Thanks and one.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
```diff
@@ -25,12 +29,20 @@ namespace otlp
...
+struct OtlpGrpcClientAsyncData;
+#endif
+
 /**
  * The OTLP gRPC client contains utility functions of gRPC.
  */
 class OtlpGrpcClient
 {
 public:
+  OtlpGrpcClient();
```
marcalff, line 42: In this class, existing method `OtlpGrpcClient::MakeCompletionQueue()` seems to be dead code.

And beside, it was not protected by ifdef ENABLE_ASYNC_EXPORT.

Please remove.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -46,16 +129,150 @@ static std::string GetFileContentsOrInMemoryContents(const std::string &file_pat
...
+                 if (call_data->grpc_async_callback)
+                 {
+                   call_data->grpc_async_callback(call_data.get());
+                 }
+
+                 // Maybe wake up blocking DelegateAsyncExport() call
+                 async_data->session_waker.notify_all();
+               });
+
+  // Can not cancle when start the request
```
marcalff, line 228: Typo: cancel

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -46,16 +129,150 @@ static std::string GetFileContentsOrInMemoryContents(const std::string &file_pat
...
+  }
+
+  return opentelemetry::sdk::common::ExportResult::kSuccess;
+}
+#endif
 }  // namespace
 
+OtlpGrpcClient::OtlpGrpcClient()
+#ifdef ENABLE_ASYNC_EXPORT
+    : is_shutdown_
```
marcalff, line 244: style: `is_shutdown(false)` will probably work better for clang-format.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -46,16 +129,150 @@ static std::string GetFileContentsOrInMemoryContents(const std::string &file_pat
...
+#ifdef ENABLE_ASYNC_EXPORT
+  std::shared_ptr<OtlpGrpcClientAsyncData> async_data;
+  async_data.swap(async_data_);
+
+  while (async_data && async_data->running_requests.load(std::memory_order_acquire) > 0)
+  {
+    std::unique_lock<std::mutex> lock{async_data->session_waker_lock};
+    async_data->session_waker.wait_for(lock, async_data->export_timeout, [async_data]() {
+      return async_data->running_requests.load(std::memory_order_acquire) <=
+             async_data->max_concurrent_requests;
```
marcalff, line 262: The condition in the wait_for loop is a bit surprising.

Should it be `running_requests == 0` instead of `running_requests < max_concurrent_requests` ?

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -140,31 +365,216 @@ OtlpGrpcClient::MakeLogsServiceStub(const OtlpGrpcClientOptions &options)
...
+  std::shared_ptr<OtlpGrpcClientAsyncData> async_data = MutableAsyncData(options);
+  return InternalDelegateAsyncExport(async_data, stub, std::move(context), std::move(arena),
+                                     std::move(request), std::move(result_callback), logs_count,
+                                     "log(s)");
+}
+
+std::shared_ptr<OtlpGrpcClientAsyncData> OtlpGrpcClient::MutableAsyncData(
+    const OtlpGrpcClientOptions &options)
+{
+  if (!async_data_)
```
marcalff, line 502: In my understanding, `MutableAsyncData()` can be called concurrently by several threads.

When 2 threads call this for the first time, when `async_data_` is still nullptr, is there a race condition here, with two threads invoking std::make_shared() and competing to assign `async_data_` ?

What if thread 1 returns async_data_ allocated first, and then thread 2 assign a second async_data_, the first shared pointer returned is no longer referenced internally, and could cause issues during flush / shutdown.

Use swap instead of an assignment, so at least the caller don't see a partially initialized async data ?

exporters/otlp/src/otlp_grpc_client.cc
marcalff: Ping - do we need to support GRPC < 1.39, or can we just document a recent GRPC is needed instead ?

marcalff (commented):
Impressive work.

Please find here more comments / questions.

exporters/otlp/src/otlp_grpc_client.cc
owent: According to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#exportbatch and in my understanding. `Export()` will never be called concurrently, and `MutableAsyncData()` will only be called in `Export()` ?

owent:
> Ping - do we need to support GRPC < 1.39, or can we just document a recent GRPC is needed instead ?
It would be better if we can suppor a low version of gRPC. We have some project with old toolchains which can only use gRPC 1.33.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
owent: Thanks, and done.

exporters/otlp/src/otlp_grpc_client.cc
owent: Thanks and fixed

exporters/otlp/src/otlp_grpc_client.cc
owent: Good point.

exporters/otlp/src/otlp_grpc_client.cc
owent: Yes and thanks, it's fixed now.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
marcalff: My comment about "not protected by ifdef ENABLE_ASYNC_EXPORT" was about method OtlpGrpcClient::MakeCompletionQueue() which is now removed, great.

I see that the constructor and destructor for `OtlpGrpcClient` are now declared with an ifdef, instead of always declaring them and using the ifdef inside the implementation.

To clarify, I think the previous code was better:

in header file:

```
class OtlpGrpcClient
{
public:
  OtlpGrpcClient();
```

in cc file:

```
OtlpGrpcClient::OtlpGrpcClient() {
ifdef ENABLE_ASYNC_EXPORT
... code ...
endif
}
```

This confusion was probably caused by me, commenting on the constructor and saying "not protected by ifdef ENABLE_ASYNC_EXPORT" when talking about MakeCompletionQueue(), sorry about this.

Please use the form you prefer, no need to go back and forth on this, this is minor anyway.

marcalff:
> > Ping - do we need to support GRPC < 1.39, or can we just document a recent GRPC is needed instead ?
> 
> @marcalff It would be better if we can suppor a low version of gRPC. We have some project with old toolchains which can only use gRPC 1.33.

Thanks for the clarification. Resolved then.

exporters/otlp/src/otlp_grpc_client.cc
marcalff: Ok, thanks for the clarification. Resolved.

marcalff (approved):
Excellent work, thanks for the fix.

I will not pretend I know grpc async in depth to fully appreciate the code here.
The patch looks good to me, based on my limited knowledge of this area.

Given the patch size, a second opinion is needed.

marcalff:
@lalitb @ThomsonTan @esigo 

Added the please review label, for a second review.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
owent: Personally, I also prefer the previous one, it produces the same symbols with or without async mode.

`OtlpGrpcClient::MakeCompletionQueue()` is used to ensure the symbols in headers of gRPC are produced in otlp_grpc_client. I'm not sure but they may also be created in `DelegateExport()`.

I will change the constructor and destructor to previous implementation.

exporters/otlp/src/otlp_grpc_client.cc
marcalff: Looks like you reverted a fix for `running_requests == 0`.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
marcalff: The last commit reverted the fix for the wait_for condition in the loop.

Should be running_requests == 0.

exporters/otlp/src/otlp_grpc_client.cc
owent: Sorry, it's fixed now.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client.h
owent: Sorry, it's fixed now.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_exporter.h
```diff
@@ -23,6 +23,8 @@ namespace exporter
 namespace otlp
 {
 
+class OtlpGrpcClient;
```
ThomsonTan, line 26: Can this be moved to `otlp_grpc_exporter_options.h` as a central place?

owent: Users use factory and `otlp_grpc*_options.h` to create OTLP gRPC exporters, but I think users do not need know the internal class `OtlpGrpcClient`?

esigo (commented):
Thanks @owent for yet another great work.
I had a minor question only.

exporters/otlp/test/otlp_grpc_log_record_exporter_test.cc
```diff
@@ -41,6 +42,135 @@ namespace exporter
 namespace otlp
 {
 
+namespace
+{
+class OtlpMockTraceServiceStub : public proto::collector::trace::v1::MockTraceServiceStub
+{
+public:
+#if defined(GRPC_CPP_VERSION_MAJOR) && \
+    (GRPC_CPP_VERSION_MAJOR * 1000 + GRPC_CPP_VERSION_MINOR) >= 1039
```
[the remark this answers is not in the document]
marcalff: See previous comment: https://github.com/open-telemetry/opentelemetry-cpp/pull/2407#issuecomment-1884451500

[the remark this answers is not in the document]
owent: Thanks, Comments ad added.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -140,31 +356,216 @@ OtlpGrpcClient::MakeLogsServiceStub(const OtlpGrpcClientOptions &options)
...
+bool OtlpGrpcClient::Shutdown(std::chrono::microseconds timeout) noexcept
+{
+  if (!async_data_)
+  {
+    return true;
+  }
+
+  if (!is_shutdown_)
+  {
+    is_shutdown_ = true;
```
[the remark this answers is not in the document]
owent: Does it need thread safe?In my understand, exporters

> is this thread safe?

Good point, It's just fine when this is not thread safety but it may confuse tsan.I change it to `atomic<bool>` now.

ThomsonTan (approved):

PR #2409, opened 2023-11-16
[FIX] Remove code to reinitialise random numbers on fork()

### LINKED ISSUE #2408: TlsRandomNumberGenerator causes memory corruption when the process is forked.
I manage an open source project https://github.com/hpcc-systems/HPCC-Platform which we adding support to open telemetry using this library.

One of the components launches large numbers (100+) of child processes at the same time.  Since adding the open telemetry library this is now the system to core when those child processes are started.  Here is an example of a stack trace from gdb:

```
#0  0x00007fc40090eb5d in read () from /usr/lib64/libc.so.6
#1  0x00007fc40089ad54 in __GI__IO_file_underflow () from /usr/lib64/libc.so.6
#2  0x00007fc400899808 in __GI__IO_file_xsgetn () from /usr/lib64/libc.so.6
#3  0x00007fc40088e1ef in fread () from /usr/lib64/libc.so.6
#4  0x00007fc4011baa50 in std::random_device::_M_getval() () from /usr/lib64/libstdc++.so.6
#5  0x00007fc40183995b in opentelemetry::v1::sdk::common::(anonymous namespace)::TlsRandomNumberGenerator::Seed() ()
   from /opt/HPCCSystems/lib/libopentelemetry_common.so
#6  0x00007fc4008e4c4e in fork () from /usr/lib64/libc.so.6
#7  0x00007fc4036aeb69 in CLinuxPipeProcess::run (this=0x7fc352cb63a0)
    at /hpcc-dev/HPCC-Platform/system/jlib/jthread.cpp:2072
```

I believe the problem is caused by the following code:

```
  TlsRandomNumberGenerator() noexcept
  {
    Seed();
    platform::AtFork(nullptr, nullptr, OnFork);
  }
```
where an onFork handler was added inside the random number generator made in 2018.

There are only a very restricted set of functions that are valid to be called at that point within the child process (https://man7.org/linux/man-pages/man3/pthread_atfork.3.html): they must be async-signal-safe, and not use any heap functions because that can cause memory corruption.  This onFork call does not obey those restrictions.  It also performs unnecessary work whenever you create a child process  (I suspect it now takes a long time to start the child processes because of the need to wait for sufficient entropy from the random number generator).

I believe the fix is to delete that call to onFork, and re-examine why that change was made.  In particular, why would a process ever call the open telemetry functions inside the child of a fork()?

ghalliday:
Fixes #2408

## Changes
Remove the AtFork hook in the random number generator which potentially causes memory corruptions (as well as being inefficient and unnecessary).

I realise the regression test case will need deleting, but have left that until the change is agreed in principle.


lalitb:
I am not sure if we should remove code to reseed the random number on fork(). The random number generator is used for generating trace_id/span_id.  If the RNG state is duplicated across forked processes, it could affect the randomness of the sampling process, potentially leading to biased sampling results.

Thinking out loud, should we have a thread local flag to indicate if a fork has occurred, and set that in fork handler. Something like this:

```cpp
class TlsRandomNumberGenerator {
  ....
  public: static FastRandomNumberGenerator & engine() noexcept {
    if (forked_) {
      Seed();
      forked_ = false;
    }
    return engine_;
  }

  private: static thread_local bool forked_;

  static void OnFork() noexcept {
    forked_ = true;
  }
}
```

This will add an extra conditional check for each span creation, so that need to be considered too, specifically when most of the applications mayn't be doing any fork.

lalitb:
Not directly related to this PR, but it is also possible to plugin a custom random-id generator if the existing one doesn't suffice -  https://github.com/open-telemetry/opentelemetry-cpp/blob/46e20a42aef521b9cc1b7207310bfefcd490741a/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L48

ghalliday:
Once a process has been forked it is legal to do very little until it calls exec() or similar to switch to a different process.  I don't think it would be legal to call any of the open telemetry functions.   See the documentation in https://man7.org/linux/man-pages/man3/pthread_atfork.3.htmlin the linked issue.

I know this change was made a while back in response to an issue, but I could not work out from the report what the application was aiming to do in the forked() child process.  I suspect the code in the application should have been changed rather than making this change here.

lalitb:
> Once a process has been forked it is legal to do very little until it calls exec() or similar to switch to a different process.

The code I shared is only setting a flag, and actual reseeding is different. We can change it to an atomic flag to be thread-safe. Do you think this is not safe/legal?

> I suspect the code in the application should have been changed rather than making this change here.

Do you mean this issue - https://github.com/opentracing-contrib/nginx-opentracing/issues/52?

ghalliday:
> > Once a process has been forked it is legal to do very little until it calls exec() or similar to switch to a different process.
> 
> The code I shared is only setting a flag in the atfork hook, and actual reseeding is deferred. We can change it to an atomic flag to be thread-safe. Do you think this is not safe/legal?
That change would also avoid the problem - both the time taken and the code - but I cannot think of any reason why any processing is required for the fork.  It also adds an impact as you say to each request.

> > I suspect the code in the application should have been changed rather than making this change here.
> 
> Do you mean changes done for this issue - [opentracing-contrib/nginx-opentracing#52](https://github.com/opentracing-contrib/nginx-opentracing/issues/52)?
Yes.  This was the change.  The documentation is clear that very little can be done inside after a fork() inside the child process.
I have looked at ingress-nginx, and can't find the code that calls fork - so I can't easily see the issue the fix was trying to address. @rnburn are you able to clarify what the issue was, and how this was causing problems with a fork?

ThomsonTan:
If `exec` is not called after the forked, would sharing the random seed from fork causing issue?

> That change would also avoid the problem - both the time taken and the code - but I cannot think of any reason why any processing is required for the fork. It also adds an impact as you say to each request.

owent:
> If `exec` is not called after the forked, would sharing the random seed from fork causing issue?
> 
> > That change would also avoid the problem - both the time taken and the code - but I cannot think of any reason why any processing is required for the fork. It also adds an impact as you say to each request.

It will make generator to generate same random sequence between processes.

lalitb:
> I am not sure if we should remove code to reseed the random number on fork(). The random number generator is used for generating trace_id/span_id. If the RNG state is duplicated across forked processes, it could affect the randomness of the sampling process, potentially leading to biased sampling results.

I believe this PR isn't ready for merging due to the above mentioned concerns. Additionally, the updates suggested in this PR could be implemented through a [custom](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/trace/random_id_generator_factory.h) random number generator. Therefore, closing this PR, but feel free to reopen it with the required modifications.

PR #2415, opened 2023-11-28
enable SSL for curl in bazel build environment

SF-Zhou:


bazel/curl.BUILD
```diff
@@ -55,5 +57,9 @@ cc_library(
             "-lpthread",
         ],
     }),
+    deps = [
+        "@boringssl//:crypto",
+        "@boringssl//:ssl",
+    ],
```
lalitb, line 63: The CMake build has a dependency on OpenSSL. Can we have the same dependency to maintain consistency?

bazel/curl.BUILD
```diff
@@ -34,6 +34,8 @@ cc_library(
     ]),
     copts = CURL_COPTS + [
         "-DOS=\"os\"",
+        "-DCURL_WITH_MULTI_SSL",
```
ThomsonTan, line 37: What is the option `CURL_WITH_MULTI_SSL` for?

bazel/curl.BUILD
owent: Agree, but gRPC also use boringssl, maybe we can use alias to switch the ssl library. (https://bazel.build/docs/configurable-attributes?#faq-select-bind)

marcalff:
The discussion on this code review is inconclusive, as there is no agreement on which SSL library to use to build CURL, and the topic needs further investigation.

Created issue #2489 to document the bug found, and closing this PR.

PR #2341, opened 2023-09-28
[WIP] Synchronous Gauge

lalitb:
Fixes #2279 , #2155 

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #2416, opened 2023-11-29
[Metrics] Make context optional for histogram instruments in Metrics SDK

### LINKED ISSUE #2155: [Metrics API] The Record method of class Histogram now must have a context param, but in the Add method of class Counter it is optional
The Record method of class Histogram now must have a context param, but in the Add method of class Counter it is optional. This brings confusion to users.

ThomsonTan:
Fixes #2155 

## Changes

This PR extracts the context related change in #2341, also the change of making `LongCounter` and `LongHistogram` as unsigned type is picked up.


lalitb (approved):
Thanks for the fix. Nicely done.

marcalff (approved):
LGTM, thanks for the fix.

PR #2420, opened 2023-11-30
[BUILD] Remove defining NOMINMAX from api

ThomsonTan:
Currently, including OpenTelemetry API requires NOMINMAX to be defined for Windows user which excludes the min/max macro from the Windows header files. There are cases when both macros are used by the customer code so requiring NOMINMAX will cause incompatibility. This PR wraps the usage of min/max from C++ standard library with parenthesis. This can avoid the naming conflict with the macro names min/max.

## Changes

### DIFF
--- a/api/CMakeLists.txt
+++ b/api/CMakeLists.txt
@@ -104,7 +104,6 @@ if(WITH_NO_GETENV)
 endif()
 
 if(WIN32)
-  target_compile_definitions(opentelemetry_api INTERFACE NOMINMAX)
   if(WITH_ETW)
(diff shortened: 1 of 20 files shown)


lalitb (approved):

marcalff (approved):
For the changes in API, totally agree, this is the most robust thing to do.

For the changes in exporters and SDK, I think the changes are not really needed, especially in `*.cc` because the `Windows.h` header file is not used and will not cause conflicts, but ok to have the extra parenthesis to have a consistent code.

Even when not needed, this will prevent two different styles to proliferate in the code base with copy and paste.

Approved, thanks for the fix.

PR #2425, opened 2023-12-02
Use std shared_ptr, unique_ptr and enable_if

### LINKED ISSUE #1877: Inconsistency between the use of std::shared_ptr and nostd::shared_ptr
**Describe your environment** 
Windows, MSVC, Using 1.8.1 tag

**Steps to reproduce**

More consitent use of either std::shared_ptr or nostd::shared_ptr (e.g. same way of using it with respect to api and sdk/etc.)

Inconsistency where one class might return/deal with std:;shared_ptr along with nostd::shared_ptr

**Additional context**
Not much, and not a blocker at all. I'm working on Windows C++ dll version and re-exposing some of (what I think) are missing factories (PeriodicExportingMetricFactory, MeterProviderFactory) such that I can have single "otel.dll" that is safe to call and work with.

Please compare this:
https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.8.1/sdk/include/opentelemetry/sdk/metrics/exemplar/filter.h#L35

With:
https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.8.1/sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir.h#L53

(and other examples)

Also I'm compiling/forcing HAVE_CPP_STDLIB so for me it doesn't matter much, but in general might be good to have consitent way of which one should be used or not (TBH, I can't fully comprehend all edge cases here, so this might've been a valid choice after all)

bogdandrutu:
Fixes #1877

### DIFF
--- a/sdk/include/opentelemetry/sdk/trace/tracer.h
+++ b/sdk/include/opentelemetry/sdk/trace/tracer.h
@@ -6,7 +6,6 @@
 #include <memory>
 
 #include "opentelemetry/common/macros.h"
-#include "opentelemetry/nostd/shared_ptr.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/instrumentationscope/instrumentation_scope.h"
 #include "opentelemetry/sdk/trace/samplers/always_on.h"
@@ -38,7 +37,7 @@ public:
                   std::unique_ptr<InstrumentationScope> instrumentation_scope =
                       InstrumentationScope::Create("")) noexcept;
 
-  nostd::shared_ptr<opentelemetry::trace::Span> StartSpan(
+  std::shared_ptr<opentelemetry::trace::Span> StartSpan(
       nostd::string_view name,
       const opentelemetry::common::KeyValueIterable &attributes,
       const opentelemetry::trace::SpanContextKeyValueIterable &links,
--- a/sdk/include/opentelemetry/sdk/common/global_log_handler.h
+++ b/sdk/include/opentelemetry/sdk/common/global_log_handler.h
@@ -3,10 +3,10 @@
 
 #pragma once
 
+#include <memory>
 #include <sstream>
 #include <utility>
 
-#include "opentelemetry/nostd/shared_ptr.h"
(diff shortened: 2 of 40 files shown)


sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -38,7 +37,7 @@ class Tracer final : public opentelemetry::trace::Tracer,
                   std::unique_ptr<InstrumentationScope> instrumentation_scope =
                       InstrumentationScope::Create("")) noexcept;
 
-  nostd::shared_ptr<opentelemetry::trace::Span> StartSpan(
+  std::shared_ptr<opentelemetry::trace::Span> StartSpan(
```
lalitb, line 40: I believe the correct approach is to use `nostd::` constructs in the SDK when implementing the Otel API interfaces, as specified in their declarations.

lalitb (changes requested):
I believe the correct approach is to use `nostd::` constructs in the SDK when implementing the Otel API interfaces, as specified in their declarations.

marcalff (changes requested):
As noted already by @lalitb,
when a virtual method returns nostd::unique_ptr<T> in the API,
the SDK implementation should also use nostd: the SDK can not change the return type, this is what is causing build failures currently, for example for GetMeter().

For code which is purely internal, and for code exposed only by the SDK or exporters, like global_log_handler.h for example, I think it is ok to use std classes, however there is this issue:

When designing `std::unique_ptr<T> do_something_internal()`,
it is very challenging to know, in advance, if the result of this call will eventually (or never) by returned all the way up to the API, which requires to either convert std to nostd, or change the code to use nostd in do_something_internal().

On a totally different topic, `std::enable_if` is about meta-programming, and never shows up in the ABI.

I think it should be safe to replace `nostd::enable_if` with `std::enable_if`, independently of what we decide for `unique_ptr` and `shared_ptr`.

According to https://en.cppreference.com/w/cpp/types/enable_if:
* `std::enable_if` is available since C++11
* `std::enable_if_t` is available since C++14

and we now require C++14,
so both should be usable.

PR #2427, opened 2023-12-04
[API] Create root span with active span

### LINKED ISSUE #2413: Ability to create a root span (span with no parent) when another span is active.
As discussed here - https://github.com/open-telemetry/opentelemetry-cpp/discussions/2410, creating the issue to evaluate/discuss support for creating a root span when another span is active. 

```cpp

{
    auto root1 = tracer->StartSpan("span 1");
    auto root2 = trace->StartSpan("span 2");

    auto token = tracer->WithActiveSpan(root1); // root 1 active
    auto child_root1 = tracer->StartSpan("child of root1"); // child of root1

    trace_api::StartSpanOptions options;
    options.parent = root2;
    auto child_root2 = tracer->StartSpan("child of root2", options); // child of root2
  
    // trying to create a span with no parent now.
    options.parent = SpanContext(false, false);
    auto root3 = tracer->StartSpan("root3", options); // root3 is child of root1 which is not desired
   
    opentelemetry::context::Context c1;
    options.parent = c1;
    root3 = tracer->StartSpan("root3", options); // root3 is child of root1 which is not desired

    trace_api::StartSpanOptions options2;
    root3 = tracer->StartSpan("root3", options2); // root3 is child of root1 which is not desired

}
```

When one can create a span that is a child to a non-active span, should it also be possible to create a span with no parent?

lalitb:
Fixes #2413 

## Changes

A safe ABIv1 approach (non-ABI/non-API breaking)  I can think of - passing Context with ("is_root_span", true) value. 

### DIFF
--- a/sdk/test/trace/tracer_test.cc
+++ b/sdk/test/trace/tracer_test.cc
@@ -1014,7 +1014,7 @@ TEST(Tracer, WithActiveSpan)
     {
       trace_api::StartSpanOptions options;
       opentelemetry::context::Context c1;
-      c1             = c1.SetValue("is_root_span", true);
+      c1             = c1.SetValue(opentelemetry::trace::kIsRootSpanKey, true);
       options.parent = c1;
       auto root_span = tracer->StartSpan("span root", options);
 
--- a/sdk/test/trace/tracer_test.cc
+++ b/sdk/test/trace/tracer_test.cc
@@ -1008,6 +1008,26 @@ TEST(Tracer, WithActiveSpan)
     spans = span_data->GetSpans();
     ASSERT_EQ(1, spans.size());
     EXPECT_EQ("span 2", spans.at(0)->GetName());
+    EXPECT_EQ(spans.at(0).get()->GetParentSpanId(), span_first->GetContext().span_id());
+    EXPECT_EQ(spans.at(0).get()->GetTraceId(), span_first->GetContext().trace_id());
+
+    {
+      trace_api::StartSpanOptions options;
+      opentelemetry::context::Context c1;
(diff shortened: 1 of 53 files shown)


sdk/test/trace/tracer_test.cc
```diff
@@ -1008,6 +1008,26 @@ TEST(Tracer, WithActiveSpan)
...
     EXPECT_EQ("span 2", spans.at(0)->GetName());
+    EXPECT_EQ(spans.at(0).get()->GetParentSpanId(), span_first->GetContext().span_id());
+    EXPECT_EQ(spans.at(0).get()->GetTraceId(), span_first->GetContext().trace_id());
+
+    {
+      trace_api::StartSpanOptions options;
+      opentelemetry::context::Context c1;
+      c1             = c1.SetValue("is_root_span", true);
+      options.parent = c1;
+      auto root_span = tracer->StartSpan("span root", options);
```
marcalff, line 1019: Suggestion:

Add this sequence of code as verbatim example of how to start a root span,
in comments in span_start_options.h

```
trace_api::StartSpanOptions options;
opentelemetry::context::Context root;
root                    = root.SetValue("is_root_span", true);
options.parent = root;
auto root_span = tracer->StartSpan("span root", options);
```

It is not trivial for someone who is not familiar with the API,
so providing code ready to copy and paste will help.

marcalff (commented):
LGTM, just one suggestion.

Nice trick to have the fix in ABIv1.

marcalff (approved):

sdk/test/trace/tracer_test.cc
lalitb: Good point. Done.

PR #2428, opened 2023-12-04
[SEMANTIC CONVENTIONS] Upgrade to semconv version 1.23.1

### LINKED ISSUE #2419: [SEMANTIC CONVENTIONS] Upgrade to semconv version 1.23.1
https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.23.1

marcalff:
Fixes #2419

## Changes

Upgrade to:

* semantic conventions 1.23.1
* build-tools 0.23.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ Increment the:
   [#2398](https://github.com/open-telemetry/opentelemetry-cpp/pull/2398)
 * [EXPORTER] Rework OTLP/HTTP and OTLP/GRPC exporter options
   [#2388](https://github.com/open-telemetry/opentelemetry-cpp/pull/2388)
+* [SEMANTIC CONVENTION] Upgrade to semconv 1.23.1
+  [#2428](https://github.com/open-telemetry/opentelemetry-cpp/pull/2428)
 
 Important changes:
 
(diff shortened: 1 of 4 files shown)


lalitb (approved):
Thanks. Difficult to identify the difference due to the change in the order of definitions. Hopefully, the generate script should be good to rely on. Else, let me know if something needs to be reviewed specifically.

marcalff:
> Thanks. Difficult to identify the difference due to the change in the order of definitions. Hopefully, the generate script should be good to rely on. Else, let me know if something needs to be reviewed specifically.

Yes, relying on the generated code too.

I think the build-tools can be improved to may be sort all symbols first, to generate them in a more predictable and stable order, to limit spurious diffs ... I will raise that with build-tools.

I use a smoke test to check the diff sanity, looked ok:

```
git diff | grep static | sort -u
```

it will list a lot of semconv with +1 and -1 lines for the same name, indicating the declaration moved but was unchanged.

lalitb:
> I think the build-tools can be improved to may be sort all symbols first, to generate them in a more predictable and stable order, to limit spurious diffs ... I will raise that with build-tools.


Agree, a more predictable order would help.

PR #2430, opened 2023-12-05
[BUILD] Improve the handling of OPENTELEMETRY_HAVE_WORKING_REGEX.

kylepl:
It used to be some places were checking for
defined(OPENTELEMETRY_HAVE_WORKING_REGEX), but it is always being set (to either 0 or 1) in api/include/opentelemetry/common/macros.h.

Thus move to just checking its value everywhere.

Additionally, even if we did have the feature off - compilation would fail, since not all the functions were ifdef'd. I assume this fails since there are not tests that have OPENTELEMETRY_HAVE_WORKING_REGEX as off.

My initial motivation for this was to reduce compile times, since specificing the definition with an std::regex adds a significant penalty to each translation unit (~700ms on my 5-year old laptop). In a separate patch, I think this should be moved to a .cc file - happy to receive advice on how (since api/ seems not to have .cc's).

### DIFF
--- a/api/include/opentelemetry/trace/trace_state.h
+++ b/api/include/opentelemetry/trace/trace_state.h
@@ -14,7 +14,7 @@
 #include "opentelemetry/nostd/unique_ptr.h"
 #include "opentelemetry/version.h"
 
-#if defined(OPENTELEMETRY_HAVE_WORKING_REGEX)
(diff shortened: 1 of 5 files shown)


lalitb (approved):
Good catch, thanks for the fix.

ThomsonTan (approved):

marcalff (approved):
Thanks for the fix.

PR #2451, opened 2023-12-13
[EXPORTER] Add option to disable Prometheus otel_scope_name and otel_scope_version attributes

### LINKED ISSUE #2442: [Prometheus Exporter] Unable to disable `otel_scope_name` attribute
According to the spec for the Prometheus exporter: 

> Prometheus exporters SHOULD provide a configuration option to disable the otel_scope_info metric and otel_scope_ labels.

I can't find any way to actually disable these. The prometheus exporter would disable them assuming the strings passed in via the scope data are empty. The data in the `InstrumentationScope` that gets passed into the exporter comes from agruments passed to `GetMeter()`. Passing a blank string for `library_name` would ostensibly disable it based on the code in the exporter, but passing a blank there results in an error:

```
[Warning] File: <>/auxil/opentelemetry-cpp/sdk/src/metrics/meter_provider.cc:54 [MeterProvider::GetMeter] Library name is empty.
```

timwoj:
Fixes #2442 (plus one bonus fix)

## Changes

- Adds a new `populate_otel_scope` option to the `PrometheusExporterOptions` object defaulting to enabled, including an environment variable to control it
- Uses the option to enable/disable the `otel_scope_name` and `otel_scope_version` attributes added to instruments in Prometheus output
- Adds a new environment variable for controlling the `populate_target_info` option of `PrometheusExporterOptions`

I opted to use (abuse) the side effect of the scope variable being null for the actual disabling of the attributes here, instead of passing the option down through the method chain to where it's used. If that's undesirable, I can switch to passing it.

### DIFF
--- a/exporters/prometheus/src/exporter_options.cc
+++ b/exporters/prometheus/src/exporter_options.cc
@@ -25,7 +25,20 @@ inline const std::string GetPrometheusDefaultHttpEndpoint()
   return exists ? endpoint : kPrometheusEndpointDefault;
 }
 
-PrometheusExporterOptions::PrometheusExporterOptions() : url(GetPrometheusDefaultHttpEndpoint()) {}
+inline bool GetPrometheusPopulateOtelScope()
+{
+  constexpr char kPrometheusPopulateOtelScope[] = "PROMETHEUS_EXPORTER_POPULATE_OTEL_SCOPE";
+
+  bool setting;
+  auto exists =
+      opentelemetry::sdk::common::GetBoolEnvironmentVariable(kPrometheusPopulateOtelScope, setting);
+
+  return exists ? setting : true;
+}
+
+PrometheusExporterOptions::PrometheusExporterOptions()
+    : url(GetPrometheusDefaultHttpEndpoint()), populate_otel_scope(GetPrometheusPopulateOtelScope())
+{}
 
 }  // namespace metrics
 }  // namespace exporter
--- a/exporters/prometheus/src/exporter_options.cc
+++ b/exporters/prometheus/src/exporter_options.cc
@@ -36,8 +36,21 @@ inline bool GetPrometheusPopulateOtelScope()
   return exists ? setting : true;
 }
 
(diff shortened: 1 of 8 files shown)


exporters/prometheus/src/exporter_options.cc
```diff
@@ -25,7 +25,33 @@ inline const std::string GetPrometheusDefaultHttpEndpoint()
   return exists ? endpoint : kPrometheusEndpointDefault;
 }
 
-PrometheusExporterOptions::PrometheusExporterOptions() : url(GetPrometheusDefaultHttpEndpoint()) {}
+inline bool GetPrometheusPopulateOtelScope()
+{
+  constexpr char kPrometheusPopulateOtelScope[] = "PROMETHEUS_EXPORTER_POPULATE_OTEL_SCOPE";
```
lalitb, line 30: nit - should we prefix both the env variables with `OTEL_`, the naming used general in the OpenTelemetry.

lalitb (approved):
LGTM. Thanks for the fix.

exporters/prometheus/src/exporter_options.cc
marcalff: The environment variable is not defined in the spec, so it becomes language specific.

Please add a `OTEL_CPP_` prefix for both new variables, to comply with:

* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#language-specific-environment-variables

marcalff (approved):
Thanks for the fix.

LGTM, see minor renaming.

marcalff:
@timwoj 

Thanks for the fixes.

Please update this branch with the latest main changes.

The auto-merge flag is enabled, this PR should merge to main once updated.

marcalff:
Reopen.

The collide with the following spec PR, which defines the option and environment variable to use:

*  https://github.com/open-telemetry/opentelemetry-specification/pull/3796

PR #2457, opened 2023-12-15
[BUILD] Skip patch alias target

### LINKED ISSUE #2446: CMake error when building with CURL 8.3.0 and higher

Ubuntu 20.04
CMake 3.18
Opentelemetry C++ 1.13.0
GCC 7.5.0
CURL 8.3.0

**Steps to reproduce**
Compile CURL 8.3.0 or higher. Then try to compile OpenTelemetry C++ with OTLP enabled (so that CURL is actually needed).

The code should compile without errors.

This is the error that CMake gives:
```
CMake Error at cmake/tools.cmake:186 (set_target_properties):
  set_target_properties can not be used on an ALIAS target.
Call Stack (most recent call first):
  cmake/patch-imported-config.cmake:25 (project_build_tools_patch_default_imported_config)
  CMakeLists.txt:685 (include)


CMake Error at cmake/tools.cmake:186 (set_target_properties):
  set_target_properties can not be used on an ALIAS target.
Call Stack (most recent call first):
  cmake/patch-imported-config.cmake:25 (project_build_tools_patch_default_imported_config)
  CMakeLists.txt:685 (include)
```

**Additional context**
I was able to find the exact commit where the build breaks: https://github.com/curl/curl/commit/1199308dbc902c52be67fc805c72dd2582520d30

My intuition tells me that, by making the target `CURL::libcurl` an ALIAS, the CMake function `project_build_tools_patch_default_imported_config` in `cmake/patch-imported-config.cmake` doesn't work anymore, as it is trying to modify a target that is an ALIAS.

owent:
Fixes #2446 

## Changes

+ Skip patch alias target of cmake.


marcalff (approved):
LGTM, thanks for the fix.

lalitb (approved):

PR #2471, opened 2023-12-27
[WIP, DEBUGGING] Fix common_foo_library when building with OPENTELEMETRY_BUILD_DLL

meastp:
We keep having build failures when trying to use OPENTELEMETRY_BUILD_DLL, see #2202 for background.

UPDATE: still trying to reproduce the build failure in CI. no luck so far. My first theory didn't work.


marcalff:
@meastp 

If I understand correctly, this PR was to reproduce the problem from #2202 in CI.

Now that the root cause is identified (thanks), can this PR be closed ?

Thanks for all the investigations.

PR #2478, opened 2024-01-06
Set is_monotonic flag for Observable Counters

### LINKED ISSUE #2469: OTLP Exporter does not set an observable counter as monotonic
Recently upgraded to Opentelemetry-cpp v1.12.0 and noticed some difference in how the OTLP Exporter exports Counters/UpDownCounters.  In the previous version (v1.9.0) I was using, all observable Counters/UpDownCounters were exported with the `is_monotonic` flag set to `true`. As a result, the Prometheus exporter in the Opentelemetry Collector Contrib mapped the OTEL Counters/UpDownCounters to Prometheus Counters. Issue #2170 changed the behavior to only set the monotonic flag for `kCounter`. As a result, in v1.12.0, only the synchronous Counter is mapped to the Prometheus Counter type by the Collector; synchronous UpDownCounter and observable Counter/UpDownCounters are mapped to Prometheus Gauge type.

My question is, shouldn't observable Counters be mapped to Prometheus Counter type? For that to happen, we would need to set the `is_monotonic` flag for both `kCounter` and `kObservableCounter`. According to the official [spec](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#sums), if the aggregation temporality is cumulative and the sum is monotonic, it MUST be converted to a Prometheus Counter.

Also, for reference, seems like the Python SDK exports observable Counters with the `is_monotonic` flag set as the Prometheus endpoint shows the instrument type as a Counter instead of a Gauge.

samin36:
Fixes #2469 

## Changes

The `is_monotonic` flag is now set to `true` for both synchronous and asynchronous/observable counters. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Increment the:
   [#2449](https://github.com/open-telemetry/opentelemetry-cpp/pull/2449)
 * [BUILD] Introduce CXX 20 CI pipeline for MSVC/Windows
   [#2450](https://github.com/open-telemetry/opentelemetry-cpp/pull/2450)
+* [EXPORTER] Set `is_monotonic` flag for Observable Counters
+  [#2478](https://github.com/open-telemetry/opentelemetry-cpp/pull/2478)
 
 Important changes:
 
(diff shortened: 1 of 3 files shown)


lalitb (approved):
LGTM. Thanks for the fix.

marcalff (approved):
LGTM, thanks for the fix.

PR #2493, opened 2024-01-18
[BUILD] Fix checks on __cplusplus under MSVC, do not assume /Zc

bcsgh:
- https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus
- The change to `.../absl/base/policy_checks.h` is a back port from the same lines in the current Abseil `HEAD`
- I _think_ the rest of the uses of `__cplusplus` are correct under MSVC, but I'm not totally sure on that.


marcalff (approved):
LGTM, thanks for the fix.

lalitb (approved):
Thanks.

PR #2496, opened 2024-01-19
[CI] Make cmake.c++20*.test actualy use C++20 and add cmake.c++23*.test

bcsgh:
Part of #2470

## Changes

The existing `cmake.c++20*.test` cases don't actually pin C++2020 but either don't use `WITH_STL=` at all or set it to `ON`. This results in a strange mix of "version" between the name of the test, the version the library _thinks_ is being used and what is actually being used by the build

Also, there was no explicit C++2023 tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -360,7 +360,7 @@ jobs:
 
   cmake_test_cxx23_gcc:
     name: CMake C++23 test(GCC)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -358,6 +358,59 @@ jobs:
         CXX_STANDARD: '20'
       run: ./ci/do_ci.sh cmake.c++20.stl.test
 
+  cmake_test_cxx23_gcc:
+    name: CMake C++23 test(GCC)
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+    - name: setup
+      env:
+        CMAKE_VERSION: 3.20.6
+      run: |
+        sudo -E ./ci/setup_ci_environment.sh
+        sudo -E ./ci/setup_cmake.sh
+    - name: run tests
+      env:
+        CXX_STANDARD: '23'
+      run: ./ci/do_ci.sh cmake.c++23.test
+    - name: run tests (enable stl)
+      env:
+        CXX_STANDARD: '23'
+      run: ./ci/do_ci.sh cmake.c++23.stl.test
+
+  cmake_test_cxx23_clang:
+    name: CMake C++23 test(Clang with libc++)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: 'recursive'
+    - name: setup
+      env:
+        CC: /usr/bin/clang
+        CXX: /usr/bin/clang++
+        CXXFLAGS: "-stdlib=libc++"
+        CMAKE_VERSION: 3.20.6
+      run: |
+        sudo -E ./ci/setup_ci_environment.sh
+        sudo -E ./ci/setup_cmake.sh
+    - name: run tests
+      env:
+        CC: /usr/bin/clang
+        CXX: /usr/bin/clang++
+        CXXFLAGS: "-stdlib=libc++"
+        CXX_STANDARD: '23'
+      run: ./ci/do_ci.sh cmake.c++23.test
+    - name: run tests (enable stl)
+      env:
+        CC: /usr/bin/clang
+        CXX: /usr/bin/clang++
+        CXXFLAGS: "-stdlib=libc++"
+        CXX_STANDARD: '23'
+      run: ./ci/do_ci.sh cmake.c++23.stl.test
+
   cmake_otprotocol_test:
     name: CMake test (with otlp-exporter)
     runs-on: ubuntu-20.04
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -229,6 +229,19 @@ elif [[ "$1" == "cmake.c++20.test" ]]; then
   cmake ${CMAKE_OPTIONS[@]}  \
         -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
         -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+        -DWITH_STL=CXX20 \
+        ${IWYU} \
+        "${SRC_DIR}"
+  eval "$MAKE_COMMAND"
+  make test
+  exit 0
+elif [[ "$1" == "cmake.c++23.test" ]]; then
+  cd "${BUILD_DIR}"
+  rm -rf *
+  cmake ${CMAKE_OPTIONS[@]}  \
+        -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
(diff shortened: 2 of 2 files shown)


bcsgh:
BTW: I make no claim to actually knowing how this works, so please don't just assume it's correct!

ci/do_ci.sh
```diff
@@ -229,6 +229,19 @@ elif [[ "$1" == "cmake.c++20.test" ]]; then
...
+  eval "$MAKE_COMMAND"
+  make test
+  exit 0
+elif [[ "$1" == "cmake.c++23.test" ]]; then
+  cd "${BUILD_DIR}"
+  rm -rf *
+  cmake ${CMAKE_OPTIONS[@]}  \
+        -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
+        -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+        -DWITH_STL=CXX23 \
```
owent, line 244: [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) is a target property, not a environment variable in cmake.We can add `-DCMAKE_CXX_STANDARD=23` here to tell cmake to use c++23 by default.

And we need to upgrade cmake to 3.20 or upper to use C++23.The `./ci/setup_cmake.sh` use 3.16 now.

ci/do_ci.sh
```diff
@@ -267,7 +280,20 @@ elif [[ "$1" == "cmake.c++20.stl.test" ]]; then
...
+  make test
+  exit 0
+elif [[ "$1" == "cmake.c++23.stl.test" ]]; then
+  cd "${BUILD_DIR}"
+  rm -rf *
+  cmake ${CMAKE_OPTIONS[@]}  \
+        -DWITH_METRICS_EXEMPLAR_PREVIEW=ON \
+        -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
+        -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+        -DWITH_STL=CXX23 \
```
owent, line 296: [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) is a target property, not a environment variable in cmake.We can add `-DCMAKE_CXX_STANDARD=23` here to tell cmake to use c++23 by default.

And we need to upgrade cmake to 3.20 or upper to use C++23.The `./ci/setup_cmake.sh` use 3.16 now.

ci/do_ci.sh
```diff
@@ -229,6 +229,19 @@ elif [[ "$1" == "cmake.c++20.test" ]]; then
   cmake ${CMAKE_OPTIONS[@]}  \
         -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
         -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+        -DWITH_STL=CXX20 \
```
owent, line 232: [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) is a target property, not a environment variable in cmake.We can add `-DCMAKE_CXX_STANDARD=20` here to tell cmake to use c++20 by default.

ci/do_ci.sh
```diff
@@ -267,7 +280,20 @@ elif [[ "$1" == "cmake.c++20.stl.test" ]]; then
         -DWITH_METRICS_EXEMPLAR_PREVIEW=ON \
         -DCMAKE_CXX_FLAGS="-Werror $CXXFLAGS" \
         -DWITH_ASYNC_EXPORT_PREVIEW=ON \
-        -DWITH_STL=ON \
+        -DWITH_STL=CXX20 \
```
owent, line 283: [`CXX_STANDARD`](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) is a target property, not a environment variable in cmake.We can add `-DCMAKE_CXX_STANDARD=20` here to tell cmake to use c++20 by default.

ci/do_ci.sh
bcsgh: Are you saying the existing usage of `CXX_STANDARD` in the other (YAML) file is not correct? Or is there something that makes that prior usage there correct but the new usage there incorrect?

(I'm slightly confused because this comment thread seems to have nothing to do with the location it's attached to.)

ci/do_ci.sh
bcsgh: [See other reply.]

ci/do_ci.sh
bcsgh: The point about `CXX_STANRDARD` seems to be redundant to the other comment.

As for upgrading CMake; how about I just comment out the C++23 section in the YAML file with a new referencing an issue that that as a TODO? I don't need C++23, I just added it out of completeness.

ci/do_ci.sh
bcsgh: [See other reply.]

ci/do_ci.sh
bcsgh: On [further inspection](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/do_ci.sh#L87-L91) it looks like the CI script promotes the `CXX_STANDARD` environment variable to the flag `-DCMAKE_CXX_STANDARD=...`. It seems like the existing usage (and by extension, this new usage) is already correct.

ci/do_ci.sh
owent: Sorry, I forgot we just set `CMAKE_CXX_STANDARD` by `CXX_STANRDARD` in `do_ci.sh` and also upgrade cmake by `CMAKE_VERSION`.Please just ignore my comment.

ci/do_ci.sh
owent: Sorry, I forgot we just set `CMAKE_CXX_STANDARD` by `CXX_STANRDARD` in `do_ci.sh` and also upgrade cmake by `CMAKE_VERSION`.Please just ignore my comment.

owent (approved):
LGTM and thanks.

.github/workflows/ci.yml
```diff
@@ -358,6 +358,59 @@ jobs:
         CXX_STANDARD: '20'
       run: ./ci/do_ci.sh cmake.c++20.stl.test
 
+  cmake_test_cxx23_gcc:
+    name: CMake C++23 test(GCC)
+    runs-on: ubuntu-20.04
```
lalitb, line 363: why not to use `ubuntu-latest`?

bcsgh: Copy/paste. I think from the c++20 version.

I'll go with whatever you suggest. (Fell free to send an edit. I'm OOO right now.)

marcalff: Using a mix of different OS releases improves coverage. To get a recent C++23 compiler, a more recent version of the distribution is most likely needed, so using ubuntu-latest seems better.

bcsgh: [Pushed.](https://github.com/open-telemetry/opentelemetry-cpp/pull/2496/commits/8c060284d52477015ce05e7cafca2bdc7d96d2c9)

marcalff: Thanks.

https://github.com/actions/runner-images/blob/ubuntu22/20240116.3/images/ubuntu/Ubuntu2204-Readme.md

has clang-15 and gcc-13

https://github.com/actions/runner-images/blob/ubuntu22/20240116.3/images/ubuntu/Ubuntu2004-Readme.md

has clang-12 and gcc-10

so this should be better.

Resolved.

marcalff (approved):
LGTM, thanks for the cleanup.

lalitb (approved):
Thanks

PR #2503, opened 2024-01-22
[BUILD] Better handling of OPENTELEMETRY_STL_VERSION under Bazel.

### LINKED ISSUE #2470: Provide a way to populate `OPENTELEMETRY_STL_VERSION` from `__cplusplus`.

bcsgh:
Fixes #2470 for Bazel.

## Changes

Make Bazel build, by default, use the "current" actual C++ version as `OPENTELEMETRY_STL_VERSION`.

Note that some compilers use some "interesting" values for `__cplusplus` when asked to compiler for the most recent version they support. This can result in slightly unexpected results, but might actually be a good thing if those values indicate less than perfect support for the version in question and cause a downgrade to a better supported library.

### DIFF
--- a/api/BUILD
+++ b/api/BUILD
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
 
 package(default_visibility = ["//visibility:public"])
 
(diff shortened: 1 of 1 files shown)


esigo (approved):
LGTM
Thanks for the PR :)

marcalff (approved):
Thanks.

PR #2507, opened 2024-01-23
[EXPORTER] Error when grpc endpoint is empty

HennerM:
## Changes

It seems like there is nothing checking whether the grpc exporter is actually empty, in which case it doesn't make sense to create a gRPC channel. This causes a problem in my application where we accidentally create a GrpcExporter with empty endpoint, which we never use because of a NOOP sampler. 

I believe it could be useful to log an error here, similar to when the endpoint is no valid URI.

### DIFF
--- a/exporters/otlp/src/otlp_grpc_client.cc
+++ b/exporters/otlp/src/otlp_grpc_client.cc
@@ -264,8 +264,13 @@ OtlpGrpcClient::~OtlpGrpcClient()
 
 std::shared_ptr<grpc::Channel> OtlpGrpcClient::MakeChannel(const OtlpGrpcClientOptions &options)
 {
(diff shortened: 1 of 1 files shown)


marcalff (approved):
Thanks.

HennerM:
Thanks @marcalff , I am just getting the CLA for my organisation through, could take a day or so

owent (approved):

marcalff:
Hi @HennerM .

Any update on the CLA ?

HennerM:
> Hi @HennerM .
> 
> Any update on the CLA ?

yes unfortunately the EasyCLA landed in the spam folder of the responsible person in my organisation, but it should get through soon now

PR #2511, opened 2024-01-25
[API] Propagators: do not overwrite the active span with a default invalid span

### LINKED ISSUE #2504: Composite propagation doesn't work because the last propagator sets the span, potentially invalid, even if the previous one finds a valid span
**Describe your environment** 
v1.13.0 using bazel

**Steps to reproduce**
Remove this line from the unit test:
https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/api/test/context/propagation/composite_propagator_test.cc#L80
And you will see that the resulting traceId is 0000000..., whereas it should be the one from the valid traceparent header line 79.

The composite propagator should only set the active_span in the context to a valid one

The composite propagator sets the active_span at each propagator in its list, whatever the outcome, so the last one always wins, even if invalid.
You can see here: https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/api/include/opentelemetry/trace/propagation/b3_propagator.h#L54
and here: https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/api/include/opentelemetry/trace/propagation/http_trace_context.h#L57

**Additional context**
I see 3 ways of fixing it.  
* disallow setting invalid spans with SetSpan? But then it might affect many parts of the code?
* CompositePropagator to do the filtering? Sounds complex because CompositePropagator does not know what is written to the context by the Extractors.
* I think the best and pretty simple is in **all** propagators to replace the line that does `return trace::SetSpan(context, sp);` with these:
```cpp
    if (span_context.IsValid()) 
    {
      return trace::SetSpan(context, sp);
    } else {
      return context;
    }
```

If you agree, I will post a PR with that.

ecourreges-orange:
[API] Fix b3, w3c and jaeger propagators: they will not overwrite the active span with a default invalid span
This is especially useful when used with CompositePropagator (#2504)

[TEST] Change wrong string "current-span" in unit tests to proper trace:kSpan which is "active_span"

Fixes #2504

## Changes
This changes how Extract returns when invalid or missing headers are provided: it now returns the original context with potentially no active_span (or not overwritten if one was present).  
But this should not have side effects on client code which calls GetSpan(context), which answers a default invalid span when it does not find an active_span

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ Increment the:
 
 ## [Unreleased]
 
+* [API] Fix b3, w3c and jaeger propagators: they will not overwrite
+  the active span with a default invalid span, which is especially useful
+  when used with CompositePropagator
+  [TEST] fix string "current-span" to trace:kSpan which is "active_span"
+  [#2511](https://github.com/open-telemetry/opentelemetry-cpp/pull/2511)
 * [REMOVAL] Remove option WITH_OTLP_HTTP_SSL_PREVIEW
   [#2435](https://github.com/open-telemetry/opentelemetry-cpp/pull/2435)
 * [BUILD] Fix removing of NOMINMAX on Windows
--- a/api/include/opentelemetry/trace/propagation/b3_propagator.h
+++ b/api/include/opentelemetry/trace/propagation/b3_propagator.h
@@ -51,7 +51,14 @@ public:
   {
     SpanContext span_context = ExtractImpl(carrier);
     nostd::shared_ptr<Span> sp{new DefaultSpan(span_context)};
-    return trace::SetSpan(context, sp);
+    if (span_context.IsValid())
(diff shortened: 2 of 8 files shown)


marcalff:
Bonjour @ecourreges-orange 

Thanks for the PR, CI build in progress.

ecourreges-orange:
> Bonjour @ecourreges-orange
> 
> Thanks for the PR, CI build in progress.

You're welcome, happy to be contributing!

marcalff (approved):
LGTM, thanks for everything:

* the initial report
* the analysis
* the fix
* the unit tests

lalitb (approved):
LGTM. Thanks for the fix.

PR #2518, opened 2024-01-31
[IGNORE ME] Poc config yaml

marcalff:
Fixes # (issue)

## Changes


examples/yaml/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+if(WITH_OTLP_HTTP)
+  include_directories(${CMAKE_SOURCE_DIR}/exporters/otlp/include)
+  add_definitions(-DOTEL_HAVE_OTLP_HTTP)
+  target_link_libraries(example_yaml opentelemetry_otlp_http_builder)
+endif()
+
+if(WITH_OTLP_GRPC)
+  include_directories(${CMAKE_SOURCE_DIR}/exporters/otlp/include)
+  add_definitions(-DOTEL_HAVE_OTLP_GRPC)
+  target_link_libraries(example_yaml opentelemetry_otlp_grpc_builder)
```
dbarker, line 38: Hi @marcalff, I took a look today and wanted to share some thoughts and initial observations of the cmake changes to help spark some discussion about how to fit this pr into the new cmake components introduced by #3220: 

**Initial Observations:** 
This PR introduces some new cmake targets. These targets include the`opentelemetry-cpp::configuration` target from `sdk/src/configuration` and `opentelemetry-cpp::init` target from `sdk/src/init` as well as `opentelemetry-cpp::*_builder` targets  (ie: opentelemetry-cpp::otlp_grpc_builder) created in the respective exporter directories.  

The `example_yaml` shows how the new targets must be linked to an executable independently (ie: the `opentelemetry-cpp::init` target and the `opentelemetry-cpp::*_builder` targets must be explicitly linked to the example).   

This PR also introduces the `WITH_CONFIGURATION` cmake option which determines if the `opentelemetry-cpp::configuration` and `opentelemetry-cpp::init` targets are created along with the main example and tests. The `opentelemetry-cpp::*_builder` targets are created independent of this option and are created if the corresponding `WITH_*` option is on (ie: `WITH_OTLP_GRPC=ON` creates the `opentelemetry-cpp::otlp_grpc_builder` target). 

**Summary of the cmake component approach so far:** 
In the new cmake component approach from #3220, each optional component is tied to a `WITH_*` option and has a fixed/required set of cmake targets defined by `cmake/component_definitions.cmake`. This creates a strong guarantee that if a component is successfully found with `find_package(opentelemetry-cpp COMPONENTS <component>)`, then all cmake targets associated with the component have been imported and can be linked. This enables failing fast at cmake configure time - if a required cmake target was not installed then the user’s build will fail on cmake configure instead of failing at build or run time.  

**Initial thoughts on options to associated new targets with components:** 
- Option 1: Create a single “configuration” component and assign all new targets to it
     - Guard all configuration related targets with the WITH_CONFIGURATION option
     - Allow `opentelemetry-cpp::*_builder` targets to be optional based on the `WITH_*` options, but link them to a common `opentelemetry-cpp::configuration_builders` target that will be the primary interface for users
     - Add all related targets to a single component called “configuration” 
     - Mark the `ryml` dependency as required by the `configuration` component.
     - Set the `opentelemetry-cpp::init`, `opentelemetry-cpp::configuration`, and `opentelemetry-cpp::configuration_builders` cmake targets as required for the component. 
- Option 2: Assign the new targets to existing components and allow them to be optional 
    - Guard all configuration related targets with the `WITH_CONFIGURATION` option
    - Add `init` and `configuration` targets to the `sdk` component
    - Set the `ryml` dependency as optional for the `sdk` component.  
    - Add the `opentelemetry-cpp::*_builder` targets to the components associated with the `WITH_*` options that guard them now. 
    - Update the cmake components framework to support marking targets as optional for each component and provide some other mechanism to inform the user which targets are imported after calling find_package(opentelemetry-cpp..). 

Just some initial thoughts after a quick look at the PR - there are definitely other options to consider. What do you think?

dbarker: Another option: 

* Option 3: Create a `configuration_common` component and optional dependent builder components
   - Add a `configuration_common` component that requires the `ryml` dependency and `init` and `configuration` targets
   - Add builder components that depend on the common component (ie: `configuration_otlp_grpc`) that require the builder targets

This would allow cmake usage like this: 
```cmake
find_package(opentelemetry-cpp COMPONENTS configuration_otlp_grpc configuration_otlp_file)
add_library(my_lib my_lib.cc)
target_link_libraries(my_lib PRIVATE
   opentelemetry-cpp::init
   opentelemetry-cpp::configuration
   opentelemetry-cpp::otlp_grpc_builder
   opentelemetry-cpp::otlp_file_builder) 
```

marcalff:
Do not merge label: merged by small parts instead.

examples/yaml/CMakeLists.txt
dbarker: Now that the components are dynamically registered with otel_add_component any option will work. Resolving.

marcalff:
> Great addition to the project!

Thanks.

According to github history for this PR, it took 273 commits to finally add 9 lines ;-)

PR #2521, opened 2024-02-01
[EXPORTERS] Fix curl calls from otlp_http_exporter for versions >=7.5 and <7.80

### LINKED ISSUE #2520: otlp http exporter fails to connect: curl option CURLOPT_PREREQFUNCTION is used on unsupported versions of curl
**Describe your environment**
Ubuntu 20.04 with its curl/libcurl v 7.68.0-1ubuntu2.21
bazel 5.4.1

**Steps to reproduce**
Created an OtlpHttpExporter

No problem connecting, unsupported options not being used

```
[Error] File: external/io_opentelemetry_cpp/ext/src/http/client/curl/http_operation_curl.cc:526 CURL, set option <20312> failed: <An unknown option was passed in to libcurl>
[Error] File: external/io_opentelemetry_cpp/exporters/otlp/src/otlp_http_client.cc:200 [OTLP HTTP Client] Session state: connection failed.An unknown option was passed in to libcurl
[Error] File: external/io_opentelemetry_cpp/exporters/otlp/src/otlp_http_client.cc:168 [OTLP HTTP Client] Session state: session create failed.
[Error] File: external/io_opentelemetry_cpp/exporters/otlp/src/otlp_http_exporter.cc:118 [OTLP HTTP Client] ERROR: Export 4 trace span(s) error: 1 
```
20312 is CURLOPT_PREREQFUNCTION you can see [here](https://github.com/curl/curl/blob/26f79419eeb22b11cc2e305dc3fa92804b692a5b/docs/libcurl/symbols-in-versions#L739) that it is supported in curl 7.80.0 and above, although otlp exporter tries to use it for 7.5 and above which is exactly my case for producing the bug 7.5<7.68<7.8

So I think simply changing the test for the curl version will fix this bug, and I will post a PR for that in a couple of minutes.

https://github.com/open-telemetry/opentelemetry-cpp/blob/d32f9609b965dbb34030c1740ddb464d93cd3065/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h#L105

https://github.com/open-telemetry/opentelemetry-cpp/blob/d32f9609b965dbb34030c1740ddb464d93cd3065/ext/src/http/client/curl/http_operation_curl.cc#L155

Not sure I can unit test that though...

ecourreges-orange:
Fixes #2520

## Changes
Just changed the `#if` from version `0x075000` to `0x078000`

Trivial change, no changelog change

### DIFF
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -102,7 +102,7 @@ private:
 
   static size_t ReadMemoryCallback(char *buffer, size_t size, size_t nitems, void *userp);
 
-#if LIBCURL_VERSION_NUM >= 0x075000
+#if LIBCURL_VERSION_NUM >= 0x078000
   static int PreRequestCallback(void *clientp,
                                 char *conn_primary_ip,
                                 char *conn_local_ip,
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -152,7 +152,7 @@ size_t HttpOperation::ReadMemoryCallback(char *buffer, size_t size, size_t nitem
   return nwrite;
 }
 
-#if LIBCURL_VERSION_NUM >= 0x075000
+#if LIBCURL_VERSION_NUM >= 0x078000
 int HttpOperation::PreRequestCallback(void *clientp, char *, char *, int, int)
 {
   HttpOperation *self = reinterpret_cast<HttpOperation *>(clientp);
@@ -1044,7 +1044,7 @@ CURLcode HttpOperation::Setup()
   }
 #endif
 
-#if LIBCURL_VERSION_NUM >= 0x075000
+#if LIBCURL_VERSION_NUM >= 0x078000
   rc = SetCurlPtrOption(CURLOPT_PREREQFUNCTION, (void *)&HttpOperation::PreRequestCallback);
   if (rc != CURLE_OK)
   {


marcalff:
To confirm:

https://github.com/curl/curl/blob/9e560d11aad028de74addc0d1edfefa5667884f4/include/curl/curlver.h#L37-L60

#define LIBCURL_VERSION_MAJOR 7
#define LIBCURL_VERSION_MINOR 80
#define LIBCURL_VERSION_PATCH 0


/* This is the numeric version of the libcurl version number, meant for easier
   parsing and comparisons by programs. The LIBCURL_VERSION_NUM define will
   always follow this syntax:


         0xXXYYZZ


   Where XX, YY and ZZ are the main version, release and patch numbers in
   hexadecimal (using 8 bits each). All three numbers are always represented
   using two digits.  1.2 would appear as "0x010200" while version 9.11.7
   appears as "0x090b07".


   This 6-digit (24 bits) hexadecimal number does not show pre-release number,
   and it is always a greater number in a more recent release. It makes
   comparisons with greater than and less than work.


   Note: This define is the full hex number and _does not_ use the
   CURL_VERSION_BITS() macro since curl's own configure script greps for it
   and needs it to contain the full number.
*/
#define LIBCURL_VERSION_NUM 0x075000

ecourreges-orange:
Thanks for your insights @marcalff , so what should we do next?  

I can try adding a log and confirming this compilation/runtime discrepancy later this week, or I can close the issue.  

I definitely wouldn't rule out that it picked up Ubuntu jammy's curl v 7.81.0 for compilation and then runs on Ubuntu focal with the older v7.68.0

ecourreges-orange:
Cf the issue, the problem doesn't have to do with my docker environment with two OSes, but the fact that the bazel build of opentelemetry compiles on curl 8.4.0 that it downloads from fresh source, and later on, my app is a shared library loaded by apache which load the curl of the system which is 7.68.0 on Ubuntu focal 20.04

PR #2531, opened 2024-02-12
Remove namespace using in ETW exporter which affects global namespace

ThomsonTan:
## Changes

There is a using of namespace `opentelemetry::exporter::etw` at the global scope in [etw_provider.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/4700e483e09eb243251cd0bb9f0559f89237b41f/exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h#L53). It causes naming clash for any one who includes `etw_provider.h` which define the class name like `Logger` class.

This PR defines an alias on the namespace `opentelemetry::exporter::etw` instead of `using` it in global scope.


lalitb (approved):
Looks Good. Thanks for the fix.

PR #2533, opened 2024-02-14
[API] Propagation: fix for hex conversion to binary for odd hex strings

karusher:
The conversion of hex to binary did not handle odd-length inputs correctly.

The output was zero-padded on the right rather than the left. On little endian machines, the second to last entry received the 0.

For example, "78cfcfec62ae9e9" would result in:

`0x00    0x00    0x00    0x00    0x00    0x00    0x00    0x00  0x78    0xcf    0xcf    0xec    0x62    0xae    0x9e    0x09`

The desired output is:

`0x00    0x00    0x00    0x00    0x00    0x00    0x00    0x00   0x07    0x8c    0xfc    0xfe    0xc6    0x2a    0xe9    0xe9`

I have attached a toy program to reproduce the problem:

[test_program.txt](https://github.com/open-telemetry/opentelemetry-cpp/files/14273830/test_program.txt)

## Changes

Corrected conversion of hex to binary for propagation

### DIFF
--- a/api/include/opentelemetry/trace/propagation/detail/hex.h
+++ b/api/include/opentelemetry/trace/propagation/detail/hex.h
@@ -61,15 +61,17 @@ inline bool HexToBinary(nostd::string_view hex, uint8_t *buffer, size_t buffer_s
   int64_t buffer_pos   = int64_t(buffer_size) - (hex_size + 1) / 2;
   int64_t last_hex_pos = hex_size - 1;
 
-  int64_t i = 0;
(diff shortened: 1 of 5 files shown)


lalitb (approved):
Thanks for the fix. While this method is always used for converting even-length hex strings (trace_id and span_id) to binary, I believe the additional check for odd lengths should not significantly affect performance.

owent (approved):

marcalff (approved):
LGTM, thanks for the fix.

PR #2535, opened 2024-02-15
[opentracing-shim] Provide span for opentracing shim inject

karusher:
We noticed that the propagation headers were not being populated correctly when using the opentracing-shim. The
trace-id, span-id and trace flags were missing.

The various propagation injectors will attempt to retrieve the Span from the context that is provided in the api call:

  void Inject(context::propagation::TextMapCarrier &carrier,
              const context::Context &context) noexcept override

The shim will provide the RuntimeContext augmented with the baggage to the Inject method. Unfortunately, the shim does not add the span to the RuntimeContext. It does not instantiate a Scope.

This commit adds a Span to the context passed into Inject by modifying the shim.

Fixes # (issue)

## Changes


chusitoo:
> cc @chusitoo

Not a subject matter expert, but this looks like a very valid fix IMHO, and it is more or less equivalent to how the Java folks do it for their OT shim too.

marcalff (approved):
Thanks for the fix.

lalitb (approved):

PR #2540, opened 2024-02-18
[EXPORTER] OTLP file exporter

### LINKED ISSUE #2482: OTLP file exporter

According to 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/file-exporter.md#file-storage-requirements .

It we have file exporter, some projects will not depend openssl, curl, or gRPC any more.

Use OTLP utils to convert data to JSON format, just like it in OTLP HTTP exporter.

**Additional context**

There are some questions I want to discuss before implementing it. 
In practice, we usually have log rotate to control the disk usage, but the specification does not say anything about it.
Can I add some extensions to it?
For example, if we configure the file pattern as `%Y-%m-%d/otlp.%N.log` and set the rotate size to three, we want to create a directory every day and rotate output file name as `otlp.0.log`, `otlp.1.log` and `otlp.2.log`. And we may also set alias name `%Y-%m-%d/otlp.log` so the `otlp.log` will always point to the latest one?

owent:
Fixes #2482 

## Changes


owent:
This PR is ready to be reviewed now. @open-telemetry/cpp-approvers @perhapsmaple

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -0,0 +1,1560 @@
...
+
+bool OtlpFileClient::ForceFlush(std::chrono::microseconds timeout) noexcept
+{
+  if (backend_)
+  {
+    return backend_->ForceFlush(timeout);
+  }
+
+  return true;
+}
```
perhapsmaple, line 1539: I think we should add the force flush mechanism for the std::ostream case as well. In case the user calls `TracerProvider::ForceFlush`, we should add something like this:

```
std::ostream &out = nostd::get<std::reference_wrapper<std::ostream>>(options_.backend_options);
out.flush();
```

owent: Good point. I didn't implement it before because some ostreams is not thread-safety and we must ignore `timeout`.
It may be more important to keep it right. I will add flush for ostream later.

owent:
> Hi @owent, thanks for doing this. This is a very nice and well done PR. I have a couple of questions for now:
> 
> 1. I'm not very familiar with this rolling strategy. Most log tailers like Promtail and Fluentbit recommend the Rename and Create rotation strategy. Has this rotation strategy been tested with any existing log tailer? See here for more information: https://grafana.com/docs/loki/latest/send-data/promtail/logrotation/
> 2. I know this is probably a bit out of scope of the opentelemetry specification, but is it possible to make a base class like `FileBackend` so that users can implement their own rotation strategies if required. The `OtlpFileBackend` is very well designed and configurable, but might not fit everyone's use case.
> 
> We will have to test it with the `otlpjsonfilereceiver` from `opentelemetry-collector-contrib` to make sure that everything works properly. I'm a little busy for the next couple days but I can help you with this after.
> 
> Thanks

Thanks.

1. I want to keep the default implementation simple. This rotation strategy is tested in our internal collector agent which is also based on otel-collector.
2. It's a good idea to leave a base class to let users to use their own log sink, maybe spdlog, boost.log or other components.I will add it later. You can review it anytime when you have time. Thanks.

CMakeLists.txt
```diff
@@ -204,6 +204,9 @@ option(WITH_OTLP_GRPC "Whether to include the OTLP gRPC exporter in the SDK"
 option(WITH_OTLP_HTTP "Whether to include the OTLP http exporter in the SDK"
        OFF)
 
+option(WITH_OTLP_FILE_PREVIOUS
+       "Whether to include the OTLP file exporter in the SDK" OFF)
+set(WITH_OTLP_FILE ${WITH_OTLP_FILE_PREVIOUS})
```
marcalff, line 209:
```suggestion
option(WITH_OTLP_FILE
       "Whether to include the OTLP file exporter in the SDK" OFF)
```

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -0,0 +1,1587 @@
...
+            "[OTLP FILE Client]     you can use FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | "
+            "FORMAT_MESSAGE_FROM_SYSTEM"
+            << " | FORMAT_MESSAGE_IGNORE_INSERTS, nullptr, " << res
+            << ", MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), "
+            << "(LPTSTR) &lpMsgBuf, 0, nullptr) to get the error message, see "
+            << "https://docs.microsoft.com/en-us/windows/desktop/api/WinBase/"
+               "nf-winbase-formatmessage and "
+            << "https://docs.microsoft.com/en-us/windows/desktop/Debug/"
+               "retrieving-the-last-error-code for more "
+               "details");
```
marcalff, line 1276: Do we really want to print these instructions in the log ?

How about just printing the `res` return code, and have these instructions as C++ comments.

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -0,0 +1,1587 @@
...
+#  else
+        OTEL_INTERNAL_LOG_ERROR("[OTLP FILE Client]     you can use strerror("
+                                << res << ") to get the error message");
```
marcalff, line 1279: Same for non windows:

Do we really want to print these instructions in the log ?

How about just printing the `res` return code, and have these instructions as C++ comments.

marcalff (approved):
Very good work, as we are used to.

See some minor changes or comments.

marcalff:
Need to resolve my 3 comments, and need a second reviewer: adding the please-review label.

CMakeLists.txt
owent: Thaanks and done.

owent:
> Need to resolve my 3 comments, and need a second reviewer: adding the please-review label.

Done

CMakeLists.txt
```diff
@@ -204,6 +204,8 @@ option(WITH_OTLP_GRPC "Whether to include the OTLP gRPC exporter in the SDK"
 option(WITH_OTLP_HTTP "Whether to include the OTLP http exporter in the SDK"
        OFF)
 
+option(WITH_OTLP_FILE "Whether to include the OTLP file exporter in the SDK"
+       OFF)
```
ThomsonTan, line 208: nit: add a newline after the option.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_file_client.h
```diff
@@ -0,0 +1,89 @@
...
+#include "opentelemetry/nostd/shared_ptr.h"
+#include "opentelemetry/sdk/common/exporter_utils.h"
+
+#include "opentelemetry/exporters/otlp/otlp_file_client_options.h"
+
+#include <chrono>
+#include <cstddef>
+#include <memory>
+
+// forward declare google::protobuf::Message
```
ThomsonTan, line 15: This is not needed as it is already declared in the otlp_file_client_options.h included at line 9

ThomsonTan (approved):

lalitb:
Nice work. Can we document this exporter as an section in [README.md](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/exporters/otlp) ?

owent:
> Nice work. Can we document this exporter as an section in [README.md](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/exporters/otlp) ?

Sorry for the missing. I will add documents tomorrow.

marcalff:
> > Nice work. Can we document this exporter as an section in [README.md](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/exporters/otlp) ?
> 
> Sorry for the missing. I will add documents ~tomorrow~ several days later. Sorry, I'm too busy recently.I will commit and push documents together with codes if there are more feedbacks about this PR.

Merging this PR now, please open a PR later for some doc when you have time.

Thanks for this work.

PR #2244, opened 2023-07-22
Valgrind errors on std::atomic variables

marcalff:
Fixes #2243 

## Changes

Using only the default constructor to initialize `std::atomic` objects is incorrect.
This is a particularity of `std::atomic` itself, that does not follow the same rules as C++ in general.

See: https://en.cppreference.com/w/cpp/atomic/atomic/atomic

In the entire code base, reviewed every use of `std::atomic`,
and changed initialization to use the _direct list initialization syntax_.

see: https://en.cppreference.com/w/cpp/language/list_initialization

This form is the most robust and maintainable:

* auditing the source code for correctness is easier
* this form guarantees a proper initialization will be done, in all cases

This fixed the bug reported on the periodic metric reader, tested under valgrind.

This fixed a missing initialization in one of the `BatchLogRecordProcessor` constructors,
found by code review.

This also fixed the following pattern, which _technically_, is incorrect:

```
std::atomic<bool> flag;
flag.store(true);
```

With the fix, code such as:

```
std::atomic<bool> flag{false};
flag.store(true);
```

is correct, because it is now legal to call `store()` as the object _**is initialized**_.

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [X] Changes in public API reviewed

PR #2553, opened 2024-02-23
[SDK] Avoid missing conditional variable update and simplify atomic bool

arekay:
Addresses two issues -
1. Fix the use of a conditional variable where a wait on the variable might not be in flight when a notify is called. This is fixed by ensuring that an associated lock is aquired before calling the notify.
2. Instead of relying on a lock an a boolean, replace the use with a single atomic boolean.

Fixes # (issue)

## Changes
* Adds a scoped lock around notify call to conditional var in periodic exporter
* Replaces a spinlock+bool with an atomic bool in metrics reader


marcalff:
Thanks for the PR.

Not a full review, but this looks good based on a first read.

Please sign the EasyCLA to pass the CI checks.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -191,7 +193,11 @@ bool PeriodicExportingMetricReader::OnShutDown(std::chrono::microseconds timeout
 {
   if (worker_thread_.joinable())
   {
-    cv_.notify_one();
+    {
+      //ensure that `cv_` is awaiting, and the update doesn't get lost
+      std::unique_lock<std::mutex> lk(cv_m_);
```
owent, line 198: Do we need lock `cv_m_` here? Should we notify background thread immediately?

arekay: Yes, this is needed to ensure that the CV is [being waited on](https://github.com/open-telemetry/opentelemetry-cpp/blob/2b357fbe027afb0c65a6ec8c938841a0f75e97f4/sdk/src/metrics/export/periodic_exporting_metric_reader.cc#L64) otherwise the notify gets lost.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -191,7 +193,11 @@ bool PeriodicExportingMetricReader::OnShutDown(std::chrono::microseconds timeout
 {
   if (worker_thread_.joinable())
   {
-    cv_.notify_one();
+    {
+      //ensure that `cv_` is awaiting, and the update doesn't get lost
```
arekay, line 197:
```suggestion
      // ensure that `cv_` is awaiting, and the update doesn't get lost
```

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -4,6 +4,7 @@
 #include "opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
 #include "opentelemetry/sdk/metrics/push_metric_exporter.h"
+#include "opentelemetry/common/spin_lock_mutex.h"
```
arekay, line 7:
```suggestion
#include "opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h"
#include "opentelemetry/common/spin_lock_mutex.h"
#include "opentelemetry/sdk/common/global_log_handler.h"
#include "opentelemetry/sdk/metrics/push_metric_exporter.h"
```

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
owent: I was thinking about a lockless way to solve this problem. Or the critical section may be a little long to make more conflict.

marcalff: If it is possible for multiple threads to call flush, should the condition broadcast with notify_all() instead ?

owent: > @owent
> 
> About related issues like #2574
> 
> If it is possible for multiple threads to call flush/shutdown, should the condition broadcast with notify_all() instead ?

Both `notify_all` and `notify_one` may be missed when `wait_for` is called between predicate function and the internal `lock` function are called. I was thinking to use a atomic int64 to record the force flush sequence and wait only when notified sequence is less than flush sequence. We may assume int64 will not overflow in a single process.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -101,6 +102,7 @@ bool PeriodicExportingMetricReader::CollectAndExportOnce()
   bool notify_force_flush = is_force_flush_pending_.exchange(false, std::memory_order_acq_rel);
   if (notify_force_flush)
   {
+    std::unique_lock<std::mutex> lk(force_flush_m_);
```
owent, line 105: May this make deadlock when `OnForceFlush` has already locked `force_flush_m_` and is waiting for `is_force_flush_notified_` ?

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
arekay: @owent do you mean tag each incoming request with a int64, and on a flush request use the current counter value to wait for all `before` requests to complete? 
The notify_all makes sense since we can have multiple waiters, and i can add that.

owent: Yes , `notify_all` would be great. #2584 may also solve the problem which this PR fixes.

ThomsonTan:
Is it possible to add a test case to reproduce the issue and verify the fix?

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
marcalff: I don't think there is a deadlock, the usage pattern looks good to me.

See https://en.cppreference.com/w/cpp/thread/condition_variable

A lock should be used before calling notify().

A lock should be used before calling wait_for().

There is no deadlock, because wait_for() takes a lock object, and internally releases the lock and re-acquires it, in my understanding.

Similar to pthread_mutex and pthread_cond.

As long as the pattern uses the same mutex for the same condition variable, it works.

marcalff (commented):
LGTM, thanks for the fix.

marcalff:
> > Is it possible to add a test case to reproduce the issue and verify the fix?
> 
> Good point. But this problem is hard to reproduce. I have no idea about how to add test case by now. Maybe someone else can help?

It is very unlikely a test case can be written for this.

From other projects using mutex and condition variables, testing requires some serious code injection in debug, to control exactly how threads executes, to block a thread at a given point. opentelemetry-cpp does not have the tooling for this.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
owent: When `PeriodicExportingMetricReader::OnForceFlush` locked `force_flush_m_` with `std::unique_lock<std::mutex> lk_cv(force_flush_m_);` and use it to wait conditional_variable to be notified(`force_flush_cv_.wait_for(lk_cv, export_interval_millis_, break_condition);` will lock it again and wait `notify()` to unlock it.), the code `std::unique_lock<std::mutex> lk(force_flush_m_);` in `PeriodicExportingMetricReader::CollectAndExportOnce` will lock it first and then call `force_flush_cv_.notify_one()`. In my understanding, the `force_flush_m_` is already locked by `PeriodicExportingMetricReader::OnForceFlush` so `force_flush_cv_.notify_one()` will never be sent unless `force_flush_cv_.wait_for(lk_cv, export_interval_millis_, break_condition);` returns with timeout. If we set a long export interval, the lock may be deadlock.

owent:
> LGTM and approved, waiting for @owent to confirm since he had comments earlier.

Sorry, I'm on holiday these days.

#2584 may also fixes the problems this PR try to slove, and it also solve other deadlock problems in traces and metrics. And it has conflicts with this PR. Do you think we can just use #2584 and drop this one?

arekay:
@owent If #2584 is coming in soon, we can wait. Otherwise, we'd appreciate it if you can merge this and then follow up with #2584 since we have been using #2553 as a patch and it has fixed the frequent crashes we were seeing.  This will give you more time to test #2584 as well.

owent:
> @owent If #2584 is coming in soon, we can wait. Otherwise, we'd appreciate it if you can merge this and then follow up with #2584 since we have been using #2553 as a patch and it has fixed the frequent crashes we were seeing. This will give you more time to test #2584 as well.

I have the last question about the deadlock problem above. Other codes looks good to me.

arekay:
> > @owent If #2584 is coming in soon, we can wait. Otherwise, we'd appreciate it if you can merge this and then follow up with #2584 since we have been using #2553 as a patch and it has fixed the frequent crashes we were seeing. This will give you more time to test #2584 as well.
> 
> I have the last question about the deadlock problem above. Other codes looks good to me.

I have not had time to look into the deadlock problem, but anecdotally, we haven't had any deadlocks and eliminated crashes due to OTEL in our deployments. I believe this would be beneficial to all, even if short-lived and updated by #2584 .

arekay:
> I have not had time to look into the deadlock problem, but anecdotally, we haven't had any deadlocks and eliminated crashes due to OTEL in our deployments. I believe this would be beneficial to all, even if short-lived and updated by #2584 .

@owent Just checking if there is any update on this. Thanks.

owent:
> > I have not had time to look into the deadlock problem, but anecdotally, we haven't had any deadlocks and eliminated crashes due to OTEL in our deployments. I believe this would be beneficial to all, even if short-lived and updated by #2584 .
> 
> @owent Just checking if there is any update on this. Thanks.

@marcalff could you please check the lock problem again when you have time, or could #2584 be merged now?

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -72,8 +72,7 @@ class MetricReader
 protected:
 private:
   MetricProducer *metric_producer_;
-  mutable opentelemetry::common::SpinLockMutex lock_;
-  bool shutdown_;
+  std::atomic<bool> shutdown_;
```
marcalff, line 75: Please check the comments in PR #2244 for details

```suggestion
  std::atomic<bool> shutdown_{false};
```

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
marcalff: > In my understanding, the `force_flush_m_` is already locked by `PeriodicExportingMetricReader::OnForceFlush` so `force_flush_cv_.notify_one()` will never be sent unless `force_flush_cv_.wait_for(lk_cv, export_interval_millis_, break_condition);` returns with timeout.
> 

I think there is no deadlock.

See:

https://en.cppreference.com/w/cpp/thread/condition_variable

and in particular:

> Any thread that intends to wait on a std::condition_variable must:
> 
>     Acquire a [std::unique_lock](http://en.cppreference.com/w/cpp/thread/unique_lock)<[std::mutex](http://en.cppreference.com/w/cpp/thread/mutex)> on the mutex used to protect the shared variable.
>     Do one of the following: 
> 
>         Check the condition, in case it was already updated and notified.
>         Call [wait](https://en.cppreference.com/w/cpp/thread/condition_variable/wait), [wait_for](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for), or [wait_until](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_until) on the std::condition_variable (atomically releases the mutex and suspends thread execution until the condition variable is notified, a timeout expires, or a [spurious wakeup](https://en.wikipedia.org/wiki/Spurious_wakeup) occurs, then atomically acquires the mutex before returning).
>         Check the condition and resume waiting if not satisfied. 
> 
>         or: 
> 
>         Use the predicated overload of [wait](https://en.cppreference.com/w/cpp/thread/condition_variable/wait), [wait_for](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for), and [wait_until](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_until), which performs the same three steps.
> 


The critical part is:

> Call [wait](https://en.cppreference.com/w/cpp/thread/condition_variable/wait), [wait_for](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_for), or [wait_until](https://en.cppreference.com/w/cpp/thread/condition_variable/wait_until) on the std::condition_variable (atomically releases the mutex and suspends thread execution until the condition variable is notified, a timeout expires, or a [spurious wakeup](https://en.wikipedia.org/wiki/Spurious_wakeup) occurs, then atomically acquires the mutex before returning).

wait_for() will release the mutex upon entry, then wait. The mutex is not held while waiting, so it does not creates a deadlock.

> If we set a long export interval, the lock may be deadlock.

This should not happen in my understanding.

marcalff:
> > @owent If #2584 is coming in soon, we can wait. Otherwise, we'd appreciate it if you can merge this and then follow up with #2584 since we have been using #2553 as a patch and it has fixed the frequent crashes we were seeing. This will give you more time to test #2584 as well.
> 
> I have the last question about the deadlock problem above. Other codes looks good to me.

I looked again in details about the last remaining concern from @owent about a possible deadlock.

After code analysis, I don't think the deadlock is possible, as the code uses a well known pattern of locking a mutex before signalling a condition variable, and the condition `wait_for()` internally releases the mutex when the wait starts, to re acquire the lock once the wait ends.

Now merging.

@arekay 

Thanks for the fix, and also for your patience, sorry this review took such a long time.

arekay:
> Thanks for the fix, and also for your patience, sorry this review took such a long time.

@marcalff @owent Thanks for all the feedback and diligence - and of course, thanks for all the work you folks put in to maintaining something all of us benefit tremendously from!

PR #2557, opened 2024-02-27
Fix observable attributes drop

### LINKED ISSUE #2548: Asynchronous metric instruments ignoring attributes
**Describe your environment** 
I am using Windows, and building with Visual Studio 2019

**Steps to reproduce**
This issue can be reproduced by making some small changes to the "metrics_simple" example. Replace the file examples/common/metrics_foo_library/foo_library.cc with the attached version. The new file simply add some random attributes in the call to observe in the callback function. Build the example and run.

I expect to see some random attributes in the output

Attributes is always empty in the output

**Additional context**
This issue seems to be introduced from commit [cf5cdaa](https://github.com/open-telemetry/opentelemetry-cpp/commit/cf5cdaae860934469ae99826939f4f583674a836)
[foo_library.cc.txt](https://github.com/open-telemetry/opentelemetry-cpp/files/14387602/foo_library.cc.txt)

lalitb:
Fixes #2548 

## Changes

PR #2472 introduced the bug to drop the attributes in Observable Instruments. This PR addresses and corrects that issue. However, there remains an unrelated, longstanding bug affecting filtering for Observable Instruments, which has never been reported. A separate bug report will be filed for that issue.


lalitb:
@marcalff @ThomsonTan @esigo - I believe we need a patch release with the fix, as this is a common use-case and multiple users have reported it (few internally in my company too). I can create one once this is reviewed if you guys agree.

sdk/src/metrics/state/observable_registry.cc
```diff
@@ -50,6 +50,7 @@ void ObservableRegistry::CleanupCallback(opentelemetry::metrics::ObservableInstr
 
 void ObservableRegistry::Observe(opentelemetry::common::SystemTimestamp collection_ts)
 {
+  static DefaultAttributesProcessor default_attribute_processor;
```
ThomsonTan, line 53: What would happen without the default_attribute_processor specified here?

lalitb: They will get dropped by the [FilteredOrderedAttributeMap](https://github.com/open-telemetry/opentelemetry-cpp/blob/84e38b215f7b166e80353ff2c60836d9d375861d/sdk/src/metrics/state/filtered_ordered_attribute_map.cc#L19) which was added in earlier PR.

esigo (approved):
LGTM
Thanks for the fix

ThomsonTan (approved):

PR #2566, opened 2024-02-28
Provide a CMake target for prometheus_exporter_utils

### LINKED ISSUE #2565: Provide a CMake prometheus_exporter_utils library target

I'm trying to use opentelemetry prometheus exporter WITHOUT the CivetWeb Webserver that is baked into prometheus-cpp.

The current target provided by this package is `prometheus_exporter` which links against `prometheus-cpp::pull` & `prometheus-cpp::core`

The `pull` dependency brings in the civetweb server in as well. Since our application already has a webserver (boost::beast) we'd like to avoid this.

If this library provided a separate target for the `prometheus/exporter_utils.cc` file - which would only link against `prometheus-cpp::core`, I'd be able to use it.

Currently, I'm using a local copy of the source and I've added this separate target for 1 cc file and it all works locally.

**Additional context**

Here's the code that I have for my PrometheusExporter - it differs from the one present in lib as it has no PrometheusCollector & no Exposer:

```cpp
#pragma once

#include <memory>
#include <mutex>
#include <sstream>
#include <string>
#include <vector>

#include <opentelemetry/exporters/prometheus/exporter_utils.h>
#include <opentelemetry/sdk/metrics/metric_reader.h>
#include <prometheus/collectable.h>
#include <prometheus/text_serializer.h>

class PrometheusExporter : public opentelemetry::sdk::metrics::MetricReader
{
public:
    opentelemetry::sdk::metrics::AggregationTemporality GetAggregationTemporality(
        opentelemetry::sdk::metrics::InstrumentType /*instrument_type*/) const noexcept override
    {
        return opentelemetry::sdk::metrics::AggregationTemporality::kCumulative;
    };

    std::vector<prometheus::MetricFamily> CollectPrometheus()
    {
        if (IsShutdown())
        {
            return {};
        }

        collection_lock_.lock();
        std::vector<prometheus::MetricFamily> result;
        auto status = Collect([&result](opentelemetry::sdk::metrics::ResourceMetrics& metric_data) {
            auto prometheus_metric_data
                = opentelemetry::exporter::metrics::PrometheusExporterUtils::TranslateToPrometheus(
                    metric_data, true, true);
            for (auto& data : prometheus_metric_data)
                result.emplace_back(data);
            return true;
        });
        collection_lock_.unlock();
        return result;
    }

    std::string serialize()
    {
        // TODO: Do we need to allocate this stream here?
        // What if we took the boost stream itself as an arg & wrote to it?
        std::ostringstream output;
        prometheus::TextSerializer().Serialize(output, CollectPrometheus());
        return output.str();
    };

private:
    mutable std::mutex collection_lock_;

    bool OnForceFlush(std::chrono::microseconds /*timeout*/) noexcept override { return true; };
    bool OnShutDown(std::chrono::microseconds /*timeout*/) noexcept override { return true; };
};

```

dufferzafar:
Fixes #2565

### DIFF
--- a/exporters/prometheus/CMakeLists.txt
+++ b/exporters/prometheus/CMakeLists.txt
@@ -31,6 +31,25 @@ target_link_libraries(
   opentelemetry_exporter_prometheus
   PUBLIC opentelemetry_metrics prometheus-cpp::pull prometheus-cpp::core)
 
+# We create a separate target for utils library
+add_library(
+  opentelemetry_exporter_prometheus_utils
+  src/exporter_utils.cc)
+
+set_target_properties(opentelemetry_exporter_prometheus_utils
+                      PROPERTIES EXPORT_NAME prometheus_exporter_utils)
+set_target_version(opentelemetry_exporter_prometheus_utils)
+
+target_include_directories(
+  opentelemetry_exporter_prometheus_utils
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+         "$<INSTALL_INTERFACE:include>")
+
+list(APPEND PROMETHEUS_EXPORTER_TARGETS opentelemetry_exporter_prometheus_utils)
+target_link_libraries(
+  opentelemetry_exporter_prometheus_utils
+  PUBLIC opentelemetry_metrics prometheus-cpp::core)
+
 if(OPENTELEMETRY_INSTALL)
   install(
     TARGETS ${PROMETHEUS_EXPORTER_TARGETS}


exporters/prometheus/CMakeLists.txt
```diff
@@ -31,6 +31,25 @@ target_link_libraries(
...
+target_include_directories(
+  opentelemetry_exporter_prometheus_utils
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
+         "$<INSTALL_INTERFACE:include>")
+
+list(APPEND PROMETHEUS_EXPORTER_TARGETS opentelemetry_exporter_prometheus_utils)
+target_link_libraries(
+  opentelemetry_exporter_prometheus_utils
+  PUBLIC opentelemetry_metrics prometheus-cpp::core)
+
```
lalitb, line 52: Targets are also added in cmake config - https://github.com/open-telemetry/opentelemetry-cpp/blob/d036d8227f6f1f988c611aa6a0312f3ccd346d63/cmake/opentelemetry-cpp-config.cmake.in#L77 But I am not sure of the implications in that case - as the symbols from exporter_utils.cc would be part of two target belonging to same package `OPENTELEMETRY_CPP_LIBRARIES`. Will let @owent comment on that :)

lalitb: I don't think that should be an issue, resolving it. Sorry for confusion :)

dufferzafar:
@lalitb Does the patch look okay apart from your original comment?

exporters/prometheus/CMakeLists.txt
owent: I think we should remove `src/exporter_utils.cc` from and link `opentelemetry_exporter_prometheus_utils` into `opentelemetry_exporter_prometheus` if we want to split this file into a standalone target.Or there may be conflict when some components link both `opentelemetry_exporter_prometheus_utils` and `opentelemetry_exporter_prometheus`.

BTW: The documents and the component name for `OPENTELEMETRY_CPP_LIBRARIES` should also be added into `cmake/opentelemetry-cpp-config.cmake.in`.

esigo:
The PR should not break existing builds for the users and this has to be investigated before merge.

marcalff:
Thanks for the PR.

Currently, opentelemetry-cpp maintains the following builds:

* CMake, with various combinations for optional features (`WITH_OTLP_HTTP`, etc), and with various combinations for technical dependencies (C++ STL, abseil, etc)
* Bazel, also with complexity related to all the dependencies used.

On top of the build systems supported by opentelemetry-cpp,
some third parties are also making their own builds, for example:

* Vcpkg
* Conan
* bazel central

Maintaining all this ecosystem, and making sure various combinations of existing features are working properly is already a challenge for opentelemetry-cpp, and a major drain on development resources.

What this PR asks for makes sense, but the problem is that this new build combination would not be tested and covered in the existing `CI`, since the new library only makes sense when used by external code not available in opentelemetry-cpp.

Adding even more complexity to an already difficult to maintain area is not desirable at this point.

Given how this combination is not covered in our CI, there is a higher risk of this build to be broken by later changes or refactoring done in the opentelemetry-cpp make files, and the break can go unnoticed.

As a result, if we (opentelemetry-cpp) accept this PR, we would just be setting expectations (to support it and maintain it) that realistically will not be met.

Closing this PR.

Overall, the need to cut the dependency from the prometheus exporter to a given web server is valid.

The best way to achieve this is to extend the prometheus exporter to support alternate web servers instead, but this is a much bigger feature and not just a makefile change to split a library.

PR #2567, opened 2024-02-28
[BUILD] Fix old style cast warning

keith:
Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/2556

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -60,7 +60,7 @@ size_t HttpOperation::WriteVectorHeaderCallback(void *ptr, size_t size, size_t n
     return 0;
   }
 
-  const unsigned char *begin = (unsigned char *)(ptr);
+  const unsigned char *begin = static_cast<unsigned char *>(ptr);
   const unsigned char *end   = begin + size * nmemb;
   self->response_headers_.insert(self->response_headers_.end(), begin, end);
 
@@ -90,7 +90,7 @@ size_t HttpOperation::WriteVectorBodyCallback(void *ptr, size_t size, size_t nme
     return 0;
   }
 
-  const unsigned char *begin = (unsigned char *)(ptr);
+  const unsigned char *begin = static_cast<unsigned char *>(ptr);
   const unsigned char *end   = begin + size * nmemb;
   self->response_body_.insert(self->response_body_.end(), begin, end);
 
@@ -810,7 +810,7 @@ CURLcode HttpOperation::Setup()
     if (ssl_options_.ssl_insecure_skip_verify)
     {
       /* 6 - DO NOT ENFORCE VERIFICATION, This is not secure. */
-      rc = SetCurlLongOption(CURLOPT_USE_SSL, (long)CURLUSESSL_NONE);
+      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<long>(CURLUSESSL_NONE));
       if (rc != CURLE_OK)
       {
         return rc;
@@ -831,7 +831,7 @@ CURLcode HttpOperation::Setup()
     else
     {
       /* 6 - ENFORCE VERIFICATION */
-      rc = SetCurlLongOption(CURLOPT_USE_SSL, (long)CURLUSESSL_ALL);
+      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<long>(CURLUSESSL_ALL));
       if (rc != CURLE_OK)
       {
         return rc;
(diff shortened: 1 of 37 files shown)


lalitb:
Thanks for the fix. Would be also good to add this warning as an error in maintainer mode CI - https://github.com/open-telemetry/opentelemetry-cpp/blob/ba183049a555aa5e103ffb63b73c04636af2751c/CMakeLists.txt#L469, this will reveal if there are other places to fix too.

lalitb (approved):

ThomsonTan (approved):

keith:
There are actually quite a few more, I will push them and that cmake change here

keith:
pushed! note there are a few cases where `reinterpret_cast` was the replacement instead

keith:
Ok I think this is ready for re-review!

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -932,40 +932,40 @@ CURLcode HttpOperation::Setup()
       return rc;
     }
 
-    rc = SetCurlPtrOption(CURLOPT_WRITEFUNCTION, (void *)&HttpOperation::WriteMemoryCallback);
+    rc = SetCurlPtrOption(CURLOPT_WRITEFUNCTION, reinterpret_cast<void *>(&HttpOperation::WriteMemoryCallback));
```
marcalff, line 935: For all the calls to `SetCurlPtrOption`, the code was using a cast because a macro was used in the past.

Now that `SetCurlPtrOption` is a real function that takes a typed `void *` parameter, I think all the casts can be removed instead.

keith: At least on this specific line that results in:

```
ext/src/http/client/curl/http_operation_curl.cc:935:50: error: cannot initialize a parameter of type 'void *' with an rvalue of type 'size_t (*)(void *, size_t, size_t, void *)' (aka 'unsigned long (*)(void *, unsigned long, unsigned long, void *)')
    rc = SetCurlPtrOption(CURLOPT_WRITEFUNCTION, &HttpOperation::WriteMemoryCallback);
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ext/src/http/client/curl/http_operation_curl.cc:509:67: note: passing argument to parameter 'value' here
CURLcode HttpOperation::SetCurlPtrOption(CURLoption option, void *value)
                                                                  ^
```

marcalff (commented):
A few more warnings indeed, thanks for the cleanup.

See a comment on SetCurlPtrOption

ext/src/http/client/curl/http_operation_curl.cc
marcalff: This specific one may need a cast, but others can be simplified I hope.

keith: this is true for the `static_cast`s in this file though, so I removed those

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1037,21 +1041,22 @@ CURLcode HttpOperation::Setup()
...
-  rc = SetCurlPtrOption(CURLOPT_PREREQFUNCTION, (void *)&HttpOperation::PreRequestCallback);
+  rc = SetCurlPtrOption(CURLOPT_PREREQFUNCTION,
+                        reinterpret_cast<void *>(&HttpOperation::PreRequestCallback));
   if (rc != CURLE_OK)
   {
     return rc;
   }
 
-  rc = SetCurlPtrOption(CURLOPT_PREREQDATA, (void *)this);
+  rc = SetCurlPtrOption(CURLOPT_PREREQDATA, this);
```
ThomsonTan, line 1059: Should `static_cast<void *>` be added here?

marcalff: No, no need for it, SetCurlPtrOption() is a otel-cpp helper with proper types, not a curl api.

marcalff (approved):
LGTM, thanks for the cleanup.

ext/src/http/client/curl/http_operation_curl.cc
ThomsonTan: But `SetCurlPtrOption` accepts `void *` instead of the typed pointer of `this`, so we assume it is safe to convert `this` to `void *` type?

marcalff: Any pointer `T *` can be converted to `void *`, no need for a cast.

marcalff:
@keith Last CI failures are in tests, because of ifdef for ABI_V2.

```
/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/trace/tracer_provider_test.cc:175:33: error: use of old-style cast [-Werror,-Wold-style-cast]
                          {"d", (unsigned int)314159},
```

Should be ready to go after this last pass.

keith:
ah thanks yea i didn't build with that locally, pushed

marcalff:
> @keith Last CI failures are in tests, because of ifdef for ABI_V2.
> 
> ```
> /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/trace/tracer_provider_test.cc:175:33: error: use of old-style cast [-Werror,-Wold-style-cast]
>                           {"d", (unsigned int)314159},
> ```
> 
> Should be ready to go after this last pass.

Search for "error:" in the CI logs.

Maintainer builds do a make -k to report as many errors as possible in one build, there are a few reported there.

keith:
thanks!

marcalff:
Thanks @keith

Issue #2572, opened 2024-03-01
[BUILD] Opentelemetry C++20 extension causes Envoy build issue


shiponcs:
*Title*: *[BUILD] Opentelemetry C++20 extension causes Envoy build issue*

*Description*:
Envoy build failed because of Opentelemetry C++20 requirement. 
```bash
$ bazel build envoy
ERROR: /source/source/extensions/tracers/opentelemetry/BUILD:14:19: Compiling source/extensions/tracers/opentelemetry/config.cc failed: (Exit 1): clang-14 failed: error executing command (from target //source/extensions/tracers/opentelemetry:config) /opt/llvm/bin/clang-14 -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics -fno-omit-frame-pointer ... (remaining 239 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
In file included from source/extensions/tracers/opentelemetry/config.cc:8:
In file included from ./source/extensions/tracers/opentelemetry/opentelemetry_tracer_impl.h:17:
In file included from bazel-out/k8-fastbuild/bin/external/io_opentelemetry_cpp/api/_virtual_includes/api/opentelemetry/logs/provider.h:10:
In file included from bazel-out/k8-fastbuild/bin/external/io_opentelemetry_cpp/api/_virtual_includes/api/opentelemetry/logs/noop.h:11:
In file included from bazel-out/k8-fastbuild/bin/external/io_opentelemetry_cpp/api/_virtual_includes/api/opentelemetry/logs/event_logger.h:7:
bazel-out/k8-fastbuild/bin/external/io_opentelemetry_cpp/api/_virtual_includes/api/opentelemetry/logs/logger.h:261:5: error: use of the 'likely' attribute is a C++20 extension [-Werror,-Wc++20-attribute-extensions]
    OPENTELEMETRY_LIKELY_IF(Enabled(severity) == false) { return false; }
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
bazel-out/k8-fastbuild/bin/external/io_opentelemetry_cpp/api/_virtual_includes/api/opentelemetry/common/macros.h:12:11: note: expanded from macro 'OPENTELEMETRY_LIKELY_IF'
[...]

PR #2580, opened 2024-03-07
[BUILD] Provide LIKELY / UNLIKELY macros

marcalff:
Contributes to #2572

## Changes

* Remove the `OPENTELEMETRY_LIKELY_IF` macro
  * This macro does not make sense because it includes the `if` itself, making it unusable for other constructs (for loops, while loops)
* Provide a `OPENTELEMETRY_LIKELY_CONDITION` macro
  * To be used instead of `OPENTELEMETRY_LIKELY_IF`.
  * Only one argument is supported instead of a `__VA_ARGS__`, to discourage `if OPENTELEMETRY_LIKELY_CONDITION (a = foo(), b = bar(), a > b)`
  * This macro supports both compiler builtins for gcc and clang, and C++20.
* Provide a `OPENTELEMETRY_UNLIKELY_CONDITION` macro
* Provide a `OPENTELEMETRY_LIKELY` macro
* Provide a `OPENTELEMETRY_UNLIKELY` macro

With this change, likely / unlikely hints can be placed anywhere (for loops, while loops, switch case, etc). These hints are not used only for if conditions.

### DIFF
--- a/api/include/opentelemetry/common/macros.h
+++ b/api/include/opentelemetry/common/macros.h
@@ -3,6 +3,82 @@
 
 #pragma once
 
+/*
+   Expected usage pattern:
+
+   if OPENTELEMETRY_LIKELY_CONDITION (ptr != nullptr)
+   {
+     do_something_likely();
+   } else {
+     do_something_unlikely();
+   }
+
+   This pattern works with gcc/clang and __builtin_expect(),
+   as well as with C++20.
+   It is unclear if __builtin_expect() will be deprecated
+   in favor of C++20 [[likely]] or not.
+
+   OPENTELEMETRY_LIKELY_CONDITION is preferred over OPENTELEMETRY_LIKELY,
+   to be revisited when C++20 is required.
+*/
+
+#if !defined(OPENTELEMETRY_LIKELY_CONDITION) && defined(__cplusplus)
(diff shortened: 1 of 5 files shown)


api/include/opentelemetry/common/macros.h
```diff
@@ -3,22 +3,36 @@
...
+#      if __has_cpp_attribute(likely)
+#        define OPENTELEMETRY_LIKELY [[likely]]
+#      endif
 #    endif
 #  endif
 #endif
-#if !defined(OPENTELEMETRY_LIKELY_IF) && (defined(__clang__) || defined(__GNUC__))
-#  define OPENTELEMETRY_LIKELY_IF(...) if (__builtin_expect(!!(__VA_ARGS__), true))
+
+#ifndef OPENTELEMETRY_LIKELY
```
owent, line 18: Why do we remove `__builtin_expect` ? In my understanding, most default STD version of opensource projects will not turn C++20 on by now.

marcalff: Restored support for `__builtin_expect`, but with `OPENTELEMETRY_LIKELY_CONDITION` instead.

owent (approved):
LGTM and thanks.

PR #2584, opened 2024-03-10
[SDK] Fix forceflush may wait for ever

### LINKED ISSUE #2574: [SDK] `BatchSpanProcessor::ForceFlush` appears to be hanging forever
**Describe your environment**
Platform: Linux/x86_64
Build system: CMake
Package manager: conda/conda-forge
OpenTelemetry version: 1.14.1

**Steps to reproduce**
Unfortunately, this is entwined into a large proprietary application and I've not been able to extricate it. We are using the OTLP exporter, BatchSpanProcessor, and we are calling Flush periodically with a timeout.

ForceFlush with a timeout should reliably complete within the timeout (roughly)

ForceFlush appears to get stuck inside the spinlock:

```
#0  0x00007f52a6c49f17 in __GI_sched_yield () at ../sysdeps/unix/syscall-template.S:120
#1  0x00007f52a84ce4ee in opentelemetry::v1::sdk::trace::BatchSpanProcessor::ForceFlush(std::chrono::duration<long, std::ratio<1l, 1000000l> >) () from /home/lidavidm/miniforge3/envs/dev/bin/../lib/libopentelemetry_trace.so
```

or 

```
   0x00007f52a84ce49e <+334>:	jne    0x7f52a84ce4bd <_ZN13opentelemetry2v13sdk5trace18BatchSpanProcessor10ForceFlushENSt6chrono8durationIlSt5ratioILl1ELl1000000EEEE+365>
   0x00007f52a84ce4a0 <+336>:	pause
=> 0x00007f52a84ce4a2 <+338>:	mov    %ebx,%eax
   0x00007f52a84ce4a4 <+340>:	and    $0x7f,%eax
   0x00007f52a84ce4a7 <+343>:	cmp    $0x7f,%eax
```

These traces appear to me to correspond with the spinlock portion. It appears to me that the spinlock portion does not respect the timeout:

https://github.com/open-telemetry/opentelemetry-cpp/blob/07f6cb54ece56691dbd2a94b0cbeec722ff6a631/sdk/src/trace/batch_span_processor.cc#L120-L137

**Additional context**
We _think_ this only occurred once we upgraded from 1.13.0 to 1.14.1. Note that we did not used to have a timeout, since we did not feel a need to. When we upgraded, we noticed it was getting stuck exporting, and added the timeout to try to avoid this (and as general good practice), but it seems there's some deeper issue as to why the export seems to never complete.

I did set a breakpoint on the actual Export method and it does appear to be called regularly, but the main thread that is blocking in ForceFlush isn't getting unblocked.

### LINKED ISSUE #2583: [SDK] BatchLogRecordProcessor::ForceFlush hangs for 10 seconds
Platform : Ubuntu 23.10
Build: opentelemetry 1.11.0 with WITH_ASYNC_EXPORT_PREVIEW enabled
Deps: curl 7.85.0

Here are the lines of code used to initialize the logger provider:
```
opentelemetry::exporter::otlp::OtlpHttpLogRecordExporterOptions opts;

auto exporter =
    opentelemetry::exporter::otlp::OtlpHttpLogRecordExporterFactory::Create(
        opts);

opentelemetry::sdk::logs::BatchLogRecordProcessorOptions batchOptions;

auto processor =
    opentelemetry::sdk::logs::BatchLogRecordProcessorFactory::Create(
        std::move(exporter),
        batchOptions);

auto resource = opentelemetry::sdk::resource::Resource::Create(attributes);

std::shared_ptr<opentelemetry::logs::LoggerProvider> provider =
    opentelemetry::sdk::logs::LoggerProviderFactory::Create(
        std::move(processor),
        resource);

opentelemetry::logs::Provider::SetLoggerProvider(provider);
```
Before quitting my application I do a `ForceFlush` with a timeout equal to 1 second.
```
auto loggerProvider =
            reinterpret_cast<opentelemetry::sdk::logs::LoggerProvider*>(
                provider);

loggerProvider->ForceFlush(std::chrono::seconds(1));

std::shared_ptr<opentelemetry::logs::LoggerProvider> none;
opentelemetry::logs::Provider::SetLoggerProvider(none);
```

When the network connection is not good, it can happen that my application hangs for 10 seconds, which corresponds to the timeout of the curl request which I don't want to change.

I'd like to know why the timeout is not taken into account.

owent:
Fixes #2574

May also fixes #2583 

## Changes

+ Using a sequence to check if a complete notification includes the datas when call `ForceFlush`.


sdk/src/logs/batch_log_record_processor.cc
```diff
@@ -110,35 +114,17 @@ bool BatchLogRecordProcessor::ForceFlush(std::chrono::microseconds timeout) noex
...
   while (!result && timeout_steady > std::chrono::steady_clock::duration::zero())
   {
-    // When is_force_flush_notified.store(true) and force_flush_cv.notify_all() is called
-    // between is_force_flush_pending.load() and force_flush_cv.wait(). We must not wait
-    // for ever
+    // When force_flush_notified_sequence.compare_exchange_strong(...) and
+    // force_flush_cv.notify_all() is called between force_flush_pending_sequence.load(...) and
+    // force_flush_cv.wait(). We must not wait for ever
     std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
     result = synchronization_data_->force_flush_cv.wait_for(lk_cv, scheduled_delay_millis_,
```
ggaudeau, line 121: If the value of `timeout` is lesser than that of `scheduled_delay_millis_`, then we'll wait longer than expected I guess

owent: Good point, it's fixed now.

owent:
@ThomsonTan Is there any problem in this PR? Maybe we can close #2553 if this PR is merged. It fixes the same problems as #2553 do and also fix the simular problems in trace and logs.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
```diff
@@ -115,20 +115,25 @@ class BatchLogRecordProcessor : public LogRecordProcessor
 
     /* Important boolean flags to handle the workflow of the processor */
     std::atomic<bool> is_force_wakeup_background_worker{false};
-    std::atomic<bool> is_force_flush_pending{false};
-    std::atomic<bool> is_force_flush_notified{false};
-    std::atomic<std::chrono::microseconds::rep> force_flush_timeout_us{0};
     std::atomic<bool> is_shutdown{false};
+    std::atomic<std::uint64_t> force_flush_pending_sequence{false};
```
ThomsonTan, line 119: nit: initialize std::atomic<uint64_t> variable to 0 instead of `false`.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
```diff
@@ -115,20 +115,25 @@ class BatchLogRecordProcessor : public LogRecordProcessor
...
+    std::atomic<std::uint64_t> force_flush_notified_sequence{false};
```
ThomsonTan, line 120: nit: the `std::` prefix for `uint64_t` is unnecessary?

sdk/src/logs/batch_log_record_processor.cc
```diff
@@ -79,21 +79,25 @@ bool BatchLogRecordProcessor::ForceFlush(std::chrono::microseconds timeout) noex
   // Now wait for the worker thread to signal back from the Export method
   std::unique_lock<std::mutex> lk_cv(synchronization_data_->force_flush_cv_m);
 
-  synchronization_data_->is_force_flush_pending.store(true, std::memory_order_release);
+  std::uint64_t current_sequence =
+      synchronization_data_->force_flush_pending_sequence.fetch_add(1, std::memory_order_release) +
```
ThomsonTan, line 83: Is the fetch_add here just the same as `synchronization_data_->force_flush_pending_sequence++`?

ThomsonTan:
BTW, can you explain in brief that why the sequence notification can break the potential forever wait? Thanks.

sdk/src/logs/batch_log_record_processor.cc
owent: In my understanding, `synchronization_data_->force_flush_pending_sequence++` use `std::memory_order_seq_cst` , we can use `std::memory_order_release` here.

owent:
> BTW, can you explain in brief that why the sequence notification can break the potential forever wait? Thanks.

When more than one threads call `ForceFlush` concurrency, `is_force_flush_notified` may be set `true` only once in background thread and one of the threads call `ForceFlush` will wait `is_force_flush_notified` for ever.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
owent: Thanks and fixed.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
owent: Thanks and fixed.

sdk/src/logs/batch_log_record_processor.cc
```diff
@@ -65,7 +65,7 @@ void BatchLogRecordProcessor::OnEmit(std::unique_ptr<Recordable> &&record) noexc
   {
     // signal the worker thread
     synchronization_data_->is_force_wakeup_background_worker.store(true, std::memory_order_release);
-    synchronization_data_->cv.notify_one();
+    synchronization_data_->cv.notify_all();
```
ThomsonTan, line 68: Could there be performance impact on the switching of `notify_one` to `notify_all` in `OnEmit`?

owent: In my understanding, there is no difference between `notify_one` and `notify_all` when there is only a single background thread.
However, we should use `notify_all` in background threads (for example, in `BatchLogRecordProcessor::OnEmit`) to wake up all waiting threads. To keep things simple and reduce the chance of errors, it's best to consistently use `notify_all` in all cases.

ThomsonTan (approved):

marcalff (approved):
Thanks for the fix.

This area also changed recently when is_shutdown was fixed to be an atomic instead of a plain boolean.

I was concerned with potential overlap and collisions, but it turns out the two fixes are really independent, which is better.

LGTM.

marcalff:
Adding ok-to-merge.

Waiting to give @lalitb and @esigo a chance to comment, and planning to merge early next week.

PR #2603, opened 2024-03-22
Mozbuild

jjjalkanen:
## Changes

With these patches, the sdk and the multithreaded example can be built as a third-party dependency for Firefox.


jjjalkanen:
Please ignore this PR

PR #2606, opened 2024-03-25
Do not link `prometheus-cpp::util` when it doesn't exist

### LINKED ISSUE #2605: `prometheus-cpp::util` is not found in prometheus-cpp 1.1.0 and before. Which will make prometheus failed to build
**Describe your environment**

Windows

**Steps to reproduce**

Build otel-cpp with prometheus-cpp 1.1.0

Success

```
[cmake] CMake Error at exporters/prometheus/CMakeLists.txt:33 (target_link_libraries):
[cmake]   Target "opentelemetry_exporter_prometheus" links to:
[cmake] 
[cmake]     prometheus-cpp::util
[cmake] 
[cmake]   but the target was not found.  Possible reasons include:
[cmake] 
[cmake]     * There is a typo in the target name.
[cmake]     * A find_package call is missing for an IMPORTED target.
[cmake]     * An ALIAS target is missing.
[cmake] 
[cmake] 
[cmake] 
```

owent:
Fixes #2605 

## Changes

+ Do not link `prometheus-cpp::util` when it doesn't exist

### DIFF
--- a/exporters/prometheus/CMakeLists.txt
+++ b/exporters/prometheus/CMakeLists.txt
@@ -30,10 +30,12 @@ endif()
 if(TARGET util)
   list(APPEND PROMETHEUS_EXPORTER_TARGETS util)
 endif()
-target_link_libraries(
-  opentelemetry_exporter_prometheus
-  PUBLIC opentelemetry_metrics prometheus-cpp::pull prometheus-cpp::core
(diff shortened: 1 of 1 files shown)


marcalff (approved):
Thanks for the fix

Issue #2334, opened 2023-09-27
Add bzlmod support


auzhva:
Since bazel 6.0, bazel added [bzlmod](https://bazel.build/external/overview?hl=en#bzlmod) dependency management and tracking system. Currently open telemetry is built by Bazel but is not published to the [Bazel Central Regestry](https://registry.bazel.build), which complicates importing open telemetry.

It would be awesome if open telemetry would be packaged to the BCR.

**Is your feature request related to a problem?**
Opentelemetry imports dependencies with own `opentelemetry_cpp_deps`, which is not portable to bzlmod.

**Describe the solution you'd like**
Create MODULE.bazel and publish to the BCR

**Describe alternatives you've considered**
Continue using existing, but it's... Hard...

esigo:
@auzhva unfortunately we could not offer any official packaging.

esigo:
@auzhva unfortunately we could not offer any official packaging.

marcalff:
While this request makes sense, we (in opentelemetry-cpp) can not provide packaging for every build system.

The best way to address this is to make a contribution to the Bazel Central Registry directly.

marcalff:
Added `help wanted` to advertise.

Do not prepare a PR for opentelemetry-cpp, contribute to https://registry.bazel.build/ instead.

keith:
:wave: I submitted a PR for the modern bazel dependency management setup here:

I know there was some discussion above about submitting the config to the BCR (which I will do if this PR is merged), but my PR here is still required to modernize the bazel config in this repo.

We could also theoretically remove the old version of doing things in bazel, since right now there is some duplication, but I am initially leaving it in my PR to not affect users on older versions of bazel. I'm happy to remove it though and direct older bazel users to prior releases of this library instead if you'd prefer!

https://github.com/open-telemetry/opentelemetry-cpp/pull/2608

keith:
still valid, linked PR should be unblocked when my other prs land

PR #2608, opened 2024-03-25
[BUILD] Add support for bzlmod

keith:
This adds support for bzlmod, which is bazel's new dependency resolution system.

Theoretically we could now remove the previous dependency management configuration which required flattening all the dependencies, but I left it here for now for users using old versions of bazel.

Fixes: https://github.com/open-telemetry/opentelemetry-cpp/issues/2334


keith:
ok I pushed an update here that seems to fix it at least with c++17, it depends on https://github.com/bazelbuild/bazel-central-registry/pull/1810

marcalff:
Hi @keith 

Could you merge with a recent main to resolve conflicts,
and investigate the remaining CI failures for bazel if any ?

Thanks,
-- Marc

keith:
The windows and asan failures here are real.

keith:
hopefully will be unblocked by https://github.com/open-telemetry/opentelemetry-cpp/pulls/keith

marcalff:
@keith 

Now that #2679 is merged, could you update this PR and resolve conflicts ?

Looking to include this in the upcoming release.

Thanks.

keith:
I think there might still be a grpc ODR issue I have to fix, but we'll see if CI fails

MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+module(
+    name = "opentelemetry-cpp",
+    version = "1.14.2",
```
marcalff, line 6:
```suggestion
    version = "1.15.0",
```

I assume this should match the current version in the opentelemetry-cpp repository.

MODULE.bazel
```diff
@@ -0,0 +1,25 @@
...
+    compatibility_level = 0,
+    repo_name = "io_opentelemetry_cpp",
+)
+
+bazel_dep(name = "abseil-cpp", version = "20240116.1", repo_name = "com_google_absl")
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "curl", version = "8.4.0")
+bazel_dep(name = "grpc", version = "1.62.1", repo_name = "com_github_grpc_grpc")
+bazel_dep(name = "nlohmann_json", version = "3.11.3", repo_name = "github_nlohmann_json")
+bazel_dep(name = "opentelemetry-proto", version = "1.1.0", repo_name = "com_github_opentelemetry_proto")
```
marcalff, line 16:
```suggestion
bazel_dep(name = "opentelemetry-proto", version = "1.3.1", repo_name = "com_github_opentelemetry_proto")
```

.github/workflows/ci.yml
```diff
@@ -537,6 +537,27 @@ jobs:
     - name: run tests
       run: ./ci/do_ci.sh bazel.test
 
+  bazel_no_bzlmod_test:
+    name: Bazel
```
marcalff, line 541: There is already a CI build named Bazel, please pick something more descriptive.

ci/do_ci.sh
```diff
@@ -459,6 +459,10 @@ EOF
   make load_plugin_example
   examples/plugin/load/load_plugin_example ${PLUGIN_DIR}/libexample_plugin.so /dev/null
   exit 0
+elif [[ "$1" == "bazel.no_bzlmod.test" ]]; then
+  bazel $BAZEL_STARTUP_OPTIONS build --enable_bzlmod=false $BAZEL_OPTIONS //...
+  bazel $BAZEL_STARTUP_OPTIONS test --enable_bzlmod=false $BAZEL_TEST_OPTIONS //...
```
marcalff, line 464: So, does this build ignores the content of file MODULE.bazel ?

Should we have another build with bzlmod enabled ?

marcalff (commented):
Thanks for the PR.

See minor fixes, and some questions.

ci/do_ci.sh
keith: bazel 7.x enables bzlmod by default, so when this merges all bazel jobs that already exist will use this new method by default. so I added this one to make sure the old way continues to work at least for this single build, and figure we can run all the other CI jobs with the new way

MODULE.bazel
keith: actually this number just needs to be replaced when it's uploaded to the BCR with a patch, so I actually just zero'd this out

MODULE.bazel
keith: looks like this version isn't in the BCR yet, so I'll have to add that there first

keith: https://github.com/bazelbuild/bazel-central-registry/pull/2252

marcalff: proto 1.3.1 is now merged in BCR, upgrading to 1.3.1

keith:
thanks!

PR #2609, opened 2024-03-25
[SDK] Remove unused variable

keith:


lalitb (approved):
Thanks for the fix. Not related to this PR, it may be good to catch unused variables in maintainer mode build - https://github.com/open-telemetry/opentelemetry-cpp/blob/02e6ad25ada84c0df7a645c954f820e38fa8c887/CMakeLists.txt#L487. cc @marcalff

PR #2616, opened 2024-03-28
[MISC] Use set -e on all shell scripts and pass shellcheck --severity=error

johnbley:
## Changes

Numerous shell scripts were not using `set -e` or equivalent.  Additionally, make all scripts pass at least `shellcheck --severity=error` (basically just argument array quoting and exit values).  I can add an additional PR for using `shellcheck` in CI.


marcalff (approved):
Thanks for the fix.

PR #2627, opened 2024-04-03
[CI] Use platform CMake

marcalff:
Use platform CMake

## Changes

In `CI`, the build scripts used to:

* download cmake installation scripts from the internet
* make the scripts executable
* execute the downloaded script

to install `CMake` version 3.16.3 by default, or `CMake` version 3.20.6

All the platforms used in `CI` have already have `CMake`, for a more recent version.

Testing on old versions instead of testing with the version that comes with the platform provides little value, since users are more likely to run into issues that were not detected in CI.

Also, downloading a script and executing it is a bad practice.

This fix uses the `CMake` version already installed on the github runner image.


ThomsonTan:
The scripts like "tools/setup-cmake.sh" can also be used in non CI environment, such as setting up the CMake to a given machine in dev environment?

marcalff:
> The scripts like "tools/setup-cmake.sh" can also be used in non CI environment, such as setting up the CMake to a given machine in dev environment?

I don't know any dev who would go out of his way to install something like this, instead of using the natural packaging commands available with the OS distribution, for example:

```bash
sudo apt install cmake
```

This `tools/setup-cmake.sh` script installs under `/usr/local` with a `make install`, which imply that to run, the whole script must be executed as root, including the `make` part.

It builds from code downloaded in `/tmp`, which may have been poisoned before the script with an extra payload.

It installs CMake `3.18.4` by default, which is ancient.

It is not documented, one has to read the code to know how to invoke it.

It is not referenced anywhere in the opentelemetry-cpp documentation, one has to find it to know it exists.

This is a long list of red flags in my opinion.

Restoring the file for now because it appear to be used from `tools/setup-buildtools.sh` and `tools/build-docker.cmd` which I missed, but I still think this should be removed at some point.

ci/setup_cmake.sh
```diff
@@ -30,17 +25,6 @@ else
...
-{
-  tmp_dir=$(mktemp -d)
-  pushd $tmp_dir
-  wget https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-Linux-x86_64.sh
-  chmod +x cmake-${CMAKE_VERSION}-Linux-x86_64.sh
-  ./cmake-${CMAKE_VERSION}-Linux-x86_64.sh --prefix=/usr/local --skip-license
-  rm cmake-${CMAKE_VERSION}-Linux-x86_64.sh
-  popd
-}
-
```
lalitb, line 43: setup_cmake.sh script installing only the googletest - this seems confusing.
For the sake of consistency with naming, should we also add `apt-get install cmake` in this script ?

lalitb (approved):
LGTM with suggestion.

ci/setup_cmake.sh
marcalff: This script is under `ci/`.

There is already another script, under `tools/setup-cmake.sh`, to also install cmake.

If the concern is that someone taking a dependency in `ci/*` will be affected, I can add a note in the changelog.

marcalff: Added changelog, and renamed to setup_googletest.sh

PR #2633, opened 2024-04-09
Upgrade to semantic convention 1.25.0

esigo:
Fixes #2614  (issue)

## Changes

upgrades semantic convention 1.25.0

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #2640, opened 2024-04-17
[SEMANTIC CONVENTIONS] Rework on semantic conventions 1.25.0

marcalff:
Fixes # (issue)

## Changes

This fix is a follow up on #2633.

Context:

* https://github.com/open-telemetry/semantic-conventions/pull/758
* Service attributes such as `service.name` are no longer defined in resources.
* Instead, these attributes are defined in the `registry` as attribute_groups.
* The resource model contains a reference to `service.name` defined in the registry

When generating resource semantic conventions,
the code generation does not fail when a unknown reference to `service.name` is found,
which seems a bug for build tools, and the semantic convention `opentelemetry::sdk::resource::SemanticConventions::kServiceName` is not generated.

This causes a breaking change for instrumented applications.

Changes:

This fix changes the command used to generate resource semantic conventions,
to use `resource,attribute_group` instead of `resource`, so that attributes groups definitions are available
to the code generation.

When invoking the code generator with `resource,attribute_group`,
the symbol `opentelemetry::sdk::resource::SemanticConventions::kServiceName` is generated.

With this fix, instrumented application code is unchanged (previous changes using `trace::SemanticConventions::kServiceName` as a work around have been reverted)

### DIFF
--- a/buildscripts/semantic-convention/generate.sh
+++ b/buildscripts/semantic-convention/generate.sh
@@ -80,7 +80,7 @@ docker run --rm \
   -v ${SCRIPT_DIR}/templates:/templates${USE_MOUNT_OPTION} \
   -v ${ROOT_DIR}/sdk/include/opentelemetry/sdk/resource/:/output${USE_MOUNT_OPTION} \
   otel/semconvgen:$GENERATOR_VERSION \
-  --only resource \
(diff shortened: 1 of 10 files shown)


esigo (approved):
LGTM
Thanks for the fix

PR #2642, opened 2024-04-19
[bazel] Enable --incompatible_disallow_empty_glob

keith:
This way downstream consumers can enable this flag as well


malkia:
I have a list in my soft fork, with few more incompatbiles that were enabled (whatever I can) - https://github.com/malkia/opentelemetry-cpp/blob/main/.bazelrc#L92 - although that specifically did not work - I think it was for the prometheus-cpp module?

keith:
depends on https://github.com/jupp0r/prometheus-cpp/pull/715 and probably some other stuff

marcalff:
> depends on [jupp0r/prometheus-cpp#715](https://github.com/jupp0r/prometheus-cpp/pull/715) and probably some other stuff

This PR in prometheus-cpp:

* https://github.com/jupp0r/prometheus-cpp/pull/715

is merged, but there is no release yet that includes it.

Opentelemetry-cpp is using the prometheus-cpp 1.2.4 tag.

Waiting on release 1.2.5 from prometheus.

marcalff:
Related:

* https://github.com/jupp0r/prometheus-cpp/issues/722

keith:
will rebase after https://github.com/open-telemetry/opentelemetry-cpp/pull/3162

marcalff (approved):
LGTM, thanks for the fix.

PR #2645, opened 2024-04-22
[CI] Simplify bazel CI setup

keith:
Doing bazel test //... is the same as doing a build first and then test. In some cases it might be better to do directly for increased parallelism.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -10,6 +10,9 @@ common --enable_platform_specific_config
 # Needed by gRPC to build on some platforms.
 build --copt -DGRPC_BAZEL_BUILD
 
+# Reduce output from bazel test
+build --test_output=errors
+
 # Set minimum supported C++ version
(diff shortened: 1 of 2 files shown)


marcalff (changes requested):
Thanks for the PR.

I have the following concern, please clarify what the behavior will be with the fix.

Before the change:

* all the code is built
* then all the tests are axecuted

In case of a build failure, it is guaranteed to be reported by the first step, before running tests.

This is in particular important for maintainers when inspecting CI logs, because investigating a test failure for test A is a waste of time if test B does not even build.

By separating build from test execution, it gets easier to resolve issues on a PR while making progress.

After the change:

Not sure what the behavior will be, and in particular if CI can report a test failure for A and hide the fact that test B does not even build, in case CI fails fast.

Please clarify how it will work with the change.

Bazel build + test is similar to CMake make + make test, so me personal preference would be to keep the code as is, unless there is a benefit I missed with this change.

PR #2652, opened 2024-04-29
[BUILD] CMakeLists.txt: Enable CMAKE_MSVC_RUNTIME_LIBRARY support

t-b:
The documentation for CMAKE_MSVC_RUNTIME_LIBRARY states [1]:

> This variable has effect only when policy CMP0091 is set to NEW prior to
> the first project() or enable_language() command that enables a language
> using a compiler targeting the MSVC ABI.

so the current usage of CMAKE_MSVC_RUNTIME_LIBRARY for vcpkg does not work at all.

Let's fix that by setting policy 91 to new if present.

[1]: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html


lalitb (approved):

lalitb (commented):
Sorry, removing the approval for the CI failure. Probably we should set the policy only if vcpkg is used (i.e, VCPKG_TOOLCHAIN is set)? cc @ThomsonTan

t-b:
> Sorry, removing the approval for the CI failure. Probably we should set the policy only if vcpkg is used (i.e, VCPKG_TOOLCHAIN is set)? cc @ThomsonTan

My usecase is not related to vcpkg, I'm just compiling with MSVC.

ThomsonTan:
The CI should have been fixed by https://github.com/open-telemetry/opentelemetry-cpp/pull/2696.

ThomsonTan (approved):

PR #2661, opened 2024-05-07
[SDK] Fix memory leak in TlsRandomNumberGenerator() constructor

### LINKED ISSUE #2660: TlsRandomNumberGenerator() can cause memory leaks
In my program, I found that the memory would continue to increase under multi-threading, and I suspected a memory leak.

In order to verify this problem, I clone this repository and wrote a `gtest` test case:
```
#include <vector>
#include <memory>
#include <thread>
void gen_traceid() {
  for (int i = 0; i < 10; ++i){
    Random::GenerateRandom64();
  }
}
TEST(RandomTest, MemoryMultiThreadGenerateRandom64) {
  printf("memory leak test");
  while (1)
  {
    std::vector<std::shared_ptr<std::thread>> vcths;
    for (int i=0; i< 10; ++i)
    {
      std::shared_ptr<std::thread> th(new std::thread(gen_traceid));
      vcths.push_back(th);
    }
    for (int i = 0; i < 10; ++i) {
      vcths[i]->join();
    }
    vcths.clear();
  }
}
```
I wrote it in `sdk/test/common/random_test.cc`, compiled it to generate `random_test`, and used `./random_test` to run it. Then I used the `top` command to observe that its memory usage will continue to increase.

The reason for this problem is that there is 
```
static thread_local TlsRandomNumberGenerator random_number_generator{};
```
in the code. When a new thread is created, a new `TlsRandomNumberGenerator()` instance will be created, and the constructor of this class will execute `::pthread_atfork(prepare , parent, child);` Memory continues to increase due to multiple executions of `pthread_atfork` .

I tried to delete `platform::AtFork` in the `TlsRandomNumberGenerator()` constructor function, and the memory usage was normal and would not increase.

hongweipeng:
Fixes #2660

## Changes


sdk/src/common/random.cc
```diff
@@ -26,12 +27,17 @@ class TlsRandomNumberGenerator
   TlsRandomNumberGenerator() noexcept
   {
     Seed();
-    platform::AtFork(nullptr, nullptr, OnFork);
+    if (!flag.test_and_set())
```
ThomsonTan, line 30: Is it possible to add a test for this?

hongweipeng: Hello, unit test has been added.

owent (approved):
LGTM

PR #2662, opened 2024-05-07
[api/logs] Return NoopLogRecord from NoopLogger

### LINKED ISSUE #2656: [API] NoopLogger causes segfault
**Describe your environment** v1.14.2

**Steps to reproduce**
```
#include "opentelemetry/logs/provider.h"

int main(int argc, char* argv[]) {
  auto logs_provider = opentelemetry::v1::logs::Provider::GetLoggerProvider();
  auto logger = logs_provider->GetLogger("test");
  auto sample = logger->CreateLogRecord();
  sample->SetAttribute("some key", "some value");
  logger->EmitLogRecord(std::move(sample));
  return 0;
}
```

Program should run successfully.

SEGV at `sample->SetAttribute` because `sample` is null.

**Additional context**
https://github.com/open-telemetry/opentelemetry-cpp/blob/da8e377be51edadad514d31c58e0ce7a2eb1b05e/api/include/opentelemetry/logs/noop.h#L37

The NoopLogger is supposed to return __usable__ objects, not a `nullptr`.

yurishkuro:
Fixes #2656

## Changes

NoopLogger was incorrectly returning nullptr, causing segfaults in client code. After this change it returns a dummy no-op log record.

### DIFF
--- a/api/include/opentelemetry/logs/noop.h
+++ b/api/include/opentelemetry/logs/noop.h
@@ -34,11 +34,34 @@ class NoopLogger final : public Logger
 public:
   const nostd::string_view GetName() noexcept override { return "noop logger"; }
 
-  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override { return nullptr; }
+  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override {
+    return nostd::unique_ptr<LogRecord>(&record_);
+  }
 
   using Logger::EmitLogRecord;
 
   void EmitLogRecord(nostd::unique_ptr<LogRecord> &&) noexcept override {}
+
+private:
+  class NoopLogRecord : public LogRecord {
+  public:
+    void SetTimestamp(common::SystemTimestamp timestamp) noexcept override {};
+    void SetObservedTimestamp(common::SystemTimestamp timestamp) noexcept override {};
+    void SetSeverity(logs::Severity severity) noexcept override {};
+    void SetBody(const common::AttributeValue &message) noexcept override {};
+    void SetAttribute(nostd::string_view key,
+                              const common::AttributeValue &value) noexcept override {};
+    void SetEventId(int64_t id, nostd::string_view name = {}) noexcept override {};
+    void SetTraceId(const trace::TraceId &trace_id) noexcept override {};
+    void SetSpanId(const trace::SpanId &span_id) noexcept override {};
+    void SetTraceFlags(const trace::TraceFlags &trace_flags) noexcept override {};
+
+    // In CreateLogRecord we return unique_ptr to the same record_ object,
+    // so we do not want to delete it when the unique_ptr is destroyed.
+    void operator delete(void*) noexcept {}
+  };
+
+  NoopLogRecord record_;
 };
 
 /**
--- a/api/include/opentelemetry/logs/noop.h
+++ b/api/include/opentelemetry/logs/noop.h
@@ -51,7 +51,7 @@ private:
     void SetBody(const common::AttributeValue & /* message */) noexcept override {}
     void SetAttribute(nostd::string_view /* key */,
                               const common::AttributeValue & /* value */) noexcept override {}
-    void SetEventId(int64_t id, nostd::string_view /* name */) noexcept override {}
+    void SetEventId(int64_t /* id */, nostd::string_view /* name */) noexcept override {}
     void SetTraceId(const trace::TraceId & /* trace_id */) noexcept override {}
     void SetSpanId(const trace::SpanId & /* span_id */) noexcept override {}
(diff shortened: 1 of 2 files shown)


api/include/opentelemetry/logs/noop.h
```diff
@@ -34,11 +34,34 @@ class NoopLogger final : public Logger
...
+    void SetTimestamp(common::SystemTimestamp timestamp) noexcept override {};
+    void SetObservedTimestamp(common::SystemTimestamp timestamp) noexcept override {};
+    void SetSeverity(logs::Severity severity) noexcept override {};
+    void SetBody(const common::AttributeValue &message) noexcept override {};
+    void SetAttribute(nostd::string_view key,
+                              const common::AttributeValue &value) noexcept override {};
+    void SetEventId(int64_t id, nostd::string_view name = {}) noexcept override {};
+    void SetTraceId(const trace::TraceId &trace_id) noexcept override {};
+    void SetSpanId(const trace::SpanId &span_id) noexcept override {};
+    void SetTraceFlags(const trace::TraceFlags &trace_flags) noexcept override {};
```
marcalff, line 57:
```suggestion
    void SetTimestamp(common::SystemTimestamp /* timestamp */) noexcept override {}
    void SetObservedTimestamp(common::SystemTimestamp /* timestamp */) noexcept override {}
    void SetSeverity(logs::Severity /* severity */) noexcept override {}
    void SetBody(const common::AttributeValue & /* message */) noexcept override {}
    void SetAttribute(nostd::string_view /* key */,
                              const common::AttributeValue & /* value */) noexcept override {}
    void SetEventId(int64_t id, nostd::string_view /* name */) noexcept override {}
    void SetTraceId(const trace::TraceId & /* trace_id */) noexcept override {}
    void SetSpanId(const trace::SpanId & /* span_id */) noexcept override {}
    void SetTraceFlags(const trace::TraceFlags & /* trace_flags */) noexcept override {}
```

This should fix failures in maintainer mode (-Wall -Werr), not tested.

api/include/opentelemetry/logs/noop.h
```diff
@@ -34,11 +34,34 @@ class NoopLogger final : public Logger
 public:
   const nostd::string_view GetName() noexcept override { return "noop logger"; }
 
-  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override { return nullptr; }
+  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override {
+    return nostd::unique_ptr<LogRecord>(&record_);
+  }
```
marcalff, line 39: This uses a member of the NoopLogger itself, which may go away, leaving the record pointer dangling.

Possible improvement:

```suggestion
  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override {
    static NoopLogRecord dummy;
  
    return nostd::unique_ptr<LogRecord>(&dummy);
  }
```

The static will never go out of scope.

marcalff: Yes, as long as it is inside a method. A static member in a class will not work.

Trying this at the moment, and still have a runtime failure in the 2 failing tests.

marcalff: From experiments, faking the new and delete is only part of the issue.

When a log record is created, used in a unique_ptr, and when the unique_ptr is destroyed,
the object pointed to is destroyed (destructor) and then memory is reclaimed (delete).

When the destructor is called on the dummy shared object, the virtual table is damaged,
so that reusing the same object for another log record leads to a crash (pure virtual method called).

Trying with a Deleter in the unique_ptr.

marcalff: And of course:

* `std::unique_ptr<LogRecord>`
* `std::unique_ptr<LogRecord, Deleter>`

are two different classes, one can not assign the second to the first, so there is no way to return a unique pointer with a custom deleter for the `NoopLogRecord`.

The conclusion so far is that unique_ptr(& dummy) owns the dummy object, and happily trashes it including the virtual table when the unique pointer goes out of scope. This prevents reusing the same dummy object for different log records.

marcalff (changes requested):
Thanks for the PR.

The idea to share a pointer on a dummy noop record, to avoid new + delete, unfortunately does not work, and causes the runtime failures seen in CI.

When logging 2 events in a row:
* CreateLogRecord() for event 1 returns a dummy record
* Delete log_record for event 1 calls the destructor, which invalidates the dummy record virtual table
* CreateLogRecord() for event 2 returns a damaged dummy record
* record->SetXXX() crashes when invoking virtual methods.

Implementing operator new() and operator delete() instead only moves the problem elsewhere, because the same crash will happen when 2 threads use the same dummy record concurrently.

yurishkuro:
@marcalff then I suggest going back to allocating a new Noop record, that at least prevents the existing behavior that returns nullptr unexpectedly, and address performance issue separately.

PR #2663, opened 2024-05-08
[EXPORTER] General cleanup for is_shutdown_ flags in exporters.

marcalff:
This is a follow up on PR #2553, to simplify `bool is_shutdown_` in general.

## Changes

* Replace every use of a spin lock + boolean by an `std::atomic<bool>`
* Replace a spin lock by a mutex, when serializing to a file in the ostream metric exporter
* Implemented the missing call to `sout_.flush()` in `OStreamMetricExporter::ForceFlush()`

### DIFF
--- a/exporters/ostream/include/opentelemetry/exporters/ostream/metric_exporter.h
+++ b/exporters/ostream/include/opentelemetry/exporters/ostream/metric_exporter.h
@@ -3,10 +3,11 @@
 
 #pragma once
 
+#include <atomic>
 #include <iostream>
+#include <mutex>
 #include <string>
 
-#include "opentelemetry/common/spin_lock_mutex.h"
 #include "opentelemetry/sdk/metrics/data/metric_data.h"
 #include "opentelemetry/sdk/metrics/export/metric_producer.h"
 #include "opentelemetry/sdk/metrics/instruments.h"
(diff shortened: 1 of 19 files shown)


lalitb (approved):
Thanks for the cleanup.

exporters/ostream/include/opentelemetry/exporters/ostream/metric_exporter.h
```diff
@@ -72,8 +73,8 @@ class OStreamMetricExporter final : public opentelemetry::sdk::metrics::PushMetr
 
 private:
   std::ostream &sout_;
-  bool is_shutdown_ = false;
-  mutable opentelemetry::common::SpinLockMutex lock_;
+  std::atomic<bool> is_shutdown_{false};
+  std::mutex serialize_lock_;
```
ThomsonTan, line 77: Why not using SpinLockMutext here?

lalitb: We were unfortunately using the `lock_` not just for shutdown :)

marcalff: The SpinLockMutex was also used to serialize when writing to sout_.

When the disk is full, a write or flush can hold forever, and using spin locks can make threads pilling up in CPU loops, which is very bad.

A mutex is more appropriate.

PR #2664, opened 2024-05-09
[API/SDK] Provider cleanup

### LINKED ISSUE #2506: [EXAMPLES] Improve ForceFlush call to also work with NoopTracerProvider
Yes, if you copy and paste code from the examples, which most users will do, you end up with the code from cleanupTracer:

```cpp
void CleanupTracer()
{
  // We call ForceFlush to prevent to cancel running exportings, It's optional.
  opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider> provider =
      trace::Provider::GetTracerProvider();
  if (provider)
  {
    static_cast<trace_sdk::TracerProvider *>(provider.get())->ForceFlush();
  }

  std::shared_ptr<opentelemetry::trace::TracerProvider> none;
  trace::Provider::SetTracerProvider(none);
}
```

Which is code that does not work when called before any initialization.  
This may happen at least in unit tests of client code (happened to me and made me search for the best solution for 1h).  
Also, it is better to call ForceFlush with a time argument so that people know they can use it, and you won't have issues posted like "why does my app not shutdown?".

I would like the example code to work in all cases, inited or not.

Changing from static_cast to dynamic cast and checking if the provider exists fixes this,
Also adding a 10s timeout to ForceFlush should cover typical cases gracefully.
```cpp
void CleanupTracer()
{
  // We call ForceFlush to prevent cancelling running exports, It's optional.
  opentelemetry::nostd::shared_ptr<opentelemetry::trace::TracerProvider> provider =
      trace::Provider::GetTracerProvider();
    if (provider)
    {
        trace_sdk::TracerProvider* tracerProvider = dynamic_cast<trace_sdk::TracerProvider*>(provider.get());
        if ( tracerProvider) {
            // Remove arg if you want to wait forever
            tracerProvider->ForceFlush(std::chrono::milliseconds(10000));
        }
    }
  std::shared_ptr<opentelemetry::trace::TracerProvider> none;
  trace::Provider::SetTracerProvider(none);
}
```

If you would like, I can provide the PR for the 4 examples that i found that have this code:
https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/examples/otlp/http_log_main.cc#L77  
https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/examples/otlp/grpc_log_main.cc#L61  
https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/examples/otlp/http_main.cc#L51
https://github.com/open-telemetry/opentelemetry-cpp/blob/a2323281adf60de8351c64861f313137bbbcebfd/examples/otlp/grpc_main.cc#L45

marcalff:
Fixes #2506

## Changes

* Removed methods related to `Flush()` and `Close()` from the `API` class `opentelemetry::trace::Tracer`, starting with ABI version 2.
  * These methods are not intended for the API, and got exposed as an oversight
  * These methods belong to the SDK instead
* Changed all `XXXProviderFactory::Create()` methods to return an `SDK` object instead of an `API` object.
  * Returning an SDK object is necessary for the application code to perform a proper initialisation, and proper shutdown of the SDK
  * In particular, returning an SDK object is required for the caller to invoke `Flush()` and `Shutdown()`.
* Removed `static_cast<SDK*>(API object)` from examples, and use cleaner code, to initialize SDK providers.
* Because the SDK change is breaking, provided a compile flag and a migration path. This allows applications:
  * To upgrade to the latest opentelemetry-cpp immediately, without code changes
  * To plan for code adjustments at a later date.


lalitb:
Good catch :)

marcalff:
Waiting for the build to finish to make sure DLL links, then this will be ready for review.

marcalff:
Add help-wanted label:

The windows DLL build fails with undefined symbols, and I can not figure out why (I don't have windows locally to investigate).

CHANGELOG.md
```diff
@@ -18,6 +18,52 @@ Increment the:
...
+  * After this fix, these methods return an SDK level object
+    (opentelemetry::sdk::trace::TracerProvider) to the caller.
+  * Returning an SDK object is necessary for the application to
+    cleanup and invoke SDK level methods, such as ForceFlush(),
+    on a provider.
+  * The application code that configures the SDK, by calling
+    the various provider factories, may need adjustment.
+  * All the examples have been updated, and in particular no
+    longer perform static_cast do convert an API object to an SDK object.
+    Please refers to examples for guidance on how to adjust.
```
marcalff, line 65:
```suggestion
    Please refer to examples for guidance on how to adjust.
```

sdk/src/logs/logger_provider_factory.cc
```diff
@@ -12,42 +14,42 @@ namespace sdk
 namespace logs
 {
 
-std::unique_ptr<opentelemetry::logs::LoggerProvider> LoggerProviderFactory::Create(
+std::unique_ptr<opentelemetry::sdk::logs::LoggerProvider> LoggerProviderFactory::Create(
```
lalitb, line 17: Though returning the `sdk::\<signal\>provider` is the right approach, not sure if we should introduce the API-breaking changes in SDK without deprecating, specifically as it may introduce code change in the application. Probably finding the way to deprecate, and keep the same method name `Create`.

marcalff: Implemented a deprecation flag, so that users can adjust the code at their own pace.

I think we should remove this flag very soon, by 1.17.0, otherwise the codebase will rot with too many ifdef.

Providing an easier migration path to avoid forcing code changes overnight is one thing, but this is still a **_migration path_**, as the old code is meant to be removed and can not stay forever.

lalitb (approved):
Thanks for the fix.

PR #2668, opened 2024-05-14
[API] Return NoopLogRecord from NoopLogger

### LINKED ISSUE #2656: [API] NoopLogger causes segfault

marcalff:
Fixes #2656

Based on PR contribution #2662 from @yurishkuro

## Changes

NoopLogger was incorrectly returning nullptr, causing segfaults in client code. After this change it returns a dummy no-op log record.

### DIFF
--- a/api/include/opentelemetry/logs/noop.h
+++ b/api/include/opentelemetry/logs/noop.h
@@ -38,7 +38,7 @@ public:
   {
     /*
      * Do not return memory shared between threads,
-     * a new + delete for each noop record can not be avoided,
+     * a `new` + `delete` for each noop record can not be avoided,
      * due to the semantic of unique_ptr.
      */
     return nostd::unique_ptr<LogRecord>(new NoopLogRecord());
--- a/api/include/opentelemetry/logs/noop.h
+++ b/api/include/opentelemetry/logs/noop.h
@@ -34,11 +34,39 @@ class NoopLogger final : public Logger
 public:
   const nostd::string_view GetName() noexcept override { return "noop logger"; }
 
-  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override { return nullptr; }
+  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override
+  {
+    /*
+     * Do not return memory shared between threads,
+     * a new + delete for each noop record can not be avoided,
+     * due to the semantic of unique_ptr.
+     */
+    return nostd::unique_ptr<LogRecord>(new NoopLogRecord());
+  }
 
   using Logger::EmitLogRecord;
 
   void EmitLogRecord(nostd::unique_ptr<LogRecord> &&) noexcept override {}
+
+private:
+  class NoopLogRecord : public LogRecord
+  {
+  public:
+    NoopLogRecord()           = default;
+    ~NoopLogRecord() override = default;
+
+    void SetTimestamp(common::SystemTimestamp /* timestamp */) noexcept override {}
+    void SetObservedTimestamp(common::SystemTimestamp /* timestamp */) noexcept override {}
+    void SetSeverity(logs::Severity /* severity */) noexcept override {}
+    void SetBody(const common::AttributeValue & /* message */) noexcept override {}
+    void SetAttribute(nostd::string_view /* key */,
+                      const common::AttributeValue & /* value */) noexcept override
+    {}
+    void SetEventId(int64_t /* id */, nostd::string_view /* name */) noexcept override {}
+    void SetTraceId(const trace::TraceId & /* trace_id */) noexcept override {}
+    void SetSpanId(const trace::SpanId & /* span_id */) noexcept override {}
+    void SetTraceFlags(const trace::TraceFlags & /* trace_flags */) noexcept override {}
+  };
 };
 
 /**
--- a/api/test/logs/logger_test.cc
+++ b/api/test/logs/logger_test.cc
@@ -28,9 +28,11 @@ TEST(Logger, GetLoggerDefault)
   auto lp = Provider::GetLoggerProvider();
   const std::string schema_url{"https://opentelemetry.io/schemas/1.11.0"};
   auto logger = lp->GetLogger("TestLogger", "opentelelemtry_library", "", schema_url);
-  auto name   = logger->GetName();
   EXPECT_NE(nullptr, logger);
+  auto name   = logger->GetName();
+  auto record = logger->CreateLogRecord();
   EXPECT_EQ(name, "noop logger");
+  EXPECT_NE(nullptr, record);
 }
 
 // Test the two additional overloads for GetLogger()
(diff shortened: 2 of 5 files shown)


yurishkuro:
> Based on PR contribution #2662 from @yurishkuro

Nit: I personally don't care but when you have new contributors (e.g. someone just starting in oss) I always try to preserve their own commits, e.g. by branching off of their branch, so that they get proper credit as contributors.  E.g. GitHub shows all authors:
image

yurishkuro (approved):

marcalff:
> > Based on PR contribution #2662 from @yurishkuro
> 
> Nit: I personally don't care but when you have new contributors (e.g. someone just starting in oss) I always try to preserve their own commits, e.g. by branching off of their branch, so that they get proper credit as contributors.

Good point, I need to improve my github skills to do that.

Now that 2662 is closed, the branch is deleted so I can't reopen it to use it.

@yurishkuro Feel like submitting a 1 line change using the web interface, to get the the git credits fixed ?

marcalff:
Do not merge label, waiting for additional commit.

Ready for review.

yurishkuro:
> Feel like submitting a 1 line change using the web interface, to get the git credits fixed ?

no it's fine, like I said, I don't care, it was more of a general comment on a nicer way of accepting contributions.

marcalff:
> > Feel like submitting a 1 line change using the web interface, to get the git credits fixed ?
> 
> no it's fine, like I said, I don't care, it was more of a general comment on a nicer way of accepting contributions.

Thanks. Point taken.

api/include/opentelemetry/logs/noop.h
```diff
@@ -34,11 +34,39 @@ class NoopLogger final : public Logger
 public:
   const nostd::string_view GetName() noexcept override { return "noop logger"; }
 
-  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override { return nullptr; }
+  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override
+  {
+    /*
+     * Do not return memory shared between threads,
+     * a new + delete for each noop record can not be avoided,
```
marcalff, line 41:
```suggestion
     * a `new` + `delete` for each noop record can not be avoided,
```

marcalff:
To add to change set upon merge:

    Co-authored-by: Yuri Shkuro <yurishkuro@users.noreply.github.com>

api/include/opentelemetry/logs/noop.h
```diff
@@ -34,11 +34,39 @@ class NoopLogger final : public Logger
...
 
-  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override { return nullptr; }
+  nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override
+  {
+    /*
+     * Do not return memory shared between threads,
+     * a `new` + `delete` for each noop record can not be avoided,
+     * due to the semantic of unique_ptr.
+     */
+    return nostd::unique_ptr<LogRecord>(new NoopLogRecord());
```
lalitb, line 44: Not sure if I fully understand the issue, but can we instead return a unique_ptr to static NoopLogRecord instance, with a custom  deleter which does nothing. This will ensure that the static instance is not deleted when the unique_ptr goes out of scope.
This way, if the LoggerProvider is not configured in SDK, and instrumented library won't do any allocation with every log call.

lalitb (changes requested):
For - https://github.com/open-telemetry/opentelemetry-cpp/pull/2668#discussion_r1603695457

I see a perf concern here if the SDK is not configured.

api/include/opentelemetry/logs/noop.h
yurishkuro: this was explored in https://github.com/open-telemetry/opentelemetry-cpp/pull/2662 and did not work, unique_ptr damages vtable of a static instance.

yurishkuro:
> I see a perf concern here if the SDK is not configured.

I'll take perf concern over segfault. Right now the noop implementation is simply not usable.

api/test/logs/logger_test.cc
```diff
@@ -28,9 +28,11 @@ TEST(Logger, GetLoggerDefault)
   auto lp = Provider::GetLoggerProvider();
   const std::string schema_url{"https://opentelemetry.io/schemas/1.11.0"};
   auto logger = lp->GetLogger("TestLogger", "opentelelemtry_library", "", schema_url);
-  auto name   = logger->GetName();
   EXPECT_NE(nullptr, logger);
+  auto name   = logger->GetName();
```
ThomsonTan, line 32: nit, check name before `logger->CreateLogRecord`.

sdk/src/logs/logger.cc
```diff
@@ -33,12 +33,6 @@ const nostd::string_view Logger::GetName() noexcept
 
 nostd::unique_ptr<opentelemetry::logs::LogRecord> Logger::CreateLogRecord() noexcept
 {
-  // If this logger does not have a processor, no need to create a log recordable
-  if (!context_)
```
ThomsonTan, line 37: Is context_ always not null?

api/include/opentelemetry/logs/noop.h
lalitb: ok thanks for the link. Yes agree multiple delete calls from different threads would be an issue.

lalitb:
> I'll take perf concern over segfault. Right now the noop implementation is simply not usable.

Yes segfault was indeed a bug to be fixed.  The fix is unfortunately adding a perf overhead for instrumented libraries when the logging is not enabled, which is not good. Not an issue with this PR, more of the constraint added by the design. I will create a tracking issue for this to be handled separately.

lalitb (approved):

sdk/src/logs/logger.cc
marcalff: It is never null, a context is always created when building a logger provider.

api/test/logs/logger_test.cc
marcalff: Thanks, fixed.

PR #2673, opened 2024-05-21
[API] DO not allow unsafe `Logger::EmitLogRecord`

### LINKED ISSUE #2666: [API] Logger::EmitLogRecord() is unsafe to use
Consider the following code:

```
        nostd::unique_ptr<opentelemetry::logs::LogRecord> record =
            logger->CreateLogRecord();
        record->SetSeverity(otel_severity);
        opentelemetry::common::AttributeValue otel_body(message);
        record->SetBody(otel_body);
        logger->EmitLogRecord(record);
```

This code builds, executes, and produces empty log records downstream.

The issue is that this method is invoked:

```
  void EmitLogRecord(ArgumentType &&... args)
  {
    nostd::unique_ptr<LogRecord> log_record = CreateLogRecord();
    if (!log_record)
    {
      return;
    }

    EmitLogRecord(std::move(log_record), std::forward<ArgumentType>(args)...);
  }
```

The log record given in arguments is ignored, and a dummy record, not populated, is used instead.

To fix this, the user code needs to change to:

```
        logger->EmitLogRecord(std::move(record));
```

which will invoke:

```
  virtual void EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record) noexcept = 0;
```

This is a major issue, because a lot of user code will be instrumented for logs, and these mistakes will happen again and again, creating the perception that opentelemetry-cpp does not work for logs.

Code such as:

```
        logger->EmitLogRecord(record);
```

**MUST** fail to build instead of ignoring the record argument.

owent:
Fixes #2666 

## Changes

+ Do not allow unsafe  `Logger::EmitLogRecord`

### DIFF
--- a/api/include/opentelemetry/logs/logger_type_traits.h
+++ b/api/include/opentelemetry/logs/logger_type_traits.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <chrono>
+#include <memory>
 #include <type_traits>
 
 #include "opentelemetry/common/attribute_value.h"
@@ -145,6 +146,10 @@ struct LogRecordSetterTrait<common::KeyValueIterable>
 template <class ValueType>
 struct LogRecordSetterTrait
 {
+  static_assert(!std::is_same<nostd::unique_ptr<LogRecord>, ValueType>::value &&
+                    !std::is_same<std::unique_ptr<LogRecord>, ValueType>::value,
+                "unique_ptr<LogRecord> is not allowed");
+
   template <class ArgumentType,
             nostd::enable_if_t<std::is_convertible<ArgumentType, nostd::string_view>::value ||
                                    std::is_convertible<ArgumentType, common::AttributeValue>::value,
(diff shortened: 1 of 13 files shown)


api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -145,6 +146,10 @@ struct LogRecordSetterTrait<common::KeyValueIterable>
 template <class ValueType>
 struct LogRecordSetterTrait
 {
+  static_assert(!std::is_same<nostd::unique_ptr<LogRecord>, ValueType>::value &&
+                    !std::is_same<std::unique_ptr<LogRecord>, ValueType>::value,
+                "unique_ptr<LogRecord> is not allowed");
```
marcalff, line 151:
```suggestion
                "unique_ptr<LogRecord> is not allowed, use std::move()");
```

Also, consider adding an assert for `LogRecord*`

marcalff (commented):
Thanks for the fix, see suggestion.

api/include/opentelemetry/logs/logger_type_traits.h
owent: Thanks, it's done

marcalff (approved):
LGTM, thanks for the fix.

PR #2675, opened 2024-05-21
[EXPORTER] OTLP file: use `fopen`, `fwrite`, `fflush` and `fclose` which are thread-safe.

### LINKED ISSUE #2674: OtlpFileSystemBackend has ThreadSanitizer problems.
`std::ofstream::write` and `std::ofstream::flush` may not be thread-safety.

owent:
Fixes #2674 

## Changes

+ Use `fwrite`, `fflush` to keep it thread-safety.
+ Add `OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND` , `OPENTELEMETRY_SANITIZER_NO_MEMORY` , `OPENTELEMETRY_SANITIZER_NO_THREAD`, `OPENTELEMETRY_SANITIZER_NO_ADDRESS`, `OPENTELEMETRY_HAVE_BUILTIN`, `OPENTELEMETRY_HAVE_FEATURE`, `OPENTELEMETRY_HAVE_ATTRIBUTE`, `OPENTELEMETRY_HAVE_CPP_ATTRIBUTE`


marcalff (commented):
Thanks for the PR.

The macros and annotations for sanitizers look good to me.

For the thread safety issue with file I/O,
I don't understand the initial problem, and the precise root cause.

In my understanding:

* file iostream operations are thread safe, when each thread uses its own stream
* file iostream operations are not thread safe when multiple threads uses the same stream
* file io operations are thread safe, when each thread uses its own `FILE` handle
* file io operations are not thread safe (in theory), when multiple threads uses the same `FILE` handle. In practive, they may appear to be safe if the implementation serializes calls.

Changing iostream to file io does not look like a proper fix for thread safety:

* If several threads can write to the same handle, do we need a mutex to serialize operations to write completely a record then ?

In particular, with:

```
  fwrite(data.data(), 1, data.size(), out.get());
  fputc('\n', out.get());
```

This does not look safe, even when fwrite() and fputc() are safe, if there are really several threads executing this code concurrently for the same handle (both records will be intertwined: 2 data, then 2 end of line).

Changing file stream with plain file io may be ok (even desirable) otherwise, but I don't understand how this is a fix for thread safety: please clarify.

owent:
> * file iostream operations are not thread safe when multiple threads uses the same stream

file iostream operations are not thread safe when multiple threads uses the same stream(we got data race from `std::basic_streambuf<CharT,Traits>::xsputn` when using tsan). But `fwrite` and `fflush` are thread safe when multiple threads uses the same FILE handle.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
@esigo Hi Ehsan. Any comments for the code review ?

esigo (approved):
LGTM
thanks for the PR :)

Issue #2619, opened 2024-03-28
Upgrade to abseil 20240116.1 (Bazel only)


marcalff:
Upgrade to abseil 20240116.1 for bazel.

keith:
have a working version in https://github.com/open-telemetry/opentelemetry-cpp/pull/2679

PR #2679, opened 2024-05-23
[bazel] Bump version and deps

keith:
This updates bazel as well as the deps that were required to be updated in order to support the newer version of bazel.

Fixes https://github.com/open-telemetry/opentelemetry-cpp/issues/2619


keith:
The DocFX failure here was a 500 from something, can someone with access restart just that job? otherwise I can re-push an empty commit

.bazelrc
```diff
@@ -4,12 +4,18 @@
...
+
 # Enable automatic configs based on platform
 common --enable_platform_specific_config
 
 # Needed by gRPC to build on some platforms.
 build --copt -DGRPC_BAZEL_BUILD
 
+# Workaround abseil libraries missing symbols
+build:windows --linkopt=/FORCE
+
```
marcalff, line 18: How is this a workaround ?

If I understand correctly:

https://learn.microsoft.com/en-us/cpp/build/reference/force-force-file-output?view=msvc-170#remarks

this will just crash at runtime.

keith: My understanding for our use case in tests here is that windows does not support no-deps cc libraries, which bazel is relying on, so we're skipping the issues at link time, which will never be satisfied, but then since they exist at runtime, the tests pass.

marcalff: Which test in particular is affected by this ?

If this is about the api singleton test (which is not executed in windows anyway),
I would rather remove this work around, and not build the singleton test under windows+bazel at all,
or even not under bazel at all, if this helps.

keith: it is about that one being _built_ on windows https://github.com/keith/opentelemetry-cpp/blob/9fedbc258e5e101d81ee889b8f7ad33574c5440c/api/test/singleton/BUILD, so yea I can disable building this on windows instead if that's what you'd prefer

keith: pushed!

bazel/repository.bzl
```diff
@@ -55,54 +46,38 @@ def opentelemetry_cpp_deps():
...
         http_archive,
-        name = "com_github_grpc_grpc_legacy",
-        sha256 = "024118069912358e60722a2b7e507e9c3b51eeaeee06e2dd9d95d9c16f6639ec",
-        strip_prefix = "grpc-1.39.1",
-        urls = [
-            "https://github.com/grpc/grpc/archive/v1.39.1.tar.gz",
-        ],
+        name = "build_bazel_apple_support",
+        sha256 = "c4bb2b7367c484382300aee75be598b92f847896fb31bbd22f3a2346adf66a80",
+        url = "https://github.com/bazelbuild/apple_support/releases/download/1.15.1/apple_support.1.15.1.tar.gz",
```
marcalff, line 63: Please clarify with comments why we need this.

Is this for extra helpers, or is it an issue for some platforms like MacOs, or ???

Also, please expand on "Load this before grpc to pull a newer version"

The goal is that people not familiar with bazel (i.e., me) should at least understand the basics, to be able to maintain this code the next time, say, grpc or abseil is upgraded for opentelemetry-cpp.

Likewise for rules_apple.3.5.1

marcalff (approved):
LGTM, thanks for the fix.

To be honest, bazel just feel like a black whole to otel-cpp maintainers currently,
which makes it challenging to maintain, so this PR is all the more appreciated.

Please add comments about `apple_support` and `rules_apple`, it will go a very long way to help us.

marcalff:
Don't pay attention to CI breaks, this is a general github issue.

api/BUILD
```diff
@@ -28,10 +23,7 @@ string_flag(
 cc_library(
     name = "api",
     hdrs = glob(["include/**/*.h"]),
-    defines = select({
-        ":with_external_abseil": ["HAVE_ABSEIL"],
-        "//conditions:default": [],
-    }) + select({
+    defines = ["HAVE_ABSEIL"] + select({
```
lalitb, line 26: Does it mean we won't be able to build the otel-cpp with the internal abseil snapshot?

lalitb: It appears that we now always use the external abseil  or stl version of `variant` type, and not the internal snapshot with bazel. This should be fine, if we can document it in (probably) [INSTALL.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/INSTALL.md#build-instructions-using-bazel)

lalitb (approved):
LGTM. with nit comment on documenting the abseil requirement.

ci/do_ci.ps1
```diff
@@ -10,6 +10,8 @@ $nproc = (Get-ComputerInfo).CsNumberOfLogicalProcessors
 
 $SRC_DIR = (Get-Item -Path ".\").FullName
 
+# Workaround https://github.com/bazelbuild/bazel/issues/18683
+$BAZEL_STARTUP_OPTIONS = "--output_base=C:\Out"
```
ThomsonTan, line 14: Is this a hard-coded path for CI environment?

keith: yes, the paths are apparently getting too long for windows so this is a reasonable workaround for CI

api/BUILD
keith: this is covered in dependencies.md fwiw

bazel/repository.bzl
keith: done

keith:
addressed all the comments here!

api/BUILD
marcalff: INSTALL.md probably needs more cleanup and revision, as it seem outdated.

I suggest to address it in a separate PR, to avoid delaying this change.

keith:
Thanks all!

PR #2691, opened 2024-06-03
[ETW] Add table name mapping for Logs other than the default Log table

### LINKED ISSUE #2680: [Proposal] Make table name configurable in OpenTelemetry C++ ETW exporter
In the current OpenTelemetry ETW Log exporter, the attribute for table name is hard-coded to `"Log"` (see below 2 links for more details). The table name is useful for grouping events in the backend, especially for the Logging signal. Based on that the ETW exporter is header-only, this table name can only be changed by patching the source code and recompiling the source code which could be inconvenient because the same source file can only have one table name defined.

https://github.com/open-telemetry/opentelemetry-cpp/blob/78d488ca44d6a9aed79e63da739b081e20c94207/exporters/etw/include/opentelemetry/exporters/etw/etw_fields.h#L128

https://github.com/open-telemetry/opentelemetry-cpp/blob/78d488ca44d6a9aed79e63da739b081e20c94207/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h#L274

Here are the 2 proposals about making this table name configurable at launch/runtime.

## 1. Reuse the library_name as table name

OpenTelemetry API has library name in [GetLogger](https://github.com/open-telemetry/opentelemetry-cpp/blob/78d488ca44d6a9aed79e63da739b081e20c94207/api/include/opentelemetry/logs/logger_provider.h#L41), but it is not used in ETW exporter (see [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/78d488ca44d6a9aed79e63da739b081e20c94207/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h#L368)). So library_name can be mapped to table name under explicit enlightenment in the provider option. The config code looks like below.

```c++
  opentelemetry::exporter::etw::TelemetryProviderOptions options {
    {"libraryNameAsTableName", true}
  };
  opentelemetry::exporter::etw::LoggerProvider logger_provider(options);
  auto logger = logger_provider.GetLogger(providerName, "my_table_name");
```

### Pros

* It doesn't break existing code for upgrading with the explicit enlightenment.
* No extra perf cost on logging besides configuration option which is only for launch time.
* It is consistent with .NET which defines the mapping in exporter [configuration](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/README.md#tablenamemappings-optional).

### Cons

* The library_name will not be able to be used for other mapping once this is enabled

## 2. Pass pre-defined attribute as table name for each Log call

We can also define a unique attribute to set the table as below.

```cpp
  logger->Info("Hello World!", common::MakeAttributes{{"unique_table_name_key", "my_event_name"}});
```

### Pros

* No need to take out library_name from unused list.

### Cons

* Extra runtime cost for each Log call.
  * Extra memory allocated for the key value pairs struct.
  * Extra CPU time on initialize the key value pairs container, and also query the unique key to extract table name in the container.
* The new convention of defining the unique key for table name causes inconsistency with other languages like .NET Geneva exporter, which does this in configuration [option](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/README.md#tablenamemappings-optional).

I prefer the option 1 above, let me know your suggestion or other options.

ThomsonTan:
Fixes #2680, this is the option 1 in the proposal with minor extension on defining a mapping for the table names. This makes it more flexible for the SDK users for configure the table name in the ETW exporter, without changing the API call in the instrumented libraries.

## Changes

### DIFF
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
@@ -145,6 +145,8 @@ class Logger : public opentelemetry::logs::Logger
    */
   std::string provId;
 
+  std::string eventName_;
+
   /**
    * @brief Encoding (Manifest, MessagePack or XML)
    */
@@ -179,10 +181,12 @@ public:
    * @param encoding ETW encoding format to use.
    */
   Logger(etw::LoggerProvider &parent,
+         nostd::string_view eventName,
          nostd::string_view providerId     = "",
          ETWProvider::EventFormat encoding = ETWProvider::EventFormat::ETW_MANIFEST)
       : opentelemetry::logs::Logger(),
         loggerProvider_(parent),
+        eventName_(eventName),
         provId(providerId.data(), providerId.size()),
         encoding(encoding),
         provHandle(initProvHandle())
@@ -271,7 +275,7 @@ public:
 #endif  // defined(ENABLE_ENV_PROPERTIES)
 
     // Populate Etw.EventName attribute at envelope level
-    evt[ETW_FIELD_NAME] = ETW_VALUE_LOG;
+    evt[ETW_FIELD_NAME] = eventName_.data();
 
 #ifdef HAVE_FIELD_TIME
     {
@@ -347,6 +351,8 @@ public:
     GetOption(options, "enableTraceId", config_.enableTraceId, true);
     GetOption(options, "enableSpanId", config_.enableSpanId, true);
     GetOption(options, "enableActivityId", config_.enableActivityId, false);
+    GetOption(options, "enableTableNameMappings", config_.enableTableNameMappings, false);
+    GetOption(options, "tableNameMappings", config_.tableNameMappings, {});
 
     // Determines what encoding to use for ETW events: TraceLogging Dynamic, MsgPack, XML, etc.
     config_.encoding = GetEncoding(options);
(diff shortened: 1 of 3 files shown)


exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
```diff
@@ -359,19 +365,30 @@ class LoggerProvider : public opentelemetry::logs::LoggerProvider
...
-    UNREFERENCED_PARAMETER(library_name);
     UNREFERENCED_PARAMETER(version);
     UNREFERENCED_PARAMETER(schema_url);
     UNREFERENCED_PARAMETER(attributes);
+
+    std::string event_name{ETW_VALUE_LOG};
+    if (config_.enableTableNameMappings)
+    {
+      auto it =
+          config_.tableNameMappings.find(std::string(library_name.data(), library_name.size()));
```
lalitb, line 382: std::string() will create a copy, can we use nostd::string_view instead? Also, unordered_map instead of map?

lalitb (approved):
Few comments for perf. LGTM once resolved.

exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
ThomsonTan: Resolve this for now as this is not in the logging hot path.

PR #2695, opened 2024-06-06
[CI] Remove benchmark overlay for vcpkg

ThomsonTan:
There is no much special config in the local vcpkg overlay of benchmark. Remove it for now and use the from the the current vcpkg.


marcalff (approved):
LGTM

PR #2699, opened 2024-06-12
[CI] Enable ARM64 build in CI

### LINKED ISSUE #1417: Enable ARM64 build in CI
ARM64 build target of opentelemetry-cpp is supported, it will be great if we have CI to track the status of ARM64 build

esigo:
Fixes #1417 (issue)

## Changes

Adds ARM CMake test to the CI.


.github/workflows/ci.yml
```diff
@@ -7,6 +7,26 @@ on:
     branches: [ main ]
 
 jobs:
+  arm_test:
+    name: CMake test on ARM (without otlp-exporter)
```
lalitb, line 11: should we do this with otlp-exporter?

lalitb (approved):

.github/workflows/ci.yml
```diff
@@ -7,6 +7,26 @@ on:
     branches: [ main ]
 
 jobs:
+  arm_test:
```
ThomsonTan, line 10: Could it be named as arm64_test? As arm_test may indicate it is 32-bit.

lalitb:
We can also try using the Github ARM runner created for otel  `otel-linux-arm64`  (refer  - [otel-linux-arm64](https://github.com/open-telemetry/community/issues/2156)

.github/workflows/ci.yml
esigo: was just testing the runner.
testing otlp now.

.github/workflows/ci.yml
esigo: renamed to `arm64_test`.

ci/setup_ci_environment.sh
```diff
@@ -7,6 +7,9 @@ set -e
 apt-get update
 apt-get install --no-install-recommends --no-install-suggests -y \
                 build-essential \
+                cmake \
+                g++-10 \
+                gcc-10 \
```
marcalff, line 12: setup_ci_environment is also used by bazel builds and clang builds.

Please define helpers like:
- setup_cmake.sh
- setup_gcc.sh (or setup_gcc10.sh)

and use them explicitly for platforms that do not provide cmake and gcc by default in the runner, like here.

It will help decouple things, allowing an easier upgrade/downgrade of gcc versions in say ubuntu independendly of arm.

ThomsonTan (approved):

ci/setup_ci_environment.sh
esigo: thanks, done.

marcalff (approved):
LGTM, thanks for the CI.

PR #2701, opened 2024-06-14
[BUILD] Fix Import Abseil-cpp

LeiZhang-Hunter:
…ink you can make a judgment when introducing abseil-cpp. If abseil is not introduced, then use find_package

Fixes # (issue)

## Changes


lalitb (approved):
LGTM. CI is failing for formatting, so need to be fixed before merging.

CMakeLists.txt
```diff
@@ -377,7 +377,10 @@ if(WITH_PROMETHEUS)
 endif()
 
 if(WITH_ABSEIL)
-  find_package(absl CONFIG REQUIRED)
+  if(TARGET absl::strings)
```
ThomsonTan, line 380: nit: put a NOT to negate the condition in `if()` so the empty if block can be avoided?

LeiZhang-Hunter: I have modified it.

marcalff (approved):
LGTM, thanks for the fix.

PR #2707, opened 2024-06-19
[Code health] Remove Unicode Text from Source files

### LINKED ISSUE #2706: Source Files contain Unicode Text
**Describe your environment** 

Branch: main (commit 25738f391bf4f7a583690d4a5c827d3803d2b6c3)

**Steps to reproduce**
```
find . -type f -name "*.h" -exec file {} \; | grep UTF-8
find . -type f -name "*.cc" -exec file {} \; | grep UTF-8
```
Source files are expected to be ASCII encoded except when unicode characters are required for tests.

```
harish@Harishs-MacBook-Air opentelemetry-cpp % find . -type f -name "*.h" -exec file {} \; | grep UTF-8  
./api/include/opentelemetry/context/context.h: C++ source text, Unicode text, UTF-8 (with BOM) text
./api/include/opentelemetry/context/runtime_context.h: C++ source text, Unicode text, UTF-8 (with BOM) text
./api/include/opentelemetry/baggage/baggage.h: C++ source text, Unicode text, UTF-8 (with BOM) text

harish@Harishs-MacBook-Air opentelemetry-cpp % find . -type f -name "*.cc" -exec file {} \; | grep UTF-8
./ext/test/http/url_parser_test.cc: c program text, Unicode text, UTF-8 text
./sdk/test/metrics/instrument_metadata_validator_test.cc: c program text, Unicode text, UTF-8 text
./opentracing-shim/src/span_shim.cc: C++ source text, Unicode text, UTF-8 text
```

**Additional context**
The headers listed above all have a BOM character at the start of the file, and `span_shim.cc` has a unicode character in a comment. I currently use an in-house build system that is built on flex which has trouble parsing UTF-8 encoded files. I think we should convert all source files to use ASCII encoding unless required. I would be happy to contribute a PR if required.

perhapsmaple:
Fixes #2706 

## Changes

Removed Unicode characters from API headers and `opentracing-shim/src/span_shim.cc`.

Escaped UTF-8 characters from test file data:

```
./ext/test/http/url_parser_test.cc: c program text, Unicode text, UTF-8 text
./sdk/test/metrics/instrument_metadata_validator_test.cc: c program text, Unicode text, UTF-8 text
```

### DIFF
--- a/api/include/opentelemetry/baggage/baggage.h
+++ b/api/include/opentelemetry/baggage/baggage.h
@@ -1,4 +1,4 @@
-﻿// Copyright The OpenTelemetry Authors
+// Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
@@ -297,3 +297,4 @@ private:
 }  // namespace baggage
 
 OPENTELEMETRY_END_NAMESPACE
+
--- a/api/include/opentelemetry/baggage/baggage.h
+++ b/api/include/opentelemetry/baggage/baggage.h
@@ -297,4 +297,3 @@ private:
 }  // namespace baggage
 
 OPENTELEMETRY_END_NAMESPACE
-
--- a/api/include/opentelemetry/context/context.h
+++ b/api/include/opentelemetry/context/context.h
@@ -1,4 +1,4 @@
-﻿// Copyright The OpenTelemetry Authors
+// Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
@@ -160,3 +160,4 @@ private:
 };
 }  // namespace context
 OPENTELEMETRY_END_NAMESPACE
+
--- a/api/include/opentelemetry/context/context.h
+++ b/api/include/opentelemetry/context/context.h
@@ -160,4 +160,3 @@ private:
 };
 }  // namespace context
 OPENTELEMETRY_END_NAMESPACE
-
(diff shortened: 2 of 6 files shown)


lalitb (approved):

marcalff:
For UTF-8 character still present, like in:

```
./ext/test/http/url_parser_test.cc: c program text, Unicode text, UTF-8 text
```

The offending source code is:

```
      {"%C3%B6%C3%A0%C2%A7%C3%96abcd%C3%84", "öà§ÖabcdÄ"},
```

Instead of using the character `ö` in the expected string, which corresponds to C3B6:

```
U+00F6	ö	c3 b6	LATIN SMALL LETTER O WITH DIAERESIS
```

could the code be changed to use unicode escape sequences like \u00F6 instead, so all the code is UTF-8 clean.

marcalff (commented):
Thanks for the fix.

See comments, to also replace UTF-8 characters present in test files.

This will affect the build when compiling with unit tests, so it is desirable to also clean test code, even if used only optionally.

marcalff:
> 
> Instead of using the character `ö` in the expected string, which corresponds to C3B6:
> 
> ```
> U+00F6	ö	c3 b6	LATIN SMALL LETTER O WITH DIAERESIS
> ```
> 
> could the code be changed to use unicode escape sequences like \u00F6 instead, so all the code is UTF-8 clean.

... and of course my suggestion does not work for windows / msvc for some reason (CI failures), sorry about that.

From what I could find, windows may need a `std::wstring` and a `L"\u00F6"` literal, but since no code in opentelemetry-cpp uses wstring currently, this is likely to open a can of worms.

A possible work around is to code the UTF-8 data instead : "\xC3\xB6", please try.

If this still fails, don't waste time on it, and remove changes from the two test files, these will be fixed separately.

Thanks

marcalff (approved):
LGTM, thanks for the fix.

PR #2717, opened 2024-06-25
[REMOVAL] Remove build option `WITH_DEPRECATED_SDK_FACTORY`

### LINKED ISSUE #2716: [REMOVAL] Remove build option `WITH_DEPRECATED_SDK_FACTORY`
CMake option `WITH_DEPRECATED_SDK_FACTORY` was introduced in opentelemetry-cpp 1.16.0.

This option is temporary, and used to avoid application code changes when upgrading to 1.16.0, to facilitate upgrade.

Remove option `WITH_DEPRECATED_SDK_FACTORY` for the next release.

Code currently using `WITH_DEPRECATED_SDK_FACTORY=OFF` is ready and not affected.

Code still using `WITH_DEPRECATED_SDK_FACTORY=ON` in release 1.16.0 will not build in 1.17.0. It must be fixed before upgrading to the next release.

marcalff:
Fixes #2716

## Changes

* Removed CMake option `WITH_DEPRECATED_SDK_FACTORY`
* Removed old references to CMake option `WITH_OTLP`
* Removed old references to CMake option `WITH_LOGS_PREVIEW`


marcalff:
Please review.

I suggest to wait at one or two weeks for possible comments on this (the removal issue is pinned), before merging to main.

CMakeLists.txt
```diff
@@ -204,13 +196,6 @@ if(NOT WITH_STL STREQUAL "OFF")
   endif()
 endif()
 
-if(DEFINED WITH_OTLP)
```
ThomsonTan, line 207: As this triggers fatal error, should fine to keep it?

marcalff: It is also safe to assume `WITH_OTLP` is no longer in use, because it would cause an error.

I prefer to finally remove old options that no longer exist, to simplify maintenance.

If we do not remove it now, when then ?

ThomsonTan:
Should this PR be included in the coming minor build (like 1.7.0) instead of patch build?

marcalff:
> Should this PR be included in the coming minor build (like 1.7.0) instead of patch build?

Current release is 1.16.0, so next patch build will be 1.16.1

This removal should only be part of release 1.17.0, and not part of the 1.16.1 patch, I agree.

ThomsonTan (approved):
LGTM. Just one note is to merging this after the patch release of 1.6.1.

PR #2720, opened 2024-06-26
[Bazel] Add missing BUILD file

tdat00:
## Fixes
- None

## Changes
* `etw_threads` example doesn't have `BUILD` file, we cannot run the example using Bazel as described here https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#build-and-run-code-examples
* This change adds the `BUILD` file for that example.


lalitb (approved):

marcalff (approved):
LGTM, thanks for the fix.

PR #2724, opened 2024-06-27
 [SDK] Added reserve for spans array in BatchSpanProcessor.

msiddhu:
## Changes 

  Added `.reserve` for `spans_arr` in BatchSpanProcessor (Small Optimization)
  
  Helps to allocate the amount of memory needed for number of records so that dynamic memory allocation doesn't happen in the consume method.
  
  `.push_back()` reallocates memory each time the method is called.
  
  Using `.reserve()` would avoid memory reallocation as already the memory is allocated.
  
  References:
  [C++ Vector push_back](https://cplusplus.com/reference/vector/vector/push_back/)
  [C++ Vector reserve](https://cplusplus.com/reference/vector/vector/reserve/)

### DIFF
--- a/sdk/src/logs/batch_log_record_processor.cc
+++ b/sdk/src/logs/batch_log_record_processor.cc
@@ -209,7 +209,7 @@ void BatchLogRecordProcessor::Export()
     }
 
     // Reserve space for the number of records
-    records_arr.reserve(num_records_to_export); 
+    records_arr.reserve(num_records_to_export);
     buffer_.Consume(num_records_to_export,
                     [&](CircularBufferRange<AtomicUniquePtr<Recordable>> range) noexcept {
                       range.ForEach([&](AtomicUniquePtr<Recordable> &ptr) {
--- a/sdk/src/logs/batch_log_record_processor.cc
+++ b/sdk/src/logs/batch_log_record_processor.cc
@@ -208,7 +208,7 @@ void BatchLogRecordProcessor::Export()
       break;
     }
 
-    // Reserve space for number of records and avoid reallocation of vector.
+    // Reserve space for the number of records
     records_arr.reserve(num_records_to_export); 
     buffer_.Consume(num_records_to_export,
                     [&](CircularBufferRange<AtomicUniquePtr<Recordable>> range) noexcept {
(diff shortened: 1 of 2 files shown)


sdk/src/trace/batch_span_processor.cc
```diff
@@ -206,6 +206,10 @@ void BatchSpanProcessor::Export()
       NotifyCompletion(notify_force_flush, exporter_, synchronization_data_);
       break;
     }
+
+    //reserves the space for the number of records to be exported
+    spans_arr.reserve(num_records_to_export);
+
```
marcalff, line 212:
```suggestion
    // reserves the space for the number of records to be exported
    spans_arr.reserve(num_records_to_export);

```

This should fix the CI failure on formatting.

marcalff (commented):
Looks good to me, thanks for the fix.

See the suggested change to fix the CI failure on code format.

msiddhu:
Thanks for suggesting. I think CI Checks should starting running now.

lalitb (approved):
Thanks,

lalitb:
Similar change can be done for BatchLogProcessor. @msiddhu just in case you can help as separate as same PR :)

msiddhu:
Thanks for pointing out. Made the suggested changes. Please review.

sdk/src/logs/batch_log_record_processor.cc
```diff
@@ -208,6 +208,8 @@ void BatchLogRecordProcessor::Export()
       break;
     }
 
+    // Reserve space for number of records and avoid reallocation of vector.
```
ThomsonTan, line 211: nit: make the spelling consistent between the 2 comments.

msiddhu: Resolved comments.

sdk/src/logs/batch_log_record_processor.cc
```diff
@@ -208,6 +208,8 @@ void BatchLogRecordProcessor::Export()
       break;
     }
 
+    // Reserve space for the number of records
+    records_arr.reserve(num_records_to_export);
```
marcalff, line 212:
```suggestion
    records_arr.reserve(num_records_to_export);
```

trailing whitespace.

marcalff (approved):
LGTM, thanks for the fix.

Issue #1647, opened 2022-10-03
[CI] Add a doxygen build


marcalff:
Add a doxygen build to CI.

The goal is to detect incorrect / missing doxygen documentation when merging a PR.

This will save a lot of time later, when publishing API documentation.

PR #2972, opened 2024-06-30
[DOC] Remove comment for unused LoggerProvider initialization params

herman5:
## Changes

Removes `@param` comments  for `id_generator` and `sampler` in the `LoggerProvider` constructor, which are not present in the constructor.


marcalff:
@herman5 Thanks for the fix.

Related, in case you are interested in more contributions:

* #1647

marcalff (approved):
LGTM, thanks for the fix.

PR #2985, opened 2024-07-02
[EXPORTER] Fix references in AttributeValueVisitor

meastp:
Just a small fix - I think the correct fix is `void operator()(nostd::string_view &&arg)` (i.e. without const), at least if it was intended to overload the forwarding reference function above . Might also need `void operator()(const nostd::string_view &arg)` if both variants should be covered.


lalitb (approved):

marcalff (approved):
LGTM, thanks for the fix.

Note that this code is only used for C++11.

Issue #2053, opened 2023-03-14
[Code health] clang-tidy cleanup


marcalff:
Run [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) on the code base, fix issues reported.

esigo:
Once cleaned, it's good to have a CI for this in another issue.

jensecrest:
Hello, I would like to work on this issue!

dbarker:
Hi @jensecrest, Thanks! 

I've logged a few issues: #3977, #3978, #3979, #3980 that would be a good place to start. Please take a look and comment on one you'd like to take on.

dbarker:
The Github clang-tidy action is uploading warning reports and is a good place to look for current clang-tidy warnings on main. 
https://github.com/open-telemetry/opentelemetry-cpp/actions/workflows/clang-tidy.yaml?query=branch%3Amain++

thc1006:
Big thanks to @marcalff and @lalitb for the quick turnarounds on
this thread over the last few days — the cleanup moved fast
because of your reviews.

Posting a status update + handover map.

**State**: `misc-use-internal-linkage` pool now at 98 abiv2 sites
across 63 files, down from ~135 at the start of these PRs (#4115
#4116 #4121 #4122 #4124 #4128 #4129). Distribution:

| Sites per file | File count |
|---|---|
| 4 | 2 |
| 3 | 5 |
| 2 | 19 |
| 1 | 37 |

**Wrap pattern, for anyone picking up the rest:**

```cpp
namespace
{
// existing class/struct/enum declarations stay here
}  // namespace
```

Things I tripped on (sharing so others don't):

1. **clang-tidy v22 flags enums too** — not just classes/structs.
   I missed `enum test_mode` in `func_http_main.cc` (1 site left
   after #4128 wraps the classes). Always grep `^enum` alongside
   `^class` and `^struct`.

2. **Wrap-boundary trap** for files with forward-declared static
   helpers (`static int foo();` early + `static int foo() {...}`
   after `main()`): the wrap MUST exclude the forward declarations
   OR include their definitions. Wrapping declarations alone
   creates `(anonymous namespace)::foo` symbols that fail to link
   against the file-scope `static foo` definitions. Hit this in
   pre-verify on #4128 (12+ undefined references); fixed by
   shrinking wrap to end before the forward decls.

3. **ABI-gated classes need asymmetric ratchet** (-N abiv1 / -M
[...]

Issue #2051, opened 2023-03-14
[CI] Add a clang-tidy build


marcalff:
Add a [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) build in CI.

esigo:
depends on #2053.

chusitoo:
I started looking into adding clang-tidy to the build, subject to first addressing issues found in cleanup #2053, which I also wanted to tackle to sort of get more acquainted with the rest of the codebase.

If all checks are enabled, the list of smells can get overwhelming very quickly; from the full list found at https://clang.llvm.org/extra/clang-tidy/checks/list.html, most of the cppcoreguidelines-* are a no-brainer IMHO. 
It might be worthwhile to also add performance-* and bugprone-*, although even then there would be some more fundamental issues related to things like methods marked _noexcept_ that themselves rely on functions that throw (i.e. trace_api::IsRootSpan calling `std::get(std::variant)` without catching _std::bad_variant_access_).

In fact, even cppcoreguidelines might need to be fune-tuned, as just running clang-tidy against api/include/opentelemetry/ produced 370 warnings, due to lots of smells related to macro usage and magic numbers.

Ultimately, in order to narrow down the scope of the cleanup that needs to be performed, a subset of checks to run against would have to be established first. It could always start as a very specific set of rules and expand the checks later on.

marcalff:
CI was implemented recently for include-what-you-use, see:

* https://github.com/open-telemetry/opentelemetry-cpp/blob/main/.github/workflows/iwyu.yml

In my understanding, both iwyu and clang-tidy rely on CMake and a compilation database, so it is likely that the CI for clang-tidy can be implemented following the include-what-you-use pattern.

Desirable features for the CI:

* Leverage CMAKE options `WITH_XYZ`, to control which code (OTLP_HTTP, etc) is covered
* Do not fail on the first error or warning, collect as much as possible (i.e., `make -k`)
* Collect all warnings in a log, and upload the log as artifact, so it can be inspected
* Count the number of remaining warnings. This is to enforce a build pass/fail rule later.

PR #2990, opened 2024-07-04
[Code Health] clang-tidy cleanup, part 1 

msiddhu:
Contributes to #2053 

## Changes

Added the `.clang-tidy` configuration file. Solved some warnings using auto FIX-IT and some manually. [Suppressed some warnings](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics) where needed (mostly in test cases, headers, and recursive functions).

Started with configurations that are a good starting point, inspired by [Google Cloud clang-tidy](https://github.com/googleapis/google-cloud-cpp/blob/main/.clang-tidy) and [AWS Palace clang-tidy](https://github.com/awslabs/palace/blob/main/.clang-tidy).

More checks could be enabled in further cleanups.

### Points to note:
1. A lot of _pass by value_ to _const pass by reference_.
2.  Addition and removal of _rvalue_ references where ever possible and needed.
3. String format changes such as  `.append` instead of `+`  ,  `'\n'` instead of `std::endl` , added missed comma's
4. Removing `using namespace` which are not used in source code.
5. Added exceptions  where it was needed and clang-tidy warnings where thrown.
6. Changing from recursive to iterative ( metrics circular buffer, for the other function (json) used suppress warnings)
7. `long` to `int64` [Google Coding conventions](https://google.github.io/styleguide/cppguide.html#Integer_Types)

### Cleanup in following Areas:

1. API: tests 
2. EXAMPLES  
3. EXPORTERS
     -   Elasticsearch
     -   Memory
     -   OStream
     -   OTLP
     -   Prometheus
     -   Zipkin
4. HTTP Client and its tests
5. SDK: Traces, Logs, Metrics and their tests.

Please review the differences in indented block as they'll be shown as complete block making it difficult to find differences. Hopefully this cleanup may provide some performance improvements.

## Remaining Work:

More cleanup, Integrate in CI .


lalitb:
@msiddhu Can you update the description with the rationale behind these changes? Seems some of the changes in the API could result in ABI incompatibility.

msiddhu:
I'm trying to add clang-tidy into CI. Creating a cleanup commit. I'm having tough time deciding what clang-tidy checks should be present. We'll be back in a couple of days after testing it.

marcalff:
Thanks for the cleanup.

Reopening and setting as draft, we definitively don't want this work to be forgotten ;-)

marcalff:
@msiddhu Thanks for working on this, this cleanup is much needed.

From experience in similar cleanup for warnings:

* https://github.com/open-telemetry/opentelemetry-cpp/pulls?q=is%3Apr+is%3Aclosed+author%3Amarcalff+warning

the complexity in resolving the cleanup is not so much individual changes, but making sure the overall task makes progress, including with reviews and merges.

Several activities are needed:

* fix existing warnings, sure,
* deliver fixes in workable patches, that can be reviewed,
* implement some CI to detect warnings in the code.

The clang-tidy cleanup will be an iterative process, the best is to do it by small size chunks, instead in one massive PR.

Please proceed as follows:

* (1), **do not ever use force-push**. This destroys the previous history on the PR, and resets the review to the very beginning. Do several commits instead, which is easier to review, and also easier to rollback if needed. The number of commits does not matter (see #2518), because the PR will be merged with a squash at the end.

* (2), this PR has some merge conflicts, because another cleanup is already in progress, with include-what-you-use, which also touches a lot of files. Please resolve the existing merge conflicts for this PR (it should be trivial). We can hold on on iwyu to avoid generating more conflicts for a while, to give time to clang-tidy to progress.

* (3), and it is a very weird thing to say: the content of the `.clang-tidy` file (which warning to report) is not important ... at this point. The starting list of settings looks good. What is more important is to merge the `.clang-tidy` file and have some CI to report warnings, to get the process to bootstrap. Once warnings are reported in CI, subsequent PRs can fix them, and the list of warnings to enforce can grow other time.

* (4), as @lalitb already pointed out, the content of `api/*` is very sensitive, and will take much longer to review. Because of this, changes touching `api/*` are better done in a separate PR, to avoid blocking progress for everything else. Note that changes to the api tests, as in `api/test/*` are ok, that part is not sensitive to ABI breaks, so just keep the fixes already implemented there.

* (5), reviewers need some time to look at changes, so the PR should not constantly change with more fixes, as it delays the review ever more. Since this PR already touches 100 files and 1K LOC, this is a good chunk to look at already, so please do not add more fixes to "part 1". Once part 1 is merged, more fixes can be implemented in a different PR of course.

From a quick look at the PR current content, the review should be fast for this, so I think we can aim to merge this chunk soon. Please resolve the merge conflicts and flag the PR as ready to review.

Again, thanks for taking this up.

marcalff:
Not sure why CI is stuck. First thing is to resolve conflicts anyway, the next commit should restart CI.

msiddhu:
@marcalff 
I think the hard CI check using the no-exceptions flag is failing. Please help me with this. What could the solutions be?

lalitb:
>I think the hard CI check using the no-exceptions flag is failing. Please help me with this. What could the solutions be?

Was there any need to add try-catch for [OTELResourceDetector::Detect()](https://github.com/msiddhu/opentelemetry-cpp/blob/2825d0372bd26146f0172801851c53b57fbd726f/sdk/src/resource/resource_detector.cc#L66)? The current CI is strict enough not to allow adding exception in the code.

lalitb:
@msiddhu - As you are more familiar with this,  checking your opinion on  how easy it would to modify the code to use the safe, non-throwing alternatives. I believe most of the error flagged by clang-tidy for missing exception handling is coming from C++ Standard Library ? Not asking you to do any change here, just trying to understand the complexity. :)

msiddhu:
Hello @lalitb , 

I feel that would be complex avoiding exceptions and using non-throwing alternatives.

For example, one of the functions got flagged:
```
/home/msiddhu/opentelemetry-cpp/exporters/elasticsearch/src/es_log_recordable.cc:238:31: warning: an exception may be thrown in function 'SetTraceId' which should not throw exceptions [bugprone-exception-escape]
  238 | void ElasticSearchRecordable::SetTraceId(const opentelemetry::trace::TraceId &trace_id) noexcept
```
[ElasticSearchRecordable::SetTraceId](https://github.com/open-telemetry/opentelemetry-cpp/blob/6dbfdb52ebca82d8705392fe88dadd9c133374e4/exporters/elasticsearch/src/es_log_recordable.cc#L238) has been flagged. 

In this particular function, the **string constructor** may produce a **bad allocation exception** due to no memory being available (which is very rare):
`std::string(trace_buf, sizeof(trace_buf)); `

In some cases, we could use other alternatives, but in others, we couldn't, in my opinion.

The attached file contains the **clang-tidy flagging related to exceptions**, which I have not yet addressed as the size of the PR is already large.

[function_exceptions.txt](https://github.com/user-attachments/files/16133946/function_exceptions.txt.txt)

Just check out what functions are flagged by clang-tidy to get an idea.

marcalff:
>     3. Noted. I had a doubt about the wanted to ask about CI, The CI will only run on the files which are touched in the particular PR right? Because running clang-tidy on every file(full project except some components) in every commit will take very long time I feel. At Least for me its taking long time to run clang-tidy.
> 

CI will run on every file, in every PR, on every commit.

How long is a full clang-tidy run taking currently, and how does it compare to a regular build ?

Runners can (and should) run clang-tidy in parallel (`make -j $(nproc)`) when possible, to reduce the total latency.

>     4. Noted. For this PR I removed the changes in API. Will create a new PR with those changes soon.
> 
>     5. Noted.
> 
> 
> #### Question: Should the CI for clang-tidy be provided in this PR or new PR for that?
> 

Please do this in a different PR. See notes in #2051

> Still one check is failing (Bazel noexcept). I'll resolve that soon and make the PR ready to review.
> 
> Please let me know if anything is needed. Thanks.

Starting the review now.

examples/etw_threads/main.cc
```diff
@@ -113,7 +113,7 @@ void beep()
 // Max number of threads to spawn
 constexpr int kMaxThreads = 10;
 
-int main(int arc, char **argv)
+int main()
```
marcalff, line 116: Why ?

If the issue is that argc and argv are not used, just comment the parameter names, but do not change the prototype of main.

exporters/ostream/test/ostream_log_test.cc
```diff
@@ -97,24 +97,28 @@ TEST(OstreamLogExporter, DefaultLogRecordToCout)
...
-      "  severity_num       : 0\n"
-      "  severity_text      : INVALID\n"
+      "  severity_num       : 0\n",
+      "  severity_text      : INVALID\n",
       "  body               : \n",
       "  resource           : \n",
-      "    telemetry.sdk.version: " OPENTELEMETRY_VERSION "\n",
+      "    telemetry.sdk.version: ",
+      OPENTELEMETRY_VERSION,
+      "\n",
```
marcalff, line 108: Works too, but the original code was correct: string concatenation done in the parser.

exporters/otlp/test/otlp_http_exporter_test.cc
```diff
@@ -158,34 +157,35 @@ class OtlpHttpExporterTestPeer : public ::testing::Test
...
-          {
-            EXPECT_EQ(GetOtlpDefaultUserAgent(), user_agent_header->second);
-          }
-
-          // let the otlp_http_client to continue
-          http_client::nosend::Response response;
-          response.Finish(*callback.get());
-        });
+        .WillOnce(
+            [&mock_session, report_trace_id](
```
marcalff, line 161: Why is the code formatted differently ?

All the formatting is already checked in CI, with clang-format.

Not sure if/why clang-tidy changed it, or if this is the result of running a possible different version of clang-format, but it will be better if clang-tidy avoids touching the code format at all.

We can't have multiple tools compete on this, generating noisy diffs.

exporters/prometheus/test/exporter_utils_test.cc
```diff
@@ -103,15 +105,13 @@ void assert_basic(prometheus_client::MetricFamily &metric,
       // Summary and Info type not supported
       ASSERT_TRUE(false);
       break;
-    case prometheus::MetricType::Untyped:
-      break;
```
marcalff, line 107: The case conveys semantic, it shows the UnTyped case was not forgotten.

Ok to remove the break and fall thru the default, but not ok to remove the case entirely.

exporters/zipkin/src/zipkin_exporter.cc
```diff
@@ -74,38 +75,54 @@ sdk::common::ExportResult ZipkinExporter::Export(
   exporter::zipkin::ZipkinSpan json_spans = {};
   for (auto &recordable : spans)
   {
-    auto rec = std::unique_ptr<Recordable>(static_cast<Recordable *>(recordable.release()));
-    if (rec != nullptr)
+    try
```
marcalff, line 78: This is a try/catch in the production code, which changes the logic.

The try is on each span in the loop, it could have been on the entire export too.

Not saying this is correct or incorrect, but a massive cleanup that contains 1000 lines of changes can not contain logic changes like this buried in the middle of it.

How to deal with exceptions to comply with noexcept will need more analysis, and should be better done separately.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -267,7 +268,7 @@ class HttpOperation
     return SetCurlPtrOption(option, list);
   }
 
-  CURLcode SetCurlLongOption(CURLoption option, long value);
+  CURLcode SetCurlLongOption(CURLoption option, int64_t value);
```
marcalff, line 271: I think this will break curl on 32bits platforms.

Curl does use `long` in the API, not int64_t.

Please revert.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -300,7 +301,7 @@ class HttpOperation
   const opentelemetry::ext::http::client::Compression &compression_;
 
   // Processed response headers and body
-  long response_code_;
+  int64_t response_code_;
```
marcalff, line 304: This relates to CURLINFO_RESPONSE_CODE, which uses a long.

Revert.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -453,7 +453,7 @@ void HttpOperation::Cleanup()
 #  define HAVE_TLS_VERSION
 #endif
 
-static long parse_min_ssl_version(std::string version)
+static int64_t parse_min_ssl_version(const std::string &version)
```
marcalff, line 456: Revert the long -> int64_t.
Keep the const reference.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -470,7 +470,7 @@ static long parse_min_ssl_version(std::string version)
   return 0;
 }
 
-static long parse_max_ssl_version(std::string version)
+static int64_t parse_max_ssl_version(const std::string &version)
```
marcalff, line 473: Revert the long -> int64_t.
Keep the const reference.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -529,7 +529,7 @@ CURLcode HttpOperation::SetCurlPtrOption(CURLoption option, void *value)
   return rc;
 }
 
-CURLcode HttpOperation::SetCurlLongOption(CURLoption option, long value)
+CURLcode HttpOperation::SetCurlLongOption(CURLoption option, int64_t value)
```
marcalff, line 532: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -729,9 +729,9 @@ CURLcode HttpOperation::Setup()
 
 #ifdef HAVE_TLS_VERSION
     /* By default, TLSv1.2 or better is required (if we have TLS). */
-    long min_ssl_version = CURL_SSLVERSION_TLSv1_2;
+    int64_t min_ssl_version = CURL_SSLVERSION_TLSv1_2;
```
marcalff, line 732: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -729,9 +729,9 @@ CURLcode HttpOperation::Setup()
...
 #else
-    long min_ssl_version = 0;
+    int64_t min_ssl_version = 0;
```
marcalff, line 734: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -755,7 +755,7 @@ CURLcode HttpOperation::Setup()
      * The CURL + openssl library may be more recent than this code,
      * and support a version we do not know about.
      */
-    long max_ssl_version = 0;
+    int64_t max_ssl_version = 0;
```
marcalff, line 758: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -773,7 +773,7 @@ CURLcode HttpOperation::Setup()
 #endif
     }
 
-    long version_range = min_ssl_version | max_ssl_version;
+    int64_t version_range = min_ssl_version | max_ssl_version;
```
marcalff, line 776: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -819,7 +819,7 @@ CURLcode HttpOperation::Setup()
     if (ssl_options_.ssl_insecure_skip_verify)
     {
       /* 6 - DO NOT ENFORCE VERIFICATION, This is not secure. */
-      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<long>(CURLUSESSL_NONE));
+      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<int64_t>(CURLUSESSL_NONE));
```
marcalff, line 822: Revert

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -840,7 +840,7 @@ CURLcode HttpOperation::Setup()
     else
     {
       /* 6 - ENFORCE VERIFICATION */
-      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<long>(CURLUSESSL_ALL));
+      rc = SetCurlLongOption(CURLOPT_USE_SSL, static_cast<int64_t>(CURLUSESSL_ALL));
```
marcalff, line 843: Revert

sdk/src/logs/logger.cc
```diff
@@ -48,44 +49,64 @@ const opentelemetry::nostd::string_view Logger::GetName() noexcept
...
+  try
+  {
+    recordable = context_->GetProcessor().MakeRecordable();
+  }
+  catch (const std::exception &e)
+  {
+    OTEL_INTERNAL_LOG_ERROR("[Logger::CreateLogRecord]"
+                            << "Error while creating recordable: " << e.what()
+                            << " Unable to create new Log Record");
+    return nullptr;
```
marcalff, line 62: Returning a nullptr is likely to break things.

Please revert all changes done to this method, this needs to be addressed separately.

sdk/src/metrics/data/circular_buffer.cc
```diff
@@ -83,13 +83,15 @@ struct AdaptingIntegerArrayCopy
 
 void AdaptingIntegerArray::Increment(size_t index, uint64_t count)
 {
-  const uint64_t result = nostd::visit(AdaptingIntegerArrayIncrement{index, count}, backing_);
-  if OPENTELEMETRY_LIKELY_CONDITION (result == 0)
+  while (true)
```
marcalff, line 86: Please explain in more details the change.

Some comments in the code are needed, very hard to understand what it actually does.

Why a while(true) loop, and is there a risk it never ends ?

lalitb:
> Are all the compilers wrong here, is the CI not strict enough, or is clang-tidy wrong ?

I think both are correct, the static analysis just goes a step further:

```bash
/tmp$ echo "#include <stdexcept>

void bar() {
    throw std::runtime_error(\"An error occurred\");
}

void foo() noexcept {
    bar(); 
}

int main() {
    foo();
    return 0;
}" > ./test_exception.cpp
$
$ g++ -Wall -Wextra -Werror -std=c++17 test_exception.cpp -o example
$ clang-tidy test_exception.cpp -checks='-*,bugprone-exception-escape' -- -std=c++17
2 warnings generated.
/tmp/test_exception.cpp:7:6: warning: an exception may be thrown in function 'foo' which should not throw exceptions [bugprone-exception-escape]
void foo() noexcept {
     ^
/tmp/test_exception.cpp:11:5: warning: an exception may be thrown in function 'main' which should not throw exceptions [bugprone-exception-escape]
int main() {
    ^
```

marcalff (changes requested):
Thanks for the cleanup.

This patch contains a lot of very good fixes, and also will improve performances with a lot of object copy that are avoided.

See some changes around CURL, it really needs to use the long type.

About try/catch blocks with exceptions, a bigger discussion is needed with maintainers to see how best to proceed, but given the topic, all the exception related work should probably be done separately.

msiddhu:
> If method foo() is flagged noexcept, and calls method bar() that can raise an exception, I would expect the compiler to raise plenty of errors about that.

`noexcept`  is for compiler optimizations, it's just tell compiler that this method won't throw any exception. But when there is an other function call it may which throws exceptions the function with `noexcept`  will call program to terminate. 

Yes, I feel about exceptions we should discuss more as they would reduce performance (Stack unwinding, Control flow). 

I'll remove the changes related to `try/catch` blocks and create a new PR after the discussion. 

Thanks @marcalff for taking time to review the long PR. Now I understood what I have done wrong. I'll update PR and have comments some comments soon.

sdk/src/metrics/data/circular_buffer.cc
msiddhu: It's avoid the recursive call. Instead makes an Iterative while loop. 

Anyway reverting this back. Will add in next small PR.

examples/etw_threads/main.cc
msiddhu: Noted. Added the **commented parameter names**.

exporters/ostream/test/ostream_log_test.cc
msiddhu: The original code is correct and this aswell. `clang-tidy `warns that there is no comma and the formatting moved them to next line. 

Reverting this.

exporters/zipkin/src/zipkin_exporter.cc
msiddhu: Understood. Reverting this.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
msiddhu: Noted and reverted.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
msiddhu: Noted and reverted.

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Noted.

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Done.

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Done

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Done

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Done

ext/src/http/client/curl/http_operation_curl.cc
msiddhu: Done

sdk/src/logs/logger.cc
msiddhu: Reverting. We'll take this in next PR

exporters/zipkin/src/zipkin_exporter.cc
```diff
@@ -105,7 +106,8 @@ sdk::common::ExportResult ZipkinExporter::Export(
     }
     return sdk::common::ExportResult::kFailure;
   }
-  return sdk::common::ExportResult::kSuccess;
+  // Commenting this return statement as the code is unreachable.
+  // return sdk::common::ExportResult::kSuccess;
```
marcalff, line 110: Good find, as both branches in the previous if/else have a return.

Please remove the line then instead of commenting it.

sdk/src/metrics/data/circular_buffer.cc
marcalff: Thanks. Fixing all the trivial issues first will reduce the overall noise, so it will be easier to address remaining (and less trivial) fixes later.

marcalff (approved):
LGTM, thanks for the cleanup.

marcalff:
Waiting for:

* tiny fix in zipskin_exporter.cc
* possible comments from other maintainers

and will merge after that.

exporters/zipkin/src/zipkin_exporter.cc
msiddhu: Removed the comment and return statement.

PR #2991, opened 2024-07-04
[BUILD] Upgrade to opentelemetry-proto 1.3.2

marcalff:
Fixes # (issue)

## Changes


docs/maintaining-dependencies.md
```diff
@@ -0,0 +1,258 @@
...
+#### file MODULE.bazel
+
+Make sure the new tag is available in bazel central:
+
+* [bazel-central](https://registry.bazel.build/modules/opentelemetry-proto)
+
+If missing, file a PR to add it, or contact the maintainer:
+
+* [repository](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/opentelemetry-proto)
+
```
lalitb, line 234: Can the upgrade proceed if the new tag is not available in bazel central? Not related to this PR, but trying to understand the role of bazel central here - Are we actually fetching the otel-proto depedency through bazel registry, or through repositories.bzl. cc @keith

lalitb (approved):
Looks good in general. Thanks for adding the upgrade steps. Few nit comments, but nothing blocking :)

docs/maintaining-dependencies.md
```diff
@@ -0,0 +1,258 @@
...
+# Maintaining dependencies
```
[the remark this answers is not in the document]
marcalff: Thanks for the comments.

This is the first section in the doc, so usage will tell if it makes sense to have one document for all dependencies, or one document per dependency, no strong opinion either way.

I don't think the fact that opentelemetry-proto comes from opentelemetry as opposed to a different organization is relevant here, it is still "external" as far as opentelemetry-cpp is concerned.

docs/maintaining-dependencies.md
```diff
@@ -0,0 +1,258 @@
...
+* [repository](https://github.com/open-telemetry/opentelemetry-proto)
+
+Check release notes at:
+
+* [release-notes](https://github.com/open-telemetry/opentelemetry-proto/releases)
+
+### Upgrade
+
+When upgrading opentelemetry-proto to a newer release,
+a few places in the code need adjustment.
```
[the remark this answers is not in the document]
marcalff: Thanks, implemented.

Added a comments section with some background.

docs/maintaining-dependencies.md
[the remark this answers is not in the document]
marcalff: Good point.

In most cases, we can upgrade opentelemetry-proto first, and implement the C++ changes later.

Added more comments to cover this.

docs/maintaining-dependencies.md
keith: Oops, the intent is that bzlmod is running on CI atm, since users should be moving to that, we do want breakages of the bzlmod only build to break CI

Fixing here https://github.com/open-telemetry/opentelemetry-cpp/pull/2995

keith: If this was running correctly the answer is yes that this has to be added to the BCR first. if bzlmod wasn't blocking on CI we still wouldn't want to update the MODULE.bazel since any HEAD users would be broken by this

marcalff:
Do not merge, blocked: waiting for this to settle in main:

* https://github.com/open-telemetry/opentelemetry-cpp/pull/2995

and then will update the doc accordingly.

docs/maintaining-dependencies.md
marcalff: Fixed the doc to reflect recent changes: CI does now depend on modules, and on bazel central.

PR #2995, opened 2024-07-08
[CI] Enable bzlmod

keith:
Somehow I missed this in my PR adding support for bzlmod

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -4,9 +4,6 @@
 # bazel configurations for running tests under sanitizers.
 # Based on https://github.com/bazelment/trunk/blob/master/tools/bazel.rc
 
-# TODO: Remove once support is added, avoid MODULE.bazel creation for now
-common --enable_bzlmod=false
-
 # Enable automatic configs based on platform
 common --enable_platform_specific_config
 
--- a/ci/do_ci.ps1
+++ b/ci/do_ci.ps1
@@ -11,8 +11,8 @@ $nproc = (Get-ComputerInfo).CsNumberOfLogicalProcessors
 $SRC_DIR = (Get-Item -Path ".\").FullName
 
 # Workaround https://github.com/bazelbuild/bazel/issues/18683
-$BAZEL_STARTUP_OPTIONS = "--output_base=C:\Out"
-$BAZEL_OPTIONS = "--copt=-DENABLE_ASYNC_EXPORT"
+$BAZEL_STARTUP_OPTIONS = "--output_base=C:\O"
+$BAZEL_OPTIONS = "--copt=-DENABLE_ASYNC_EXPORT --compilation_mode=dbg"
 $BAZEL_TEST_OPTIONS = "$BAZEL_OPTIONS --test_output=errors"
 
 if (!(test-path build)) {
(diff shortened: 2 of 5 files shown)


examples/http/BUILD
```diff
@@ -15,6 +15,7 @@ cc_binary(
             "-DEFAULTLIB:advapi32.lib",
             "-DEFAULTLIB:crypt32.lib",
             "-DEFAULTLIB:Normaliz.lib",
+            "-DEFAULTLIB:wldap32.lib",
```
ThomsonTan, line 18: What is this new introduced dependency for?`

keith: seems like there's some reliance on ldap that must have been autolinked before?

```
curl.lib(ldap.obj) : error LNK2019: unresolved external symbol __imp_ldap_search_s referenced in function _ldap_free_urldesc
```

not entirely sure since i don't have a windows machine to debug on. must be coming from a dependency somewhere

keith: ah actually i see, it's from curl and that's a bug where we should probably disable it.

keith: waiting on https://github.com/bazelbuild/bazel-central-registry/pull/2381

ci/do_ci.ps1
```diff
@@ -11,8 +11,8 @@ $nproc = (Get-ComputerInfo).CsNumberOfLogicalProcessors
 $SRC_DIR = (Get-Item -Path ".\").FullName
 
 # Workaround https://github.com/bazelbuild/bazel/issues/18683
-$BAZEL_STARTUP_OPTIONS = "--output_base=C:\Out"
-$BAZEL_OPTIONS = "--copt=-DENABLE_ASYNC_EXPORT"
```
keith, line 15: this continues to be a wild hack. I linked the new path issue in the upstream issue, it should be improved once grpc moves its deps into bzlmod, but right now the paths are artificially longer. absolutely wild stuff.

.bazelrc
```diff
@@ -4,9 +4,6 @@
 # bazel configurations for running tests under sanitizers.
 # Based on https://github.com/bazelment/trunk/blob/master/tools/bazel.rc
 
-# TODO: Remove once support is added, avoid MODULE.bazel creation for now
-common --enable_bzlmod=false
-
```
marcalff, line 9: Is there a way to support bazel modules on all platforms except windows, due to the file length disaster ?

keith: I don't think we can restrict it per platform in the BCR, so probably better we find the issues here in case users on windows try?

marcalff: I meant here, in this file:

somehow have this only for windows:

```
common --enable_bzlmod=false
```

so we can move on, with the windows limitation to be lifted later.

keith: oh, yea we definitely could, but now that things are passing i think we're probably fine? i think it's only likely to re-break if we update deps or something, so i don't think it would block normal development

marcalff:
Also, "This branch is out-of-date with the base branch", please merge recent changes from main.

.bazelrc
marcalff: Works for me. Waiting on @ThomsonTan and/or @lalitb opinion.

marcalff (approved):
Not a bazel expert, but looks ok to me.

Thanks for the work on bazel, with supporting PRs in BCR.

keith:
green after rebase! 🎉

ThomsonTan (approved):

PR #2999, opened 2024-07-10
[Metrics SDK] Fix hash calculation for nostd::string 

### LINKED ISSUE #2997: Cardinality limit on metrics (otel_metrics_overflow)
### Discussed in https://github.com/open-telemetry/opentelemetry-cpp/discussions/2993

<div type='discussions-op-text'>

<sup>Originally posted by **xjanin** July  8, 2024</sup>
Hi,
I've instrumented my application with opentelemetry-cpp (metrics only), and I use tags in my metrics. When testing with a low rate of metrics recording, I see my metrics with the expected tags in the prometheus export on the opentelemetry collector. However, when testing the application with a high rate of requests, and so a high rate of metrics recording, I get metrics aggregated in time series with no tags except "otel_metrics_overflow".

My understanding is that this should happens if the sdk collector has to collect metrics with a tag cardinality superior to 2000 in a collection cycle (by default).
Edit: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#cardinality-limits

However, my tag cardinality doesn't increase with the number of request processed, and in the opentelemetry collector that my application uses, I don't see 2000 time series in the prometheus export.

So my questions are : 
- Do I have a wrong understanding of what "otel_metrics_overflow" means ?
- Is there a way of monitoring the cardinality of the metrics produced by my application ?

Thank you and best regards,
Xavier</div>

lalitb:
Fixes #2997 

## Changes

**Step 1**:

Added test to validate the hash creation issue, where the Hash is being calculated on the address of key , and not it's value as below:

`TEST(AttributesHashMap, HashWithKeyValueIterable)`
```cpp
 // Create mock KeyValueIterable instances with the same content but different variables
  MockKeyValueIterable attributes1({{make_unique_string("k1"), make_unique_string("v1")},
                                    {make_unique_string("k2"), make_unique_string("v2")}});
  MockKeyValueIterable attributes2({{make_unique_string("k1"), make_unique_string("v1")},
                                    {make_unique_string("k2"), make_unique_string("v2")}});
  MockKeyValueIterable attributes3({{make_unique_string("k1"), make_unique_string("v1")},
                                    {make_unique_string("k2"), make_unique_string("v2")},
                                    {make_unique_string("k3"), make_unique_string("v3")}});
```

Most existing tests directly use `key` and `value` as literals, and these literals share a common address from the read-only data segment. This masked the issue, and it is also the most common use case where metric attributes are statically created and used during measurement recording.

CI failure -  https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/9870060037/job/27255004557?pr=2999

```
[ RUN      ] AttributesHashMap.HashWithKeyValueIterable
/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/metrics/attributes_hashmap_test.cc:142: Failure
Expected equality of these values:
  hash1
    Which is: 7942177087367374725
  hash2
    Which is: 7942177087698536463

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/metrics/attributes_hashmap_test.cc:144: Failure
Expected equality of these values:
  hash1
    Which is: 7942177087367374725
  hash3
    Which is: 7942177087497598195
```

**Step 2**:
The fix is to replace the hash calculation:
```cpp
GetHash(seed, key.data());
```
with

```cpp
GetHash(seed, key);
```
and so directly use the custom hash implementation for [nostd::string_view](https://github.com/open-telemetry/opentelemetry-cpp/blob/6dbfdb52ebca82d8705392fe88dadd9c133374e4/api/include/opentelemetry/nostd/string_view.h#L214). Though it comes with a cost, as this custom hash-implementation creates an intermediate string. This needs to be revisited.


marcalff:
Is there a way to make sure that `GetHash(seed, key);` is the same, with:

* `const char *key="foo";`
* `std::string_view key{"foo"};`
* `nostd::string_view key{"foo"};`

Maybe instantiate the GenHash template explicitly for all string types that matter.

sdk/include/opentelemetry/sdk/common/attributemap_hash.h
```diff
@@ -71,7 +71,7 @@ inline size_t GetHashForAttributeMap(
         {
           return true;
         }
-        GetHash(seed, key.data());
```
ThomsonTan, line 74: In the `GetHash` function, could we assert that `arg` should not be a pointer?

lalitb: We can add it, but since this is an internal API, it might not be necessary as long as we ensure it's used correctly

lalitb: Thinking again, we do need to support the `GetHash` function with char pointer, as the [AttributeValue ](https://github.com/open-telemetry/opentelemetry-cpp/blob/6dbfdb52ebca82d8705392fe88dadd9c133374e4/api/include/opentelemetry/common/attribute_value.h#L37)enum contains it.

lalitb:
> Is there a way to make sure that `GetHash(seed, key);` is the same, with:
> 
> * `const char *key="foo";`
> * `std::string_view key{"foo"};`
> * `nostd::string_view key{"foo"};`
> * `std::string key{"foo"};`
> 
> Maybe instantiate the GenHash template explicitly for all string types that matter.

I believe we template specialization for `char *`, as it is the valid [AttributeValue ](https://github.com/open-telemetry/opentelemetry-cpp/blob/6dbfdb52ebca82d8705392fe88dadd9c133374e4/api/include/opentelemetry/common/attribute_value.h#L37) type, rest should work with default hashing implementation, I can update the tests to validate it.

sdk/include/opentelemetry/sdk/common/attributemap_hash.h
ThomsonTan: If we do need supporter pointer, perhaps create a dedicated specialization for it? Then a difference hash function could be applied to pointers.

lalitb: yes, it is done in the latest commit.

lalitb:
bazel-asan is failing. Will check that- 

//sdk/test/metrics:attributes_hashmap_test                               FAILED in 0.5s
  /home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/_main/bazel-out/k8-fastbuild-asan/testlogs/sdk/test/metrics/attributes_hashmap_test/test.log

marcalff:
LGTM, thanks for the fix.

PR #3001, opened 2024-07-11
[CI] Add a clang-tidy build

### LINKED ISSUE #2051: [CI] Add a clang-tidy build

msiddhu:
Fixes #2051 

## Changes

- Added `clang-tidy-18` checks to GitHub Workflow. 
- Using the existing `.clang-tidy` file for checks which is added in #2990
- The opentracing_shim (build `WITH_OPENTRACING_SHIM` fails as it's taking the` .clang-tidy` file and has the warningsAsErrors flag enabled) and the etw exporter [#1423] are currently excluded from the build.
- The workflow **does not auto-fix any code**. It simply writes the **warnings** to an output file (`clang-tidy-output.log`) and prints the number of warnings present.

### Problems
Currently installing **gRPC** is eating up a lot of time. Open for suggestions to optimize this. 

### Note
The build contains features which are in currently in preview. Should modify the workflow file when the feature is released as stable.

### Remaining Work:

- Add more checks and resolve the warnings via multiple cleanups. 
- Compare the warnings count of main branch and the current PR. 


.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  clang-tidy:
+    runs-on: ubuntu-latest
```
marcalff, line 11: Better use:

runs-on: ubuntu-24.04

clang18 is already there.

Note: ubuntu-latest is only 22.04, because 24.04 is not marked as stable yet in github.

.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,74 @@
...
+          -DWITH_METRICS_EXEMPLAR_PREVIEW=ON \
+          -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+          -DWITH_OTLP_GRPC=ON \
+          -DWITH_OTLP_GRPC_SSL_MTLS_PREVIEW=ON
+          cd ..
+
+      - name: Run Clang-Tidy
+        run: |
+          chmod +x tools/clang-tidy.sh
```
marcalff, line 62: Do not use chmod here.

Instead, do chmod +x on the file and git add, git commit.

.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          sudo apt update -y
+          sudo apt install -y --no-install-recommends --no-install-suggests \
+          build-essential
+          sudo -E ./ci/setup_cmake.sh
+          sudo -E ./ci/setup_ci_environment.sh
+          sudo -E ./ci/setup_googletest.sh
+          sudo -E ./ci/install_abseil.sh
+          sudo -E ./ci/install_protobuf.sh
+          sudo -E ./ci/setup_grpc.sh -m -p protobuf -p abseil-cpp
```
marcalff, line 31: This will install dependencies from source, rebuilding a lot of code.

For clang-tidy, using the most up to date versions of dependencies is not relevant.

Better install packages that come with the platform, see apt install from the iwyu build.

tools/clang-tidy.sh
```diff
@@ -0,0 +1,59 @@
...
+# - .git
+# - _deps
+# - build
+# - out
+# - .vs
+# - opentelemetry_logo.png
+# - TraceLoggingDynamic.h:
+# - opentelemetry-shim
+
+FIND="find . -name third_party -prune -o -name opentracing-shim -prune -o -name tools -prune -o -name .git -prune -o -name _deps -prune -o -name build -prune -o -name out -prune -o -name .vs -prune -o -name opentelemetry_logo.png -prune -o -name TraceLoggingDynamic.h -prune -o -name '*.h' -o -name '*.cc' -print"
```
marcalff, line 20: So, this script:

* finds all the code, whether included in the CMake build or not
* uses CMake to generate compile_commands.json
* Iterate on each file seen in the repository, to run clang-tidy on it.

This does not work well with generated code, code that is platform dependent, etc.

Please investigate this instead:

https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

The driver here should be CMake, which invokes clang-tidy on each file that are part of the build (WITH_OTLP_HTTP, etc)

marcalff (commented):
Thanks for adding clang-tidy to the CI.

See some comments.

.github/workflows/clang-tidy.yaml
msiddhu: Done.

.github/workflows/clang-tidy.yaml
msiddhu: @marcalff  Is there any way to install `grpc ` from package managers? I don't see any. Please let me know if there are any. Because building `gRPC` from source code takes lot of time.

tools/clang-tidy.sh
msiddhu: Thanks for the suggestion.  Added this flag:

 `-DCMAKE_CXX_CLANG_TIDY="clang-tidy-18"`
 
 This may be helpful if we wanted to do with each PR changes: [clang tidy diff](https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/clang-tidy-diff.py)

.github/workflows/clang-tidy.yaml
msiddhu: Removed the file as we are using flag in `cmake`.

.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,79 @@
...
+          -DWITH_ZIPKIN=ON \
+          -DWITH_ELASTICSEARCH=ON \
+          -DWITH_OTLP_HTTP_COMPRESSION=ON \
+          -DWITH_EXAMPLES=ON \
+          -DWITH_EXAMPLES_HTTP=ON \
+          -DBUILD_W3CTRACECONTEXT_TEST=ON \
+          -DWITH_METRICS_EXEMPLAR_PREVIEW=ON \
+          -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+          -DCMAKE_CXX_CLANG_TIDY="clang-tidy-18"
```
marcalff, line 59: Not sure this is working properly, I don't see clang-tidy warnings in the log. Are you sure this is the correct binary name ?

There are build warnings from clang, yes, but the tidy warnings detected earlier are not there.

I suggest to add something like:

```
which clang-tidy
clang-tidy --version
```
in the worklow, to make sure clang-tidy is effectively installed.

This will help to debug CI.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [CI] Added clang-tidy to the CI pipeline, which outputs
+the number of warnings and generates a log file containing the
+details of these warnings [#2051](https://github.com/open-telemetry/opentelemetry-cpp/issues/2051)
```
marcalff, line 20: For the CHANGELOG, only use the issue title, with a link.

marcalff (commented):
This is taking shape.

Somehow the CI log do not show clang-tidy warnings, this needs some troubleshooting.

Make sure it works properly locally first, to validate CMAKE_CXX_CLANG_TIDY works.

.github/workflows/clang-tidy.yaml
marcalff: Let's put this aside from now, the same issue exists with the maintainers build for warnings, and iwyu.

.github/workflows/clang-tidy.yaml
msiddhu: Yes, I'll add this. But, 

https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/9909180749/job/27376671743

I can see the clang-tidy warnings for the job. Not sure why you are not able to view them.

marcalff: With iwyu, cmake runs iwyu first, and then runs the compiler (clang), to build a binary.

I suspect this works the same for clang-tidy, these warnings are from the compiler (clang).

For example, we know there are warnings about long/int64_t in the curl code, or warnings about exception in noexecpt methods ... where are they ?

marcalff: Locally, I get:

```
[  4%] Building CXX object api/test/context/propagation/CMakeFiles/composite_propagator_test.dir/composite_propagator_test.cc.o
/data/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/test/context/propagation/composite_propagator_test.cc:36:29: warning: C-style casts are discouraged; use constructor call syntax [google-readability-casting]
   36 |     auto it = headers_.find(std::string(key));
      |                             ^~~~~~~~~~~
      |                             std::string
/data/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/test/context/propagation/composite_propagator_test.cc:45:14: warning: C-style casts are discouraged; use constructor call syntax [google-readability-casting]
   45 |     headers_[std::string(key)] = std::string(value);
      |              ^~~~~~~~~~~
      |              std::string
/data/malff/CODE/MARC_GITHUB/opentelemetry-cpp/api/test/context/propagation/composite_propagator_test.cc:45:34: warning: C-style casts are discouraged; use constructor call syntax [google-readability-casting]
   45 |     headers_[std::string(key)] = std::string(value);
      |                                  ^~~~~~~~~~~
      |                                  std::string
[  4%] Linking CXX executable composite_propagator_test
```

so `-DCMAKE_CXX_CLANG_TIDY="clang-tidy"` is working.

.github/workflows/clang-tidy.yaml
```diff
@@ -0,0 +1,79 @@
...
+#                         -DWITH_ABSEIL=ON \
+#                    -DWITH_OTLP_GRPC=ON \
+#                    -DWITH_OTLP_GRPC_SSL_MTLS_PREVIEW=ON \
+
+      - name: clang-tidy
+        run: |
+          cd build
+          make 2>&1 | tee -a clang-tidy.log
```
marcalff, line 67: This only captures stdout.

warnings are printed in stderr, and lost.

use `2>&1`

tools/clang-tidy.sh
marcalff: Checked locally that:

```
-DCMAKE_CXX_CLANG_TIDY="clang-tidy-unknown"
```

causes the build to fail when the binary is not found, so this validates that clang-tidy-18 effectively exists on the github runner, and is correct.

.github/workflows/clang-tidy.yaml
msiddhu: This is due to TOPDIR which I set which makefile detects. And removed it now. We can see the warnings.

msiddhu:
A still have a problem. I'm not able to suppress  the **third_party**. Please help me with this if you know anyway. Or could revert back to original approach without using the  `-DCMAKE_CXX_CLANG_TIDY` .

I have tried different combinations of regex but the clang-tidy is still not suppressing the warnings.

marcalff:
> A still have a problem. I'm not able to suppress the **third_party**. Please help me with this if you know anyway. Or could revert back to original approach without using the `-DCMAKE_CXX_CLANG_TIDY` .
> 
> I have tried different combinations of regex but the clang-tidy is still not suppressing the warnings.
> 
> Even, this is taking longer to run in CI

See this code in the top level `CMakeList.txt`:

```
if(WITH_PROMETHEUS)
  find_package(prometheus-cpp CONFIG QUIET)
  if(NOT prometheus-cpp_FOUND)
    message(STATUS "Trying to use local prometheus-cpp from submodule")
    if(EXISTS ${PROJECT_SOURCE_DIR}/third_party/prometheus-cpp/.git)
      set(SAVED_ENABLE_TESTING ${ENABLE_TESTING})
      set(ENABLE_TESTING OFF)
      add_subdirectory(third_party/prometheus-cpp)
      set(ENABLE_TESTING ${SAVED_ENABLE_TESTING})
    else()
```

The same way the prometheus-cpp sub directory is build with `ENABLE_TESTING` = `OFF`, there should be a way to build that sub directory without clang-tidy, as in:

```
      set(CMAKE_CXX_CLANG_TIDY "")
```

Save and restore the setting, same as SAVED_ENABLE_TESTING.

msiddhu:
Thanks alot for the help. @marcalff 

- Added the `set(CMAKE_CXX_CLANG_TIDY "") ` in required third_parties. 
- Enabled some more checks. Currently the warnings are 542. and takes 18 min to run the check.
- In the next cleanups we could adjust the configuration file to get our correct requirements. 

Thanks.

marcalff (approved):
LGTM, thanks for the CI

lalitb:
> Currently installing gRPC is eating up a lot of time. Open for suggestions to optimize this.

The [bazel-windows](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/9980864851/job/27583243807?pr=3001) builds, which take ~ 30mins, provide enough buffer for CI tasks that are slower but still relatively fast. It's also worth considering using Alpine Linux, as it includes the gRPC package, and the opentelemetry-demo is already [utilizing it](https://github.com/open-telemetry/opentelemetry-demo/blob/a76f3377b1a3c0e427fd69fbfbf6350dc3eabe4b/src/currencyservice/Dockerfile#L19).

msiddhu:
Thanks @lalitb. I'll check this and include the OTLP_GRPC in the next clean up.

PR #3003, opened 2024-07-12
Remove libcurl from Bazel

### LINKED ISSUE #2072: Inconsistency between Bazel & CMake
**Describe your environment** In OTLP project, there are two different build mechanisms which one of them could be used for compilation different targets. Also, it would be helpful to maintain different third party packages from their source codes.

**Steps to reproduce**
For CMake, there is a file(third_party_release) which contains which third party library is linked with which version. Same for Bazel, the file(repository.bzl) contains same information. However, both file contains different third-party versions as an information.

For example;

- benchmark=v1.5.3 was defined for CMake but for Bazel it was defined "benchmark-1.6.0".
- CURL library was not defined as a depedency in the CMake.

Same versions must be used for even different build mechanisms.

There are different versions for Bazel and CMake which could cause inconsistency.

**Additional context**

haidong:
Fixes #2072

## Changes

The benchmark version is consistent between CMake and bazel. This PR removes the libcurl in repository.bzl.

### DIFF
--- a/bazel/repository.bzl
+++ b/bazel/repository.bzl
@@ -128,19 +128,6 @@ def opentelemetry_cpp_deps():
         ],
     )
 
-    # libcurl (optional)
-    maybe(
-        http_archive,
-        name = "curl",
-        build_file = "@io_opentelemetry_cpp//bazel:curl.BUILD",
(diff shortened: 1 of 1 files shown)


esigo:
@haidong thanks for the PR. please note that if you remove curl from bazel, you'd need to install curl in the system.

esigo:
Closing as CI is not fixes.

PR #3012, opened 2024-07-25
[EXPORTER] Ignore exception when create thread in OTLP file exporter.

### LINKED ISSUE #3011: OTLP File exporter may crash if create thread failed
**Describe your environment**

OS: Linux

**Steps to reproduce**
We have a internal prometheus file exporter which just like OTLP file exporter, and it may lead to crash when creating thread failed.
I think there is simular problem in OTLP file exporters.

**Additional context**

```bash
(gdb) bt
#0  0x00007f2882895387 in raise () from /lib64/libc.so.6
#1  0x00007f2882896a78 in abort () from /lib64/libc.so.6
#2  0x00007f2882f90a95 in __gnu_cxx::__verbose_terminate_handler() () from /lib64/libstdc++.so.6
#3  0x00007f2882f8ea06 in ?? () from /lib64/libstdc++.so.6
#4  0x00007f2882f8d9b9 in ?? () from /lib64/libstdc++.so.6
#5  0x00007f2882f8e624 in __gxx_personality_v0 () from /lib64/libstdc++.so.6
#6  0x00007f288ad2b8e3 in ?? () from /lib64/libgcc_s.so.1
#7  0x00007f288ad2be17 in _Unwind_Resume () from /lib64/libgcc_s.so.1
#8  0x00007f288f69f226 in rpc::telemetry::exporter::metrics::PrometheusFileBackend::AddMetricData (this=<optimized out>, data=...)
    at /data/devops/workspace/p-4943d876cb734a609b55ad5c740c36ff/src/pool/0/server/main/src/server_frame/rpc/telemetry/exporter/prometheus_file_exporter.cpp:636
#9  0x00007f288f69bdbd in rpc::telemetry::exporter::metrics::PrometheusFileExporter::Export (this=<optimized out>, data=...)
    at /data/devops/workspace/p-4943d876cb734a609b55ad5c740c36ff/src/pool/0/server/main/src/server_frame/rpc/telemetry/exporter/prometheus_file_exporter.cpp:1028
```

`PrometheusFileBackend::AddMetricData` calls `SpawnBackgroundWorkThread()` and the implementation just like `OtlpFileSystemBackend::SpawnBackgroundWorkThread()`.

owent:
Fixes #3011 

## Changes

+ Ignore exception when create thread failed.

### DIFF
--- a/exporters/otlp/src/otlp_file_client.cc
+++ b/exporters/otlp/src/otlp_file_client.cc
@@ -20,7 +20,6 @@
 #include "opentelemetry/exporters/otlp/protobuf_include_suffix.h" // IWYU pragma: keep
 // clang-format on
 
-#include "opentelemetry/common/macros.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/nostd/variant.h"
 #include "opentelemetry/sdk/common/base64.h"
--- a/exporters/otlp/src/otlp_file_client.cc
+++ b/exporters/otlp/src/otlp_file_client.cc
@@ -20,6 +20,7 @@
 #include "opentelemetry/exporters/otlp/protobuf_include_suffix.h" // IWYU pragma: keep
 // clang-format on
 
+#include "opentelemetry/common/macros.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/nostd/variant.h"
 #include "opentelemetry/sdk/common/base64.h"
@@ -47,6 +48,9 @@
 #include <thread>
 #include <utility>
 #include <vector>
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
+#  include <exception>
+#endif
 
 #if !defined(__CYGWIN__) && defined(_WIN32)
 #  ifndef WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


exporters/otlp/src/otlp_file_client.cc
```diff
@@ -20,6 +20,7 @@
 #include "opentelemetry/exporters/otlp/protobuf_include_suffix.h" // IWYU pragma: keep
 // clang-format on
 
+#include "opentelemetry/common/macros.h"
```
marcalff, line 23: macros.h is not needed.

By convention, including version.h also provides macros.h, because it contains the magic line:

```
#include "opentelemetry/common/macros.h"  // IWYU pragma: export
```

This was done because every file depends on version.h anyway, so code can reliably make use of various defines like OPENTELEMETRY_HAVE_EXCEPTIONS without having to remember to include macros.h.

marcalff (approved):
LGTM, thanks for the fix.

This is mostly code for robustness: if a thread can not be started, we log an error instead of taking the application down, but opentelemetry will execute in degraded mode, possibly not recording data.

In the long term, we may need to pay closer attention to warnings reported by clang-tidy about exceptions raised in noexcept methods, to find and resolve similar issues that may exist elsewhere already.

cc @msiddhu

exporters/otlp/src/otlp_file_client.cc
marcalff: Reported by include-what-you-use as:

```
/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_file_client.cc should remove these lines:
- #include "opentelemetry/common/macros.h"  // lines 23-23
```

because of the export pragma in version.h.

owent: Thanks, it's removed now.

PR #3015, opened 2024-08-02
[BUILD] Update the version in MODULE.bazel

BYVoid:
## Changes

Keep the version in `MODULE.bazel` consistent with other places (for example https://github.com/open-telemetry/opentelemetry-cpp/commit/baecbb95bd63df53e0af16e87bc683967962c5f8).

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -3,7 +3,7 @@
 
 module(
     name = "opentelemetry-cpp",
-    version = "0",
+    version = "1.16.1",
     compatibility_level = 0,
     repo_name = "io_opentelemetry_cpp",
 )


MODULE.bazel
```diff
@@ -3,7 +3,7 @@
 
 module(
     name = "opentelemetry-cpp",
-    version = "0",
+    version = "1.16.1",
```
lalitb, line 6: Isn't `0` means to fetch the latest version from the bazel repository which at present is `1.16.0` -https://registry.bazel.build/modules/opentelemetry-cpp? Not a bazel expert, but wondering if we need to explicitly add a version here.

BYVoid: `0` has no meaning. It does not represent the latest version.

Because the upstream package does not set it correctly, Bazel registry has to patch it: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/opentelemetry-cpp/1.16.0/patches/0001-Fix-Version-in-MODULE.bazel.patch

Like all other places in the code base storing the version number, it would be better if this place has the same value in future.

lalitb (approved):

marcalff (approved):
LGTM, thanks for the fix.

Now, we will need to remember to update this when making new releases then.

PR #3020, opened 2024-08-06
[BUILD] Add abi_version_no bazel flag.

### LINKED ISSUE #3019: Specify OPENTELEMETRY_ABI_VERSION_NO with Bazel
`OPENTELEMETRY_ABI_VERSION_NO` is by default `1` in Bazel, and there is no where to configure it, unlike CMake (`WITH_ABI_VERSION_2`).

Consider adding a Bazel flag to allow changing `OPENTELEMETRY_ABI_VERSION_NO` as in https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/BUILD.

I tried to patch the Bazel build files.

BYVoid:
Fixes #3019

## Changes

This change allows Bazel users to change the `OPENTELEMETRY_ABI_VERSION_NO`.

To specify the `OPENTELEMETRY_ABI_VERSION_NO` in a client project, simply add `build --@opentelemetry-cpp//api:abi_version_no=2` in the `.bazelrc`.

### DIFF
--- a/api/BUILD
+++ b/api/BUILD
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "int_flag", "string_flag")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -35,6 +35,9 @@ cc_library(
         ":set_cxx_stdlib_2020": ["OPENTELEMETRY_STL_VERSION=2020"],
         ":set_cxx_stdlib_2023": ["OPENTELEMETRY_STL_VERSION=2023"],
         "//conditions:default": [],
+    }) + select({
+        ":abi_version_no_1": ["OPENTELEMETRY_ABI_VERSION_NO=1"],
+        ":abi_version_no_2": ["OPENTELEMETRY_ABI_VERSION_NO=2"],
     }),
     strip_include_prefix = "include",
     tags = ["api"],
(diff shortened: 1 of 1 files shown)


marcalff:
@keith Please take a look (Bazel)

api/BUILD
```diff
@@ -61,3 +64,18 @@ bool_flag(
     build_setting_default = False,
     deprecation = "The value of this flag is ignored. Bazel builds always depend on Abseil for its pre-adopted `std::` types. You should remove this flag from your build command.",
 )
+
+int_flag(
+    name = "abi_version_no",
+    build_setting_default = 1,
```
keith, line 70: this is fine as is IMO but you can also add `make_variable = "ABI_VERSION_NO`, then add `toolchains = [":abi_version_no"],` to the rule above then you can use it like `defines = ["OPENTELEMETRY_ABI_VERSION_NO=$(ABI_VERSION_NO)"],` which allows you to remove the select

marcalff: Thanks for looking at it.

Actually, I prefer the current flags then, because it matches what is documented for CMake:

```
#
# We do not want to have WITH_ABI_VERSION = "1" or "2", and instead prefer two
# distinct flags, WITH_ABI_VERSION_1 and WITH_ABI_VERSION_2.
#
# This allows:
#
# * to have a specific option description for each ABI
# * to mark experimental/stable/deprecated on flags, for clarity
# * to search for exact abi usage move easily, discouraging:
#
#   * cmake -DWITH_ABI_VERSION=${ARG}
#
# While not supported, having distinct WITH_ABI_VERSION_1 and WITH_ABI_VERSION_2
# flags also opens the possibility to support multiple ABI concurrently, should
# that become necessary.
#
```

marcalff (approved):
LGTM, thanks for the fix.

api/BUILD
keith: note that the UX wouldn't change, only how it was consumed in the rule above, but yea sounds fine to me

lalitb (approved):

PR #3027, opened 2024-08-14
[SDK] Support empty histogram buckets

lalitb:
## Changes

As per the specs, it should be possible to configure the histogram aggregation with empty buckets - 

https://github.com/open-telemetry/opentelemetry-specification/blob/3a60fdae9392a57803b9f40325257db8ccba44a1/specification/metrics/data-model.md#histogram

``
Bucket counts are optional. A Histogram without buckets conveys a population in terms of only the sum and count, and may be interpreted as a histogram with single bucket covering (-Inf, +Inf).
``

As of now, we default to the standard bucket boundaries. Fixing the code for it.


marcalff (approved):
LGTM, thanks for the fix.

sdk/test/metrics/histogram_test.cc
```diff
@@ -134,6 +134,66 @@ TEST(Histogram, DoubleCustomBuckets)
...
+  std::unique_ptr<View> view{
+      new View("view1", "view1_description", instrument_unit, AggregationType::kHistogram, config)};
+  std::unique_ptr<InstrumentSelector> instrument_selector{
+      new InstrumentSelector(InstrumentType::kHistogram, instrument_name, instrument_unit)};
+  std::unique_ptr<MeterSelector> meter_selector{new MeterSelector("meter1", "version1", "schema1")};
+  mp.AddView(std::move(instrument_selector), std::move(meter_selector), std::move(view));
+
+  auto h = m->CreateDoubleHistogram(instrument_name, instrument_desc, instrument_unit);
+
+  h->Record(5, {});
```
ThomsonTan, line 161: nit: add some negative double value to be recorded in the test?

lalitb: Histogram only records positive values. Have modified the test to validate that the negative double values get ignored.

ThomsonTan (approved):

PR #3029, opened 2024-08-14
[API] Add synchronous gauge

### LINKED ISSUE #2279: Add Synchronous Gauge instrument to the metrics API/SDK

There's plenty of use cases described in the [specification issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2318).

I expect to have two new functions in the class `Meter` called `CreateInt64Gauge()` and `CreateDoubleGauge()` along with the template class `Gauge`. They should behave like the ObservableGauge instrument, but being synchronous instead of asynchronous.

This is described as an experimental feature in the OpenTelemetry specification: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#gauge

**Additional context**

uuzay:
Fixes #2279 

## Changes

Added synchronous gauge implementation according to the specification.


api/include/opentelemetry/metrics/meter.h
```diff
@@ -91,6 +94,25 @@ class Meter
...
+
+  virtual nostd::unique_ptr<Gauge<int64_t>> CreateInt64Gauge(
+      nostd::string_view name,
+      nostd::string_view description = "",
+      nostd::string_view unit        = "") noexcept = 0;
+
+  virtual nostd::unique_ptr<Gauge<double>> CreateDoubleGauge(
+      nostd::string_view name,
+      nostd::string_view description = "",
+      nostd::string_view unit        = "") noexcept = 0;
```
marcalff, line 114: These are new virtual methods in the API, which breaks the ABI.

Please use:

```
#if OPENTELEMETRY_ABI_VERSION_NO >= 2
#endif
```

to define them in ABI version 2 only, as ABI version 1 can not change.

marcalff (commented):
Thanks for the feature.

This is not a full review yet,
only some preliminary comments.

Please fix:

* declarations of new virtual methods using ABI version 2
* build break issues found in CI

More extensive review to follow.

api/include/opentelemetry/metrics/meter.h
```diff
@@ -91,6 +94,27 @@ class Meter
       nostd::string_view description = "",
       nostd::string_view unit        = "") noexcept = 0;
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  /**
+   * Creates a Gauge with the passed characteristics and returns a unique_ptr to that Counter.
```
ThomsonTan, line 99: Return a unique_ptr to that Gauge?

api/include/opentelemetry/metrics/meter.h
uuzay: Thanks for the quick feedback. 

I updated the code accordingly.

BTW I believe it still won't pass the format test since I could not find clang-format-10 for the debian distribution I'm using.

api/include/opentelemetry/metrics/meter.h
uuzay: Fixed the comment, thanks.

api/include/opentelemetry/metrics/meter.h
marcalff: There is a helper docker image to format:

```
docker run --rm --privileged=true --volume ${PWD}:/otel otel/cpp_format_tools
```

Run this from the PR top level directory.

uuzay: Thanks. As the last issue, a test case in Bazel valgrind CI (otlp_file_exporter_test) is failing. I'm not sure how it is related to this PR since it seems to be trace related. Could someone check and give me some pointers on how to fix maybe?

marcalff: This test failure is a known existing issue, please ignore it.

uuzay: Formatting should be fixed now.

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -19,6 +19,9 @@ enum class InstrumentType
   kCounter,
   kHistogram,
   kUpDownCounter,
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  kGauge,
+#endif
```
marcalff, line 24: It will be simpler to just declare this enum value all the time, and not conditionally.

The sdk headers can change, and having kGauge declared even in ABI v1 is not disruptive.

It will just be dead code in ABI v1.

Please adjust the switch/cases accordingly, to avoid the ifdef

marcalff (commented):
Thanks for the feature.

The code structure looks good to me overall.

I defer to @lalitb to check the logic around aggregation temporality in particular,
as I am not too familiar with this area.

lalitb:
@uuzay Thanks for the PR. Had a quick glance to the code, this looks similar to #2341 (correct me if not). And I believe this won't work for delta temporality. To validate, can we update the PR to test for it?

sdk/test/metrics/sync_metric_storage_gauge_test.cc
```diff
@@ -0,0 +1,189 @@
...
+      });
+  EXPECT_EQ(count_attributes, 2);  // RackA and RackB
+#else
+  EXPECT_TRUE(true);
+#endif
+}
+
+INSTANTIATE_TEST_SUITE_P(WritableMetricStorageTestDouble,
+                         WritableMetricStorageTestFixture,
+                         ::testing::Values(AggregationTemporality::kCumulative));
```
lalitb, line 189: Update the test here to include the delta temporality.

punya:
> @uuzay Thanks for the PR. Had a quick glance to the code, this looks similar to #2341 (correct me if not). And I believe this won't work for delta temporality. To validate, can we update the PR to test for it?

If I understand [the relevant section of the spec](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#temporality) correctly, Gauges (unlike UpDownCounters and Histograms) don't support a choice of temporality.

uuzay:
I was thinking the same @punya. 

It also mentions:
> Gauges do not provide an aggregation semantic, instead “last sample value” is used when performing operations like temporal alignment or adjusting resolution. ([source](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#gauge))

if that's related. 

@lalitb can you please clarify if delta temporality is actually needed?

lalitb:
> @lalitb can you please clarify if delta temporality is actually needed?

Sorry i was not clear enough. I meant testing for temporality with last value aggregation. The current test is only doing for cumulative.

lalitb:
@uuzay - Do you think we can modify the PR to restrict configuring the delta temporality for Last Value as of now - Just thrown an error at the startup if this is enabled through View or otherwise. With that, we should be good to go through this PR.

uuzay:
Hi @lalitb,
Sorry for the delays, I was both on a holiday period and also had other important tasks to take care of. 

I'm planning to make the necessary changes regarding the ABI macros this week. 
For delta temporality, I think what you suggested makes sense for now.

uuzay:
@lalitb I'm trying to figure out where would be an appropriate place to throw the error and would appreciate some pointers.

I was thinking if SyncMetricStorage or TemporalMetricStorage constructor would be a good place. 
Another idea that comes to mind is modifying DeltaTemporalitySelector in otlp_metric_utils to send a warning log and select cumulative temporality instead but I'm not sure if this covers all of the cases.

Thanks in advance.

lalitb:
> Hi @lalitb, I'm trying to figure out where would be an appropriate place to throw the error and would appreciate some pointers.
> 
> I was thinking if SyncMetricStorage or TemporalMetricStorage constructor would be a good place. Another idea that comes to mind is modifying DeltaTemporalitySelector in otlp_metric_utils to send a warning log and select cumulative temporality instead but I'm not sure if this covers all of the cases.
> 
> Thanks in advance.

@uuzay Yes that should work. Another easy place could be [MetricCollector::GetAggregationTemporality](https://github.com/open-telemetry/opentelemetry-cpp/blob/994761988eed36a41014de6291d470c738190ae0/sdk/src/metrics/state/metric_collector.cc#L36). This part of code is called for both periodic reader and Prometheus, so should cover all (future) scenarios.

marcalff:
@uuzay Please merge a recent main into this PR, to get the merge conflict resolved.
This should force the CI to build also, somehow it got stuck.
Thanks.

lalitb (approved):
LGTM. Thanks for working on this.

marcalff (approved):
LGTM, thanks for the feature.

api/include/opentelemetry/metrics/sync_instruments.h
```diff
@@ -247,5 +247,82 @@ class UpDownCounter : public SynchronousInstrument
...
+/* A Gauge instrument that records values. */
+template <class T>
+class Gauge : public SynchronousInstrument
+{
+
+public:
+  /**
+   * Record a value
+   *
+   * @param value The measurement value. May be positive, negative or zero.
```
ThomsonTan, line 260: not blocking the merge. Seems `T` should be integer or float types? If so, could it be asserted explicitly like below?

```
static_assert(std::is_same_v<T, int64_t> || std::is_same_v<T, double>, "Gauge<T> can only be instantiated with int64_t or double");
```

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -19,6 +19,7 @@ enum class InstrumentType
   kCounter,
   kHistogram,
   kUpDownCounter,
+  kGauge,
```
ThomsonTan, line 22: Is inserting a new `InstrumentType` of `kGauge` a SDK breaking change and also break the ABI? Should it be included only for `OPENTELEMETRY_ABI_VERSION_NO >= 2`?

ThomsonTan (changes requested):
Want to make sure if the adding of `kGauge` is ABI breaking change.

sdk/include/opentelemetry/sdk/metrics/instruments.h
lalitb: It won't be ABI breaking, as this is not part of API surface. It could be breaking change for custom exporters which use the `switch` on this `enum` without handling `default` case, but this can be added in the changelog.

marcalff: This has been discussed before:

https://github.com/open-telemetry/opentelemetry-cpp/pull/3029#discussion_r1730141170

The code as written is fine:

* no change in the API, this kGauge is part of the SDK
* the SDK ABI is changed, and we do not guarantee SDK ABI compatibility

I don't think we should document that in the changelog for the SDK, by definition every SDK change can (and most of the time does) break the SDK ABI.

@ThomsonTan Please remove the change requested then, as this blocks the merge.

ThomsonTan:
Apologize for blocking the PR. I think this is the exactly the issue I raised before.

The user application builds fine with either new SDK headers/old SDK libs, or old SDK headers/new SDK libs, either one will trigger a runtime break which is very hard to investigate like memory corruption.

It is true that the SDK doesn't guarantee ABI compatibility, but we don't need break this if not necessary. In this specific case, either wrap the new `kGauge` in the ABI version 2 macro, or we can just append it to the end of the enum list, so we don't trigger an potential break.

Even it is documented that SDK doesn't support ABI combability, it may be ignored by common users as this actually works in most of the time. A better solution will be that we check the SDK version and SDK lib version, and raise error (like block linking) if both versions mismatch.

How to you think?

marcalff:
> The user application builds fine with either new SDK headers/old SDK libs, or old SDK headers/new SDK libs, either one will trigger a runtime break which is very hard to investigate like memory corruption.

The only way for this to happen is to link dynamically with a shared library/DLL for the SDK itself, using a library from a different version compared to header files used during the application build.

Any application compiling the with the SDK header files and linking statically with the matching SDK library will be just fine.

Wrapping kGauge with ifdef causes another list of issues as well, because user code may break as well if not using the same ifdef.

I think that we can move kGauge at the end of the enum, to minimize binary compatibility issues and in practice avoid breaks for this case alone.

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -19,6 +19,7 @@ enum class InstrumentType
...
   kObservableCounter,
   kObservableGauge,
   kObservableUpDownCounter
```
marcalff, line 25:
```suggestion
  kObservableUpDownCounter,
  kGauge
```

marcalff:
> I think that we can move kGauge at the end of the enum, to minimize binary compatibility issues and in practice avoid breaks for this case alone.

Implemented as code review comment.

@ThomsonTan PTAL.

ThomsonTan:
> > The user application builds fine with either new SDK headers/old SDK libs, or old SDK headers/new SDK libs, either one will trigger a runtime break which is very hard to investigate like memory corruption.
> 
> The only way for this to happen is to link dynamically with a shared library/DLL for the SDK itself, using a library from a different version compared to header files used during the application build.
> 
> Any application compiling the with the SDK header files and linking statically with the matching SDK library will be just fine.
> 
> Wrapping kGauge with ifdef causes another list of issues as well, because user code may break as well if not using the same ifdef.
> 
> I think that we can move kGauge at the end of the enum, to minimize binary compatibility issues and in practice avoid breaks for this case alone.

Looks good to me then, thanks.

PR #3030, opened 2024-08-15
[SDK] PeriodicExportingMetricReader: future is never set, blocks until timeout

### LINKED ISSUE #3026: [SDK] PeriodicExportingMetricReader: future is never set, blocks until timeout
`PeriodicExportingMetricReader::CollectAndExportOnce()` future is never set and `CollectAndExportOnce()` probably blocks until timeout.

https://github.com/open-telemetry/opentelemetry-cpp/blob/9e062b5bdd0920e1d6f5e7ffadc468e0b7ec945b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc#L106

```
    std::promise<void> sender;
    auto receiver = sender.get_future();
```
nothing ever sets receiver's value. suggestion is to pass sender into the thread and have the thread call `sender.set_value()` when finished, in order to unblock the below `receiver.wait_for()` as soon as possible.

I think the saving grace here is the `break` when the future eventually times out.

owent:
Fixes #3026

## Changes

### DIFF
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -120,7 +120,7 @@ bool PeriodicExportingMetricReader::CollectAndExportOnce()
             return true;
           });
 
-          sender.set_value();
+          const_cast<std::promise<void> &>(sender).set_value();
         }));
 
     std::future_status status;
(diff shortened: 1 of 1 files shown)


marcalff (approved):
LGTM, thanks for the fix.

esigo (approved):
LGTM
thanks for the PR :)

ThomsonTan (approved):

PR #3042, opened 2024-08-29
Change EmitLogRecord to return success/fail

xman1979:
Summary:
Minor change to the API to return a boolean (true if succeed) when sending a log. so caller can consume and do proper retry, otherwise, there is no way for caller to know if a log has been sent successfully or not

Test Plan:
Please suggest what tests I need to run, thanks!

Reviewers:

Subscribers:

Tasks:

Tags:

Fixes # (issue)

## Changes

### DIFF
--- a/api/include/opentelemetry/logs/logger.h
+++ b/api/include/opentelemetry/logs/logger.h
@@ -43,8 +43,9 @@ public:
    * Emit a Log Record object
    *
    * @param log_record
+   * @return true if the record was successfully exported, false otherwise
    */
-  virtual void EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record) noexcept = 0;
+  virtual bool EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record) noexcept = 0;
 
   /**
    * Emit a Log Record object with arguments
@@ -65,18 +66,18 @@ public:
    *  span<pair<string_view, AttributeValue>> -> attributes(return type of MakeAttributes)
    */
   template <class... ArgumentType>
-  void EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record, ArgumentType &&...args)
+  bool EmitLogRecord(nostd::unique_ptr<LogRecord> &&log_record, ArgumentType &&...args)
   {
     if (!log_record)
     {
-      return;
+      return false;
     }
 
     IgnoreTraitResult(
         detail::LogRecordSetterTrait<typename std::decay<ArgumentType>::type>::template Set(
             log_record.get(), std::forward<ArgumentType>(args))...);
 
-    EmitLogRecord(std::move(log_record));
+    return EmitLogRecord(std::move(log_record));
   }
 
   /**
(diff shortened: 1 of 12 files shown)


owent:
Thanks for the PR. According to the specification, <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#enabled> , the return value should also consider whether it's enabled for input arguments. And maybe we change the return type of add all `Trace`, `Debug` , `Info`, `Warn`, `Error`, `Fatal`.

lalitb:
I think as per the specs, [EmitLogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#emit-a-logrecord) shouldn't return anything. And in the case of batch processor, there is no way for the `EmitLogRecord` to know if the log is exported successfully or not. The [retry logic](https://github.com/open-telemetry/opentelemetry-specification/blob/6672dbc97ddeb34f36c020a0f0a30323c8bc4d95/specification/logs/sdk.md#export) (if any) should be the part of the exporter, and not the application.

xman1979:
> I think as per the specs, [EmitLogRecord](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/bridge-api.md#emit-a-logrecord) shouldn't return anything. Also, in case of batch processor, there is no way for the `EmitLogRecord` to know if the log is exported successfully or not. The [retry logic](https://github.com/open-telemetry/opentelemetry-specification/blob/6672dbc97ddeb34f36c020a0f0a30323c8bc4d95/specification/logs/sdk.md#export) (if any) should be the part of the exporter, and not the application.

Thanks for pointing out implementing the retry in exporter, do you think it is reasonable to add retry policy in  OtlpHttpLogRecordExporterOptions? 

But In general, what is the rational explantation for EmitLogRecord not to return anything?  I don't see that is mentioned in the spec? I would think it is beneficial for application to know if the log is successfully sent out or not? even if we implemented the retry in the exporter,  it can still fail if we exhaust all the retries. 

For the batch processor, if I know this can fail silently, I would not consider to use it when reliability is a top priority in my use case.

lalitb:
> I would think it is beneficial for application to know if the log is successfully sent out or not? even if we implemented the retry in the exporter, it can still fail if we exhaust all the retries.

Yes, and the application can attach custom log handler to get the status - https://github.com/open-telemetry/opentelemetry-cpp/blob/f02d70bdb82985bea713b283bdd3da9937cb2172/docs/public/sdk/GettingStarted.rst#logging-and-error-handling.

punya:
> But In general, what is the rational explantation for EmitLogRecord not to return anything?  I don't see that is mentioned in the spec?

EmitLogRecord places the log into the processing pipeline, and typically returns before the processing pipeline (including batching and long-running IO) flushes the log to its destination outside the process. Thus it is infeasible for the return value of EmitLogRecord to accurately convey whether the log was "sent successfully."

The spec says

> A returned value of true means the Logger is enabled for the provided arguments, and a returned value of false means the Logger is disabled for the provided arguments.

This is about how the logging system is configured (e.g. the log level), not about runtime errors (e.g. network timeouts).

marcalff:
Thanks @xman1979 for the proposed contribution.

As commented by many in this PR, the API for EmitLogRecord should _not_  return an error code.

The fundamental design is as follows:

* an application without OpenTelemetry instrumentation works a given way.
* the same application instrumented with OpenTelemetry should continue to work the same way.

In particular, the instrumentation can discard everything entirely because the configuration does not enable logs, or it can be configured to collect log events but fail at runtime.

The proper place to add code to handle runtime failures is not at the API level, but inside the SDK or exporter implementation: for example, with a retry logic.

Closing this PR, as what it intends to implement is not compatible with the opentelemetry specifications.

Issue #1831, opened 2022-12-01
MetricProducer support


dashpole:
### Problem Statement

https://github.com/open-telemetry/opentelemetry-specification/pull/2951 defined an experimental MetricProducer interface, which can be used to support metric bridges. Note that the specifications are experimental, so it isn't critical that it is implemented at this time.

Changes:

* [ ] Define a MetricProducer interface, following https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricproducer
* [ ] Define a RegisterProducer(metricProducer) function on MetricReader, following https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#registerproducermetricproducer

punya:
Note that `MetricProducer` has been stable for about a year now, so it might be worth prioritizing this work.

PR #3044, opened 2024-09-01
[SDK] Update MetricProducer interface to match spec

punya:
Following https://github.com/open-telemetry/opentelemetry-specification/blob/v1.36.0/specification/metrics/sdk.md#metricproducer.

Updates #1831

## Changes

Updates `MetricProducer` to return a batch of metric points instead of receiving a callback.

* According to the spec, `Produce` SHOULD also take a Resource parameter, but I have chosen not to do that (because existing implementations pull the resource from the metric context). I'm not familiar with this area so I'd appreciate any advice.

* `MetricFilter` support isn't stable yet so I've kept it out of this initial PR.

### DIFF
--- a/sdk/src/metrics/metric_reader.cc
+++ b/sdk/src/metrics/metric_reader.cc
@@ -26,7 +26,7 @@ bool MetricReader::Collect(
   if (!metric_producer_)
   {
     OTEL_INTERNAL_LOG_WARN(
-        "MetricReader::Collect Cannot invoke Collect(). No MetricProducer registered for "
+        "MetricReader::Collect Cannot invoke Produce(). No MetricProducer registered for "
         "collection!")
     return false;
   }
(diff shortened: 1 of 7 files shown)


sdk/src/metrics/metric_reader.cc
```diff
@@ -36,7 +36,9 @@ bool MetricReader::Collect(
     OTEL_INTERNAL_LOG_WARN("MetricReader::Collect invoked while Shutdown in progress!");
   }
 
-  return metric_producer_->Collect(callback);
+  auto result  = metric_producer_->Produce();
+  auto success = callback(result.points_);
```
marcalff, line 40: This can invoke callback on failures.

Per the spec:

> When the Produce operation fails, the MetricProducer MAY return successfully collected results and a failed reasons list to the caller.

If calling callback on the part that was successfully collected is intended, please add  a comment that explains this, otherwise the code just looks broken at first read.

marcalff (commented):
Thanks for the fix.

Not a full review, just some preliminary comments.

marcalff (approved):
LGTM, thanks for the fix.

PR #3049, opened 2024-09-04
[EXAMPLE] update collector example with debug exporter

codeboten:
The logging exporter was renamed debug in 2023, the logging exporter will be removed in the near future, updating the example accordingly.


marcalff (approved):
Thanks for the upgrade.

PR #3053, opened 2024-09-05
[TEST] update references to logging exporter

codeboten:
## Changes

Missed some references in yesterday's PR and updated the verbosity to `detailed` where it was missed. This exporter has been replaced by the debug exporter and will be removed soon

Related to https://github.com/open-telemetry/opentelemetry-collector/pull/11037


marcalff (approved):
Thanks for the fix.

PR #3095, opened 2024-10-12
[BUILD] Add a CMake option to disable shared libs

tobim:
This option is useful to support platforms that do not support dynamic libraries and `dlopen`.

Fixes # (issue)

## Changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,7 +287,7 @@ option(WITH_ASYNC_EXPORT_PREVIEW "Whether to enable async export" OFF)
 option(WITH_METRICS_EXEMPLAR_PREVIEW
        "Whether to enable exemplar within metrics" OFF)
 
-option(OPENTELEMETRY_STATIC_ONLY
+option(OPENTELEMETRY_SKIP_DYNAMIC_LOADING_TESTS
        "Whether to build test libraries that are always linked as shared libs"
        OFF)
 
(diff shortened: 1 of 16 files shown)


tobim:
It is finished. The option only makes sense to use without `-DBUILD_SHARED_LIBS=ON`. Sorry for not making that clear.

CMakeLists.txt
```diff
@@ -287,6 +287,10 @@ option(WITH_ASYNC_EXPORT_PREVIEW "Whether to enable async export" OFF)
 option(WITH_METRICS_EXEMPLAR_PREVIEW
        "Whether to enable exemplar within metrics" OFF)
 
+option(OPENTELEMETRY_STATIC_ONLY
```
lalitb, line 290: should we name it to (say) - `OPENTELEMETRY_SKIP_DYNAMIC_LOADING_TESTS` - if this is to skip generating libraries for specific tests. Or any other name, OPENTELEMETRY_STATIC_ONLY is too generic.

examples/plugin/plugin/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+if(OPENTELEMETRY_SKIP_DYNAMIC_LOADING_TESTS)
```
ThomsonTan, line 4: Probably better to check this in the parent CMakeList.txt, around `add_subdirectory(plugin)`?

tobim: I updated the PR to do it this way.

marcalff (approved):
LGTM, thanks for the fix.

ThomsonTan (approved):

PR #3108, opened 2024-10-22
[EXPORTER] Add logging for async gRPC errors

santigimeno:
## Changes

If there's a gRPC error when exporting telemetry data via `gRPC` with `ENABLE_ASYNC_EXPORT` set, this error is not printed via logs in the same way as it's done when exporting the data [synchronously](https://github.com/open-telemetry/opentelemetry-cpp/blob/7402ed9bac9265ea22ddb1dd4e61029f745f7436/exporters/otlp/src/otlp_grpc_exporter.cc#L114-L117). Print the error stored in `grpc::Status` in the same way.


exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -207,14 +207,19 @@ static sdk::common::ExportResult InternalDelegateAsyncExport(
...
-               [call_data, async_data](::grpc::Status grpc_status) {
+               [call_data, async_data, export_data_name](::grpc::Status grpc_status) {
                  --async_data->running_requests;
                  ++async_data->finished_request_counter;
 
                  call_data->grpc_status = grpc_status;
                  if (call_data->grpc_status.ok())
                  {
                    call_data->export_result = opentelemetry::sdk::common::ExportResult::kSuccess;
+                 } else {
```
marcalff, line 218:
```suggestion
                 }
                 else
                 {
```

This should fix the clang-format errors in CI.

marcalff (approved):
LGTM, thanks for the fix.

See proposed change to fix CI.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -207,14 +207,19 @@ static sdk::common::ExportResult InternalDelegateAsyncExport(
...
+                    OTEL_INTERNAL_LOG_ERROR("[OTLP GRPC Client] ERROR: Export "
+                                            << export_data_name << " failed with status_code: \""
+                                            << grpc_status.error_code() << "\" error_message: \""
+                                            << grpc_status.error_message() << "\"");
```
lalitb, line 222: Not related to this PR, but we need to figure out some way of throttling, as this can otherwise flood the logs for the non-recoverable issues.

lalitb (approved):

owent (approved):
LGTM and thanks

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -216,6 +216,13 @@ static sdk::common::ExportResult InternalDelegateAsyncExport(
                  {
                    call_data->export_result = opentelemetry::sdk::common::ExportResult::kSuccess;
                  }
+                 else
+                 {
+                    OTEL_INTERNAL_LOG_ERROR("[OTLP GRPC Client] ERROR: Export "
+                                            << export_data_name << " failed with status_code: \""
+                                            << grpc_status.error_code() << "\" error_message: \""
+                                            << grpc_status.error_message() << "\"");
```
marcalff, line 224:
```suggestion
                   OTEL_INTERNAL_LOG_ERROR("[OTLP GRPC Client] ERROR: Export "
                                           << export_data_name << " failed with status_code: \""
                                           << grpc_status.error_code() << "\" error_message: \""
                                           << grpc_status.error_message() << "\"");
```

PR #3113, opened 2024-10-28
[EXPORTER] Log SSL Connection Information

### LINKED ISSUE #2250: Log SSL Connection Information

**Request**
I have successfully set up the HTTP metric/trace exporter to use SSL/TLS, but I am not seeing any connection specific information logged anywhere. With debug log level, I see the `[OTLP HTTP Client] Session state: connected` message, but that does not contain sufficient information.

Upon establishing an HTTPS connection, the negotiated TLS version and the Cipher suite should be logged. Logging the entire handshake would also be nice to have for debugging purposes.

chusitoo:
Fixes #2250

## Changes

Enable curl verbose output and add debug function callback to intercept the "interesting" logs.

This functionality is controlled by the compile-time flag/option `WITH_CURL_LOGGING=ON` but can also be enabled at runtime by setting `OtlpHttpExporterOptions::console_debug = true`.

By default, all the headers and other information are not available and protected by a compile-time flag since they could contain sensitive information.

An example of info logging related to TLS handshake, available by default:

```
[Info] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:593 SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / X25519 / RSASSA-PSS
```

An example of debug logging, enabled via `-DCURL_DEBUG`:

```
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => POST /v1/traces HTTP/2
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => Host: localhost:4318
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => Accept: */*
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => Content-Type: application/x-protobuf
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => User-Agent: OTel-OTLP-Exporter-Cpp/1.17.0
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:613 Send header => Content-Length: 238
[Debug] File: /otel/exporters/otlp/src/otlp_http_client.cc:226 [OTLP HTTP Client] Session state: sending request
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:602 We are completely uploaded and fine
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:620 Recv header => HTTP/2 200
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:620 Recv header => content-type: application/x-protobuf
[Debug] File: /otel/ext/src/http/client/curl/http_operation_curl.cc:620 Recv header => content-length: 2
```

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -263,7 +263,7 @@ HttpOperation::HttpOperation(opentelemetry::ext::http::client::Method method,
                              bool is_raw_response,
                              std::chrono::milliseconds http_conn_timeout,
                              bool reuse_connection,
-                             bool needs_to_log)
+                             bool is_log_enabled)
     : is_aborted_(false),
       is_finished_(false),
       is_cleaned_(false),
@@ -283,7 +283,7 @@ HttpOperation::HttpOperation(opentelemetry::ext::http::client::Method method,
       request_nwrite_(0),
       session_state_(opentelemetry::ext::http::client::SessionState::Created),
       compression_(compression),
-      needs_to_log_(needs_to_log),
+      is_log_enabled_(is_log_enabled),
       response_code_(0)
 {
   /* get a curl handle */
@@ -658,7 +658,7 @@ CURLcode HttpOperation::Setup()
     return rc;
   }
 #else
-  rc = SetCurlLongOption(CURLOPT_VERBOSE, static_cast<long>(needs_to_log_ || kEnableCurlLogging));
+  rc = SetCurlLongOption(CURLOPT_VERBOSE, static_cast<long>(is_log_enabled_ || kEnableCurlLogging));
   if (rc != CURLE_OK)
   {
     return rc;
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -598,6 +598,8 @@ int HttpOperation::CurlLoggerCallback(const CURL * /* handle */,
     {
       OTEL_INTERNAL_LOG_ERROR(text_to_log);
     }
+// This guard serves as a catch-all block for all other less interesting output that should
+// remain available for maintainer internal use and for debugging purposes only.
 #ifdef CURL_DEBUG
     else
     {
@@ -605,6 +607,8 @@ int HttpOperation::CurlLoggerCallback(const CURL * /* handle */,
     }
 #endif  // CURL_DEBUG
   }
+// Same as above, this guard is meant only for internal use by maintainers, and should not be used
+// in production (information leak).
 #ifdef CURL_DEBUG
   else if (type == CURLINFO_HEADER_OUT)
   {
@@ -658,7 +662,7 @@ CURLcode HttpOperation::Setup()
     return rc;
   }
 #else
-  rc = SetCurlLongOption(CURLOPT_VERBOSE, static_cast<long>(is_log_enabled_ || kEnableCurlLogging));
+  rc = SetCurlLongOption(CURLOPT_VERBOSE, (is_log_enabled_ && kEnableCurlLogging) ? 1L : 0L);
   if (rc != CURLE_OK)
   {
     return rc;
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -600,16 +600,16 @@ int HttpOperation::CurlLoggerCallback(const CURL * /* handle */,
     }
 // This guard serves as a catch-all block for all other less interesting output that should
 // remain available for maintainer internal use and for debugging purposes only.
-#ifdef CURL_DEBUG
+#ifdef OTEL_CURL_DEBUG
     else
     {
       OTEL_INTERNAL_LOG_DEBUG(text_to_log);
     }
-#endif  // CURL_DEBUG
+#endif  // OTEL_CURL_DEBUG
   }
 // Same as above, this guard is meant only for internal use by maintainers, and should not be used
 // in production (information leak).
-#ifdef CURL_DEBUG
+#ifdef OTEL_CURL_DEBUG
   else if (type == CURLINFO_HEADER_OUT)
   {
     static const auto kHeaderSent = nostd::string_view("Send header => ");
@@ -630,7 +630,7 @@ int HttpOperation::CurlLoggerCallback(const CURL * /* handle */,
     static const auto kHeaderRecv = nostd::string_view("Recv header => ");
     OTEL_INTERNAL_LOG_DEBUG(kHeaderRecv << text_to_log);
   }
-#endif  // CURL_DEBUG
+#endif  // OTEL_CURL_DEBUG
 
   return 0;
 }
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -649,6 +649,15 @@ CURLcode HttpOperation::Setup()
   curl_error_message_[0] = '\0';
   curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_ERRORBUFFER, curl_error_message_);
 
+// Support for custom debug function callback was added in version 7.9.6 so we guard against
+// exposing the default CURL output by keeping verbosity always disabled in lower versions.
+#if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 9, 6)
+  rc = SetCurlLongOption(CURLOPT_VERBOSE, 0L);
+  if (rc != CURLE_OK)
+  {
+    return rc;
+  }
+#else
   rc = SetCurlLongOption(CURLOPT_VERBOSE, static_cast<long>(needs_to_log_ || kEnableCurlLogging));
   if (rc != CURLE_OK)
   {
@@ -661,6 +670,7 @@ CURLcode HttpOperation::Setup()
   {
     return rc;
   }
+#endif
 
   // Specify target URL
   rc = SetCurlStrOption(CURLOPT_URL, url_.c_str());
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -22,6 +22,7 @@
 #include "opentelemetry/ext/http/client/curl/http_client_curl.h"
 #include "opentelemetry/ext/http/client/curl/http_operation_curl.h"
 #include "opentelemetry/ext/http/client/http_client.h"
+#include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
 #include "opentelemetry/version.h"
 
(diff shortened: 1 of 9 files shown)


ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -581,7 +641,14 @@ CURLcode HttpOperation::Setup()
   curl_error_message_[0] = '\0';
   curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_ERRORBUFFER, curl_error_message_);
 
-  rc = SetCurlLongOption(CURLOPT_VERBOSE, 0L);
+  rc = SetCurlLongOption(CURLOPT_VERBOSE, 1L);
```
lalitb, line 644: Setting CURLOPT_VERBOSE to 1L is useful for debugging but not sure if it's best practice to keep enabled in production, as it can significantly impact performance in high-throughput scenarios. Would it make sense to make this optional, controlled by an environment variable, configuration option, or macro?

chusitoo: Agreed, I think that would make sense as the general rule. 

If the requirement is to expose the TLS version / cipher information in the logs, then `CURLOPT_VERBOSE` is absolutely required in order to use a custom debug function callback, as explained in [CURLOPT_DEBUGFUNCTION](https://curl.se/libcurl/c/CURLOPT_DEBUGFUNCTION.html)

I haven't found anything obvious in terms of option switches / macros that I could repurpose for this so maybe introducing something like a `WITH_CURL_LOGGING` flag would be an acceptable approach.

lalitb: yes `WITH_CURL_LOGGING ` looks good to me.

marcalff: I agree, we need a compile time flag like `WITH_CURL_LOGGING`.

A runtime flag as well, to turn the verbose logging on/off would be nice, but this can be a separate discussion and PR.

chusitoo: Apologies for the sudden silence, it's been a busy week so far.

I actually had this similarly implemented with a runtime flag for another project. I think I can have this done in the same PR, just need to run this through the full CI to make sure I didn't miss anything.

If both compile-time and runtime flags are to be supported, the way I see it is that runtime should take precedence as there could be certain cases where this logging info would be useful but, as a general use case, omitting it would fall back to the value set at compile time via `WITH_CURL_LOGGING`.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -581,11 +649,28 @@ CURLcode HttpOperation::Setup()
...
+// Support for custom debug function callback was added in version 7.9.6 so we guard against
+// exposing the default CURL output by keeping verbosity always disabled in lower versions.
+#if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 9, 6)
   rc = SetCurlLongOption(CURLOPT_VERBOSE, 0L);
   if (rc != CURLE_OK)
   {
     return rc;
   }
+#else
+  rc = SetCurlLongOption(CURLOPT_VERBOSE, static_cast<long>(is_log_enabled_ || kEnableCurlLogging));
```
marcalff, line 661: Two concerns here:

* not sure if casting a boolean to a long will always give 0L / 1L, so a plain if statement to set a verbose local variable will be safer
* should it be `is_log_enabled && kEnableCurlLogging` instead ?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -569,8 +572,73 @@ CURLcode HttpOperation::SetCurlOffOption(CURLoption option, curl_off_t value)
...
+      OTEL_INTERNAL_LOG_ERROR(text_to_log);
+    }
+#ifdef CURL_DEBUG
+    else
+    {
+      OTEL_INTERNAL_LOG_DEBUG(text_to_log);
+    }
+#endif  // CURL_DEBUG
+  }
+#ifdef CURL_DEBUG
```
marcalff, line 608: Please clarify (with a comment) the intent of CURL_DEBUG, which is separate from ENABLE_CURL_LOGGING.

As I understand it, ENABLE_CURL_LOGGING is exposed in CMake, to add the new logging feature.

On top of that, CURL_DEBUG seem to be a private developer flag (not exposed in cmake), for a developper to debug the new logging feature itself, if needed.

So, ENABLE_CURL_LOGGING may be used optionally in production, but CURL_DEBUG should not be used in production.

Is this accurate ?

marcalff (commented):
Great work, thanks for the feature.

Reusing console_debug for this is a good idea.

See some comments or questions, with approval to follow once resolved.

ext/src/http/client/curl/http_operation_curl.cc
chusitoo: Regarding the explicit cast, it looked fairly safe given that the CURL API seems to only care about the value being 0 (disabled) or otherwise. Will update to a good ol' if statement, as requested.

As for the logic, my impression was that we wanted to have the runtime drive this behavior regardless of what is set at compile-time, but if omitting `console_debug`, it would still output logs provided `WITH_CURL_LOGGING` option was set.

If, what is ultimately needed is that the compile-time flag dictate the runtime behavior, then I guess it would need to be AND'ed, instead.

Let me know what is the preferred approach here and I will push all the required changes at the same time.

ext/src/http/client/curl/http_operation_curl.cc
chusitoo: That is correct, `CURL_DEBUG` is a guard that comes in handy when debugging something, where the full verbosity would be required. Making it controllable by the user could lead to unintentionally logging too much / sensitive information, such as request body, auth tokens, other sensitive in/out headers, etc.

Nothing would prevent one from "punching" more holes in the callback function, if this is deemed necessary by future requirements, by simply adding a few more `else if` statements outside the guarded block.

marcalff: Thanks for the clarifications.

Do not change any code then, just add a comment that CURL_DEBUG is for maintainers, and should not be used in production (information leak).

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Regarding the OR (`||`) or AND (`&&`) logic for the test:

* using the patch as written (OR), one can set the default at compile time to log details, without having to change the `console_debug` attribute for each exporter, which requires an application change.
* this sounds "easier" to use, but I think there is a twist.

Assume an application has a trace exporter and a metric exporter, both using OTLP HTTP to different endpoints.

By setting `WITH_CURL_LOGGING=ON`, both exporter are now traced, always. There is no way to turn off any of this tracing, even when debugging is no longer needed.

For this reason, I think the AND logic is better.

For people who wants to trace everything, it forces the application to use something like:

```
OtlpHttpExporterOptions opt;
#ifdef MY_OWN_DEBUG_FLAG
opt.console_debug = true; // or read from an application specific configuration
#else
opt.console_debug = false;
#endif
```

so in effect, the logic that was present in http_operation_curl.cc is moved to the application space, where the application (not opentelemetry-cpp) decides if/how it wants tracing to be done.

Yes, it is a code change for the application, to benefit from the feature.

However, this also allows to trace a metric exporter independently of a trace exporter, which I think is much more convenient to debug things, compared to the OR logic, especially when exporters points to different endpoints.

And when debugging is no longer needed ("or read from an application specific configuration"), tracing can be turned off, without having to recompile.

Please change the logic to AND then.

ext/src/http/client/curl/http_operation_curl.cc
chusitoo: Sure, I've added this statement almost word by word.

ext/src/http/client/curl/http_operation_curl.cc
chusitoo: That is a convincing argument, thanks for explaining in detail. Changes made to AND the flags and get rid of the cast.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -569,8 +572,77 @@ CURLcode HttpOperation::SetCurlOffOption(CURLoption option, curl_off_t value)
...
+#endif  // CURL_DEBUG
+  }
+// Same as above, this guard is meant only for internal use by maintainers, and should not be used
+// in production (information leak).
+#ifdef CURL_DEBUG
+  else if (type == CURLINFO_HEADER_OUT)
+  {
+    static const auto kHeaderSent = nostd::string_view("Send header => ");
+
+    while (!text_to_log.empty() && !std::iscntrl(text_to_log[0]))
```
lalitb, line 617: Do we need check for control character here? As this is HTTP header, it would be already be properly formatted having only printable chars

lalitb (approved):
LGTM. Thanks for the contribution.

marcalff (approved):
LGTM, thanks for the feature.

ext/src/http/client/curl/http_operation_curl.cc
chusitoo: This was mostly meant for checking for CR or LF instead of using isspace which would also bail out of the while loop on other criteria. It is probably not required but harmless too. I will need to check if it could be dropped entirely.

lalitb: Ok thanks, don't see much concern here. We can leave it here.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -569,8 +572,77 @@ CURLcode HttpOperation::SetCurlOffOption(CURLoption option, curl_off_t value)
...
+#ifdef CURL_DEBUG
+    else
+    {
+      OTEL_INTERNAL_LOG_DEBUG(text_to_log);
+    }
+#endif  // CURL_DEBUG
+  }
+// Same as above, this guard is meant only for internal use by maintainers, and should not be used
+// in production (information leak).
+#ifdef CURL_DEBUG
```
ThomsonTan, line 612: Seems the `CURL_DEBUG` is not from curl, instead, it is for private usage here? If so, could it be named as more specific like `OTEL_CURL_DEBUG`?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -569,8 +572,77 @@ CURLcode HttpOperation::SetCurlOffOption(CURLoption option, curl_off_t value)
...
+      OTEL_INTERNAL_LOG_ERROR(text_to_log);
+    }
+// This guard serves as a catch-all block for all other less interesting output that should
+// remain available for maintainer internal use and for debugging purposes only.
+#ifdef CURL_DEBUG
+    else
+    {
+      OTEL_INTERNAL_LOG_DEBUG(text_to_log);
+    }
+#endif  // CURL_DEBUG
```
marcalff, line 608:
```suggestion
#endif  // OTEL_CURL_DEBUG
```

ext/src/http/client/curl/http_operation_curl.cc
marcalff, line 612:
```suggestion
#ifdef OTEL_CURL_DEBUG
```

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -569,8 +572,77 @@ CURLcode HttpOperation::SetCurlOffOption(CURLoption option, curl_off_t value)
...
+        text_to_log = text_to_log.substr(pos + 1);
+      }
+    }
+  }
+  else if (type == CURLINFO_HEADER_IN)
+  {
+    static const auto kHeaderRecv = nostd::string_view("Recv header => ");
+    OTEL_INTERNAL_LOG_DEBUG(kHeaderRecv << text_to_log);
+  }
+#endif  // CURL_DEBUG
```
marcalff, line 633:
```suggestion
#endif  // OTEL_CURL_DEBUG
```

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Good point, fixed on behalf of @chusitoo

PR #3117, opened 2024-10-31
Allow cross-compilation with protobuf

tobim:
The upstream config module from protobuf always declares targets for the protobuf library and protoc of the same platform. This means one of the two is always wrong, as protoc needs to be run on the build machine, but the library must be linked for the "host" platform.

A small change in the CMake scaffold allows the user to pass `-DPROTOBUF_PROTOC_EXECUTABLE=<path/to/protoc>` to cmake when creating the build tree.

Fixes # (issue)

## Changes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -419,20 +419,21 @@ if(WITH_OTLP_GRPC
       endif()
     endif()
   endif()
-  # Latest Protobuf imported targets and without legacy module support
-  if(TARGET protobuf::protoc)
-    project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
-                                              protobuf::protoc)
-    # If protobuf::protoc is not a imported target, then we use the target
-    # directly for fallback
-    if(NOT PROTOBUF_PROTOC_EXECUTABLE)
-      set(PROTOBUF_PROTOC_EXECUTABLE protobuf::protoc)
+  if(NOT PROTOBUF_PROTOC_EXECUTABLE)
+    # Latest Protobuf imported targets and without legacy module support
+    if(TARGET protobuf::protoc)
+      project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
+                                                protobuf::protoc)
+      # If protobuf::protoc is not a imported target, then we use the target
+      # directly for fallback
+      if(NOT PROTOBUF_PROTOC_EXECUTABLE)
+        set(PROTOBUF_PROTOC_EXECUTABLE protobuf::protoc)
+      endif()
+    elseif(Protobuf_PROTOC_EXECUTABLE)
+      # Some versions of FindProtobuf.cmake uses mixed case instead of uppercase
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -419,20 +419,21 @@ if(WITH_OTLP_GRPC
...
+      endif()
+    elseif(Protobuf_PROTOC_EXECUTABLE)
+      # Some versions of FindProtobuf.cmake uses mixed case instead of uppercase
+      set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
     endif()
-  elseif(Protobuf_PROTOC_EXECUTABLE)
-    # Some versions of FindProtobuf.cmake uses mixed case instead of uppercase
-    set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
   endif()
-  include(CMakeDependentOption)
```
tobim, line 435: This is unrelated to the rest of the change, but seeing that the module isn't used anywhere in the project I thought I might as well clean it up.

CMakeLists.txt
```diff
@@ -419,20 +419,21 @@ if(WITH_OTLP_GRPC
...
-  if(TARGET protobuf::protoc)
-    project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
-                                              protobuf::protoc)
-    # If protobuf::protoc is not a imported target, then we use the target
-    # directly for fallback
-    if(NOT PROTOBUF_PROTOC_EXECUTABLE)
-      set(PROTOBUF_PROTOC_EXECUTABLE protobuf::protoc)
+  if(NOT PROTOBUF_PROTOC_EXECUTABLE)
+    # Latest Protobuf imported targets and without legacy module support
+    if(TARGET protobuf::protoc)
```
owent, line 424: Could you please also check and set `PROTOBUF_PROTOC_EXECUTABLE`?
`PROTOBUF_PROTOC_EXECUTABLE` is for the old version of cmake, and now it's `PROTOBUF_PROTOC_EXECUTABLE`.

owent:
@tobim Could you please update this PR? Find scripts of some packages are already be added in other PRs.

CMakeLists.txt
tobim: The intent of this change is so that one cat supply `-DPROTOBUF_PROTOC_EXECUTABLE=<path/to/protoc>` when invoking `cmake` on the command line.
I assume you mean I should also make `-DProtobuf_PROTOC_EXECUTABLE` possible, but I don't see a good reason for it. Having to options that do the same thing just creates confusion.

malkia: For thus project, doesn't this also need precompiled grpc protoc plugin?

marcalff:
Given that #3186 is merged, I assume this PR is no longer needed.

Please confirm if the issue is resolved, and close this pr.

tobim:
Yes, it works now.

PR #3120, opened 2024-11-04
[BUILD] allow building with -DWITH_OTLP_HTTP_COMPRESSION=OFF without zlib

cfstras:
## Issue
Currently, when building `-DWITH_OTLP_HTTP_COMPRESSION=OFF`, and not providing the libz dependency, we get this error:

```
...\opentelemetry-cpp\ext\src\http\client\curl\http_
client_curl.cc(6,10): error C1083: Cannot open include file: 'zconf.h': No such file or directory [...opentelemetry-cpp\build\ext\src\http\client\curl\opentelemetry_http_cli
ent_curl.vcxproj]
```

## Changes

This fixes the issue by moving the `#include <zconf.h>` inside the corresponding #ifdef block, letting the build pass.

Do I need to do any of these? LMK, happy to :)

> * [ ] `CHANGELOG.md` updated for non-trivial changes
> * [ ] Unit tests have been added
> * [ ] Changes in public API reviewed

### DIFF
--- a/ext/src/http/client/curl/http_client_curl.cc
+++ b/ext/src/http/client/curl/http_client_curl.cc
@@ -3,7 +3,6 @@
 
 #include <curl/curl.h>
 #include <curl/curlver.h>
-#include <zconf.h>
 #include <atomic>
 #include <chrono>
 #include <cstddef>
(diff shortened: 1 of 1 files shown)


marcalff:
This makes sense.

Note however that `ENABLE_OTLP_COMPRESSION_PREVIEW` is what we call a "feature flag":

* the extra code path is protected with an ifdef, for now
* building without the feature is allowed, for now

By the time the compression code is deemed stable, the feature flag will be removed, so the dependency at build time on zlib will become mandatory.

marcalff (approved):
LGTM, thanks for the fix.

ThomsonTan (approved):

PR #3123, opened 2024-11-05
[MAINTAINER] Add devcontainer

### LINKED ISSUE #3118: Instructions for running CI tests locally via Docker do not work
Based on the [current instructions](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ci/README.md) for running the CI tests locally, I should be able to run
```sh
./ci/run_docker.sh 
```
from the root of the project to enter a docker shell where I can run tests manually.

However running this command results in an error 
```
unable to prepare context: unable to evaluate symlinks in Dockerfile path /Projects/opentelemetry-cpp/ci/Dockerfile: no such file or directory
```

This is probably coming from [this line](https://github.com/open-telemetry/opentelemetry-cpp/blob/f30ab925c82ca817461ec9b5a6a268be0cfc3960/ci/run_docker.sh#L10) in the shell script which seems like it is explicitly looking for a Dockerfile in the `ci` directory and currently there is none.

esigo:
Fixes #3118 (issue)

## Changes
`bazelisk-linux-amd64 test --copt=-DENABLE_LOGS_PREVIEW --test_output=errors --cache_test_results=no --copt=-DENABLE_TEST //exporters/otlp/...`

`bash ci/do_ci.sh cmake.exporter.otprotocol.test`


CONTRIBUTING.md
```diff
@@ -60,6 +60,91 @@ bazel build //examples/simple:example_simple
...
+container (`devcontainer`) environment to streamline testing and development
+for this project. With the DevContainer, you can work in a consistent environment
+configured with all the necessary dependencies and tools.
+
+#### Prerequisites
+
+Before getting started, ensure you have the following installed:
+
+* **Docker**: DevContainers require Docker for containerization.
+* **Visual Studio Code (VSCode)** with the **Remote - Containers** extension.
```
psx95, line 75: Haven't used DevContainers before, but I imagine it would be possible to do this without relying on VS Code ?

Please correct me if I'm wrong, but I imagine we could simply exec into a running docker container as well and run CI tests from there ?

I would prefer if we not put in instructions that are IDE specific.

ci/README.md
```diff
@@ -1,7 +1,9 @@
 # Building and running tests as a developer
 
-CI tests can be run on docker by invoking the script `./ci/run_docker.sh
-./ci/do_ci.sh {TARGET}` where the targets are:
+CI tests can be run inside
+[devcontainer](../CONTRIBUTING.md#devcontainer-setup-for-project)
+by invoking the script
```
psx95, line 5: Could we also add a Dockerfile that creates a container with the current codebase and required dependencies to run the CI tests within that container ?

.devcontainer/Dockerfile.dev
```diff
@@ -0,0 +1,28 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+FROM otel/cpp_format_tools
+
+ARG GRPC_VERSION=v1.55.0
```
owent, line 6: Should we use the same versions in `third_party_release`?

CONTRIBUTING.md
esigo: of course you could build the image manually yourself and run the tests.
I update `run_docker.sh` script too.

ci/README.md
esigo: there are too many combinations, it doesn't make sense to have several docker files for each.

.devcontainer/Dockerfile.dev
esigo: the versions could be changer when building the image or in https://github.com/open-telemetry/opentelemetry-cpp/blob/6512381df4f6986a59d4045f0f9ba598cac68232/.devcontainer/devcontainer.json#L10-L13
not sure if we need to force versions on build environment.

marcalff (approved):
LGTM, thanks for the tooling and doc.

PR #3131, opened 2024-11-06
[EXPORTER]: Elasticsearch exporter put log resource in root instead of under 'resources'

### LINKED ISSUE #3119: Elasticsearch Log exporter "resources" are under "resources" key while it should be under the "root"
Not exactly a bug but this is to continue the work I have done here https://github.com/open-telemetry/opentelemetry-cpp/issues/3091

It is open to discussion but I think, like for `attributes` that `resources` should also be by default in the root object and that it would be up to the user to add different levels if necessary.

This would make sense as for instance in the resources we would put `user.id` which is common to all logs and that would  not make sense to "manually" add it through the log `attributes` every time we call `EmitLogRecord`

ShadowMaxLeb:
In order to give the user full flexibility on what and where to put resources for logs, remove the fact that it was forced to always be under the `resource`.
This allows to have resources such as `client.id` that will be under the root and comliant with ECS format.

Closes #3119

## Changes

Log resources are set under the root object and not under `resources` anymore.


marcalff (approved):
LGTM, thanks for the fix.

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
```diff
@@ -31,13 +31,8 @@ class ElasticSearchRecordable final : public sdk::logs::Recordable
...
    * being "attributes" and "resources"
    */
-  void WriteKeyValue(nostd::string_view key,
-                     const opentelemetry::common::AttributeValue &value,
-                     const std::string &name);
-
-  void WriteKeyValue(nostd::string_view key,
-                     const opentelemetry::sdk::common::OwnedAttributeValue &value,
-                     const std::string &name);
+  void WriteValue(const opentelemetry::sdk::common::OwnedAttributeValue &value,
```
lalitb, line 34: nit - do we need to revisit the method comment above?

lalitb (approved):

exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_recordable.h
ShadowMaxLeb: Yes probably
This is not needing anymore so I guess we just either right a small comment or just no comment at all since the `WriteValue` methods are kind of self explanatory ?

ShadowMaxLeb: I did some changes in the comment. Let me now if this is good for you.

marcalff:
@ShadowMaxLeb

In the CHANGELOG.md, please add an entry for important changes, to document that resources are no longer added under "root/resource" but are added under root instead.

I will merge this PR after that.

Thanks for the fix.

ShadowMaxLeb:
@marcalff done

marcalff:
> @marcalff done

Thanks.

I meant a section similar to this:

```
Important changes:

* [API] Jaeger Propagator should not be deprecated
  [#3086](https://github.com/open-telemetry/opentelemetry-cpp/pull/3086)

  * Deprecation of the Jaeger propagator, as announced on 2023-01-31
    in version 1.8.2, is now reverted.
  * This deprecation turned out to be not justified,
    as the Jaeger propagator can be used without the (now removed)
    Jaeger exporter.
```

but since there is a note in the changelog, I will notice it and edit it when doing the next release, so this will do for now.

Merging.

PR #3136, opened 2024-11-11
Set up for opentelemetry_proto_grpc_lib_type being shared

### LINKED ISSUE #3126: Build on IBM AIX
Opentelemetry-cpp doesn't build on IBM AIX out-of-the-box

Opentelemetry-cpp should build on AIX without needing locally-applied patches

**Additional context**

tjcw:
Caution, this patch may cause one of the tests to go into a loop

Fixes #3126
Buildability for AIX
## Changes
cmake/opentelemetry-proto.cmake to make opentelemetry_proto_grpc library link with gRPC::grpc++ if building shared

This PR split off from #3127

### DIFF
--- a/cmake/opentelemetry-proto.cmake
+++ b/cmake/opentelemetry-proto.cmake
@@ -348,7 +348,8 @@ if(WITH_OTLP_GRPC)
   target_link_libraries(opentelemetry_proto_grpc PUBLIC opentelemetry_proto)
 
   get_target_property(grpc_lib_type gRPC::grpc++ TYPE)
-  if(grpc_lib_type STREQUAL "SHARED_LIBRARY")
+  get_target_property(opentelemetry_proto_grpc_lib_type opentelemetry_proto_grpc TYPE)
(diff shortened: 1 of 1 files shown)


marcalff (changes requested):
The following CI:

CMake test (build shared libraries with otlp-exporter and static gRPC)

systematically fails on this test:

        Start  44: exporter.otlp.OtlpGrpcExporterFactoryTest.BuildTest

which hangs for up to 6 hours, and is then killed by the github infrastructure.

No explanations for the root cause at this point, but evidence from the recent CI runs in the opentelemetry-cpp repository seem to point at this change, as other branches do not show this.

It could be that this change is correct and exposed an existing bug elsewhere, or that somehow this change is incorrect / incomplete.

In any case, we need this resolved before merging, to avoid breaking CI.

tjcw:
Asking a colleague, it seems we no longer use the function that is set up by this PR. So I can close it. Please reopen another matching PR if you want to investigate why it causes the CI test to go into a loop.

Issue #2641, opened 2024-04-18
[SDK] Add simple scope configuration to Tracer, Meter, Logger


marcalff:
Per spec change:

* https://github.com/open-telemetry/opentelemetry-specification/pull/3877

psx95:
I could take this up, if there are no takers. Feel free to assign this to me.

PR #3137, opened 2024-11-12
[SDK] Add tracer scope configurator

psx95:
Contributes to #2641 

Adds scope configurator for Tracers.

## Changes
 - Updates the Trace SDK to add support for [TracerConfigurator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracerconfigurator) as per the spec changes.
 - Adds a TracerConfig class that can be used to configure a tracer's behavior as per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracerconfig).

Some optional, good-to-have convenience functions were recommended by the spec to accommodate common use-cases - these **have not been added** in this PR.

*See the [TracerConfigurator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#tracerconfig) spec for these recommendations. They should be added in a future PR.*


sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -103,8 +105,10 @@ class Tracer final : public opentelemetry::trace::Tracer,
 private:
   // order of declaration is important here - instrumentation scope should destroy after
   // tracer-context.
+  std::unique_ptr<TracerConfig> tracer_config_;
```
marcalff, line 108: Replace with:
```
TracerConfig tracer_config_;
```

There is no need to allocate memory and use a unique pointer, the tracer can own it's own configuration directly.

sdk/src/trace/tracer.cc
```diff
@@ -36,11 +38,16 @@ namespace sdk
 {
 namespace trace
 {
+const std::shared_ptr<opentelemetry::trace::NoopTracer> Tracer::kNoopTracer =
+    std::make_shared<opentelemetry::trace::NoopTracer>();
```
marcalff, line 42: Why a shared_ptr ?

This memory will never be reclaimed.

This should be enough:

```
opentelemetry::trace::NoopTracer kNoopTracer();
```

sdk/include/opentelemetry/sdk/trace/tracer_config.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Returns a ScopeConfigurator that can be used to get the default tracer configurator.
+   * A tracer configurator computes TracerConfig based on the InstrumentationScope. The default
+   * tracer configurator unconditionally computes the default TracerConfig for all scopes.
+   *
+   * @return a static constant TracerConfig that represents a tracer configurator with default
+   * behavior.
+   */
+  static const instrumentationscope::ScopeConfigurator<TracerConfig> &DefaultConfigurator();
```
marcalff, line 60: To move in a separate class.

A TracerConfig represents a configuration.

This class should not know how / when / by whom a configuration is created, and should not even know of ScopeConfigurator.

sdk/include/opentelemetry/sdk/trace/tracer_config.h
```diff
@@ -0,0 +1,71 @@
...
+
+private:
+  explicit TracerConfig(const bool disabled = false) : disabled_(disabled) {}
+  bool disabled_;
+  static const TracerConfig kDefaultConfig;
+  static const TracerConfig kDisabledConfig;
+  static const instrumentationscope::ScopeConfigurator<TracerConfig> kDefaultTracerConfigurator;
```
marcalff, line 67: Remove kDefaultTracerConfigurator.

sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h
```diff
@@ -158,6 +159,8 @@ class InstrumentationScope
   InstrumentationScopeAttributes attributes_;
 };
 
+template <class T>
+using ScopeConfigurator = std::function<T(const InstrumentationScope &)>;
```
marcalff, line 163: A couple of comments.

(1) 

ScopeConfigurator<T> should be an abstract class, with a virtual configure() method.

This will help to define different kinds of configurators later, including the various builtins helpers.

(2)

One of the configurators is the always enabled one, so implement a AlwaysOnScopeConfigurator sub class.

Have a static AlwaysOnScopeConfigurator::Create() method, to return an instance.

(3)

Likewise, AlwaysOffScopeConfigurator::Create()

(4)

All this should be in dedicated headers, so instrumentation_scope.h should be unchanged.

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -43,7 +46,10 @@ class TracerContext
           opentelemetry::sdk::resource::Resource::Create({}),
       std::unique_ptr<Sampler> sampler = std::unique_ptr<AlwaysOnSampler>(new AlwaysOnSampler),
       std::unique_ptr<IdGenerator> id_generator =
-          std::unique_ptr<IdGenerator>(new RandomIdGenerator())) noexcept;
+          std::unique_ptr<IdGenerator>(new RandomIdGenerator()),
+      std::unique_ptr<instrumentationscope::ScopeConfigurator<TracerConfig>> tracer_configurator =
+          std::make_unique<instrumentationscope::ScopeConfigurator<TracerConfig>>(
+              TracerConfig::DefaultConfigurator())) noexcept;
```
marcalff, line 52: Replace with (assuming the default is enabled):

tracer_configurator = AlwaysOnScopeConfigurator::Create();

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -77,6 +83,8 @@ class TracerContext
    */
   const opentelemetry::sdk::resource::Resource &GetResource() const noexcept;
 
+  instrumentationscope::ScopeConfigurator<TracerConfig> &GetTracerConfigurator() const noexcept;
```
marcalff, line 86: The return type should probably be a const reference, not just a reference.

sdk/src/trace/tracer.cc
```diff
@@ -51,6 +58,10 @@ nostd::shared_ptr<opentelemetry::trace::Span> Tracer::StartSpan(
   opentelemetry::trace::SpanContext parent_context = GetCurrentSpan()->GetContext();
   if (nostd::holds_alternative<opentelemetry::trace::SpanContext>(options.parent))
   {
+    if (!tracer_config_->IsEnabled())
+    {
+      return kNoopTracer->StartSpan(name, attributes, links, options);
+    }
```
marcalff, line 64: In my understanding, this should be moved earlier, not done only for the SpanContext alternative.

If the tracer is disabled, it is disabled for all types of spans.

sdk/include/opentelemetry/sdk/trace/tracer_config.h
```diff
@@ -0,0 +1,71 @@
...
+   * A tracer configurator computes TracerConfig based on the InstrumentationScope. The default
+   * tracer configurator unconditionally computes the default TracerConfig for all scopes.
+   *
+   * @return a static constant TracerConfig that represents a tracer configurator with default
+   * behavior.
+   */
+  static const instrumentationscope::ScopeConfigurator<TracerConfig> &DefaultConfigurator();
+
+private:
+  explicit TracerConfig(const bool disabled = false) : disabled_(disabled) {}
```
marcalff, line 63: Do not use default values `= false`.

The choice of the default value should not be nested deep in the code, but bubble up in the call stack, up to where the decision is made to use a default.

This will be in the tracer provider factory, when no scope configurator is provided.

sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h
psx95: Sounds good, I had a class before, but removed since it was easily replaceable by a function. I was expecting to put the built-in helpers in a static utility class, but this approach might look better.

I will give this a try.

sdk/include/opentelemetry/sdk/trace/tracer_config.h
psx95: This is a private constructor, so it cannot be called externally and the value of the default is dictated by the spec.

sdk/src/trace/tracer.cc
psx95: I was trying to avoid making multiple copies of NoopTracer, but looking at it again, shared_ptr might actually have more overhead due to ref counts.

psx95:
Thanks for the detailed review, I will go over the suggestions and make necessary changes.

sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h
marcalff: In the mean time, the spec configuration repository is defining the yaml bits for this:

* https://github.com/open-telemetry/opentelemetry-configuration/pull/140

Instead of "AlwaysOn" / "AlwaysOff", the following is probably better:

class ScopeConfigurator;
class DefaultScopeConfigurator : public ScopeConfigurator;

and later

class NamedScopeConfigurator : public ScopeConfigurator; (for name matching)

sdk/src/trace/tracer.cc
psx95: Ah yes, it should be. Missed this.

sdk/src/trace/tracer.cc
psx95: I think I remember the main reason for why I put this as `shared_ptr`.

The [NoopTracer](https://github.com/open-telemetry/opentelemetry-cpp/blob/150256c1720b279a1dfc7c380d17e13e90cee15d/api/include/opentelemetry/trace/noop.h#L94) inherits from `std::enable_shared_from_this<NoopTracer>`. 

And IIUC, this means an exception would be thrown if no shared_ptr exists at the time of usage. I just removed the shared_ptr and the tests started failing with `std::bad_weak_ptr` exception.

I'm not super familiar with `enable_shared_from_this`, so please let me know if my understanding is incorrect.

sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h
psx95: My one concern with this approach is that it would require us to create and maintain multiple classes which wont necessary hold state and could therefore also be replaced by functions.

Based on these suggestions, I have created a Builder based approach that allows for helper functions to add functionality to a scope configurator. The existing tests show how this change would look like from the user's perspective.

Please take a look and let me know your thoughts on that.

sdk/test/trace/tracer_config_test.cc
```diff
@@ -0,0 +1,89 @@
...
+                  "https://opentelemetry.io/schemas/v1.18.0",
+                  {attr1}))
+        .get(),
+    std::move(instrumentation_scope::InstrumentationScope::Create(
+                  "test_scope_4",
+                  "0",
+                  "https://opentelemetry.io/schemas/v1.18.0",
+                  {attr1, attr2, attr3}))
+        .get(),
+};
```
marcalff, line 68: This construct is a problem.

Either this is static test data, where the `instrumentation_scopes` array is initialized with _**data**_, not code.

Ot it is constructed with code, with statements adding elements in the array.

As written, this will be very challenging to debug and interpret stack traces, should it fail at runtime.

sdk/test/trace/tracer_test.cc
```diff
@@ -484,6 +495,137 @@ TEST(Tracer, SpanSetEvents)
...
+  // Custom configurator 1 evaluates version first and enables the tracer
+  const auto span_foo_scope_with_version_1 = tracer_foo_scope_with_version_1->StartSpan("span 1");
+  auto &span_foo_scope_with_version_ref_1  = *span_foo_scope_with_version_1.get();
+  EXPECT_EQ(typeid(span_foo_scope_with_version_ref_1), typeid(opentelemetry::sdk::trace::Span));
+
+  // Custom configurator 2 evaluates the name first and therefore disables the tracer without
+  // evaluating other condition
+  const auto span_foo_scope_with_version_2 = tracer_foo_scope_with_version_2->StartSpan("span 1");
+  auto &span_foo_scope_with_version_ref_2  = *span_foo_scope_with_version_2.get();
+  EXPECT_EQ(typeid(span_foo_scope_with_version_ref_2), typeid(trace_api::NoopSpan));
```
marcalff, line 622: Possible alternative to using OPENTELEMETRY_RTTI_ENABLED, not tested, to investigate:

* Create a noop_span using NoopTracerProvider::GetTracer() and NoopTracer::StartSpan
* Compare the pointer to the span object with the noop span.

This relies on the fact that noop spans are always the same object, per:

```
    // Don't allocate a no-op span for every StartSpan call, but use a static
    // singleton for this case.
    static nostd::shared_ptr<trace::Span> noop_span(new trace::NoopSpan{this->shared_from_this()});

    return noop_span;
```

Alternate possibility:

* check for IsRecordable(), should be false for noop span and true for real spans.

marcalff (commented):
Thanks for the patch, this is making good progress.

See another round comments on the code and tests.

sdk/test/trace/tracer_config_test.cc
psx95: This is static test data. 

The only way to construct InstrumentationScope objects is to use the static `Create` method which returns a `nostd::unique_ptr`, which is why the std::array accepts a pointer type and why a call to the `std::move` is required.

I can extract the creation of objects to separate lines, just like I have done for attributes. 

That might make it easier to debug stacktraces but the construction of `instrumentation_scope` objects will remain the same.

sdk/test/trace/tracer_test.cc
psx95: I like the idea of comparing to the noop_span directly. 

According to the spec, `isRecording()` might return true for a real span too - for instance if the span has ended. So I am preferring to use direct comparisons with the noop_span. 

One small caveat with this approach is that the test is now dependent on the implementation of the `NoopTracer`, but I feel it is OK for now.

psx95:
Hi @marcalff, 
I have addressed all your comments, please take a look.

The CI check for `CI / CMake -> exporter proto` first failed in this [recently merged PR](https://github.com/open-telemetry/opentelemetry-cpp/pull/3254). 

The check was passing before the changes from main were merged into this.

sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h
```diff
@@ -0,0 +1,141 @@
...
+     * in order.
+     * @param scope_matcher a function that returns true if the scope being evaluated matches the
+     * criteria defined by the function.
+     * @param scope_config the scope configuration to return for the matched scope.
+     * @return this
+     */
+    Builder AddCondition(std::function<bool(const InstrumentationScope &)> scope_matcher,
+                         T scope_config)
+    {
+      conditions_.push_back(Condition{scope_matcher, scope_config});
```
chusitoo, line 51: Minor nit (feel free to ignore); consider emplace_back to construct the Condition directly inside the vector

sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h
```diff
@@ -0,0 +1,141 @@
...
+     * evaluated. If the scope name matches to the provided string, then the provided scope
+     * configuration is applied to the scope.
+     * @param scope_name The scope name to which the config needs to be applied.
+     * @param scope_config The scope config for the matching scopes.
+     * @return this
+     */
+    Builder AddConditionNameEquals(nostd::string_view scope_name, T scope_config)
+    {
+      std::function<bool(const InstrumentationScope &)> name_equals_matcher =
+          [scope_name](const InstrumentationScope &scope_info) {
```
chusitoo, line 66: Passing `string_view` in the context looks dangerous, because one is not guaranteed that the contents of scope_name will still be around when the matcher is invoked. One possibility would be to own a copy

```suggestion
          [scope_name=std::string(scope_name)](const InstrumentationScope &scope_info) {
```

sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h
```diff
@@ -0,0 +1,141 @@
...
+            return scope_info.GetName() == scope_name;
+          };
+      conditions_.push_back(Condition{name_equals_matcher, scope_config});
+      return *this;
```
chusitoo, line 70: Be careful as this seems to create a copy of the builder every time (along with its conditions vector) so it could get expensive when there are more than a handful of conditions.

sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h
psx95: Initially avoided `emplace_back` since that would require writing a constructor in the `Condition` struct.

sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h
psx95: This is a good catch, thanks.

I have made the method return a reference type to prevent unneeded copies. Let me know if this alleviates your concerns.

psx95:
Hi @chusitoo,
Thanks for the review! I have addressed your comments.

Please take a look.

chusitoo (commented):
@psx95 thanks for the fixes. These changes look good to me 👍

sdk/test/trace/tracer_config_test.cc
```diff
@@ -0,0 +1,95 @@
...
+
+static std::pair<opentelemetry::nostd::string_view, opentelemetry::common::AttributeValue> attr1 = {
+    "accept_single_attr", true};
+static std::pair<opentelemetry::nostd::string_view, opentelemetry::common::AttributeValue> attr2 = {
+    "accept_second_attr", "some other attr"};
+static std::pair<opentelemetry::nostd::string_view, opentelemetry::common::AttributeValue> attr3 = {
+    "accept_third_attr", 3};
+
+static instrumentation_scope::InstrumentationScope test_scope_1 =
+    *instrumentation_scope::InstrumentationScope::Create("test_scope_1");
```
marcalff, line 49: It is weird to invoke Create() and then use the object behind the unique ptr, but given how InstrumentationScope does not expose public constructors, this is the best the test can do.

Approved as is then.

marcalff (approved):
LGTM, thanks for the feature.

PR #3146, opened 2024-11-15
[EXPORTER] fix clang-tidy warnings in UrlParser

sjinks:
## Changes

This PR fixes issues found by `clang-tidy` in UrlParser.

### DIFF
--- a/ext/include/opentelemetry/ext/http/common/url_parser.h
+++ b/ext/include/opentelemetry/ext/http/common/url_parser.h
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <string>
+#include <utility>
 
 #include "opentelemetry/version.h"
 
@@ -34,7 +35,7 @@ public:
   std::string query_;
   bool success_;
 
-  UrlParser(std::string url) : url_(url), success_(true)
+  UrlParser(std::string url) : url_(std::move(url)), success_(true)
   {
     if (url_.length() == 0)
     {
@@ -50,15 +51,16 @@ public:
     }
     else
     {
-      scheme_ = std::string(url_.begin() + cpos, url_.begin() + pos);
+      scheme_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                            url_.begin() + static_cast<std::string::difference_type>(pos));
       cpos    = pos + 3;
     }
 
     // credentials
-    size_t pos1 = url_.find_first_of("@", cpos);
-    size_t pos2 = url_.find_first_of("/", cpos);
+    size_t pos1 = url_.find_first_of('@', cpos);
     if (pos1 != std::string::npos)
     {
+      size_t pos2 = url_.find_first_of('/', cpos);
       // TODO - handle credentials
       if (pos2 == std::string::npos || pos1 < pos2)
       {
@@ -80,7 +82,8 @@ public:
     {
       // port present
       is_port = true;
-      host_   = std::string(url_.begin() + cpos, url_.begin() + pos);
+      host_   = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                            url_.begin() + static_cast<std::string::difference_type>(pos));
       cpos    = pos + 1;
     }
     pos = url_.find_first_of("/?", cpos);
@@ -97,7 +100,9 @@ public:
       }
       else
       {
-        host_ = std::string(url_.begin() + cpos, url_.begin() + url_.length());
+        host_ =
+            std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                        url_.begin() + static_cast<std::string::difference_type>(url_.length()));
       }
       return;
     }
@@ -109,7 +114,8 @@ public:
     }
     else
     {
-      host_ = std::string(url_.begin() + cpos, url_.begin() + pos);
+      host_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                          url_.begin() + static_cast<std::string::difference_type>(pos));
     }
     cpos = pos;
 
@@ -118,21 +124,27 @@ public:
       pos = url_.find('?', cpos);
       if (pos == std::string::npos)
       {
-        path_  = std::string(url_.begin() + cpos, url_.begin() + url_.length());
+        path_ =
+            std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                        url_.begin() + static_cast<std::string::difference_type>(url_.length()));
         query_ = "";
       }
       else
       {
-        path_  = std::string(url_.begin() + cpos, url_.begin() + pos);
-        cpos   = pos + 1;
-        query_ = std::string(url_.begin() + cpos, url_.begin() + url_.length());
+        path_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                            url_.begin() + static_cast<std::string::difference_type>(pos));
+        cpos  = pos + 1;
+        query_ =
+            std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                        url_.begin() + static_cast<std::string::difference_type>(url_.length()));
       }
       return;
     }
     path_ = std::string("/");
     if (url_[cpos] == '?')
     {
-      query_ = std::string(url_.begin() + cpos, url_.begin() + url_.length());
+      query_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                           url_.begin() + static_cast<std::string::difference_type>(url_.length()));
     }
   }
 
@@ -209,7 +221,7 @@ public:
         hex[1]      = encoded[++pos];
 
         char *endptr;
-        long value = strtol(hex, &endptr, 16);
+        int value = static_cast<int>(std::strtol(hex, &endptr, 16));
 
         // Invalid input: no valid hex characters after '%'
         if (endptr != &hex[2])
--- a/ext/include/opentelemetry/ext/http/common/url_parser.h
+++ b/ext/include/opentelemetry/ext/http/common/url_parser.h
@@ -51,8 +51,7 @@ public:
     }
     else
     {
-      scheme_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
-                            url_.begin() + static_cast<std::string::difference_type>(pos));
+      scheme_ = url_.substr(cpos, pos - cpos);
       cpos    = pos + 3;
     }
 
@@ -74,16 +73,19 @@ public:
     {
       // port missing. Used default 80 / 443
       if (scheme_ == "http")
+      {
         port_ = 80;
-      if (scheme_ == "https")
+      }
+      else if (scheme_ == "https")
+      {
         port_ = 443;
+      }
     }
     else
     {
       // port present
       is_port = true;
-      host_   = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
-                            url_.begin() + static_cast<std::string::difference_type>(pos));
+      host_   = url_.substr(cpos, pos - cpos);
       cpos    = pos + 1;
     }
     pos = url_.find_first_of("/?", cpos);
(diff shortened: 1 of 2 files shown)


ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -34,7 +35,7 @@ class UrlParser
   std::string query_;
   bool success_;
 
-  UrlParser(std::string url) : url_(url), success_(true)
+  UrlParser(std::string url) : url_(std::move(url)), success_(true)
```
sjinks, line 38: > Parameter 'url' is passed by value and only copied once; consider moving it to avoid unnecessary copies

https://clang.llvm.org/extra/clang-tidy/checks/performance/unnecessary-value-param.html

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -50,15 +51,16 @@ class UrlParser
     }
     else
     {
-      scheme_ = std::string(url_.begin() + cpos, url_.begin() + pos);
+      scheme_ = std::string(url_.begin() + static_cast<std::string::difference_type>(cpos),
+                            url_.begin() + static_cast<std::string::difference_type>(pos));
```
sjinks, line 55: This one and similar issues:

> Narrowing conversion from 'size_t' (aka 'unsigned long') to signed type 'difference_type' (aka 'long') is implementation-defined

https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/narrowing-conversions.html

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -50,15 +51,16 @@ class UrlParser
...
       cpos    = pos + 3;
     }
 
     // credentials
-    size_t pos1 = url_.find_first_of("@", cpos);
-    size_t pos2 = url_.find_first_of("/", cpos);
+    size_t pos1 = url_.find_first_of('@', cpos);
```
sjinks, line 60: > 'find_first_of' called with a string literal consisting of a single character; consider using the more effective overload accepting a character

https://clang.llvm.org/extra/clang-tidy/checks/performance/faster-string-find.html

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -209,7 +221,7 @@ class UrlDecoder
         hex[1]      = encoded[++pos];
 
         char *endptr;
-        long value = strtol(hex, &endptr, 16);
+        int value = static_cast<int>(std::strtol(hex, &endptr, 16));
```
sjinks, line 224: The `google-runtime-int` rule suggests replacing 'long' with 'int64', which seems overkill. Because two hexadecimal characters give a value in the [0; 255] range, downcasting to `int` is safe.

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: It looks correct, but my concern here is proliferation of casts in the code base, that makes the code harder to read, hiding the rotten tree in the forest should there be a broken cast somewhere.

Could this be simplified to something like:

```
scheme_ = url_.substr(cpos, pos);
```

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: Good point on using the `int` type.

How about `std::strtoi()` then ?

sjinks: It looks like there is only std::atoi() but it can only work with base 10 numbers 

https://en.cppreference.com/w/cpp/string/byte

ext/include/opentelemetry/ext/http/common/url_parser.h
sjinks: Yeah, this approach is much better. I have updated the code.

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -72,15 +73,19 @@ class UrlParser
     {
       // port missing. Used default 80 / 443
       if (scheme_ == "http")
+      {
         port_ = 80;
-      if (scheme_ == "https")
+      }
+      else if (scheme_ == "https")
```
sjinks, line 79: Added `else` here because otherwise, we will have two comparisons for `http`, and the second one will be unnecessary because we already have the value.

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -118,21 +119,21 @@ class UrlParser
       pos = url_.find('?', cpos);
       if (pos == std::string::npos)
       {
-        path_  = std::string(url_.begin() + cpos, url_.begin() + url_.length());
+        path_  = url_.substr(cpos, url_.length());
```
sjinks, line 122: Here and below: strictly speaking, it should be `(cpos, url_.length - cpos)` but because we want all remaining characters, we can save a few CPU cycles by not subtracting `cpos`.

sjinks:
I also tried to optimize `UrlDecoder::Decode()` a bit: https://godbolt.org/z/qnv1oqhcn (`Decode2` is the original version, `Decode` is the first attempt, `Decode3` is the final code)

sjinks:
@marcalff a bit of offtopic: I would like to make tests more readable by using parametrized tests. Among other things, this will show which of the test data failed.

Something like this:

```cpp
class UrlDecoderTests : public ::testing::TestWithParam<std::tuple<std::string, std::string>>
{};

INSTANTIATE_TEST_SUITE_P(
    SampleValues,
    UrlDecoderTests,
    testing::Values(std::make_tuple("Authentication=Basic xxx", "Authentication=Basic xxx"),
                    std::make_tuple("Authentication=Basic%20xxx", "Authentication=Basic xxx"),
                    std::make_tuple("%C3%B6%C3%A0%C2%A7%C3%96abcd%C3%84",
                                    "\xc3\xb6\xc3\xa0\xc2\xa7\xc3\x96\x61\x62\x63\x64\xc3\x84"),
                    std::make_tuple("%2x", "%2x"),
                    std::make_tuple("%20", " "),
                    std::make_tuple("text%2", "text%2")));

TEST_P(UrlDecoderTests, BasicTests)
{
  const auto encoded  = std::get<0>(GetParam());
  const auto expected = std::get<1>(GetParam());
  const auto actual   = http_common::UrlDecoder::Decode(encoded);

  EXPECT_EQ(actual, expected);
}
```

instead of

```cpp
TEST(UrlDecoderTests, BasicTests)
{
  std::map<std::string, std::string> testdata{
      {"Authentication=Basic xxx", "Authentication=Basic xxx"},
      {"Authentication=Basic%20xxx", "Authentication=Basic xxx"},
      {"%C3%B6%C3%A0%C2%A7%C3%96abcd%C3%84",
       "\xc3\xb6\xc3\xa0\xc2\xa7\xc3\x96\x61\x62\x63\x64\xc3\x84"},
      {"%2x", "%2x"},
      {"%20", " "},
      {"text%2", "text%2"},
  };

  for (auto &testsample : testdata)
  {
    ASSERT_EQ(http_common::UrlDecoder::Decode(testsample.first), testsample.second);
    ASSERT_TRUE(http_common::UrlDecoder::Decode(testsample.first) == testsample.second);
  }
}
```

The output will look like this (I intentionally made the test fail):
```
[----------] 6 tests from SampleValues/UrlDecoderTests
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/0
[       OK ] SampleValues/UrlDecoderTests.BasicTests/0 (0 ms)
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/1
[       OK ] SampleValues/UrlDecoderTests.BasicTests/1 (0 ms)
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/2
[       OK ] SampleValues/UrlDecoderTests.BasicTests/2 (0 ms)
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/3
[       OK ] SampleValues/UrlDecoderTests.BasicTests/3 (0 ms)
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/4
/home/volodymyr/work/github/opentelemetry-cpp/ext/test/http/url_parser_test.cc:252: Failure
Expected equality of these values:
  actual
    Which is: " "
  expected
    Which is: "-"

[  FAILED  ] SampleValues/UrlDecoderTests.BasicTests/4, where GetParam() = ("%20", "-") (0 ms)
[ RUN      ] SampleValues/UrlDecoderTests.BasicTests/5
[       OK ] SampleValues/UrlDecoderTests.BasicTests/5 (0 ms)
[----------] 6 tests from SampleValues/UrlDecoderTests (0 ms total)

[==========] 7 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 6 tests.
[  FAILED  ] SampleValues/UrlDecoderTests.BasicTests/4, where GetParam() = ("%20", "-")
```

See http://google.github.io/googletest/advanced.html#value-parameterized-tests

Is this something you would accept as a PR?

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: Sorry, typo. I meant `std::stoi`

https://en.cppreference.com/w/cpp/string/basic_string/stol

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: Even simpler, when we want all the remaining substring:

```
path_ = url_.substr(cpos);
```

https://en.cppreference.com/w/cpp/string/basic_string/substr

> (e.g. if count == npos), the returned substring is [pos, [size()](https://en.cppreference.com/w/cpp/string/basic_string/size))

ext/include/opentelemetry/ext/http/common/url_parser.h
```diff
@@ -193,36 +194,32 @@ class UrlDecoder
...
-        char hex[3] = {0};
-        hex[0]      = encoded[++pos];
-        hex[1]      = encoded[++pos];
-
-        char *endptr;
-        long value = strtol(hex, &endptr, 16);
-
-        // Invalid input: no valid hex characters after '%'
-        if (endptr != &hex[2])
+        if (hi != -1 && lo != -1)
```
marcalff, line 215: Previously, in this case:

```
// Invalid input: no valid hex characters after '%'
```

parsing stopped.

With this change, parsing continues, which is a change of behavior.

add an `else` and `return encoded;` ?

sjinks: Done; I have added two test cases as well.

ext/include/opentelemetry/ext/http/common/url_parser.h
sjinks: Updated the code.

ext/include/opentelemetry/ext/http/common/url_parser.h
sjinks: `std::stoi()` is a wrapper around `std::stol()`. It performs additional error checking and throws an exception if something is wrong. We would need a try/catch block in this case.

marcalff:
> @marcalff a bit of offtopic: I would like to make tests more readable by using parametrized tests. Among other things, this will show which of the test data failed.
> 

(...)

> 
> Is this something you would accept as a PR?

For a new test case added, I would definitively take it, this is good code.

For a PR that refactors test cases alone, say to investigate a hard to reproduce spurious failure to pinpoint which case is failing exactly, I would also take it.

The problem here is that we can not have refactoring both of production code (the url parser) and of unit tests for it, at the same time in the same PR.

This introduces the risk of adding a bug that goes undetected, even when admittedly the test case refactoring looks clean, so the probability of a bug in the unit test alone hiding a bug in the production code is very low.

That being said, I like the refactored test, as it will show how to write parameterized tests in opentelemetry-cpp in general, serving as an example for others, so I think it is important to have it.

As long as the unit test is refactored in a different PR (I think this is what you meant), I will (gladly) take it.
While at it, please add a test case for `Invalid input: no valid hex characters after '%'`, so the current behavior is captured and covered.

The test PR can be reviewed and merged immediately.

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: I see, thanks for investigating it.

I like the Decode refactoring better anyway, so this is resolved differently since the code no longer uses `std::strtol`

ext/include/opentelemetry/ext/http/common/url_parser.h
marcalff: Thanks, much cleaner. Waiting for CI to complete, then merging.

marcalff:
> I also tried to optimize `UrlDecoder::Decode()` a bit: https://godbolt.org/z/qnv1oqhcn (`Decode2` is the original version, `Decode` is the first attempt, `Decode3` is the final code)

The URL parser is used once at program startup, to parse the configuration of a OTLP exporter for example, so performance is hardly a concern, as this is not in any critical path.

The refactored code is also much cleaner, so even if this is a bit more than just fixing a clang-tidy warning, it is a good change to have (TLDR: approved).

marcalff (approved):
LGTM, thanks for the cleanup

Issue #3110, opened 2024-10-24
Please add number version in SONAMEs.


santiagorr:
Dear OpenTelemetry Team,

While packaging opentelemetry-cpp for Debian, I noticed the SDK's SONAME lacks a numerical version. While I understand the library has been carefully designed and considering ABI stability, IMO, it would be useful to add a numerical version in case you need to introduce breaking changes in the future. Bumping the SONAME would help to have smooth library transitions.

Could you please consider this?

Than you

marcalff:
Thanks for the report.

Agreed, opentelemetry-cpp needs to have this.

sjinks:
A test script to check for ABI changes between different releases (adapt as needed to add more tags or to build various exporters):

```sh
#!/bin/sh

mkdir -p dumps
first=v1.10.0
tags="v1.11.0 v1.12.0 v1.13.0 v1.14.0 v1.14.1 v1.14.2 v1.15.0 v1.16.0 v1.16.1 v1.17.0"
for tag in ${first} ${tags}; do
	rm -rf build
	git checkout "${tag}"
	cmake -B build -DBUILD_TESTING=OFF -DWITH_BENCHMARK=OFF -DCMAKE_CXX_FLAGS="-Og -g" -DBUILD_SHARED_LIBS=ON -DWITH_EXAMPLES=OFF
	cmake --build build -j
	cmake --install build --prefix "dist/${tag}"
	for i in "dist/${tag}/lib/"*.so; do abi-dumper "${i}" -o "dumps/${tag}-$(basename "${i}").dump" -lver "${tag}"; done
	git reset --hard "${tag}"
done

prev="${first}"
for tag in ${tags}; do
	for lib in "dist/${prev}/lib/"*.so; do
		name="$(basename "${lib}" .so)"
		base="$(basename "${lib}")"
		abi-compliance-checker -l "${name}" -old "dumps/${prev}-${base}.dump" -new dumps/${tag}-${base}.dump --extended
	done
	prev="${tag}"
done
```

sjinks:
The ABI has changed several times:

### libopentelemetry_metrics

Binary compatibility: 99.8%
Source compatibility: 99.96%
Total binary compatibility problems: 10, warnings: 4
Total source compatibility problems: 8, warnings: 0
Report: compat_reports/libopentelemetry_metrics/v1.10.0_to_v1.11.0/compat_report.html

Binary compatibility: 99.8%
Source compatibility: 99.8%
[...]

PR #3147, opened 2024-11-15
Fix lifetime for sdk::ReadWriteLogRecord

### LINKED ISSUE #2651: common::MakeAttributes creates garbled output for std::string
**Describe your environment**
OS: Debian bookworm
GCC: 12.2.0
Version: 054b0dc2 ([RELEASE] Release opentelemetry-cpp version 1.15.0 (#2639), 2024-04-21), also present in 1.11.0.

**Steps to reproduce**
- Download patch to ostream test [0001-Show-error-with-garbled-output.zip](https://github.com/open-telemetry/opentelemetry-cpp/files/15151907/0001-Show-error-with-garbled-output.zip)
- unzip 0001-Show-error-with-garbled-output.patch.zip
- git am 0001-Show-error-with-garbled-output.patch
- cmake -S . -B build -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=ON -DWITH_OTLP_GRPC=ON -DWITH_OTLP_HTTP=ON -DWITH_ABSEIL=ON -DWITH_BENCHMARK=OFF -DCMAKE_CXX_STANDARD=17
- cmake --build build --target ostream_log_test && ctest --test-dir build -R IntegrationTestWithEventId --verbose

Outputs:

```
...
539: {
539:   timestamp          : 0
539:   observed_timestamp : 1714407256398793342
539:   severity_num       : 5
539:   severity_text      : DEBUG
539:   body               : Hello {key1} {key2}
539:   resource           : 
539:     service.name: unknown_service
539:     telemetry.sdk.version: 1.15.0
539:     telemetry.sdk.name: opentelemetry
539:     telemetry.sdk.language: cpp
539:   attributes         : 
539:     key2: �fA�U
539: 
539: ��A����abcd
539:     key1: 123
539:   event_id           : 12345678
539:   event_name         : test_event_id
539:   trace_id           : 00000000000000000000000000000000
539:   span_id            : 0000000000000000
539:   trace_flags        : 00
539:   scope              : 
539:     name             : opentelelemtry_library
539:     version          : 1.15.0
539:     schema_url       : https://opentelemetry.io/schemas/1.11.0
539:     attributes       : 
539:       scope.attr.key: 123
539: }
...
```

If I use the `SimpleLogRecordProcessor` there is no garbled output. And if I pass in a `const char*` via `someString.c_str()` and use the batched processor I also get garbled output.

That the strings I sent are outputted properly.

Garbled output.

**Additional context**

I would also be interested to know if the bug is limited to the ostream exporter or also applies to other exporters, we want tot use otlp with grpc and http in production.

This looks like a lifetime issue with regarding common::AttributeValue.

### LINKED ISSUE #3135: heap-use-after-free with BatchLogRecordProcessor
**Describe your environment**
OpenTelemetry v1.17.0

**Steps to reproduce**
Sample code: https://github.com/sjinks/otel-heap-use-after-free
Sample run: https://github.com/sjinks/otel-heap-use-after-free/actions/runs/11758831588/job/32757617053

There should be no errors.

The application crashed because of the heap-use-after-free-error

```
=5853==ERROR: AddressSanitizer: heap-use-after-free on address 0x6030000005c8 at pc 0x55707b183c86 bp 0x7f82e12fe770 sp 0x7f82e12fdf38
READ of size 6 at 0x6030000005c8 thread T1
    #0 0x55707b183c85 in __interceptor_strlen (/home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/build/bug-repro+0x46c85) (BuildId: 3dbd37ecfe9134972fe6a448909a1b651fdd1e62)
    #1 0x7f82e3b3ceac in std::basic_ostream<char, std::char_traits<char> >& std::operator<<<std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*) (/lib/x86_64-linux-gnu/libstdc++.so.6+0x13ceac) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)
    #2 0x55707b247a33 in void opentelemetry::v1::exporter::ostream_common::print_value<char const*>(char const* const&, std::ostream&) /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/exporters/ostream/include/opentelemetry/exporters/ostream/common_utils.h:25:8
    #3 0x55707b2435cb in auto opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&)::'lambda'(auto&&)::operator()<char const* const&>(auto&&) const /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.
    #4 0x55707b24ab4d in auto std::__invoke_impl<void, opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&)::'lambda'(auto&&), char const* const&>(std::__invoke_other, opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, doub
    #5 0x55707b248277 in std::__invoke_result<auto, char const* const&>::type std::__invoke<opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&)::'lambda'(auto&&), char const* const&>(auto&&, char const* const&) /usr/include/c++/11/bits/invoke.h:96:40
    #6 0x55707b243a33 in std::__detail::__variant::__gen_vtable_impl<std::__detail::__variant::_Multi_array<std::__detail::__variant::__deduce_visit_result<void> (*)(opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&)::'lambda'(auto&&)&&, std::variant<bool, int, long, uns
    #7 0x55707b243f3a in decltype(auto) std::__do_visit<std::__detail::__variant::__deduce_visit_result<void>, opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&)::'lambda'(auto&&), std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd:
    #8 0x55707b243fa1 in std::invoke_result<auto, std::conditional<is_lvalue_reference_v<std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&>, std::variant_alternative<0ul, std::remove_reference<decltype(__as(declval<std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd
    #9 0x55707b23f9a8 in opentelemetry::v1::exporter::ostream_common::print_value(std::variant<bool, int, long, unsigned int, double, char const*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<bool const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned int const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<double const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<opentelemetry::v1::nostd::string_view const, 18446744073709551615ul>, unsigned long, opentelemetry::v1::nostd::span<unsigned long const, 18446744073709551615ul>, opentelemetry::v1::nostd::span<unsigned char const, 18446744073709551615ul> > const&, std::ostream&) /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/exporters/ostream/include/opentelemetry/exporters/ostream/commo
    #10 0x55707b23e5a4 in opentelemetry::v1::exporter::logs::OStreamLogRecordExporter::Export(opentelemetry::v1::nostd::span<std::unique_ptr<opentelemetry::v1::sdk::logs::Recordable, std::default_delete<opentelemetry::v1::sdk::logs::Recordable> >, 18446744073709551615ul> const&) /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/exporters/ostream/src/log_record_exporter.cc:111:57
    #11 0x55707b276fe7 in opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::Export() /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/sdk/src/logs/batch_log_record_processor.cc:223:22
    #12 0x55707b276cc6 in opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::DoBackgroundWork() /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/sdk/src/logs/batch_log_record_processor.cc:178:11
    #13 0x55707b27c429 in void std::__invoke_impl<void, void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*>(std::__invoke_memfun_deref, void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*&&)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*&&) /usr/include/c++/11/bits/invoke.h:74:46
    #14 0x55707b27c37c in std::__invoke_result<void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*>::type std::__invoke<void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*>(void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*&&)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*&&) /usr/include/c++/11/bits/invoke.h:96:40
    #15 0x55707b27c2dc in void std::thread::_Invoker<std::tuple<void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*> >::_M_invoke<0ul, 1ul>(std::_Index_tuple<0ul, 1ul>) /usr/include/c++/11/bits/std_thread.h:259:26
    #16 0x55707b27c22b in std::thread::_Invoker<std::tuple<void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*> >::operator()() /usr/include/c++/11/bits/std_thread.h:266:20
    #17 0x55707b27c1c3 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::*)(), opentelemetry::v1::sdk::logs::BatchLogRecordProcessor*> > >::_M_run() /usr/include/c++/11/bits/std_thread.h:211:20
    #18 0x7f82e3adc252  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xdc252) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)
    #19 0x7f82e3694ac2  (/lib/x86_64-linux-gnu/libc.so.6+0x94ac2) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #20 0x7f82e372684f  (/lib/x86_64-linux-gnu/libc.so.6+0x12684f) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
0x6030000005c8 is located 24 bytes inside of 30-byte region [0x6030000005b0,0x6030000005ce)
freed by thread T0 here:
    #0 0x55707b22bfbd in operator delete(void*) (/home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/build/bug-repro+0xeefbd) (BuildId: 3dbd37ecfe9134972fe6a448909a1b651fdd1e62)
    #1 0x7f82e3ac46f2 in std::runtime_error::~runtime_error() (/lib/x86_64-linux-gnu/libstdc++.so.6+0xc46f2) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)
    #2 0x7f82e3629d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
previously allocated by thread T0 here:
    #0 0x55707b22b75d in operator new(unsigned long) (/home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/build/bug-repro+0xee75d) (BuildId: 3dbd37ecfe9134972fe6a448909a1b651fdd1e62)
    #1 0x7f82e3af4978 in std::string::_Rep::_S_create(unsigned long, unsigned long, std::allocator<char> const&) (/lib/x86_64-linux-gnu/libstdc++.so.6+0xf4978) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)
Thread T1 created by T0 here:
    #0 0x55707b1d9e0c in __interceptor_pthread_create (/home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/build/bug-repro+0x9ce0c) (BuildId: 3dbd37ecfe9134972fe6a448909a1b651fdd1e62)
    #1 0x7f82e3adc328 in std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) (/lib/x86_64-linux-gnu/libstdc++.so.6+0xdc328) (BuildId: e37fe1a879783838de78cbc8c80621fa685d58a2)
    #2 0x55707b2764d0 in opentelemetry::v1::sdk::logs::BatchLogRecordProcessor::BatchLogRecordProcessor(std::unique_ptr<opentelemetry::v1::sdk::logs::LogRecordExporter, std::default_delete<opentelemetry::v1::sdk::logs::LogRecordExporter> >&&, opentelemetry::v1::sdk::logs::BatchLogRecordProcessorOptions const&) /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/sdk/src/logs/batch_log_record_processor.cc:58:7
    #3 0x55707b26c501 in opentelemetry::v1::sdk::logs::BatchLogRecordProcessorFactory::Create(std::unique_ptr<opentelemetry::v1::sdk::logs::LogRecordExporter, std::default_delete<opentelemetry::v1::sdk::logs::LogRecordExporter> >&&, opentelemetry::v1::sdk::logs::BatchLogRecordProcessorOptions const&) /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/vcpkg/buildtrees/opentelemetry-cpp/src/v1.17.0-b9ea5cab17.clean/sdk/src/logs/batch_log_record_processor_factory.cc:25:63
    #4 0x55707b22e18b in main /home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/repro.cpp:19:22
    #5 0x7f82e3629d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
SUMMARY: AddressSanitizer: heap-use-after-free (/home/runner/work/otel-heap-use-after-free/otel-heap-use-after-free/build/bug-repro+0x46c85) (BuildId: 3dbd37ecfe9134972fe6a448909a1b651fdd1e62) in __interceptor_strlen
Shadow bytes around the buggy address:
  0x0c067fff8060: fa fa fd fd fd fa fa fa fd fd fd fa fa fa fd fd
  0x0c067fff8070: fd fa fa fa fd fd fd fa fa fa fd fd fd fa fa fa
  0x0c067fff8080: fd fd fd fa fa fa 00 00 06 fa fa fa 00 00 03 fa
  0x0c067fff8090: fa fa 00 00 07 fa fa fa 00 00 00 00 fa fa 00 00
  0x0c067fff80a0: 00 fa fa fa 00 00 00 fa fa fa fd fd fd fa fa fa
=>0x0c067fff80b0: fd fd fd fa fa fa fd fd fd[fd]fa fa 00 00 00 fa
  0x0c067fff80c0: fa fa fd fd fd fa fa fa fd fd fd fa fa fa fd fd
  0x0c067fff80d0: fd fd fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff80e0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff80f0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c067fff8100: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==5853==ABORTING
```

**Additional context**
```cpp
    opentelemetry::sdk::logs::BatchLogRecordProcessorOptions options{2048, std::chrono::milliseconds(5000), 512};
    auto processor = opentelemetry::sdk::logs::BatchLogRecordProcessorFactory::Create(std::move(exporter), options);

//...

    try {
        throw std::runtime_error("error");
    }
    catch (const std::exception &e) {
        auto logger = opentelemetry::logs::Provider::GetLoggerProvider()->GetLogger("example");
        logger->Error(e.what());
    }
```

The bug happens when the `catch` block is left before the `BatchLogRecordProcessor` processes the data.

My understanding is that the log message is stored `ReadWriteLogRecord::body_` which is `opentelemetry::common::AttributeValue`, a `nostd::variant` holding a `const char*` value. When `e` goes out of the scope and gets destructed, the memory pointed to by `e.what()` is also freed. However, because `AttributeValue` does not own the data, it is left with a dangling pointer.

The fix will probably be to replace `AttributeValue` with `OwnedAttributeValue` (this will affect `body_` and `attributes_map_`).

Refs:
* https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.17.0/sdk/include/opentelemetry/sdk/logs/read_write_log_record.h#L191
* https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.17.0/sdk/include/opentelemetry/sdk/logs/read_write_log_record.h#L190

owent:
Fixes #3135 
Fixes #2651

## Changes

+ Add `MixedAttributeMap` to store both `std::unordered_map<std::string, opentelemetry::common::AttributeValue>` and `AttributeMap`
+ Copy both `OwnedAttributeValue`, `std::vector<nostd::string_view>` and `bool` array for `ReadWriteLogRecord`.

### DIFF
--- a/sdk/include/opentelemetry/sdk/common/attribute_utils.h
+++ b/sdk/include/opentelemetry/sdk/common/attribute_utils.h
@@ -6,6 +6,7 @@
 #include <cstdint>
 #include <initializer_list>
 #include <map>
+#include <memory>
 #include <string>
 #include <unordered_map>
 #include <utility>
@@ -48,6 +49,8 @@ using OwnedAttributeValue = nostd::variant<bool,
                                            uint64_t,
                                            std::vector<uint64_t>,
                                            std::vector<uint8_t>>;
+
+using OwnedAttributeView = nostd::variant<std::vector<nostd::string_view>, std::unique_ptr<bool[]>>;
 
 enum OwnedAttributeType
 {
@@ -295,6 +298,248 @@ public:
   }
 
 private:
+  AttributeConverter converter_;
+};
+
+/**
+ * Class for storing attributes.
+ */
+struct MixedAttributeMapStorage
+{
+  std::unordered_map<std::string, opentelemetry::common::AttributeValue> attributes;
+  AttributeMap owned_attributes;
+  std::unordered_map<std::string, OwnedAttributeView> owened_attributes_view;
+};
+
+/**
+ * Set an owned copy (OwnedAttributeValue) and attribute view of a non-owning AttributeValue.
+ */
+class MixedAttributeViewSetter
+{
+public:
+  inline MixedAttributeViewSetter(const nostd::string_view &key,
+                                  MixedAttributeMapStorage &storage,
+                                  AttributeConverter &converter) noexcept
+      : key_(&key), storage_(&storage), converter_(&converter)
+  {}
+
+  void operator()(bool v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(int32_t v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(uint32_t v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(int64_t v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(uint64_t v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(double v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    storage_->attributes[std::string(*key_)]       = v;
+  }
+
+  void operator()(nostd::string_view v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::string>(owned_value);
+  }
+
+  void operator()(const char *v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::string>(owned_value).c_str();
+  }
+
+  void operator()(nostd::span<const uint8_t> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<uint8_t>>(owned_value);
+  }
+
+  void operator()(nostd::span<const bool> v)
+  {
+    storage_->owned_attributes[std::string(*key_)] = (*converter_)(v);
+    if (v.empty())
+    {
+      storage_->attributes[std::string(*key_)] = nostd::span<const bool>{};
+    }
+    else
+    {
+      std::unique_ptr<bool[]> owned_view{new bool[v.size()]};
+      for (size_t i = 0; i < v.size(); i++)
+      {
+        owned_view[i] = v[i];
+      }
+
+      storage_->attributes[std::string(*key_)] =
+          nostd::span<const bool>{owned_view.get(), v.size()};
+      storage_->owened_attributes_view[std::string(*key_)] = std::move(owned_view);
+    }
+  }
+
+  void operator()(nostd::span<const int32_t> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<int32_t>>(owned_value);
+  }
+
+  void operator()(nostd::span<const uint32_t> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<uint32_t>>(owned_value);
+  }
+
+  void operator()(nostd::span<const int64_t> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<int64_t>>(owned_value);
+  }
+
+  void operator()(nostd::span<const uint64_t> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<uint64_t>>(owned_value);
+  }
+
+  void operator()(nostd::span<const double> v)
+  {
+    auto &owned_value                        = storage_->owned_attributes[std::string(*key_)];
+    owned_value                              = (*converter_)(v);
+    storage_->attributes[std::string(*key_)] = nostd::get<std::vector<double>>(owned_value);
+  }
+
+  void operator()(nostd::span<const nostd::string_view> v)
+  {
+    auto &owned_value = storage_->owned_attributes[std::string(*key_)];
+    owned_value       = (*converter_)(v);
+
+    if (v.empty())
+    {
+      storage_->attributes[std::string(*key_)] = nostd::span<const nostd::string_view>{};
+    }
+    else
+    {
+      auto &owned_view   = storage_->owened_attributes_view[std::string(*key_)];
+      owned_view         = std::vector<nostd::string_view>{};
+      auto &owned_vector = nostd::get<std::vector<nostd::string_view>>(owned_view);
+
+      owned_vector.reserve(v.size());
+      for (auto &data : nostd::get<std::vector<std::string>>(owned_value))
+      {
+        owned_vector.push_back(data);
+      }
+
+      storage_->attributes[std::string(*key_)] =
+          nostd::span<const nostd::string_view>{owned_vector.data(), owned_vector.size()};
+    }
+  }
+
+private:
+  const nostd::string_view *key_;
+  MixedAttributeMapStorage *storage_;
+  AttributeConverter *converter_;
+};
+
+/**
+ * Class for storing attributes and attribute view.
+ */
+class MixedAttributeMap
+{
+public:
+  // Construct empty attribute map
+  MixedAttributeMap() {}
+
+  // Construct attribute map and populate with attributes
+  MixedAttributeMap(const opentelemetry::common::KeyValueIterable &attributes) : MixedAttributeMap()
+  {
+    attributes.ForEachKeyValue(
+        [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
+          nostd::visit(MixedAttributeViewSetter(key, storage_, converter_), value);
+          return true;
+        });
+  }
+
+  // Construct attribute map and populate with optional attributes
+  MixedAttributeMap(const opentelemetry::common::KeyValueIterable *attributes) : MixedAttributeMap()
+  {
+    if (attributes != nullptr)
+    {
+      attributes->ForEachKeyValue(
+          [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
+            nostd::visit(MixedAttributeViewSetter(key, storage_, converter_), value);
+            return true;
+          });
+    }
+  }
+
+  // Construct map from initializer list by applying `SetAttribute` transform for every attribute
+  MixedAttributeMap(
+      std::initializer_list<std::pair<nostd::string_view, opentelemetry::common::AttributeValue>>
+          attributes)
+      : MixedAttributeMap()
+  {
+    for (auto &kv : attributes)
+    {
+      nostd::visit(MixedAttributeViewSetter(kv.first, storage_, converter_), kv.second);
+    }
+  }
+
+  // Returns a reference to this map
+  const std::unordered_map<std::string, opentelemetry::common::AttributeValue> &GetAttributes()
+      const noexcept
+  {
+    return storage_.attributes;
+  }
+
+  const AttributeMap &GetOwnedAttributes() const noexcept { return storage_.owned_attributes; }
+
+  // Convert non-owning key-value to owning std::string(key) and OwnedAttributeValue(value)
+  void SetAttribute(nostd::string_view key,
+                    const opentelemetry::common::AttributeValue &value) noexcept
+  {
+    nostd::visit(MixedAttributeViewSetter(key, storage_, converter_), value);
+  }
+
+  void Reserve(AttributeMap::size_type size)
+  {
+    storage_.attributes.reserve(size);
+    storage_.owned_attributes.reserve(size);
+  }
+
+  AttributeMap::size_type Size() const noexcept { return storage_.attributes.size(); }
+
+  bool Empty() const noexcept { return storage_.attributes.empty(); }
+
+private:
+  MixedAttributeMapStorage storage_;
   AttributeConverter converter_;
 };
 
--- a/sdk/include/opentelemetry/sdk/logs/readable_log_record.h
+++ b/sdk/include/opentelemetry/sdk/logs/readable_log_record.h
@@ -75,7 +75,12 @@ public:
    * Get body field of this log.
    * @return the body field for this log.
    */
-  virtual const opentelemetry::common::AttributeValue &GetBody() const noexcept = 0;
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  virtual const common::OwnedAttributeValue &
+#else
+  virtual const opentelemetry::common::AttributeValue &
+#endif
+  GetBody() const noexcept = 0;
 
   /**
    * Get the Event id.
@@ -111,7 +116,11 @@ public:
    * Get attributes of this log.
    * @return the body field of this log
    */
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  virtual const std::unordered_map<std::string, common::OwnedAttributeValue> &
+#else
   virtual const std::unordered_map<std::string, opentelemetry::common::AttributeValue> &
+#endif
   GetAttributes() const noexcept = 0;
 
   /**
--- a/sdk/include/opentelemetry/sdk/logs/readable_log_record.h
+++ b/sdk/include/opentelemetry/sdk/logs/readable_log_record.h
@@ -75,7 +75,7 @@ public:
    * Get body field of this log.
    * @return the body field for this log.
    */
-#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+#if !defined(OPENTELEMETRY_DEPRECATED_SDK_LOG_RECORD)
   virtual const common::OwnedAttributeValue &
 #else
   virtual const opentelemetry::common::AttributeValue &
@@ -116,7 +116,7 @@ public:
    * Get attributes of this log.
    * @return the body field of this log
    */
-#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+#if !defined(OPENTELEMETRY_DEPRECATED_SDK_LOG_RECORD)
   virtual const std::unordered_map<std::string, common::OwnedAttributeValue> &
 #else
   virtual const std::unordered_map<std::string, opentelemetry::common::AttributeValue> &
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,13 +63,15 @@ Important changes:
 * [SDK] Fix lifetime for sdk::ReadWriteLogRecord
   [#3147](https://github.com/open-telemetry/opentelemetry-cpp/pull/3245)
 
-  * With ABI version 2, `opentelemetry::sdk::logs::ReadableLogRecord::GetAttributes()`
+  * `opentelemetry::sdk::logs::ReadableLogRecord::GetAttributes()`
     will returns a `std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>`
     instead of a `std::unordered_map<std::string, opentelemetry::common::AttributeValue>`
     to keep the same type as the `opentelemetry::sdk::trace::SpanData`. And
     `opentelemetry::sdk::logs::ReadableLogRecord::GetBody()` will returns a
     `const opentelemetry::sdk::common::OwnedAttributeValue &` instead of a
-    `const opentelemetry::common::AttributeValue &`.
+    `const opentelemetry::common::AttributeValue &` now, we switch to the old
+    APIs by add `-DWITH_DEPRECATED_SDK_LOG_RECORD=ON` for cmake or add
+    `--copt -DOPENTELEMETRY_DEPRECATED_SDK_LOG_RECORD` for bazel.
 
 ## [1.20 2025-04-01]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,22 @@ Increment the:
 * [CMAKE] Add generated protobuf headers to the opentelemetry_proto target
   [#3400](https://github.com/open-telemetry/opentelemetry-cpp/pull/3400)
 
+* [SDK] Fix lifetime for sdk::ReadWriteLogRecord
+  [#3147](https://github.com/open-telemetry/opentelemetry-cpp/pull/3245)
+
+Important changes:
+
+* [SDK] Fix lifetime for sdk::ReadWriteLogRecord
+  [#3147](https://github.com/open-telemetry/opentelemetry-cpp/pull/3245)
+
+  * With ABI version 2, `opentelemetry::sdk::logs::ReadableLogRecord::GetAttributes()`
+    will returns a `std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>`
+    instead of a `std::unordered_map<std::string, opentelemetry::common::AttributeValue>`
+    to keep the same type as the `opentelemetry::sdk::trace::SpanData`. And
+    `opentelemetry::sdk::logs::ReadableLogRecord::GetBody()` will returns a
+    `const opentelemetry::sdk::common::OwnedAttributeValue &` instead of a
+    `const opentelemetry::common::AttributeValue &`.
+
 ## [1.20 2025-04-01]
 
 * [BUILD] Update opentelemetry-proto version
@@ -198,6 +214,9 @@ Important changes:
     `Provider` classes, to benefit from this feature.
 
   * All the example code has been updated to reflect the new usage.
+
+* [SDK] Implement spec: MetricFilter
+  [#3235](https://github.com/open-telemetry/opentelemetry-cpp/pull/3235)
 
 ## [1.19 2025-01-22]
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,8 +121,10 @@ if(VCPKG_CHAINLOAD_TOOLCHAIN_FILE)
   include("${VCPKG_CHAINLOAD_TOOLCHAIN_FILE}")
 endif()
 
-option(WITH_ABI_VERSION_1 "ABI version 1" ON)
+include(CMakeDependentOption)
 option(WITH_ABI_VERSION_2 "EXPERIMENTAL: ABI version 2 preview" OFF)
+cmake_dependent_option(WITH_ABI_VERSION_1 "ABI version 1" ON
+                       "NOT WITH_ABI_VERSION_2" OFF)
 
 file(READ "${CMAKE_CURRENT_LIST_DIR}/api/include/opentelemetry/version.h"
      OPENTELEMETRY_CPP_HEADER_VERSION_H)
@@ -490,7 +492,6 @@ if(WITH_OTLP_GRPC
     # Some versions of FindProtobuf.cmake uses mixed case instead of uppercase
     set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
   endif()
-  include(CMakeDependentOption)
 
   message(STATUS "PROTOBUF_PROTOC_EXECUTABLE=${PROTOBUF_PROTOC_EXECUTABLE}")
   set(SAVED_CMAKE_CXX_CLANG_TIDY ${CMAKE_CXX_CLANG_TIDY})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,13 @@ message(STATUS "OPENTELEMETRY_VERSION=${OPENTELEMETRY_VERSION}")
 
 option(WITH_NO_DEPRECATED_CODE "Do not include deprecated code" OFF)
 
+option(WITH_DEPRECATED_SDK_LOG_RECORD "Use deprecated SDK log record" OFF)
+
+if(WITH_DEPRECATED_SDK_LOG_RECORD)
+  message(
+    WARNING "WITH_DEPRECATED_SDK_LOG_RECORD=ON is temporary and deprecated")
+endif()
+
 set(WITH_STL
     "OFF"
     CACHE STRING "Which version of the Standard Library for C++ to use")
--- a/ci/do_ci.ps1
+++ b/ci/do_ci.ps1
@@ -67,6 +67,7 @@ switch ($action) {
     cmake $SRC_DIR `
       -DVCPKG_TARGET_TRIPLET=x64-windows `
       -DOPENTELEMETRY_BUILD_DLL=1 `
+      -DWITH_DEPRECATED_SDK_LOG_RECORD=ON `
      "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_DIR/scripts/buildsystems/vcpkg.cmake"
     $exit = $LASTEXITCODE
     if ($exit -ne 0) {
@@ -221,6 +222,7 @@ switch ($action) {
     cmake $SRC_DIR `
       -DVCPKG_TARGET_TRIPLET=x64-windows `
       -DWITH_ASYNC_EXPORT_PREVIEW=ON `
+      -DWITH_DEPRECATED_SDK_LOG_RECORD=ON `
       "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_DIR/scripts/buildsystems/vcpkg.cmake"
     $exit = $LASTEXITCODE
     if ($exit -ne 0) {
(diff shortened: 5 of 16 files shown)


sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -215,6 +218,248 @@ class OrderedAttributeMap : public std::map<std::string, OwnedAttributeValue>
   AttributeConverter converter_;
 };
 
+/**
+ * Class for storing attributes.
+ */
+struct MixedAttributeMapStorage
+{
+  std::unordered_map<std::string, opentelemetry::common::AttributeValue> attributes;
+  AttributeMap owened_attributes;
```
sjinks, line 227: Looks like a typo:

```suggestion
  AttributeMap owned_attributes;
```

sjinks:
I don't know if this is critical, but this change breaks the ABI:

* Problems with Data Types, High Severity
  * `read_write_log_record.h`, namespace `opentelemetry::v1::sdk::logs`, class `ReadWriteLogRecord`
    * Change: Size of this class has been increased from 176 bytes to 464 bytes.
    * Effect:
      1. An object of this class can be allocated by the applications and old size will be hardcoded at the compile time. Call of any exported constructor will break the memory of neighboring objects on the stack or heap.
      2. The memory layout and size of subclasses will be changed.
* Problems with Data Types, Low Severity
  * `read_write_log_record.h`, namespace `opentelemetry::v1::sdk::logs`, class `ReadWriteLogRecord`
    * Change: Type of field `attributes_map_` has been changed from `std::unordered_map<std::__cxx11::basic_string<char>, absl::otel_v1::variant<bool>, std::hash<std::__cxx11::basic_string<char> >, std::equal_to<std::__cxx11::basic_string<char> > >` (56 bytes) to `opentelemetry::v1::sdk::common::MixedAttributeMap` (184 bytes)
    * Change: Type of field `body_` has been changed from `opentelemetry::v1::common::AttributeValue` (24 bytes) to `opentelemetry::v1::sdk::common::MixedAttributeMap` (184 bytes)
    * Effect: Size of the inclusive type has been changed.

To reproduce:
```sh
cmake -B build -DBUILD_SHARED_LIBS=ON -DCMAKE_CXX_FLAGS="-Og -g3" --fresh
cmake --build build -j$(nproc) --clean-first
cmake --install build --prefix ORIG
wget https://patch-diff.githubusercontent.com/raw/open-telemetry/opentelemetry-cpp/pull/3147.diff
patch -p1 < 3147.diff
cmake --build build -j$(nproc)
cmake --install build --prefix NEW
mkdir dumps
for i in ORIG/lib/*.so; do abi-dumper "$i" -o "dumps/orig-$(basename "$i").dump" -lver orig; done
for i in NEW/lib/*.so; do abi-dumper "$i" -o "dumps/new-$(basename "$i").dump" -lver new; done
for i in NEW/lib/*.so; do abi-compliance-checker -l "$(basename "$i")" -old "dumps/orig-$(basename "$i").dump" -new "dumps/new-$(basename "$i").dump; done
```

It will show smth like
```
...
Preparing, please wait ...
Comparing ABIs ...
Comparing APIs ...
Creating compatibility report ...
Binary compatibility: 85.3%
Source compatibility: 100%
Total binary compatibility problems: 1, warnings: 2
Total source compatibility problems: 0, warnings: 0
Report: compat_reports/libopentelemetry_logs.so/orig_to_new/compat_report.html
...
```

Screenshot_20241116_131753

sdk/include/opentelemetry/sdk/common/attribute_utils.h
owent: Thanks, it's fixed now.

owent:
> I don't know if this is critical, but this change breaks the ABI:
> 
> * Problems with Data Types, High Severity
>   
>   * `read_write_log_record.h`, namespace `opentelemetry::v1::sdk::logs`, class `ReadWriteLogRecord`
>     
>     * Change: Size of this class has been increased from 176 bytes to 464 bytes.
>     * Effect:
>       
>       1. An object of this class can be allocated by the applications and old size will be hardcoded at the compile time. Call of any exported constructor will break the memory of neighboring objects on the stack or heap.
>       2. The memory layout and size of subclasses will be changed.
> * Problems with Data Types, Low Severity
>   
>   * `read_write_log_record.h`, namespace `opentelemetry::v1::sdk::logs`, class `ReadWriteLogRecord`
>     
>     * Change: Type of field `attributes_map_` has been changed from `std::unordered_map<std::__cxx11::basic_string<char>, absl::otel_v1::variant<bool>, std::hash<std::__cxx11::basic_string<char> >, std::equal_to<std::__cxx11::basic_string<char> > >` (56 bytes) to `opentelemetry::v1::sdk::common::MixedAttributeMap` (184 bytes)
>     * Change: Type of field `body_` has been changed from `opentelemetry::v1::common::AttributeValue` (24 bytes) to `opentelemetry::v1::sdk::common::MixedAttributeMap` (184 bytes)
>     * Effect: Size of the inclusive type has been changed.
> 
> To reproduce:
> 
> ```shell
> cmake -B build -DBUILD_SHARED_LIBS=ON -DCMAKE_CXX_FLAGS="-Og -g3" --fresh
> cmake --build build -j$(nproc) --clean-first
> cmake --install build --prefix ORIG
> wget https://patch-diff.githubusercontent.com/raw/open-telemetry/opentelemetry-cpp/pull/3147.diff
> patch -p1 < 3147.diff
> cmake --build build -j$(nproc)
> cmake --install build --prefix NEW
> mkdir dumps
> for i in ORIG/lib/*.so; do abi-dumper "$i" -o "dumps/orig-$(basename "$i").dump" -lver orig; done
> for i in NEW/lib/*.so; do abi-dumper "$i" -o "dumps/new-$(basename "$i").dump" -lver new; done
> for i in NEW/lib/*.so; do abi-compliance-checker -l "$(basename "$i")" -old "dumps/orig-$(basename "$i").dump" -new "dumps/new-$(basename "$i").dump; done
> ```
> 
> It will show smth like
> 
> ```
> ...
> Preparing, please wait ...
> Comparing ABIs ...
> Comparing APIs ...
> Creating compatibility report ...
> Binary compatibility: 85.3%
> Source compatibility: 100%
> Total binary compatibility problems: 1, warnings: 2
> Total source compatibility problems: 0, warnings: 0
> Report: compat_reports/libopentelemetry_logs.so/orig_to_new/compat_report.html
> ...
> ```
> 
> Screenshot_20241116_131753

If I didn't miss anything, I think we only need to keep ABI compatibility for api. These changes keep API compatibility for sdk and exporters.

sjinks:
> I think we only need to keep ABI compatibility for API.

I noticed that the library does not have a SOVERSION and thought that the changes could break the existing consumers (I know that [Alpine Linux ships OpenTelemetry C++ libraries](https://pkgs.alpinelinux.org/contents?name=opentelemetry-cpp&repo=testing&branch=edge&arch=x86_64)). That's why I asked :-)

owent:
> > I think we only need to keep ABI compatibility for API.
> 
> I noticed that the library does not have a SOVERSION and thought that the changes could break the existing consumers (I know that [Alpine Linux ships OpenTelemetry C++ libraries](https://pkgs.alpinelinux.org/contents?name=opentelemetry-cpp&repo=testing&branch=edge&arch=x86_64)). That's why I asked :-)

If we build otel-cpp with cmake, we can use `-DOTELCPP_VERSIONED_LIBS=ON` to set `SOVERSION`. Could you please try this option?

owent:
> > I think we only need to keep ABI compatibility for API.
> 
> I noticed that the library does not have a SOVERSION and thought that the changes could break the existing consumers (I know that [Alpine Linux ships OpenTelemetry C++ libraries](https://pkgs.alpinelinux.org/contents?name=opentelemetry-cpp&repo=testing&branch=edge&arch=x86_64)). That's why I asked :-)

If we build otel-cpp with cmake, we can use `-DOTELCPP_VERSIONED_LIBS=ON` to set `SOVERSION`. I'm not familar with bazel

> > I think we only need to keep ABI compatibility for API.
> 
> I noticed that the library does not have a SOVERSION and thought that the changes could break the existing consumers (I know that [Alpine Linux ships OpenTelemetry C++ libraries](https://pkgs.alpinelinux.org/contents?name=opentelemetry-cpp&repo=testing&branch=edge&arch=x86_64)). That's why I asked :-)

Thanks. I noticed #3110 and agree with the SOVERSION idea. Let's continue discussing it there.

lalitb:
@owent Thanks for the PR. With the use of  `MixedAttributeMap` construct, `ReadWriteLogRecord` will have now maintain both owning and non-owning types for `attributes` and `body` in its implementation due to the use of the `MixedAttributeMap` construct. Which mean even if some *custom* exporter needs to maintain non-owning type for the specific use-case, there would be extra baggage of the owning type. How about we have two different recordables - 
- ReadWriteLogRecord in its existing form which maintain non-owning values.
- ReadWriteLogRecordData - (similar to SpanData) which maintain owning values. The ostream exporter will use this as recordable.

owent:
> @owent Thanks for the PR. With the use of the `MixedAttributeMap` construct, `ReadWriteLogRecord` now maintains both owning and non-owning values for attributes and body. While this ensures flexibility, it also introduces extra overhead when only non-owning types are required by specific custom exporters.
> 
> How about having two separate recordables:
> 
> * **ReadWriteLogRecord** -  Retains the current behavior, maintaining non-owning values for attributes and body.
> * **LogRecordData** (Similar to [SpanData](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.17.0/sdk/include/opentelemetry/sdk/trace/span_data.h)) - this new structure would store owning values, and used for ostream exporter.

Good point. I will try to implement them later.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -215,6 +218,248 @@ class OrderedAttributeMap : public std::map<std::string, OwnedAttributeValue>
   AttributeConverter converter_;
 };
 
+/**
+ * Class for storing attributes.
+ */
+struct MixedAttributeMapStorage
```
lalitb, line 224: @owent  Do we still need this structure?

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -215,6 +218,248 @@ class OrderedAttributeMap : public std::map<std::string, OwnedAttributeValue>
...
+{
+  std::unordered_map<std::string, opentelemetry::common::AttributeValue> attributes;
+  AttributeMap owned_attributes;
+  std::unordered_map<std::string, OwnedAttributeView> owened_attributes_view;
+};
+
+/**
+ * Set an owned copy (OwnedAttributeValue) and attribute view of a non-owning AttributeValue.
+ */
+class MixedAttributeViewSetter
```
lalitb, line 234: @owent - Do we still need this class?

owent: It's still used by `MixedAttributeMap` , which is used by `LogRecordData`.

lalitb: @owent  As we agreed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/3147#issuecomment-2492705149), can we implement the LogRecordData to only contain the owning types, not the mixed types. The motive is to keep the public API simple and minimal. And this will be consistent with SpanData implementation.

owent: I think there are some differents between `LogRecordData` and `SpanData`.

`SpanData::GetAttributes` returns `std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>` but `LogRecordData::GetAttributes` returns `std::unordered_map<std::string, opentelemetry::common::AttributeValue>`.
We need storage to store `span<T>` in `opentelemetry::common::AttributeValue`, or do you think we can break the API of 
`LogRecordData::GetAttributes` ?

owent: Any idea about keep the mixed types or change the API of `ReadableLogRecord::GetAttributes`?

owent: I just raised another approach which will change the return types in ABI v2 to keep the same as LogRecordData.
Could you please review again? @lalitb @sjinks @marcalff

marcalff:
Another note to myself that I really really need to review this.

@owent Could you refresh the patch and merge with a recent main ? Also, make sure there are no changes under third/party.

Thanks.

owent:
> Another note to myself that I really really need to review this.
> 
> @owent Could you refresh the patch and merge with a recent main ? Also, make sure there are no changes under third_party.
> 
> Thanks.

Thanks, conflicts are resolved.

marcalff:
I don't understand all the reasoning that lead to this current patch, it looks way too complicated to me.

Either I totally missed something, or it can be done in a simpler way.

Please see a different attempt to resolve the same issues, and comment:

* #3417

owent:
> Is the update to the prometheus-cpp submodule expected?

Sorry, it's a mistake and it's removed now.

sdk/include/opentelemetry/sdk/logs/readable_log_record.h
```diff
@@ -75,7 +75,12 @@ class ReadableLogRecord : public Recordable
    * Get body field of this log.
    * @return the body field for this log.
    */
-  virtual const opentelemetry::common::AttributeValue &GetBody() const noexcept = 0;
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  virtual const common::OwnedAttributeValue &
+#else
+  virtual const opentelemetry::common::AttributeValue &
+#endif
+  GetBody() const noexcept = 0;
```
marcalff, line 83: I disagree with using `OPENTELEMETRY_ABI_VERSION_NO` here:

* The content of api/include has not changed in this PR, the `ABI` for the instrumentation interface is strictly the same before and after the fix, so there is no `ABI` change. The symbol `OPENTELEMETRY_ABI_VERSION_NO` is dedicated to the API part, it is not meant to keep track of SDK ABI changes, that can happen independently.

* Using a flag to provide a migration path is fine. If we want to do it, the flag should be similar to `WITH_DEPRECATED_SDK_FACTORY` / `OPENTELEMETRY_DEPRECATED_SDK_FACTORY` introduced by PR #2664 for example. Do not use `OPENTELEMETRY_ABI_VERSION_NO` for this.

Now, there is more to this:

* This change does not affect the instrumentation interface, so no instrumented code will be broken by this.

* This change does not affect the SDK configuration interface (defining an exporter, etc), so no application simply using the SDK to configure opentelemetry-cpp will be affected.

* The only way some user code can be affected is if the user code implements its own log record exporter, and reuses the `ReadableLogRecord` class from the SDK instead of defining its own recordable for its own implementation.

* Even if this happen to be the case, the user log record exporter is then by definition broken and crashing, the same way the ostream exporter is crashing today, precisely because this class implementation uses stale pointers.

In conclusion, I don't think using a feature flag to provide the fix (or not) resolves anything. To the contrary, it complicates the implementation.

marcalff (changes requested):
See comments, do not use the ABI version here.

sdk/include/opentelemetry/sdk/logs/readable_log_record.h
owent: I agree about the flag. However, regarding compatibility, the issue isn't a crash. The return type has changed, so if users implement their own exporters using `ReadableLogRecord`, their code won't compile. I think we can allow users time to migrate from the old SDK to the new version, rather than introducing an immediate breaking change.

owent:
There are a lot of commits before, I also rebase and squash it from mian again.

sdk/include/opentelemetry/sdk/logs/readable_log_record.h
marcalff: > The return type has changed, so if users implement their own exporters using `ReadableLogRecord`, their code won't compile.

Correct.

But also note that if a user has such code in a custom exporter, that exporter should be currently crashing, when using stale pointers in body or attributes.

Forcing users to fix a broken custom exporter implementation might not be that bad after all, even if the visible effect is a build break.

> I think we can allow users time to migrate from the old SDK to the new version, rather than introducing an immediate breaking change.

Assuming we want to take this path, it requires:

* to introduce a build flag, to expose the new return type versus the old one.
* to announce building with the old return type is deprecated, and will be removed in the future
* to raise a build deprecation warning, to force users to notice
* to document the migration path, and explain using the build flag is preferable
* in the next release, to remove the build flag, and only keep the new implementation
* in the next release notes, document that users have to adjust their code (they had 1 full release to do so)

This is technically possible, and has been done in the past.

But this has been done in cases where:

* the SDK configuration api got changed
* every application configuring the SDK was known to be affected

so in short when the amount of code / users affected was significant.

Here, the SDK configuration has not changed, and most if not all applications will require no changes at all.

Again, the only impact is for applications that:

* defines their own log record exporter
* reuse an SDK class in their implementation (instead of implementing their own recordable)

Taking dependencies on the SDK implementation implies some risks, applications doing this should be aware of possible consequences.

I understand this particular point is a matter of assessment and opinions, but in this case I think having the full build flag and migration path for a use case that is expected to be very narrow (custom exporter), is overkill.

owent:
Close as #3417 is more simple when breaking the ABI and APIs is acceptable.

marcalff:
> Close as #3417 is more simple when breaking the ABI and APIs is acceptable.

Thanks @owent

PR #3150, opened 2024-11-18
[Code health] Perform cppcheck cleanup

### LINKED ISSUE #2297: [CI] Add a C++ static code analyser in the build

chusitoo:
Fixes #2297

## Changes

- Cleaned up code based on current settings for CI to pass
- Some warnings are either false positives and/or fail in v2.13 but no longer reported in v2.14+, although this would require building cppcheck from source
- Also suppressed _exporters/etw/include/opentelemetry/exporters/etw/TraceLoggingDynamic.h_ as it has dependencies on Windows macros (partially covered by microsoft_sal.cfg but not entirely). An alternative to suppressing it, maybe as a future improvement, could be to supply a custom cfg file with rules for cppcheck to ingest.

## Remaining issues

To summarize the 9 outstanding false positives currently being reported by cppcheck v2.13, can be mitigated by using a newer release, more precisely:

**v2.14** fixes
- ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h:310:8: style: Virtual function 'CancelAllSessions' is called from destructor '~HttpClient()' at line 206. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h:88:8: style: Virtual function 'Shutdown' is called from destructor '~BatchLogRecordProcessor()' at line 325. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/multi_log_record_processor.h:46:8: style: Virtual function 'ForceFlush' is called from destructor '~MultiLogRecordProcessor()' at line 33. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/multi_log_record_processor.h:56:8: style: Virtual function 'Shutdown' is called from destructor '~MultiLogRecordProcessor()' at line 34. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/batch_span_processor.h:86:8: style: Virtual function 'Shutdown' is called from destructor '~BatchSpanProcessor()' at line 322. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/multi_span_processor.h:124:8: style: Virtual function 'Shutdown' is called from destructor '~MultiSpanProcessor()' at line 140. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/simple_processor.h:77:8: style: Virtual function 'Shutdown' is called from destructor '~SimpleSpanProcessor()' at line 88. Dynamic binding is not used. [virtualCallInConstructor]

**v2.16** fixes
- sdk/include/opentelemetry/sdk/common/circular_buffer.h:188:16: portability: 'data' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined. [arithOperationsOnVoidPointer]
- sdk/include/opentelemetry/sdk/common/circular_buffer.h:190:50: portability: 'data' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined. [arithOperationsOnVoidPointer]


.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  cppcheck:
+    runs-on: ubuntu-latest
```
marcalff, line 12: ubuntu-latest is only 22.04 at this point, because 24.04 is not yet flagged stable.

Please try ubuntu-24.04 instead, in hope to get a more recent cppcheck.

.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,76 @@
...
+            libbenchmark-dev
+
+      - name: Prepare CMake
+        run: |
+          mkdir build && cd build
+          CC="clang" CXX="clang++" cmake ..
+
+      - name: Run cppcheck
+        run: |
+          cppcheck \
```
marcalff, line 41: Please add another statement, to run `cppcheck --version`, so we can see the actual version used in CI logs.
This will help debugging.

marcalff (commented):

Please add a separate PR for the CI alone, which can be merged right away in the main branch. Comment out the last step, to not fail the build on errors, this will be enforced when cleanup is completed.

This will allow us reviewers to actually see the current cppcheck errors in the log, which will facilitate review, as we won't have to ask "why ?" or "what was the error ?" on various lines fixed.

After CI is merged, this PR will contain only the cleanup fixes then.

Comments on actual code cleanup to follow.

api/include/opentelemetry/baggage/baggage_context.h
```diff
@@ -27,7 +27,7 @@ inline nostd::shared_ptr<Baggage> GetBaggage(const context::Context &context) no
 }
 
 inline context::Context SetBaggage(context::Context &context,
-                                   nostd::shared_ptr<Baggage> baggage) noexcept
+                                   const nostd::shared_ptr<Baggage> &baggage) noexcept
```
marcalff, line 30: Providing details because this is not trivial, as justification, for all reviewers:

* ABI change (the prototype is different)
* on a helper API
* which is inlined anyway, expanded in the caller compile unit

Hence:

* there is **no** ABI change overall, as no API implemented by the SDK is touched

Approved, this is safe IMO.

.github/workflows/cppcheck.yml
chusitoo: Definitely! It did not make a difference in terms of issues flagged but the versions of cppcheck were different. 

In ubuntu-latest it installed v2.7 whereas with ubuntu-24.04 we get 2.13, which is what I was already testing with locally.

api/include/opentelemetry/context/context.h
```diff
@@ -29,16 +29,14 @@ class Context
   // hold a shared_ptr to the head of the DataList linked list
   template <class T>
   Context(const T &keys_and_values) noexcept
-  {
-    head_ = nostd::shared_ptr<DataList>{new DataList(keys_and_values)};
-  }
+      : head_{nostd::shared_ptr<DataList>{new DataList(keys_and_values)}}
+  {}
```
marcalff, line 33: Approved as ABI safe, implementation change only.

api/include/opentelemetry/context/context.h
```diff
@@ -29,16 +29,14 @@ class Context
...
 
   // Creates a context object from a key and value, this will
   // hold a shared_ptr to the head of the DataList linked list
   Context(nostd::string_view key, ContextValue value) noexcept
-  {
-    head_ = nostd::shared_ptr<DataList>{new DataList(key, value)};
-  }
+      : head_{nostd::shared_ptr<DataList>{new DataList(key, value)}}
+  {}
```
marcalff, line 39: Approved as ABI safe, implementation change only.

.github/workflows/cppcheck.yml
chusitoo: Done in CI only PR #3151

api/include/opentelemetry/context/context.h
```diff
@@ -92,22 +90,21 @@ class Context
...
   {
-  public:
-    char *key_;
+    char *key_ = nullptr;
 
-    nostd::shared_ptr<DataList> next_;
+    size_t key_length_ = 0UL;
 
-    size_t key_length_;
+    nostd::shared_ptr<DataList> next_;
```
marcalff, line 99: Why reordering the members here ?

chusitoo: If I recall, it started complaining about member ordering when I added the copy ctor that it asked [here](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/11902741932/job/33168372488?pr=3151#step:4:55)

I just re-ran the the test with `next_` back up above `key_length_` and cppcheck seems to be more accepting of it, so it must've been something else while the changes were still work in progress on my end.

Do you want me to revert this ordering?

api/include/opentelemetry/context/context.h
```diff
@@ -133,8 +130,17 @@ class Context
       key_        = new char[key.size()];
       key_length_ = key.size();
       memcpy(key_, key.data(), key.size() * sizeof(char));
-      value_ = value;
       next_  = nostd::shared_ptr<DataList>{nullptr};
+      value_ = value;
+    }
+
+    DataList(const DataList &other)
```
marcalff, line 137: Ok to have a copy constructor, this clears the cppcheck warning, and is safer in case of refactoring.

I suspect this code to be unreachable.

api/include/opentelemetry/context/propagation/global_propagator.h
```diff
@@ -32,7 +32,7 @@ class OPENTELEMETRY_EXPORT GlobalTextMapPropagator
     return nostd::shared_ptr<TextMapPropagator>(GetPropagator());
   }
 
-  static void SetGlobalPropagator(nostd::shared_ptr<TextMapPropagator> prop) noexcept
+  static void SetGlobalPropagator(const nostd::shared_ptr<TextMapPropagator> &prop) noexcept
```
marcalff, line 35: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/context/runtime_context.h
```diff
@@ -152,7 +152,8 @@ class OPENTELEMETRY_EXPORT RuntimeContext
    *
    * @param storage a custom runtime context storage
    */
-  static void SetRuntimeContextStorage(nostd::shared_ptr<RuntimeContextStorage> storage) noexcept
+  static void SetRuntimeContextStorage(
+      const nostd::shared_ptr<RuntimeContextStorage> &storage) noexcept
```
marcalff, line 156: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/logs/event_id.h
```diff
@@ -19,9 +19,9 @@ namespace logs
 class EventId
 {
 public:
-  EventId(int64_t id, nostd::string_view name) noexcept : id_{id}
+  EventId(int64_t id, nostd::string_view name) noexcept
+      : id_{id}, name_{nostd::unique_ptr<char[]>{new char[name.length() + 1]}}
```
marcalff, line 23: ABI safe, internal implementation change.

api/include/opentelemetry/logs/provider.h
```diff
@@ -39,7 +39,7 @@ class OPENTELEMETRY_EXPORT Provider
   /**
    * Changes the singleton LoggerProvider.
    */
-  static void SetLoggerProvider(nostd::shared_ptr<LoggerProvider> tp) noexcept
+  static void SetLoggerProvider(const nostd::shared_ptr<LoggerProvider> &tp) noexcept
```
marcalff, line 42: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/logs/provider.h
```diff
@@ -60,7 +60,7 @@ class OPENTELEMETRY_EXPORT Provider
   /**
    * Changes the singleton EventLoggerProvider.
    */
-  static void SetEventLoggerProvider(nostd::shared_ptr<EventLoggerProvider> tp) noexcept
+  static void SetEventLoggerProvider(const nostd::shared_ptr<EventLoggerProvider> &tp) noexcept
```
marcalff, line 63: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/metrics/provider.h
```diff
@@ -38,7 +38,7 @@ class Provider
   /**
    * Changes the singleton MeterProvider.
    */
-  static void SetMeterProvider(nostd::shared_ptr<MeterProvider> tp) noexcept
+  static void SetMeterProvider(const nostd::shared_ptr<MeterProvider> &tp) noexcept
```
marcalff, line 41: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/plugin/tracer.h
```diff
@@ -20,7 +20,7 @@ class DynamicLibraryHandle;
 class Span final : public trace::Span
 {
 public:
-  Span(std::shared_ptr<trace::Tracer> &&tracer, nostd::shared_ptr<trace::Span> span) noexcept
+  Span(std::shared_ptr<trace::Tracer> &&tracer, const nostd::shared_ptr<trace::Span> &span) noexcept
```
marcalff, line 23: Here changing the span constructor does not change the object layout or virtual table, should be ok.

This plugin class in particular is not inherited from in the SDK, changing the constructor is ok.

Approved as ABI safe.

Beside, plugin/tracer.h is not used as far as I know.

api/include/opentelemetry/trace/default_span.h
```diff
@@ -63,7 +63,7 @@ class DefaultSpan : public Span
 
   nostd::string_view ToString() const noexcept { return "DefaultSpan"; }
 
-  DefaultSpan(SpanContext span_context) noexcept : span_context_(span_context) {}
+  DefaultSpan(const SpanContext &span_context) noexcept : span_context_(span_context) {}
```
marcalff, line 66: Not sure if this one is safe, due to:

```
opentracing-shim/src/tracer_shim.cc:        new opentelemetry::trace::DefaultSpan(span_context)};
```

api/include/opentelemetry/trace/context.h
```diff
@@ -34,7 +34,8 @@ inline bool IsRootSpan(const context::Context &context) noexcept
 }
 
 // Set Span into explicit context
-inline context::Context SetSpan(context::Context &context, nostd::shared_ptr<Span> span) noexcept
+inline context::Context SetSpan(context::Context &context,
+                                const nostd::shared_ptr<Span> &span) noexcept
```
marcalff, line 38: Helper inlined.
Approved as ABI safe.

api/include/opentelemetry/trace/span_context.h
```diff
@@ -41,7 +41,7 @@ class SpanContext final
               SpanId span_id,
               TraceFlags trace_flags,
               bool is_remote,
-              nostd::shared_ptr<TraceState> trace_state = TraceState::GetDefault()) noexcept
+              const nostd::shared_ptr<TraceState> &trace_state = TraceState::GetDefault()) noexcept
```
marcalff, line 44: Class SpanContext is used almost everywhere.

Not sure if this change breaks the ABI or not, this needs more analysis.

api/include/opentelemetry/context/context.h
marcalff: Yes, please revert. Better not take chances with binary layout if we can avoid it.

In theory, theory and practice are the same.
In practice, they are different.

marcalff: Thinking more about it, I think changing the order of members is a breaking change:

* an instrumented library A built with otel-cpp v1.17.0 (without this change)
* an instrumented library B built with otel-cpp v1.18.0 (with this change)

should see the same bits in the same place, when manipulating a runtime context using their own copy of `class Context`, which includes `DataList`, when both libraries A and B are linked in the same application.

The runtime context is shared between all instrumented libraries, because it comes from a global singleton.

Same for the baggage propagator.

marcalff:
@chusitoo The CI is now merged to main, please merge main into this PR to resolve the conflicts.

api/include/opentelemetry/context/context.h
chusitoo: I remember now; the issue came up not with cppcheck but with `-Werror=reorder` in maintainer mode CI checks and it felt more natural to initialize the key and key length back-to-back so I had swapped the order of the member variables instead the order of init in the copy ctor.

In any case, they are reverted to their original order and the change is in now in the copy constructor to respect the original order of initialization.

api/include/opentelemetry/trace/span_context.h
chusitoo: I think there would be a way to compromise on the direct requirement from cppcheck in order to ensure ABI compatibility. I reverted this change (pass by const ref -> pass by value) and move the copy of that shared_ptr into the member variable. 

This should be comparable since it doesn't waste making an extra copy and incrementing/decrementing the use count. 
Verified that cppcheck agrees with this change and no longer reports it as problematic.

api/include/opentelemetry/trace/default_span.h
chusitoo: Reverted this change, just to be on the safe side. Instead, we move the copy into the member variable, similar to the latest change for [span_context.h](https://github.com/open-telemetry/opentelemetry-cpp/pull/3150#discussion_r1849343531)

.github/workflows/cppcheck.yml
```diff
@@ -73,4 +58,5 @@ jobs:
           set +e
           COUNT=`grep -c -E "\[.+\]" cppcheck.log`
           echo "cppcheck reported ${COUNT} warning(s)"
-          if [ $COUNT -ne 0 ] ; then exit 1 ; fi
+          # TODO: uncomment to enforce failing the build
+          # if [ $COUNT -ne 0 ] ; then exit 1 ; fi
```
chusitoo: Should this check be re-enabled now?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -904,9 +898,11 @@ OtlpHttpClient::createSession(
       {
         std::cerr << error_message << '\n';
       }
-      OTEL_INTERNAL_LOG_ERROR(error_message.c_str());
+      OTEL_INTERNAL_LOG_ERROR(error_message);
 
-      return opentelemetry::sdk::common::ExportResult::kFailure;
+      const auto result = opentelemetry::sdk::common::ExportResult::kFailure;
+      result_callback(result);
```
marcalff, line 904: Here and in other places, the call to result_callback() has moved a bit.

Could you provide more details ?

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -309,6 +307,7 @@ class HttpClient : public opentelemetry::ext::http::client::HttpClient
   std::shared_ptr<opentelemetry::ext::http::client::Session> CreateSession(
       nostd::string_view url) noexcept override;
 
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 310: Please remove the suppression.

Instead, if this still raises an warning in cppcheck, we need to find where from, and fix the caller code.

sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h
```diff
@@ -85,6 +85,7 @@ class BatchLogRecordProcessor : public LogRecordProcessor
    *
    * NOTE: Timeout functionality not supported yet.
    */
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 88: Remove the suppression, and identify/fix callers.

sdk/include/opentelemetry/sdk/logs/multi_log_record_processor.h
```diff
@@ -53,6 +54,7 @@ class MultiLogRecordProcessor : public LogRecordProcessor
    * shutdown before giving up and returning failure.
    * @return true if the shutdown succeeded, false otherwise
    */
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 57: Remove the suppression, and identify/fix callers.

sdk/include/opentelemetry/sdk/trace/batch_span_processor.h
```diff
@@ -83,6 +83,7 @@ class BatchSpanProcessor : public SpanProcessor
    *
    * NOTE: Timeout functionality not supported yet.
    */
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 86: Remove the suppression, and identify/fix callers.

sdk/include/opentelemetry/sdk/trace/multi_span_processor.h
```diff
@@ -121,6 +121,7 @@ class MultiSpanProcessor : public SpanProcessor
     return result;
   }
 
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 124: Remove the suppression, and identify/fix callers.

sdk/include/opentelemetry/sdk/trace/simple_processor.h
```diff
@@ -74,6 +74,7 @@ class SimpleSpanProcessor : public SpanProcessor
     return true;
   }
 
+  // cppcheck-suppress [virtualCallInConstructor]
```
marcalff, line 77: Remove the suppression, and identify/fix callers.

api/include/opentelemetry/trace/default_span.h
marcalff: Thanks, the change is much better now.

.github/workflows/cppcheck.yml
marcalff: Yes, please do.

If there are unresolved warnings left, it is ok to test with some margin, say  `count > 5`.

The point is to prevent new errors to come in with more merges, while giving us time to resolve possibly more complex issues in follow up PRs and ultimately have `count > 0`.

marcalff (commented):
Thanks for the cleanup.

All the code under api/ looks good to me, but I would like to have a second opinion (@lalitb , @ThomsonTan , @esigo) due to the risk of breaking the ABI.

See new comments for non api code.

exporters/otlp/src/otlp_http_client.cc
chusitoo: Absolutely. The scan reports an _[accessMoved]_ warning [here](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/11948772032/job/33306927417#step:4:440)

This is due to the move that happens on higher `OtlpHttpClient::Export()` [L753](https://github.com/open-telemetry/opentelemetry-cpp/pull/3150/files/e9b225295e05a2baf71708e6486036042ed69542#diff-eb7f99ebc152d9cc0f4781cc0a0388907e4d4b58271dba85eba99fd16b880674R753) which looks like it is likely undefined behavior calling it from that point on.

Since the **_moved from_** result_callback appears to be used only to consume the result from `createSession()`, my solution was to move the invocation closer to the call sites where it's been **_moved to_**.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
chusitoo: Please note that this and all other suppressions of `[virtualCallInConstructor]` are false positives no longer reported in v2.14+, as highlighted in the [changelog summary](https://github.com/danmar/cppcheck/releases/tag/2.14.0).

I will revert, as requested. As a future improvement, using a more updated version of cppcheck (built from source) should mitigated these warnings.

chusitoo: @marcalff I am noticing that the changes I had put in circular_buffer.h to suppress both `[arithOperationsOnVoidPointer]` are no longer reported in v2.16, so I am wondering if these false positives waiting to be addressed should be documented somewhere. I think this would be beneficial to avoid repeat investigation when the fix is updating the scanner version.

.github/workflows/cppcheck.yml
chusitoo: Made the `WARNING_LIMIT = 10` to match your PR for iwyu part 5 and to accommodate the 9 false positives that are no longer suppressed. I also thought it would be a good idea to make it stand out a bit more when the count is below threshold but not zero, by emitting a warning annotation.

image

marcalff: Very nice.

I am applying this for include-what-you-use as well.

exporters/otlp/src/otlp_http_client.cc
marcalff: Good call on `result_callback`, thanks for finding and fixing this.

marcalff (approved):
LGTM, excellent cleanup.

@lalitb , @ThomsonTan , @esigo,

Please double check the code under `api/` in particular, due to the risk of breaking the ABI.

I left comments to indicate why I think each change is safe.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
chusitoo: Good suggestion, thanks!

To summarize the 9 outstanding false positives currently being reported by cppcheck v2.13, can be mitigated by using a newer release, more precisely:

**v2.14** fixes
- ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h:310:8: style: Virtual function 'CancelAllSessions' is called from destructor '~HttpClient()' at line 206. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/batch_log_record_processor.h:88:8: style: Virtual function 'Shutdown' is called from destructor '~BatchLogRecordProcessor()' at line 325. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/multi_log_record_processor.h:46:8: style: Virtual function 'ForceFlush' is called from destructor '~MultiLogRecordProcessor()' at line 33. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/logs/multi_log_record_processor.h:56:8: style: Virtual function 'Shutdown' is called from destructor '~MultiLogRecordProcessor()' at line 34. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/batch_span_processor.h:86:8: style: Virtual function 'Shutdown' is called from destructor '~BatchSpanProcessor()' at line 322. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/multi_span_processor.h:124:8: style: Virtual function 'Shutdown' is called from destructor '~MultiSpanProcessor()' at line 140. Dynamic binding is not used. [virtualCallInConstructor]
- sdk/include/opentelemetry/sdk/trace/simple_processor.h:77:8: style: Virtual function 'Shutdown' is called from destructor '~SimpleSpanProcessor()' at line 88. Dynamic binding is not used. [virtualCallInConstructor]

**v2.16** fixes
- sdk/include/opentelemetry/sdk/common/circular_buffer.h:188:16: portability: 'data' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined. [arithOperationsOnVoidPointer]
- sdk/include/opentelemetry/sdk/common/circular_buffer.h:190:50: portability: 'data' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined. [arithOperationsOnVoidPointer]

api/include/opentelemetry/context/context.h
```diff
@@ -133,8 +130,17 @@ class Context
...
+        : key_(new char[other.key_length_]),
+          next_(other.next_),
+          key_length_(other.key_length_),
+          value_(other.value_)
+    {
+      memcpy(key_, other.key_, other.key_length_ * sizeof(char));
```
sjinks, line 143:
```suggestion
      std::memcpy(key_, other.key_, other.key_length_ * sizeof(char));
```

because we `#include <cstring>`, not `<string.h>`

lalitb (approved):
reviewed for ABI breaking changes in API. Looks good from that prospective.

PR #3153, opened 2024-11-19
[TEST] refactor UrlParser tests to use value-paramterized tests

sjinks:
As discussed in [#3146](https://github.com/open-telemetry/opentelemetry-cpp/pull/3146#issuecomment-2486663185)

## Changes

This PR refactors the UrlParser tests to make them more readable and use value-parametrized tests.


ext/test/http/url_parser_test.cc
```diff
@@ -2,248 +2,136 @@
...
 #include <string>
-#include <utility>
+#include <tuple>
 
 #include "opentelemetry/ext/http/common/url_parser.h"
 
 namespace http_common = opentelemetry::ext::http::common;
 
-inline const char *BoolToString(bool b)
+struct ParsedUrl
```
sjinks, line 12: I used a dedicated structure instead of `std::map<std::string, std::string>` because map's `operator[]` may change the map (if a key does not exist, it will create a default-constructed value and put it into the map). Test data must be immutable.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
Waiting a bit for possible comments, then will merge.

Issue #2483, opened 2024-01-10
Add support for embedded devices?


proffalken:
Before opening a feature request against this repo, consider whether the feature should/could be implemented in the [other OpenTelemetry client libraries](https://github.com/open-telemetry/). If so, please [open an issue on opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/new) first.

**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
Many embedded devices run C++ as their primary programming language.  This includes both hobbyist devices such as Arduino's and industrial IoT devices based on chips from various vendors.

Whilst each chip has it's own "special requirements", being able to add Logs, Traces, and Metrics to embedded devices (especially those that have a robust data connection to the outside world!) would provide huge insight into how microcontroller-based embedded devices interact with the systems they talk to.

**Describe alternatives you've considered**

   * Creating a w3c Trace Context TraceID manually as part of the application code and using existing logging libraries to send that data to OTEL collector for processing

**Additional context**

   * Common transports from embedded IoT devices include HTTP and MQTT
   * LoRaWAN and other 5G technologies are also options, however payload restrictions will limit what can be sent over some of these options, so I'd recommend only using WiFi, Cellular, or Ethernet as "outbound" options.

marcalff:
As long as the code builds for the target architecture, and as long as the device has network connectivity to send data out, I don't see what is fundamentally different from a "regular" deployment.

Could you clarify what you think is missing in particular in opentelemetry-cpp ?

Without knowing in detail:

* the device to use
* the device platform
* the compiler to use, possibly a cross compiler
* the libraries available

there is not much opentelemetry-cpp can do to "support embedded devices", this statement is too general.

Still, I would love to see a `tic-tac-toe` app instrumented, as a demo.

proffalken:
Apologies, I'd missed this comment.

As an initial target device I'd love to get this running on an ESP32, something like [this one](https://www.amazon.co.uk/ESP-32S-Development-2-4GHz-Bluetooth-Antenna/dp/B071JR9WS9/)

My knowledge of c++ and embedded systems is "just enough to get by", so I'm not too sure how I'd get the SDK into a suitable state for use with these kinds of devices, however being able to trace from the call to the sensor right through to the interaction with the backend storage would be amazing!

The compiler would be via ESPressif's toolchain and using [PlatformIO](https://platformio.org/) in an ideal situation.

[...]

PR #3155, opened 2024-11-20
[BUILD] Fix build for esp32

albkharisov:
Fix for toolchains that alias `int32_t` to `long int`. In this way we don't have `int` in variant of attributes.
If it's defined as strict int32_t, int64_t etc, I think it's better to have strict types everywhere.
ESP-IDF (v5.3.1) toolchain fails to compile as they don't have int in attributes variant list.

Fixes # (issue)

## Changes


marcalff (changes requested):
Approved the workflow to run CI, but blocking the merge.

I would like first to understand the issue in more details,
to decide the best course of action for the fix.

albkharisov:
@marcalff 
Okay, here are the details:

I've integrated the opentelemetry-cpp library for ESP32 project. In the middle of this process main difficulty was to make it buildable on ESP-IDF (v5.3.1) toolchain.
ESP-IDF compiler is: `xtensa-esp-elf-g++ (crosstool-NG esp-13.2.0_20240530) 13.2.0`

Repo was configured as follows:
```
cmake -B build \
    -DCMAKE_TOOLCHAIN_FILE=/opt/esp-idf2/tools/cmake/toolchain-esp32s3.cmake \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_POSITION_INDEPENDENT_CODE=FALSE \
    -DWITH_ZIPKIN=OFF \
    -DBUILD_TESTING=OFF \
    -DWITH_OTLP_GRPC=OFF \
    -DWITH_OTLP_HTTP=OFF \
    -DBUILD_SHARED_LIBS=OFF \
    -DCMAKE_INSTALL_PREFIX=./build/optl \
    -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
    -DWITH_EXAMPLES=OFF \
    -DCMAKE_SYSTEM_PROCESSOR=riscv
```
CMAKE_SYSTEM_PROCESSOR is here only to pass cmake configure checks, it doesn't affect the OT project. I suppose we have to add here 'xtensa' too.
Toolchain file is common:
```
include($ENV{IDF_PATH}/tools/cmake/utilities.cmake)

set(CMAKE_SYSTEM_NAME Generic)

set(CMAKE_C_COMPILER xtensa-esp32s3-elf-gcc)
set(CMAKE_CXX_COMPILER xtensa-esp32s3-elf-g++)
set(CMAKE_ASM_COMPILER xtensa-esp32s3-elf-gcc)
set(_CMAKE_TOOLCHAIN_PREFIX xtensa-esp32s3-elf-)

remove_duplicated_flags("-mlongcalls \
                         -fno-builtin-memcpy -fno-builtin-memset -fno-builtin-bzero \
                         -fno-builtin-stpcpy -fno-builtin-strncpy \
                         ${CMAKE_C_FLAGS}" UNIQ_CMAKE_C_FLAGS)
set(CMAKE_C_FLAGS "${UNIQ_CMAKE_C_FLAGS}" CACHE STRING "C Compiler Base Flags" FORCE)
remove_duplicated_flags("-mlongcalls \
                         -fno-builtin-memcpy -fno-builtin-memset -fno-builtin-bzero \
                         -fno-builtin-stpcpy -fno-builtin-strncpy \
                         ${CMAKE_CXX_FLAGS}" UNIQ_CMAKE_CXX_FLAGS)
set(CMAKE_CXX_FLAGS "${UNIQ_CMAKE_CXX_FLAGS}" CACHE STRING "C++ Compiler Base Flags" FORCE)
remove_duplicated_flags("-mlongcalls ${CMAKE_ASM_FLAGS}" UNIQ_CMAKE_ASM_FLAGS)
set(CMAKE_ASM_FLAGS "${UNIQ_CMAKE_ASM_FLAGS}" CACHE STRING "ASM Compiler Base Flags" FORCE)
```

Result compile command:
```
/opt/espressif/tools/xtensa-esp-elf/esp-13.2.0_20240530/xtensa-esp-elf/bin/xtensa-esp32s3-elf-g++ \
    -DOPENTELEMETRY_ABI_VERSION_NO=1 \
    -I/home/albert/work/optl/api/include \
    -I/home/albert/work/optl/sdk/include \
    -I/home/albert/work/optl/sdk \
    -I/home/albert/work/optl/ext/include \
    -mlongcalls \
    -fno-builtin-memcpy \
    -fno-builtin-memset \
    -fno-builtin-bzero \
    -fno-builtin-stpcpy \
    -fno-builtin-strncpy \
    -O3 \
    -DNDEBUG \
    -Wno-error=deprecated-declarations \
    -o \
    CMakeFiles/opentelemetry_trace.dir/tracer_context.cc.obj \
    -c \
    /home/albert/work/optl/sdk/src/trace/tracer_context.cc
```
This code compiles perfect with host compiler (which is GCC 13.2 by the way, and this stressed me a lot how can 2 similar compilers with exact same flags produce different output!).

But with ESP-IDF it produces an error:
```
In file included from /home/albert/work/optl/sdk/include/opentelemetry/sdk/common/empty_attributes.h:10,
                 from /home/albert/work/optl/sdk/include/opentelemetry/sdk/trace/recordable.h:11,
                 from /home/albert/work/optl/sdk/include/opentelemetry/sdk/trace/multi_recordable.h:10,
                 from /home/albert/work/optl/sdk/include/opentelemetry/sdk/trace/multi_span_processor.h:11,
                 from /home/albert/work/optl/sdk/src/trace/tracer_context.cc:11:
/home/albert/work/optl/api/include/opentelemetry/common/key_value_iterable_view.h: In instantiation of 'bool opentelemetry::v1::common::KeyValueIterableView<T>::ForEachKeyValue(opentelemetry::v1::nostd::function_ref<bool(opentelemetry::v1::nostd::string_view, absl::otel_v1::variant<bool, long int, long long int, long unsigned int, double, const char*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<const bool, 4294967295>, opentelemetry::v1::nostd::span<const long int, 4294967295>, opentelemetry::v1::nostd::span<const long long int, 4294967295>, opentelemetry::v1::nostd::span<const long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const double, 4294967295>, opentelemetry::v1::nostd::span<const opentelemetry::v1::nostd::string_view, 4294967295>, long long unsigned int, opentelemetry::v1::nostd::span<const long long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const unsigned char, 4294967295> >)>) const [with T = std::array<std::pair<std::__cxx11::basic_string<char>, int>, 0>]':
/home/albert/work/optl/api/include/opentelemetry/common/key_value_iterable_view.h:64:8:   required from here
/home/albert/work/optl/api/include/opentelemetry/common/key_value_iterable_view.h:71:20: error: no match for call to '(opentelemetry::v1::nostd::function_ref<bool(opentelemetry::v1::nostd::string_view, absl::otel_v1::variant<bool, long int, long long int, long unsigned int, double, const char*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<const bool, 4294967295>, opentelemetry::v1::nostd::span<const long int, 4294967295>, opentelemetry::v1::nostd::span<const long long int, 4294967295>, opentelemetry::v1::nostd::span<const long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const double, 4294967295>, opentelemetry::v1::nostd::span<const opentelemetry::v1::nostd::string_view, 4294967295>, long long unsigned int, opentelemetry::v1::nostd::span<const long long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const unsigned char, 4294967295> >)>) (const std::__cxx11::basic_string<char>&, const int&)'
   71 |       if (!callback(iter->first, iter->second))
      |            ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/albert/work/optl/api/include/opentelemetry/common/key_value_iterable.h:7,
                 from /home/albert/work/optl/sdk/include/opentelemetry/sdk/common/attribute_utils.h:15,
                 from /home/albert/work/optl/sdk/include/opentelemetry/sdk/resource/resource.h:8,
                 from /home/albert/work/optl/sdk/src/trace/tracer_context.cc:9:
/home/albert/work/optl/api/include/opentelemetry/nostd/function_ref.h:89:5: note: candidate: 'R opentelemetry::v1::nostd::function_ref<R(Args ...)>::operator()(Args ...) const [with R = bool; Args = {opentelemetry::v1::nostd::string_view, absl::otel_v1::variant<bool, long int, long long int, long unsigned int, double, const char*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<const bool, 4294967295>, opentelemetry::v1::nostd::span<const long int, 4294967295>, opentelemetry::v1::nostd::span<const long long int, 4294967295>, opentelemetry::v1::nostd::span<const long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const double, 4294967295>, opentelemetry::v1::nostd::span<const opentelemetry::v1::nostd::string_view, 4294967295>, long long unsigned int, opentelemetry::v1::nostd::span<const long long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const unsigned char, 4294967295> >}]'
   89 |   R operator()(Args... args) const { return invoker_(callable_, std::forward<Args>(args)...); }
      |     ^~~~~~~~
/home/albert/work/optl/api/include/opentelemetry/nostd/function_ref.h:89:20: note:   no known conversion for argument 2 from 'const int' to 'absl::otel_v1::variant<bool, long int, long long int, long unsigned int, double, const char*, opentelemetry::v1::nostd::string_view, opentelemetry::v1::nostd::span<const bool, 4294967295>, opentelemetry::v1::nostd::span<const long int, 4294967295>, opentelemetry::v1::nostd::span<const long long int, 4294967295>, opentelemetry::v1::nostd::span<const long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const double, 4294967295>, opentelemetry::v1::nostd::span<const opentelemetry::v1::nostd::string_view, 4294967295>, long long unsigned int, opentelemetry::v1::nostd::span<const long long unsigned int, 4294967295>, opentelemetry::v1::nostd::span<const unsigned char, 4294967295> >'
   89 |   R operator()(Args... args) const { return invoker_(callable_, std::forward<Args>(args)...); }
      |                ~~~~^~~~~~~~
```

The cause of this error is that [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/common/empty_attributes.h#L29) you have
```
static const
opentelemetry::common::KeyValueIterableView<
      std::array<std::pair<std::string, int>, 0>>
      kEmptyAttributes(array);
```
and `int` here is crucial.

The thing is that [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/common/attribute_value.h#L37) we have a set of allowed values for Attribute:
```
using AttributeValue =
    nostd::variant<bool,
                   int32_t,
                   int64_t,
                   uint32_t,
                   double,
                   const char *,
                   nostd::string_view,
                   nostd::span<const bool>,
                   nostd::span<const int32_t>,
                   nostd::span<const int64_t>,
                   nostd::span<const uint32_t>,
                   nostd::span<const double>,
                   nostd::span<const nostd::string_view>,
                   // Not currently supported by the specification, but reserved for future use.
                   // Added to provide support for all primitive C++ types.
                   uint64_t,
                   // Not currently supported by the specification, but reserved for future use.
                   // Added to provide support for all primitive C++ types.
                   nostd::span<const uint64_t>,
                   // Not currently supported by the specification, but reserved for future use.
                   // See https://github.com/open-telemetry/opentelemetry-specification/issues/780
                   nostd::span<const uint8_t>>;
```

And it appeals to int32_t.
If we do g++ -E -P we can see the difference in 2 compilers (host and ESP-IDF):
image

And it results in failing detection int resolution for variant in `AttributeValue`.

I think the best solution is to change all random non-defined `int/long int/long long int` values to strict sized types. And I'm saying this not because I'm related to embedded field, but because you specified strict types for `AttributeValue`, and specifying non-strict in other part seems inconsistent.

Possible side-effects:
I don't think it could lead to something unspecified because int on PC-systems is unlikely be different from `int32_t`, but on embedded systems it would be better to specify exact type `int32_t` as it could lead to more predictability.
The worst what can happen: projects compiled on toolchains with int == int64 could have some side-effects.

P.S.
It's not the last PR for bringing up OT on ESP32. This will only fix compile problems. I also have 2 suggestions about:
* possibility to specify NO_OS value for CMake, because there's no AtFork in pthread realization for ESP32, so [fork_unix.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/common/platform/fork_unix.cc#L16) is not suitable.
* there is a slight [typo in cmake.in](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/opentelemetry-cpp-config.cmake.in#L108): if (OR val OR val)

P.P.S.
I also want to add packages for ESP registry to have easy and seamless integration OT into ESP32 projects.

albkharisov:
I don't know how to put this to already created PR, but this can fix this abandoned issue [#2483](https://github.com/open-telemetry/opentelemetry-cpp/issues/2483)

marcalff:
Thanks for all the details, looking.

marcalff:
@albkharisov 

Could you clarify, what is the value of `sizeof(int)` for this platform, and what does INT_MAX looks like in header files (limits.h) ?

`short int` is 16 bits, `long int` is 32, `long long int` is 64, according to the screen shot.

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -257,7 +257,7 @@ class SpanData final : public Recordable
                 opentelemetry::common::SystemTimestamp timestamp =
                     opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now()),
                 const opentelemetry::common::KeyValueIterable &attributes =
-                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int>>(
+                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int32_t>>(
                         {})) noexcept override
   {
     SpanDataEvent event(std::string(name), timestamp, attributes);
```
marcalff, line 263: This is a virtual method, which `override` the parent.

A better fix is to remove the default value entirely (remove `= ...`), as this is taken care of in class Recordable.

Could you check if it resolves the issue for your platform ?

sdk/include/opentelemetry/sdk/common/empty_attributes.h
```diff
@@ -21,12 +21,12 @@ namespace sdk
...
 static const opentelemetry::common::KeyValueIterableView<
-    std::array<std::pair<std::string, int>, 0>> &
+    std::array<std::pair<std::string, int32_t>, 0>> &
 GetEmptyAttributes() noexcept
 {
-  static const std::array<std::pair<std::string, int>, 0> array{};
+  static const std::array<std::pair<std::string, int32_t>, 0> array{};
   static const opentelemetry::common::KeyValueIterableView<
-      std::array<std::pair<std::string, int>, 0>>
+      std::array<std::pair<std::string, int32_t>, 0>>
```
marcalff, line 29: Agreed.

marcalff:
> P.S. It's not the last PR for bringing up OT on ESP32. This will only fix compile problems. I also have 2 suggestions about:
> 
>     * possibility to specify NO_OS value for CMake, because there's no AtFork in pthread realization for ESP32, so [fork_unix.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/common/platform/fork_unix.cc#L16) is not suitable.
> 

Ok, let's see this in a different PR.

>     * there is a slight [typo in cmake.in](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/opentelemetry-cpp-config.cmake.in#L108): if (OR val OR val)
> 

Thanks for reporting it.
Addressed in #3165

> 
> P.P.S. I also want to add packages for ESP registry to have easy and seamless integration OT into ESP32 projects.

Great, thanks for that.

albkharisov:
> @albkharisov
> 
> Could you clarify, what is the value of `sizeof(int)` for this platform, and what does INT_MAX looks like in header files (limits.h) ?
> 
> `short int` is 16 bits, `long int` is 32, `long long int` is 64, according to the screen shot.

@marcalff 

```
I (502) Test: sizeof(short int) : 2
I (503) Test: sizeof(int) : 4
I (503) Test: sizeof(long int) : 4
I (503) Test: sizeof(long long int) : 8
I (504) Test: INT_MAX : 2147483647
I (504) Test: INT_MIN : 2147483648
I (504) Test: UINT_MAX : 4294967295
```
It's 4 bytes, but it doesn't help since it doesn't have an alias for `int32_t`, so for ESP-IDF g++ this std::variant looks like set of (long int, long long int, ...), without `int`.

sdk/include/opentelemetry/sdk/trace/span_data.h
albkharisov: Fixed, pushed. Yes this helps on my platform.
But I had to remove 2 default params, since second can't be not-defaulted, while first has default value.

albkharisov:
It fails to compile test with this.
Error is here: [span_data_test.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/test/trace/span_data_test.cc#L72)

@marcalff 
What's best fix?
* Fix test
* Revert my last commit

marcalff:
> It fails to compile test with this. Error is here: [span_data_test.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/test/trace/span_data_test.cc#L72)
> 

Yes, saw that.

This exposed a different issue: the existing header files are not so clean when it comes to overload resolution, with helpers conflicting with override methods using default values.

> @marcalff What's best fix?
> 
>     * Fix test
> 
>     * Revert my last commit
> 
> 
> UPD: fixed test

Changing the test is only a work around, because the original test code is supposed to work.

I am reverting the change back to your original patch, as it is good enough.

The various overrides need some cleanup, but I prefer to address that in a separate PR, and not delay this one because of unrelated issues.

sdk/test/trace/span_data_test.cc
```diff
@@ -69,7 +69,8 @@ TEST(SpanData, Set)
   data.SetStartTime(now);
   data.SetDuration(std::chrono::nanoseconds(1000000));
   data.SetAttribute("attr1", static_cast<int64_t>(314159));
-  data.AddEvent("event1", now);
+  data.AddEvent("event1", now,
+                opentelemetry::common::KeyValueIterableView<std::map<std::string, int32_t>>({}));
```
marcalff, line 73:
```suggestion
  data.AddEvent("event1", now);
```

Reverting to original commit

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -254,11 +254,8 @@ class SpanData final : public Recordable
   }
 
   void AddEvent(nostd::string_view name,
-                opentelemetry::common::SystemTimestamp timestamp =
-                    opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now()),
-                const opentelemetry::common::KeyValueIterable &attributes =
-                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int>>(
-                        {})) noexcept override
+                opentelemetry::common::SystemTimestamp timestamp,
+                const opentelemetry::common::KeyValueIterable &attributes) noexcept override
```
marcalff, line 258:
```suggestion
                opentelemetry::common::SystemTimestamp timestamp =
                    opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now()),
                const opentelemetry::common::KeyValueIterable &attributes =
                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int32_t>>) noexcept override
```

Reverting to original commit

sdk/include/opentelemetry/sdk/trace/span_data.h
```diff
@@ -257,8 +257,7 @@ class SpanData final : public Recordable
                 opentelemetry::common::SystemTimestamp timestamp =
                     opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now()),
                 const opentelemetry::common::KeyValueIterable &attributes =
-                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int>>(
-                        {})) noexcept override
+                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int32_t>>) noexcept override
```
marcalff, line 260:
```suggestion
                    opentelemetry::common::KeyValueIterableView<std::map<std::string, int32_t>>(
                        {})) noexcept override
```

PR #3106, opened 2024-10-17
[TEST] Fix linking order and gmock linking

owent:
Fixes #3102

## Changes

+ Change the link order of gtest

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #3158, opened 2024-11-22
[BUILD] Visual Studio GMOCK_LIB issues and explicit c++17 

bkuschel:
- Change in PR #3106 broke MSVC build. GMOCK_LIB-NOTFOUND from cmake. 
- In some cases (unsure why), c++17 has to be set explicitly for MSVC

## Changes

- Copy over similar working code into cmake files dealing with GMOCK_LIB
- Explicitly add c++17 support for MSVC


CMakeLists.txt
```diff
@@ -94,7 +94,10 @@ else()
   endif()
 endif()
 message(STATUS "Building for architecture ARCH=${ARCH}")
-
+# Force C++17
+if(MSVC)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++17")
```
marcalff, line 99: We can not have that, this will break everybody.

marcalff (changes requested):
Thanks for the patch.

This needs more investigations, to understand the root cause.

CMakeLists.txt
bkuschel: I'll try to find an alternative

bkuschel: @marcalff The only other way I can get it to compile with vs2022 is to add "-DWITH_STL=CXX14"
Other wise i get

```
warning STL4038: The contents of <string_view> are available only with C++17 or later.
C:\Users\Administrator\Git\opentelemetry-cpp\api\include\opentelemetry/std/string_view.h(18): error C2039: 'string_view': is not a member of 'std'
```

This is during `stl` build, `nostd` works without this (of course).

owent: You can use `-DCMAKE_CXX_STANDARD=17` when running cmake command to use C++17 only.

owent:
This problem should already be fixed by #3203 , could you please try the main branch?

marcalff:
> This problem should already be fixed by #3203 , could you please try the main branch?

As indicated earlier, the issues related to GMOCK_LIB have been resolved already.

Now closing this PR, since it is not needed.

If somehow the latest code in opentelemetry-cpp still does not work, please file an issue with the current findings.

PR #3164, opened 2024-11-22
[EXPORTER] Refactor `ElasticSearchRecordable`

sjinks:
## Changes

* Simplify `ElasticSearchRecordable::WriteValue()` by using the [ADL Serializer](https://github.com/nlohmann/json?tab=readme-ov-file#how-do-i-convert-third-party-types);
* Simplify `ElasticSearchRecordable::SetTimestamp()`;
* Refactor `ElasticSearchRecordable::SetTraceId()` and `ElasticSearchRecordable::SetSpanId()` by using constants for array sizes instead of hardcoded values.

* [x] Unit tests have been added (in another PR)


exporters/elasticsearch/src/es_log_recordable.cc
```diff
@@ -9,6 +11,27 @@
...
 #include "opentelemetry/trace/trace_id.h"
 
+namespace nlohmann
+{
+template <>
+struct adl_serializer<opentelemetry::sdk::common::OwnedAttributeValue>
+{
+  static void to_json(json &j, const opentelemetry::sdk::common::OwnedAttributeValue &v)
+  {
+    opentelemetry::nostd::visit([&j](const auto &value) { j = value; }, v);
```
sjinks, line 21: A possible breaking change here is that the original `WriteValue()` did not handle spans. If this is important, I can add an `if` condition to follow the old behavior.

exporters/elasticsearch/src/es_log_recordable.cc
```diff
@@ -180,28 +70,21 @@ void ElasticSearchRecordable::SetTimestamp(
...
-  std::strftime(bufferDate, sizeof(bufferDate), "%Y-%m-%dT%H:%M:%S", &tm);
   auto microseconds =
       std::chrono::duration_cast<std::chrono::microseconds>(timePoint.time_since_epoch()) %
       std::chrono::seconds(1);
 
-  char bufferMilliseconds[millisecondsSize];
-  std::snprintf(bufferMilliseconds, sizeof(bufferMilliseconds), ".%06ld",
+  char dateStr[dateSize + 1];  // example: 2024-10-18T07:26:00.123456Z
+  std::snprintf(dateStr, sizeof(dateStr), "%04d-%02d-%02dT%02d:%02d:%02d.%06ldZ", tm.tm_year + 1900,
+                tm.tm_mon + 1, tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec,
```
sjinks, line 86: We can format the date in one go; there's no need to do unnecessary concatenations.

Also, because `nlohmann::json` handles `const char*`, there is no need to construct an `std::string`.

exporters/elasticsearch/src/es_log_recordable.cc
```diff
@@ -180,28 +76,17 @@ void ElasticSearchRecordable::SetTimestamp(
...
       std::chrono::duration_cast<std::chrono::microseconds>(timePoint.time_since_epoch()) %
       std::chrono::seconds(1);
 
-  char bufferMilliseconds[millisecondsSize];
-  std::snprintf(bufferMilliseconds, sizeof(bufferMilliseconds), ".%06ld",
-                static_cast<long>(microseconds.count()));
-
-  std::strcat(bufferDate, bufferMilliseconds);
-  std::strcat(bufferDate, "Z");
+  std::ostringstream oss;
```
owent, line 85: Why change it to `put_time` and `ostringstream`? In my understandarding, using `strftime` and `snprintf` with just one static buffer block will have better performence than using `ostringstream` and `put_time` with dynamic memory allocation. And there are also some compatiblity problems with `put_time` and old compiler's STL.

sjinks: A newer version of gcc complained:

```
warning: ‘__builtin___snprintf_chk’ output may be truncated before the last format character [-Wformat-truncation=]
   47 |     std::snprintf(bufferMilliseconds, sizeof(bufferMilliseconds), ".%06ld",
      |                                                                          ^
In function ‘int snprintf(char*, size_t, const char*, ...)’,
/usr/include/x86_64-linux-gnu/bits/stdio2.h:54:35: note: ‘__builtin___snprintf_chk’ output between 8 and 9 bytes into a destination of size 8
   54 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   55 |                                    __glibc_objsize (__s), __fmt,
      |                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   56 |                                    __va_arg_pack ());
      |                                    ~~~~~~~~~~~~~~~~~
```

so I tried to come up with something safe that does not involve hardcoded sizes, compiles cleanly, does not have a potential for buffer overflows, and is as fast as `std::format("{:%FT%T%Ez}", timePoint)`.

Anyway, I think I have found a faster solution not involving `strcat` and hardcoded buffer sizes.

owent: This can be fixed by declaring `bufferMilliseconds` as `char bufferMilliseconds[millisecondsSize + 1];` . And change the snprinf calling to `std::snprintf(bufferMilliseconds, sizeof(bufferMilliseconds) - 1, ".%06ld", static_cast<long>(microseconds.count()));` . Thera are a little different between the bebaviour of snprinf in MSVC and GCC.

owent (approved):
LGTM and thanks.

marcalff (approved):
LGTM, thanks for the fix.

exporters/elasticsearch/src/es_log_recordable.cc
ShadowMaxLeb: I just bumped into your changes and I was wondering if this could help putting json objects within the root json ?
The modifications I did was kind of a workaround so we could add keys like `user.id` that would be visible within ELK like the documentation asks for but the key was still `user.id` and not an object `user` under `root` that would have a key `id`

My hope would be to be able to be able to create an attribute or resource that contains a json object for some keys:

```
auto attr = { "user":  R"({"id": 1234}"_json};
```

PR #3167, opened 2024-11-25
[BUILD] Upgrade cmake

### LINKED ISSUE #3116: Clarifications needed around WITH_ABSEIL
How to use opentelemetry-cpp with or without ABSEIL is unclear.

This is because:

* using the opentelemetry-cpp ABI with abseil
* using an exporter that indirectly depends on abseil (namely, the OTLP GRPC exporter)
* using abseil in an application that happens to use opentelemetry-cpp header files

all collide into the same flag, `WITH_ABSEIL`, while each topic is a separate concern.

We need to clarify what is supported, what is not supported, and how to build cleanly a coherent binary while separating each concern.

### LINKED ISSUE #3166: [CMake] Upgrade cmake_minimum_required to 3.16
The new cmake (3.30+) will produce a warning like this:

```bash
CMake Deprecation Warning at CMakeLists.txt:4 (cmake_minimum_required):Compatibility with CMake < 3.10 will be removed from a future version of
CMake.

Update the VERSION argument <min> value or use a ...<max> suffix to tell
CMake that the project does not need compatibility with older versions.CMake (cmake_minimum_required)
```

Since cmake 3.16, there have been a lot of changes of CONFIG package names in FindXXX scripts, and a lot of policies changes.
I just want to discuss if we can upgrade the minimum version to cmake to 3.16 or upper.

We use `cmake_minimum_required(VERSION 3.24)` and `cmake_minimum_required(VERSION 3.27)` to avoid writing too much compact codes to keep the same behiour in old and new cmake. It's OK for me to upgrade the minimal requirement to 3.24 or 3.27, but I think we can discuss if it will bring difficulty to other cases and if it's worth doing the upgrade.

owent:
Fixes #3166

## Changes

+ Upgrade the minimal requirement of cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,21 +1,36 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.10)
 
-# See https://cmake.org/cmake/help/v3.3/policy/CMP0057.html required by certain
-# versions of gtest
+# These policies belowed are used to keep a more strick behavior of cmake
+cmake_policy(SET CMP0054 NEW)
+cmake_policy(SET CMP0067 NEW)
+
+# See https://cmake.org/cmake/help/latest/policy/CMP0057.html required by
+# certain versions of gtest
 cmake_policy(SET CMP0057 NEW)
 
-# See https://cmake.org/cmake/help/v3.12/policy/CMP0074.html required by certain
-# version of zlib which CURL depends on.
+# See https://cmake.org/cmake/help/latest/policy/CMP0074.html required by
+# certain version of zlib which CURL depends on.
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
   cmake_policy(SET CMP0074 NEW)
+endif()
+
+# Allow to use normal variable for option()
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13")
+  cmake_policy(SET CMP0077 NEW)
 endif()
 
 # Prefer CMAKE_MSVC_RUNTIME_LIBRARY if possible
 if(POLICY CMP0091)
   cmake_policy(SET CMP0091 NEW)
+endif()
+
+# MSVC RTTI flag /GR should not be not added to CMAKE_CXX_FLAGS by default. @see
+# https://cmake.org/cmake/help/latest/policy/CMP0117.html
+if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.20.0")
+  cmake_policy(SET CMP0117 NEW)
 endif()
 
 project(opentelemetry-cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,14 +3,6 @@
 
 cmake_minimum_required(VERSION 3.10)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.9)
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30")
+  cmake_minimum_required(VERSION 3.10)
+else()
+  cmake_minimum_required(VERSION 3.9)
+endif()
```
marcalff, line 8:
```suggestion
cmake_minimum_required(VERSION 3.10)
```

Could we make this not conditional, and just require 3.10 ?

marcalff: Also, the policy below:

```
cmake_policy(SET CMP0057 NEW)
```

is outdated in my understanding, because it will always to NEW with >= 3.3

Should that part be removed ?

owent: According to <https://cmake.org/cmake/help/latest/policy/CMP0057.html>, cmake >= 3.3 just add this policy but the default behavior is OLD.So I think `cmake_policy(SET CMP0057 NEW)` is still needed here.

owent: I also add some policy settings to use cmake a more modern way.Could you please review it to check if it's OK?

marcalff: There is something I do not understand here.

CMake policy CMP0054 was added in 3.1

The doc https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required for CMAKE_MINIMUM_REQUIRED says:

> All policies known to the running version of CMake and introduced in the min (or max, if specified) version or earlier will be set to use NEW behavior.

So, `cmake_minimum_required(VERSION 3.10)` should set policy CMP0054 to NEW, because that policy was added in 3.1, and 3.1 <= 3.10.

Why the need to set CMP0057 explicitly ?

Likewise for CMP0057 (3.3), CMP0067 (3.8)

CMakeLists.txt
```diff
@@ -1,23 +1,38 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.10)
 
-# See https://cmake.org/cmake/help/v3.3/policy/CMP0057.html required by certain
-# versions of gtest
+# These policies belowed are used to keep a more strick behavior of cmake
```
lalitb, line 6:
```suggestion
# The below policies are used to keep a more strict behavior of cmake
```

CMakeLists.txt
```diff
@@ -1,23 +1,38 @@
...
+cmake_policy(SET CMP0054 NEW)
```
lalitb, line 7: Can we add comment on what these policies do, though I understand some existing policies are missing the comment :)

lalitb (approved):

CMakeLists.txt
owent: Thanks, I missed this message before. These policies are cleaned now.

CMakeLists.txt
owent: These old policies are removed now.

marcalff (approved):
LGTM, thanks for the fix.

PR #3171, opened 2024-11-26
[PROMETHEUS_EXPORTER] Fix default for emitting otel_scope attributes

### LINKED ISSUE #3170: Prometheus Exporter: otel_scope attributes not present by default
I noticed that https://github.com/open-telemetry/opentelemetry-cpp/pull/2479 changed the default from `otel_scope` attributes being populated to being off by default. From my reading of the code, it seems like this was an oversight.

yashykt:
Fixes #3170

## Changes

Fixes the default behavior of emitting `otel_scope` attributes.

### DIFF
--- a/exporters/prometheus/src/exporter_options.cc
+++ b/exporters/prometheus/src/exporter_options.cc
@@ -33,7 +33,7 @@ inline bool GetPrometheusWithoutOtelScope()
   auto exists =
       opentelemetry::sdk::common::GetBoolEnvironmentVariable(kPrometheusWithoutOtelScope, setting);
 
-  return exists ? setting : true;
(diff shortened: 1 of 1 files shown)


marcalff:
From the spec:

> A Prometheus Exporter MAY support a configuration option to produce metrics without a [scope info](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1) metric, or scope labels. The option MAY be named without_scope_info, and MUST be false by default.

so this PR is indeed the expected behavior.

marcalff (approved):
LGTM, thanks for the fix.

PR #3175, opened 2024-11-26
[POC] Better control of threads executed by opentelemetry-cpp

### LINKED ISSUE #3174: [Feat] Better control of threads executed by opentelemetry-cpp
## Context

The opentelemetry-cpp uses threads internally, for various components.

For example:

* the batch processor uses a thread to process records from the span buffer.
* the periodic metric reader uses a thread to trigger periodic collection, and a child thread to collect observable metrics.
* the log processor uses a thread to process records from a log buffer.
* every exporter based on CURL uses a thread internally, to process HTTP requests and responses.

## Problem

An application that links with the opentelemetry-cpp library may have additional constraints, and may want to control at a very low level how the code executes within a thread.

For example:

* the application may want to name threads, to be visible in the operating system under an application controlled name.
* the application may need to initialize thread local storage, in particular when custom code is executed within a thread spawned by opentelemetry. Typically, callbacks for observable metrics may need TLS keys in place to execute properly.
* the application may need to control thread priorities, or bind internal opentelemetry threads to some CPUs.
* the application may need to control the current network namespace, when a thread opens a network connection.

Currently, there is no way to implement any of this from an application that uses opentelemetry-cpp.

marcalff:
Fixes #3174

## Changes

### DIFF
--- a/ext/src/http/client/curl/http_client_curl.cc
+++ b/ext/src/http/client/curl/http_client_curl.cc
@@ -189,7 +189,18 @@ HttpClient::HttpClient()
       next_session_id_{0},
       max_sessions_per_connection_{8},
       scheduled_delay_milliseconds_{std::chrono::milliseconds(256)},
-      curl_global_initializer_(HttpCurlGlobalInitializer::GetInstance())
+      curl_global_initializer_(HttpCurlGlobalInitializer::GetInstance()),
+      background_thread_instrumentation_(nullptr)
+{}
+
+HttpClient::HttpClient(
+    const std::shared_ptr<sdk::common::ThreadInstrumentation> &thread_instrumentation)
+    : multi_handle_(curl_multi_init()),
+      next_session_id_{0},
+      max_sessions_per_connection_{8},
+      scheduled_delay_milliseconds_{std::chrono::milliseconds(256)},
+      curl_global_initializer_(HttpCurlGlobalInitializer::GetInstance()),
+      background_thread_instrumentation_(thread_instrumentation)
 {}
 
 HttpClient::~HttpClient()
@@ -345,6 +356,11 @@ void HttpClient::MaybeSpawnBackgroundThread()
 
   background_thread_.reset(new std::thread(
       [](HttpClient *self) {
+        if (self->background_thread_instrumentation_ != nullptr)
+        {
+          self->background_thread_instrumentation_->OnStart();
+        }
+
         int still_running = 1;
         while (true)
         {
@@ -451,6 +467,11 @@ void HttpClient::MaybeSpawnBackgroundThread()
               break;
             }
           }
+        }
+
+        if (self->background_thread_instrumentation_ != nullptr)
+        {
+          self->background_thread_instrumentation_->OnEnd();
         }
       },
       this));
--- a/sdk/include/opentelemetry/sdk/common/thread_instrumentation.h
+++ b/sdk/include/opentelemetry/sdk/common/thread_instrumentation.h
@@ -1,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace common
+{
+
+class ThreadInstrumentation
+{
+public:
+  ThreadInstrumentation()          = default;
+  virtual ~ThreadInstrumentation() = default;
+
+  virtual void OnStart()    = 0;
+  virtual void OnEnd()      = 0;
+  virtual void BeforeWait() = 0;
+  virtual void AfterWait()  = 0;
+  virtual void BeforeLoad() = 0;
+  virtual void AfterLoad()  = 0;
+};
+
+}  // namespace common
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
@@ -20,6 +20,7 @@
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/nostd/variant.h"
 #include "opentelemetry/sdk/common/exporter_utils.h"
+#include "opentelemetry/sdk/common/thread_instrumentation.h"
 #include "opentelemetry/version.h"
 
 // forward declare google::protobuf::Message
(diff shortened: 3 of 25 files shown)


owent:
Do you think we can implement a worker pool to share threads among multiple exporters and processors? Since most of them simply wait for a timeout, there is no need to create a separate thread for each component.

marcalff:
> Do you think we can implement a worker pool to share threads among multiple exporters and processors? Since most of them simply wait for a timeout, there is no need to create a separate thread for each component.

For some applications, running in a multi threaded environment, it is critical to control how a thread executes in finer details.

For example, I have:

* (a) an OTLP HTTP trace exporter
* (b) several OTLP HTTP metric exporters
* (c) an OTLP HTTP log exporter

When the exporter uses CURL to post to the endpoint, it makes a TCP/IP connection.

(a), (b) and (c) do need to use different TCP/IP stacks, or different named networks, to connect to their respective endpoints.

This is achieved by calling `setns(fd, CLONE_NEWNET)` in each thread, where fd is a file descriptor describing the named network to use.

Having dedicated threads (as currently) makes this easier.

If execution of each exporter was to be multiplexed and executed in the same worker thread, the context will need to change for each exporter, calling setns() many more time, and introducing more complexity to counter the effect of a pool worker thread, so it won't help but instead get in the way.

setns() is just an example, there can be other use cases, like binding to a CPU, etc.

Overall, the need is for the main application to inject arbitrary code in the opentelemetry-cpp execution code path, for opentelemetry threads.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -428,6 +441,11 @@ bool HttpClient::MaybeSpawnBackgroundThread()
...
   background_thread_.reset(new std::thread(
       [](HttpClient *self) {
+        if (self->background_thread_instrumentation_ != nullptr)
+        {
+          self->background_thread_instrumentation_->OnStart();
+        }
+
         int still_running = 1;
         std::chrono::system_clock::time_point last_free_job_timepoint =
             std::chrono::system_clock::now();
```
owent, line 451: Do we need to call `BeforeWait` and `AfterWait` before and after `curl_multi_poll` or `curl_multi_wait`? These calls may block until an event occurs or a timeout is reached.

owent:
Could please also add this instrumentation for OTLP file exporters?

ext/src/http/client/curl/http_client_curl.cc
marcalff: Yes, good point. Implemented.

marcalff:
> Could please also add this instrumentation for OTLP file exporters?

Sure, implemented.

Every place where opentelemetry-cpp starts a new `std::thread` is now instrumented.

marcalff:
An example of using this feature, to set thread names to the operating system on Linux:

```
ps -Lp 199922 -o pid,tid,comm
```

```
    PID     TID COMMAND
 199922  199922 (redacted)
 ...
 199922  201152 otel_bsp-1
 199922  201153 otel_pmr-1
 199922  201154 otel_pmr-2
 199922  201157 otel_blrp-1
 199922  201163 otlp_traces
 199922  201165 otlp_logs
 199922  201179 otlp_metrics
```

In order, threads are:

* the batch span processor
* the periodic exporting meter reader (2 instances)
* the batch log record processor
* a trace otlp http curl thread
* a logs otlp http curl thread
* a metrics otlp http curl thread

owent (approved):
LGTM and thanks.

sdk/include/opentelemetry/sdk/common/thread_instrumentation.h
```diff
@@ -0,0 +1,92 @@
...
+/*
+ * This feature is experimental, protected by a _PREVIEW flag.
+ */
+#ifdef ENABLE_THREAD_INSTRUMENTATION_PREVIEW
+  virtual void OnStart() {}
+  virtual void OnEnd() {}
+  virtual void BeforeWait() {}
+  virtual void AfterWait() {}
+  virtual void BeforeLoad() {}
+  virtual void AfterLoad() {}
```
lalitb, line 86: Wondering if there's a way to avoid virtual calls for these methods to reduce their associated cost. One potential approach is to allow users to provide a callback that the SDK invokes during thread lifecycle events. For example:

```cpp
    class ThreadInstrumentation {
        ThreadInstrumentation( std::function<> callback = DefaultCallback)
            : callback_(std::move(callback)) {}
        bool OnOperation(ThreadOperation opr, ThreadContext &context) {
             return callback_(operation, context);
        }
   
        private:
        std::function<bool(ThreadOperation, const ThreadContext&)> callback_;
    }
```
`ThreadContext` could carry information like `thread_id` or be omitted if unnecessary. This approach lets the SDK call `OnOperation` internally while allowing the callback to define behavior for specific stages. 

Not a blocker as I don't see these virtual methods to be called in user hot-path, but if this is something we need to consider?

marcalff: Most of the time, the instrumentation will be absent (nullptr), so that:

```
if (worker_thread_instrumentation_ != nullptr)
{
  worker_thread_instrumentation->OnXXX();
}
```

will not even enter the if block, so it is definitively not in the hot path.

When some instrumentation is provided, the virtual function will land in user code, or in the noop default implementation.

I expect that using a virtual function or a callback std::function to be equivalent, in term of overhead: the call will be dynamic (using a function pointer), and not static with the target known at compile time. A virtual method versus a callback std::function versus plain old C function pointers is just a different coding style, they all boil down to the same thing (a function pointer).

marcalff: About passing thread context:

The subclass of ThreadInstrumentation is free to hold state, and can even keep per thread instance state in thread local storage if it really wants to, for complex cases.

This pushes complexity to the user (and disclosure, this is how I use it).

For calls that are paired, we could help a bit to propagate state like this:

```
void *opaque_wait_context = nullptr;

if (worker_thread_instrumentation_ != nullptr)
{
  opaque_wait_context = worker_thread_instrumentation->BeforeWait();
}

// do something

if (worker_thread_instrumentation_ != nullptr)
{
  worker_thread_instrumentation->AfterWait(opaque_wait_context);
}
```

and likewise for OnStart/OnEnd, and BeforeLoad/AfterLoad.

With this opaque pointer, the user code is free to use it (or not), and it avoids complexity with thread local storage.

@lalitb 

I can implement this if you agree with it, or we can revise it later, let me know what you prefer.

lalitb: > I expect that using a virtual function or a callback std::function to be equivalent, in term of overhead: the call will be dynamic (using a function pointer), and not static with the target known at compile time. A virtual method versus a callback std::function versus plain old C function pointers is just a different coding style, they all boil down to the same thing (a function pointer).

true. 

> I can implement this if you agree with it, or we can revise it later, let me know what you prefer.

I think we can revisit it later.

examples/otlp/http_instrumented_main.cc
```diff
@@ -0,0 +1,357 @@
...
+  std::string name{"otlp_http_metric_example"};
+  foo_library::observable_counter_example(name);
+
+  std::cout << "Shutting down opentelemetry-cpp" << std::endl << std::flush;
+
+  CleanupLogger();
+  CleanupMetrics();
+  CleanupTracer();
+
+  std::cout << "Done" << std::endl << std::flush;
```
marcalff, line 356:
```suggestion
  std::cout << "Done" << std::endl << std::flush;
  return 0;
```

PR #3178, opened 2024-12-02
[EXPORTER] Optimize OTLP HTTP compression

### LINKED ISSUE #2570: [EXPORTER] optimize OTLP HTTP compression
Investigate and/or implement possible performance improvements.

Related to this idea from #2530

> Do you think we can use the in-place compression of the data without using a separate output buffer compressed_body, as suggested here - https://stackoverflow.com/questions/12398377/is-it-possible-to-have-zlib-read-from-and-write-to-the-same-memory-buffer/12412863#12412863. This is suggested by "Mark Adler", the author of zlib, and would be better memory optimzation in the hot-path of upload.

chusitoo:
Fixes #2570

## Changes

- Added the in-place version of deflate suggested by zlib author. 
- Reworked some parts to resolve warnings and make it more C++ like.
- Minor refactoring to the logic in SendRequest to catch and report the error via the callback.
- Removed some redundant temporary string constructors where appropriate.

Another alternative to in-place compression was also considered, which consists in statically allocating the vector that would be used as output to deflate (i.e. `compressed_body` in the original code) and swapping it with the original body after deflate was done, essentially, reusing the original body vector for the subsequent call to `SendRequest`.

```bash
------------------------------------------------------
Benchmark            Time             CPU   Iterations
------------------------------------------------------
BM_Original      32782 ns        32782 ns        21061
BM_WithSwap      16769 ns        16769 ns        40696
BM_InPlace       16960 ns        16960 ns        42027
```

The very minimal benchmark showed that the performance was not that far off compared to the in-place version from this PR, but it would obviously fall short in more realistic scenarios, since it will likely require resizing this scratch vector every now and then if the input body being processed is larger.


ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -76,44 +149,47 @@ void Session::SendRequest(
...
-    zs.next_out  = compressed_body.data();
+    z_stream zs{};
+    zs.zalloc = Z_NULL;
+    zs.zfree  = Z_NULL;
+    zs.opaque = Z_NULL;
 
     // ZLIB: Have to maually specify 16 bits for the Gzip headers
-    const int window_bits = 15 + 16;
+    static constexpr int kWindowBits = MAX_WBITS + 16;
+    static constexpr int kMemLevel   = MAX_MEM_LEVEL;
```
chusitoo, line 159: Here, MAX_MEM_LEVEL appears to be 9 and, from the official documentation

> The memLevel parameter specifies how much memory should be allocated for the internal compression state. memLevel=1 uses minimum memory but is slow and reduces compression ratio; memLevel=9 uses maximum memory for optimal speed. The default value is 8. 

The default level is not exposed unfortunately (header zutil.h no available for including) but the code to drive the default is:

```cpp
#if MAX_MEM_LEVEL >= 8
#  define DEF_MEM_LEVEL 8
#else
#  define DEF_MEM_LEVEL  MAX_MEM_LEVEL
#endif
/* default memLevel */
```

My intuition was to not hardcode the "default" value 8 and use the max level setting provided by the macro but I can revert this back to a hardcoded value of 8 as in the current compression code.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -76,44 +149,47 @@ void Session::SendRequest(
...
     }
-    else
+
+    if (stream != Z_OK)
     {
       if (callback)
       {
-        callback->OnEvent(opentelemetry::ext::http::client::SessionState::CreateFailed, "");
+        callback->OnEvent(opentelemetry::ext::http::client::SessionState::CreateFailed,
+                          zs.msg ? zs.msg : "");
```
chusitoo, line 182: This is just a precaution; the documentation states that `msg` is null when there are no errors. Given that we check for an error return code before even attempting to use this variable, the ternary if is probably redundant, but it does not hurt keeping it as another safety net in case the invariant is broken by the custom deflateInPlace code or future versions of the library, for instance.

The list of possible messages to be reported:

```cpp
z_const char * const z_errmsg[10] = {
    (z_const char *)"need dictionary",     /* Z_NEED_DICT       2  */
    (z_const char *)"stream end",          /* Z_STREAM_END      1  */
    (z_const char *)"",                    /* Z_OK              0  */
    (z_const char *)"file error",          /* Z_ERRNO         (-1) */
    (z_const char *)"stream error",        /* Z_STREAM_ERROR  (-2) */
    (z_const char *)"data error",          /* Z_DATA_ERROR    (-3) */
    (z_const char *)"insufficient memory", /* Z_MEM_ERROR     (-4) */
    (z_const char *)"buffer error",        /* Z_BUF_ERROR     (-5) */
    (z_const char *)"incompatible version",/* Z_VERSION_ERROR (-6) */
    (z_const char *)""
};
```

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -76,44 +149,47 @@ void Session::SendRequest(
...
 
     if (stream == Z_OK)
     {
-      deflate(&zs, Z_FINISH);
-      deflateEnd(&zs);
-      compressed_body.resize(zs.total_out);
-      http_request_->SetBody(compressed_body);
+      auto size = static_cast<uInt>(http_request_->body_.size());
+      auto max  = size;
+      stream    = deflateInPlace(&zs, http_request_->body_.data(), size, &max);
```
chusitoo, line 168: Here, `max == size` which means that, if I interpret the comments from the library author correctly, when there is a lot of incompressible data, an auxiliary buffer is used to complete the deflate operation so that it can report the current max value and bail out with an error code. My understanding is that this would require calling `deflateInPlace` with an input buffer that has been resized to accommodate for the expected output, which would be larger than the original payload.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -896,7 +896,7 @@ OtlpHttpClient::createSession(
   // Parse uri and store it to cache
   if (http_uri_.empty())
   {
-    auto parse_url = opentelemetry::ext::http::common::UrlParser(std::string(options_.url));
+    const auto &parse_url = opentelemetry::ext::http::common::UrlParser(options_.url);
```
owent, line 899: Why change it to `const auto &`? In my understanding, `const auto &` can not refer to a temporary object here, because it's may be de destroyed after this sentense.
`T f = T(..);` will construct T only once and will not call copy/move constructor. So I think it's useless here.

owent (changes requested):

exporters/otlp/src/otlp_http_client.cc
chusitoo: I made it _const_ because it is never modified later in function, and _reference_ as a future assurance that it doesn't create a copy, since copy elision is not guaranteed prior to C++17.

I believe it is not a problem with temporary in this case because the [lifetime extension](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary) would kick in to bind the prvalue to the scope of the function it is being used in. 

Although, I agree, it would've been a problem if we were to pass around the pointer to its `.data()` once the function scope ends. 

Seeing this could cause some confusion, I removed the reference to make the intent clearer.

owent: Thanks, my mistake

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -57,11 +60,85 @@ nostd::shared_ptr<HttpCurlGlobalInitializer> HttpCurlGlobalInitializer::GetInsta
...
+  // kick start the process with a temporary output buffer -- this allows deflate to consume a large
+  // chunk of input data in order to make room for output data there
+  strm->next_in  = buf;
+  strm->avail_in = len;
+  if (*max < len)
+  {
+    *max = len;
+  }
+  strm->next_out  = temp.data();
+  strm->avail_out = std::min(static_cast<decltype(z_stream::avail_out)>(temp.size()), *max);
```
owent, line 83:
```cpp
#ifdef min
#undef min
#endif
```

Should we add `#undef min`? Windows SDK may define `min` and will cause conflict sometimes.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -57,11 +60,85 @@ nostd::shared_ptr<HttpCurlGlobalInitializer> HttpCurlGlobalInitializer::GetInsta
...
+  auto ret        = deflate(strm, Z_FINISH);
+  if (ret == Z_STREAM_ERROR)
+  {
+    return ret;
+  }
+
+  // if we can, copy the temporary output data to the consumed portion of the input buffer, and then
+  // continue to write up to the start of the consumed input for as long as possible
+  auto have = strm->next_out - temp.data();  // number of bytes in temp[]
+  if (have <= (strm->avail_in ? len - strm->avail_in : *max))
```
owent, line 93: In my understandarding, compare a signed interger with a  unsigned integer will cause a warning. No idea why this warning does not be triggered.

ext/src/http/client/curl/http_client_curl.cc
marcalff: Given that min is a macro on windows, a work around is `(std::min)(a, b)`, like we do currently for max.

ext/src/http/client/curl/http_client_curl.cc
marcalff: For `ENABLE_OTLP_COMPRESSION_PREVIEW` to work, add `-DWITH_OTLP_HTTP_COMPRESSION=ON` to CMake in the maintainer builds.

ext/src/http/client/curl/http_client_curl.cc
chusitoo: Thanks @marcalff, I used this approach. @owent let me know if this is acceptable.

ext/src/http/client/curl/http_client_curl.cc
chusitoo: I've not seen compile warnings related to this but. indeed, the type of *have* here is `ptrdiff_t`. At first glance, it does not seem possible that the logic yield a negative number so it could but made unsigned but, given that L96 uses it to advance through the buffer, I think it is more reasonable to keep its intended type and to cast the right hand side of condition in L93 instead.

ext/src/http/client/curl/http_client_curl.cc
owent: Yes, but the variable `max` may still have problem. Or we can rename it?

marcalff: max is a macro in windows, so only the syntax max(...), with parenthesis, will cause breaks.

I think this should be fine.

chusitoo: I went ahead and renamed it to avoid any surprises

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -57,11 +60,85 @@ nostd::shared_ptr<HttpCurlGlobalInitializer> HttpCurlGlobalInitializer::GetInsta
...
+// Original source:
+// https://stackoverflow.com/questions/12398377/is-it-possible-to-have-zlib-read-from-and-write-to-the-same-memory-buffer/12412863#12412863
+int deflateInPlace(z_stream *strm, unsigned char *buf, uint32_t len, uint32_t *max_len)
+{
+  // must be large enough to hold zlib or gzip header (if any) and one more byte -- 11 works for the
+  // worst case here, but if gzip encoding is used and a deflateSetHeader() call is inserted in this
+  // code after the deflateReset(), then the 11 needs to be increased to accommodate the resulting
+  // gzip header size plus one
+  std::array<unsigned char, 11> temp{};
+
```
marcalff, line 73: The original code from Mark Adler has:

```
    /* initialize deflate stream and point to the input data */
    ret = deflateReset(strm);
    if (ret != Z_OK)
        return ret;
```

was it forgotten, or is this not needed and why ?

chusitoo: From the [zlib manual](https://www.zlib.net/manual.html#Advanced):

> ZEXTERN int ZEXPORT deflateReset(z_streamp strm);
>
> This function is equivalent to deflateEnd followed by deflateInit, but does not free and reallocate the internal compression state. The stream will leave the compression level and any other attributes that may have been set unchanged. total_in, total_out, adler, and msg are initialized.
>
> deflateReset returns Z_OK if success, or Z_STREAM_ERROR if the source stream state was inconsistent (such as zalloc or state being NULL).

Thus, the use of `deflateReset()` does not fit the use case well here because it implies calling `deflateInit()`.
Since we additionally need to supply a _windowBits_ argument, which is only supported in `deflateInit2()`, the setup was kept "out" of this in-place deflate method as originally implemented in `SendRequest()` on [L165](https://github.com/open-telemetry/opentelemetry-cpp/pull/3178/files/fadff90c126777a4c68480c8c725847f2963d7a6#diff-21cfd22f6082a873195bac59990cb9f311c98451cba9f2ac3cb48555c5f960eaR165) followed by `deflateEnd()` on [L191](https://github.com/open-telemetry/opentelemetry-cpp/pull/3178/files/fadff90c126777a4c68480c8c725847f2963d7a6#diff-21cfd22f6082a873195bac59990cb9f311c98451cba9f2ac3cb48555c5f960eaR191), once the compression is complete.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
Waiting on @owent to clear the change needed flag and approve.

owent (approved):

owent:
> Waiting on @owent to clear the change needed flag and approve.

LGTM, but  I'm not familiar with zlib APIs.I'm not sure about the details of the usage.

ext/test/http/curl_http_test.cc
```diff
@@ -531,3 +532,105 @@ TEST_F(BasicCurlHttpTests, FinishInAsyncCallback)
...
+
+  auto session = session_manager->CreateSession("http://127.0.0.1:19000");
+  auto request = session->CreateRequest();
+  request->SetUri("post/");
+  request->SetMethod(http_client::Method::Post);
+
+  http_client::Body body;
+
+  // Original source:
+  // http://blog.chenshuo.com/2014/05/incompressible-zlibdeflate-data.html
```
marcalff, line 603: From a technical point, this test `GzipIncompressibleData` is very good.

The problem here is that the code used below to generate incompressible data comes from this blog, which reads as:

> Copyright 2009 [Shuo Chen's Blog](http://blog.chenshuo.com/) All rights reserved.

aka, not open source.

As a result, and it pains me to say that because this is good work:

Please remove test `GzipIncompressibleData` entirely, unless a different source can be found to generate test data.

marcalff (changes requested):
Blocking merge, to resolve concern with `GzipIncompressibleData`

ext/test/http/curl_http_test.cc
marcalff: @chusitoo 
Possible remediation, just define a static array of 256 bytes, with all different values in it, as test data.

chusitoo: That's a good catch. Apologies, I should've checked the fine print.

I had already been able to reproduce this using uniform distributions but was more inclined to use the code snippet from that blog post because it was not subject to change from one run to another.

As you rightfully suggested, and for consistency, I think the trade-off could be to generate an incompressible blob once and statically define it in the unit test, instead of relying on the randomly-generated data for every run.

marcalff (approved):
LGTM, thanks for the fix.

Great work.

PR #3198, opened 2024-12-10
[SDK] Do not frequently create and destroy http client threads

### LINKED ISSUE #3189: [HTTP Client] Reduce the frequency of creating and destroying background threads in HTTP Client.
The HTTP Client may frequently create and destroy threads when it's not always busy, which will cost a lot of resource.
Could we keep the background thread for a while and destroy only when it's free for some more time? Just like the background thread in OTLP file exporter.

I can take this, if it's accepted.

Wait a longer time to destroy HTTP Client's background thread.

**Additional context**
I can casually try to optimize the issue in #3172.

xiehuc:
Fixes #3189

## Changes

keep the background thread for a while and destroy only when it's free for 1 minute.
Just like the background thread in OTLP file exporter.


ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -416,6 +420,23 @@ void HttpClient::MaybeSpawnBackgroundThread()
...
           }
 
+          if (still_running > 0)
+          {
+            continue;
+          }
+
+          // If there is no pending jobs, Exit flush thread if there is not data to flush more
+          // than one minute.
+          if (self->background_thread_wait_for_ != std::chrono::milliseconds{0})
```
owent, line 430: We should avoid delays after calling `Shutdown`, or graceful exiting will take too long.

owent (changes requested):

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -416,6 +424,30 @@ void HttpClient::MaybeSpawnBackgroundThread()
...
 
+          std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
+          if (still_running > 0)
+          {
+            last_free_job_timepoint = now;
+            need_wait_more          = false;
+            continue;
+          }
+
+          std::chrono::milliseconds wait_for;
```
owent, line 435: `std::chrono::milliseconds wait_for = std::chrono::milliseconds::zero();` ?

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -366,6 +356,9 @@ class HttpClient : public opentelemetry::ext::http::client::HttpClient
   std::unique_ptr<std::thread> background_thread_;
   std::chrono::milliseconds scheduled_delay_milliseconds_;
 
+  std::chrono::milliseconds background_thread_wait_for_;
+  std::atomic<bool> is_shutdown;
```
owent, line 360: Nit: `is_shutdown` -> `is_shutdown_`

owent (changes requested):

owent (approved):
LGTM and thanks for your contribution.

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
```diff
@@ -366,6 +357,9 @@ class HttpClient : public opentelemetry::ext::http::client::HttpClient
   std::unique_ptr<std::thread> background_thread_;
   std::chrono::milliseconds scheduled_delay_milliseconds_;
 
+  std::chrono::milliseconds background_thread_wait_for_;
+  std::atomic<bool> is_shutdown_;
```
marcalff, line 361:
```suggestion
  std::atomic<bool> is_shutdown_{false};
```

See #2244 for rationale.

ThomsonTan: With this, then set it in constructure is uncessary.

ext/test/http/curl_http_test.cc
```diff
@@ -531,3 +533,51 @@ TEST_F(BasicCurlHttpTests, FinishInAsyncCallback)
...
+  request->SetUri("get/");
+  auto handler = std::make_shared<GetEventHandler>();
+  session->SendRequest(handler);
+  std::this_thread::sleep_for(std::chrono::milliseconds{10});  // let it enter poll state
+  auto beg = std::chrono::system_clock::now();
+  http_client->FinishAllSessions();
+  http_client.reset();
+  // when use background_thread_wait_for_ should have no side effort on elegant quit
+  // it shoule less than scheduled_delay_milliseconds_
+  // because ci machine may slow, some take 10ms, so we assert it less than 20ms
```
marcalff, line 553:
```suggestion
  // when background_thread_wait_for_ is used, it should have no side effect on elegant quit
  // wait should be less than scheduled_delay_milliseconds_
  // Due to load on CI hosts (some take 10ms), we assert it is less than 20ms
```

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -189,11 +189,15 @@ HttpClient::HttpClient()
       next_session_id_{0},
       max_sessions_per_connection_{8},
       scheduled_delay_milliseconds_{std::chrono::milliseconds(256)},
+      background_thread_wait_for_{std::chrono::minutes{1}},
       curl_global_initializer_(HttpCurlGlobalInitializer::GetInstance())
-{}
+{
+  is_shutdown_.store(false);
+}
```
marcalff, line 196:
```suggestion
{}
```

ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h
marcalff: Yes, constructor is fixed as well.

marcalff:
CI failures in functional tests are unrelated, it also fails in main.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -502,6 +536,27 @@ void HttpClient::ScheduleRemoveSession(uint64_t session_id, HttpCurlEasyResource
...
+  is_shutdown_.store(true, std::memory_order_release);
+  std::unique_ptr<std::thread> background_thread;
+  {
+    std::lock_guard<std::mutex> lock_guard{background_thread_m_};
+    background_thread.swap(background_thread_);
+  }
+
+  if (background_thread && background_thread->joinable())
+  {
+    background_thread->join();
```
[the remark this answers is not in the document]
xiehuc: yes, you are right, since WaitBackgroundThreadExit just used in unittest, it doesn't affect before, i can add it.

marcalff (approved):
LGTM, thanks for the fix.

PR #3214, opened 2024-12-17
[SDK] Fix include instrumentation scope attributes in equal method

### LINKED ISSUE #3208: Providers do not check instrumentation scope attributes when returning cached loggers, tracers, or meters
**Describe your environment** 
main branch at 6ed0651a42c88e347ff37b0f3171ab0f31d9edcb

**Background** 
After working with the otel python api/sdk a bit and reviewing the [1.40.0 OTEL](https://opentelemetry.io/docs/specs/otel/) spec it seems that the c++ implementation of the GetTracer, GetLogger, and GetMeter may not be fully meeting the spec with respect to these methods returning distinct instances based on all parameters including the attributes. Reading the spec I'd expect that calling GetTracer with different attributes would return different tracers but that is not the case. 

**Steps to reproduce**
The following test illustrates the issue. 

```
TEST(TracerProviderTest, DistinctTracers)
{    
    auto processor = opentelemetry::sdk::trace::SimpleSpanProcessorFactory::Create(std::move(opentelemetry::exporter::trace::OStreamSpanExporterFactory::Create()));
    auto provider = opentelemetry::sdk::trace::TracerProviderFactory::Create(std::move(processor));

    auto tracer_1 = provider->GetTracer("my_library", "1.0.0", "my_schema_url");
    auto tracer_2 = provider->GetTracer("my_library", "1.0.0", "my_schema_url", {{ "instance", "one" }});
    auto tracer_3 = provider->GetTracer("my_library", "1.0.0", "my_schema_url", {{ "instance", "two" }});
    
    ASSERT_NE(tracer_1, nullptr);
    ASSERT_NE(tracer_2, nullptr);
    ASSERT_NE(tracer_3, nullptr);

    EXPECT_NE(tracer_1, tracer_2) << "providers must return distinct tracers if there are any differences in parameters passed to GetTracer";
    EXPECT_NE(tracer_2, tracer_3) << "providers must check for any difference in attributes too";   
}

```

Test passes

Test fails

```
 [==========] Running 1 test from 1 test suite.
 [----------] Global test environment set-up.
 [----------] 1 test from TracerProviderTest
 [ RUN      ] TracerProviderTest.DistinctTracers
 Expected: (tracer_1) != (tracer_2), actual: 1 vs 1
 providers must return distinct tracers if there are any differences in parameters passed to GetTracer
 Expected: (tracer_2) != (tracer_3), actual: 1 vs 1
 providers must check for any difference in attributes too
 [  FAILED  ] TracerProviderTest.DistinctTracers (0 ms)
 [----------] 1 test from TracerProviderTest (0 ms total)
 [==========] 1 test from 1 test suite ran. (0 ms total)
 [  PASSED  ] 0 tests.
 [  FAILED  ] TracerProviderTest.DistinctTracers
```

**Additional context**
The current 1.40.0 OTEL specification has language that seems to imply any difference in parameters passed to [get_logger](https://opentelemetry.io/docs/specs/otel/logs/api/#get-a-logger), [get_tracer](https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer), and [get_meter](https://opentelemetry.io/docs/specs/otel/metrics/api/#get-a-meter) must result in a distinct instance. 

> The term identical applied to Tracers describes instances where all parameters are equal. The term distinct applied to Tracers describes instances where at least one parameter has a different value.

> The term identical applied to Meters describes instances where all parameters are equal. The term distinct applied to Meters describes instances where at least one parameter has a different value.

> The term identical applied to Loggers describes instances where all parameters are equal. The term distinct applied to Loggers describes instances where at least one parameter has a different value.

The [`InstrumentationScope::equal`](https://github.com/open-telemetry/opentelemetry-cpp/blob/6ed0651a42c88e347ff37b0f3171ab0f31d9edcb/sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h#L118) method doesn't account for attributes

dbarker:
Fixes #3208

## Changes

This PR updates the GetTracer, GetMeter, and GetLogger methods to include instrumentation scope attributes in the equality check before returning a cached instance. 

 - Adds AttributeEqualToVisitor to sdk/common/attribute_utils.h that compares an OwnedAttributeValue to a AttributeValue 
 - Adds EqualTo method to the AttributeMap class to support checking equality with a KeyValueIterable 
 - Updates the InstrumenationScope::equal method to accept attributes and evaluate them 
 - Updates tests for Get<Logger,Tracer,Meter> methods to check for instance uniqueness based on the OTEL spec. 
 - Updates tests for InstrumentationScope and AttributeMap to cover the equal and EqualTo methods


sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -105,6 +105,60 @@ struct AttributeConverter
...
+struct AttributeEqualToVisitor
+{
+  // Different types are not equal including containers of different element types
+  template <typename T, typename U>
+  bool operator()(const T &, const U &) const noexcept
+  {
+    return false;
+  }
+
+  // Compare the same arithmatic types
```
marcalff, line 122:
```suggestion
  // Compare the same arithmetic types
```

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -105,6 +105,60 @@ struct AttributeConverter
...
+                  const nostd::span<const nostd::string_view> &value) const noexcept
+  {
+    return owned_value.size() == value.size() &&
+           std::equal(owned_value.begin(), owned_value.end(), value.begin(),
+                      [](const std::string &owned_element, nostd::string_view element) {
+                        return owned_element == element;
+                      });
+  }
+
+  // Compare nostd::span<const T> and std::vector<T> for arithmatic types
```
marcalff, line 152:
```suggestion
  // Compare nostd::span<const T> and std::vector<T> for arithmetic types
```

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -162,8 +216,36 @@ class AttributeMap : public std::unordered_map<std::string, OwnedAttributeValue>
...
+          }
+          return false;
+        });
+
+    return is_equal;
+  }
+
 private:
   AttributeConverter converter_;
+  AttributeEqualToVisitor equal_to_visitor_;
```
marcalff, line 248: Could the visitor be a local object in the `EqualTo()` method, instead of a member of the class ?

This would reduce the memory footprint of the attribute map.

dbarker: Right it is a struct with no data. Follows the existing style as the converter.

dbarker:
CI has passed. Ready for review.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -162,8 +216,36 @@ class AttributeMap : public std::unordered_map<std::string, OwnedAttributeValue>
...
+    {
+      return false;
+    }
+
+    const bool is_equal = attributes.ForEachKeyValue(
+        [this](nostd::string_view key,
+               const opentelemetry::common::AttributeValue &value) noexcept {
+          // Perform a linear search to find the key assuming the map is small
+          // This avoids temporary string creation from this->find(std::string(key))
+          for (const auto &kv : *this)
```
owent, line 231: Do you think we can use `this->find(kv.first)` here for better performance?

dbarker: It seems we have to start the search from the `KeyValueIterable::ForEachKeyValue` method and it provides a `nostd::string_view` key while the key type of the map is `std::string`.  So we either have to iterate directly like this or create a temp string every time  as in `auto iter = this->find(std::string(key))`. I chose the first option since it I've heard it could be faster than find for small maps :)  and I didn't feel like string allocation (and its potential to throw) was ideal.  I could profile both options if it seems worth it. 

C++20 gives the ["Heterogeneous comparison lookup" version of the `std::unordered_map::find` method](https://en.cppreference.com/w/cpp/container/unordered_map/find#Version_3) of unordered_map which is really what we need here.  Here's a neat example https://godbolt.org/z/Yqsx4rsY1

marcalff: Thanks for the details. Looks ok to me as is.

marcalff (approved):
LGTM, thanks for the fix.

Very well done, and with good test coverage.

dbarker:
I realized i missed a comment and update to operator== of the InstrumentationScope. Will push a commit shortly.

dbarker:
> I realized i missed a comment and update to operator== of the InstrumentationScope. Will push a commit shortly.

Done. CI passed.

PR #3219, opened 2024-12-22
[EXPORTER] Support handling retry-able errors for OTLP/gRPC

### LINKED ISSUE #2049: Support handling Retryable error for OTLP exporter (OTLP/gRPC and OTLP/HTTP)
Hi,

If OTel collector is down or due to some other reason if OTLP exporter received Retryable error code from OTel collector, it has to be handled and the OTLP exporter has to retry sending the same trace span. Currently there is no code in OTLP exporter(both OTLP/gRPC and OTLP/HTTP) to handle retryable error code in opentelmetry-cpp sdk. As currently retryable error code is not handled for OTLP exporters, so it will discard the trace span when OTel collector returned retryable error code).

Error codes that are retryable and non-retryable is mentioned in OTLP spec at :
https://opentelemetry.io/docs/reference/specification/protocol/otlp/

OTLP/GRPC: https://opentelemetry.io/docs/reference/specification/protocol/otlp/#failures
OTLP/HTTP : https://opentelemetry.io/docs/reference/specification/protocol/otlp/#failures-1

I this this can be a candidate for feature.

discussion link: https://github.com/open-telemetry/opentelemetry-cpp/discussions/2042

Thanks,
Mukesh

chusitoo:
Fixes #2049

## Changes

This change introduces a [retry policy](https://grpc.io/docs/guides/retry/#retry-configuration) in OTLP/gRPC exporter for select [failures](https://opentelemetry.io/docs/specs/otlp/#failures) via the [gRPC service config mechanism](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retry-policy-capabilities)

- Add support to set retry values via environment variables.
- Enabled by default, using the same configuration values as in OTel java and dotnet.
- Users can opt-out of the retry capabilities by zeroing out any (or all) of the retry settings.
- Set service config JSON when creating gRPC channel if all parameters are non-zero.

The changes to support retries for OTLP/HTTP exporter are addressed in #3223


exporters/otlp/CMakeLists.txt
```diff
@@ -317,7 +317,7 @@ if(BUILD_TESTING)
     add_executable(otlp_grpc_exporter_test test/otlp_grpc_exporter_test.cc)
     target_link_libraries(
       otlp_grpc_exporter_test ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}
-      ${GMOCK_LIB} opentelemetry_exporter_otlp_grpc)
+      ${GMOCK_LIB} opentelemetry_exporter_otlp_grpc gRPC::grpc++)
```
chusitoo, line 320: Required for retryable integration tests using real grpc server

owent: We can only link grpc directly when grpc is built as a dynamic library.Or there will be conflict in some environments.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -62,6 +62,18 @@ struct OtlpGrpcClientOptions
   // Concurrent requests
   std::size_t max_concurrent_requests;
 #endif
+
+  /** The maximum number of call attempts, including the original attempt. */
+  std::uint32_t retry_policy_max_attempts{};
+
+  /** The initial backoff delay between retry attempts, random between (0, initial_backoff). */
+  float retry_policy_initial_backoff{};
```
owent, line 70: Why not use `std::chrono::duration<>` here?

chusitoo: I had that as a chrono duration initially, but it was not really of any use for otlp/grpc, so it was moved to otlp/http, where it is being required: https://github.com/open-telemetry/opentelemetry-cpp/pull/3223/files#diff-3b1304d527adad8c09adf1b71b4025f6515ed48aae1d65665f45afdaf25195d4R229

exporters/otlp/CMakeLists.txt
chusitoo: Should this be made optional in that case? What is the right approach?

exporters/otlp/src/otlp_environment.cc
```diff
@@ -80,6 +80,34 @@ static bool GetStringDualEnvVar(const char *signal_name,
   return exists;
 }
 
+static std::uint32_t GetUintEnvVarOrDefault(opentelemetry::nostd::string_view signal_env,
```
marcalff, line 83: Code changes in this file are also present in #3223

Planning to review and merge the OTLP HTTP PR first.

exporters/otlp/test/otlp_grpc_exporter_test.cc
```diff
@@ -17,17 +17,23 @@
 // That is because `std::result_of` has been removed in C++20.
 
 #  include "opentelemetry/exporters/otlp/otlp_grpc_exporter.h"
-
+#  include "opentelemetry/exporters/otlp/otlp_grpc_exporter_factory.h"
 #  include "opentelemetry/exporters/otlp/protobuf_include_prefix.h"
+#  include "opentelemetry/nostd/shared_ptr.h"
```
marcalff, line 22: Move nostd/shared_ptr.h outside of the protobuf include prefix + suffix block.

marcalff (commented):
Great work, thanks.

More comments to follow.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
owent: In my understanding, the exporting of both otlp/http and otlp/grpc will cost much more CPU than type conversion here. I think it's more important to make it clear what this parameters means(We don't know the meaning and the unit of this variable by just the name and comments here), and also float number has EPS and is more imprecise.
What do you think about it?

chusitoo: IMHO, precision is probably very subjective since examples of normal use cases are limited to a single decimal place (and this is how it is formatted here before passing the config settings to grpc library), which seems logical given that measuring backoff in tens of milliseconds or lower is probably a very niche requirement.

I think there is some truth in that chrono duration makes the type more descriptive. Part of the reasoning I went back to float was because I could not find a common place where I could alias this to a more descriptive name without having to repeat it in at least one more header file (for instance, otlp_environment.h and http_client.h).

For now, I will revert/update this in https://github.com/open-telemetry/opentelemetry-cpp/pull/3223 until it is approved/merged to avoid duplicating all these work in progress changes for common code bits...

exporters/otlp/test/otlp_grpc_exporter_test.cc
chusitoo: I never bothered to notice that these includes carried special meaning. Good to know!

exporters/otlp/src/otlp_environment.cc
chusitoo: All changes from the above PR were backported here, as well.

The only minor annoyance with the current use of the WITH_OTLP_RETRY_PREVIEW flag on CI is that none of the maintainer builds have WITH_OTLP_GRPC enabled, so the new code would not be unit tested.

Would it be reasonable also enabling this new feature flag, for instance with "cmake.exporter.otprotocol.test", in order to have at least one build that exercises this code?

marcalff: > All changes from the above PR were backported here, as well.
> 
> The only minor annoyance with the current use of the WITH_OTLP_RETRY_PREVIEW flag on CI is that none of the maintainer builds have WITH_OTLP_GRPC enabled, so the new code would not be unit tested.
> 
> Would it be reasonable also enabling this new feature flag, for instance with "cmake.exporter.otprotocol.test", in order to have at least one build that exercises this code?

Sure, feel free to add it where it makes sense.

We should add WITH_OTLP_GRPC to the maintainer builds as well (in a separate PR), these builds are meant to have a maximum number of features covered, and should also execute the new gRPC functional tests.

exporters/otlp/CMakeLists.txt
owent: Could you try build opentelemetry_exporter_otlp_grpc as dynamic library and build gRPC as static library?

chusitoo: Correct me if I'm wrong but I think this scenario is covered by CI job "cmake.exporter.otprotocol.shared_libs.with_static_grpc.test"

In any case, I tested with the suggested combination:
```sh
$ find /opt/cmake.exporter.otprotocol.shared_libs.with_static_grpc.test/exporters/otlp/ -name "libopentelemetry_exporter_otlp_grpc*"
/opt/cmake.exporter.otprotocol.shared_libs.with_static_grpc.test/exporters/otlp/libopentelemetry_exporter_otlp_grpc_metrics.so
/opt/cmake.exporter.otprotocol.shared_libs.with_static_grpc.test/exporters/otlp/libopentelemetry_exporter_otlp_grpc.so
/opt/cmake.exporter.otprotocol.shared_libs.with_static_grpc.test/exporters/otlp/libopentelemetry_exporter_otlp_grpc_client.so
/opt/cmake.exporter.otprotocol.shared_libs.with_static_grpc.test/exporters/otlp/libopentelemetry_exporter_otlp_grpc_log.so
$ find /usr/ -name "libgrpc*"
/usr/local/lib/libgrpc++.a
/usr/local/lib/libgrpcpp_channelz.a
/usr/local/lib/libgrpc++_alts.a
/usr/local/lib/libgrpc++_unsecure.a
/usr/local/lib/libgrpc_plugin_support.a
/usr/local/lib/libgrpc++_error_details.a
/usr/local/lib/libgrpc++_reflection.a
/usr/local/lib/libgrpc.a
/usr/local/lib/libgrpc_unsecure.a
```

The tests still succeed with the grpc version installed via _./setup_grpc.sh_ so maybe this issue happens with old versions of the lib, or perhaps it doesn't manifest itself in unit testing. I can try doing some more testing in the coming days to try to reproduce this.

owent: Maybe this unit test only use mock APIs and do not actually use the global variables in gRPC, it's hard to reproduce, maybe it's safe to link gRPC to unit test programs. It will not take effect on library so I think we can continue to pay attention to see if the problems of global variable conflict will recur.

marcalff:
Please merge from main to resolve conflicts.
Now that the OTLP HTTP PR is merged, resuming review for OTLP GRPC.

exporters/otlp/CMakeLists.txt
marcalff: `otlp_grpc_exporter_test` is a binary for a unit test, so the opentelemetry-cpp library itself (the OTLP GRPC exporter) is not affected here.

resolving this comment for now, to revisit if somehow this test fails in CI.

marcalff (approved):
LGTM, thanks for the feature.

marcalff:
Looking to merge first #3248, so that maintainers test do cover GRPC in functional tests.

This is to ensure that changes from this PR do work properly in functional tests.

exporters/otlp/test/otlp_grpc_exporter_test.cc
```diff
@@ -355,7 +353,206 @@ TEST_F(OtlpGrpcExporterTestPeer, ConfigUnknownInsecureFromEnv)
...
+
+#  ifdef ENABLE_OTLP_RETRY_PREVIEW
+struct TestTraceService : public opentelemetry::proto::collector::trace::v1::TraceService::Service
+{
+  TestTraceService(std::vector<grpc::StatusCode> status_codes) : status_codes_(status_codes) {}
+
+  inline grpc::Status Export(
+      grpc::ServerContext *context,
+      const opentelemetry::proto::collector::trace::v1::ExportTraceServiceRequest *request,
+      opentelemetry::proto::collector::trace::v1::ExportTraceServiceResponse *response) override
```
marcalff, line 416:
```suggestion
      grpc::ServerContext *,
      const opentelemetry::proto::collector::trace::v1::ExportTraceServiceRequest *,
      opentelemetry::proto::collector::trace::v1::ExportTraceServiceResponse *) override
```

Build warnings in maintainer mode.

marcalff:
@chusitoo 

Thanks for all your recent work, which is now shipped as part of opentelemetry-cpp 1.19.0

Please consider becoming a CNCF member.
If you choose to apply, @lalitb and myself @marcalff will be your sponsors.

Process:

* https://github.com/open-telemetry/community/blob/main/guides/contributor/membership.md

Example application:

* https://github.com/open-telemetry/community/issues/1122

Regards,
-- Marc

Issue #3172, opened 2024-11-26
Test BasicCurlHttpTests.FinishInAsyncCallback is unstable in CI.


marcalff:
```
[ RUN      ] BasicCurlHttpTests.FinishInAsyncCallback
ext/test/http/curl_http_test.cc:525: Failure
Value of: handlers[i]->got_response_
  Actual: false
Expected: true

[  FAILED  ] BasicCurlHttpTests.FinishInAsyncCallback (5 ms)
[----------] 15 tests from BasicCurlHttpTests (7643 ms total)

[----------] Global test environment tear-down
[==========] 15 tests from 1 test suite ran. (7643 ms total)
[  PASSED  ] 14 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] BasicCurlHttpTests.FinishInAsyncCallback

 1 FAILED TEST
===========================================================================
```

To investigate, unclear if this is a failure due to the test itself, or the production code tested.

owent:
Not sure if it's related to CPU cache line. Should we try to declare got_response_ as `atomic<bool>` to see if it will happen again?

marcalff:
Reopen: the test is still unstable, even after using atomics.

```
[ RUN      ] BasicCurlHttpTests.FinishInAsyncCallback
ext/test/http/curl_http_test.cc:645: Failure
Value of: handlers[i]->got_response_.load(std::memory_order_acquire)
  Actual: false
Expected: true

[  FAILED  ] BasicCurlHttpTests.FinishInAsyncCallback (4 ms)
```

PR #3223, opened 2024-12-30
[EXPORTER] Support handling retry-able errors for OTLP/HTTP

### LINKED ISSUE #2049: Support handling Retryable error for OTLP exporter (OTLP/gRPC and OTLP/HTTP)

chusitoo:
Contributes to #2049 

## Changes

This change introduces a retry mechanism for OTLP/HTTP for select [failures](https://opentelemetry.io/docs/specs/otlp/#failures-1), mimicking the same [exponential backoff approach](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff) used in OTLP/gRPC.

- Add support to set retry values via environment variables.
- Enabled by default, using the same configuration values as in OTel dotnet, java and js.
- Users can opt-out of the retry capabilities by zeroing out any (or all) of the retry settings.
- Similar to OTLP/gRPC, retries are transparent to the user so only the last attempt is "bubbled up" as the actual response.

The changes to support retries for OTLP/gRPC exporter are addressed in #3219


ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1198,11 +1247,6 @@ CURLcode HttpOperation::Send()
 
   CURLcode code = curl_easy_perform(curl_resource_.easy_handle);
   PerformCurlMessage(code);
-  if (CURLE_OK != code)
```
chusitoo, line 1201: This looked redundant

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1248,7 +1292,7 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   async_data_->callback = std::move(callback);
 
   session->GetHttpClient().ScheduleAddSession(session->GetSessionId());
-  return code;
+  return CURLE_OK;
```
chusitoo, line 1295: I think this might be cleaner than returning `code` given that it can only be CURLE_OK, as otherwise it would've been returned to the caller on L1277

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -426,6 +432,49 @@ void HttpOperation::Cleanup()
...
+  };
+
+  const auto is_retryable = std::find(kRetryableStatusCodes.cbegin(), kRetryableStatusCodes.cend(),
+                                      response_code_) != kRetryableStatusCodes.cend();
+
+  return is_retryable && (last_curl_result_ == CURLE_OK) &&
+         (retry_attempts_ < retry_policy_.max_attempts);
+}
+
+std::chrono::system_clock::time_point HttpOperation::NextRetryTime()
```
chusitoo, line 451: The exponential backoff logic is modeled after the gRPC retry policy so that both OTLP gRPC and HTTP behave more or less consistently.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -474,13 +474,20 @@ bool HttpClient::MaybeSpawnBackgroundThread()
...
+              if (operation)
               {
                 // Session can not be destroyed when calling PerformCurlMessage
                 auto hold_session = session->shared_from_this();
-                session->GetOperation()->PerformCurlMessage(result);
+                operation->PerformCurlMessage(result);
+
+                if (operation->IsRetryable())
+                {
+                  self->pending_to_retry_sessions_.push_front(session);
```
chusitoo, line 489: I believe this is safe being lock-free because it is only managed by the background thread and not public. In case the session is removed by `doAbortSessions()` or `doRemoveSessions()`, the pointer would be ignored and removed when processed by `doRetrySessions()`

exporters/otlp/src/otlp_environment.cc
```diff
@@ -80,6 +80,34 @@ static bool GetStringDualEnvVar(const char *signal_name,
...
 
+static std::uint32_t GetUintEnvVarOrDefault(opentelemetry::nostd::string_view signal_env,
+                                            opentelemetry::nostd::string_view generic_env,
+                                            std::uint32_t default_value)
+{
+  std::string value;
+
+  if (GetStringDualEnvVar(signal_env.data(), generic_env.data(), value))
+  {
+    return static_cast<std::uint32_t>(std::strtoul(value.c_str(), nullptr, 10));
```
marcalff, line 91: Here, any garbage in the environment variable that does not fit into a ul string will be silently ignored.

Please implement `GetUIntEnvironmentVariable` in sdk/common instead, and log warnings when invalid strings are found.

See existing code for Bool and Duration.

exporters/otlp/src/otlp_environment.cc
```diff
@@ -80,6 +80,34 @@ static bool GetStringDualEnvVar(const char *signal_name,
...
+
+static float GetFloatEnvVarOrDefault(opentelemetry::nostd::string_view signal_env,
+                                     opentelemetry::nostd::string_view generic_env,
+                                     float default_value)
+{
+  std::string value;
+
+  if (GetStringDualEnvVar(signal_env.data(), generic_env.data(), value))
+  {
+    return std::strtof(value.c_str(), nullptr);
```
marcalff, line 105: Likewise, implement `GetFloatEnvironmentVariable` in sdk/common.

marcalff (commented):
Excellent work.

This is a first pass of comments, more to follow, 
see some changes on environment variables

sdk/src/common/env_variables.cc
```diff
@@ -193,6 +296,52 @@ bool GetStringEnvironmentVariable(const char *env_var_name, std::string &value)
...
+  static constexpr auto kDefaultValue = 0U;
+  std::string raw_value;
+  bool exists = GetRawEnvironmentVariable(env_var_name, raw_value);
+  if (!exists || raw_value.empty())
+  {
+    value = kDefaultValue;
+    return false;
+  }
+
+  if (!ParseNumber(raw_value.c_str(), value))
```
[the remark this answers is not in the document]
chusitoo: No problem, I misunderstood the previous comment in that `strtoul` and `strtof` were not good enough to report any errors

exporters/otlp/src/otlp_environment.cc
```diff
@@ -1125,6 +1157,174 @@ std::string GetOtlpDefaultLogsCompression()
...
+    return value;
+  }
+
+  return 5U;
+}
+
+std::uint32_t GetOtlpDefaultMetricsRetryMaxAttempts()
+{
+  constexpr char kSignalEnv[]  = "OTEL_EXPORTER_OTLP_METRICS_RETRY_MAX_ATTEMPTS";
+  constexpr char kGenericEnv[] = "OTEL_EXPORTER_OTLP_RETRY_MAX_ATTEMPTS";
```
marcalff, line 1177: These new environment variables are not in the spec (yet).

Please use the OTEL_CPP_xxx namespace, as in `OTEL_CPP_EXPORTER_OTLP_METRICS_RETRY_MAX_ATTEMPTS`, and likewise for all friends.

Once the spec is extended, this will be revisited to use the official names.

marcalff (commented):
Continuing review,

environment variables not in the spec should use the OTEL_CPP_xxx namespace.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -769,6 +788,43 @@ bool HttpClient::doRemoveSessions()
...
+    {
+      auto easy_handle = operation->GetCurlEasyHandle();
+      curl_multi_remove_handle(multi_handle_, easy_handle);
+      curl_multi_add_handle(multi_handle_, easy_handle);
+      retry_it = decltype(retry_it){pending_to_retry_sessions_.erase(std::next(retry_it).base())};
+      has_data = true;
+    }
+    else
+    {
+      ++retry_it;
```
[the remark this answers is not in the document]
chusitoo: My impression would be that we want to "reschedule" as many operation that are ready as possible. 
This is because `curl_multi_info_read` on L465 runs in a loop and should be able to handle several handles "simultaneously". 
Otherwise, we'd need to break out of that loop and look here again to see who is next.

A tradeoff to short-circuiting this list could to stop and exit at the first handle that is still scheduled sometime "in the future". This assumes that everything else until the head of the list would be newer and therefore not ready. I don't believe that it is possible in practice to have handles that are out of order, but I also did not spend enough time trying to exclude that possibility.

sdk/src/common/env_variables.cc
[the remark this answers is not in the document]
chusitoo: Implemented more or less as described above. I had to add some extra checks because `strtoul` only fails when out of range for positive values. For negative values, it just wraps around so need to manually check if value falls outside `uint` range given that it returns `ulong`. Also, changed it to `strtoull`  to make this check possible in MSVC, since the former function behaves differently on Windows, i.e. `sizeof(uint) == sizeof(ulong)`

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
chusitoo: After a second thought, I think it is safe to break here. I believe the background thread can not simply exit because the easy handle that needs retrying was not removed (this is intended) and `Cleanup()` was not called on the operation as it is retryable.

The side-effect in this case is that, if some session that could've been ready for retry earlier, it would wait for slightly longer than needed (maybe an extra 256ms for the poll / wait).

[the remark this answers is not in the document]
chusitoo: Change made, unlikely to be the source of performance issues, in any case.

marcalff (commented):
Excellent work.

The only part remaining at this point is to add a feature flag:

In CMakeList.txt, define a new option `WITH_OTLP_RETRY_PREVIEW`, default OFF

The same option name should be fine for both HTTP and gRPC.

In api/CMakeList.txt, add a target_compile_definitions (aka, a define), for `ENABLE_OTLP_RETRY_PREVIEW`

Use ifdef ENABLE_OTLP_RETRY_PREVIEW to protect the code path for this feature. A good place could be inside the implementation of IsRetryable().

Adjust the unit tests using the same ifdef, to make sure they pass when compiling with and without WITH_OTLP_RETRY_PREVIEW.

Having a feature flag is needed because:

* the environment variables are not in the spec yet, so this code is considered experimental and subject to change
* the code itself is new, so this mitigates risks for people in production with the OTLP exporter. Adoption will be voluntary since the default is OFF.

Much later, the feature flag will be ON by default, and then removed, assuming we get this in the spec at some point, to declare the whole thing stable.

marcalff (approved):
LGTM

Excellent work, thanks for the feature.

marcalff:
Ok to ignore CI failurews for test `BasicCurlHttpTests.FinishInAsyncCallback`, due to #3172

marcalff:
@chusitoo 

Sorry for the merge conflicts.

This should be trivial, as we both added some members in the same place, let me know if you need help with resolving.

Please note that there is a new constructor in ext, for HttpClient, so your changes will be needed there as well.

I will review again after conflict resolution anyway.

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
owent: But `still_running` will not be set to `1` when `curl_multi_poll` or `curl_multi_wait` just time out. And the background thread may exit when `pending_to_retry_sessions_` is not empty.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_record_exporter_options.h
```diff
@@ -101,6 +102,18 @@ struct OPENTELEMETRY_EXPORT OtlpHttpLogRecordExporterOptions
 
   /** Compression type. */
   std::string compression;
+
+  /** The maximum number of call attempts, including the original attempt. */
+  std::uint32_t retry_policy_max_attempts{};
+
+  /** The initial backoff delay between retry attempts, random between (0, initial_backoff). */
+  std::chrono::duration<float> retry_policy_initial_backoff{};
```
owent, line 110: Just wondering, why not use `std::chrono::microseconds`, which just like other components.

chusitoo: This is partly due to the fact that OTLP gRPC and HTTP shame the same retry policy settings (i.e types).
There are no constraints for HTTP, but if we look at the gRPC service config, it defines the initial backoff and max backoff as a `google.protobuf.Duration` type, which is described as follows:

> A Duration represents a signed, fixed-length span of time represented as a count of seconds and fractions of seconds at nanosecond resolution. It is independent of any calendar and concepts like "day" or "month". It is related to Timestamp in that the difference between two Timestamp values is a Duration and it can be added or subtracted from a Timestamp.

This suggests that this type supports a decimal representation of seconds and my understanding is that std:chrono::*seconds are all helpers of `std::chrono::duration<some integral type, some ratio>`.

One approach is to convert to std::milli, std::micro, etc, when parsing these from the environment. 
Another solution is to keep the representation as `std::chrono::duration<some floating type, ratio<1>>` and be able to work with "decimal seconds" the same way chrono works with the builtin helpers.
My intuition was the the latter was a bit more natural to work with, and simply duration_cast to milliseconds when calculating the next retry time, which is a single place in the code.

Again, this is a side-effect of trying to keep a close representation to the protobuf duration type while only needing it for calculations in OTLP/HTTP and not really using the type to do anything useful in OTLP/gRPC (which is why at some point I had reverted to simply using plain float types).

This out of the way, I have no preference and could switch to using one of the builtin helper chrono types if you think it makes more sense.

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
chusitoo: I do not fully follow. Just like the other checks, doRetrySessions() is called in twice; the second check happens when `still_running` = 0, just before the background thread is about to break out of the loop:

```cpp
          if (still_running == 0)
          {
            std::lock_guard<std::mutex> lock_guard{self->background_thread_m_};
            // Double check, make sure no more pending sessions after locking background thread
            // management

            ... other checks ...

            // Check if pending easy handles can be retried
            if (self->doRetrySessions())
            {
              still_running = 1;
            }

            // If there is no pending jobs, we can stop the background thread.
            if (still_running == 0)
            {
            ...
```

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_record_exporter_options.h
owent: I think we can use `sdk::common::GetDurationEnvironmentVariable` to get duration settings from environment variables(which just like `GetDurationDualEnvVar` in `otlp_environment.cc`) or export and use `sdk::common::GetTimeoutFromString` to parse a duration from string.

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
owent: But `doRetrySessions()` still returns `false` when all retry sessions are pending?

marcalff:
@chusitoo 

Please resolve the merge conflicts, or indicate if you need help with this.

Thanks (and sorry for the conflicts)

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
chusitoo: Thanks, I see your point now. Maybe doRetrySessions could return `true` if the collection is not empty, even though nothing could be ready to retry at that precise moment, but this is not optimal because the first check would be spamming with false positives. Let me think about a better approach.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_record_exporter_options.h
chusitoo: The timeout type returned by `GetDurationDualEnvVar` is a duration with an typename `Rep` being of type long.
Let's say assume backoff is now parsed as a value representing 1500 milliseconds, instead of 1.5 seconds.
This would still require a conversion to duration<float> later on when passed in service config to configure gRPC retries, because it expects a value of the form "1.5s", which is a bit ugly as shown here: https://godbolt.org/z/qcxM5dozo

In addition to that, I think working with duration<float> in this case makes more sense for OTLP/HTTP also because it retains the same type when doing arithmetic operations to calculate the next retry, so less type casting down the line.

chusitoo:
> @chusitoo
> 
> Please resolve the merge conflicts, or indicate if you need help with this.
> 
> Thanks (and sorry for the conflicts)

Apologies for the delay, having a hard time finding time for this during the week.

Conflicts resolved but I have to look into addressing a concern raised in [this discussion](https://github.com/open-telemetry/opentelemetry-cpp/pull/3223#discussion_r1918429786)

ext/src/http/client/curl/http_client_curl.cc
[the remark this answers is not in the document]
chusitoo: I think a good compromise for this would be to remain status quo for the first check but report everything that is pending (i.e. future retries) for the second check. Hopefully, that's and acceptable solution.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_log_record_exporter_options.h
marcalff: I think the code as is looks good.

Having consistency between HTTP and GRPC is a good thing.

Also, this part is protected by a PREVIEW flag, and is considered experimental.

Note that the spec does not even gives details about configuration parameters, even less units to express timeouts.

Once this part is clarified in the spec, we will have to adjust anyway, so this is not something blocking for now, and not worth spend too much time on, because whatever we decide now is not guaranteed to be the final desired result.

Marking this conversation resolved.

ext/src/http/client/curl/http_client_curl.cc
```diff
@@ -811,6 +830,42 @@ bool HttpClient::doRemoveSessions()
   return has_data;
 }
 
+bool HttpClient::doRetrySessions(bool report_all)
+{
```
marcalff, line 834: nit, optional:

Please use ifdef ENABLE_OTLP_RETRY_PREVIEW as well in HttpClient::doRetrySessions(), to protect this code path.

marcalff:
To summarize remaining outstanding issues for this review:

1)
I have a nit comment on HttpClient::doRetrySessions(bool)
@chusitoo please take a look.

2)
HttpClient::doRetrySessions(bool) now takes a bool report_all parameter.

@owent 
This should resolve the concern you raised, please take a look.

owent (approved):
LGTM now, thanks.

PR #3227, opened 2025-01-04
[TEST] Functional tests for OTLP/gRPC with mutual TLS 

### LINKED ISSUE #2208: [TEST] Functional tests for OTLP GRPC MTLS export
This is a follow up on:
* https://github.com/open-telemetry/opentelemetry-cpp/pull/2120

Add functional tests for validating OTLP gRPC with client certificate chain  + Server ca certificate. 
Ref- https://github.com/open-telemetry/opentelemetry-cpp/tree/main/functional/otlp

chusitoo:
Fixes #2208 

## Changes

- There was only a main for http functional tests so a _func_grpc_main_ was added following a similar pattern
- Add debug log line in OTLP grpc exporter to capture result in tests
- Reuse existing CI step "cmake_otprotocol_test" since it is the only one enabling WITH_OTLP_GRPC_SSL_MTLS_PREVIEW
- Minor tweaks to scripts to accommodate using both test binaries

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,6 +452,16 @@ jobs:
       run: |
         sudo ./ci/setup_grpc.sh
         ./ci/do_ci.sh cmake.exporter.otprotocol.test
+        - name: generate test cert
+    - name: generate test cert      
+      env:
+        CFSSL_VERSION: 1.6.3
+      run: |
+        sudo -E ./tools/setup-cfssl.sh
+        (cd ./functional/cert; ./generate_cert.sh)
+    - name: run func test
+      run: |
+        (cd ./functional/otlp; ./run_test.sh)
 
   cmake_modern_protobuf_grpc_with_abseil_test:
     name: CMake test (with modern protobuf,grpc and abseil)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,8 +452,7 @@ jobs:
       run: |
         sudo ./ci/setup_grpc.sh
         ./ci/do_ci.sh cmake.exporter.otprotocol.test
-        - name: generate test cert
-    - name: generate test cert      
+    - name: generate test cert
       env:
         CFSSL_VERSION: 1.6.3
       run: |
--- a/functional/otlp/func_grpc_main.cc
+++ b/functional/otlp/func_grpc_main.cc
@@ -1,0 +1,588 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stdint.h>
+#include <stdio.h>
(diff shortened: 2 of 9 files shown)


functional/otlp/run_test_mode.sh
```diff
@@ -22,27 +22,33 @@ set -e
...
+[ -z "${TEST_EXECUTABLE}" ] && echo "Please specify TEST_EXECUTABLE name" && exit 1
+
+[ -z "${TEST_URL}" ] && echo "Please specify TEST_URL endpoint (without scheme)" && exit 1
+
 export CERT_DIR=../cert
 
-export TEST_BIN_DIR=${BUILD_DIR}/functional/otlp/
+export TEST_BIN_DIR="${BUILD_DIR}/functional/otlp/"
+
+[ ! -f "${TEST_BIN_DIR}/${TEST_EXECUTABLE}" ] && echo "::notice::Executable ${TEST_EXECUTABLE} not built in this configuration" && exit 0
```
chusitoo, line 33: Because this script gets its inputs from run_test.sh and all other CI checks that run these tests only build WITH_OTLP_HTTP, I thought this would be the easiest solution to supporting existing checks that only test HTTP functionality as well this check that does gRPC, as well.

For visibility, this notice is also shown in the annotations section:

image

functional/otlp/func_grpc_main.cc
```diff
@@ -0,0 +1,821 @@
...
+  internal_log::GlobalLogHandler::SetLogLevel(internal_log::LogLevel::Debug);
+
+  rc = run_test_case(opt_test_name);
+  return rc;
+}
+
+void set_common_opts(otlp::OtlpGrpcExporterOptions &opts)
+{
+  opts.endpoint            = opt_endpoint;
+  opts.timeout             = std::chrono::milliseconds{100};
```
chusitoo, line 430: Setting a timeout to speed things up a bit, since some tests might go on for 5+ seconds before erroring out.
The observable behavior is that a "deadline exceeded" will be output as a result instead of the actual reason. This is accounted for in `expect_connection_failed()`.

If deemed more appropriate, it could be reverted so as to let the tests fail normally.

.github/workflows/ci.yml
```diff
@@ -452,6 +452,15 @@ jobs:
       run: |
         sudo ./ci/setup_grpc.sh
         ./ci/do_ci.sh cmake.exporter.otprotocol.test
+    - name: generate test cert
```
chusitoo, line 455: Reusing this test because it is the only one enabling WITH_OTLP_GRPC_SSL_MTLS_PREVIEW and also WITH_OTLP_GRPC so it sort of made sense

marcalff (approved):
LGTM, thanks for the tests.

PR #3236, opened 2025-01-09
[SDK] Optimize Metric Processing for Single Collector with Delta Temporality

lalitb:
This introduces a fast path for cases with a single collector/reader using delta temporality. It skips unnecessary intermediate processing and directly generates metrics for callback, reducing memory usage and CPU overhead. This change is supposed to improve performance for common single-collector scenarios while retaining existing functionality for other cases.

## Changes


sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -51,6 +51,36 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   AggregationTemporality aggregation_temporarily =
       collector->GetAggregationTemporality(instrument_descriptor_.type_);
 
+  // Fast path for single collector with delta temporality and counter, updown-counter, histogram
+  if (collectors.size() == 1 && collector->GetAggregationTemporality(
+                                    instrument_descriptor_.type_) == AggregationTemporality::kDelta)
```
marcalff, line 56: Nit: use the `aggregation_temporarily` local variable, from the previous line, in the if test.

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -51,6 +51,36 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
...
+    delta_metrics->GetAllEnteries(
+        [&metric_data](const MetricAttributes &attributes, Aggregation &aggregation) {
+          PointDataAttributes point_data_attr;
+          point_data_attr.point_data = aggregation.ToPoint();
+          point_data_attr.attributes = attributes;
+          metric_data.point_data_attr_.emplace_back(std::move(point_data_attr));
+          return true;
+        });
+
+    return callback(metric_data);
```
marcalff, line 81: No idea if this can happen or not, does the code in the fast path needs to pay attention to previous metrics in `unreported_metrics_` ?

If not, consider to add a comment to explain why, and/or an assert to make sure there are no unreported metrics.

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -51,6 +51,36 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   AggregationTemporality aggregation_temporarily =
       collector->GetAggregationTemporality(instrument_descriptor_.type_);
 
+  // Fast path for single collector with delta temporality and counter, updown-counter, histogram
+  if (collectors.size() == 1 && collector->GetAggregationTemporality(
+                                    instrument_descriptor_.type_) == AggregationTemporality::kDelta)
+  {
+    // If no metrics, early return
+    if (delta_metrics->Size() == 0)
```
ThomsonTan, line 59: Could this be moved out to be in the common code path above the check on `collectors.size() == 1`?

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -51,6 +51,35 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   AggregationTemporality aggregation_temporarily =
       collector->GetAggregationTemporality(instrument_descriptor_.type_);
 
+  // Fast path for single collector with delta temporality and counter, updown-counter, histogram
+  // This path doesn't need to aggregated-with/contribute-to the unreported_metric_, as there is
+  // no other reader configured to collect those data.
+  if (collectors.size() == 1 && aggregation_temporarily == AggregationTemporality::kDelta)
```
ThomsonTan, line 57: Short circuit return if `collectors.size() == 0`?

sdk/src/metrics/state/temporal_metric_storage.cc
lalitb: Thanks done.

sdk/src/metrics/state/temporal_metric_storage.cc
lalitb: The fast-path doesn't need to consider the unreported_metrics, I have added comments to make it clear.

sdk/src/metrics/state/temporal_metric_storage.cc
lalitb: Tried that, but seems the unit-tests for the path for multiple collectors is failing with the change, and I don't want to touch that path for now :) Will look into it separate of this PR.

sdk/src/metrics/state/temporal_metric_storage.cc
lalitb: Yes, can be done, but this is separate of the current issue handled? Will add it as separate PR.

ThomsonTan (approved):

marcalff (approved):

PR #3240, opened 2025-01-10
[BUILD] Fix error ‘uint8_t’ does not name a type with gcc-15

### LINKED ISSUE #3239: Compiling `severity.h` with gcc-15 fails
**Describe your environment** 

Reported originally in https://bugs.gentoo.org/946146 for opentelemetry-cpp-1.16.1, but still reproducible with 1.18.0.

Build with gcc-15 fails with:
```
/var/tmp/portage/dev-cpp/opentelemetry-cpp-1.16.1/work/opentelemetry-cpp-1.16.1/api/include/opentelemetry/logs/severity.h:49:7: error: ‘uint8_t’ does not name a type
   49 | const uint8_t kMaxSeverity = 255;
      |       ^~~~~~~
```

With some other gcc/clang versions it worked due to transitive includes, but it was very unpredictable (for example is available after inclusion if `string_view` with libc++, or `ostream` with libstdc++, but not for gcc-13). Here is a conformance view if you are interested: https://godbolt.org/z/hEs18z8Pd

See linked pull-request with a fix.

AngryLoki:
Fixes #3239

## Changes

Added missing `cstdint` header for `uint8_t` type in `api/include/opentelemetry/logs/severity.h`.

### DIFF
--- a/api/include/opentelemetry/logs/severity.h
+++ b/api/include/opentelemetry/logs/severity.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <cstdint>
+
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/version.h"
 


marcalff (approved):
LGTM, thanks for the fix.

lalitb (approved):

PR #3245, opened 2025-01-15
[SDK] Support OTEL_SDK_DISABLED environment variable

### LINKED ISSUE #1631: [SDK] Support OTEL_SDK_DISABLED environment variable
The SDK spec recently added a new environment variable, OTEL_SDK_DISABLED

See open-telemetry/opentelemetry-specification#2679

Implement support for OTEL_SDK_DISABLED, for all signals.

marcalff:
Fixes #1631

## Changes


sdk/src/logs/provider.cc
```diff
@@ -0,0 +1,44 @@
...
+{
+  bool disabled = opentelemetry::sdk::common::GetSdkDisabled();
+
+  if (!disabled)
+  {
+    opentelemetry::logs::Provider::SetLoggerProvider(lp);
+  }
+}
+
+void Provider::SetEventLoggerProvider(
```
ThomsonTan, line 31: Do we still need this? Seems evet has been removed from the spec?

marcalff: Thanks,

Removed all changes related to `EventLogger`, since this is going away.

lalitb:
@marcalff Thanks for a PR. Just a different thought. Instead of providing a set-provider methods within the SDK which check for the disabled env-variable, wouldn't it be better to more granular control within get_logger/tracer/meter methods to return noop objects if the SDK is disabled?

marcalff:
> @marcalff Thanks for the PR. Just a thought—rather than including set-provider methods in the SDK that check for the `disabled` , wouldn't it be better to offer more granular control within the `get_logger/tracer/meter` methods, allowing them to return no-op objects when the SDK is disabled? This way, users wouldn't need to use a new api, and it also ensures that new loggers or tracers won't function if the environment variable is disabled after the application starts.


@lalib

Thanks for the suggestion.

I considered it as well, but decided against it for the following reasons:

* In terms or robustness, technical risk, and performance overhead,
  it is better to talk to the default Noop provider when the SDK is
  disabled, rather that invoke code in a real SDK implementation that will
  degrade itself to a Noop.

* For SDK code that opentelemetry-cpp has control of, this may work,
  but there are unknowns with SDK classes implemented by third parties,
  which still may not honor OTEL_SDK_DISABLED.

* Implementing a check in the SDK on get_tracer() implies that the SDK
  always honor environment variables defined in the spec.
  This is invalidated by the declarative configuration project
  (config.yaml), because the spec says explicitly that every environment
  variable beside OTEL_EXPERIMENTAL_CONFIG_FILE is to be ignored
  when a yaml configuration file is used.
  Implementing the check inside get_tracer() in the SDK would break this.
  See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#declarative-configuration

As a result, the current proposal:

* Puts the test for OTEL_SDK_DISABLED in the SetTracerProvider() code path,
  when a configured SDK is installed by the application main().

* Will work to support declarative configuration later, I plan to add
  opentelemetry::sdk::trace::Provider::SetTracerProviderNoEnv(),
  to be used with a config.yaml file.

* Leaves the SDK execution (Get() code path) unchanged,
  and independent on how the SDK was configured.

As for environment variables changing during the process execution,
I don't think there is any expectation for this to work.
For example, changing an endpoint URL after the SDK is already configured
will have no effect.

In my understanding,
environment variables are meant to be a simple way to change the SDK
configuration without having to change code in the main application,
but are not meant to change during execution.
Honoring a change in OTEL_SDK_DISABLED between start and end span will cause
chaos, for unclear benefit.

CHANGELOG.md
```diff
@@ -18,6 +18,47 @@ Increment the:
...
+    * opentelemetry::sdk::logs::Provider::SetLoggerProvider()
+
+  * These methods do support the `OTEL_SDK_DISABLED` environment variable,
+    unlike the corresponding existing API Provider classes.
+
+  * Applications are encouraged to migrate from the API to the SDK
+    `Provider` classes, to benefit from this feature.
+
+  * Note that the API methods will no longer be public starting
+    with OPENTELEMETRY_ABI_VERSION >= 2, only the SDK methods will
```
lalitb, line 42: I think we shouldn't remove the API method. As the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracerprovider) requires this. It should be instead ok to recommend user to use the SDK provided methods if they need support to enable/disable SDK.

``
Normally, the TracerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default TracerProvider.
``

lalitb:
Thanks @marcalff for explanation. The changes look good to me - just a comment regarding removal of API provided methods.

CHANGELOG.md
marcalff: Thanks for the review.

For this PR, I reverted the changes that removed the Setter from the API, to avoid blocking the OTEL_SDK_DISABLED change itself.

The spec needs revisions in my opinion, so I filed an issue for this:

* https://github.com/open-telemetry/opentelemetry-specification/issues/4400

Once the spec discussion is settled, we can revisit this part.

sdk/test/logs/logger_sdk_test.cc
```diff
@@ -303,3 +309,46 @@ TEST(LoggerSDK, EventLog)
...
+  auto test_provider = nostd::shared_ptr<opentelemetry::logs::LoggerProvider>(new TestProvider());
+  logs_sdk::Provider::SetLoggerProvider(test_provider);
+
+  auto lp = logs_api::Provider::GetLoggerProvider();
+
+  // Check that the implementation was pushed by calling TestLogger's GetName()
+  nostd::string_view schema_url{"https://opentelemetry.io/schemas/1.11.0"};
+  auto logger = lp->GetLogger("TestLogger", "opentelelemtry_library", "", schema_url);
+  ASSERT_EQ("test logger", logger->GetName());
+}
```
lalitb, line 354: Not sure if I am missing something, is this test related to getting logger/provider with OTEL_SDK_ENABLED?

lalitb (approved):

sdk/test/logs/logger_sdk_test.cc
marcalff: This test was in the API, in logger_test.cc, and was moved to the SDK when the setter was no longer in the API.

Now that the changes in API are reverted, this test is no longer needed in the SDK, I will clean this up.

Good catch.

Issue #2431, opened 2023-12-05
[BUILD] Improve compilation speeds by avoiding std::regex instantiation in header file


kylepl:
As a follow-up to https://github.com/open-telemetry/opentelemetry-cpp/pull/2430, I realized that [`std::regex instantiation`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/api/include/opentelemetry/trace/trace_state.h#L252) takes a decent chunk of time (like 500-750ms on my dev laptop). and given it does not need to exist (since it is needed only once, not for _every_ translation unit that users Tracer), seems like a nice thing to improve.

I'm looking for advice on how to do so - basically where to put a definition in a .cc file. All of `/api/` is just headers from what I can see, and so looking for guidance about that.

Thanks!

marcalff:
@kylepl 

Thanks for the previous cleanup on regex.

For architectural reasons, the entire opentelemetry-cpp `API` **MUST** be header only, this is a very strong constraint. We even managed the not so trivial feat to implement header only singletons in C++.

Regarding the use of regex in the SDK, this part is trivial, and I think all uses are located in a *.cc file already.

About the use of regex in the API, the only place that depends on it comes from trace_state.h.

The following methods:

* TraceState::FromHeader()
* TraceState::Get()
* TraceState::Set()

all depends indirectly on regex.

Now, because trace_state.h is used in span_context.h, and because span_context.h is used in many places, indeed the regex code is compiled in many translation units.

Possible techniques to cut down dependencies like this are:

* move code in a different helper class, using friends declarations if needed, and using a forward declaration to cut
* move the inlined, in class implementation of a method to an inlined, outside of class implementation, defined in a separate header file (trace_state_inline.h)

The main challenge here is to do so without breaking the API, so that existing user code calling my_trace_state->Set() still build and run properly.

From a quick investigation on this topic, I can't think of a viable fix to limit the regex exposure from trace_state.h (without breaking changes that is).

A totally different solution can be to also force `OPENTELEMETRY_HAVE_WORKING_REGEX` to be 0, so that the alternate (non regex) code is used instead, even when a "working" regex is available.

This would require yet another cmake compiling option.

kylepl:
I appreciate the detailed response!

For my local builds, I am already using `OPENTELEMETRY_HAVE_WORKING_REGEX=0`, which does do the job.

[...]

PR #3276, opened 2025-02-15
Fix compilation with Regex being disabled

kylepl:
## Changes

Fix compilation with Regex disabled:
* It was missing an include for std::any_of
* Avoid declaring unused constants (at least my LLVM errors out)

## Testing

I compiled it locally, and it works - but I assume either not-regex isn't covered by testing, our my use of LLVM 19 in my own code base means it was being more strict.

### DIFF
--- a/sdk/src/metrics/instrument_metadata_validator.cc
+++ b/sdk/src/metrics/instrument_metadata_validator.cc
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#include <algorithm>
 #include <string>
 
 #include "opentelemetry/nostd/string_view.h"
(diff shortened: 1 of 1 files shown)


sdk/src/metrics/instrument_metadata_validator.cc
```diff
@@ -1,6 +1,9 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#if OPENTELEMETRY_HAVE_WORKING_REGEX
```
lalitb, line 4: The CI is failing for this - you need to include `instrument_metadata_validator.h` before using this macro. The changes look good otherwise.

lalitb:
>  but I assume either not-regex isn't covered by testing

Yes, this was previously tested with GCC 4.8, but it was removed from CI after C++11 support was dropped.

kylepl:
Thanks for the comment. I've moved where the check is, into the else of the existing part.

kylepl:
> Yes, this was previously tested with GCC 4.8, but it was removed from CI after C++11 support was dropped.

Does this mean... we should just drop support for this? As in, is the condition it is disabled in ever hit? Perhaps just by people like me that disable it. Note that I do it just to save compilation time (a la https://github.com/open-telemetry/opentelemetry-cpp/issues/2431).

owent:
> > Yes, this was previously tested with GCC 4.8, but it was removed from CI after C++11 support was dropped.
> 
> Does this mean... we should just drop support for this? As in, is the condition it is disabled in ever hit? Perhaps just by people like me that disable it. Note that I do it just to save compilation time (a la #2431).

I prefer to keep the old implementation because we have some old toolchains in  some peoduct environment.I can maintain this toolchain but don't want to have a large patch file.
I'm not sure but in my understanding, just a macro checking 
 won't significantly increase compile time,template expansion and new header files will cost more.

owent (approved):
LGTM, and thanks.

lalitb (approved):

PR #3282, opened 2025-02-20
[SDK] Add logger scope configurator

### LINKED ISSUE #2641: [SDK] Add simple scope configuration to Tracer, Meter, Logger

psx95:
Fixes #2641 

Adds scope configurator for Loggers.

Similar work was done for Tracers in #3137 and Meters in #3268.

## Changes

 - Updates the Logging SDK to add support for [LoggerConfigurator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfigurator) as per the spec changes.
 -  Adds a LoggerConfig class that can be used to configure a logger's behavior as per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfig).


sdk/include/opentelemetry/sdk/logs/logger_config.h
```diff
@@ -0,0 +1,60 @@
...
+   * @return a static constant LoggerConfig that represents a logger configured with default
+   * behavior.
+   */
+  static LoggerConfig Default();
+
+private:
+  explicit LoggerConfig(const bool disabled = false) : disabled_(disabled) {}
+
+  bool disabled_;
+  static const LoggerConfig kDisabledConfig;
```
owent, line 55: Just wondering, why not use a static local variable for `kDisabledConfig` and `kDefaultConfig` to avoid unnecessary creation and ensure stable constructor and destructor order?

sdk/include/opentelemetry/sdk/logs/logger_config.h
```diff
@@ -0,0 +1,60 @@
...
+   * Returns if the Logger is enabled or disabled. Loggers are enabled by default.
+   * @return a boolean indicating if the Logger is enabled. Defaults to true.
+   */
+  bool IsEnabled() const noexcept;
+
+  /**
+   * Returns a LoggerConfig that represents an enabled Logger.
+   * @return a static constant LoggerConfig that represents an enabled logger.
+   */
+  static LoggerConfig Enabled();
```
owent, line 34: Because this class can be used as a parameter of factory, which is public and exported. I think we should use `OPENTELEMETRY_EXPORT` to export all public static functions and default constructor, destructor and assignment operators.

psx95: The comments around this Macro do not provide the details about its usages. What does "exporting" public static functions mean in this context ?

sdk/include/opentelemetry/sdk/logs/logger_config.h
psx95: I was not aware of the concept of static local variable in C++. Seems to be a good idea.

sdk/include/opentelemetry/sdk/logs/logger_config.h
psx95: Thank you for the explanation, I have made this change.

Based on this, I think similar change should be made for [meter_config](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/meter_config.cc) and [tracer_config](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/trace/tracer_config.cc) ?

I think private static  variables could be applied there as well.

psx95: Created issue to track this work for tracer and meter config. 

https://github.com/open-telemetry/opentelemetry-cpp/issues/3302

marcalff (approved):
LGTM, thanks for the feature.

PR #3285, opened 2025-02-21
[API] Deprecate event logger

### LINKED ISSUE #3193: Deprecate event logger
As Event API and SDK are deprecated in the spec (https://github.com/open-telemetry/opentelemetry-specification/pull/4319), we should deprecate [EventLogger](https://github.com/open-telemetry/opentelemetry-cpp/blob/23562e6da33155419c84e3f62254c738d74daa09/api/include/opentelemetry/logs/event_logger.h#L20).

owent:
Fixes #3193

## Changes

- Remove event logger and provider in ABI v2
- Deprecate event logger and provider in ABI v1
- Add some attributes to `.clang-format` to fix some format error on Windows.


.vscode/settings.json
```diff
@@ -0,0 +1,153 @@
...
+    "xtree": "cpp",
+    "xutility": "cpp",
+    "pointers": "cpp",
+    "multi_span": "cpp",
+    "span_ext": "cpp",
+    "string_span": "cpp",
+    "gsl_byte": "cpp"
+  } //,
+  //"editor.formatOnSave": false
+}
```
lalitb, line 153: is this file intended?

.clang-format
```diff
@@ -59,3 +59,15 @@ IndentPPDirectives: AfterHash
...
+AttributeMacros:
+  - OPENTELEMETRY_UNLIKELY
+  - OPENTELEMETRY_LIKELY
+  - OPENTELEMETRY_MAYBE_UNUSED
+  - OPENTELEMETRY_DEPRECATED
+  - OPENTELEMETRY_API_SINGLETON
+  - OPENTELEMETRY_LOCAL_SYMBOL
+  - OPENTELEMETRY_EXPORT
+  - OPENTELEMETRY_SANITIZER_NO_MEMORY
+  - OPENTELEMETRY_SANITIZER_NO_THREAD
```
lalitb, line 72: Can we move this formatting in separate PR. I don't think it is related to deprecating event logger.

lalitb:
@owent thanks for the PR. The event logger related changes look good in general. However, can we move the clang formatting related changes out of scope of this PR? we can have separate PR for that if needed.

.vscode/settings.json
marcalff: This file contains path locals to the dev machine, so it should probably not be added, at least as is.

@owent Please confirm if this is intended, or remove.

marcalff (approved):
This change contains two independent parts:

* deprecation (abi v1) and removal (abi v2) of the event logger
* clang-format changes with compiler attributes macros

Approving this PR because both look ok.

See comments on vscode file.

@owent For future work, please separate independent changes in different PRs.

This helps with the git history, and helps when a backport is needed to an older branch for some fixes.

lalitb (approved):
LGTM.

.vscode/settings.json
owent: Sorry, this file is commited by mistake, It's removed now

owent:
> This change contains two independent parts:
> 
> * deprecation (abi v1) and removal (abi v2) of the event logger
> * clang-format changes with compiler attributes macros
> 
> Approving this PR because both look ok.
> 
> See comments on vscode file.
> 
> @owent For future work, please separate independent changes in different PRs.
> 
> This helps with the git history, and helps when a backport is needed to an older branch for some fixes.

@lalitb @marcalff The format changes are included into this PR because there will be strange format style in some files of log's codes. They may cause some confuse.
I found that clang-format doesn't recognize our attribute macros as attributes but sometimes as class names. Including these format changes should help reduce confusion.

api/include/opentelemetry/semconv/exception_attributes.h
```diff
@@ -26,8 +26,7 @@ namespace exception
  * It's no longer recommended to record exceptions that are handled and do not escape the scope of a
  * span.
  */
-OPENTELEMETRY_DEPRECATED
-static constexpr const char *kExceptionEscaped = "exception.escaped";
+OPENTELEMETRY_DEPRECATED static constexpr const char *kExceptionEscaped = "exception.escaped";
```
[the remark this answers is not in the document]
owent: Yes, these changes are made by clang-format after add `AttributeMacros` into `.clang-format`

.clang-format
owent: I found that clang-format doesn't recognize our attribute macros as attributes but sometimes as class names, which will make clang-format make strange style formatting for this PR. So my suggestion is including these settings in this PR to reduce confuse.

PR #3286, opened 2025-02-21
[ETW] Add configuration to export 64-bit integer as timestamp

ThomsonTan:
## Changes

Add macro `OPENTELEMETRY_ATTRIBUTE_TIMESTAMP_PREVIEW` to ETW exporter to support exporting a given 64-bit integer as timestamp type, because timestamp is not supported in the attribute type.

### DIFF
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
@@ -323,6 +323,33 @@ public:
     }
     evt[ETW_FIELD_LOG_SEVERITY_NUM] = static_cast<uint32_t>(severity);
     evt[ETW_FIELD_LOG_BODY]         = std::string(body.data(), body.length());
+
+#if defined(OPENTELEMETRY_ATTRIBUTE_TIMESTAMP_PREVIEW)
+
+    for (const auto &attr : cfg.timestampAttributes)
+    {
+      auto it = evt.find(attr);
+      if (it != evt.end())
+      {
+        auto value_index = it->second.index();
+        if (value_index != exporter_etw::PropertyType::kTypeInt64 &&
+            value_index != exporter_etw::PropertyType::kTypeUInt64)
+        {
+          continue;
+        }
+        int64_t filetime = value_index == exporter_etw::PropertyType::kTypeUInt64
+                        ? nostd::get<uint64_t>(it->second)
+                        : nostd::get<int64_t>(it->second);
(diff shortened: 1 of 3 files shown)


exporters/etw/include/opentelemetry/exporters/etw/etw_logger.h
```diff
@@ -323,6 +323,33 @@ class Logger : public opentelemetry::logs::Logger
...
+        {
+          continue;
+        }
+        int64_t filetime = value_index == exporter_etw::PropertyType::kTypeUInt64
+                        ? nostd::get<uint64_t>(it->second)
+                        : nostd::get<int64_t>(it->second);
+        constexpr int64_t FILETIME_EPOCH_DIFF = 11644473600LL; // Seconds from 1601 to 1970
+        constexpr int64_t HUNDRED_NANOSECONDS_PER_SECOND = 10000000LL;
+        int64_t unix_time_seconds = (filetime / HUNDRED_NANOSECONDS_PER_SECOND) - FILETIME_EPOCH_DIFF;
+        int64_t unix_time_nanos = unix_time_seconds * 1'000'000'000 + (filetime % HUNDRED_NANOSECONDS_PER_SECOND) * 100;
```
lalitb, line 346: Do we need to check for the overflow, as `unix_time_seconds` can be large value in uint64_t, and we are multiplying with 1'000'000'000 ?

ThomsonTan: I am not sure how should overflow be handled here, but it seems it is converted into `time64_t` which doesn't take the signing of `int64_t` into account (see [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/3212b0f68faf1878431c3956025e702ab849ccd5/exporters/etw/include/opentelemetry/exporters/etw/utils.h#L258), probably passing overflowed value is fine here, which will just be treated as sort of invalid timestamp and returns some new or old timestamp.

Please feel free to correct me.

lalitb: I did some test here - https://godbolt.org/z/dGoen89bh, and can see the overflow happening for the filetime > year 2280. The existing [code ](https://github.com/open-telemetry/opentelemetry-cpp/blob/3212b0f68faf1878431c3956025e702ab849ccd5/exporters/etw/include/opentelemetry/exporters/etw/utils.h#L258) is using the conversion for the current timestamp (span start time, log timestamp etc), which should be fine as it won't overflow. However, with timestamp coming as attribute can have any value, and this can overflow.
This should be fine for now, but good to add some code comment so it can be revisited.

lalitb (approved):

PR #3304, opened 2025-03-20
fix: buffer overrun in the gRPC sample project

lukeina2z:
Building the sample on Windows, the gRPC sample server/client app fails to function correctly. The propagated trace context is corrupted, containing garbage characters after the trace parent string.

The string buffer from gRPC cannot be safely treated as a null-terminated C-string. To ensure safe copying, we must use its length property instead.


ThomsonTan (approved):

marcalff (approved):
LGTM, thanks for the fix.

Issue #1998, opened 2023-02-23
OTLP example fails to build with error "undefined reference to `grpc::Status::OK' "


VivekSubr:
**Describe your environment** 
WSL ubuntu jammy

**Steps to reproduce**
Install grpc as per doc: https://grpc.io/docs/languages/cpp/quickstart/
Build opentelemetry,
```
cmake .. -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_CXX_STANDARD=17 -DBUILD_SHARED_LIBS=ON -DWITH_ABSEIL=OFF -DWITH_OTLP=ON -DBUILD_TESTING=OFF
cmake --build . --target all

```

**What is the expected behavior?**
Build should go through

**What is the actual behavior?**
Got error,
```
[ 79%] Linking CXX executable example_otlp_grpc
cd /home/vivek/Client-Server/module/opentelemetry-cpp/build/examples/otlp && /usr/bin/cmake -E cmake_link_script CMakeFiles/example_otlp_grpc.dir/link.txt --verbose=1
/usr/bin/c++ -rdynamic CMakeFiles/example_otlp_grpc.dir/grpc_main.cc.o -o example_otlp_grpc  -Wl,-rpath,/home/vivek/Client-Server/module/opentelemetry-cpp/build/examples/common/foo_library:/home/vivek/Client-Server/module/opentelemetry-cpp/build/exporters/otlp:/home/vivek/Client-Server/module/opentelemetry-cpp/build/sdk/src/trace:/home/vivek/Client-Server/module/opentelemetry-cpp/build/sdk/src/resource:/home/vivek/Client-Server/module/opentelemetry-cpp/build/sdk/src/metrics:/home/vivek/Client-Server/module/opentelemetry-cpp/build/sdk/src/common ../common/foo_library/libcommon_foo_library.so ../../exporters/otlp/libopentelemetry_exporter_otlp_grpc.so ../../exporters/otlp/libopentelemetry_otlp_recordable.so ../../sdk/src/trace/libopentelemetry_trace.so ../../sdk/src/resource/libopentelemetry_resources.so ../../sdk/src/metrics/libopentelemetry_metrics.so ../../sdk/src/common/libopentelemetry_common.so ../../exporters/otlp/libopentelemetry_exporter_otlp_grpc_client.so ../../libopentelemetry_proto.a /usr/lib/x86_64-linux-gnu/libprotobuf.a
[...]

PR #3307, opened 2025-03-21
[BUILD] fix build of otlp_grpc_exporter_test with modern grpc

### LINKED ISSUE #3306: unable to build with grpc 1.71.0
**Describe your environment** 
main at a57063f7a2

GRPC_VERSION="v1.71.0" 
PROTOBUF_VERSION="29.0" 
ABSEIL_CPP_VERSION="20240722.1"

**Steps to reproduce**

1. Build a docker image with the grpc, protobuf, and abseil versions above
```sh
docker build --build-arg GRPC_VERSION="v1.71.0" --build-arg PROTOBUF_VERSION="29.0" --build-arg ABSEIL_CPP_VERSION="20240722.1" --build-arg USER_UID="$(id -u)" --build-arg USER_GID="$(id -g)"  -t opentelemetry-cpp-dev:latest -f ./.devcontainer/Dockerfile.dev .
```
2. Run a bash session and build the otlp grpc exporter
```sh
docker run -it -v "$PWD:/workspaces/opentelemetry-cpp" opentelemetry-cpp-dev:latest bash

cmake -S . -B ~/build -DWITH_OTLP_GRPC=ON
cmake --build ~/build --parallel
ctest --test-dir ~/build

```

build and test passes

Link error from the `otlp_grpc_exporter_test`
```
/usr/bin/ld: ../../libopentelemetry_proto_grpc.so: undefined reference to `grpc::GetGlobalCallbackHook()'
collect2: error: ld returned 1 exit status
make[2]: *** [exporters/otlp/CMakeFiles/otlp_grpc_exporter_test.dir/build.make:210: exporters/otlp/otlp_grpc_exporter_test] Error 1
make[1]: *** [CMakeFiles/Makefile2:6475: exporters/otlp/CMakeFiles/otlp_grpc_exporter_test.dir/all] Error 2
```

**Additional context**
The `otlp_grpc_exporter_test` adds a link to `gRPC::grpc++` and shouldn't

dbarker:
Fixes #3306

## Changes
- makes grpc link publicly to the grpc client 
- remove otlp_grpc_exporter_test's link to gRPC::grpc++
- removes extra protobuf links to tests 
- adds curl link to the zipkin test

### DIFF
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -321,7 +321,7 @@ if(BUILD_TESTING)
     add_executable(otlp_grpc_exporter_test test/otlp_grpc_exporter_test.cc)
     target_link_libraries(
       otlp_grpc_exporter_test ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}
-      ${GMOCK_LIB} opentelemetry_exporter_otlp_grpc gRPC::grpc++)
+      ${GMOCK_LIB} opentelemetry_exporter_otlp_grpc)
     gtest_add_tests(
       TARGET otlp_grpc_exporter_test
       TEST_PREFIX exporter.otlp.
--- a/exporters/otlp/CMakeLists.txt
+++ b/exporters/otlp/CMakeLists.txt
@@ -38,18 +38,11 @@ if(WITH_OTLP_GRPC)
   # targets that depend on opentelemetry_proto_grpc.
   target_link_libraries(
     opentelemetry_exporter_otlp_grpc_client
-    PUBLIC opentelemetry_sdk opentelemetry_common opentelemetry_ext
+    PUBLIC opentelemetry_sdk opentelemetry_common
            # gRPC::grpc++ must be linked before opentelemetry_proto_grpc.
-           opentelemetry_proto_grpc
-    PRIVATE gRPC::grpc++)
+           opentelemetry_proto_grpc gRPC::grpc++
+    PRIVATE opentelemetry_ext)
 
-  get_target_property(GRPC_INCLUDE_DIRECTORY gRPC::grpc++
-                      INTERFACE_INCLUDE_DIRECTORIES)
-  if(GRPC_INCLUDE_DIRECTORY)
-    target_include_directories(
-      opentelemetry_exporter_otlp_grpc_client BEFORE
-      PUBLIC "$<BUILD_INTERFACE:${GRPC_INCLUDE_DIRECTORY}>")
-  endif()
   target_include_directories(
     opentelemetry_exporter_otlp_grpc_client
     PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
@@ -290,9 +283,8 @@ endif()
 
 if(BUILD_TESTING)
   add_executable(otlp_recordable_test test/otlp_recordable_test.cc)
-  target_link_libraries(
-    otlp_recordable_test ${GTEST_BOTH_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT}
-    opentelemetry_otlp_recordable protobuf::libprotobuf)
+  target_link_libraries(otlp_recordable_test ${GTEST_BOTH_LIBRARIES}
+                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_otlp_recordable)
   gtest_add_tests(
     TARGET otlp_recordable_test
     TEST_PREFIX exporter.otlp.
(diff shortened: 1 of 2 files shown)


owent (approved):
LGTM, thanks.
This issue was introduced by PR #3219. Could you please add some comments describing how to reproduce it?  
Otherwise, other contributors might reintroduce the same issue in the future, and we would have difficulty identifying it.

dbarker:
> LGTM, thanks. This issue was introduced by PR #3219. Could you please add some comments describing how to reproduce it? Otherwise, other contributors might reintroduce the same issue in the future, and we would have difficulty identifying it.

Thanks! Here are the steps to reproduce the build error in a linux container. 

1. build and run a dev container setting the latest grpc version with compatible protobuf and abseil versions. 
````
cd <path-to-opentelemetry-cpp>
docker build --build-arg GRPC_VERSION="v1.71.0" --build-arg PROTOBUF_VERSION="29.0" --build-arg ABSEIL_CPP_VERSION="20240722.1" --build-arg USER_UID="$(id -u)" --build-arg USER_GID="$(id -g)"  -t opentelemetry-cpp-dev:latest -f ./.devcontainer/Dockerfile.dev .
docker run -it -v "$PWD:/workspaces/opentelemetry-cpp" opentelemetry-cpp-dev:latest bash
````
2. build the otlp exporter and run tests
```
cmake -S . -B ~/build -DWITH_OTLP_GRPC=ON
cmake --build ~/build 
ctest --test-dir ~/build
```

I'll log an issue to upgrade to the latest release of vcpkg and use newer versions of grpc in ci. The link error solved by this PR should exist in grpc v1.67.0+.

**Comment by dbarker** [comment]
The [build failure](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/13996263060/job/39192105339?pr=3307#step:4:5135) is in an unrelated test of the otlp file exporter that [passed](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/13984803212/job/39156548025#step:4:5069) in the previous run on the PR.

**Comment by dbarker** [comment]
@owent  after more testing I found that the grpc link to the client should be public since the client is a public interface and exposes grpc headers. Making the link public also allows the tests to get the transitive dependency (solves a link error when building shared libraries on later versions of grpc). While I was in that file I've also removed extra protobuf links to tests since opentelemetry_proto exposes it publicly. I've tested this on old versions of grpc (1.49.2) and the latest (1.71.0). The zipkin exporter test needed curl and I've added it the branch.

File: exporters/otlp/CMakeLists.txt [hunk 1]
```diff
@@ -38,18 +38,11 @@ if(WITH_OTLP_GRPC)
...
-    PRIVATE gRPC::grpc++)
-
-  get_target_property(GRPC_INCLUDE_DIRECTORY gRPC::grpc++
-                      INTERFACE_INCLUDE_DIRECTORIES)
-  if(GRPC_INCLUDE_DIRECTORY)
-    target_include_directories(
-      opentelemetry_exporter_otlp_grpc_client BEFORE
-      PUBLIC "$<BUILD_INTERFACE:${GRPC_INCLUDE_DIRECTORY}>")
-  endif()
+           opentelemetry_proto_grpc gRPC::grpc++
```
owent, line 43: gRPC client definitions are not exposed to users, so it's fine to keep the gRPC dependency private. Alternatively, we could use `PUBLIC $<BUILD_INTERFACE:gRPC::grpc++> PRIVATE $<INSTALL_INTERFACE:gRPC::grpc++>` to expose the include directories internally. Symbol conflicts don't only cause issues during compilation or linking but will result in runtime crashes.

More details can be found at #1603 , #1891 , #1998

owent (changes requested):

dbarker:
After more testing I found that building against grpc 1.71.0 is working on the current main branch. I'm not able to reproduce the link error originally reported [example here](https://github.com/dbarker/opentelemetry-cpp/actions/runs/13982683037/job/39150838265#step:5:1782).

I suspect the root issue was an ill-configured grpc install in my docker container and github runner. I had a typo on the arguments to the setup-grpc.sh script and `abseil` should be `abseil-cpp` https://github.com/open-telemetry/opentelemetry-cpp/blob/2c9b68fbaeef95bae9d3b5df20e6cf29f62c34d8/.devcontainer/Dockerfile.dev#L33

I'm going to close this PR and open separate PRs for: 
1. Making the grpc headers public in the installed otlp grpc client target (so grpc will not need to be linked explicit when creating a shared client) 
2. Fix the missing curl link to the zipkin exporter test
3. Fix the grpc install in the devcontainer docker file.

dbarker:
Closing.

exporters/otlp/CMakeLists.txt
owent: Do you think we can move `OtlpGrpcClientReferenceGuard` out of `otlp_grpc_client.h`, so we don't need to include `otlp_grpc_client.h` in headers of `OtlpGrpcClientFactory`?

dbarker: I think we'd need to remove all grpc from `otlp_grpc_client.h` using pmpl and some generic status codes. Alternatively we could give the client a base class and use it in the constructors of the exporters and return from the factory - then deprecate the constructors and create method with the OtlpGrpcClient shard ptr.

dbarker: actually #3321 is a good solution as long as the client methods are accessed outside the scope of the exporters. good call.

PR #3318, opened 2025-03-24
[BUILD] Remove `WITH_ABSEIL`

### LINKED ISSUE #3116: Clarifications needed around WITH_ABSEIL

### LINKED ISSUE #3309: API VariantTest.Get test fails on windows when WITH_ABSEIL=ON
**Describe your environment** 
main at b4328c0bc10
windows 2019 (setup with `./ci/setup_windows_ci_environment.ps1`)
WITH_ABSEIL=ON

**Steps to reproduce**
build and test setting WITH_ABSEIL=ON

test passes

test fails
```sh
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from VariantTest
[ RUN      ] VariantTest.Get
Running main() from D:\a\opentelemetry-cpp\opentelemetry-cpp\tools\vcpkg\buildtrees\gtest\src\v1.14.0-bcf93537a8.clean\googletest\src\gtest_main.cc
D:\a\opentelemetry-cpp\opentelemetry-cpp\api\test\nostd\variant_test.cc(52): error: Death test: { nostd::get<float>(w); }
    Result: threw an exception.
 Error msg:
[  DEATH   ] 
[  DEATH   ] D:\a\opentelemetry-cpp\opentelemetry-cpp\api\test\nostd\variant_test.cc(52):: Caught std::exception-derived exception escaping the death test statement. Exception message: Bad variant access
[  DEATH   ] 

[  FAILED  ] VariantTest.Get (8 ms)
[----------] 1 test from VariantTest (8 ms total)

[==========] 1 test from 1 test suite ran. (9 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] VariantTest.Get
```

**Additional context**
There are no windows ci jobs that test `WITH_ABSEIL=ON`

owent:
Fixes #3116 

Maybe also fixes #3309 

## Changes


marcalff (commented):
The WITH_ABSEIL removal itself looks ok.

Not sure why the file opentelemetry-cpp-config.cmake.in was reformated.

Should we change the formatting scripts to ignore this file instead ?

.devcontainer/README.md
dbarker, line 1: We'll still need to install abseil for the protobuf and grpc build in the dev container.

owent: Sorry for the mistake.It's restored now.

marcalff:
In terms of stability and risk mitigation, there is:

* #3328
* #3220
* #3318 (this PR)

I suspect there will be minor merge conflicts between the WITH_ABSEIL removal and the CMake COMPONENTS refactoring, so this raise the question in which order to merge (for one, the cmake.in template has moved)

@owent Can this PR wait after the opentelemetry-cpp 1.20 release, so we merge all cmake changes at once ?

@dbarker FYI

marcalff (approved):
LGTM, thanks for the cleanup.

dbarker:
I'm okay merging this before the components pr and working through the conflicts, they should be manageable.

CMakeLists.txt
```diff
@@ -400,10 +398,8 @@ if(WITH_PROMETHEUS)
   endif()
 endif()
 
-if(WITH_ABSEIL)
-  if(NOT TARGET absl::strings)
-    find_package(absl CONFIG REQUIRED)
-  endif()
+if(WITH_OTLP_GRPC)
+  find_package(absl CONFIG REQUIRED)
```
ThomsonTan, line 402: Why is absl only required for `WITH_OTLP_GRPC`, not for all cases?

owent:
> In terms of stability and risk mitigation, there is:
> 
> * [[RELEASE] Release opentelemetry-cpp 1.20.0 #3328](https://github.com/open-telemetry/opentelemetry-cpp/pull/3328)
> * [[INSTALL] add cmake components to the package #3220](https://github.com/open-telemetry/opentelemetry-cpp/pull/3220)
> * [Remove `WITH_ABSEIL` #3318](https://github.com/open-telemetry/opentelemetry-cpp/pull/3318) (this PR)
> 
> I suspect there will be minor merge conflicts between the WITH_ABSEIL removal and the CMake COMPONENTS refactoring, so this raise the question in which order to merge (for one, the cmake.in template has moved)
> 
> @owent Can this PR wait after the opentelemetry-cpp 1.20 release, so we merge all cmake changes at once ?
> 
> @dbarker FYI

OK, this is a small PR and it can be merged last, I will resolve the conflicts when the time comes.

marcalff:
All, there is more to this, as it just turned out.

Ubuntu 20.04 is being removed in 3 days, so we MUST upgrade ci as we are still using it.

When upgrading CI, the abseil builds now fails, even when the abseil version used is the same.

From a quick look, this sounds due to a compiler version change (gcc 9 to 11), confusing the abseil code to think it can use std::in_place_t, which is a C++17 construct. It seems the abseil code does not honor the fact we build in C++14 then, and assumes std::in_place_t is ok to use.

Long story short:

* I am very reluctant to debug and fix this upgrade to support some abseil new combination we have not been exposed to before, just for 3 days, given WITH_ABSEIL is going to be removed anyway
* I am now considering to merge WITH_ABSEIL removal now, before shipping opentelemetry-cpp 1.20, to avoid the issue entirely that comes with the ubuntu ci upgrade from 20.04

@open-telemetry/cpp-approvers @dbarker Any thoughts ?

CMakeLists.txt
owent: otel-cpp's API do not depend abseil-cpp's headers now, and there is no conlict paths between internal version and external version  of abseil-cpp.

There are some components use both protobuf and it may include abseil-cpp's headers which have the same path as the internal version. That's why we need abseil-cpp when we use a new version of protobuf.

The protobuf's package file will call `find_package(absl CONFIG REQUIRED)` and import it when need. But some versions of gRPC do not import it, so we only need to import it when using gRPC.

ThomsonTan (approved):

ThomsonTan:
As @dbarker mentioned, I agree on merging this after the release if it could break vcpkg port in potential.

dbarker:
I've added a comment here https://github.com/open-telemetry/opentelemetry-cpp/pull/3330#discussion_r2017837337. I'm optimistic that the ci upgrade can proceed by adding the CXX_STANDARD=14 env var to the failing jobs.

marcalff:
> Would disabling the 20.04 ci runners before the 3 days, then create the release without this PR, and work on the ci upgrade post release work?

Yes, this is the safest path I think.

marcalff:
Do not merge (now) tag: To merge after opentelemetry-cpp 1.20 is released

PR #3322, opened 2025-03-25
[METRICS SDK] Fix hash collision in MetricAttributes

### LINKED ISSUE #3060: Hash collision risk of metric data aggregation
Hello, I have concerns regarding the hash collision issue with the current implementation of metric data aggregation, related to  this PR #1993. 

https://github.com/open-telemetry/opentelemetry-cpp/blob/c6d394547b012077998b6bd74665145488e85ffe/sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h#L201

The aggregation map uses size_t as the key type, and the hash code of the attribute set is treated as a unique identifier. However, if two different attribute sets happen to generate the same hash code, it could lead to collisions, and I haven’t seen any handling for this scenario in the current logic.

Why not consider using the `MetricAttributes` instead of its hash code as the key of the `unordered_map`?

Thanks

ThomsonTan:
Fixes #3060 

## Changes

attributes_hashmap_benchmark shows flat result before and after this change.


sdk/include/opentelemetry/sdk/common/attributemap_hash.h
```diff
@@ -74,27 +74,6 @@ inline size_t GetHashForAttributeMap(const OrderedAttributeMap &attribute_map)
   return seed;
 }
 
-// Calculate hash of keys and values of KeyValueIterable, filtered using callback.
-inline size_t GetHashForAttributeMap(
```
ThomsonTan, line 78: It is not expected to calculate hash on `KeyValueIterable` by traversing it, because the order of the elements should not affect the hashing. `MetricAttributes` need to be constructed from it before hash. So remove this function.

lalitb:
@ThomsonTan - Could you also include the benchmark results from before and after in the comments. A slight regression is acceptable, as it's a tradeoff for fixing collision.

ThomsonTan:
> @ThomsonTan - Could you also include the benchmark results from before and after in the comments. A slight regression is acceptable, as it's a tradeoff for fixing collision.

Here are the before/after benchmark result. It looks like that the wall time doubled while the CPU time reduced slightly.

* Before

```bash
./measurements_benchmark
2025-03-26T00:43:08+00:00
Running measurements_benchmark
Run on (16 X 2445.43 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 32768 KiB (x1)
Load Average: 1.85, 3.23, 1.73
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
BM_MeasurementsTest    2044809 ns       148990 ns         3856
```

* After

```bash
measurements_benchmark
2025-03-26T00:44:26+00:00
Running measurements_benchmark
Run on (16 X 2445.43 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 32768 KiB (x1)
Load Average: 1.02, 2.64, 1.65
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
BM_MeasurementsTest    4287068 ns        95297 ns         1000
```

ThomsonTan:
@lalitb it seems the major perf gap is from the below original code. It doesn't sort the incoming `KeyValueIterator` before hashing and query, and the new code does the sort before the hash lookup.

```cpp
  Aggregation *GetOrSetDefault(const opentelemetry::common::KeyValueIterable &attributes,
                               const AttributesProcessor *attributes_processor,
                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback,
                               size_t hash)
  {
    auto it = hash_map_.find(hash);
    if (it != hash_map_.end())
    {
      return it->second.second.get();
    }
```

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -197,8 +209,13 @@ class AttributesHashMap
...
   size_t Size() { return hash_map_.size(); }
 
+#ifdef UNIT_TESTING
+  size_t BucketCount() { return hash_map_.bucket_count(); }
+  size_t BucketSize(size_t n) { return hash_map_.bucket_size(n); }
+#endif
+
 private:
-  std::unordered_map<size_t, std::pair<MetricAttributes, std::unique_ptr<Aggregation>>> hash_map_;
+  std::unordered_map<MetricAttributes, std::unique_ptr<Aggregation>, CustomHash> hash_map_;
```
maxmarkin01, line 218: can we store MetricAttributes as a pointer? it's going to be a lot copying on rehashing...

ThomsonTan: Do you mean make the key as `std::shared_ptr<MetricAttributes>`, it may work with a more complicated custom hash. But I think `MetricsAttributes` should be created and moved into the `std::unordered_map`, so copying and rehashing should be in rare.

I remember the the exporting process does make a copy of `MetricsAttributes`, but there is an optimization for the common scenario. @lalitb any thoughts on this?

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -71,108 +76,115 @@ class AttributesHashMap
    */
   Aggregation *GetOrSetDefault(const opentelemetry::common::KeyValueIterable &attributes,
                                const AttributesProcessor *attributes_processor,
-                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback,
-                               size_t hash)
+                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback)
   {
-    auto it = hash_map_.find(hash);
+    MetricAttributes attr{attributes, attributes_processor};
```
lalitb, line 81: can we add a TODO here for possible optimization - as for now it will copy the attribute list in memory even if it already exist in the hashmap.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
lalitb: I believe we are already avoiding deep copy with std::move while inserting the MetricsAttributes, also during rehashing, the hashmp will move(not copy) the MetricAttributes objects to their new locations. In general, I do see options for optimization at other places, but we can visit it separate to the PR.

sdk/include/opentelemetry/sdk/metrics/state/filtered_ordered_attribute_map.h
```diff
@@ -22,22 +30,65 @@ class AttributesProcessor;  // IWYU pragma: keep
...
+    return _hash == other._hash && static_cast<const OrderedAttributeMap &>(*this) ==
+                                       static_cast<const OrderedAttributeMap &>(other);
+  }
+
+  size_t GetHash() const { return _hash; }
+
+  void UpdateHash() { _hash = GetHashForAttributeMap(*this); }
+
+private:
+  size_t _hash = (std::numeric_limits<size_t>::max)();
```
lalitb, line 79: should it be `hash_` ?

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
ThomsonTan: Added TODO.

marcalff:
Found no issues on this patch (LGTM), but I don't know the metric aggregation code well enough to approve.
Deferring code review to @lalitb , will merge once approved.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -71,108 +76,117 @@ class AttributesHashMap
...
+    return result.first->second.get();
   }
 
-  Aggregation *GetOrSetDefault(std::function<std::unique_ptr<Aggregation>()> aggregation_callback,
-                               size_t hash)
+  Aggregation *GetOrSetDefault(const MetricAttributes &attributes,
+                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback)
   {
-    auto it = hash_map_.find(hash);
+    auto it = hash_map_.find(attributes);
```
lalitb, line 103: This can be changed to:
```cpp
  return GetOrSetDefault(MetricAttributes{attributes}, std::move(aggregation_callback));
```

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -71,108 +76,117 @@ class AttributesHashMap
...
    * Set the value for given key, overwriting the value if already present
    */
   void Set(const opentelemetry::common::KeyValueIterable &attributes,
            const AttributesProcessor *attributes_processor,
-           std::unique_ptr<Aggregation> aggr,
-           size_t hash)
+           std::unique_ptr<Aggregation> aggr)
   {
-    auto it = hash_map_.find(hash);
+    MetricAttributes attr{attributes, attributes_processor};
```
lalitb, line 142: This method can be simplified:
```cpp
  MetricAttributes attr{attributes, attributes_processor};
  Set(std::move(attr), std::move(aggr));
```

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -71,108 +76,117 @@ class AttributesHashMap
...
-      hash_map_[hash] = {attr, std::move(aggr)};
+      hash_map_[std::move(attr)] = std::move(aggr);
     }
   }
 
-  void Set(const MetricAttributes &attributes, std::unique_ptr<Aggregation> aggr, size_t hash)
+  void Set(const MetricAttributes &attributes, std::unique_ptr<Aggregation> aggr)
   {
-    auto it = hash_map_.find(hash);
+    auto it = hash_map_.find(attributes);
```
lalitb, line 161: This method can be simplified:
```cpp
  Set(MetricAttributes{attributes}, std::move(aggr));

```

sdk/test/metrics/attributes_hashmap_test.cc
```diff
@@ -84,49 +81,57 @@ TEST(AttributesHashMap, BasicTests)
...
-  // Expect the hashes to be the same because the content is the same
-  EXPECT_EQ(hash1, hash2);
-  // Expect the hashes to be the same because the content is the same
-  EXPECT_EQ(hash1, hash3);
+  }
+  EXPECT_EQ(total_active_buckets, 1);
+  EXPECT_EQ(total_elements, 4);
+}
+
+std::string make_unique_string(const char *str)
```
lalitb, line 132: nit -Do we need this method?

lalitb (approved):
LGTM. Thanks @ThomsonTan

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
ThomsonTan: The suggestion always createa a copy, but the current overload for `const MetricAttributes&` will not create a copy if the `hash_map_.find()` succeeds. Let me know if anything missed here.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
ThomsonTan: Also want to avoid creatign the temp var if the `find()` does succeed.

sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h
```diff
@@ -71,108 +76,117 @@ class AttributesHashMap
    */
   Aggregation *GetOrSetDefault(const opentelemetry::common::KeyValueIterable &attributes,
                                const AttributesProcessor *attributes_processor,
-                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback,
-                               size_t hash)
+                               std::function<std::unique_ptr<Aggregation>()> aggregation_callback)
   {
-    auto it = hash_map_.find(hash);
+    // TODO: avoid constructing MetricAttributes from KeyValueIterable for
```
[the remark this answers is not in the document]
ThomsonTan: The same as the below comment, want to avoid creating temporary when the `find()` does succeed.

https://github.com/open-telemetry/opentelemetry-cpp/pull/3322#discussion_r2023502591

marcalff:
@lalitb @ThomsonTan Assuming CI is clean, ok to merge ?

ThomsonTan:
> @lalitb @ThomsonTan Assuming CI is clean, ok to merge ?

Yes, it is Ok to merge it if CI is clean again.

PR #3330, opened 2025-03-28
[CI] Upgrade CI to ubuntu 22.04

### LINKED ISSUE #3329: Upgrade CI, Ubuntu 20.04 unsupported after 2025-04-01
Ubuntu 20.04 is no longer supported, CI needs to upgrade:

* https://github.com/actions/runner-images/issues/11101

marcalff:
Fixes #3329

## Changes

* Upgrade CI from ubuntu 20.04 to ubuntu 22.04
* Disable CI tests using abseil, to be revisited after PR #3318

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,23 +301,21 @@ jobs:
       run: |
         ./ci/do_ci.sh cmake.with_async_export.test
 
-  cmake_abseil_stl_test:
-    name: CMake test (with abseil)
-    runs-on: ubuntu-22.04
-    env:
-      CXX_STANDARD: '14'
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        submodules: 'recursive'
-    - name: setup
-      run: |
-        sudo -E ./ci/setup_googletest.sh
-        sudo -E ./ci/setup_ci_environment.sh
-    - name: run cmake tests (enable abseil-cpp)
-      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -303,7 +303,7 @@ jobs:
 
   cmake_abseil_stl_test:
     name: CMake test (with abseil)
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
```
dbarker, line 307: @marcalff unlike the `do_ci.sh` script, the `install_abseil.sh` script does not set a default `CMAKE_CXX_STANDARD=14` cmake arg. Try adding the `CXX_STANDARD=14` environment variable to the jobs that are installing abseil and failing. By default on 22.04 abseil will be building in cxx17 and opentelemetry-cpp in cxx14 if no CXX_STANDARD env var is set.

marcalff: Thanks for the suggestion. I tried, but it still fails. I temporarily disabled the 2 offending CI instead.

dbarker: Disabling sounds reasonable. I had missed that the script was called with sudo but also needed the `-E` to preserve the environment variables.

marcalff: Ah, this can explain it. Thanks for the investigations.

@dbarker A reminder, please file a PR to become an approver

owent (approved):

PR #3274, opened 2025-02-13
[BUILD] Fixes compatibility of type_traits

owent:
Fixes #3273 

## Changes

+ Make type traits the same when `OPENTELEMETRY_STL_VERSION>=2011` or not
+ Do not use auto as parameter of lambda to support legacy toolchain.

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [x] Changes in public API reviewed

PR #3337, opened 2025-04-02
Fix defining OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS

georgthegreat:
Fixes # (issue)

## Changes

Upon updating to 1.20.0 we have encountered a bunch of the following errors:
```
contrib/libs/opentelemetry-cpp/api/include/opentelemetry/nostd/type_traits.h:145:73: error: builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Werror,-Wdeprecated-builtins]
  145 |   static constexpr bool value = std::is_copy_constructible<T>::value && __has_trivial_copy(T);
```

(we use clang-18 to build our code).
It turned out that some old compatibility code got broken.

@marcalff, according to README.md opentelemetry-cpp does no longer support standards below C++11, hence we can always use these type_traits from the standard library.
Can we drop the C++03 compat and simplify the things even further?


marcalff:
@georgthegreat Thanks for the PR.

Concerning the issue, we also use clang-18 in our CI, and this never broke this way.
I really would like to understand the root cause of this, could you clarify what are the values for:

* the C++ standard version used at compile time (C++11, 14, 17, ...)
* The build system used: CMake, bazel, plain makefiles, ...
* If this breaks with compiling opentelemetry-cpp itself, or when compiling an application using it
* If OPENTELEMETRY_STL_VERSION is defined in the build, and to what value
* If OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS is defined (I assume it was not, please confirm)

Concerning the fix, I agree it should be simplified.
I will add review comments to make this even simpler.

api/include/opentelemetry/config.h
```diff
@@ -9,6 +9,6 @@
 #  define OPENTELEMETRY_HAS_INCLUDE(x) __has_include(x)
 #endif
 
-#if !defined(__GLIBCXX__) || OPENTELEMETRY_HAS_INCLUDE(<codecvt>)  // >= libstdc++-5
+#if defined(OPENTELEMETRY_STL_VERSION) && OPENTELEMETRY_STL_VERSION > 2011
```
marcalff, line 12: The test should probably be `>= 2011`.

Now, given that `OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS` is used in exactly one place, and given that it relie on the C++11 std headers anyway, this can be simplified further:

* please remove the code that sets `OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS` in this config.h file
* please change the only place where it is used with `ifdef OPENTELEMETRY_HAVE_STD_TYPE_TRAITS` instead.

marcalff (changes requested):
Thanks for the PR.

See suggestion to simplify even more.

marcalff:
Seem also related to `WITH_STL`, we need to clarify whether to test `__cplus_plus` or `OPENTELEMETRY_STL_VERSION` to check for C++11.

marcalff:
Related to:

* #3274

georgthegreat:
@marcalff, we use clang-18, `-std=c++20` and `-DOPENTELEMETRY_STL_VERSION=2023` (I see the divergence here but I don't see how it could lead to the issue).

I have removed `OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS` as you suggested.

api/include/opentelemetry/config.h
georgthegreat: Did as suggested

georgthegreat:
@marcalff let's merge it.

api/include/opentelemetry/nostd/type_traits.h
```diff
@@ -133,7 +133,7 @@ using is_nothrow_swappable = detail::swappable::is_nothrow_swappable<is_swappabl
  *  std::is_trivialy_copy_assignable
  *  std::is_trivialy_move_assignable
  */
-#ifdef OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS
+#if defined(OPENTELEMETRY_STL_VERSION) && OPENTELEMETRY_STL_VERSION >= 2011
```
owent, line 136: The fallback implementation relies on GCC's API, making it exclusive to GCC. It should be updated to support all compilers.

owent (changes requested):

api/include/opentelemetry/nostd/type_traits.h
georgthegreat: This is a dead code, opentelemetry does not support cxx03 standard / libstdc++ 5.

marcalff: @georgthegreat 

In theory, yes, this is dead code.
In practice, no, this code is not totally dead now, there are historical reasons to maintain this part working.

What we need here is a fix that solves the trivial cases of building with recent compilers, using `std:: type` traits when they are available, instead of using the alternate `nostd::` implementation.

The issue is probably that we need to do this independently of `WITH_STL` : that is, test `__cplusplus` instead of `OPENTELEMETRY_STL_VERSION`.

Now, for old cases when `std::` classes are not available, like this old work around for gcc 4.8, the code needs to implement the work around as before.

georgthegreat: I hope I understood you correctly.

owent: `#if !defined(OPENTELEMETRY_HAVE_STD_TYPE_TRAITS)` can be removed now.

api/include/opentelemetry/nostd/type_traits.h
```diff
@@ -6,19 +6,9 @@
 #if defined(OPENTELEMETRY_STL_VERSION)
 #  if OPENTELEMETRY_STL_VERSION >= 2011
 #    include "opentelemetry/std/type_traits.h"
-#    define OPENTELEMETRY_HAVE_STD_TYPE_TRAITS
 #  endif
 #endif
 
-#if !defined(OPENTELEMETRY_HAVE_STD_TYPE_TRAITS)
-#  include <array>
```
owent, line 14: `#if !defined(OPENTELEMETRY_HAVE_STD_TYPE_TRAITS)` Can be removed, but these headers are still required.

georgthegreat:
I am done with it.
I definitely have more interesting work to do compared to supporting 12 year old compiler.

Issue #3334, opened 2025-04-01
CI broken by CMake upgrade in ubuntu images


marcalff:
https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20250330.1

CMake was upgraded from 3.31.6 to 4.0.0

This drops support for CMake 3.5, breaking the build.

marcalff:
Broken by:

* https://github.com/actions/runner-images/commit/6d939a3ab352a54a021dd67b071577287b6f14a5

marcalff:
Win22 is impacted as well:

* https://github.com/actions/runner-images/commit/8b1b76fa4ae5e40d263f99bd87596e68fc6ae466

marcalff:
> The CMake 4.0 upgrade was rolled back on github runners and they now pin 3.31.x
> 
>     * [[Ubuntu] Pin CMake Version to 3.31.6 to fix Compatibility Issues in 4.0 actions/runner-images#11933](https://github.com/actions/runner-images/pull/11933)
> 
>     * [[windows] Pin the cmake version to 3.31.6 actions/runner-images#11936](https://github.com/actions/runner-images/pull/11936)
> 
>     * [[macOS] Pin Cmake to 3.31.6 due to a backward compatibility issue in 4.0 actions/runner-images#11940](https://github.com/actions/runner-images/pull/11940)
> 
> 
> [@marcalff](https://github.com/marcalff) we can likely revert the changes to the `ci.yml` workflow from [#3336](https://github.com/open-telemetry/opentelemetry-cpp/pull/3336) now. Let me know if that seems like the right next step. I can submit the PR.

@dbarker Thanks for the follow up.

Please do a partial revert then, this should speed up CI, and avoid downloading binaries from elsewhere during the build.

See #3480

marcalff:
Now resolved, closing

PR #3349, opened 2025-04-06
[CMAKE] bump cmake minimum required version to 3.14

### LINKED ISSUE #3348: Bump cmake min version to 3.14
The current cmake minimum version is 3.10. However it is not possible to use this version to build opentelemetry-cpp as some of the cmake files, ci scripts, and dependencies rely on newer versions. 

There is no test for the minimum version in ci currently. 

Bump the version minimum to 3.14, update cmake files/scripts to support that version, and add a test in ci. 

Bump to cmake minimum version 3.16 but doing this after the new versions of googletest, protobuf, and grpc are released seems reasonable. 

**Additional context**
Current dependencies rely on newer versions of cmake already: 
- googletest requires cmake version 3.13 as of [gtest v1.14](https://github.com/google/googletest/blob/f8d7d77c06936315286eb55f8de22cd23c188571/CMakeLists.txt#L4) 
- prometheus-cpp requires cmake version 3.14 [since v1.0.0](https://github.com/jupp0r/prometheus-cpp/blob/4ea303fa66e4c26dc4df67045fa0edf09c2f3077/CMakeLists.txt#L2)

Google has bumped the min version to 3.16 on the main branches of googletest, protobuf, and grpc according to their [cxx-support-matrix](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md).

dbarker:
Fixes #3348 

## Changes

- updates setup_cmake* scripts to verify the cmake version installed
- bumps the cmake min version to 3.14 and fixes issues found when testing with that version 
- updates the cmake install test github workflow to test specific cmake versions including the min version 


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.14)
```
dbarker, line 4: Rationale for bumping to 3.14
 - prometheus-cpp 1.0.0 requires cmake 3.14
 - googletest 1.14 requires cmake 3.13
 - Marc's comment on the last version bump implies 3.14 is okay for his project - https://github.com/open-telemetry/opentelemetry-cpp/issues/3166#issuecomment-2499138533
 
caveats
- The otlp file exporter tests fail on windows unless built with cmake 3.15 see [ExportToFileSystemRotateIndexTest and ExportToFileSystemRotateByTimeTest failures](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/14295690169/job/40062540599#step:4:2575)
- The  `CMAKE_FIND_PACKAGE_PREFER_CONFIG` option comes with 3.15 so changes in this file are required to properly find protobuf 3.22.0+ with cmake 3.14.

.github/workflows/cmake_install.yml
```diff
@@ -26,10 +26,11 @@ jobs:
       - name: Run DLL Tests
         run: ./ci/do_ci.ps1 cmake.dll.install.test
 
-  windows_2019_vcpkg:
-    name: Windows 2019 vcpkg cxx14 (static libs)
+  windows_2019_vcpkg_submodule_min_cmake:
+    name: Windows 2019 vcpkg submodule versions minimum cmake cxx14 (static libs)
     runs-on: windows-2019
     env:
+      CMAKE_VERSION: '3.15.0'
```
dbarker, line 33: Windows tests only pass with cmake 3.15.0+ currently. This is likely due to the addition of policy [CMP0091]( https://cmake.org/cmake/help/latest/policy/CMP0091.html#policy:CMP0091) in cmake 3.15

.github/workflows/cmake_install.yml
```diff
@@ -135,12 +138,17 @@ jobs:
...
       run: ./ci/do_ci.sh cmake.install.test
 
-  ubuntu_2204_script_build_grpc_1_49_2:
-    name: Ubuntu 22.04 script grpc 1.49.2 cxx14 (static libs - shared libs)
+  ubuntu_2204_minimum:
+    name: Ubuntu 22.04 minimum versions cxx14 (static libs - shared libs)
     runs-on: ubuntu-22.04
     env:
       INSTALL_TEST_DIR: '/home/runner/install_test'
+      CMAKE_VERSION: '3.14.0'
```
dbarker, line 146: All tests pass in Ubuntu 22.02 with cmake 3.14

CMakeLists.txt
```diff
@@ -420,9 +421,18 @@ if(WITH_OTLP_GRPC
    OR WITH_OTLP_HTTP
    OR WITH_OTLP_FILE)
 
-  find_package(Protobuf)
-  # Protobuf 3.22 or upper require abseil-cpp, we can find it in
-  # opentelemetry-cpp-config.cmake
+  # Note: remove once cmake 3.14 is no longer supported including the
+  # CMakeFindDependencyMacro is required with cmake 3.14 due to an issue with
+  # the gRPCConfig.cmake file.
+  include(CMakeFindDependencyMacro)
```
marcalff, line 427: Is this a CMake issue with 3.14, or a gRPC issue ?

I found this, which looks related:

* https://github.com/grpc/grpc/pull/33361

In my understanding, this include should stay, as a work around for find_package(gRPC), even with cmake >= 3.15.

Please clarify / update the comments / provide a link to the relevant issue accordingly.

marcalff (approved):
LGTM, see minor comments.

CMakeLists.txt
```diff
@@ -420,9 +421,18 @@ if(WITH_OTLP_GRPC
...
+
+  # Protobuf 3.22 and up requires abseil-cpp. This is found using the CONFIG
+  # search mode. In cmake 3.14 we must manually try config mode first then fall
+  # back to MODULE mode
+  find_package(Protobuf CONFIG)
```
owent, line 432: Why not use `set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)` to prioritize finding CONFIG packages?

CMakeLists.txt
```diff
@@ -800,10 +810,10 @@ if(gRPC_FOUND)
   message(STATUS "gRPC: ${gRPC_VERSION}")
 endif()
 if(CURL_FOUND)
-  message(STATUS "CURL: ${CURL_VERSION}")
+  message(STATUS "CURL: ${CURL_VERSION_STRING}")
```
owent, line 813: `CURL_VERSION_STRING` is superseded by `CURL_VERSION` after cmake 4.0.

ci/setup_grpc.sh
```diff
@@ -144,8 +146,8 @@ echo "Building gRPC ${install_grpc_version}"
 echo "CMake build options:" "${GRPC_BUILD_OPTIONS[@]}"
 
 cmake "${GRPC_BUILD_OPTIONS[@]}" ..
-cmake --build . -j$(nproc)
-cmake --install .
+make -j $(nproc)
```
owent, line 149: Why use `make -j` here? From my understanding, `cmake --build . -j` offers better compatibility, supporting Ninja and other generators.

`cmake --install .` also offers better compatibility than `make install`.

install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_executable(exporters_elasticsearch_test
                ${INSTALL_TEST_SRC_DIR}/test_exporters_elasticsearch.cc)
 target_link_libraries(
   exporters_elasticsearch_test
-  PRIVATE opentelemetry-cpp::elasticsearch_log_record_exporter GTest::gtest
-          GTest::gtest_main)
+  PRIVATE opentelemetry-cpp::elasticsearch_log_record_exporter GTest::GTest
```
owent, line 20: The CONFIG package of gtest only offers `GTest::gtest` and `GTest::gtest_main`. So I think them should not be replaced by old target names.

owent (changes requested):

TheMenisFactor:
I don't understand what this means

CMakeLists.txt
dbarker: CMAKE_FIND_PACKAGE_PREFER_CONFIG is still called. Shared more in this comment 
https://github.com/open-telemetry/opentelemetry-cpp/pull/3349/files#r2031566633

CMakeLists.txt
dbarker: > Is this a CMake issue with 3.14, or a gRPC issue ?
I'm not sure. The tests passed with cmake 3.16 without adding this line on the same version of grpc. I don't think keeping this include causes problems  - I'll updated the comment.

CMakeLists.txt
dbarker: I've pushed a changes to check if CURL_VERSION is defined and set it to CURL_VERSION_STRING to keep the version variable consistent on versions cmake 3.14+

CMakeLists.txt
```diff
@@ -800,10 +810,10 @@ if(gRPC_FOUND)
...
 endif()
 if(ZLIB_FOUND)
-  message(STATUS "ZLIB: ${ZLIB_VERSION}")
+  message(STATUS "ZLIB: ${ZLIB_VERSION_STRING}")
```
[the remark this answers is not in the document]
dbarker: I've pushed a changes to check if ZLIB_VERSION is defined and set it to ZLIB_VERSION_STRING to keep the version variable consistent on cmake 3.14+

ci/setup_cmake.ps1
```diff
@@ -12,4 +12,37 @@ choco uninstall cmake cmake.install -y --remove-dependencies --skip-autouninstal
...
+    exit 1
+}
+if ($ctestVersion -ne $CMAKE_VERSION) {
+    Write-Error "CTest version mismatch: expected $CMAKE_VERSION, installed $ctestVersion"
+    exit 1
+}
+if ($cpackVersion -ne $CMAKE_VERSION) {
+    Write-Error "CPack version mismatch: expected $CMAKE_VERSION, installed $cpackVersion"
+    exit 1
+}
```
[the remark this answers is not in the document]
marcalff: Actually, the format is ok: `*.ps1` file, runs on Windows, uses windows file format for CRLF.

.github/workflows/cmake_install.yml
```diff
@@ -101,12 +103,17 @@ jobs:
         BUILD_SHARED_LIBS: 'OFF'
       run: ./ci/do_ci.sh cmake.install.test
 
-  ubuntu_2204_script_build_grpc_1_55_0:
-    name: Ubuntu 22.04 script grpc 1.55.0 cxx17 (static libs - shared libs)
+  ubuntu_2204_stable:
```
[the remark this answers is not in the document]
dbarker: The `setup_cmake*` scripts default to cmake 3.31.6 now. The scripts in the main ci.yml workflow also use the default. 

I'm not apposed to setting CMAKE_VERSION in every job explicitly if others feel strongly.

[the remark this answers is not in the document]
dbarker: The cmake version can be found in the ci logs at two places: 

[In the install step](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/14316461203/job/40123633379#step:3:21)
```bash
Verifying installed versions...
cmake version: 3.31.6 detected
ctest version: 3.31.6 detected
cpack version: 3.31.6 detected
```
[Run test step](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/14316461203/job/40123633379#step:4:119)
```bash
-- ---------------------------------------------
-- versions
-- ---------------------------------------------
-- CMake: 3.31.6
-- GTest: 1.16.0
-- benchmark: 1.9.2
-- Abseil: 20240722
-- Protobuf: 29.3.0
-- gRPC: 1.71.0
-- CURL: 8.6.0
-- ZLIB: 1.2.12
-- nlohmann-json: 3.11.3
-- prometheus-cpp: 1.3.0
-- ---------------------------------------------
```

.github/workflows/cmake_install.yml
dbarker: Added a comment

install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
owent: If we use CONFIG pacakge  of gtest, the GTest::GTest do not exists,even with cmake below 3.20, we have building failures before and can use ${GTEST_BOTH_LIBRARIES} here.

dbarker: > If we use CONFIG pacakge of gtest, the GTest::GTest do not exists,even with cmake below 3.20, we have building failures before and can use ${GTEST_BOTH_LIBRARIES} here.

This PR won't impact the main unit tests as their targets and method of finding gtest remain unchanged. 

In this PR, the module search for gtest and legacy targets are passing with: 

  **cmake:**  3.14, 3.15, 3.22, 3.28, and 3.31
  **gtest:**  1.11, 1.14, and 1.16
  **packages from:** apt, brew, conan, vcpkg and our custom build scripts

I'm not sure if the gtest cmake config file is installed in all those cases. The apt packages for Protobuf don't install the cmake config file for example. 

If you have a build configuration that would fail with the current tests I'd be happy to add it. I'd like to cover those types of cases.

owent: I have reproduced it by building gtest from source, without any patch, and using CONFIG package only.
According to `https://github.com/google/googletest/tree/main/googletest#incorporating-into-an-existing-cmake-project`, it also uses `GTest::gtest`.
Here is the `GTestTargets-release.cmake` in my installed gtest CONFIG package:

```cmake
#----------------------------------------------------------------
# Generated CMake target import file for configuration "Release".
#----------------------------------------------------------------

# Commands may need to know the format version.
set(CMAKE_IMPORT_FILE_VERSION 1)

# Import target "GTest::gtest" for configuration "Release"
set_property(TARGET GTest::gtest APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(GTest::gtest PROPERTIES
  IMPORTED_IMPLIB_RELEASE "${_IMPORT_PREFIX}/lib/gtest-rel.lib"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/bin/gtest-rel.dll"
  )

list(APPEND _cmake_import_check_targets GTest::gtest )
list(APPEND _cmake_import_check_files_for_GTest::gtest "${_IMPORT_PREFIX}/lib/gtest-rel.lib" "${_IMPORT_PREFIX}/bin/gtest-rel.dll" )

# Import target "GTest::gtest_main" for configuration "Release"
set_property(TARGET GTest::gtest_main APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(GTest::gtest_main PROPERTIES
  IMPORTED_IMPLIB_RELEASE "${_IMPORT_PREFIX}/lib/gtest_main-rel.lib"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/bin/gtest_main-rel.dll"
  )

list(APPEND _cmake_import_check_targets GTest::gtest_main )
list(APPEND _cmake_import_check_files_for_GTest::gtest_main "${_IMPORT_PREFIX}/lib/gtest_main-rel.lib" "${_IMPORT_PREFIX}/bin/gtest_main-rel.dll" )

# Import target "GTest::gmock" for configuration "Release"
set_property(TARGET GTest::gmock APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(GTest::gmock PROPERTIES
  IMPORTED_IMPLIB_RELEASE "${_IMPORT_PREFIX}/lib/gmock-rel.lib"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/bin/gmock-rel.dll"
  )

list(APPEND _cmake_import_check_targets GTest::gmock )
list(APPEND _cmake_import_check_files_for_GTest::gmock "${_IMPORT_PREFIX}/lib/gmock-rel.lib" "${_IMPORT_PREFIX}/bin/gmock-rel.dll" )

# Import target "GTest::gmock_main" for configuration "Release"
set_property(TARGET GTest::gmock_main APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(GTest::gmock_main PROPERTIES
  IMPORTED_IMPLIB_RELEASE "${_IMPORT_PREFIX}/lib/gmock_main-rel.lib"
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/bin/gmock_main-rel.dll"
  )

list(APPEND _cmake_import_check_targets GTest::gmock_main )
list(APPEND _cmake_import_check_files_for_GTest::gmock_main "${_IMPORT_PREFIX}/lib/gmock_main-rel.lib" "${_IMPORT_PREFIX}/bin/gmock_main-rel.dll" )

# Commands beyond this point should not need to know the version.
set(CMAKE_IMPORT_FILE_VERSION)
```

.github/workflows/cmake_install.yml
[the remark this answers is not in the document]
psx95: I don't have a strong opinion on whether the version should be explicitly specified either, but if the version is only being specified on certain jobs due to some technical constraints (*like CI fails if we the default cmake version is used*), I do think it is worthwhile to document why the version is being explicitly mentioned on those jobs.

This would be helpful in the future if and when we re-visit this file to upgrade or modify the jobs.

I'm not an expert on CMake, so I'll defer the final decision to you though 👍🏻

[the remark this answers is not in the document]
dbarker: Thanks for the feedback. I've updated the cmake install test workflow to set the version explicitly for all jobs. 

> but if the version is only being specified on certain jobs due to some technical constraints (like CI fails if we the default cmake version is used), I do think it is worthwhile to document why the version is being explicitly mentioned on those jobs.

Great point. The only cases I'm aware of where ci fails based on the cmake version is windows with 3.14 (comment added to the CMake Install Test github workflow) and with cmake 4.0.0 (documented in #3334).

install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
dbarker: > I'm not sure why vcpks works with CMAKE_FIND_PACKAGE_PREFER_CONFIG=TRUE.

This may makes sense. There is special handling of windows in the main cmake file and it will call find_package(GTest). The FindGTest module will use the config search internally with cmake 3.20+. 

Interestingly...The `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is patched by the community opentelemetry-cpp portfile to be off. See https://github.com/microsoft/vcpkg/blob/master/ports/opentelemetry-cpp/cmake-quirks.diff

I'd expect that patch to be causing problems for the vcpkg portfile users if building with modern protobuf 3.22+. This PR should help by still doing the config search before using the legacy protobuf module.

owent (approved):
LGTM now, thanks.

.github/workflows/cmake_install.yml
```diff
@@ -8,10 +8,11 @@ on:
     branches: [ main ]
 
 jobs:
-  windows_2022_vcpkg:
-    name: Windows 2022 vcpkg cxx17 (static libs - dll)
+  windows_2022_vcpkg_submodule:
+    name: Windows 2022 vcpkg submodule versions cxx17 (static libs - dll)
     runs-on: windows-2022
     env:
+      CMAKE_VERSION: '3.31.6'
```
lalitb, line 15: Why do we need cmake 3.31.6 here?

dbarker: It is not needed. The default cmake version set by the `setup_cmake.ps1` script is 3.31.6. Do you think it be set to a different version or just implicitly use the default?

lalitb: Thanks for the context. I think we can let it be here, just to be explicit.

.github/workflows/cmake_install.yml
```diff
@@ -47,6 +50,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       INSTALL_TEST_DIR: '/home/runner/install_test'
+      CMAKE_VERSION: '3.28.3'
```
lalitb, line 53: Sorry if I am out of context here. Why do we have 3.28.3 here, rather than default 3.31.6 used in setup_cmake.sh?

.github/workflows/cmake_install.yml
```diff
@@ -71,12 +75,17 @@ jobs:
...
+    name: Ubuntu 24.04 latest versions cxx20 (static libs)
     runs-on: ubuntu-24.04
     env:
       INSTALL_TEST_DIR: '/home/runner/install_test'
+      CMAKE_VERSION: '3.31.6'
       CXX_STANDARD: '20'
+      GOOGLETEST_VERSION: '1.16.0'
+      ABSEIL_CPP_VERSION: '20240722.1'
+      PROTOBUF_VERSION: '29.0'
+      GRPC_VERSION: 'v1.71.0'
```
lalitb, line 88: Can we add comment on rational behind using these versions - as they are different from what is specified here - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/third_party_release

.github/workflows/cmake_install.yml
dbarker: My intent was to make sure we are testing the currently packaged versions of cmake with [Ubuntu 24.04 - cmake 3.28](https://launchpad.net/ubuntu/noble/+package/cmake) and [Ubuntu 22.04 - cmake 3.22](https://launchpad.net/ubuntu/jammy/+source/cmake). My expectation is that many users on Ubuntu will start with `apt install cmake`.

CMakeLists.txt
```diff
@@ -420,15 +421,31 @@ if(WITH_OTLP_GRPC
...
 
-  find_package(Protobuf)
-  # Protobuf 3.22 or upper require abseil-cpp, we can find it in
-  # opentelemetry-cpp-config.cmake
+  # Including the CMakeFindDependencyMacro resolves an error from
+  # gRPCConfig.cmake on some grpc versions. See
+  # https://github.com/grpc/grpc/pull/33361 for more details.
+  include(CMakeFindDependencyMacro)
+
+  # Protobuf 3.22+ and depends on abseil-cpp and must be found using the CONFIG
```
ThomsonTan, line 429:
```suggestion
  # Protobuf 3.22+ depends on abseil-cpp and must be found using the CONFIG
```

.github/workflows/cmake_install.yml
dbarker: Will do. These are just the latest versions available (with the protobuf and abseil version picked from the submodules on the grpc tag).

CMakeLists.txt
ThomsonTan: And what does it mean for "must be found using the CONFIG search mode"? Seems module mode is also tried.

.github/workflows/cmake_install.yml
dbarker: Added additional comments on the versions.

CMakeLists.txt
dbarker: Updated the comment to be clear this is the cmake find_package search mode.

ci/install_protobuf.sh
```diff
@@ -70,7 +70,7 @@ if [ -e "../CMakeLists.txt" ]; then
 else
   cmake ../cmake "${CPP_PROTOBUF_BUILD_OPTIONS[@]}"
 fi
-cmake --build . -j $(nproc)
-cmake --install .
+make -j $(nproc)
```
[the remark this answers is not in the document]
lalitb: ok please ignore, just saw the comment here - https://github.com/open-telemetry/opentelemetry-cpp/pull/3349/files#r2031479219. Probably keep a TODO here with the reason for using make, so that someone can fix later.

lalitb (approved):

CMakeLists.txt
```diff
@@ -36,8 +36,9 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
-# Prefer cmake CONFIG to auto resolve dependencies. This is important to
-# properly find protobuf versions 3.22.0 and above
+# Note: CMAKE_FIND_PACKAGE_PREFER_CONFIG requires cmake 3.15. Prefer cmake
```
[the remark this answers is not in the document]
dbarker: Agreed. 3.16 would be best but I wanted to make the smallest bump possible for now. My true motive was to add the tests with lower versions of cmake to make sure my recent changes didn't affect anyone :)

.github/workflows/cmake_install.yml
```diff
@@ -236,10 +264,11 @@ jobs:
           ./ci/verify_packages.sh
 
   macos_14_conan_stable:
-    name: macOS 14 conan stable cxx17 (static libs)
+    name: macOS 14 conan stable versions cxx17 (static libs)
     runs-on: macos-14
     env:
       INSTALL_TEST_DIR: '/Users/runner/install_test'
+      CMAKE_VERSION: '3.28.3'
```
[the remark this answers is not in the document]
dbarker: I honestly don't know what the default is. The recipe has [is 4.0.0](https://formulae.brew.sh/formula/cmake) which would fail to build grpc. CMake 3.28.0 here is set to be consistent with the configuration for the Ubuntu 24.04 stable job.

INSTALL.md
```diff
@@ -23,7 +23,9 @@ You can link OpenTelemetry C++ SDK with libraries provided in
   repository. To install Git, consult the [Set up
   Git](https://help.github.com/articles/set-up-git/) guide on GitHub.
 - [CMake](https://cmake.org/) for building opentelemetry-cpp API, SDK with their
-  unittests. We use CMake version 3.15.2 in our build system. To install CMake,
+  unittests. The minimum CMake version is 3.14.
+  Cmake 3.15+ is recommended on Windows due to known ci test failures with 3.14.
```
ThomsonTan, line 27:
```suggestion
  CMake 3.15+ is recommended on Windows due to known ci test failures with 

> 3.14.

```

install/conan/conanfile_stable.txt
```diff
@@ -30,7 +30,7 @@ pcre2/*:with_bzip2=False
 
 [test_requires]
 gtest/1.14.0
-benchmark/1.9.1
+benchmark/1.8.3
```
ThomsonTan, line 33: Same as above.

INSTALL.md
dbarker: Why add the new lines and quote here?

install/conan/conanfile_stable.txt
dbarker: Benchmark 1.8.3 is the version that gets installed on Ubuntu 24.04 and is primarily used in the ci.yml workflow. 

https://launchpad.net/ubuntu/noble/+source/benchmark

INSTALL.md
ThomsonTan: Removed the new line which seems inserted by accident.

ci/install_protobuf.sh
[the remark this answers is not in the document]
dbarker: Looking at those scripts they were never setting a different cmake generator and appear to always have defaulted to make. I'd lean towards adding a github feature ticket instead of todo comments, but will defer to you @lalitb .

install/conan/conanfile_stable.txt
```diff
@@ -8,7 +8,7 @@ abseil/20230125.3
 protobuf/3.21.12
 grpc/1.54.3
 nlohmann_json/3.10.5
-prometheus-cpp/1.3.0
+prometheus-cpp/1.2.4
```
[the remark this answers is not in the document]
dbarker: > Why is version downgraded here?
The conan based install tests have a "latest" and "stable" conanfile in `install/conan`. The conan file for "latest" includes  prometheus-cpp/1.3.0. I downgraded prometheus-cpp version in the "stable" conan file to ensure the previous version was also tested.

PR #3355, opened 2025-04-09
Fixes glibc++ 5 checking

### LINKED ISSUE #3354: 1.20 no longer builds due to deprecated stuff being activated (solution suggestion included)
The problem is as follows (compiler output from clang, but similar for gcc), building with -std=c++23. And otel configured to use this standard.

[ 23%] Building CXX object _deps/opentelemetry-cpp-build/sdk/src/trace/CMakeFiles/opentelemetry_trace.dir/tracer_context.cc.o
In file included from /home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/sdk/src/trace/tracer_context.cc:9:
In file included from /home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h:7:
In file included from /home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/sdk/include/opentelemetry/sdk/instrumentationscope/instrumentation_scope.h:9:
In file included from /home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/api/include/opentelemetry/common/key_value_iterable_view.h:18:
/home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/api/include/opentelemetry/nostd/type_traits.h:145:73: error: builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Werror,-Wdeprecated-builtins]
  145 |   static constexpr bool value = std::is_copy_constructible<T>::value && __has_trivial_copy(T);
      |                                                                         ^
/home/ldco/Projects/supernative/build/ClangRelease/_deps/opentelemetry-cpp-src/api/include/opentelemetry/nostd/type_traits.h:157:70: error: builtin __has_trivial_assign is deprecated; use __is_trivially_assignable instead [-Werror,-Wdeprecated-builtins]
  157 |   static constexpr bool value = std::is_copy_assignable<T>::value && __has_trivial_assign(T);
      |                                                                      ^

The root cause of this is that the file "api/include/opentelemetry/nostd/type_traits.h" has changed. In the previous versions, most of the file content was not active, due to the #if !defined(OPENTELEMETRY_HAVE_STD_TYPE_TRAITS) at the top, which only ended at the bottom of the file.
In 1.20 this is ending 5 lines further from where it starts. So much more code in this file is now active, till we reach the following section:

#ifdef OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS
using std::is_trivially_copy_assignable;
using std::is_trivially_copy_constructible;
using std::is_trivially_move_assignable;
using std::is_trivially_move_constructible;
#else
template <typename T>
struct is_trivially_copy_constructible
{
  static constexpr bool value = std::is_copy_constructible<T>::value && __has_trivial_copy(T);
};

template <typename T>
struct is_trivially_move_constructible
{
  static constexpr bool value = __is_trivial(T);
};

template <typename T>
struct is_trivially_copy_assignable
{
  static constexpr bool value = std::is_copy_assignable<T>::value && __has_trivial_assign(T);
};

template <typename T>
struct is_trivially_move_assignable
{
  static constexpr bool value = __is_trivial(T);
};
#endif

In this file, now and also in previous versions "OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS" is NOT defined. But in previous versions this did not matter for this file, since top ifdef disabled all the rest of the file.

Now who defines this, this is still the same as in the previous versions: api/include/opentelemetry/config.h
#if !defined(__GLIBCXX__) || OPENTELEMETRY_HAS_INCLUDE(<codecvt>)  // >= libstdc++-5
#  define OPENTELEMETRY_TRIVIALITY_TYPE_TRAITS
#endif

The define is active, I have checked, HOWEVER this file is not being included in type_trait.h.
If for example add the include, near the top:#include "opentelemetry/config.h"

=====> the builds is fine again.

owent:
Fixes #3354 

## Changes


api/include/opentelemetry/config.h
```diff
@@ -3,12 +3,9 @@
...
 
-#ifndef __has_include
-#  define OPENTELEMETRY_HAS_INCLUDE(x) 0
-#else
-#  define OPENTELEMETRY_HAS_INCLUDE(x) __has_include(x)
-#endif
+#include <type_traits>  // IWYU pragma: keep
 
-#if !defined(__GLIBCXX__) || OPENTELEMETRY_HAS_INCLUDE(<codecvt>)  // >= libstdc++-5
+#if !defined(__GLIBCXX__) || (defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE >= 7) || \
```
dbarker, line 8: How can we reproduce the build error in ci and verify the change?

owent: This is just for legacy toolchains, which is removed in otel-cpp's repo , but still used in community projects.

owent: Thanks to @agrawroh for testing it in #3384 .

dbarker (approved):
LGTM Thanks. Thanks for testing @agrawroh

lalitb:
The PR looks good overall!
Just one small suggestion — instead of deleting `config.h`, should we move these changes there?
`config.h` is a public header, and some users might still have a dependency on it (directly or indirectly).

ThomsonTan:
> looks good overall!
> Just one small suggestion — instead of deleting `config.h`, should we move these changes there?
> `config.h` is a public header, and some users might still have a dependency on it (directly or indirectly).

Is the original file name also kind of misleading, because the `config.h` is usually used to store the status of configuration process? If so, probably it is a good time to remove it then it will not conflict the name if we add it later.

lalitb:
> Is the original file name also kind of misleading, because the config.h is usually used to store the status of configuration process? If so, probably it is a good time to remove it then it will not conflict the name if we add it later.

I see your point about the naming — agree that `config.h` could have been better named originally.

However, since `config.h` is already part of the public API in stable releases, removing it would technically be a breaking change for downstream users. Even if it was meant to be used internally, we never documented that. I would recommend keeping `config.h` for now, even if it's mostly empty, and cleaning it up properly in separate PR as part of ABIv2.

owent:
> > Is the original file name also kind of misleading, because the config.h is usually used to store the status of configuration process? If so, probably it is a good time to remove it then it will not conflict the name if we add it later.
> 
> I see your point about the naming — agree that `config.h` could have been better named originally.
> 
> However, since `config.h` is already part of the public API in stable release, removing it would technically be a breaking change for downstream users. Even if it was meant to be used internally, we never documented that. I would recommend keeping `config.h` for now, even if it's mostly empty, and cleaning it up properly in separate PR as part of ABIv2.

Good point, I restore config.h, but only add error message for ABI v2 and warning messages for v1. How do you think about it?

lalitb (approved):
Thanks for the fix.

PR #3357, opened 2025-04-09
[API] Add Enabled method to Tracer

### LINKED ISSUE #3149: [API] Add functionality in Tracer to report if it is enabled
OpenTelemetry Specification specifies that a [Tracer **SHOULD** provide functions to report if it is enabled or not](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracer).

This spec change was added in [PR#4063 ](https://github.com/open-telemetry/opentelemetry-specification/pull/4063) and it goes well with the spec compliance changes requested in issue #2641.

psx95:
Fixes #3149 

## Changes

Adds a method to Tracer (ABI version >= 2) that allows the user to check if the tracer is enabled or not. A disabled tracer does not produce spans.

This change enhances [tracer spec-compatibility](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#tracer).

### DIFF
--- a/api/include/opentelemetry/trace/tracer.h
+++ b/api/include/opentelemetry/trace/tracer.h
@@ -163,6 +163,19 @@ public:
     }
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  /**
+   * Reports if the tracer is enabled or not. A disabled tracer will not create spans.
+   *
+   * The returned value can change over time so the instrumentation authors must call this method
+   * every time before creating a span to potentially avoid performing computationally expensive
+   * operations.
+   *
+   * @since ABI_VERSION 2
+   */
+  virtual bool Enabled() noexcept = 0;
+#endif
+
 #if OPENTELEMETRY_ABI_VERSION_NO == 1
 
   /*
--- a/api/include/opentelemetry/trace/tracer.h
+++ b/api/include/opentelemetry/trace/tracer.h
@@ -66,7 +66,7 @@ public:
   template <class T,
             class U,
             nostd::enable_if_t<common::detail::is_key_value_iterable<T>::value> * = nullptr,
-            nostd::enable_if_t<detail::is_span_context_kv_iterable<U>::value>   * = nullptr>
+            nostd::enable_if_t<detail::is_span_context_kv_iterable<U>::value> *   = nullptr>
   nostd::shared_ptr<Span> StartSpan(nostd::string_view name,
                                     const T &attributes,
                                     const U &links,
--- a/api/include/opentelemetry/plugin/tracer.h
+++ b/api/include/opentelemetry/plugin/tracer.h
@@ -104,6 +104,13 @@ public:
     return nostd::shared_ptr<trace::Span>{new (std::nothrow) Span{this->shared_from_this(), span}};
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  bool Enabled() noexcept override
+  {
+    return true;
+  }
+#endif
+
 #if OPENTELEMETRY_ABI_VERSION_NO == 1
 
   void ForceFlushWithMicroseconds(uint64_t timeout) noexcept override
--- a/api/include/opentelemetry/plugin/tracer.h
+++ b/api/include/opentelemetry/plugin/tracer.h
@@ -105,10 +105,7 @@ public:
   }
 
 #if OPENTELEMETRY_ABI_VERSION_NO >= 2
-  bool Enabled() noexcept override
-  {
-    return true;
-  }
+  bool Enabled() noexcept override { return true; }
 #endif
 
 #if OPENTELEMETRY_ABI_VERSION_NO == 1
--- a/api/include/opentelemetry/trace/noop.h
+++ b/api/include/opentelemetry/trace/noop.h
@@ -108,6 +108,13 @@ public:
     return noop_span;
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  bool Enabled() noexcept override
+  {
+    return false;
+  }
+#endif
+
 #if OPENTELEMETRY_ABI_VERSION_NO == 1
 
   void ForceFlushWithMicroseconds(uint64_t /*timeout*/) noexcept override {}
--- a/api/include/opentelemetry/trace/noop.h
+++ b/api/include/opentelemetry/trace/noop.h
@@ -109,10 +109,7 @@ public:
   }
 
 #if OPENTELEMETRY_ABI_VERSION_NO >= 2
-  bool Enabled() noexcept override
-  {
-    return false;
-  }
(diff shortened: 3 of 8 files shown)


marcalff (approved):
LGTM, thanks for the fix.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -163,6 +163,18 @@ class Tracer
     }
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  /**
+   * Reports if the tracer is enabled or not. A disabled tracer will not create spans.
+   *
+   * The instrumentation authors must call this method very time before creating a span to
```
ThomsonTan, line 170:
```suggestion
   * The instrumentation authors must call this method every time before creating a span to
```

dbarker, line 170: nit: "very" -> "every"

api/include/opentelemetry/trace/tracer.h
```diff
@@ -163,6 +163,18 @@ class Tracer
...
+   * potentially avoid performing computationally expensive operations for disabled tracers.
+   *
+   * @since ABI_VERSION 2
+   */
+  virtual bool Enabled() noexcept = 0;
```
dbarker, line 175: Q: should this be const?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -590,6 +590,20 @@ class Tracer : public opentelemetry::trace::Tracer,
     return result;
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  /**
+   * Reports if the tracer is enabled or not. A disabled tracer will not create spans.
+   * Note: The etw_tracer currently does not accept a TracerConfig and can therefore not be disabled
+   * based on the instrumentation scope.
+   *
+   * The instrumentation authors must call this method very time before creating a span to
```
ThomsonTan, line 599:
```suggestion
   * The instrumentation authors must call this method every time before creating a span to
```

api/include/opentelemetry/trace/tracer.h
lalitb, line 175: Just wondering if need to do anything specific for this - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#enabled

-- clip
 Parameters can be added in the future, therefore, the API MUST be structured in a way for parameters to be added.


Adding new overload or modifying this method in future would break the ABIv2 compatibility, unless we are fine with that. cc @marcalff 
Else, one option could be to provide an overloaded method which accepts Context (similar to Logger::Enabled), which would provide a mechanism for passing additional parameters in the future.

But I am even fine to break the ABIv2 compatibility, as this API is still not stable in the specs.

api/include/opentelemetry/trace/tracer.h
lalitb: I don't think this constraint is mandatory according to the specifications. Ideally, instrumentation authors should have the flexibility to decide when to call the Enabled() method - particularly in cases where creating a span would require expensive pre-computation operations.

marcalff: Good catch, this is not mandatory, only a helper for expensive operations.

@psx95 Please reword.

api/include/opentelemetry/trace/tracer.h
marcalff: Re:

>  Parameters can be added in the future, therefore, the API MUST be structured in a way for parameters to be added.

This is wishful thinking from the spec, but ignores the reality of C++ constraints.

Both the spec for Enabled() and the ABI_VERSION_2 are not stable, this can be changed any time, and will be just fine.

Ok to merge as is (with a const)

lalitb (approved):

api/include/opentelemetry/trace/tracer.h
ThomsonTan: Also if the purpose is avoid heavy computation for the user to call this `Enabled()`, I think the call to `Enabled()` involves virtual dispatch which seems not a light operation. 

Could we save the enabled in the API class so enabled check just query this class member? And also provide API for SDK to set it?

lalitb: +1 I think @ThomsonTan implemented something similar for Log SDK, good to refer that too.

psx95: I think I misunderstood [the spec here](https://github.com/open-telemetry/opentelemetry-specification/blame/8bbac83c0c994c0a750cc688935fc1cae7fd39b3/specification/trace/api.md#L216), I will reword it, I agree instrumentation author should have flexibility on when to call this 👍🏻

dbarker (approved):
LGTM. Thanks

api/include/opentelemetry/trace/tracer.h
psx95: Regarding avoiding the virtual dispatch call: I'm not sure if the associated overhead is huge to move implementation to the API (I might be wrong here, so please let me know). 

Unlike the Logger, in which the implementation was completely moved to the API in a self-sufficient way, it might be difficult to provide an accurate implementation in API itself. 
As @ThomsonTan mentioned, we would need to add a protected method in the API which the SDKs can call to update the value of the saved variable. 

So we won't have a way to force implementations of the API to call the `UpdateEnabled(bool enabled)` method. If the SDK implementations forget to call this method, the API might return incorrect state of the tracers.
Its a minor concern, bur are we ok with this approach ?

psx95: @ThomsonTan I have implemented your suggestion in [this commit](https://github.com/open-telemetry/opentelemetry-cpp/pull/3357/commits/8037eaff29376f9442a0a62322501535d99ea771). 

 - Pros: It avoids the virtual methods that lead to dynamic dispatch calls.
 - Cons: It forces the SDK implementations to explicitly call `UpdateEnabled(bool)` to keep status up-to-date.

On second look, it seems like a good trade-off to me. Let me know your thoughts on this.

api/include/opentelemetry/trace/tracer.h
```diff
@@ -209,6 +209,30 @@ class Tracer
...
+   *
+   * This method should be used by SDK implementations to indicate the tracer's updated state
+   * whenever a tracer transitions from enabled to disabled state and vice versa.
+   *
+   * @param enabled The new state of the tracer. False would indicate that the tracer is no longer
+   * enabled and will not produce as
+   *
+   * @since ABI_VERSION 2
+   */
+  void UpdateEnabled(const bool enabled) noexcept { this->enabled_ = enabled; }
```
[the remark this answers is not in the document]
psx95: Ah, good point. I didn't consider multi-threading.

ThomsonTan (approved):

PR #3358, opened 2025-04-09
[SDK] support aggregation of identical instruments

### LINKED ISSUE #3356: identical instruments are not aggregated by the SDK
**Describe your environment**
main at 4d52ab2c4d89

**Steps to reproduce**
- Create a meter
- Create a counter "my_counter"
- Create a second instance of "my_counter"
- Add 1 to both with different attributes 

Example test (added to meter_test.cc): 

```cpp
TEST(MeterTest, IdenticalInstrumentTests)
{
  MetricReader *metric_reader_ptr = nullptr;
  auto meter                      = InitMeter(&metric_reader_ptr);
  auto counter1         = meter->CreateUInt64Counter("my_counter");
  auto counter2         = meter->CreateUInt64Counter("my_counter");

  counter1->Add(1, {{"key", "value1"}});
  counter2->Add(1, {{"key", "value2"}});
   
  size_t count = 0;
  metric_reader_ptr->Collect([&count](ResourceMetrics &metric_data) {
    EXPECT_EQ(metric_data.scope_metric_data_.size(), 1);
    EXPECT_EQ(metric_data.scope_metric_data_[0].metric_data_.size(), 1);
    EXPECT_EQ(metric_data.scope_metric_data_[0].metric_data_[0].point_data_attr_.size(), 2);
    return true;
  });
}

```

The test passes. Metric streams are created and exported for each attribute set of the counter.

The first counter instance's count is not exported and the test fails with no warnings or errors from the internal logger.

```
320: Note: Google Test filter = MeterTest.IdenticalInstrumentTests
320: [==========] Running 1 test from 1 test suite.
320: [----------] Global test environment set-up.
320: [----------] 1 test from MeterTest
320: [ RUN      ] MeterTest.IdenticalInstrumentTests
320: /workspaces/opentelemetry-cpp/sdk/test/metrics/meter_test.cc:391: Failure
320: Expected equality of these values:
320:   metric_data.scope_metric_data_[0].metric_data_[0].point_data_attr_.size()
320:     Which is: 1
320:   2
320: 
320: [  FAILED  ] MeterTest.IdenticalInstrumentTests (1 ms)
320: [----------] 1 test from MeterTest (1 ms total)
320: 
320: [----------] Global test environment tear-down
320: [==========] 1 test from 1 test suite ran. (1 ms total)
320: [  PASSED  ] 0 tests.
320: [  FAILED  ] 1 test, listed below:
320: [  FAILED  ] MeterTest.IdenticalInstrumentTests
320: 
320:  1 FAILED TEST
1/1 Test #320: metrics.MeterTest.IdenticalInstrumentTests ...***Failed    0.00 sec
```

**Additional context**
The spec defines identical instruments as: 

```
The term _identical_ applied to
Instruments describes instances where all [identifying
fields](./api.md#instrument) are equal.  The term _distinct_ applied
to Instruments describes instances where at least one field value is
different.
```

The spec requires identical instruments to be aggregated and exported by the sdk. 
See: [sdk.md#L905](https://github.com/open-telemetry/opentelemetry-specification/blob/9c8c30631b0e288de93df7452f91ed47f6fba330/specification/metrics/sdk.md?plain=1#L905)

The meter has a map `storage_registry_` that uses the instrument name only as a key. When a new sync instrument is created the storage registry entry for the key is set even if the key exists. See [meter.cc#L499](https://github.com/open-telemetry/opentelemetry-cpp/blob/4d52ab2c4d89981cc7b1d5228e40d684e9efe448/sdk/src/metrics/meter.cc#L499)

dbarker:
Fixes #3356 

**Why?** 
The spec requires identical instruments to be aggregated and duplicate instruments to be detected and exported. 
- Instruments are identical if all identifying fields equal: case-insensitive `name`, `kind`, `unit`, `description`
- Instruments are duplicates if the case-insensitive names match but another field is different. 

Current behavior: 
- Instrument names are treated as case sensitive and the only identifier used to track instruments by meter.
- Identical instruments are not aggregated. 
   - If a user creates another instrument with the same name then the first instrument's data is no longer collected/exported. 
- Duplicate instruments are not detected and reported to the user 

New behavior: 
- Instruments are identified based on all identifying fields. 
- Names are treated as case-insensitive per the spec and name case conflicts result in the first-seen instrument by name being used when all other fields equal. 
- Identical instruments are aggregated 
- Duplicate instruments are detected and reported to the user. Duplicates are functional instruments (will export data) per the spec. 

## Changes

- checks if an instrument already has a storage object in the meter's registry before creating a new one. 
- add tests for the spec requirements on aggregation of identical instruments
- add tests and log messages following the spec on [Duplicate instrument registration](https://github.com/open-telemetry/opentelemetry-specification/blob/9c8c30631b0e288de93df7452f91ed47f6fba330/specification/metrics/sdk.md?plain=1#L887). 


sdk/src/metrics/meter.cc
```diff
@@ -596,6 +662,73 @@ std::vector<MetricData> Meter::Collect(CollectorHandle *collector,
   return metric_data_list;
 }
 
+// Implementation of the log message recommended by the SDK specification for duplicate instruments.
+// See
+// https://github.com/open-telemetry/opentelemetry-specification/blob/9c8c30631b0e288de93df7452f91ed47f6fba330/specification/metrics/sdk.md?plain=1#L882
+void Meter::WarnOnDuplicateInstrument(const sdk::instrumentationscope::InstrumentationScope *scope,
```
dbarker, line 668: Example warning message intended to satisfy [metrics/sdk.md#L892](https://github.com/open-telemetry/opentelemetry-specification/blob/8bbac83c0c994c0a750cc688935fc1cae7fd39b3/specification/metrics/sdk.md?plain=1#L892)

> 
>  If the potential conflict involves multiple `description`
>    properties, setting the `description` through a configured View
>    SHOULD avoid the warning.

```bash
332: [Meter::WarnOnDuplicateInstrument] Creating a duplicate instrument of the same case-insensitive name. This may cause semantic errors in the data exported from this meter.
332: Different instrument descriptions found. Consider configuring a View to change the description of the duplicate instrument.
332: Scope: 
332:   name="test_meter"
332:   schema_url=""
332:   version=""
332: Existing instrument: 
332:   name="observable_counter"
332:   description="desc"
332:   unit="unit"
332:   kind="DoubleObservableCounter"
332: Duplicate instrument: 
332:   name="observable_counter"
332:   description="another_desc"
332:   unit="unit"
332:   kind="DoubleObservableCounter"
```

sdk/src/metrics/meter.cc
```diff
@@ -596,6 +662,73 @@ std::vector<MetricData> Meter::Collect(CollectorHandle *collector,
...
+          << "\nDuplicate instrument: " << InstrumentDescriptorLogStreamable{new_instrument});
+      return;
+    }
+  }
+}
+
+// Implementation of the log message recommended by the SDK specification for name case conflicts.
+// See
+// https://github.com/open-telemetry/opentelemetry-specification/blob/9c8c30631b0e288de93df7452f91ed47f6fba330/specification/metrics/sdk.md?plain=1#L910
+void Meter::WarnOnNameCaseConflict(const sdk::instrumentationscope::InstrumentationScope *scope,
```
dbarker, line 714: Example warning message intended to satisfy [metrics/sdk.md#L917](https://github.com/open-telemetry/opentelemetry-specification/blob/8bbac83c0c994c0a750cc688935fc1cae7fd39b3/specification/metrics/sdk.md?plain=1#L917)

> 
> The `name` of an Instrument is defined to be
> case-insensitive. If an SDK uses a case-sensitive encoding to represent this
> `name`, a duplicate instrument registration will occur when a user passes
> multiple casings of the same `name`. When this happens, the Meter MUST return
> an instrument using the first-seen instrument name and log an appropriate error
> as described above.

```bash
328: [Meter::WarnOnNameCaseConflict] Instrument name case conflict detected on creation. 
Returning the existing instrument with the first-seen instrument name. 
To resolve this warning consider configuring a View to rename the duplicate instrument.
328: Scope: 
328:   name="test_meter"
328:   schema_url=""
328:   version=""
328: Existing instrument: 
328:   name="OBServable_CounTER"
328:   description="desc"
328:   unit="unit"
328:   kind="DoubleObservableCounter"
328: Duplicate instrument: 
328:   name="observable_counter"
328:   description="desc"
328:   unit="unit"
328:   kind="DoubleObservableCounter"
```

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -65,6 +67,110 @@ struct InstrumentDescriptor
   InstrumentValueType value_type_;
 };
 
+struct InstrumentDescriptorUtil
+{
+  static bool CaseInsensitiveEquals(const std::string &lhs, const std::string &rhs) noexcept
```
ThomsonTan, line 72: This is for for ascii string comparison only which is good for instrumentation scope name, probably mention it in the function name like `CaseInsensitiveAsciiEquals`?

dbarker: Thanks for the feedback. I'll add `Ascii` it to the name. 

It wasn't clear if there were any other immediate use cases for this function so I put it here in just for use with instrument descriptors. There may be other use cases. Would it fit better in another more common header file? 

Looking quickly at the [spec on InstrumentationScope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) and it doesn't say if the scope name is case-sensitive or not.

sdk/src/metrics/meter.cc
```diff
@@ -486,18 +524,31 @@ std::unique_ptr<SyncWritableMetricStorage> Meter::RegisterSyncMetricStorage(
...
         }
-        auto multi_storage = static_cast<SyncMultiMetricStorage *>(storages.get());
-
-        auto storage = std::shared_ptr<SyncMetricStorage>(new SyncMetricStorage(
-            view_instr_desc, view.GetAggregationType(), &view.GetAttributesProcessor(),
+        std::shared_ptr<SyncMetricStorage> sync_storage{};
+        auto storage_iter = storage_registry_.find(view_instr_desc);
+        if (storage_iter != storage_registry_.end())
+        {
+          WarnOnNameCaseConflict(GetInstrumentationScope(), storage_iter->first, view_instr_desc);
```
dbarker, line 531: If the instrument is found with `storage_registry_.find(view_instr_desc)` then it is either identical OR a duplicate with a name (case) conflict to an existing instrument. In the case of a [name conflict](https://github.com/open-telemetry/opentelemetry-specification/blob/b42c863c24830bbac0313dd19517448398140943/specification/metrics/sdk.md?plain=1#L915) the spec requires returning the first-seen instrument of the given name. This section will do that by setting the storage ptr from the first-seen instrument to the new instrument.

sdk/src/metrics/meter.cc
```diff
@@ -486,18 +524,31 @@ std::unique_ptr<SyncWritableMetricStorage> Meter::RegisterSyncMetricStorage(
...
+          // static_pointer_cast is okay here. If storage_registry_.find is successful
+          // InstrumentEqualNameCaseInsensitive ensures that the
+          // instrument type and value type are the same for the existing and new instrument.
+          sync_storage = std::static_pointer_cast<SyncMetricStorage>(storage_iter->second);
+        }
+        else
+        {
+          WarnOnDuplicateInstrument(GetInstrumentationScope(), storage_registry_, view_instr_desc);
```
dbarker, line 539: If the instrument is not found with `storage_registry_.find(view_instr_desc)` then it is either a new unique instrument OR a [duplicate](https://github.com/open-telemetry/opentelemetry-specification/blob/b42c863c24830bbac0313dd19517448398140943/specification/metrics/sdk.md?plain=1#L876) (same case-insensitive `name` but with a different `kind`, `unit`, or `description`). In either case the spec requires creating a functional instrument even if it will cause a semantic error in exported data. In the case of a duplicate the spec recommends emitting a warning to the internal logger. 

> A _duplicate instrument registration_ occurs when more than one Instrument of
> the same `name` is created for identical
> Meters from the same MeterProvider but they have different identifying
> fields.
> 
> Whenever this occurs, users still need to be able to make measurements with the
> duplicate instrument. This means that the Meter MUST return a functional
> instrument that can be expected to export data even if this will cause
> semantic error in the data
> model.

sdk/include/opentelemetry/sdk/metrics/instruments.h
dbarker: I've pushed a commit to update the function name.

sdk/include/opentelemetry/sdk/metrics/instruments.h
```diff
@@ -65,6 +67,112 @@ struct InstrumentDescriptor
...
+  }
+};
+
+struct InstrumentEqualNameCaseInsensitive
+{
+  bool operator()(const InstrumentDescriptor &lhs, const InstrumentDescriptor &rhs) const noexcept
+  {
+    const bool names_match =
+        InstrumentDescriptorUtil::CaseInsensitiveAsciiEquals(lhs.name_, rhs.name_);
+    const bool kinds_match = (lhs.type_ == rhs.type_) && (lhs.value_type_ == rhs.value_type_);
```
ThomsonTan, line 144: Seems the variable definition could defeat short circuit, as all these bool variables will be evaluated unconditionally.

dbarker: Thanks Tom, I've updated those functions to short circuit.

ThomsonTan (approved):

dbarker:
Hi @lalitb, Could you take a look at this PR and share your feedback? Thanks!

lalitb:
> Hi @lalitb, Could you take a look at this PR and share your feedback? Thanks!

Apologies for the delay—will review and share feedback in the next couple of days.

lalitb (approved):
@dbarker Thank you for your PR. This is very nicely done, thoroughly covers all test scenarios :)

Issue #3364, opened 2025-04-16
GTest Printer functions


yashykt:
**Is your feature request related to a problem?**
I am looking for PrintTo / AbslStringify functions for public types in OpenTelemetry C++ to help with GTest output.

**Describe the solution you'd like**
What do you want to happen instead? What is the expected behavior?

Currently, without any helper functions, GTest will just print out the raw bytes which is not very helpful.

**Describe alternatives you've considered**
Which alternative solutions or features have you considered?

Refer https://github.com/grpc/grpc/commit/43272155f2141ec464a5eebb2979b45eefda15b4. GTest requires that `PrintTo` be defined in the same namespace as the type. I feel uncomfortable declaring something in opentelemetry's namespace even if it is only for test.

PR #3365, opened 2025-04-16
Add GTest PrintTo function for PointDataAttributes

yashykt:
Addresses #3364 for my immediate needs

## Changes

Added a `PrintTo` method for `PointDataAttributes`. (We could probably do this for all the other types, but for my immediate needs, this is enough.) 

I'm not sure where the maintainers would prefer such functions to be placed. Open to suggestions.


ThomsonTan:
Seems some format is applied by accident?

lalitb:
> Seems some format is applied by accident?

Actually, complete python venv has been added to the PR :facepalm:

yashykt:
my bad, and I thought I was careful :D

lalitb:
@yashykt - I so see some concerns maintaining a test-only helper that isn’t covered by the spec in the repo, as we want to keep the public API minimal and clean. Wouldn’t it make more sense for users of the library to add their own PrintTo implementation in their test code, within the appropriate otel namespace, if they need custom GTest output?

yashykt:
> Wouldn’t it make more sense for users of the library to add their own PrintTo implementation in their test code, within the appropriate otel namespace, if they need custom GTest output?

Adding methods to some other library's namespace is generally not recommended due to namespace pollution, conflicts, and general maintainability. But I guess I'll leave the decision to y'all. Do you feel comfortable with users adding their own methods to `opentelemetry::sdk::metrics`?

Alternative, we could create a separate `test_util` library?

yashykt:
ack

PR #3375, opened 2025-04-23
[DOC] Update doc comments to pass -WDocumention check

### LINKED ISSUE #3370: The documentation comments are not matching the code
**Describe your environment** 
N/A
**Steps to reproduce**
N/A

documentation comments are match with code.

A lot of comments do not match with the code.
For example:
```c++
/**
   * Writes a log with a severity of debug.
   * @tparam args Arguments which can be used to set data of log record by type.
   *  string_view                             -> body
   *  AttributeValue                          -> body
   *  SpanContext                             -> span_id,trace_id and trace_flags
   *  SpanId                                  -> span_id
   *  TraceId                                 -> trace_id
   *  TraceFlags                              -> trace_flags
   *  SystemTimestamp                         -> timestamp
   *  system_clock::time_point                -> timestamp
   *  KeyValueIterable                        -> attributes
   *  Key value iterable container            -> attributes
   *  span<pair<string_view, AttributeValue>> -> attributes(return type of MakeAttributes)
   */
  void Debug(ArgumentType &&...args) noexcept
  {
    static_assert(
        !detail::LogRecordHasType<Severity, typename std::decay<ArgumentType>::type...>::value,
        "Severity is already set.");
    this->EmitLogRecord(Severity::kDebug, std::forward<ArgumentType>(args)...);
  }
```
In this case, it should be `@param args` .

**Additional context**
I will list all the similar error bellow and fix them one by one with separated PRs.
- [ ] opentelemetry-cpp/api/include/opentelemetry/logs/logger.h

What's more, we should add some check to prevent these cases.

zurex:
Fixes #3370 

## Changes
* Update doc comments to compile with `-Werror -Wdocumentation`

This reduces the number of documentation warnings.
Enforcement in CI to be implemented separately.


api/include/opentelemetry/common/key_value_iterable.h
```diff
@@ -44,12 +44,12 @@ class NoopKeyValueIterable : public KeyValueIterable
 
   /**
    * Iterate over key-value pairs
-   * @param callback a callback to invoke for each key-value. If the callback returns false,
+   * A callback to invoke for each key-value. If the callback returns false,
    * the iteration is aborted.
```
lalitb, line 48: I know it's existing comment, but it's incorrect as the callback is never used here. Perhaps change the comment to:
```cpp
/**
 * No-op implementation: does not invoke the callback, even if key-value pairs are present.
 * @return true without iterating or invoking the callback
 */
```

api/include/opentelemetry/common/attribute_value.h
```diff
@@ -23,7 +23,7 @@ namespace common
 ///    (IEEE 754-1985) or signed 64 bit integer.
 ///  - Homogenous arrays of primitive type values.
 ///
-/// \warning
+/// \warning The OpenTelemetry C++ API does not support the following attribute
```
lalitb, line 26: this comment seems unfinished?

api/include/opentelemetry/common/key_value_iterable.h
zurex: Thanks for your suggestion. Will fix it in next commit.

CMakeLists.txt
```diff
@@ -603,6 +603,14 @@ if(OTELCPP_MAINTAINER_MODE)
     add_compile_options(-Werror)
     add_compile_options(-Wextra)
 
+    # Check if the compiler supports -Wdocumentation
+    include(CheckCXXCompilerFlag)
+    check_cxx_compiler_flag("-Wdocumentation" CLANG_SUPPORTS_DOCUMENTATION)
+    if(CLANG_SUPPORTS_DOCUMENTATION)
+      message(STATUS "Adding -Wdocumentation flag for Clang")
+      add_compile_options(-Wdocumentation)
```
dbarker, line 611: If there is a test command we can recommend contributors run locally to check for documentation issues before creating the pull request please add a note to [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CONTRIBUTING.md#how-to-send-pull-requests).

sdk/include/opentelemetry/sdk/trace/recordable.h
```diff
@@ -59,7 +59,7 @@ class Recordable
    * @param name the name of the attribute
    * @param value the attribute value
    */
-  virtual void SetAttribute(nostd::string_view key,
+  virtual void SetAttribute(nostd::string_view name,
```
dbarker, line 62: The argument name should stay as "key" for this one. The comment can be updated "name" -> "key". The term "key" is used in a lot of other methods and tests wrt attributes.

api/include/opentelemetry/common/key_value_iterable_view.h
```diff
@@ -130,7 +130,7 @@ inline static const common::KeyValueIterable &MakeAttributes(
 /**
  * Utility function to help to make a attribute view from a key-value iterable object
  *
- * @param attributes
+ * @param arg The key-value iterable object
```
dbarker, line 133: The `arg` argument still holds attributes, but must be of an `ArgumentType` type that evaluates to true with the `common::detail::is_key_value_iterable` type trait. This one may make sense to change `arg` -> `attributes` in the method and add a `@tparam` comment about the `ArgumentType`.

CMakeLists.txt
zurex: Good point, will add it to contributing guide.

api/include/opentelemetry/semconv/url_attributes.h
```diff
@@ -40,15 +40,14 @@ static constexpr const char *kUrlFragment = "url.fragment";
...
- *   <li><a
- * href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth">@code
- * AWSAccessKeyId @endcode</a></li> <li><a
- * href="https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html#RESTAuthenticationQueryStringAuth">@code
- * Signature @endcode</a></li> <li><a
- * href="https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token">@code sig
- * @endcode</a></li> <li><a
- * href="https://cloud.google.com/storage/docs/access-control/signed-urls">@code X-Goog-Signature
- * @endcode</a></li>
+ *   <li>
```
ThomsonTan, line 43: Is this re-format? Wondering whether it is necessary.

api/include/opentelemetry/context/runtime_context.h
```diff
@@ -56,7 +56,7 @@ class OPENTELEMETRY_EXPORT RuntimeContextStorage
 
   /**
    * Set the current context.
-   * @param the new current context
+   * @param context new current context
```
psx95, line 59: Perhaps it should be `The new current context`.

api/include/opentelemetry/semconv/incubating/system_metrics.h
```diff
@@ -277,12 +277,13 @@ static inline nostd::shared_ptr<metrics::ObservableInstrument> CreateAsyncDouble
...
- * <a
- * href="https://learn.microsoft.com/archive/blogs/askcore/windows-performance-monitor-disk-counters-explained#windows-performance-monitor-disk-counters-explained">"Disk%
- * Idle Time"</a> performance counter: @code uptime * (100 - "Disk\% Idle Time") / 100 @endcode</li>
+ * parallel are not counted). Measured as: <ul> 
+ *  <li>Linux: Field 13 from 
+ *    <a href="https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats">procfs-diskstats</a>
+ *  </li>
+ *  <li>Windows: The complement of 
+ *    <a href="https://learn.microsoft.com/archive/blogs/askcore/ windows-performance-monitor-disk-counters-explained#windows-performance-monitor-disk-counters-explained">"Disk% Idle Time"</a> 
+ * performance counter: @code uptime * (100 - "Disk\% Idle Time") / 100 @endcode</li>
```
psx95, line 286: Quick Question: Is this auto-formatted? 
If **not**, I think `<ul>`  should start from a new line.

api/include/opentelemetry/semconv/url_attributes.h
```diff
@@ -67,21 +66,18 @@ static constexpr const char *kUrlFull = "url.full";
 static constexpr const char *kUrlPath = "url.path";
 
 /**
- * The <a href="https://www.rfc-editor.org/rfc/rfc3986#section-3.4">URI query</a> component
+ * The <a href="https://www.rfc-editor.org/rfc/rfc3986#section-3.4">URI query</a> component @link https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token @code sig @endcode @endlink
```
psx95, line 69: Is the end result of this supposed to be: 
"The URI query component sig" 
with sig linking to https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token ? 

Does Doxygen `@link` link to HTML pages?  I don't see that functionality mentioned on the [documentation page](https://www.doxygen.nl/manual/commands.html#cmdlink)

exporters/memory/include/opentelemetry/exporters/memory/in_memory_span_exporter.h
```diff
@@ -81,7 +81,7 @@ class InMemorySpanExporter final : public opentelemetry::sdk::trace::SpanExporte
   }
 
   /**
-   * @param timeout an optional value containing the timeout of the exporter
+   * Timeout is an optional value containing the timeout of the exporter
```
psx95, line 84: I think `@param timeout` should remain.

Also good to add a small brief: "Attempt to shut down the in-memory span exporter." for the sake of completion.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -265,7 +257,7 @@ class HttpOperation
    * Perform curl message, this function only can be called in the polling thread and it can only
    * be called when got a CURLMSG_DONE.
    *
-   * @param code
+   * @param code CURLcode
```
psx95, line 260: This looks incomplete ?

sdk/include/opentelemetry/sdk/common/atomic_unique_ptr.h
```diff
@@ -61,7 +61,7 @@ class AtomicUniquePtr
 
   /**
    * Atomically swap the pointer with another.
-   * @param ptr the pointer to swap with
+   * @param other the pointer to swap with
```
psx95, line 64: Nit: To maintain consistency, start the param definition with a capital letter. 
`the pointer ...` &rarr; `The pointer ...` 

Also, in other places.

psx95 (changes requested):
Thanks for the PR! 

Left some comments.

sdk/include/opentelemetry/sdk/trace/recordable.h
zurex: OK, will do

api/include/opentelemetry/common/key_value_iterable_view.h
zurex: Will fix it in next commit

api/include/opentelemetry/semconv/url_attributes.h
zurex: It will throw exception if we use clang with `-Werror -WDocumention` flags. As i mentioned in the description section, integration as source code in downstream projects is a common way. These projects may enable some strict checks, we need to make sure opentelemetry code could works well under strict mode,

api/include/opentelemetry/semconv/url_attributes.h
zurex: You could check this guide: https://www.doxygen.nl/manual/autolink.html. By my understanding, we should keep HTML 'a' tag in this scenario.

psx95: Yes, the 'a' tag should be kept, I was talking about the `link` tag used here: `@link https://learn.microsoft.com/azure/storage/common/storage-sas-overview#sas-token @code sig @endcode @endlink`.

zurex: Sorry i misread it. You are right, we should replace the `@link` with a tag

ThomsonTan:
If the current format does break `-Werror -WDocumention`, it will also be good to add a CI pipeline with both option enabled, to ensure we have all the issues fixed and also guard future regression on it.

api/include/opentelemetry/semconv/incubating/process_attributes.h
```diff
@@ -133,8 +133,7 @@ static constexpr const char *kProcessInteractive = "process.interactive";
  * <p>
  * Control groups (cgroups) are a kernel feature used to organize and manage process resources. This
  * attribute provides the path(s) to the cgroup(s) associated with the process, which should match
- * the contents of the <a
- * href="https://man7.org/linux/man-pages/man7/cgroups.7.html">/proc/[PID]/cgroup</a> file.
+ * the contents of the <a href="https://man7.org/linux/man-pages/man7/cgroups.7.html">/proc/[PID]/cgroup</a> file.
```
marcalff, line 136: Everything under api/include/opentelemetry/semconv is generated code, as documented by:

```
/*
 * DO NOT EDIT, this is an Auto-generated file from:
 * buildscripts/semantic-convention/templates/registry/semantic_attributes-h.j2
 */
```

Depending on what needs to be fixed, changes could go in:

* the semconv repository, in the metadata associated with a semantic convention
* in the weaver code generation tool, to generate code better formatted for doxygen
* in the weaver configuration files in opentelemetry-cpp, used to drive code generation (possibly with new helpers more doxygen friendy to format comments)

Generated code should never by fixed manually, the fixes will be lost at every code generation.

What was the original build warning reported here ?
The anchor URL is split into several lines, but this still looks legit and valid.

marcalff (commented):
Thanks for the cleanup.

See preliminary comments on generated code.

marcalff:
Related:

* https://github.com/open-telemetry/opentelemetry-cpp/issues/1647

api/include/opentelemetry/semconv/incubating/process_attributes.h
zurex: Thanks for reminder and your suggestion, will check this part carefully:)

api/include/opentelemetry/common/key_value_iterable.h
zurex: Fixed.

sdk/include/opentelemetry/sdk/trace/recordable.h
zurex: Hi @dbarker , i checked the parameter names in this file and it looks like all of them use `name`. Would you mind double confirm it?

api/include/opentelemetry/common/key_value_iterable_view.h
zurex: fixed

api/include/opentelemetry/context/runtime_context.h
zurex: fixed.

api/include/opentelemetry/semconv/incubating/process_attributes.h
zurex: @marcalff the original build shows could not find another a tag since the close tag of a in another line.

sdk/include/opentelemetry/sdk/trace/recordable.h
dbarker: > Hi @dbarker , i checked the parameter names in this file and it looks like all of them use `name`. Would you mind double confirm it?

Hi @zurex. Yes I can confirm :) 

The spec uses the term `key`
https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes

The API  for `trace::Span` and derived classes from `sdk::trace::Recordable` also use the term `key` wrt attributes. 
[opentelemetry/trace/span.h#L78](https://github.com/open-telemetry/opentelemetry-cpp/blob/ed91e718f720e7209fcbb55fe37987888d45f58f/api/include/opentelemetry/trace/span.h#L78)
[multi_recordable.h#L69](https://github.com/open-telemetry/opentelemetry-cpp/blob/ed91e718f720e7209fcbb55fe37987888d45f58f/sdk/include/opentelemetry/sdk/trace/multi_recordable.h#L69)

zurex: Thanks @dbarker for confirm. Which means the parameter name in these lines need to be changed to `key ` by my understanding? https://github.com/open-telemetry/opentelemetry-cpp/blob/ed91e718f720e7209fcbb55fe37987888d45f58f/sdk/include/opentelemetry/sdk/trace/recordable.h#L65C2-L105C4

dbarker: The span `AddEvent` methods are correct as is and comments should stay with the term `name`.  The trace API spec is a good reference. 

https://opentelemetry.io/docs/specs/otel/trace/api/#add-events

zurex:
> Thanks for the cleanup PR.
> 
> Please remove all changes done on generated code for semantic conventions.
> 
> If there are unresolved doxygen issues, this is fine, the CI will need to wait longer before enforcing `-Wdocumentation` in maintainer mode.
> 
> Cleanup work (speaking from experience with warning cleanup, include-what-you-use, ...) is typically done incrementally, so the fixes available here should be merged now, without waiting for the entire build to be fixed.

Thanks for suggestions, kind of busy these days :(.

marcalff (approved):
LGTM, thanks for the fix.

See minor cleanup.

sdk/include/opentelemetry/sdk/trace/recordable.h
zurex: Fixed.

exporters/memory/include/opentelemetry/exporters/memory/in_memory_span_exporter.h
```diff
@@ -81,12 +81,14 @@ class InMemorySpanExporter final : public opentelemetry::sdk::trace::SpanExporte
...
-   * @param timeout an optional value containing the timeout of the exporter
+   * Attempt to shut down the in-memory span exporter.
+   * @param timeout Timeout is an optional value containing the timeout of the exporter
    * note: passing custom timeout values is not currently supported for this exporter
    * @return Returns the status of the operation
    */
-  bool Shutdown(std::chrono::microseconds /* timeout */) noexcept override
+  bool Shutdown(std::chrono::microseconds timeout) noexcept override
   {
+    (void)timeout;
```
marcalff, line 91:
```suggestion
  bool Shutdown(std::chrono::microseconds timeout OPENTELEMETRY_MAYBE_UNUSED) noexcept override
  {
```

sdk/include/opentelemetry/sdk/common/atomic_unique_ptr.h
marcalff, line 64:
```suggestion
   * @param other The pointer to swap with
```

api/include/opentelemetry/common/key_value_iterable.h
```diff
@@ -43,13 +43,12 @@ class NoopKeyValueIterable : public KeyValueIterable
   ~NoopKeyValueIterable() override = default;
 
   /**
-   * Iterate over key-value pairs
-   * @param callback a callback to invoke for each key-value. If the callback returns false,
-   * the iteration is aborted.
-   * @return true if every key-value pair was iterated over
+   * No-op implementation: does not invoke the callback, even if key-value pairs are present.
```
psx95, line 46: Nit: Technically, the `@param` should still be there.

psx95 (approved):
Thanks for addressing the comments! 
Overall it looks good, just a few minor nits remaining.

PR #3383, opened 2025-04-26
Optimize PeriodicExportingMetricReader Thread Usage

### LINKED ISSUE #3230: [Metrics SDK] Optimize Thread Usage in PeriodicExportingMetricReader
The current implementation of [PeriodicExportingMetricReader::CollectAndExportOnce](https://github.com/open-telemetry/opentelemetry-cpp/blob/3b8934619e81ebde4048f1d64ac06ec2e402211f/sdk/src/metrics/export/periodic_exporting_metric_reader.cc#L107) spawns a new thread for each export operation. This design was initially based on the assumption that the reader should handle multiple exporter requests for separate time series concurrently. However, this behavior is not supported by the specifications and leads to unnecessary resource consumption, especially when the export interval is short.

The proposal is to manage export execution on a single background worker thread, similar to [DoBackgroundWork](https://github.com/open-telemetry/opentelemetry-cpp/blob/3b8934619e81ebde4048f1d64ac06ec2e402211f/sdk/src/logs/batch_log_record_processor.cc#L152) in the batch processor.

ColeVanOphem:
Fixes #3230

## Changes

Optimizes thread usage in `PeriodicExportingMetricReader` by:

* Removing the separate thread previously used for `CollectAndExportOnce`. Collection and export now run synchronously in the worker thread.
* Simplifying the timeout logic within the collection callback, replacing the previous `std::future`-based mechanism.
* Refactoring the worker thread's wait loop using `std::condition_variable`.

### DIFF
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -98,8 +98,8 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
     worker_thread_instrumentation_->OnStart();
   }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-  do
+  std::unique_lock<std::mutex> lk(cv_m_);
+  while(true)
   {
     auto start = std::chrono::steady_clock::now();
 
@@ -134,8 +134,6 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
       worker_thread_instrumentation_->BeforeWait();
     }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-    std::unique_lock<std::mutex> lk(cv_m_);
     cv_.wait_for(lk, remaining_wait_interval_ms, [this]() {
       if (is_force_wakeup_background_worker_.load(std::memory_order_acquire))
       {
@@ -151,8 +149,11 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
       worker_thread_instrumentation_->AfterWait();
     }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-  } while (IsShutdown() != true);
+    if(IsShutdown())
+    {
+      return;
+    }
+  }
 
 #ifdef ENABLE_THREAD_INSTRUMENTATION_PREVIEW
   if (worker_thread_instrumentation_ != nullptr)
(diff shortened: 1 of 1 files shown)


sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -151,8 +149,11 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
       worker_thread_instrumentation_->AfterWait();
     }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-  } while (IsShutdown() != true);
+    if(IsShutdown())
+    {
+      return;
```
owent, line 154: Don't we need call `worker_thread_instrumentation_->OnEnd()` below?

lalitb: +1, probably we can just `break` from here instead of `return`.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -151,8 +141,11 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
       worker_thread_instrumentation_->AfterWait();
     }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-  } while (IsShutdown() != true);
+    if (IsShutdown())
```
ThomsonTan, line 144: With the `return` changed back to `break`, is it the same as the original `do...while` loop? If so, just restore it to the `do...while` loop which seems more concise?

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -98,8 +90,8 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
     worker_thread_instrumentation_->OnStart();
   }
 #endif /* ENABLE_THREAD_INSTRUMENTATION_PREVIEW */
-
-  do
+  std::unique_lock<std::mutex> lk(cv_m_);
```
dbarker, line 93: Thanks for the PR! 
Ideally we can scope the mutex lock as tightly as possible. Can it go right above the `cv_.wait_for(...)` inside the loop?

dbarker (approved):
Thanks for the contribution @ColeVanOphem. The change looks good. Please add a changelog entry and consider adding test cases/checks for the `PeriodicExportingMetricReader` if missing or impacted by this PR.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -164,61 +152,39 @@ void PeriodicExportingMetricReader::DoBackgroundWork()
...
+    this->Collect([this, &start](ResourceMetrics &metric_data) {
+      auto end = std::chrono::steady_clock::now();
+      if ((end - start) > this->export_timeout_millis_)
+      {
+        OTEL_INTERNAL_LOG_ERROR(
+            "[Periodic Exporting Metric Reader] Collect took longer configured time: "
+            << this->export_timeout_millis_.count() << " ms, and timed out");
+        return false;
+      }
+      this->exporter_->Export(metric_data);
```
lalitb, line 177: Note to self - the timeout interval `export_timeout_millis_` is checked at best effort basis between `collect` and `export`. This means, if the collect and/or export takes indefinite time, we can be in blocked state. This is specifically relevant to export method, as its execution time is not in the control of the SDK. The export operation runs to completion once started, regardless of how long it takes.

Nothing for this PR, as this is the existing behavior, but something to discuss further.

lalitb (approved):
LGTM. Thanks for the PR :)

ThomsonTan:
@ColeVanOphem please let us know if could add an entry to CHANGELOG as @dbarker mentioned, then the PR is ready for merging.

PR #3391, opened 2025-05-07
[SDK] Set small initial value for AttributesHashMap::hash_map_

### LINKED ISSUE #3390: The default size of AttributesHashMap::hash_map_ could be too large
The current `AttributesHashMap::hash_map_` has unspecified bucket count which makes it implementation defined. The implementation may starts with some large value like 13 on GCC which wastes memory for scenario where only a few attributes are used. For such scenario, it will be helpful to start with explicit small number of buckets.

```cpp
std::unordered_map<MetricAttributes, std::unique_ptr<Aggregation>, CustomHash> hash_map_;
```

ThomsonTan:
Fixes #3390 

## Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [SDK] Set small initial value for AttributesHashMap::hash_map_
+  [#3391](https://github.com/open-telemetry/opentelemetry-cpp/pull/3391)
+
 * [SDK] Base2 exponential histogram aggregation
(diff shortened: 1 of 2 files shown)


lalitb (approved):

marcalff (approved):

ThomsonTan:
The initial large bucket count for attribute hash map could be mitigated by recording multiple dimensions to one instruments. I think it is still rare to have only one or two attribute sets on an instrument.

Let me move this to draft and do some more experiment on it.

ThomsonTan:
More benchmark needed for this tuning, close it for now.

PR #3403, opened 2025-05-12
[SDK] Add credentials option to OTLP gRPC client (#3402)

### LINKED ISSUE #3402: Add credentials option to OTLP gRPC client
## The Problem
There is no way to specify gRPC [“call credentials”](https://grpc.io/docs/guides/auth/#credential-types) for the OTLP exporter. This is required when using short-lived tokens that must be refreshed which cannot be set at creation time using the [`metadata` option](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h#L50).

This is required by the [Google Cloud OTLP endpoint](https://cloud.google.com/stackdriver/docs/instrumentation/migrate-to-otlp-endpoints) that uses Access Token credentials.

## Proposed Solution
I would like to add a `credentials` option to [`OtlpGrpcClientOptions`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h) that allows specifying a `ChannelCredentials` object, rather than it being created in [`OtlpGrpcClient::MakeChannel`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/src/otlp_grpc_client.cc) with either `grpc::SslCredentials()` or `grpc::InsecureChannelCredentials()`.

It would allow using a custom [`MetadataCredentialsPlugin`](https://grpc.github.io/grpc/cpp/classgrpc_1_1_metadata_credentials_plugin.html) subclass to support arbitrary authentication methods, or an existing implementation like `GoogleDefaultCredentials()` for use with GCP APIs.
This is similar to the solution used by OpenTelemetry Python – its [OTLP exporters](https://opentelemetry-python.readthedocs.io/en/latest/exporter/otlp/otlp.html#id4) have a `credentials` parameter.

## Alternatives Considered
* Allow passing a gRPC channel into the exporter, rather than just its credentials. This would be similar to the [Go implementation](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc), and has the additional benefit of better supporting mocking the client for unit testing. However it’s slightly more effort for users, so could be added in addition to the credentials option?
* Support subclassing `OtlpGrpcClient` – currently there are no `virtual` methods, and the method I’d like to override `MakeChannel` is static. If it was made `virtual`, users could subclass the client to produce the gRPC channel however they need. However this could be harder to maintain compatilibity if the implementaiton of the client needs to change in the future.
* Change existing exporter constructors from private to public, that allow passing a gRPC service stub. These are currently used in tests – I’m not sure what happens here if the `OtlpGrpcClient` is different to the stub actually used by the exporter.
* Allow calling a function to generate headers like [Java implementation](https://github.com/open-telemetry/opentelemetry-java/pull/4630). This is not easily compatible with gRPC credentials objects without boilerplate to extract from them and insert into metadata.

DouglasHeriot:
Adds a `credentials` option to [`OtlpGrpcClientOptions`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h) that allows specifying a `ChannelCredentials` object, rather than it being created in [`OtlpGrpcClient::MakeChannel`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/otlp/src/otlp_grpc_client.cc) with either `grpc::SslCredentials()` or `grpc::InsecureChannelCredentials()`.

It allows using a custom [`MetadataCredentialsPlugin`](https://grpc.github.io/grpc/cpp/classgrpc_1_1_metadata_credentials_plugin.html) subclass to support arbitrary authentication methods, or an existing implementation like `GoogleDefaultCredentials()` for use with the [GCP OTLP API](https://cloud.google.com/stackdriver/docs/instrumentation/migrate-to-otlp-endpoints).

This is similar to the solution used by OpenTelemetry Python – its [OTLP exporters](https://opentelemetry-python.readthedocs.io/en/latest/exporter/otlp/otlp.html#id4) have a `credentials` parameter.

Fixes #3402

## Changes


DouglasHeriot:
Am I correct that the [ABI Policy](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/abi-policy.md) says this change is exempt because it is to the SDK, rather then API?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -43,6 +45,9 @@ struct OtlpGrpcClientOptions
   std::string ssl_client_cert_string;
 #endif
 
+  /** Use custom ChannelCredentials, instead of the SSL options above. */
+  std::shared_ptr<grpc::ChannelCredentials> credentials;
```
malkia, line 49: std::optional instead? (or the `nostd` one?)

DouglasHeriot: I don’t think we need the complexity of wrapping it in an optional. It would lead to two different unset states – `std::optional` set to `nullopt`, and `std::optional` containing a `nullptr`.

`std::shared_ptr` defaults to `nullptr` to represent the unset state, similar to how other unset options are represented as empty strings.

This is passed to `grpc::CreateCustomChannel` that takes a `std::shared_ptr<grpc::ChannelCredentials>`.

lalitb: `std::optional` is not available in C++14, which is minimum requirement for the SDK.

lalitb:
> Am I correct that the [ABI Policy](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/abi-policy.md) says this change is exempt because it is to the SDK, rather then API?

That's correct.

lalitb (approved):
LGTM. We need one more approval to get this through.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <grpcpp/security/credentials.h>
```
dbarker, line 6: With cmake, grpc is currently a private dependency [exporters/otlp/CMakeLists.txt#L44](https://github.com/open-telemetry/opentelemetry-cpp/blob/1e33bb595c2142ee65f87b9f2a671f27d07bad68/exporters/otlp/CMakeLists.txt#L44). Adding this header to the options will cause a test failure in the [otlp grpc component install test](https://github.com/open-telemetry/opentelemetry-cpp/blob/1e33bb595c2142ee65f87b9f2a671f27d07bad68/install/test/src/test_exporters_otlp_grpc.cc#L16) . We could consider forward declaring the `grpc::ChannelCredentials`. @owent what do you think?

dbarker: The design is such that you should be able to configure the options and use the client factory to create a client without leaking the grpc headers. 

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/cpp-sdk-factory-design.md#case-study-using-factory-and-shared-grpc-client-between-otlp-grpc-exporters

lalitb: Good catch. forward declaration would be option, but let's wait for @owent response :)

marcalff: Not sure a forward declaration will be sufficient.

`std::shared_ptr` will need to see the class destructor, which imply an actual include instead of a forward declaration.

owent: It's OK to import gRPC's include directory. Just don't link it as public dependency.
`std::shared_ptr` do not need class's destructor, only `std::unuque_ptr` in C++23 and upper need it.

DouglasHeriot: Ok, I’ll give forward declaration a try, and only add the gRPC include path if necessary.

marcalff (commented):
Please clarify:

* How does this proposal fits with OTEP https://github.com/open-telemetry/opentelemetry-specification/pull/4500

* This seem to cover the programmatic configuration only. I assume configuration of custom credentials using environment variables is out of scope at this point, please confirm.

* How, probably not yet but in the long term, would custom credential work with declarative configuration (see https://github.com/open-telemetry/opentelemetry-configuration/blob/f27ee13bbc7ee8156954aea79b1d8bfb0312b7cc/schema/common.json#L82), where the Grpc configuration is expressed in yaml ?


I support this change, but we need to make sure it fits the overall picture.

exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -392,7 +392,11 @@ std::shared_ptr<grpc::Channel> OtlpGrpcClient::MakeChannel(const OtlpGrpcClientO
   }
 #endif  // ENABLE_OTLP_RETRY_PREVIEW
 
-  if (options.use_ssl_credentials)
+  if (options.credentials)
+  {
+    channel = grpc::CreateCustomChannel(grpc_target, options.credentials, grpc_arguments);
+  }
+  else if (options.use_ssl_credentials)
```
ThomsonTan, line 399: Could a warning be emit here if both `credentials` and `use_ssl_credentials` are configured in the option, to notify that `credential` is preferred.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
owent: For `std::shared_ptr<grpc::ChannelCredentials>`, I think forward declaration is preferable.
So we can reduce a lot of the including headers if user do not need it.

marcalff (changes requested):
Thanks for the PR.

To be able to merge this to the main branch, the following changes are necessary.

1)

Please try a forward declaration instead of an include of grpc++ headers.
User code creating a grpc exporter may not have the grpc++ headers installed,
so this change can be breaking, especially for people who will not be using this feature.

This will also resolve the CI issues.

2)

Given that this feature is not stable in the specs, it needs a feature flag.

Suggested naming:

* CMake option `WITH_OTLP_GRPC_CREDENTIAL_PREVIEW`
* define `ENABLE_OTLP_GRPC_CREDENTIAL_PREVIEW`

See the existing option `WITH_OTLP_GRPC_SSL_MTLS_PREVIEW` for how to implement this.

DouglasHeriot:
Ok, I’ve had some time to work on this again. I’ve addressed the failing tests and comments above with a forward declaration. I’ve also added a warning log if both `credentials` and `use_ssl_credentials` are configured.

It is behind a feature flag supported by both CMake `-DWITH_OTLP_GRPC_CREDENTIAL_PREVIEW=ON` and Bazel `--@io_opentelemetry_cpp//exporters/otlp:with_otlp_grpc_credential_preview=true`. The current CI tests appear a bit inconsistent – in CMake I’ve enabled it where `WITH_OTLP_GRPC_SSL_MTLS_PREVIEW` was, but this does not appear to be enabled in any of the Bazel tests. So I’ve enabled it in Bazel where `ENABLE_METRICS_EXEMPLAR_PREVIEW` is.

How this related to OTEP, environment variables and other types of declarative configuration is an open question. I see this new `credentials` option as a low-level option that these other higher-level specifications can use to be implemented on top of. It will take some consideration how best to approach this, given the wide variety of possible [gRPC credential configurations](https://grpc.io/docs/guides/auth/), and support for application-specific custom credential objects.

exporters/otlp/src/otlp_grpc_client.cc
DouglasHeriot: Done

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
DouglasHeriot: Using a forward declaration has worked.

PR #3413, opened 2025-05-15
[SDK] Fix MetricProducer interface

### LINKED ISSUE #3412: SDK's MetricProducer API needs cleanup
https://github.com/open-telemetry/opentelemetry-cpp/pull/3235 added an optional parameter to the constructor of an interface.

This added parameter is unused, and doesn't seem useful in the interface. It should be the responsibility of the implementations of this class to accept a metric filter as a parameter.

In its current form, both the child class `MetricCollector` and the interface `MetricProducer` have a member `metric_filter_` which seems to be an oversight.

yashykt:
Fixes #3412

## Changes

Cleans up the `MetricProducer` interface by removing the unused `metric_filter_` parameter.


yashykt:
@IcySteam

IcySteam (changes requested):
Thanks for the PR Yash, having two instances of `metric_filter_` indeed looks to be an oversight.

Should we remove the `metric_filter_` member from `MetricCollector` instead, in order to closer to match the description of the [specs](https://github.com/open-telemetry/opentelemetry-specification/pull/3566)?

yashykt:
> Should we remove the metric_filter_ member from MetricCollector instead, in order to closer to match the description of the https://github.com/open-telemetry/opentelemetry-specification/pull/3566?

I was not involved in the creation of the spec, but I don't think they intended to add member variables to interfaces (since that no longer would be a pure interface) . I think they just meant that child classes of MetricCollector should take in metric_filter_. Either that or as a parameter to `Produce`

IcySteam (approved):

lalitb (approved):

psx95 (approved):

owent (approved):

PR #3421, opened 2025-05-16
[CMAKE] update cmake files in examples directory

### LINKED ISSUE #3399: Use the public read-only `opentelemetry-cpp` namespace cmake targets in examples
#3368 adds cmake read-only [ALIAS targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#alias-targets) for public use when integrating opentelemetry-cpp directly (with `FetchContent` or `add_subdirectory`). The examples should use these targets (`opentelemetry-cpp::api`, ..) instead of the internal mutable targets (`opentelemetry_api`, ...).

dbarker:
Fixes #3399

Updates the examples' cmake files to use the read-only ALIAS targets of `opentelemetry-cpp` added by https://github.com/open-telemetry/opentelemetry-cpp/pull/3368 and to use explicitly scoped target properties. 

## Changes
- Use public read-only `opentelemetry-cpp` ALIAS targets (ie: `opentelemetry-cpp::api` in place of `opentelemetry_api`)
- Use `target_compile_definitions` in place of `add_definitions` 
- Use `target_include_directories` in place of `include_directories`
- Add explicit scope keywords (ie: `PRIVATE`, `PUBLIC`) to `target_link_libraries`
- Remove some redundant target linking and includes
  - Rely on the transitive `PUBLIC` target properties (definitions, linked libraries, include directories) 
-  Replace `CMAKE_THREAD_LIBS_INIT` with the `Threads::Threads` target where needed, otherwise remove it and rely on the `opentelemetry-cpp::common` target's public transitive link to Threads.  
- Update the dll target (`opentelemetry-cpp::opentelemetry_cpp`) to include the directories for headers it depends on. Update the `ext_dll` component install test to verify.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 add_subdirectory(common)
-include_directories(common)
+
 if(WITH_OTLP_GRPC
    OR WITH_OTLP_HTTP
    OR WITH_OTLP_FILE)
--- a/examples/batch/CMakeLists.txt
+++ b/examples/batch/CMakeLists.txt
@@ -1,9 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
-
-include_directories(${CMAKE_SOURCE_DIR}/exporters/ostream/include)
-
 add_executable(batch_span_processor_example main.cc)
 
-target_link_libraries(batch_span_processor_example ${CMAKE_THREAD_LIBS_INIT}
-                      opentelemetry_exporter_ostream_span opentelemetry_trace)
+target_link_libraries(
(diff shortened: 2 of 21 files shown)


owent (approved):
LGTM.

BTW: Not a problem of this PR, but should `OPENTELEMETRY_BUILD_IMPORT_DLL` be interface definition of `opentelemetry_cpp` in `ext/src/dll/CMakeLists.txt`)? @ThomsonTan

ThomsonTan:
> `OPENTELEMETRY_BUILD_IMPORT_DLL`



> LGTM.
> 
> BTW: Not a problem of this PR, but should `OPENTELEMETRY_BUILD_IMPORT_DLL` be interface definition of `opentelemetry_cpp` in `ext/src/dll/CMakeLists.txt`)? @ThomsonTan

Yes, I think this is good suggestion. Opened below issue for tracking.

https://github.com/open-telemetry/opentelemetry-cpp/issues/3436

PR #3426, opened 2025-05-20
[CMAKE] remove global include_directories usage and rely on target properties

dbarker:
More cmake cleanup. 

## Changes
- Remove more global `include_directories` calls in favor of `target_include_directories`. Make sure the targets relying on these global includes link/include what they need. 
- Make the api target an interface link to the sdk target
- Remove usage of the sdk headers in the `foo_*` example libraries (These example libraries should link to the api only)

### DIFF
--- a/exporters/zipkin/CMakeLists.txt
+++ b/exporters/zipkin/CMakeLists.txt
@@ -1,8 +1,6 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-include_directories(include)
-add_definitions(-DWITH_CURL)
 add_library(
   opentelemetry_exporter_zipkin_trace
   src/zipkin_exporter.cc src/zipkin_exporter_factory.cc src/recordable.cc)
--- a/sdk/test/common/CMakeLists.txt
+++ b/sdk/test/common/CMakeLists.txt
@@ -47,7 +47,7 @@ if(WITH_BENCHMARK)
 
   add_executable(circular_buffer_benchmark circular_buffer_benchmark.cc)
   target_link_libraries(circular_buffer_benchmark benchmark::benchmark
-                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_api)
+                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_common)
 
   add_executable(attributemap_hash_benchmark attributemap_hash_benchmark.cc)
   target_link_libraries(attributemap_hash_benchmark benchmark::benchmark
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ Increment the:
 * [CMAKE] Add generated protobuf headers to the opentelemetry_proto target
   [#3400](https://github.com/open-telemetry/opentelemetry-cpp/pull/3400)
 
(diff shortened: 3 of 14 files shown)


exporters/zipkin/CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-include_directories(include)
-add_definitions(-DWITH_CURL)
```
ThomsonTan, line 5: Is this definition of `WITH_CURL` not used and safe to remove?

marcalff (approved):
LGTM

sdk/test/common/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if(WITH_BENCHMARK)
 
   add_executable(circular_buffer_benchmark circular_buffer_benchmark.cc)
   target_link_libraries(circular_buffer_benchmark benchmark::benchmark
-                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_api)
+                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_common)
```
ThomsonTan, line 50: Is it better to name it as `opentelemetry_sdk_common` instead of `opentelemetry_common`? Just raise this for awareness and no need to take action in this PR.

exporters/zipkin/CMakeLists.txt
dbarker: I don't think WITH_CURL is used anywhere.

ThomsonTan:
Seems this is a breaking change if the user relies on the `include_directories()` behavior?

ThomsonTan:
> > Seems this is a breaking change if the user relies on the `include_directories()` behavior?
> 
> @ThomsonTan I don't expect any breaking changes to users. The scope of the `include_directories` call covers all targets defined in the same CMakeLists.txt file and the sub directories it adds. User code shouldn't be included in this scope.

FYI, here is one case which relied on `include_directories()` to work, but needs to switch to `target_link_libraries()` now.

https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/551

dbarker:
> > > Seems this is a breaking change if the user relies on the `include_directories()` behavior?
> > 
> > 
> > @ThomsonTan I don't expect any breaking changes to users. The scope of the `include_directories` call covers all targets defined in the same CMakeLists.txt file and the sub directories it adds. User code shouldn't be included in this scope.
> 
> FYI, here is one case which relied on `include_directories()` to work, but needs to switch to `target_link_libraries()` now.
> 
> [open-telemetry/opentelemetry-cpp-contrib#551](https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/551)

That is a good catch. One of the drawbacks with the global include_directories (and why it is good to move towards target_include_directories) is that it can hide issues like this since the global includes are not transitive with the targets. The fix you made in https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/551 is a good one. 

Another instance where issues were hidden is with the foo_* library examples. They had included the sdk headers but did not link to the sdk targets (and shouldn't have). If a user copied those examples and the cmake files to build in their project - they would have previously failed to find the sdk headers at compile time.

Issue #3405, opened 2025-05-13
Add test for building otel-cpp shared libraries against shared libraries of grpc and protobuf


dbarker:
There is a testing gap for building otel-cpp shared libraries against shared libraries for protobuf and grpc at the moment. 

Building `opentelemetry-cpp` as shared libraries against newer versions of protobuf and grpc requires those dependencies to also be built as shared libraries. Tests fail when building otel-cpp shared libraries on Linux against grpc 1.67.1 and protobuf 5.27.0 static libraries. Those tests pass with older versions of grpc (1.54) and protobuf (3.21) though.

dbarker:
> In my understanding, It's OK to ling static gRPC and protobuf only to one dynamic/executable. That's why we only link all gRPC's libraries into `otlp_grpc_client` and make it private. And also we can not link it into executable again.
> It's the same for protobuf, could you please provide all the linking relationship about protobuf and gRPC here?

Something has changed in a newer release of protobuf that causes issues with linking static protobuf libs to shared otel-cpp libs. Some unit tests will fail when trying this with protobuf 5.27.0. I'll post a draft PR to reproduce the issue in ci. 

The current tests in ci cover linking shared otel-cpp libs with static protobuf libs on older versions (protobuf 3.23 and 3.21) but not newer versions (5.27+)

malkia:
Why such old versions of grpc / protobuf are still targeted? gRPC is at 1.70.2, and protobuf 31.0.

dbarker:
> Why such old versions of grpc / protobuf are still targeted? gRPC is at 1.70.2, and protobuf 31.0.

Just using grpc 1.67 and protobuf 5.27 to demonstrate the errors in building otel-cpp shared libs (the result is the same with the latest versions). 

CI is testing static builds with up to grpc 1.72 and protobuf 29.3 successfully. It seems reasonable to officially target the latest versions once the shared library build issues are resolved and tests added.

malkia:
FYI: protobuf 31.0 deprecates MSVC (due to compiler not supporting long paths). Clang-cl still works.

owent:
Do you build gRPC with the same version of protobuf as it's used in otel-cpp?
Could you please help to test if <https://github.com/owent/opentelemetry-cpp/tree/link_problems_of_grpc> can solve this problem?

dbarker:
[...]

PR #3427, opened 2025-05-20
Test building opentelemetry-cpp as shared libraries against shared protobuf and grpc libraries

dbarker:
Test for #3405 - This draft PR is just to demonstrate the current state of building opentelemetry-cpp shared libraries against newer versions of protobuf and grpc. 

## Changes

- adds cmake install tests for 
   - Building shared otel-cpp libs against static protobuf v5.27 and grpc v1.67 libs - This is known to fail tests and the otlp cmake targets do not function as expected (either segfault or just fail to export data)
   - Building shared otel-cpp libs against shared protobuf v5.27 and grpc v1.67 libs - The otlp grpc tests fail to build, but the otlp grpc cmake targets can be used in an application and work as expected. 

### DIFF
--- a/.github/workflows/cmake_install.yml
+++ b/.github/workflows/cmake_install.yml
@@ -270,7 +270,7 @@ jobs:
         run: ./ci/do_ci.sh cmake.opentracing_shim.install.test
 
   ubuntu_2404_conan_latest:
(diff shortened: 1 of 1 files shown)


dbarker:
Closing this as addressed by #3435

PR #3432, opened 2025-05-21
[CMAKE] Remove third-party version mismatch warning

dbarker:
This is a new warning added but is unreliable when building and installing opentelemetry-cpp against third party submodules or external clones. In those cases determining the dependency version is not reliable or implemented in all cases. 

This check can be revisited in the future. 

## Changes

- remove the warning


owent (approved):

ThomsonTan:
Is there any known issue on these we only have a few dependencies?

dbarker:
> Is there any known issue on these we only have a few dependencies?

The dependencies that may be included as submodules or src and built in-tree (add_subdirectory) won't populate the CMake `<package>_VERSION` variable (which is populated when the dependency is found with `find_package`).  Those include `nlohmann-json`, `opentracing-cpp`, and `prometheus-cpp`. In at least one case (vcpkg package used by ci of `prometheus-cpp`) the find_package call doesn't populate the `<package>_VERSION` variable.

Since there are testing gaps in ci for all the options to include a third party dependency supported by otel-cpp cmake, it is better to remove this version check warning for now and reevaluate a check like this once tests can be added.

ThomsonTan:
> > Is there any known issue on these we only have a few dependencies?
> 
> The dependencies that may be included as submodules or src and built in-tree (add_subdirectory) won't populate the CMake `<package>_VERSION` variable (which is populated when the dependency is found with `find_package`). Those include `nlohmann-json`, `opentracing-cpp`, and `prometheus-cpp`. In at least one case (vcpkg package used by ci of `prometheus-cpp`) the find_package call doesn't populate the `<package>_VERSION` variable.
> 
> Since there are testing gaps in ci for all the options to include a third party dependency supported by otel-cpp cmake, it is better to remove this version check warning for now and reevaluate a check like this once tests can be added.

Thanks for sharing the details. Could we file an issue to track the revisiting of the issue?

ThomsonTan (approved):

PR #3435, opened 2025-05-22
[BUILD] Fixes grpc linking for OTLP exporter's tests

### LINKED ISSUE #3405: Add test for building otel-cpp shared libraries against shared libraries of grpc and protobuf

owent:
Force build opentelemetry_proto and OTLP exporters as static libraries when protobuf or grpc are built static.

Fixes #3405 

## Changes

+ Build opentelemetry_proto and OTLP exporters as static libraries when protobuf or gRPC are statically built.


cmake/opentelemetry-proto.cmake
```diff
@@ -221,6 +221,39 @@ elseif(PROTOBUF_VERSION AND PROTOBUF_VERSION VERSION_LESS "3.16")
...
+      ".*(protobuf).*")
+    foreach(protobuf_lib_file ${protobuf_lib_files})
+      if(protobuf_lib_file MATCHES
+         "(^|[\\\\\\/])[^\\\\\\/]*protobuf[^\\\\\\/]*.a$")
+        set(protobuf_lib_type "STATIC_LIBRARY")
+        break()
+      endif()
+    endforeach()
+  endif()
+else()
```
dbarker, line 245: Is this else to handle `Protobuf_LIBRARIES` needed? Since cmake 3.9 the FindProtobuf module imports the `protobuf::libprotobuf` target and very old versions (see [v16.2](https://github.com/protocolbuffers/protobuf/blob/c18f5e71d86063fd6cea2c47cd7ab4131db5c9e2/cmake/install.cmake#L8)) install the target for import with `protobuf-config.cmake`.

cmake/opentelemetry-proto.cmake
```diff
@@ -334,8 +368,9 @@ endif()
 if(WITH_OTLP_GRPC)
   add_library(
```
dbarker, line 369: Line 385 in this file likely needs some review and changes with this PR. It adds `gRPC::gRPC++` as a public link to the `opentelemetry_proto_grpc` target if `grpc_lib_type` is `SHARED_LIBRARY`. The `opentelemetry_proto_grpc` target is then linked publicly to the `opentelemetry_exporter_otlp_grpc_client` which will make `gRPC::gRPC++` public to users of the client target. This contradicts the private grpc link made by the `opentelemetry_exporter_otlp_grpc_client`.

cmake/tools.cmake
```diff
@@ -220,3 +222,106 @@ function(project_build_tools_patch_default_imported_config)
...
+        __project_build_tools_recursive_scan_imported_locations(
+          ${__TARGET_LINK_LIBRARIES_SELECTED})
+      endif()
+    endif()
+  endforeach()
+  math(EXPR __project_build_tools_recursive_call_level
+       "${__project_build_tools_recursive_call_level} - 1")
+endmacro()
+
+function(project_build_tools_recursive_scan_imported_locations OUTPUT_VAR)
```
dbarker, line 313: Which build scenario does this function address? Is there an example ci test that will fail without it?

cmake/opentelemetry-proto.cmake
```diff
@@ -221,6 +221,39 @@ elseif(PROTOBUF_VERSION AND PROTOBUF_VERSION VERSION_LESS "3.16")
...
+if(TARGET protobuf::libprotobuf)
+  get_target_property(protobuf_lib_type protobuf::libprotobuf TYPE)
+  # protobuf_lib_type may be "INTERFACE_LIBRARY" in some build systems, such as
+  # conan.
+  if(protobuf_lib_type STREQUAL "INTERFACE_LIBRARY")
+    project_build_tools_recursive_scan_imported_locations(
+      protobuf_lib_files TARGET_NAME protobuf::libprotobuf TARGET_MATCH
+      ".*(protobuf).*")
+    foreach(protobuf_lib_file ${protobuf_lib_files})
+      if(protobuf_lib_file MATCHES
```
dbarker, line 238: This should handle other platforms we support (`.lib` ... )

exporters/otlp/CMakeLists.txt
```diff
@@ -23,10 +33,34 @@ target_link_libraries(opentelemetry_otlp_recordable
 
 if(WITH_OTLP_GRPC)
   find_package(gRPC REQUIRED)
+  # gRPC uses numerous global variables, which can lead to conflicts when a
```
dbarker, line 36: Consider moving the definition of `grpc_lib_type` to `opentelemetry-proto.cmake` for consistency with `protobuf_lib_type`.

cmake/opentelemetry-proto.cmake
owent: Each dll/exe of PE ABI has it's own copy of symbols and heap. So I think it won't cause conflicts in PE.

cmake/opentelemetry-proto.cmake
owent: Some crossing toolchains set `Protobuf_LIBRARIES` and `Protobuf_PROTOC_EXECUTABLE` without target. I think we should support them.

cmake/tools.cmake
owent: We can use #3427 to test it. But I wonder if we have better way instead of installing ninja each time in CI?

cmake/opentelemetry-proto.cmake
dbarker: How do we test it? I don't think we have a ci test that doesn't import the `protobuf::libprotobuf` target currently.

cmake/opentelemetry-proto.cmake
dbarker: The case I'm considering is that conan can be used on Windows and may result in `protobuf_lib_type` being set to `INTERFACE_LIBRARY` initially. Then the match for `.a` would fail and `protobuf_lib_type` would never be set to `STATIC_LIBRARY`.  Then `OPENTELEMETRY_OTLP_TARGETS_LIB_TYPE` would not be set to `STATIC`

exporters/otlp/CMakeLists.txt
owent: Good point. Done.

owent:
> Thanks for working on this @owent. I've added some feedback. Can you cherry-pick the tests commit from the testing PR #3427 into this PR so we can verify the changes? 
> 
> I've tested this PR locally and the tests now pass with otel-cpp shared libs built with protobuf/grpc static libs (grpc 1.67 protobuf 5.27). The tests still fail to build with otel-cpp shared libs built with protobuf/grpc shared libs though. 

The problem of building  shared libraries with conan is solved  now.
And I also cherry-pick #3427 into this PR to check the  building.
Could  you  please  have  a look when you have time?@dbarker

cmake/tools.cmake
dbarker: The recursion in this method is concerning and hard to review. 

Digging a bit deeper it seems the need for handling grpc and protobuf targets as INTERFACE_LIBRARY types is a conan quirk and has been addressed in their new CMakeConfigDeps generator that imports the real targets (https://docs.conan.io/2/incubating.html#new-cmakeconfigdeps-generator). 

The following function can extract the actual type from the `CONAN_LIB` target (conan's protobuf::libprotobuf links to a CONAN_LIB target). We could use simple regex on the conan syntax like this (or maybe just find the first real target inside the generator expressions and use its type). 

Can you replace the recursion with something simple like this? 
  
```cmake
function(get_type_from_interface_target IN_TARGET_NAME OUT_TARGET_TYPE)
  get_target_property(
      _interface_libs ${IN_TARGET_NAME} INTERFACE_LINK_LIBRARIES)
  foreach(_linked_lib IN LISTS _interface_libs)
    if(_linked_lib MATCHES "CONAN_LIB::[A-Za-z0-9_+]+(_[A-Za-z0-9_+]+)*")
      if(TARGET ${CMAKE_MATCH_0})
        get_target_property(_lib_type "${CMAKE_MATCH_0}" TYPE)
        if(_lib_type)
          set(${OUT_TARGET_TYPE} "${_lib_type}" PARENT_SCOPE)
          return()
        endif()
      endif()
    endif()
  endforeach()
endfunction()
```

cmake/opentelemetry-proto.cmake
```diff
@@ -345,7 +384,30 @@ if(WITH_OTLP_GRPC)
...
+  # gRPC uses numerous global variables, which can lead to conflicts when a
+  # user's dynamic libraries, executables, and otel-cpp are all built as dynamic
+  # libraries and linked against a statically built gRPC library. This may
+  # result in crashes. To prevent such conflicts, we also need to build
+  # opentelemetry_exporter_otlp_grpc_client as a static library.
   get_target_property(grpc_lib_type gRPC::grpc++ TYPE)
+  # grpc_lib_type may be "INTERFACE_LIBRARY" in some build systems, such as
+  # conan.
+  if(grpc_lib_type STREQUAL "INTERFACE_LIBRARY")
+    project_build_tools_recursive_scan_imported_locations(
```
dbarker: The `INTERFACE_LINK_LIBRARIES` for gRPC::grpc++ with conan includes: 

```cmake
$<$<CONFIG:Debug>:>;$<$<CONFIG:Debug>:CONAN_LIB::grpc_gRPC_grpc++_grpc++_DEBUG>
```
The `CONAN_LIB::grpc_gRPC_grpc++_grpc++_DEBUG` target has the actual type (`SHARED_LIBRARY` or `STATIC_LIBRARY`)

Using a simple search to find the first real target under the conan interface target is a direct way to get at the type.

cmake/opentelemetry-proto.cmake
```diff
@@ -221,6 +221,43 @@ elseif(PROTOBUF_VERSION AND PROTOBUF_VERSION VERSION_LESS "3.16")
...
+# protobuf uses numerous global variables, which can lead to conflicts when a
+# user's dynamic libraries, executables, and otel-cpp are all built as dynamic
+# libraries and linked against a statically built protobuf library. This may
+# result in crashes. To prevent such conflicts, we also need to build
+# opentelemetry_exporter_otlp_grpc_client as a static library.
+if(TARGET protobuf::libprotobuf)
+  get_target_property(protobuf_lib_type protobuf::libprotobuf TYPE)
+  # protobuf_lib_type may be "INTERFACE_LIBRARY" in some build systems, such as
+  # conan.
+  if(protobuf_lib_type STREQUAL "INTERFACE_LIBRARY")
```
dbarker: The `INTERFACE_LINK_LIBRARIES` property of the conan `protobuf::libprotobuf` target includes: 
```cmake
$<$<CONFIG:Debug>:>;$<$<CONFIG:Debug>:CONAN_LIB::protobuf_protobuf_libprotobuf_protobufd_DEBUG>
```
The `CONAN_LIB::protobuf_protobuf_libprotobuf_protobufd_DEBUG` target is a real one and has the right type.

cmake/tools.cmake
owent: I'm not sure if other build systems have the similar behavior. So I prefer to use a more common way. What do you think about it?

dbarker: We shouldn't add recursion to the cmake code. 

It seems like a specific case for the conan CMakeDeps generator that is changing soon. It also seems reasonable to not support this case from conan and direct users to the new CMakeConfigDeps generator, which will import the actual protobuf and grpc targets. Alternatively they can build all as shared or all as static. 

How about removing this handling of INTERFACE_LIBRARY types for protobuf and grpc then just fall back to the current behavior of not explicitly setting the linkage type?

cmake/opentelemetry-proto.cmake
owent: Do you think we should also support other build systems which may have similar alias as conan?

dbarker: As far as I know, creating only interface targets is unique to this one Conan generator (CMakeDeps). 

Checking the type property of the target in the INTERFACE _LINK_LIBRARIES of the interface target will work. We can do the regex match on libprotobuf and grpc++ to find the real target without searching for CONAN_LIB. This would be one for loop and remove the recursion.

If new cases come up later they could be addressed in a separate PR. What do you think?

owent: I just removed recursive scanning, it's simpler now.

cmake/tools.cmake
owent: I removed recursive scanning. Could you review when you have time?
The link names include nested '$<...>' expressions (more than one level deep).
Rather than hardcoding for Conan specifically, I think we should implement a general solution that would support similar patterns used by other build systems.

dbarker: Thanks @owent. I took another look. Can you share more about why this should be a general solution? 

My understanding is that this is a specific case caused by just one generator of conan (CMakeDeps) and it is not a general problem. I'm concerned with how we will test and maintain the more complex general solution.

cmake/opentelemetry-proto.cmake
owent: Yes, we have no crossing compiling checking yet.  It requires a cross toolchain.
The problems were reported by some users before. Sorry but I forgot the details.
I also compiled otel-cpp for android and iOS before. We have a pipeline which can check the compiling but it's not opensource and it will take a long time on every build.

It's welcome to add CI jobs for cross compiling but I think we can do it in a seperate PR.

cmake/tools.cmake
owent: We cannot maintain every component system, but many component systems have similar implementation approaches. Just like some protocols, algorithms, or semantic transformations, there are many best practices and engineering conventions.

We can provide mechanisms to offer these features to meet this type of requirement. Conan is just one component system we found that implements this approach. However, the C++ component environment is not unified, and end users may have many smaller, even private implementations that also adopt this approach.

Among the component systems I have encountered, there are some based on secondary development of Conan, they can import package of conan and add some their own private features. I'm not sure whether they will continue to use the Conan name (most open source licenses do not allow the use of the original name when redistributing). I think we should not hardcode the name for just one build system like Conan.

dbarker: > We cannot maintain every component system

Thanks for the details. I agree we should not try to maintain support for all possible packages managers and CMake configurations they present. If easier to have a conversation live on slack I'd be happy to meet this week. 

To move forward with this PR and unblock upgrading to the latest grpc and protobuf I'm proposing: 
1. Remove handling of `protobuf::libprotobuf` and `gRPC::grpc++` targets as `INTERFACE_LIBRARY` targets from this PR and in that case default to the prior behavior of leaving the `add_library` `<type>` undefined (let `BUILD_SHARED_LIBS` determine the type).
2. Remove the `Run Tests (shared libs)` step from the `ubuntu_2404_conan_latest` job and move that step to the [`ubuntu_2404_latest`](https://github.com/open-telemetry/opentelemetry-cpp/blob/3d901a81354435904e3b9bd5c4179c934afdbb6a/.github/workflows/cmake_install.yml#L91C3-L91C21) job. 
3. [Optional] Open an issue to collect community feedback on handling building shared otel-cpp against static protobuf/grpc (highlighting the known issue with the conan CMakeDeps generator and ask for feedback on other package manager configurations that don't import real protobuf or grpc targets). 

Rationale: 
- The most important change from this PR is to support the nominal case where grpc and protobuf targets are imported as real targets (native cmake installs, vcpkg, homebrew, conan with the CMakeConfigDeps generator).  
-  The code to support the case of these targets being INTERFACE_LIBRARY types adds a lot of complexity and the use case cannot be fully tested or defined. We'll struggle to refactor/change this CMake code or ever remove it in absence of a concrete test or use case definition. 
   -  It is unclear that the solution in the PR will handle all expected CMake configurations - user customized static/shared suffixes and other variations possible with different link library hierarchies and more complex generator expressions. 

- Currently we only have one known case with the conan CMakeDeps generator. Conan users can select the CMakeConfigDeps generator to import the real targets and don't need the fix to be in otel-cpp.  
- No users have requested a solution to the problem of building shared otel-cpp against static protobuf/grpc with the conan CMakeDeps generator. They may have already found a solution with the tools provided by conan

cmake/opentelemetry-proto.cmake
dbarker: Creating a separate PR for a cross compiling test sounds perfect.

cmake/tools.cmake
owent: Thanks for the explain. I'm not familar with conon. I agree it's better not to make otel-cpp too complex if there is another solution for conan's users.
The INTERFACE_LIBRARY supporting is removed and the test steps of building shared libraries with conan are also removed now. Could you please review it again?

cmake/opentelemetry-proto.cmake
```diff
@@ -302,6 +322,7 @@ endif()
 set(OPENTELEMETRY_PROTO_TARGETS opentelemetry_proto)
 add_library(
   opentelemetry_proto
+  ${OPENTELEMETRY_OTLP_PROTO_LIB_TYPE}
```
dbarker: `OPENTELEMETRY_OTLP_PROTO_LIB_TYPE` is never set and can be removed since the library type (STATIC, SHARED) is included in the `OTELCPP_PROTO_TARGET_OPTIONS`

cmake/tools.cmake
dbarker: Thanks for updating this @owent. I've tested locally building otel-cpp shared libs against grpc 1.71 and protobuf 1.30 static libs. The tests pass.

cmake/opentelemetry-proto.cmake
dbarker: @owent this looks like the last change needed then I'll approve. Thanks.

owent: Thanks, it's removed.

PR #3457, opened 2025-06-02
Fix: Use shared_ptr internally for AttributesProcessor to prevent use-after-free

### LINKED ISSUE #3456: Potential use-after-free of AttributesProcessor in metric storage after MeterProvider shutdown
There is a potential use-after-free bug in the OpenTelemetry C++ SDK metrics implementation.

**Problem**:

- Metric storage class(`SyncMetricStorage`) store a raw pointer to the AttributesProcessor instance, which is owned by the associated View.

- The View (and its `AttributesProcessor`) is owned by the `ViewRegistry`, which is managed by the `MeterContext` and ultimately the `MeterProvider`.

- When the `MeterProvider` is shut down or destroyed, the `ViewRegistry` and its `View`s are deleted, invalidating the `AttributesProcessor` pointer inside any metric storage objects.

- If metric instruments are recorded after shutdown (especially from other threads, background tasks, or due to application logic errors), this leads to use-after-free, undefined behavior, and possible segmentation faults.

lalitb:
Fixes #3456

## Changes

This PR changes metric storage and View class to use `std::shared_ptr<const AttributesProcessor>` internally, while the public View API continues to accept a `std::unique_ptr`. The unique pointer is promoted to a shared pointer inside the View constructor. This prevents use-after-free when metrics are recorded after `MeterProvider` shutdown or destruction.

Each metric recording operation now accesses the `AttributesProcessor` via a `std::shared_ptr`. This adds a minimal overhead due to shared pointer reference counting, but ensures memory safety, so should be acceptable.

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
+++ b/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
@@ -58,7 +58,7 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
 public:
   SyncMetricStorage(InstrumentDescriptor instrument_descriptor,
                     const AggregationType aggregation_type,
-                    const AttributesProcessor *attributes_processor,
+                    std::shared_ptr<const AttributesProcessor> attributes_processor,
 #ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
                     ExemplarFilterType exempler_filter_type,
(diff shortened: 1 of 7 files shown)


ThomsonTan:
Thanks for raising the PR. Could you add a dedicated test to validate the seg fault?

lalitb:
> Thanks for raising the PR. Could you add a dedicated test to validate the seg fault?

done

marcalff (approved):
LGTM

PR #3460, opened 2025-06-03
[CMAKE] Unify third party dependency management

dbarker:
Update third party dependency management in CMake. 

## Changes
- Add an `otel_add_external_project` function unify how third party dependencies are found/fetched. 
  - Create a cmake script file for each dependency to ensure all required targets are imported/installed
  - Enable fetching from submodules or git repo with the version tags in the `third_party_release file`. 
- Remove the `install_windows_deps` CMake function and ARCH detection
   - This seems intended to support bootstrapping vcpkg windows packages within the otel-cpp CMake file. 
   - This extra logic seems untested in ci and not required since vcpkg users can set the CMAKE_TOOLCHAIN_FILE as expected
   - The bootstrapping use case is addressed through fetching all dependencies with standard CMake FetchContent from github. 
- Clean up cmake dependency linking
   - Use the standard `CURL::libcurl` target
   - Use the explicit `prometheus-cpp` `pull` and `core` targets
   - Use either the shared or static `opentracing-cpp` target 
- Add cmake options to force fetch dependencies or to require all local dependencies (following the model used by Protobuf)
   - If only local dependencies should be used then set `OTELCPP_REQUIRE_LOCAL_DEPENDENCIES=ON` to prevent fetching from github
   - If only fetching from github is desired then set `OTELCPP_FORCE_FETCH_DEPENDENCIES=ON`
- Update github ci workflow to test fetching all dependencies from github (no git submodules).
  -  Clean up the googletest install script to build with the selected cxx standard and no longer install curl, zlib,and nlohmann-json. 
- Add the ms-gsl dependency to the api component if `WITH_GSL=ON` and test in ci

### DIFF
--- a/.github/workflows/cmake_install.yml
+++ b/.github/workflows/cmake_install.yml
@@ -19,6 +19,7 @@ jobs:
       CMAKE_VERSION: '3.31.6'
       # cxx17 is the default for windows-2022
       CXX_STANDARD: '17'
+      CMAKE_EXTRA_ARGS: "-DWITH_STL=CXX17 -DOTELCPP_REQUIRE_LOCAL_DEPENDENCIES=ON"
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
@@ -45,6 +46,7 @@ jobs:
       CMAKE_VERSION: '3.15.0'
       # cxx14 is the default for windows-2019
       CXX_STANDARD: '14'
+      CMAKE_EXTRA_ARGS: "-DWITH_GSL=ON -DOTELCPP_REQUIRE_LOCAL_DEPENDENCIES=ON"
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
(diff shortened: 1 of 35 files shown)


ThomsonTan:
Why do we need add a dedicated function `otel_add_external_project` for third party, could just use `otel_add_component`?

dbarker:
> Why do we need add a dedicated function `otel_add_external_project` for third party, could just use `otel_add_component`?

@ThomsonTan They serve two different purposes. 

The `otel_add_external_project` function wraps the CMake `find_package` and `FetchContent_*` functions and sets some properties needed for install. The rationale for this is to unify support for importing the required targets by finding packages or fetching from git submodules/repos.

The `otel_add_component` function supports grouping otel-cpp targets to create otel-cpp components for install/packaging and this step depends on having valid third party dependency targets imported.

dbarker:
Closing this PR in favor of breaking up the changes into multiple PRs: 
1. Adding a third-party install cmake project and script (#3486)
2. Find/Fetch googletest and benchmark (#3485)
3. Find/Fetch prometheus-cpp and cleanup to explicitly link core and pull targets (#3522)  
4. Find/Fetch Curl and find zlib (#3526)
5. Find/Fetch OpenTracing and support the opentracing-static target (#3518) 
6. Find/Fetch nlohmann-json and cmake cleanup (#3523)
7. Find/Fetch ms-gsl and test (#3521)
8. Find/Fetch Protobuf/Grpc (#3533) 
9. Setting PROJECT_VERSION for opentelemetry-cpp and cache OPENTELEMETRY_* variables for in-tree builds (#3605)

TODO:
- Fix API only cmake builds (api only must override all other options) and test

Issue #2481, opened 2024-01-10
[CONFIG] Implement file configuration


marcalff:
The spec for file configuration is shaping up, opentelemetry-cpp needs to start planning for this.

* https://github.com/open-telemetry/oteps/pull/225
* https://github.com/open-telemetry/opentelemetry-specification/pull/3805
* https://github.com/open-telemetry/opentelemetry-specification/pull/3804

Spec:  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/file-configuration.md

PR #3467, opened 2025-06-09
[CONFIGURATION] File configuration - trace model

marcalff:
Contributes to #2481

This is a partial fix, that implements supporting declarations.

## Changes

* Implement the trace model for file configuration.


sdk/include/opentelemetry/sdk/configuration/configuration.h
```diff
@@ -0,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <string>
```
dbarker, line 7: I think `<memory>` is missing due to std::unique_ptr being used.  

~~If possible we can add the code from these PRs to the iwyu and clang-tidy jobs. Showing they add no new warnings should speed up review quite a bit~~.  (edit - iwyu is being run in the full PR)

sdk/include/opentelemetry/sdk/configuration/configuration.h
```diff
@@ -0,0 +1,70 @@
...
+  std::unique_ptr<AttributeLimitsConfiguration> attribute_limits;
+  std::unique_ptr<LoggerProviderConfiguration> logger_provider;
+  std::unique_ptr<MeterProviderConfiguration> meter_provider;
+  std::unique_ptr<PropagatorConfiguration> propagator;
+  std::unique_ptr<TracerProviderConfiguration> tracer_provider;
+  std::unique_ptr<ResourceConfiguration> resource;
+  // Ignored: instrumentation
+
+private:
+  std::unique_ptr<Document> m_doc;
```
dbarker: nit - `m_doc` -> `doc_` to be consistent with the convention used elsewhere in ./sdk

sdk/include/opentelemetry/sdk/configuration/otlp_file_span_exporter_configuration.h
```diff
@@ -0,0 +1,36 @@
...
+
+// YAML-SCHEMA: schema/common.json
+// YAML-NODE: ExperimentalOtlpFileExporter
+class OtlpFileSpanExporterConfiguration : public SpanExporterConfiguration
+{
+public:
+  OtlpFileSpanExporterConfiguration()           = default;
+  ~OtlpFileSpanExporterConfiguration() override = default;
+
+  void Accept(SpanExporterConfigurationVisitor *visitor) const override
```
dbarker, line 23: q: should these methods be `noexcept` or do we expect to throw during configuration?

sdk/include/opentelemetry/sdk/configuration/configuration.h
marcalff: I don't understand the comment itself, what is missing ?

The include-what-you-use build is covered in PR #2518 and is clean on the same file.

It is not feasible to have include-what-you-use here, because the headers added are not used yet.

dbarker: updated the comment (missing `<memory>`)

marcalff: It seems include-what-you-use has trouble with header only classes, because there is no foo.cc to trigger the check for foo.h.

I am making a general pass to add string, memory, vector when needed, thanks for spotting this.

dbarker: Yeah it is running on the builder and should be checking this file. Maybe there is a setting we need to change in iwyu? 
https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/15534054991/job/43729025261#step:8:1248

sdk/include/opentelemetry/sdk/configuration/otlp_http_span_exporter_configuration.h
```diff
@@ -0,0 +1,44 @@
...
+
+  std::string endpoint;
+  std::string certificate_file;
+  std::string client_key_file;
+  std::string client_certificate_file;
+  std::unique_ptr<HeadersConfiguration> headers;
+  std::string headers_list;
+  std::string compression;
+  size_t timeout{0};
+  enum_otlp_http_encoding encoding{protobuf};
```
dbarker, line 39: nit - this should follow the Google naming convention to match the code in sdk.
```.h
enum enum_otlp_http_encoding
{
  protobuf,
  json
};
```
to
```.h
enum class EnumOtlpHttpEncoding
{
  kProtobuf,
  kJson
};
```

sdk/include/opentelemetry/sdk/configuration/configuration.h
marcalff: Missing headers added.

sdk/include/opentelemetry/sdk/configuration/otlp_file_span_exporter_configuration.h
marcalff: When translating the C++ representation of yaml node `xyz` in file `config.yaml` (as in, C++ `XyzConfiguration objects`) to objects in the SDK proper:

* Some feature might be totally missing in opentelemetry-cpp (not supported)
* Some feature might be missing due to compilation options (WITH_XYZ=OFF)

so exceptions will happen in these cases.

No change, as we expect exceptions.

sdk/include/opentelemetry/sdk/configuration/otlp_http_span_exporter_configuration.h
marcalff: It is preferable to match the naming conventions used in the yaml schema itself, as defined by the configuration repository:

* https://github.com/open-telemetry/opentelemetry-configuration/tree/main/schema

This helps, because each C++ class is a direct representation of the corresponding yaml node, so maintaining the C++ code in sync gets easier.

At some point later, when creating SDK objects from the configuration classes, there will be some translation, like from enum_otlp_http_encoding to EnumOtlpHttpEncoding in this case.

Also, keep in mind that the configuration repository might introduce new features (in yaml) that are not supported in the opentelemetry SDK, like a third enum entry for example.

This is possible:

```
enum enum_otlp_http_encoding
{
  protobuf,
  json,
  new_cool_feature
};
```

```
enum class EnumOtlpHttpEncoding
{
  kProtobuf,
  kJson
};
```

enum `enum_otlp_http_encoding` is strictly to represent the yaml schema, regardless of what the SDK implementation supports in `EnumOtlpHttpEncoding`.

These two are related, but different.

sdk/include/opentelemetry/sdk/configuration/configuration.h
marcalff: Fixed.

dbarker (approved):
LGTM

sdk/include/opentelemetry/sdk/configuration/configuration.h
```diff
@@ -0,0 +1,71 @@
...
+  bool disabled{false};
+  std::string log_level;
+
+  std::unique_ptr<AttributeLimitsConfiguration> attribute_limits;
+  std::unique_ptr<LoggerProviderConfiguration> logger_provider;
+  std::unique_ptr<MeterProviderConfiguration> meter_provider;
+  std::unique_ptr<PropagatorConfiguration> propagator;
+  std::unique_ptr<TracerProviderConfiguration> tracer_provider;
+  std::unique_ptr<ResourceConfiguration> resource;
+  // Ignored: instrumentation
```
lalitb, line 67: Why is this ignored?

sdk/include/opentelemetry/sdk/configuration/batch_span_processor_configuration.h
```diff
@@ -0,0 +1,41 @@
...
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+// YAML-SCHEMA: schema/tracer_provider.json
+// YAML-NODE: BatchSpanProcessor
```
lalitb, line 20: Should this configuration class be generated automatically from the schema? Given that the class structure closely mirrors the schema, automating this could help reduce manual maintenance and ensure it stays in sync as the schema evolves.

sdk/include/opentelemetry/sdk/configuration/configuration.h
marcalff: The instrumentation yaml node is not stable yet, it is still in development.

Also, this is to provide data to auto instrumentation in each languages ... C++ does not support any that I know of.

sdk/include/opentelemetry/sdk/configuration/batch_span_processor_configuration.h
marcalff: I considered to generate automatically some code, but decided against it:

* The schema mostly can be automated, but there are still some parts that need to be expressed explicitly. For example, the fact that an "exporter" can be one of otlp_http, otlp_grpc or otlp_file is represented by inheritance in C++, while it is a one-of list in yaml. Convincing an automated tool to map a `yaml` model to a `C++` model carries a lot of complexity of its own, and is not trivial.
* Still on the exporter example, introducing C++ classes for extensions points, which are by definition not in the yaml model, will be even more challenging.

Also to consider:

* Today, the yaml to C++ mapping looks simple, but it will not stay this way.
* Later, some yaml format will decide that property xxx is required in schema version A, but optional in schema version B. Property yyy will be accepted, but only starting with version C. When that happens, the C++ classes needed to represent the yaml schema as a whole will no longer be a 1-to-1 translation of a given version of the schema. The parsing code will then need to look at the schema version to parse yaml directly.
* Having C++ classes automatically generated will become a problem then, because it lacks some flexibility.

And to clarify, we do not want to generate different C++ classes for different schema versions, as this will duplicate the entire code base (yaml parsing, SDK instantiation) for each schema version.

The current proposal looks like the best option, in my opinion, and is the most robust to expected or unexpected changes in the yaml schema.

marcalff:
@lalitb @ThomsonTan @esigo Please review

sdk/include/opentelemetry/sdk/configuration/batch_span_processor_configuration.h
lalitb: Thanks Marc, that makes sense. I agree—if the mapping between YAML and C++ is going to get more complex, trying to automate it might actually make things harder.

If the schema doesn’t change much, keeping the config class updated by hand should be fine. The only risk is missing a change and the class getting out of sync with the schema, but as long as we’re careful, I think the current approach works.

sdk/include/opentelemetry/sdk/configuration/span_limits_configuration.h
```diff
@@ -0,0 +1,29 @@
...
+// YAML-NODE: SpanLimits
+class SpanLimitsConfiguration
+{
+public:
+  size_t attribute_value_length_limit;
+  size_t attribute_count_limit;
+  size_t event_count_limit;
+  size_t link_count_limit;
+  size_t event_attribute_count_limit;
+  size_t link_attribute_count_limit;
```
lalitb: I am not sure if otel-cpp implementation supports all these configurations. Do we still need to add them?

marcalff: The goal is for the yaml parser to translate all the yaml syntax into C++ model data (the *_configuration classes), so I would like to have complete coverage there, to make sure the parser understand the full syntax.

Now, down the line, when it is time to instantiate a SDK object from the model, of course there will be some features not supported by the opentelemetry-cpp SDK, so that warnings or errors will be raised then.

That part of the code (create SDK objects from the model) is not in the scope for this PR, it will be done in PRs to come that consume the model.

PR #3469, opened 2025-06-09
Increase code coverage of iwyu and clang-tidy ci jobs

dbarker:
## Changes

- Turns on abiv2 and all exporter options for iwyu and clang-tidy

### DIFF
--- a/.github/workflows/iwyu.yml
+++ b/.github/workflows/iwyu.yml
@@ -62,9 +62,18 @@ jobs:
             -DWITH_OTLP_FILE=ON \
             -DWITH_OPENTRACING=ON \
             -DWITH_OTLP_HTTP_COMPRESSION=ON \
-            -DWITH_THREAD_INSTRUMENTATION=ON \
             -DWITH_ZIPKIN=ON \
             -DWITH_PROMETHEUS=ON \
+            -DWITH_ELASTICSEARCH=ON \
+            -DWITH_HTTP_CLIENT_CURL=ON \
+            -DWITH_ABI_VERSION_1=OFF \
+            -DWITH_ABI_VERSION_2=ON \
+            -DWITH_METRICS_EXEMPLAR_PREVIEW=ON \
+            -DWITH_ASYNC_EXPORT_PREVIEW=ON \
+            -DWITH_THREAD_INSTRUMENTATION_PREVIEW=ON \
+            -DWITH_OTLP_GRPC_SSL_MTLS_PREVIEW=ON \
+            -DWITH_OTLP_GRPC_CREDENTIAL_PREVIEW=ON \
+            -DWITH_OTLP_RETRY_PREVIEW=ON \
             ..
 
       - name: iwyu_tool
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -32,10 +32,15 @@ jobs:
           sudo apt install -y --no-install-recommends --no-install-suggests \
           build-essential \
           cmake \
+          zlib1g-dev \
           libssl-dev \
           libcurl4-openssl-dev \
+          nlohmann-json3-dev \
+          libabsl-dev \
           libprotobuf-dev \
+          libgrpc++-dev \
           protobuf-compiler \
+          protobuf-compiler-grpc \
           libgmock-dev \
           libgtest-dev \
           libbenchmark-dev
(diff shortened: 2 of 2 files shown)


.github/workflows/iwyu.yml
```diff
@@ -87,7 +96,7 @@ jobs:
           readonly WARNING_COUNT=`grep -c "include-what-you-use reported diagnostics:" iwyu.log`
           echo "include-what-you-use reported ${WARNING_COUNT} warning(s)"
           # Acceptable limit, to decrease over time down to 0
-          readonly WARNING_LIMIT=0
+          readonly WARNING_LIMIT=48
```
lalitb: Are these warnings coming from ABIv2 and default disabled features, and the goal is to have it 0 eventually?

dbarker: Yeah these 48 warnings are coming from code under #ifdefs and turned on with the cmake options added. Targeting zero warnings makes sense.

lalitb: Thanks for confirming. Seems there is indeed a risk that clang-tidy won't analyze code in the #else branches of #ifdef blocks when we enable all those CMake options. If that's the case, should we have two different clang-tidy runs - one with features enabled and another with features disabled?

Not a blocker for this PR - just thinking out loud.

dbarker: Multiple runs makes sense. How about 1) abiv1 no preview features, 2) abiv2 no preview features, 3) abiv2 with all preview features?

marcalff: For clang-tidy, ok to have only one run in CI, with maximum code exposure (all preview flags on). No strong opinion about ABIv1 vs ABIv2.

Please note, however, that the clang-tidy warnings are generated ... and not looked at, the CI is in place to detect issues, the but cleanup still needs to happen.

For include-what-you-use, ABIv1 is stable while ABIv2 is in development, so we absolutely need coverage for ABIv1 to make sure it is spotless.

I suggest:

* iwyu ABIv1 with no preview
* iwyu ABIv1 with all preview
* iwyu ABIv2 with all preview
* clang-tidy ABIv1 with all preview

There is no need to have more clang-tidy combinations while cleanup has not started yet.

To consider separately, I think we should promote *_PREVIEW flags to stable and decrease the number of combinations here.

marcalff: > Yeah these 48 warnings are coming from code under #ifdefs and turned on with the cmake options added. Targeting zero warnings makes sense.

Exposing more code leads to more warnings to fix.

I just fixed 121 issues in grpc, see #3465.

dbarker: > I suggest:
> 
> iwyu ABIv1 with no preview
> iwyu ABIv1 with all preview
> iwyu ABIv2 with all preview
> clang-tidy ABIv1 with all preview

Sounds like a good plan. I'll move this to draft to implement that matrix. 

> To consider separately, I think we should promote *_PREVIEW flags to stable and decrease the number of combinations here.

This makes sense. We can consider creating a policy that after some number of releases we promote the feature out of preview. 

> Please note, however, that the clang-tidy warnings are generated ... and not looked at, the CI is in place to detect issues, the but cleanup still needs to happen.

For clang-tidy warnings we can start enabling warnings as errors at the folder (or target level if needed) and in a release or so have warnings as errors enabled for the whole code base. There are some good ones in there to fix :)

dbarker: I've added the matrix runs. There was a slight difference in clang-tidy warnings between abiv1 and abiv2 so I left that job as a matrix to keep testing both. The clang-tidy jobs will fail if new warnings are added beyond the limit so these should provide value. 

Ready again for review :)

test_common/cmake/component-options.cmake
```diff
@@ -0,0 +1,27 @@
...
+if(NOT DEFINED ENABLE_COMPONENTS)
+    set(ENABLE_COMPONENTS ON)
+endif()
+
+set(WITH_OTLP_GRPC ${ENABLE_COMPONENTS} CACHE BOOL "" FORCE)
+set(WITH_OTLP_HTTP ${ENABLE_COMPONENTS} CACHE BOOL "" FORCE)
+set(WITH_OTLP_FILE ${ENABLE_COMPONENTS} CACHE BOOL "" FORCE)
+
+set(WITH_OTLP_HTTP_COMPRESSION ${ENABLE_COMPONENTS} CACHE BOOL "" FORCE)
+set(WITH_CURL_LOGGING ${ENABLE_COMPONENTS} CACHE BOOL "" FORCE)
```
marcalff: `WITH_OTLP_HTTP_COMPRESSION` and `WITH_CURL_LOGGING` are poorly named, these are feature flag _PREVIEW options.

Move to preview-options.cmake

marcalff:
@dbarker See minor comment about preview options.

test_common/cmake/component-options.cmake
dbarker: Thanks Marc. I've pushed a fix.

PR #3486, opened 2025-06-17
[CMAKE] Add thirdparty install cmake project and install bash script

dbarker:
The third-party package installs for ci testing have a few challenges/limitations: 
1. The packages are installed by four bash scripts and creating consistent builds (common cxx standard, cmake args, ...) can be challenging. 
2. Versions are set by script defaults or environment variables in ci yml and don't use the git tags from `third_party_release`. 
3. Creating compatible installs of abseil, protobuf, and grpc can be challenging. Script input args need to be correct to avoid creating duplicate/conflicting builds of abseil and protobuf with grpc. 

The goal of this PR is to make managing/maintaining third party dependencies for CI testing easier. 

## Changes

- creates a `opentelemetry-cpp-thirdparty-install` CMake project and adds tags files for minimum supported, "stable", and latest versions. 
- creates a install_thirdparty.sh script for Linux
- use the third-party install project/script in the devcontainer and ci jobs that test otlp components. 


.devcontainer/Dockerfile.dev
```diff
@@ -46,14 +36,17 @@ ENV USER_UID=${USER_UID}
...
 
+COPY install /opt/install
 COPY ./.devcontainer/customize_container.sh /tmp/opentelemetry_cpp/devcontainer/customize_container.sh
 RUN /tmp/opentelemetry_cpp/devcontainer/customize_container.sh
 RUN apt install -y npm && npm install -g markdownlint-cli@0.44.0
 
 USER devuser
 
 WORKDIR  /workspaces/opentelemetry-cpp
+RUN cd /opt && bash ci/install_thirdparty.sh --install-dir /home/devuser/third-party/install-stable --tags-file install/cmake/third_party_stable
```
dbarker, line 47: Installing the dependencies with user privileges and under the /home/devuser directory then setting CMAKE_PREFIX_PATH makes it easier for developers to change/test versions of dependencies in the devcontainer.

.github/workflows/ci.yml
```diff
@@ -84,32 +82,25 @@ jobs:
   cmake_gcc_maintainer_sync_test:
     name: CMake gcc 14 (maintainer mode, sync)
     runs-on: ubuntu-24.04
+    env:
```
dbarker, line 85: Moving the environment variables to the job level ensures the dependencies and otel-cpp are built with the same c++ standard and compiler.

.github/workflows/ci.yml
```diff
@@ -951,7 +911,7 @@ jobs:
 
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
     - name: setup
-      run: sudo apt remove needrestart && sudo ./ci/install_format_tools.sh #refer: https://github.com/actions/runner-images/issues/9937
+      run: sudo ./ci/install_format_tools.sh
```
dbarker, line 914: The `needrestart` package no longer needs to be removed. Opportunistic cleanup.

install/cmake/third_party_latest
```diff
@@ -0,0 +1,18 @@
+# Copyright The OpenTelemetry Authors
```
dbarker, line 1: There are four files now with git tags for third-party repositories. The intent is to move all third party versions out of github workflow yml files into these tag files.  
- `third_party_release`
- `install/cmake/third_party_minimum`
- `install/cmake/third_party_stable`
- `install/cmake/third_party_latest`

In the future we can consider maintaining  `third_party_release` as the "latest" versions while keeping the `third_party_minimum` version file to test the range supported. Then it seems reasonable to remove the "stable" and "latest" files.

third_party_release
```diff
@@ -4,20 +4,24 @@
...
-#
+#  - manually edit the release git tag below.
+#  - update the matching git submodule to the new tag.
+#  - update the Bazel build files
+#     - opentelemetry-cpp/MODULE.bazel
+#     - opentelemetry-cpp/bazel/repository.bzl
 
-gRPC=v1.49.2
-abseil=20240116.1
+abseil=20220623.2
```
dbarker, line 17: Updated this file to add the missing dependencies.  The intent is to not upgrade anything in this file with this PR. 
- Git tags for curl and zlib were selected with no strong rationale (interested in feedback on what these should be). 
- Git tags for abseil and protobuf were selected based on the grpc submodule versions at the grpc tag.

marcalff (approved):
LGTM, nice work.

PR #3493, opened 2025-06-24
[CodeHealth] Fix clang-tidy warnings part 1

dbarker:
Contributes to #2053

## Changes

Fix some clang-tidy warnings: 
- metric sdk (aggregation)
- logs sdk (processors and tests)
- otlp file client
- otlp http client
- misc-unused-using-decls warnings
- performance-avoid-endl warnings
- cppcoreguidelines-avoid-const-or-ref-data-members warnings for const data members

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 215
+            warning_limit: 172
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 216
+            warning_limit: 173
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@002fdce3c6a235733a90a27c80493a3241e56863 # v2.12.1


sdk/src/metrics/aggregation/lastvalue_aggregation.cc
```diff
@@ -50,12 +47,12 @@ std::unique_ptr<Aggregation> LongLastValueAggregation::Merge(
       nostd::get<LastValuePointData>(delta.ToPoint()).sample_ts_.time_since_epoch())
   {
     LastValuePointData merge_data = nostd::get<LastValuePointData>(ToPoint());
-    return std::unique_ptr<Aggregation>(new LongLastValueAggregation(std::move(merge_data)));
+    return std::unique_ptr<Aggregation>(new LongLastValueAggregation(merge_data));
```
dbarker, line 50: Fixes: 
```bash
warning: std::move of the variable 'merge_data' of the trivially-copyable type 'LastValuePointData' has no effect [performance-move-const-arg]
```

sdk/include/opentelemetry/sdk/metrics/aggregation/lastvalue_aggregation.h
```diff
@@ -43,16 +42,15 @@ class DoubleLastValueAggregation : public Aggregation
 {
 public:
   DoubleLastValueAggregation();
-  DoubleLastValueAggregation(LastValuePointData &&);
```
dbarker, line 46: `LastValuePointData` is trivially-copyable and std::move has no effect.

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -1588,11 +1593,9 @@ class OPENTELEMETRY_LOCAL_SYMBOL OtlpFileOstreamBackend : public OtlpFileAppende
 public:
   explicit OtlpFileOstreamBackend(const std::reference_wrapper<std::ostream> &os) : os_(os) {}
 
-  ~OtlpFileOstreamBackend() override {}
```
dbarker, line 1591: Fixes 
```bash
warning: class 'OtlpFileOstreamBackend' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]
```

sdk/include/opentelemetry/sdk/metrics/aggregation/sum_aggregation.h
```diff
@@ -22,7 +22,6 @@ class LongSumAggregation : public Aggregation
 {
 public:
   LongSumAggregation(bool is_monotonic);
-  LongSumAggregation(SumPointData &&);
```
dbarker, line 25: `SumPointData` is trivially-copyable and std::move has no effect.

sdk/src/metrics/aggregation/lastvalue_aggregation.cc
owent: I bellieve it's a mistake of clang. But it's OK to copy `LastValuePointData` because it's small and will not cost a lot.

owent (approved):

sdk/src/metrics/aggregation/lastvalue_aggregation.cc
dbarker: Thanks for the review. The class is trivially copyable, but we could consider disabling this warning with clang-tidy and keep the std::move. I'd prefer to keep the warning enabled and remove the moves that don't have any effect.

marcalff (approved):
LGTM, thanks for the cleanup.

PR #3495, opened 2025-06-24
[EXPORTER] Add bytes support for OTLP recordables

### LINKED ISSUE #3487: Support bytes_value type for LogRecord Body
I see that Log Record body cannot take in bytes_value.

void OtlpLogRecordable::SetBody(const opentelemetry::common::AttributeValue &message) noexcept
{
  OtlpPopulateAttributeUtils::PopulateAnyValue(proto_record_.mutable_body(), message);
}

I see that OtlpPopulateAttributeUtils::PopulateAnyValue doesn't call set_bytes_value for any data type.
But, the otlp proto has this option. 

I was planning to send another protobuf as log body in Otel Resource Logs. I tried from golang client and it's working. But wanted the same in cpp.

I would like to have the same behaviour from cpp clients as well.

Trying to enhance the server side to accept array_value as cpp clients can send array_value.

owent:
Fixes #3487

## Changes

- Add bytes support for OTLP recordables.

> The API interface support long time ago, but after the protos are updated, we seem to forget to implement it.
> See <https://github.com/open-telemetry/opentelemetry-specification/issues/780> and <https://github.com/open-telemetry/opentelemetry-proto/pull/297> for details.


lalitb:
Unfortunately, as a side effect, this change also enables support for populating byte/uint8 types as span attribute values, which is inconsistent with the specs:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute

Per the spec, an attribute value MUST be one of the following:

- string
- boolean
- double (IEEE 754)
- signed 64-bit integer
- Or an array of the above types (array must be homogeneous)

Supporting raw bytes or uint8 is not permitted for span attributes.

It’s better to enable populating uint8_t as OTLP attributes or body only for logs, not for spans.
Alternatively, instead of this PR, document that byte data should be encoded as base64 and set as a string attribute. The custom exporter or backend can handle base64 decoding if needed.

lalitb (changes requested):
As mentioned - https://github.com/open-telemetry/opentelemetry-cpp/pull/3495#issuecomment-3001498137

owent:
> Unfortunately, as a side effect, this change also enables support for populating byte/uint8 types as span attribute values, which is inconsistent with the specs:
> 
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute
> 
> Per the spec, an attribute value MUST be one of the following:
> 
> * string
> * boolean
> * double (IEEE 754)
> * signed 64-bit integer
> * Or an array of the above types (array must be homogeneous)
> 
> Supporting raw bytes or uint8 is not permitted for span attributes.
> 
> It’s better to enable populating uint8_t as OTLP attributes or body only for logs, not for spans. Alternatively, instead of this PR, document that byte data should be encoded as base64 and set as a string attribute. The custom exporter or backend can handle base64 decoding if needed.

Thanks, only log record use bytes now.

lalitb (approved):
Thanks.

PR #3496, opened 2025-06-24
[CodeHealth] Fix clang-tidy warnings part 2

dbarker:
Contributes to #2053

## Changes

Fixes warnings in the following categories: 
- bugprone-signed-char-misuse
- bugprone-suspicious-missing-comma
- bugprone-forward-declaration-namespace
- bugprone-use-after-move 
- cppcoreguidelines-explicit-virtual-functions
- cppcoreguidelines-rvalue-reference-param-not-moved
- cppcoreguidelines-use-default-member-init
- misc-unused-parameters
- performance-move-const-arg
- performance-unnecessary-value-param
- performance-avoid-endl


exporters/otlp/src/otlp_grpc_client.cc
```diff
@@ -519,31 +522,40 @@ OtlpGrpcClient::MakeLogsServiceStub()
 grpc::Status OtlpGrpcClient::DelegateExport(
     proto::collector::trace::v1::TraceService::StubInterface *stub,
     std::unique_ptr<grpc::ClientContext> &&context,
-    std::unique_ptr<google::protobuf::Arena> && /*arena*/,
+    std::unique_ptr<google::protobuf::Arena> &&arena,
     proto::collector::trace::v1::ExportTraceServiceRequest &&request,
     proto::collector::trace::v1::ExportTraceServiceResponse *response)
 {
-  return stub->Export(context.get(), request, response);
+  auto context_ptr = std::move(context);
```
dbarker: I'm not thrilled with this fix. The alternative I've considered is to change the interface of `DelegateExport`to: 

````.h
grpc::Status OtlpGrpcClient::DelegateExport(
    proto::collector::trace::v1::TraceService::StubInterface *stub,
    std::unique_ptr<grpc::ClientContext> &context,
    const proto::collector::trace::v1::ExportTraceServiceRequest &request,
    proto::collector::trace::v1::ExportTraceServiceResponse *response)
````
- context becomes a non-const reference
- remove passing in arena since it is not used
- request becomes a const reference

File: sdk/test/logs/logger_provider_sdk_test.cc [hunk 2]
```diff
@@ -224,6 +224,7 @@ class DummyProcessor : public LogRecordProcessor
     return std::unique_ptr<Recordable>(new DummyLogRecordable());
   }
 
+  // NOLINTNEXTLINE(cppcoreguidelines-rvalue-reference-param-not-moved)
   void OnEmit(std::unique_ptr<Recordable> && /* record */) noexcept override {}
```
marcalff: Instead of ignoring the moved parameter, maybe implement as:

auto record_ptr = std::move(record);

I assume this will be more correct, and prevent leaks in valgrind runs for example.

(This applies to other places as well)

exporters/otlp/src/otlp_grpc_client.cc
dbarker: Looking into this more the interface for the grpc client will need to change. In some cases the arena is moved to a scope that would have it destroyed before the messages (request and response) it created. This shouldn't be allowed. I'll revert the grpc client/exporter warning fixes in this PR and open a separate PR for the otlp grpc fixes which will have changes to its method interfaces.

sdk/test/logs/logger_provider_sdk_test.cc
dbarker: Thanks Marc. Fixed now.

exporters/otlp/src/otlp_grpc_client.cc
dbarker: cppcoreguidelines-rvalue-reference-param-not-moved warning fixes to otlp grpc client/exporters have been reverted. Will link the new PR when posted.

marcalff (approved):
LGTM, thanks for the cleanup.

examples/otlp/http_instrumented_main.cc
```diff
@@ -103,46 +103,45 @@ class MyThreadInstrumentation : public opentelemetry::sdk::common::ThreadInstrum
     {
       std::cout << ", priority " << priority_;
     }
-    std::cout << std::endl << std::flush;
+    std::cout << "\n" << std::flush;
```
ThomsonTan, line 106: Why is `\n` preferred than `std::endl`?

dbarker: `std::endl` is slow because it forces an immediate flush of the output buffer every time it's used. `\n` just adds a newline character to the buffer, which is faster as it avoids frequent system calls. 

We currently have the clang-tidy warning category `performance-avoid-endl` enabled it warns on use of `std::endl`.

PR #3501, opened 2025-06-28
[CodeHealth] Fix clang tidy warnings part 4

dbarker:

## Changes
Address clang-tidy warning categories
- google-runtime-int (nolint used for curl operations with long)
- google-readability-casting
- cppcoreguidelines-macro-to-enum
- cppcoreguidelines-prefer-member-initializer
- performance-unnecessary-copy-initialization

### DIFF
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter_options.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter_options.h
@@ -65,10 +65,10 @@ struct OPENTELEMETRY_EXPORT OtlpHttpExporterOptions
 
 #ifdef ENABLE_ASYNC_EXPORT
   /** Max number of concurrent requests. */
-  std::size_t max_concurrent_requests;
+  std::size_t max_concurrent_requests{64};
 
   /** Max number of requests per connection. */
-  std::size_t max_requests_per_connection;
+  std::size_t max_requests_per_connection{8};
 #endif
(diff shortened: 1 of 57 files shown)


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter_options.h
```diff
@@ -65,10 +65,10 @@ struct OPENTELEMETRY_EXPORT OtlpHttpExporterOptions
 
 #ifdef ENABLE_ASYNC_EXPORT
   /** Max number of concurrent requests. */
-  std::size_t max_concurrent_requests;
+  std::size_t max_concurrent_requests{64};
```
marcalff: With this change, the default value 64 moved from the cc file to the header file.

It then becomes part of the application binary, when building with the SDK headers.

The day the SDK and exporter becomes a shared library (we are not there yet), this will be a problem.

Could the member initialization stay in the *.cc file instead, using `max_concurrent_requests(64)` in the constructor implementation ?

dbarker: Good catch. Fixed.

marcalff (approved):
LGTM.

This:
```
// NOLINTNEXTLINE(google-runtime-int)
```
is indeed the best fix for curl.

Changing the code to please the google style breaks curl on 32bits machines.

PR #3506, opened 2025-06-30
[CodeHealth] Fix clang-tidy warnings part 5

dbarker:

## Changes

Fixes clang-tidy warnings with the OTLP File component
- cppcoreguidelines-slicing
- cppcoreguidelines-avoid-const-or-ref-data-members
- cppcoreguidelines-special-member-functions

### DIFF
--- a/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_file_client.h
+++ b/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_file_client.h
@@ -40,6 +40,10 @@ public:
                           OtlpFileClientRuntimeOptions &&runtime_options);
 
   ~OtlpFileClient();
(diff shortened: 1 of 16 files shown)


marcalff (approved):
LGTM

marcalff:
According to CI, this solves only 1 warning (85 instead of 86).

To reevaluate once part 4 is merged.

dbarker:
> According to CI, this solves only 1 warning (85 instead of 86).
> 
> To reevaluate once part 4 is merged.

Yes. It may be due to using an older version of clang-tidy in CI. So we won't see the satisfying drop in warnings from this but they do need to be fixed. :) 

I found the warnings using the clangd extension to vscode while fixing the slicing warning. CI only reports the slicing warning for some reason.

PR #3507, opened 2025-07-01
[CodeHealth] Fix clang-tidy warnings part 6

dbarker:

## Changes

Fix clang-tidy warnings in otlp grpc and otlp http classes: 
- cppcoreguidelines-avoid-const-or-ref-data-members
- cppcoreguidelines-special-member-functions
- cppcoreguidelines-rvalue-reference-param-not-moved
- performance-unnecessary-value-param


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -64,14 +71,14 @@ struct OtlpGrpcClientOptions
   std::string user_agent;
 
   /** max number of threads that can be allocated from this */
-  std::size_t max_threads;
+  std::size_t max_threads{};
```
dbarker, line 74: some values in this class were not initialized. The initial values are overridden by constructors of the derived classes.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -230,28 +234,15 @@ class OtlpHttpClient
     std::shared_ptr<opentelemetry::ext::http::client::Session> session;
     std::shared_ptr<opentelemetry::ext::http::client::EventHandler> event_handle;
 
-    inline HttpSessionData() = default;
+    HttpSessionData() = default;
```
dbarker, line 237: inline keywords and the explicit move constructor and move operator were removed. Moves will use the implicit default implementations.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -230,28 +234,15 @@ class OtlpHttpClient
...
 
-    inline explicit HttpSessionData(
-        std::shared_ptr<opentelemetry::ext::http::client::Session> &&input_session,
-        std::shared_ptr<opentelemetry::ext::http::client::EventHandler> &&input_handle)
+    explicit HttpSessionData(
```
dbarker, line 239: The arguments for the explicit constructor were never moved and created warnings when passed as rvalue references. Changed to pass by value for the shared ptrs.  This fixes  cppcoreguidelines-rvalue-reference-param-not-moved warnings

marcalff (approved):
LGTM

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
owent: I think it's preferable to change the constructor to `HttpSessionData(std::shared_ptr<opentelemetry::ext::http::client::Session> &&input_session, std::shared_ptr<opentelemetry::ext::http::client::EventHandler> &&input_handle): session(std::move(input_session)), event_handle(std::move(input_handle)) {}` rather than simply removing the rvalue reference. Otherwise, updates to the internal atomic counter could lead to a tenfold increase in CPU usage due to cache line misses.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_metric_exporter_options.h
```diff
@@ -34,7 +34,6 @@ namespace otlp
 struct OPENTELEMETRY_EXPORT OtlpHttpMetricExporterOptions
 {
   OtlpHttpMetricExporterOptions();
```
owent, line 36: Explicit constructors and destructors should be retained for classes declared with `OPENTELEMETRY_EXPORT` to ensure the symbol is defined in only one unit, thereby reducing linker load and size of symbol table.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
dbarker: Thanks for the feedback @owent. I'll make this change.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_metric_exporter_options.h
dbarker: Thanks. I'll review this.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
dbarker: Fixed

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_metric_exporter_options.h
dbarker: I've added back the explicit default destructors that were implemented in cc files. This requires adding default move and copy constructors and assignment operators.

marcalff: @owent Please take a look and approve, to confirm if this is resolved. Thanks.

owent: Sorry for the delay.
In my understanding, using `ctor = default` or `operator=(...) = default` in headers still generate weak symbols in unix like platforms . I think we should keep it in `.cc` to prevent redundancy symbols. What do you think about it?

dbarker: > What do you think about it?

We don't have a consistent approach to this in the otlp code and it's not clear to me if the benefits (reducing linker load and size of symbol table) are worth the verbosity of explicitly defining all the special functions and creating cc files for simple options classes. For example some classes like the OtlpFileExporterRuntimeOptions don't have a cc file and use ctor=default while still declaring OPENTELEMETRY_EXPORT. 

What do you think is the right path forward?

owent (approved):

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_metric_exporter_options.h
owent: You are right. This is not a problem of this PR, I think we can do the cleanup in another one.

BTW: We worked extensively to reduce exported symbols in our large Linux project, which will significantly decrease compile and link times. Unlike some platforms, Unix-like ABIs don't use import/export declarations for classes or functions; instead, they rely on global or local (hidden) visibility. When symbols are declared with `visibility=default`, code and debug information are generated in every module that uses them, leading to duplication across dynamic libraries and potentially combined instances during linking.

marcalff: @owent 

Thanks for the context.

This looks indeed as something we can address separately.

PR #3512, opened 2025-07-01
Add bundle version of utf8_range to validate attributes

### LINKED ISSUE #3491: Need validation/debug logs for protobuf setters
**Describe your environment** 
Not related to env, any new version of otel-cpp

**Steps to reproduce**
Export span with non utf-8 attribute. 

We have a large system with hundreds of traces per second, and see intermittent error logs due to non utf-8 attributes. 

[libprotobuf ERROR google/protobuf/wire_format_lite.cc:577] String field 'opentelemetry.proto.trace.v1.Span.Event.name' contains invalid UTF-8 data when serializing a protocol buffer. Use the 'bytes' type if you intend to send raw bytes.

But it's very hard to root cause this, to help in these cases we believe otel-cpp should add validations/debug logs for at least set_string_value here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/9708acb694deb260beccb8578378bedcbbfbb1a8/exporters/otlp/src/otlp_populate_attribute_utils.cc#L41

Currently there are no checks for any of the setter functions return values and what otel-cpp does is just discard the attribute that errored and export the rest.

owent:
Fixes #3491 

## Changes

- [x] Add bundle version of utf8_range to validate attribute values
- [ ] Update the implementation for the new SPEC(2026-03)


owent:
The clang-tidy check failure does not appear to be caused by this PR.

sdk/src/common/CMakeLists.txt
```diff
@@ -1,8 +1,14 @@
...
-set(COMMON_SRCS random.cc global_log_handler.cc env_variables.cc base64.cc
-                disabled.cc)
+set(COMMON_SRCS
+    random.cc
+    global_log_handler.cc
+    env_variables.cc
+    base64.cc
+    disabled.cc
+    attribute_validity.cc
+    internal/utf8_range/uft8_range.cc)
```
dbarker, line 11: Vendoring this library and always running validation by default adds some maintenance and performance overhead. Ideally we can just get the required package/targets along with protobuf and then string validation can be tested as a preview feature and controlled by the users. 

Can we require the `utf8_range` package to be found/fetched like the other dependencies? 

The package is within protobuf source and built with protobuf. For CMake builds it can be installed using the  `utf8_range_ENABLE_INSTALL` CMake option (See [protobuf/third_party/utf8_range/CMakeLists.txt#L10](https://github.com/protocolbuffers/protobuf/blob/3d4adad5c4c4e6a6f9f038769b8c90716065b0e4/third_party/utf8_range/CMakeLists.txt#L10) ) and then found with find_package. If protobuf is fetched then we can use `FetchContent_Declare` on the source directory `${protobuf_SOURCE_DIR}/third_party/utf8_range` and build the target directly. Bazel builds relying on the protobuf packaged utf8_range also seem doable. 

What do you think?

sdk/src/resource/resource.cc
```diff
@@ -27,8 +30,28 @@ Resource::Resource(const ResourceAttributes &attributes) noexcept
...
+      OTEL_INTERNAL_LOG_WARN("[Resource] Invalid attribute key "
+                             << kv.first << ". This attribute will be ignored.");
+      continue;
+    }
+
+    if (!common::AttributeValidator::IsValid(kv.second))
+    {
+      OTEL_INTERNAL_LOG_WARN("[Resource] Invalid attribute value for "
+                             << kv.first << ". This attribute will be ignored.");
+      continue;
```
dbarker, line 49: Should these attribute values that are not valid UTF-8 sequences be set to byte values here?

Spec language ([specification/common/attribute-type-mapping.md#L109](https://github.com/open-telemetry/opentelemetry-specification/blame/854b4df2d8591af249bb12488d52238c2fde2e07/specification/common/attribute-type-mapping.md#L109)): 
> String Values
> String values which are valid UTF-8 sequences SHOULD be converted to AnyValue's [string_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L31) field.
> 
> String values which are not valid Unicode sequences SHOULD be converted to AnyValue's [bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L37) with the bytes representing the string in the original order and format of the source string.

sdk/src/trace/span.cc
```diff
@@ -106,6 +112,20 @@ void Span::SetAttribute(nostd::string_view key, const common::AttributeValue &va
...
+  if (!sdk::common::AttributeValidator::IsValid(key))
+  {
+    OTEL_INTERNAL_LOG_WARN("[Trace Span] Invalid span attribute key "
+                           << key << ". This attribute will be ignored.");
+    return;
+  }
+
+  if (!sdk::common::AttributeValidator::IsValid(value))
+  {
+    OTEL_INTERNAL_LOG_WARN("[Trace Span] Invalid span attribute value for "
```
dbarker, line 123: Same question as above: ignore or set as bytes?

sdk/include/opentelemetry/sdk/common/attribute_validity.h
```diff
@@ -0,0 +1,152 @@
...
+  OPENTELEMETRY_EXPORT static bool IsAllValid(const OrderedAttributeMap &attributes) noexcept;
+
+  OPENTELEMETRY_EXPORT static void Filter(AttributeMap &attributes, nostd::string_view log_hint);
+
+  OPENTELEMETRY_EXPORT static void Filter(OrderedAttributeMap &attributes,
+                                          nostd::string_view log_hint);
+};
+
+/**
+ * Supports internal iteration over a collection of key-value pairs and filtering of invalid
```
dbarker, line 98: If we have to place invalid utf8 sequences into byte attribute values per the spec should the IsValid check always be performed when constructing an attribute map (attribute_utils.h)?

dbarker:
Hey @owent, sorry for the long delay in review. I've added some comments above. Two main questions: 
- Can we treat `utf8_range` as a standard dependency (whose source comes with protobuf and can be installed with protobuf)?
- Do we always need to put invalid utf8 strings into byte values of attributes?

lalitb:
@owent - Is it possible to scope this fix within OTLP instead of adding SDK-wide validation/framework changes?

The original issue seems much narrower: OTLP/protobuf serialization is where invalid UTF-8 actually becomes a problem. So I’d expect a smaller fix closer to the exporter boundary, for example validating right before writing protobuf string fields / set_string_value(...), and handling invalid data there, rather than introducing a broader SDK-wide mechanism. If we go that route, I’m also not sure we need to bring in `utf8_range` in this broader way.

sdk/src/resource/resource.cc
owent: Thanks, the spec seems changed after the first version of this PR.I will change it to follow the latest spec.

sdk/src/common/CMakeLists.txt
owent: utf8_range was moved into protobuf’s source tree starting from protobuf v22, and it is only included in CMake installation when the protobuf binaries are built.
If an older protobuf version is used, or if only the libraries are built for cross-compilation, utf8_range may not be installed and therefore cannot be found by find_package.

Enforcing the use of FetchContent_Declare may also introduce ABI compatibility issues when users link our project against a prebuilt version of protobuf.

For this reason, I imported it in the same way as the bundled version of abseil-cpp.

Alternatively, do you think we can disable these checks, even though this may result in some records being dropped?

sdk/include/opentelemetry/sdk/common/attribute_validity.h
owent: Thanks, it looks like the spec changed after the initial version of this PR.
In the earlier version, bytes were allowed in log attributes, but not in traces, metrics, or resources.
I’ll review the spec again over the next few days and update the implementation to match the latest version.

owent:
> @owent - Is it possible to scope this fix within OTLP instead of adding SDK-wide validation/framework changes?
> 
> The original issue seems much narrower: OTLP/protobuf serialization is where invalid UTF-8 actually becomes a problem. So I’d expect a smaller fix closer to the exporter boundary, for example validating right before writing protobuf string fields / set_string_value(...), and handling invalid data there, rather than introducing a broader SDK-wide mechanism. If we go that route, I’m also not sure we need to bring in `utf8_range` in this broader way.

Thanks. This is also due to a change in the spec.
The latest baggage API spec (https://opentelemetry.io/docs/specs/otel/baggage/api/) also requires UTF-8 strings. Do you think we can ignore this requirement?

sdk/src/common/CMakeLists.txt
dbarker: > utf8_range was moved into protobuf’s source tree starting from protobuf v22, and it is only included in CMake installation when the protobuf binaries are built.
> If an older protobuf version is used, or if only the libraries are built for cross-compilation, utf8_range may not be installed and therefore cannot be found by find_package.

We currently have [v3.21](https://github.com/open-telemetry/opentelemetry-cpp/blob/2eba0949c8b7a452c9d4563ce7b20453995bf1a3/install/cmake/third_party_minimum#L12) as the minimum supported/tested protobuf version (it is a 4 year old release) and we can bump it to 3.22 if needed, but only supporting utf8 validation for otlp attributes if utf8_range is installed or fetched seems reasonable. 

If protobuf (v3.22+) is fetched now (by default otel-cpp will fetch protobuf [v6.31](https://github.com/open-telemetry/opentelemetry-cpp/blob/2eba0949c8b7a452c9d4563ce7b20453995bf1a3/third_party_release#L19)) the utf_range CMake targets ([`utf8_validity`](https://github.com/protocolbuffers/protobuf/blob/4b0c3aacf0657fbf38253b38918d3358dd4319ec/third_party/utf8_range/CMakeLists.txt#L20)) will be available to link to. 

> Enforcing the use of FetchContent_Declare may also introduce ABI compatibility issues when users link our project against a prebuilt version of protobuf.

The utf8_range package is independent and used as a static dependency in protobuf and will be used as a private dependency in otel-cpp OTLP exporters. ABI compatibility with protobuf may not be of concern in this case. Either way I think we use the same version bundled with protobuf.

> For this reason, I imported it in the same way as the bundled version of abseil-cpp.

Makes sense. Just raising the alternative to optionally enable UTF-8 validation when the SDK is built with OTLP support and installing/importing/fetching utf8_range along with the protobuf dependency. This could be less maintenance for otel-cpp. 

> Alternatively, do you think we can disable these checks, even though this may result in some records being dropped?

The spec on OTLP attribute type mapping ([specification/common/attribute-type-mapping.md](https://github.com/open-telemetry/opentelemetry-specification/blame/854b4df2d8591af249bb12488d52238c2fde2e07/specification/common/attribute-type-mapping.md#L109)) uses `Should` for UTF8 validation so it appears to be an optional requirement of the SDK. I think we should make it a compile time option that is off by default for now and evaluate performance/stability (similar to other _PREVIEW options).

sdk/include/opentelemetry/sdk/common/attribute_validity.h
dbarker: Looking at the spec it may only apply utf8 validation to OTLP type mapping. If that is the case we may be able to make a more isolated change to [`OtlpPopulateAttributeUtils::PopulateAnyValue`](https://github.com/open-telemetry/opentelemetry-cpp/blob/2eba0949c8b7a452c9d4563ce7b20453995bf1a3/exporters/otlp/src/otlp_populate_attribute_utils.cc#L41) to support this as @lalitb suggests.

owent:
I’m very busy these days. I’ll continue working on this next week and fully reimplement it. Sorry for the delay.

owent:
Close this PR; #3991 will reimplement the feature per the latest spec.

PR #3518, opened 2025-07-05
[CMAKE] Address the vcpkg opentelemetry-cpp port CMake patches

### LINKED ISSUE #3401: Address cmake issues requiring patches by the vcpkg port
The community vcpkg port has several patches to `opentelemetry-cpp` cmake files: 

1. Don't prefer the CONFIG search mode globally and remove the VCPKG_CHAINLOAD_TOOLCHAIN_FILE include
    - PR: https://github.com/microsoft/vcpkg/pull/41742
    - Patch: https://github.com/microsoft/vcpkg/blob/master/ports/opentelemetry-cpp/cmake-quirks.diff 
3. Support the OpenTracing static library target
    - PR: https://github.com/microsoft/vcpkg/pull/44037
    - Patch: https://github.com/microsoft/vcpkg/blob/master/ports/opentelemetry-cpp/fix-target_link.patch

   
These should be addressed so the patches can be removed.

dbarker:
Fixes #3401

Fix CMake issues requiring patches to the vcpkg opentelemetry-cpp port. 

## Changes

- Support finding or fetching OpenTracing
- Support linking to OpenTracing shared library or static library targets
- Don't set `CMAKE_FIND_PACKAGE_PREFER_CONFIG`
- Don't include `VCPKG_CHAINLOAD_TOOLCHAIN_FILE`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,6 +497,14 @@ if((NOT WITH_API_ONLY) AND USE_NLOHMANN_JSON)
   include("${opentelemetry-cpp_SOURCE_DIR}/cmake/nlohmann-json.cmake")
 endif()
 
+#
+# Do we need OpenTracing ?
+#
+
+if(WITH_OPENTRACING)
+  include("${opentelemetry-cpp_SOURCE_DIR}/cmake/opentracing-cpp.cmake")
+endif()
+
 if(OTELCPP_MAINTAINER_MODE)
   if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     message(STATUS "Building with gcc in maintainer mode.")
@@ -701,6 +709,10 @@ endif()
 if(prometheus-cpp_FOUND)
   message(STATUS "prometheus-cpp: ${prometheus-cpp_VERSION}")
 endif()
+if(WITH_OPENTRACING)
+  message(
+    STATUS "opentracing-cpp: ${OpenTracing_VERSION} (${OpenTracing_PROVIDER})")
+endif()
 message(STATUS "---------------------------------------------")
 
 include("${opentelemetry-cpp_SOURCE_DIR}/cmake/otel-install-functions.cmake")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -36,11 +36,6 @@ project(opentelemetry-cpp)
 # Mark variables as used so cmake doesn't complain about them
 mark_as_advanced(CMAKE_TOOLCHAIN_FILE)
 
-# Note: CMAKE_FIND_PACKAGE_PREFER_CONFIG requires cmake 3.15. Prefer cmake
-# CONFIG search mode to find dependencies. This is important to properly find
-# protobuf versions 3.22.0 and above due to the abseil-cpp dependency.
-set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
```
dbarker, line 42: We can remove this now that GTest, Protobuf, and gRPC are found with the CONFIG search mode explicitly.

CMakeLists.txt
```diff
@@ -701,6 +700,10 @@ endif()
 if(prometheus-cpp_FOUND)
   message(STATUS "prometheus-cpp: ${prometheus-cpp_VERSION}")
 endif()
+if(WITH_OPENTRACING)
```
marcalff, line 705: Should the test be if "opentracing_found" instead of WITH_OPENTRACING, for consistency with other tests ?

dbarker: > Should the test be if "opentracing_found" instead of WITH_OPENTRACING, for consistency with other tests ?

When building a package in tree with FetchContent the `<package>_FOUND` variable isn't set (typically only find_package sets it). 

As I add support for FetchContent with each dependency I'll change these checks to use the `WITH_*` option or to just check that the `<package>_VERSION` variable is set before logging.

CMakeLists.txt
```diff
@@ -497,6 +488,14 @@ if((NOT WITH_API_ONLY) AND USE_NLOHMANN_JSON)
   include("${opentelemetry-cpp_SOURCE_DIR}/cmake/nlohmann-json.cmake")
 endif()
 
+#
+# Do we need OpenTracing ?
```
ThomsonTan, line 492: The official vcpkg port still supports it as `opentelemetry-cpp[opentracing]`, but I think we could deprecate it, or plan on that.

PR #3519, opened 2025-07-06
[CONFIGURATION] File configuration - yaml parser

marcalff:
Contributes to #2481

This is a partial fix, that implements the yaml parser.

## Changes

* Implement the yaml parser as a wrapper on top of rapidyaml.
* Implement variables substitutions in the yaml parser.


sdk/include/opentelemetry/sdk/configuration/document_node.h
```diff
@@ -0,0 +1,167 @@
...
+  virtual void Next()                                                  = 0;
+  virtual std::string Name() const                                     = 0;
+  virtual std::unique_ptr<DocumentNode> Value() const                  = 0;
+  virtual bool Equal(const PropertiesNodeConstIteratorImpl *rhs) const = 0;
+};
+
+class DocumentNodeConstIterator
+{
+public:
+  DocumentNodeConstIterator(DocumentNodeConstIteratorImpl *impl) : impl_(impl) {}
```
dbarker: Can we use a `std::unique_ptr` for impl? This would make ownership transfer explicit in this constructor.

sdk/include/opentelemetry/sdk/configuration/document_node.h
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+private:
+  DocumentNodeConstIteratorImpl *impl_;
+};
+
+class PropertiesNodeConstIterator
+{
+public:
+  PropertiesNodeConstIterator(PropertiesNodeConstIteratorImpl *impl) : impl_(impl) {}
```
dbarker: Same as above. Can we use std::unique_ptr for impl to make transfer of ownership explicit and remove the need to manage raw pointers?

sdk/include/opentelemetry/sdk/configuration/invalid_schema_exception.h
```diff
@@ -0,0 +1,32 @@
...
+namespace sdk
+{
+namespace configuration
+{
+
+class InvalidSchemaException : public std::runtime_error
+{
+public:
+  InvalidSchemaException(const std::string &msg) : std::runtime_error(msg) {}
+  InvalidSchemaException(InvalidSchemaException &&)                      = default;
```
dbarker: nit - if all five special functions are default we can remove these declarations and use the implicitly defined implementations.

sdk/include/opentelemetry/sdk/configuration/ryml_document.h
```diff
@@ -0,0 +1,40 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <memory>
+#include <ryml.hpp>
```
dbarker, line 7: q: Will the ryml dependency be public for users (will users need to include this ryml_document.h header) ?

sdk/src/configuration/document_node.cc
```diff
@@ -0,0 +1,262 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <cctype>
```
dbarker, line 4: q: Is it reasonable to add a test covering the document_node.h to this PR? Reviewing this implementation in the context of a test would be helpful.

sdk/src/configuration/yaml_configuration_parser.cc
```diff
@@ -0,0 +1,99 @@
...
+namespace configuration
+{
+
+std::unique_ptr<Configuration> YamlConfigurationParser::ParseFile(const std::string &filename)
+{
+  std::string input_file = filename;
+
+  if (input_file.empty())
+  {
+    const char *env_var = std::getenv("OTEL_EXPERIMENTAL_CONFIG_FILE");
```
dbarker: q: should we use the common environment utilities from `sdk/common/env_variables.h` and respect the `NO_GETENV` declaration?

sdk/include/opentelemetry/sdk/configuration/ryml_document.h
marcalff: Good question. The rapid yaml parser (ryaml.hpp) is hidden in the implementation, users should never see it.

What users see is just an abstract Document with DocumentNode, hiding totally the parser implementation.

sdk/src/configuration/document_node.cc
marcalff: Adding tests for DocumentNode alone is not practical, as this is just one single yaml node in a document.

Please see parser yaml tests added in PR #3520, these cover a full document parsing, which means navigating on all the nodes of a document.

In particular, see test `yaml_test.cc` which covers all the variables substitutions, implemented in the yaml parser itself.

sdk/src/configuration/yaml_configuration_parser.cc
marcalff: Good question, I need to investigate for `NO_GETENV`.

sdk/include/opentelemetry/sdk/configuration/document_node.h
marcalff: Yes, it looks better with unique_ptr. Fixed.

sdk/include/opentelemetry/sdk/configuration/document_node.h
marcalff: Fixed.

sdk/include/opentelemetry/sdk/configuration/invalid_schema_exception.h
marcalff: Fixed.

sdk/src/configuration/yaml_configuration_parser.cc
marcalff: Now using `sdk::common::GetStringEnvironmentVariable()`.

Not really for `NO_GETENV` which is a very odd feature, but more for the platform issues on Windows.

marcalff:
@dbarker Thanks for the first round of comments, all addressed. Please take another look.

sdk/src/configuration/ryml_document.cc
```diff
@@ -0,0 +1,70 @@
...
+    OTEL_INTERNAL_LOG_ERROR("[Ryml Document] Parse failed with exception: " << e.what());
+    return doc;
+  }
+  catch (...)
+  {
+    OTEL_INTERNAL_LOG_ERROR("[Ryml Document] Parse failed with unknown exception.");
+    return doc;
+  }
+
+  RymlDocument *ryml_doc = new RymlDocument(tree);
```
dbarker: nit - we should be able to use `doc = std::make_unique<RymlDocument>(tree);` and remove explicit use of new and raw pointers here.

sdk/src/configuration/ryml_document.cc
```diff
@@ -0,0 +1,70 @@
...
+#include "opentelemetry/sdk/configuration/ryml_document_node.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+std::unique_ptr<Document> RymlDocument::Parse(const std::string &source, const std::string &content)
```
dbarker, line 24: q: Do we expect exceptions to escape this method (Should this method be marked `noexcept`)?  It looks like it is catching all the configuration exceptions and the tests seem to expect configuration parsing errors to result in the `std::unique_ptr<Document>` returned to be null instead of an exception.

dbarker, line 24: q: can the top level parse methods be marked `noexcept`?

sdk/src/configuration/ryml_document_node.cc
```diff
@@ -0,0 +1,505 @@
...
+}
+
+size_t RymlDocumentNode::AsInteger()
+{
+  OTEL_INTERNAL_LOG_DEBUG("RymlDocumentNode::AsInteger()");
+
+  if (!node_.is_val() && !node_.is_keyval())
+  {
+    OTEL_INTERNAL_LOG_ERROR("Yaml: not scalar");
+    throw InvalidSchemaException("");
```
dbarker: nit - (optional) It looks like the exception messages are more general (or empty strings) while the detailed messages a user would need to debug their configuration parse issue are logged separately at the throw site. If the exceptions are caught and their messages also logged from the top level parse methods, consider removing the throw site log message and adding the user level debugging info to the exception (which can be logged when caught).

sdk/src/configuration/ryml_document.cc
dbarker: Generally we should prefer std::make_unique over new for creating unique_ptrs per [CppCoreGuidelines#r23](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r23-use-make_unique-to-make-unique_ptrs)

sdk/include/opentelemetry/sdk/configuration/document_node.h
```diff
@@ -0,0 +1,176 @@
...
+  virtual bool GetBoolean(const std::string &name, bool default_value) = 0;
+
+  virtual std::size_t GetRequiredInteger(const std::string &name)                    = 0;
+  virtual std::size_t GetInteger(const std::string &name, std::size_t default_value) = 0;
+
+  virtual double GetRequiredDouble(const std::string &name)               = 0;
+  virtual double GetDouble(const std::string &name, double default_value) = 0;
+
+  virtual std::string GetRequiredString(const std::string &name)                           = 0;
+  virtual std::string GetString(const std::string &name, const std::string &default_value) = 0;
```
dbarker: q: Can/should any of these accessors (GetString, GetDouble, GetInteger, GetBoolean) be marked const?

sdk/src/configuration/ryml_document_node.cc
```diff
@@ -0,0 +1,505 @@
...
+
+RymlPropertiesNodeConstIteratorImpl::~RymlPropertiesNodeConstIteratorImpl() {}
+
+void RymlPropertiesNodeConstIteratorImpl::Next()
+{
+  OTEL_INTERNAL_LOG_DEBUG("RymlPropertiesNodeConstIteratorImpl::Next()");
+  ++index_;
+}
+
+std::string RymlPropertiesNodeConstIteratorImpl::Name() const
```
dbarker, line 471: q: could this return a string_view instead of allocating a string each time?

sdk/src/configuration/ryml_document_node.cc
```diff
@@ -0,0 +1,505 @@
...
+  if (value.empty())
+  {
+    OTEL_INTERNAL_LOG_ERROR("Yaml: string value is empty: " << name);
+    throw InvalidSchemaException(name);
+  }
+
+  return value;
+}
+
+std::string RymlDocumentNode::GetString(const std::string &name, const std::string &default_value)
```
dbarker: q: would returning a string_view work here or is a copy needed?

dbarker: I see the DoSubstitution creates new strings and those must be returned by value.

dbarker:
> @dbarker Thanks for the first round of comments, all addressed. Please take another look.

@marcalff Thanks for adding all the tests and updates. The test coverage looks great. I've added some new comments - most can be optional or applied later to keep the configuration PRs moving.

sdk/src/configuration/ryml_document.cc
marcalff: Fixed by using std::make_unique everywhere.

sdk/src/configuration/ryml_document.cc
marcalff: This needs further analysis.

One one hand, the application may want to handle exceptions itself, for example to print a yaml format error in a configuration UI, so to do this it should see exceptions.

On the other hand, if we (in the opentelemetry-cpp library) do not print exceptions in the log, we will never see them when someone file a bug report to say "my config.yaml does not work", making things impossible to investigate.

I would like to defer until exceptions have more context, like a position (file, line, column) and a node name in the yaml stream, so it is easier to locate errors in the yaml document, and decide then the best course of action (maybe catch, log, re throw).

sdk/src/configuration/ryml_document_node.cc
marcalff: Thanks for spotting this.

Code like `OTEL_INTERNAL_LOG_ERROR` was added very early, to debug, but it should no longer be present.

Fixed all the sites to use only throw, with more details.

sdk/src/configuration/ryml_document_node.cc
marcalff: In theory, yes, because there is a name in the rapid yaml node, so we could return a view on the data.

In practice however, `std::string_view` is C++17 only, and I would prefer not getting into `nostd::string_view`.

The yaml parsing code will be used once at application startup, to read a config file.

Once the application is up and running, this code will no longer be used, so the performance overhead is not critical for this, in my opinion.

sdk/src/configuration/ryml_document_node.cc
marcalff: Indeed. Because of susbstitution, the string data may not be available, to build a view on it.

sdk/include/opentelemetry/sdk/configuration/document_node.h
marcalff: Yes. All fixed.

dbarker (approved):
LGTM. Thanks!

PR #3520, opened 2025-07-06
[CONFIGURATION] File configuration - configuration parser

marcalff:
Contributes to #2481

This is a partial fix, that implements the configuration parser.

## Changes

* Implement the configuration parser.

### DIFF
--- a/sdk/src/configuration/configuration_parser.cc
+++ b/sdk/src/configuration/configuration_parser.cc
@@ -1,0 +1,1721 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stddef.h>
+#include <cstddef>
+#include <fstream>
+#include <map>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "opentelemetry/sdk/common/global_log_handler.h"
+#include "opentelemetry/sdk/configuration/aggregation_configuration.h"
+#include "opentelemetry/sdk/configuration/always_off_sampler_configuration.h"
+#include "opentelemetry/sdk/configuration/always_on_sampler_configuration.h"
+#include "opentelemetry/sdk/configuration/attribute_limits_configuration.h"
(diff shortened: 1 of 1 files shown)


sdk/src/configuration/configuration_parser.cc
```diff
@@ -0,0 +1,1721 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <stddef.h>
```
dbarker, line 4: Marc, is it reasonable to add a test to this PR that covers file configuration parsing? Going through this parser.cc in the context of a test will be helpful for review.

marcalff: Parser yaml tests now added, see files under sdk/test

sdk/src/configuration/configuration_parser.cc
```diff
@@ -0,0 +1,1766 @@
...
+  }
+
+  OTEL_INTERNAL_LOG_ERROR("OtlpHttpEncoding: name = " << name);
+  throw InvalidSchemaException("Illegal OtlpHttpEncoding");
+}
+
+static std::unique_ptr<StringArrayConfiguration> ParseStringArrayConfiguration(
+    const std::unique_ptr<DocumentNode> &node)
+{
+  std::unique_ptr<StringArrayConfiguration> model(new StringArrayConfiguration);
```
dbarker: nit - using std::make_unique in place of new for all the unique_ptr creation cases should be preferred. We are using std::make_unique elsewhere in the code base so there shouldn't be a problem with migrating from new to make_unique.

dbarker: The rationale for this is to follow C++ Core Guideline R23. 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r23-use-make_unique-to-make-unique_ptrs

marcalff: Good point. Fixed everywhere.

dbarker (approved):
LGTM. Thanks!

PR #3522, opened 2025-07-07
[CMAKE] Add CMake script to find or fetch prometheus-cpp

dbarker:
Fixes # (issue)

## Changes

- Adds CMake script to find or fetch prometheus-cpp
- Explicitly link to the core and pull targets from the prometheus exporter.

### DIFF
--- a/exporters/prometheus/CMakeLists.txt
+++ b/exporters/prometheus/CMakeLists.txt
@@ -1,10 +1,6 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-if(NOT TARGET prometheus-cpp::core)
-  find_package(prometheus-cpp CONFIG REQUIRED)
-endif()
-
 add_library(
   opentelemetry_exporter_prometheus
   src/exporter.cc src/exporter_options.cc src/exporter_factory.cc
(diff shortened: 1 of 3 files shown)


exporters/prometheus/CMakeLists.txt
```diff
@@ -19,22 +15,10 @@ target_include_directories(
...
-  list(APPEND PROMETHEUS_EXPORTER_TARGETS util)
-endif()
-set(PROMETHEUS_CPP_TARGETS prometheus-cpp::pull prometheus-cpp::core)
-if(TARGET prometheus-cpp::util)
-  list(APPEND PROMETHEUS_CPP_TARGETS prometheus-cpp::util)
-endif()
-target_link_libraries(opentelemetry_exporter_prometheus
-                      PUBLIC opentelemetry_metrics ${PROMETHEUS_CPP_TARGETS})
+target_link_libraries(
+  opentelemetry_exporter_prometheus
```
owent, line 19: The target names are used for compatibility with an older Prometheus version, as the target naming in Prometheus changed before.

dbarker: Thanks for the context. We're testing with prometheus-cpp 1.1.0 to 1.3.0 in ci and this cmake change is compatible with those versions. 

If we go back further (to prometheus-cpp v1.0.0) then there are compiler errors unrelated to cmake and c++ changes will be needed. These errors are due to the introduction of the Info metric from version v1.1.0. 

```
/workspaces/opentelemetry-cpp/exporters/prometheus/src/exporter_utils.cc:620:55: error: 'Info' is not a member of 'prometheus::MetricType'
  620 |   metric_family.type = prometheus_client::MetricType::Info;
      |                                                       ^~~~
/workspaces/opentelemetry-cpp/exporters/prometheus/src/exporter_utils.cc:624:10: error: 'struct prometheus::ClientMetric' has no member named 'info'
  624 |   metric.info.value                       = 1.0;
```

Given v1.0.0 is very old (released November 2021), I think we should communicate that v1.0.0 is not supported it at this point.

owent: Thanks for these testing. LGTM when we already don't support old prometheus-cpp.

owent (approved):
LGTM

PR #3526, opened 2025-07-09
[CMAKE] Add CMake scripts to find or fetch curl and find zlib

dbarker:
Fixes # (issue)

## Changes

- Add cmake script to find or fetch curl
- Add cmake script to find zlib
- Test fetching curl from github in CI

### DIFF
--- a/cmake/curl.cmake
+++ b/cmake/curl.cmake
@@ -1,0 +1,76 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Import the curl target (CURL::libcurl).
+# 1. Find an installed curl package
+# 2. Use FetchContent to fetch and build curl from GitHub
+
+# Find the curl package with the default search mode
+find_package(CURL QUIET)
+set(CURL_PROVIDER "find_package")
+
+if(NOT CURL_FOUND)
+  FetchContent_Declare(
+    "curl"
+    GIT_REPOSITORY "https://github.com/curl/curl.git"
+    GIT_TAG "${curl_GIT_TAG}"
+    )
+  set(CURL_PROVIDER "fetch_repository")
+
+  if(OPENTELEMETRY_INSTALL)
+    set(_CURL_DISABLE_INSTALL OFF)
+  else()
+    set(_CURL_DISABLE_INSTALL ON)
+  endif()
+
+  if(DEFINED BUILD_SHARED_LIBS)
+    set(_SAVED_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
+  endif()
+
+  set(CURL_DISABLE_INSTALL ${_CURL_DISABLE_INSTALL} CACHE BOOL "" FORCE)
+  set(CURL_USE_LIBPSL OFF CACHE BOOL "" FORCE)
+  set(BUILD_CURL_EXE OFF CACHE BOOL "" FORCE)
+  set(BUILD_LIBCURL_DOCS OFF CACHE BOOL "" FORCE)
+  set(BUILD_MISC_DOCS OFF CACHE BOOL "" FORCE)
+  set(ENABLE_CURL_MANUAL OFF CACHE BOOL "" FORCE)
+  set(BUILD_SHARED_LIBS ON CACHE BOOL "" FORCE)
+
+  FetchContent_MakeAvailable(curl)
+
+  # Restore BUILD_SHARED_LIBS
+  if(DEFINED _SAVED_BUILD_SHARED_LIBS)
+    set(BUILD_SHARED_LIBS ${_SAVED_BUILD_SHARED_LIBS} CACHE BOOL "" FORCE)
+  else()
+    unset(BUILD_SHARED_LIBS CACHE)
+  endif()
+
+  # Set the CURL_VERSION variable from the git tag.
+  string(REGEX REPLACE "^v([0-9]+\\.[0-9]+\\.[0-9]+)$" "\\1" CURL_VERSION "${curl_GIT_TAG}")
+
+  # disable iwyu and clang-tidy
+  foreach(_curl_target libcurl_shared libcurl_static)
+    if(TARGET ${_curl_target})
+      set_target_properties(${_curl_target} PROPERTIES CXX_INCLUDE_WHAT_YOU_USE ""
+                                                     CXX_CLANG_TIDY "")
+    endif()
+  endforeach()
+endif()
+
+# Set the CURL_VERSION from the legacy CURL_VERSION_STRING Required for CMake
+# versions below 4.0
+if(NOT CURL_VERSION AND CURL_VERSION_STRING)
+  set(CURL_VERSION ${CURL_VERSION_STRING})
+endif()
+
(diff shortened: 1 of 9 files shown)


cmake/curl.cmake
```diff
@@ -0,0 +1,76 @@
...
+if(NOT CURL_VERSION AND CURL_VERSION_STRING)
+  set(CURL_VERSION ${CURL_VERSION_STRING})
+endif()
+
+# Add the main CURL::libcurl alias target if missing. Prefer the shared target followed by the static target
+if(NOT TARGET CURL::libcurl)
+  if(TARGET libcurl_shared)
+    add_library(CURL::libcurl ALIAS libcurl_shared)
+  elseif(TARGET libcurl_static)
+    add_library(CURL::libcurl ALIAS libcurl_static)
```
owent, line 70: I think we shouldn't add alias here. If users use http exporters and CONFIG package of curl, the exported `opentelemetry-cpp::http_client_curl` will linking `CURL::libcurl` but it does not exist in users's repo, only exists in otel-cpp.

dbarker: Thanks for the review. The intent with adding the alias here is to provide consistent behavior if older curl versions are used. The CURL::libcurl target is the public target imported for curl. 

The CURL::libcurl alias  target is defined by [curl-config.cmake.in#L40](https://github.com/curl/curl/blob/8eab2b70860e285525130f69b22ac5567292c659/CMake/curl-config.cmake.in#L40) and the FindCURL module. Finding curl with find_package will import this target. 

When building curl in-tree with otel-cpp the alias will also be created by the curl CMake (added several years ago). See [lib/CMakeLists.txt#L288](https://github.com/curl/curl/blob/8eab2b70860e285525130f69b22ac5567292c659/lib/CMakeLists.txt#L288). So this alias should only be created by the otel-cpp curl.cmake script if building an old version of curl.

owent: The alias in [curl-config.cmake.in](https://github.com/curl/curl/blob/8eab2b70860e285525130f69b22ac5567292c659/CMake/curl-config.cmake.in#L40) is available only for curl 8.3.0 and above. Do you think we should support older version of curl?

RH 8 or CentOS 8 has curl 7.61, ,Debian 12 has curl 7.88. I think many products do not use such a new curl version.

dbarker: We should test with older curl versions. The minimum tested in ci now is 7.81 but we can lower it to 7.61 if that seems appropriate. 

Since we're not specifying the CMake search mode for curl it will use the module first. The module will import the CURL::libcurl  target since our minimum CMake version is 3.14.

owent: I mean, users can use CONFIG package explicitly.

We build everything with `"-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON"` in our own build system. Which will make all packages find CONFIG package first.(The CONFIG packages have better dependency graph). This alias will break the linking and we have to make another patch to support the internal alias in otel-cpp.

Agree to add a test for this case, maybe we can add a CI job with `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` .

dbarker: I'd be in favor of adding a new test case, but I wouldn't expect setting `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` to impact the import of the CURL::libcurl target. 
The behavior for finding an installed curl package is not changed in this PR (curl is found with the find_package default search mode). 

After some digging into the curl CMake code it seems safe for otel cpp to link to the standard CURL::libcurl target even for old versions (back to 8+ year old releases) and independent of the search mode used.

- Old versions (like 7.61) will install the CURL::libcurl  target and it is imported by the curl-config.cmake file when the CONFIG search mode is used. See [lib/CMakeLists.txt#L110](https://github.com/curl/curl/blob/432eb5f5c254ee8383b2522ce597c9219877923e/lib/CMakeLists.txt#L110)

-  When the MODULE search mode is used the target will be imported since CMake 3.12. See https://cmake.org/cmake/help/v4.1/module/FindCURL.html. 

- Since CMake 3.17 the FindCURL module will find the curl config.cmake file and use it to import the target.

Am I missing a case where the CURL:: libcurl target would not be imported when calling `find_package(CURL)`?

The curl.cmake file added by this PR will only create the CURL::libcurl alias target if no installed curl package is found and only if a user defines an old version of curl before adding otel-cpp to their build tree using `FetchContent_Declare(curl GIT_TAG <old version tag> ...)`. The fetch content declaration made by otel-cpp will use the version defined by the third_party_release file (curl 8.12, which already creates the alias). If this path is taken the curl will still install if OPENTELEMETRY_INSTALL=ON and then the standard CURL target can be found at the install path.

cmake/curl.cmake
```diff
@@ -0,0 +1,76 @@
...
+
+# Find the curl package with the default search mode
+find_package(CURL QUIET)
+set(CURL_PROVIDER "find_package")
+
+if(NOT CURL_FOUND)
+  FetchContent_Declare(
+    "curl"
+    GIT_REPOSITORY "https://github.com/curl/curl.git"
+    GIT_TAG "${curl_GIT_TAG}"
```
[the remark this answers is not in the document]
dbarker: The curl_GIT_TAG variable is created when reading the third_party_release file in the main CMakeLists.txt file of opentelemetry-cpp. 

The FetchContent path is tested in the CI workflow by FetchContent test.

cmake/curl.cmake
```diff
@@ -0,0 +1,76 @@
...
+
+  # Restore BUILD_SHARED_LIBS
+  if(DEFINED _SAVED_BUILD_SHARED_LIBS)
+    set(BUILD_SHARED_LIBS ${_SAVED_BUILD_SHARED_LIBS} CACHE BOOL "" FORCE)
+  else()
+    unset(BUILD_SHARED_LIBS CACHE)
+  endif()
+
+  # Set the CURL_VERSION variable from the git tag.
+  string(REGEX REPLACE "^curl-([0-9]+)_([0-9]+)_([0-9]+)$" "\\1.\\2.\\3" CURL_VERSION "${curl_GIT_TAG}")
```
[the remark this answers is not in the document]
dbarker: See comment above about the definition of the git tag variable and test in ci. The format of the git tag is consistent throughout curl releases and this file assumes it will remain in this format for future releases. The CURL version variable is only informative and used in logging the build configuration.

cmake/curl.cmake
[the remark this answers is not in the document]
dbarker: It seems Copilot doesn't or can't reply to responses in these threads. Resolving.

cmake/curl.cmake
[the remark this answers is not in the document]
dbarker: Resolving - the curl tag format is assumed to be consistent and the impact is low if the version format for curl tags changes. This can be managed in the PRs for curl upgrades.

cmake/curl.cmake
owent: Sorry, it's my mistake, I forgot curl had changed the library name after 8.3. The `CURL::libcurl` is always available for all versions. LGTM now.

PR #3533, opened 2025-07-10
[CMAKE] Add CMake scripts to find or fetch protobuf and grpc

### LINKED ISSUE #3554: [CMake] Discovery of grpc_cpp_plugin does not work with ALIAS targets
It would be good if it works like alias target fallback of protoc. I know that the root cause is, from the fact that gRPC's cmake is not well written. But IMO it would be better to have it.

**Describe your environment**

- otel-cpp: latest (tried v1.22.2 and same code in main branch HEAD)
- compiler: GCC 15 (Fedora 42), 11 (RHEL 8)
- cmake: 3.29.x with ninja
- os: linux

**Steps to reproduce**

- Add absl, protobuf, gRPC as FetchContent
- Set `gRPC::` namespaced ALIAS targets to gRPC's
  (as it is not defined so when it is used with add_subdirectory and similar things)
- Configure & build it

I am attaching full cmake usage example all the way below.

Even with non-imported but aliased targets, it would be better to work without problem.

The (path) variable `gRPC_CPP_PLUGIN_EXECUTABLE` which is from `gRPC::grpc_cpp_plugin` becomes empty so protoc step of OTel-gRPC protos fail. 

**Additional context**

It would be better if this gRPC cpp plugin discovery steps

https://github.com/open-telemetry/opentelemetry-cpp/blob/c4068645558eeae5d88e5e67ed4fa3d0b0b2908a/cmake/opentelemetry-proto.cmake#L181-L186

works like this protoc discovery steps.

https://github.com/open-telemetry/opentelemetry-cpp/blob/c4068645558eeae5d88e5e67ed4fa3d0b0b2908a/CMakeLists.txt#L401-L412

And below is the example of CMake-FetchContent usage.

```
    message(STATUS "gRPC: Using FetchContent")
    FetchContent_Declare(
        absl
        GIT_REPOSITORY https://github.com/grpc/grpc.git
        GIT_TAG        v1.39.1
        SOURCE_SUBDIR third_party/abseil-cpp
        EXCLUDE_FROM_ALL
        OVERRIDE_FIND_PACKAGE
    )
    FetchContent_Declare(
        protobuf
        GIT_REPOSITORY https://github.com/grpc/grpc.git
        GIT_TAG        v1.39.1
        SOURCE_SUBDIR third_party/protobuf/cmake
        EXCLUDE_FROM_ALL
        OVERRIDE_FIND_PACKAGE
    )
    FetchContent_Declare(
        gRPC
        GIT_REPOSITORY https://github.com/grpc/grpc.git
        GIT_TAG        v1.39.1
        EXCLUDE_FROM_ALL
        OVERRIDE_FIND_PACKAGE
    )

    set(gRPC_ABSL_PROVIDER "package")
    set(gRPC_PROTOBUF_PROVIDER "package")
    set(ABSL_PROPAGATE_CXX_STD ON)
    set(RE2_BUILD_TESTING OFF)
    set(gRPC_BUILD_TESTS OFF)

    FetchContent_MakeAvailable(absl)
    FetchContent_MakeAvailable(protobuf)
    FetchContent_MakeAvailable(gRPC)

    ## grpc does not offer namespaced targets, so we create them
    add_library(gRPC::grpc++ ALIAS grpc++)
    add_library(gRPC::grpc++_reflection ALIAS grpc++_reflection)
    add_executable(gRPC::grpc_cpp_plugin ALIAS grpc_cpp_plugin)

    FetchContent_Declare(
        opentelemetry-cpp
        GIT_REPOSITORY https://github.com/open-telemetry/opentelemetry-cpp.git
        GIT_TAG v1.22.2
        OVERRIDE_FIND_PACKAGE
    )

    set(WITH_OTLP_GRPC ON)
    set(WITH_OTLP_HTTP ON)
    set(WITH_OTLP_FILE ON)
    set(WITH_NO_DEPRECATED_CODE ON)

    set(BUILD_TESTING OFF)
    set(WITH_EXAMPLES OFF)
    set(WITH_BENCHMARK OFF)
    
    FetchContent_MakeAvailable(opentelemetry-cpp)
```

dbarker:
Fixes #3554

## Changes

- Adds CMake scripts for protobuf and grpc to find installed packages OR fetch from github and build in-tree with otel-cpp
- Test fetching grpc and protobuf in the CI FetchContent test

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -37,10 +37,10 @@ jobs:
       run: |
         sudo -E ./ci/setup_ci_environment.sh
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
+      uses: github/codeql-action/init@d6bbdef45e766d081b84a2def353b0055f728d3e # v3.29.3
       with:
        languages: cpp
     - name: Autobuild
-      uses: github/codeql-action/autobuild@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
+      uses: github/codeql-action/autobuild@d6bbdef45e766d081b84a2def353b0055f728d3e # v3.29.3
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
+      uses: github/codeql-action/analyze@d6bbdef45e766d081b84a2def353b0055f728d3e # v3.29.3
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -47,6 +47,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@181d5eefc20863364f96762470ba6f862bdef56b # v3.29.2
+        uses: github/codeql-action/upload-sarif@d6bbdef45e766d081b84a2def353b0055f728d3e # v3.29.3
         with:
           sarif_file: results.sarif
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Increment the:
 * [SDK] Update default exemplar reservoir size for exponential histograms
   [#3551](https://github.com/open-telemetry/opentelemetry-cpp/pull/3551)
 
+* [SDK] Implements options for the ParentBasedSampler with default values
+  [#3553](https://github.com/open-telemetry/opentelemetry-cpp/pull/3553)
+
 ## [1.22 2025-07-11]
 
 * [DOC] Udpate link to membership document
--- a/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/parent.h
@@ -9,6 +9,8 @@
 #include "opentelemetry/common/key_value_iterable.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/trace/sampler.h"
+#include "opentelemetry/sdk/trace/samplers/always_off.h"
+#include "opentelemetry/sdk/trace/samplers/always_on.h"
 #include "opentelemetry/trace/span_context.h"
 #include "opentelemetry/trace/span_metadata.h"
 #include "opentelemetry/trace/trace_id.h"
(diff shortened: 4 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
     - name: install dependencies
       run: |
         sudo -E apt-get update
-        sudo -E apt-get install -y zlib1g-dev libabsl-dev libprotobuf-dev libgrpc++-dev protobuf-compiler protobuf-compiler-grpc
+        sudo -E apt-get install -y zlib1g-dev
```
dbarker, line 76: In this test now all dependencies are fetched (from the git submodule if it exists and otherwise from github) with the exception of zlib.

install/cmake/CMakeLists.txt
```diff
@@ -232,6 +232,7 @@ if(grpc IN_LIST _THIRDPARTY_PACKAGE_LIST)
                "-DRE2_BUILD_TESTING=OFF"
                "-DgRPC_ZLIB_PROVIDER=package"
                "-DgRPC_PROTOBUF_PROVIDER=package"
+               "-DgRPC_PROTOBUF_PACKAGE_TYPE=CONFIG"
```
dbarker, line 235: To enable fetching both grpc and protobuf from github and building in-tree with otel-cpp we need to find or fetch grpc before protobuf. If Protobuf is found using the MODULE mode then the `protobuf::libprotobuf` target type is set to "UNKNOWN_LIBRARY" and the linking to the grpc client triggers a known (and now fixed) issue with grpc v1.49.2 where `grpc::CreateCustomChannel` will hang indefinitely. The work around for that issue is to ensure grpc finds protobuf using the CONFIG mode which is set here. This is not a problem for later versions of grpc.

CMakeLists.txt
```diff
@@ -344,83 +344,23 @@ if(WITH_PROMETHEUS)
...
-        )
-      endif()
-    endif()
-  endif()
-  # Latest Protobuf imported targets and without legacy module support
-  if(TARGET protobuf::protoc)
-    if(CMAKE_CROSSCOMPILING AND Protobuf_PROTOC_EXECUTABLE)
-      set(PROTOBUF_PROTOC_EXECUTABLE ${Protobuf_PROTOC_EXECUTABLE})
-    else()
-      project_build_tools_get_imported_location(PROTOBUF_PROTOC_EXECUTABLE
```
dbarker, line 405: We can use the generator expression `$<TARGET_FILE:protobuf::protoc>` to get the imported file location. This is the method grpc also uses. See [grpc/cmake/protobuf.cmake#L78](https://github.com/grpc/grpc/blob/e99ffb02ac07c996a35472ba7eb7edf736ddc1a2/cmake/protobuf.cmake#L78)

CMakeLists.txt
```diff
@@ -344,83 +344,23 @@ if(WITH_PROMETHEUS)
   include("${opentelemetry-cpp_SOURCE_DIR}/cmake/prometheus-cpp.cmake")
 endif()
 
+#
+# Do we need protobuf and/or gRPC ?
```
ThomsonTan, line 348: What is this comment for? Protobuf is a dependence of OTLP, but gRPC is not.

dbarker: These "Do we need .. ?" comments were in the CMake file previously I continued the comment style here. I'd be okay removing all of them as the CMake code is pretty self explanatory now with includes for each dependency CMake script.

cmake/protobuf.cmake
```diff
@@ -0,0 +1,73 @@
...
+
+  endif()
+endif()
+
+if(NOT TARGET protobuf::libprotobuf)
+  message(FATAL_ERROR "A required protobuf target (protobuf::libprotobuf) was not imported")
+endif()
+
+if(CMAKE_CROSSCOMPILING)
+  find_program(PROTOBUF_PROTOC_EXECUTABLE protoc)
```
owent: `Protobuf_PROTOC_EXECUTABLE` allows users to specify the protoc executable when it is not in `PATH`, typically during cross-compilation. `PROTOBUF_PROTOC_EXECUTABLE` is retained for backward compatibility with older CMake versions.
Do we still need to support the legacy build scripts? If not, I think we should use `Protobuf_PROTOC_EXECUTABLE` instead of `PROTOBUF_PROTOC_EXECUTABLE`, which is used in [FindProtobuf](https://cmake.org/cmake/help/latest/module/FindProtobuf.html) of the latest CMake.

dbarker: Thanks @owent I'll take a look at Protobuf_PROTOC_EXECUTABLE vs PROTOBUF_PROTOC_EXECUTABLE.

dbarker: I've updated the file to use `Protobuf_PROTOC_EXECUTABLE`. I think a follow up PR can remove the reliance on this variable by using the newer `protobuf_generate` function (added in CMake 3.13) in place of the custom commands in `opentelemetry-proto.cmake` and grpc example.  

@owent can you take another look?

PR #3534, opened 2025-07-11
[ADMIN] Remove file .github/repository-settings.md

reyang:
Related to https://github.com/open-telemetry/admin/pull/116.

## Changes

This file is no longer needed as we're now using the admin repo to manage the settings.


marcalff:
Indeed. This file has never been used.

PR #3537, opened 2025-07-13
[CONFIGURATION] File configuration - component registry

marcalff:
Contributes to #2481

This is a partial fix, that implements the component registry.

## Changes

* Implement the component registry:
  * Core component are made available using a build function, to avoid link dependencies on optional components.
  * Extension points are registered by name

### DIFF
--- a/.github/workflows/cmake_install.yml
+++ b/.github/workflows/cmake_install.yml
@@ -88,7 +88,7 @@ jobs:
       run: ./ci/do_ci.sh cmake.install.test
 
   ubuntu_2404_latest:
-    name: Ubuntu 24.04 latest versions cxx20 (static libs)
+    name: Ubuntu 24.04 latest versions cxx20 (static libs - shared libs)
     runs-on: ubuntu-24.04
     env:
       INSTALL_TEST_DIR: '/home/runner/install_test'
@@ -111,6 +111,10 @@ jobs:
     - name: Run Tests (static libs)
       env:
         BUILD_SHARED_LIBS: 'OFF'
+      run: ./ci/do_ci.sh cmake.install.test
+    - name: Run Tests (shared libs)
+      env:
+        BUILD_SHARED_LIBS: 'ON'
       run: ./ci/do_ci.sh cmake.install.test
 
   ubuntu_2204_stable:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [TEST] Shared otel-cpp libs linked to latest static protobuf and grpc
+  [#3544](https://github.com/open-telemetry/opentelemetry-cpp/pull/3544)
+
 ## [1.22 2025-07-11]
 
 * [DOC] Udpate link to membership document


sdk/include/opentelemetry/sdk/init/extension_span_processor_builder.h
```diff
@@ -0,0 +1,33 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "opentelemetry/sdk/configuration/extension_span_processor_configuration.h"
+#include "opentelemetry/sdk/trace/processor.h"
+#include "opentelemetry/version.h"
```
dbarker: nit - many of these headers are missing some header includes. Many are missing `<memory>` for unique_ptr.

sdk/include/opentelemetry/sdk/init/console_push_metric_exporter_builder.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#include "opentelemetry/sdk/configuration/console_push_metric_exporter_configuration.h"
+#include "opentelemetry/sdk/metrics/push_metric_exporter.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace init
```
dbarker: q: Are all the declarations made under the `init` namespace and files in the `init` folder specifically for implementation of the declarative configuration feature/spec? 

If so the `init` folder being separate from the `configuration` folder and namespaces being orthogonal may feel unexpected.

sdk/include/opentelemetry/sdk/init/console_log_record_exporter_builder.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#include "opentelemetry/sdk/configuration/console_log_record_exporter_configuration.h"
+#include "opentelemetry/sdk/logs/exporter.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace init
```
dbarker: nit - abbreviation (ie: `init` instead of `initialization`) should be generally avoided per the style guide (https://google.github.io/styleguide/cppguide.html#General_Naming_Rules). There are not many abbreviations in otel-cpp namespaces/folders (although well known acronyms are like api/sdk/otlp are used).

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+  Registry &operator=(const Registry &other) = default;
+
+  ~Registry() = default;
+
+  /* Core optional components. */
+
+  const OtlpHttpSpanExporterBuilder *GetOtlpHttpSpanBuilder() { return m_otlp_http_span_builder; }
+  void SetOtlpHttpSpanBuilder(const OtlpHttpSpanExporterBuilder *builder)
+  {
+    m_otlp_http_span_builder = builder;
```
dbarker: nit - member naming style `m_` -> trailing `_`

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+  const OtlpHttpLogRecordExporterBuilder *GetOtlpHttpLogRecordBuilder()
+  {
+    return m_otlp_http_log_record_builder;
+  }
+  void SetOtlpHttpLogRecordBuilder(const OtlpHttpLogRecordExporterBuilder *builder)
+  {
+    m_otlp_http_log_record_builder = builder;
+  }
+
+  const OtlpGrpcLogRecordExporterBuilder *GetOtlpGrpcLogRecordBuilder()
```
dbarker: q: can / should any of these accessors be const?

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+  void AddExtensionLogRecordExporterBuilder(const std::string &name,
+                                            ExtensionLogRecordExporterBuilder *builder);
+
+  const ExtensionLogRecordProcessorBuilder *GetExtensionLogRecordProcessorBuilder(
+      const std::string &name);
+  void AddExtensionLogRecordProcessorBuilder(const std::string &name,
+                                             ExtensionLogRecordProcessorBuilder *builder);
+
+private:
+  const OtlpHttpSpanExporterBuilder *m_otlp_http_span_builder{nullptr};
```
dbarker: nit - const members will create a clang-tidy warning (ci may not be catching this until we upgrade clang-tidy or change some clang-tidy options). See https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

marcalff: This is a const pointer, not just a const, I think it will be ok.

sdk/include/opentelemetry/sdk/init/console_push_metric_exporter_builder.h
marcalff: Everything under the init folder is for the declarative configuration feature.

The content of init is more dedicated to creating sdk objects, while the content of configuration is more dedicated to yaml parsing.

Now that I look at it, I think having a single configuration folder would be simpler.

Please confirm you are ok with this, before I make all the changes ;-)

sdk/include/opentelemetry/sdk/init/console_log_record_exporter_builder.h
marcalff: Code to be moved under configuration anyway.

sdk/include/opentelemetry/sdk/init/console_push_metric_exporter_builder.h
dbarker: Moving the code under configuration sounds good to me. If it makes sense to have sub folders under configuration that is okay too.

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+  const OtlpFilePushMetricExporterBuilder *m_otlp_file_push_metric_builder{nullptr};
+  const ConsolePushMetricExporterBuilder *m_console_metric_builder{nullptr};
+  const PrometheusPullMetricExporterBuilder *m_prometheus_metric_builder{nullptr};
+
+  const OtlpHttpLogRecordExporterBuilder *m_otlp_http_log_record_builder{nullptr};
+  const OtlpGrpcLogRecordExporterBuilder *m_otlp_grpc_log_record_builder{nullptr};
+  const OtlpFileLogRecordExporterBuilder *m_otlp_file_log_record_builder{nullptr};
+  const ConsoleLogRecordExporterBuilder *m_console_log_record_builder{nullptr};
+
+  std::map<std::string, TextMapPropagatorBuilder *> m_propagator_builders;
```
dbarker: q: Can the values of these maps be owning ptrs (shared or unique)? If so then the static global scope builder objects in the cc file could be removed.

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+class ExtensionPullMetricExporterBuilder;
+
+class OtlpHttpLogRecordExporterBuilder;
+class OtlpGrpcLogRecordExporterBuilder;
+class OtlpFileLogRecordExporterBuilder;
+class ConsoleLogRecordExporterBuilder;
+class ExtensionLogRecordExporterBuilder;
+class ExtensionLogRecordProcessorBuilder;
+
+class Registry
```
dbarker: q: Is there only one Registry allowed per process or can there be many registries?

sdk/src/init/registry.cc
```diff
@@ -0,0 +1,251 @@
...
+{
+public:
+  std::unique_ptr<opentelemetry::context::propagation::TextMapPropagator> Build() const override
+  {
+    auto result = std::make_unique<opentelemetry::trace::propagation::JaegerPropagator>();
+    return result;
+  }
+};
+
+static TraceContextBuilder trace_context_builder;
```
dbarker: q: Can these static objects be removed? 

Ideally we don't have static storage duration for these if it can be avoided. Generally to be avoided per the style guide if we can (https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables)

sdk/src/init/registry.cc
```diff
@@ -0,0 +1,251 @@
...
+  m_propagator_builders.insert(entry);
+
+  entry = {"b3", &b3_builder};
+  m_propagator_builders.insert(entry);
+
+  entry = {"b3multi", &b3_multi_builder};
+  m_propagator_builders.insert(entry);
+
+  entry = {"jaeger", &jaeger_builder};
+  m_propagator_builders.insert(entry);
```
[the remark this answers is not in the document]
dbarker: I'm okay with this change but certainly optional.

sdk/src/init/registry.cc
```diff
@@ -0,0 +1,251 @@
...
+    builder = search->second;
+  }
+  return builder;
+}
+
+void Registry::AddTextMapPropagatorBuilder(const std::string &name,
+                                           TextMapPropagatorBuilder *builder)
+{
+  std::pair<std::string, TextMapPropagatorBuilder *> entry{name, builder};
+  m_propagator_builders.insert(entry);
```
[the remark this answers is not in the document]
dbarker: consider accepting this change if overriding when a key is inserted twice is the intended behavior or add a check for the key's existence and throw if it should be an error.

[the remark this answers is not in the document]
dbarker: Copilot's change cannot be accepted directly since `insert_or_assign` is a c++17 feature.

sdk/src/init/registry.cc
```diff
@@ -0,0 +1,251 @@
...
+    builder = search->second;
+  }
+  return builder;
+}
+
+void Registry::AddExtensionSpanProcessorBuilder(const std::string &name,
+                                                ExtensionSpanProcessorBuilder *builder)
+{
+  std::pair<std::string, ExtensionSpanProcessorBuilder *> entry{name, builder};
+  m_span_processor_builders.insert(entry);
```
dbarker: q: should this override any existing element with the same key or throw if that is not allowed?

sdk/src/init/registry.cc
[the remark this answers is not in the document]
dbarker: It may be better to create the builder objects as shared_ptrs then add them to the map in one initializer list. Still optional.

sdk/include/opentelemetry/sdk/init/registry.h
dbarker: You are right. `const T*` seems okay but `T* const` would likely trigger the warning.

sdk/include/opentelemetry/sdk/init/registry.h
```diff
@@ -0,0 +1,231 @@
...
+  /* Core optional components. */
+
+  const OtlpHttpSpanExporterBuilder *GetOtlpHttpSpanBuilder() { return m_otlp_http_span_builder; }
+  void SetOtlpHttpSpanBuilder(const OtlpHttpSpanExporterBuilder *builder)
+  {
+    m_otlp_http_span_builder = builder;
+  }
+
+  const OtlpGrpcSpanExporterBuilder *GetOtlpGrpcSpanBuilder() { return m_otlp_grpc_span_builder; }
+  void SetOtlpGrpcSpanBuilder(const OtlpGrpcSpanExporterBuilder *builder)
```
dbarker: q: what object owns the builder?

sdk/src/init/registry.cc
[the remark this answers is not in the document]
dbarker: @lalitb fyi. Is there a way to restrict Copilot to only use C++14 and lower features?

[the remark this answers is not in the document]
ThomsonTan: Should it possible to @ copilot with the feedback, but somehow is it not showing in in the list.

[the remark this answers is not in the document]
lalitb: >  Is there a way to restrict Copilot to only use C++14 and lower features?

We need to add instructions for the Copilot, which it can use for reviewing the PRs - https://docs.github.com/en/enterprise-cloud@latest/copilot/how-tos/agents/copilot-code-review/configuring-coding-guidelines#creating-a-coding-guideline.

It has to be done in repo setting's page - and none of use are the admins to update it. I think this is now done through https://github.com/open-telemetry/admin - will raise a PR with instructions there.

sdk/include/opentelemetry/sdk/init/extension_span_processor_builder.h
marcalff: Fixed.

sdk/include/opentelemetry/sdk/init/console_push_metric_exporter_builder.h
marcalff: Moved all the code under configuration.

sdk/include/opentelemetry/sdk/init/registry.h
marcalff: Fixed.

sdk/include/opentelemetry/sdk/init/registry.h
marcalff: Yes. All fixed.

sdk/include/opentelemetry/sdk/init/registry.h
marcalff: All changed to use std::unique_ptr.

This means that instead of registering a pointer to a global singleton, the caller calls std::make_unique to build an object, and gives it to the registry, who owns it.

sdk/include/opentelemetry/sdk/init/registry.h
marcalff: The registry defines what components or extension plugins are allowed and present in the binary.

The config.yaml files is allowed to pick only from the content of the registry, so for example, if a given exporter is not added in the registry, it will not be available even if config.yaml asks for it.

Typically, this can be for the process, so only one registry is needed, but there is no reasons to forbid using several registries at once, if that becomes necessary.

For an example of the intended use, see PR #2518, file example/yaml/main.cc

sdk/src/init/registry.cc
marcalff: All removed.

The registry now owns the objects with std::unique_ptr, instead of keeping a raw pointer to a singleton.

sdk/src/init/registry.cc
marcalff: Fixed to override elements, so this has the same semantic as for core (non extension) builders.

Renamed the methods while at it, to use Get and Set.

sdk/include/opentelemetry/sdk/init/registry.h
marcalff: Now that the registry uses std::unique_ptr, the registry owns it.

sdk/src/init/registry.cc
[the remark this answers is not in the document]
marcalff: This part got reworked, because the registry now uses std::unique_ptr.

sdk/src/init/registry.cc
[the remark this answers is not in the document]
marcalff: Fixed by using erase + insert

marcalff:
@dbarker Thanks for the first round of review.

All comments addressed, please take another look.

sdk/src/init/registry.cc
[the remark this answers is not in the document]
dbarker: > Should it possible to @ copilot with the feedback, but somehow is it not showing in in the list.

@ThomsonTan I see the same thing and cannot interact with Copilot agent in these threads. Based on the docs it seems only those with write access to the project can.  

> Limits who can assign tasks to Copilot. Only users with write access to the repository can trigger Copilot to work. Comments from users without write access are never presented to the agent.

https://docs.github.com/en/copilot/concepts/about-copilot-coding-agent#risk-copilot-can-push-code-changes-to-your-repository

marcalff:
@Copilot Please take another look.

sdk/include/opentelemetry/sdk/configuration/console_log_record_exporter_builder.h
```diff
@@ -0,0 +1,36 @@
...
+#include "opentelemetry/sdk/logs/exporter.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+class ConsoleLogRecordExporterBuilder
```
dbarker, line 18: q: Is it important to have specific base classes at the level of the final builder (ie: `ConsoleLogRecordExporterBuilder`) or can we reduce the number of base classes to just those return types from the Build methods (ie: `LogRecordExporterBuilder`)?

sdk/include/opentelemetry/sdk/configuration/registry.h
```diff
@@ -0,0 +1,283 @@
...
+
+  const ExtensionLogRecordProcessorBuilder *GetExtensionLogRecordProcessorBuilder(
+      const std::string &name) const;
+
+  void SetExtensionLogRecordProcessorBuilder(
+      const std::string &name,
+      std::unique_ptr<ExtensionLogRecordProcessorBuilder> &&builder);
+
+private:
+  std::unique_ptr<OtlpHttpSpanExporterBuilder> otlp_http_span_builder_;
```
dbarker, line 250: q: Related to the question on base classes above. Would it make sense to move all of these into a set of maps for each type of sdk object being built? For example if there was a base builder class for each sdk type built then could the known derived builders and custom extension builders leverage the same registration interface and maps? 

```cpp
std::map<std::string, std::unique_ptr<TextMapPropagatorBuilder>> propagator_builders_;
std::map<std::string, std::unique_ptr<SpanSamplerBuilder>> sampler_builders_;
std::map<std::string, std::unique_ptr<SpanExporterBuilder>> span_exporter_builders_;
std::map<std::string, std::unique_ptr<SpanProcessorBuilder>> span_processor_builders_;
std::map<std::string, std::unique_ptr<PushMetricExporterBuilder>>
      push_metric_exporter_builders_;
std::map<std::string, std::unique_ptr<PullMetricExporterBuilder>>
      pull_metric_exporter_builders_;
std::map<std::string, std::unique_ptr<LogRecordExporterBuilder>>
      log_record_exporter_builders_;
std::map<std::string, std::unique_ptr<LogRecordProcessorBuilder>>
      log_record_processor_builders_;
```

sdk/include/opentelemetry/sdk/configuration/registry.h
```diff
@@ -0,0 +1,283 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <map>
+#include <string>
+
+#include "opentelemetry/sdk/configuration/console_log_record_exporter_builder.h"
```
dbarker, line 9: nit - If it is desirable to forward declare the base classes then consider moving the set method implementations to the cc file.

sdk/include/opentelemetry/sdk/configuration/console_log_record_exporter_builder.h
marcalff: a: Yes.

Consider `ConsoleLogRecordExporterBuilder` and `OtlpHttpLogRecordExporterBuilder` for example.

The result of the Build method is already abstract, as in `LogRecordExporter`.

However, the parameters given to Build() are different: `ConsoleLogRecordExporterConfiguration` versus `OtlpHttpLogRecordExporterConfiguration`.

Each builder needs the exact configuration parameters that applies to it, and these are different between different builders.

This whole thing is strongly typed (and as a result a bit repetitive), to prevent misuse.

sdk/include/opentelemetry/sdk/configuration/registry.h
marcalff: This can not be abstracted, at least not without giving up on type safety, because different builders for the same familly of sdk objects do take different input data as configuration.

sdk/include/opentelemetry/sdk/configuration/registry.h
marcalff: The include is a side effect of using std:unique_ptr in the registry.

A raw pointer can use forward declarations, but a unique_ptr needs to see a destructor, hence the full header file for builders added in registry.h

PR #3549, opened 2025-07-17
[SDK] Implement env var configuration for PeriodicExportingMetricReader

### LINKED ISSUE #3515: Support periodic exporting MetricReader config from env vars
According to the specs [[1]](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#periodic-exporting-metricreader) the periodic metric reader should read configuration from environment variables. opentelemetry-cpp does not use these environment variables currently [[2]](https://github.com/open-telemetry/opentelemetry-cpp/blob/f4897b2a5f8d9b65ff777c549f16f3952a1eebab/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h#L16-L17)

[1] https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#periodic-exporting-metricreader
[2] https://github.com/open-telemetry/opentelemetry-cpp/blob/f4897b2a5f8d9b65ff777c549f16f3952a1eebab/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h#L16-L17

mathieurobin1:
Fixes #3515 

## Changes

Implements the configuration of `PeriodicExportingMetricReader` via environment variables with default values.

It introduces support for:
*   `OTEL_METRIC_EXPORT_INTERVAL` (Duration)
*   `OTEL_METRIC_EXPORT_TIMEOUT` (Timeout)

**Known Limitation / Discussion Point:**

During implementation, I identified that the existing utility function `common::GetDurationEnvironmentVariable` is not fully compliant with the specification for `Timeout` types.

*   **Specification:** A `Timeout` value of `0` SHOULD be interpreted as "infinite".
*   **Current Implementation:** The underlying parser (`GetTimeoutFromString`) explicitly rejects a value of `0` as invalid.

To keep the scope of this PR focused on the `PeriodicExportingMetricReaderOptions`, I have used the existing utility for both environment variables. This means that `OTEL_METRIC_EXPORT_TIMEOUT=0` is currently not handled as an infinite timeout but will instead fall back to the default value.

I believe fixing the common utility function should be addressed in a separate issue/PR to ensure it doesn't introduce unintended side effects elsewhere.

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
+++ b/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
@@ -5,7 +5,6 @@
 
 #include <chrono>
 
-#include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/version.h"
 
 OPENTELEMETRY_BEGIN_NAMESPACE
@@ -17,14 +16,11 @@ namespace metrics
 constexpr std::chrono::milliseconds kExportIntervalMillis = std::chrono::milliseconds(60000);
 constexpr std::chrono::milliseconds kExportTimeOutMillis  = std::chrono::milliseconds(30000);
 
-std::chrono::milliseconds GetEnvDuration(nostd::string_view env_var_name,
-                                         std::chrono::milliseconds default_value);
-
 /**
  * Struct to hold PeriodicExortingMetricReader options.
  */
 
-struct PeriodicExportingMetricReaderOptions
+struct OPENTELEMETRY_EXPORT PeriodicExportingMetricReaderOptions
 {
   /* The time interval between two consecutive exports. */
   std::chrono::milliseconds export_interval_millis;
--- a/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
+++ b/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
@@ -3,9 +3,10 @@
 
 #pragma once
 
-#include "opentelemetry/version.h"
-
 #include <chrono>
+
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/version.h"
 
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace sdk
@@ -16,6 +17,9 @@ namespace metrics
 constexpr std::chrono::milliseconds kExportIntervalMillis = std::chrono::milliseconds(60000);
 constexpr std::chrono::milliseconds kExportTimeOutMillis  = std::chrono::milliseconds(30000);
 
+std::chrono::milliseconds GetEnvDuration(nostd::string_view env_var_name,
+                                         std::chrono::milliseconds default_value);
+
 /**
  * Struct to hold PeriodicExortingMetricReader options.
  */
@@ -23,11 +27,12 @@ constexpr std::chrono::milliseconds kExportTimeOutMillis  = std::chrono::millise
 struct PeriodicExportingMetricReaderOptions
 {
   /* The time interval between two consecutive exports. */
-  std::chrono::milliseconds export_interval_millis =
-      std::chrono::milliseconds(kExportIntervalMillis);
+  std::chrono::milliseconds export_interval_millis;
 
   /*  how long the export can run before it is cancelled. */
-  std::chrono::milliseconds export_timeout_millis = std::chrono::milliseconds(kExportTimeOutMillis);
+  std::chrono::milliseconds export_timeout_millis;
+
+  PeriodicExportingMetricReaderOptions();
 };
 
 }  // namespace metrics
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
@@ -24,10 +24,12 @@ std::chrono::milliseconds GetEnvDuration(nostd::string_view env_var_name,
   }
   return default_value;
 }
+
 PeriodicExportingMetricReaderOptions::PeriodicExportingMetricReaderOptions()
     : export_interval_millis(GetEnvDuration("OTEL_METRIC_EXPORT_INTERVAL", kExportIntervalMillis)),
       export_timeout_millis(GetEnvDuration("OTEL_METRIC_EXPORT_TIMEOUT", kExportTimeOutMillis))
 {}
+
 }  // namespace metrics
 }  // namespace sdk
 OPENTELEMETRY_END_NAMESPACE
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
@@ -1,0 +1,33 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <chrono>
+
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/sdk/common/env_variables.h"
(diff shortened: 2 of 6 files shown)


marcalff:
I changed the title to [SDK], because I use the [CONFIGURATION] tag is for everything related to declarative configuration (aka, config.yaml)

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
```diff
@@ -24,10 +28,11 @@ struct PeriodicExportingMetricReaderOptions
 {
   /* The time interval between two consecutive exports. */
   std::chrono::milliseconds export_interval_millis =
-      std::chrono::milliseconds(kExportIntervalMillis);
+      GetEnvDuration("OTEL_METRIC_EXPORT_INTERVAL", kExportIntervalMillis);
```
marcalff: Please move this initialization to the *.cc file.

It should not be in the header, we don't want to expose more symbols to the client application code that configures the sdk.

Please move the kExportIntervalMillis constant to *.cc as well.

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
```diff
@@ -24,10 +28,11 @@ struct PeriodicExportingMetricReaderOptions
...
 
   /*  how long the export can run before it is cancelled. */
-  std::chrono::milliseconds export_timeout_millis = std::chrono::milliseconds(kExportTimeOutMillis);
+  std::chrono::milliseconds export_timeout_millis =
```
marcalff: Likewise, move to *.cc

marcalff:
> **Known Limitation / Discussion Point:**
> 
> During implementation, I identified that the existing utility function `common::GetDurationEnvironmentVariable` is not fully compliant with the specification for `Timeout` types.
> 
>     * **Specification:** A `Timeout` value of `0` SHOULD be interpreted as "infinite".
> 
>     * **Current Implementation:** The underlying parser (`GetTimeoutFromString`) explicitly rejects a value of `0` as invalid.
> 
> 
> To keep the scope of this PR focused on the `PeriodicExportingMetricReaderOptions`, I have used the existing utility for both environment variables. This means that `OTEL_METRIC_EXPORT_TIMEOUT=0` is currently not handled as an infinite timeout but will instead fall back to the default value.
> 
> I believe fixing the common utility function should be addressed in a separate issue/PR to ensure it doesn't introduce unintended side effects elsewhere.

Looks like you found an existing bug.

Please use `common::GetDurationEnvironmentVariable` so that all duration variables have the same behavior, instead of having multiple and slightly different helpers.

The issue with 0 / infinite can be fixed separately.

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
mathieurobin1: Thanks for your feedback, I'll move the initialization in a constructor in the .cc.

However, I can't move the kExportIntervalMillis and kExportTimeOutMillis to the *.cc file since they are used in the periodic_exporting_metric_reader.cc in case the interval is inferior to the timeout. So I would suggest to leave them in the .h. 
I could also create another .h file for these two constants and include it but I would say the first option is better ?

marcalff: > However, I can't move the kExportIntervalMillis and kExportTimeOutMillis to the *.cc file since they are used in the periodic_exporting_metric_reader.cc in case the interval is inferior to the timeout.

Thanks for the details.
Ok to leave constants where they are then.

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader_options.h
```diff
@@ -16,18 +17,22 @@ namespace metrics
 constexpr std::chrono::milliseconds kExportIntervalMillis = std::chrono::milliseconds(60000);
 constexpr std::chrono::milliseconds kExportTimeOutMillis  = std::chrono::milliseconds(30000);
 
+std::chrono::milliseconds GetEnvDuration(nostd::string_view env_var_name,
+                                         std::chrono::milliseconds default_value);
+
```
marcalff: This is no longer needed, remove this declaration from the header file.

marcalff:
The include-what-you-use failures can be seen in CI, go to "Summary" for the CI job, and download artifacts that contains the full logs.

Search for `Warning: include-what-you-use reported diagnostics`.

Logs provided below

```Warning: include-what-you-use reported diagnostics:

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/metrics/periodic_exporting_metric_reader_test.cc should add these lines:
#include <stdlib.h>
#include <ratio>
```

Also, there is a build break for tests on Windows. See existing tests for how to use setenv.

mathieurobin1:
Thanks for your feedback, forgot to include env_variables.h for Windows build. 
Should be good now :)

lalitb (approved):
Thanks.

marcalff:
Build failure in windows CI:

```
metrics_ostream.obj : error LNK2019: unresolved external symbol "public: __cdecl opentelemetry::v1::sdk::metrics::PeriodicExportingMetricReaderOptions::PeriodicExportingMetricReaderOptions(void)" (??0PeriodicExportingMetricReaderOptions@metrics@sdk@v1@opentelemetry@@QEAA@XZ) referenced in function "void __cdecl `anonymous namespace'::InitMetrics(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const &)" (?InitMetrics@?A0x81c1a3c3@@YAXAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z) [D:\a\opentelemetry-cpp\opentelemetry-cpp\build\install_test\examples-test\examples\metrics_simple\metrics_ostream_example.vcxproj]
```

To fix it:

* add `OPENTELEMETRY_EXPORT`, see file `otlp_http_metric_exporter_options.h` for an example
* adjust `ext/src/dll/input.src` if necessary

Since a new patch is needed to fix this, I am adding minor nitpicks as well.

Code looks good overall, will approve once CI is clean.

sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
```diff
@@ -0,0 +1,33 @@
...
+std::chrono::milliseconds GetEnvDuration(nostd::string_view env_var_name,
+                                         std::chrono::milliseconds default_value)
+{
+  std::chrono::system_clock::duration duration;
+  if (common::GetDurationEnvironmentVariable(env_var_name.data(), duration))
+  {
+    return std::chrono::duration_cast<std::chrono::milliseconds>(duration);
+  }
+  return default_value;
+}
```
marcalff, line 26: nitpick: (style) add an empty line

sdk/src/metrics/export/periodic_exporting_metric_reader_options.cc
```diff
@@ -0,0 +1,33 @@
...
+PeriodicExportingMetricReaderOptions::PeriodicExportingMetricReaderOptions()
+    : export_interval_millis(GetEnvDuration("OTEL_METRIC_EXPORT_INTERVAL", kExportIntervalMillis)),
+      export_timeout_millis(GetEnvDuration("OTEL_METRIC_EXPORT_TIMEOUT", kExportTimeOutMillis))
+{}
```
marcalff, line 31: nitpick: (style) add an empty line

marcalff (approved):
LGTM.

Thanks for the fix, and the misc cleanup.

marcalff:
@mathieurobin1 Congratulations on your first patch. Merged.

PR #3550, opened 2025-07-17
[CONFIGURATION] File configuration - sdk builder

marcalff:
Contributes to #2481

This is a partial fix, that implements the sdk builder.

## Changes

* Implement the sdk builder:
  * Create opentelemetry-cpp SDK objects from the configuration model.

### DIFF
--- a/sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
+++ b/sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
@@ -1,0 +1,44 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <memory>
+
+#include "opentelemetry/context/propagation/text_map_propagator.h"
+#include "opentelemetry/sdk/configuration/configuration.h"
+#include "opentelemetry/sdk/configuration/registry.h"
+#include "opentelemetry/sdk/logs/logger_provider.h"
+#include "opentelemetry/sdk/metrics/meter_provider.h"
+#include "opentelemetry/sdk/resource/resource.h"
+#include "opentelemetry/sdk/trace/tracer_provider.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+class ConfiguredSdk
+{
+public:
+  static std::unique_ptr<ConfiguredSdk> Create(
+      std::shared_ptr<Registry> registry,
+      const std::unique_ptr<opentelemetry::sdk::configuration::Configuration> &model);
+
+  ConfiguredSdk() : resource(opentelemetry::sdk::resource::Resource::GetEmpty()) {}
+
+  void Install();
+  void UnInstall();
+
+  opentelemetry::sdk::resource::Resource resource;
+  std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> tracer_provider;
+  std::shared_ptr<opentelemetry::context::propagation::TextMapPropagator> propagator;
+  std::shared_ptr<opentelemetry::sdk::metrics::MeterProvider> meter_provider;
+  std::shared_ptr<opentelemetry::sdk::logs::LoggerProvider> logger_provider;
+};
+
+}  // namespace configuration
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
+++ b/sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
@@ -20,6 +20,11 @@ namespace sdk
 namespace configuration
 {
 
+/**
+ * This class represents a fully configured SDK.
+ * A SDK contains various objects, like propagators and providers for each
+ * signals, that collectively describe the opentelemetry configuration.
+ */
 class ConfiguredSdk
 {
 public:
@@ -27,28 +32,27 @@ public:
       std::shared_ptr<Registry> registry,
       const std::unique_ptr<opentelemetry::sdk::configuration::Configuration> &model);
 
-  ConfiguredSdk(
-      opentelemetry::sdk::resource::Resource &&resource,
-      const std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> &tracer_provider,
-      const std::shared_ptr<opentelemetry::context::propagation::TextMapPropagator> &propagator,
-      const std::shared_ptr<opentelemetry::sdk::metrics::MeterProvider> &meter_provider,
-      const std::shared_ptr<opentelemetry::sdk::logs::LoggerProvider> &logger_provider)
-      : resource_(std::move(resource)),
-        tracer_provider_(tracer_provider),
-        propagator_(propagator),
-        meter_provider_(meter_provider),
-        logger_provider_(logger_provider)
-  {}
+  ConfiguredSdk() : resource(opentelemetry::sdk::resource::Resource::GetEmpty()) {}
 
-  void Init();
-  void Cleanup();
+  /**
+   * Install the SDK, so that an instrumented application can make calls
+   * to it.
+   * This methods sets the global provider singletons to point to the SDK.
+   */
+  void Install();
 
-private:
-  opentelemetry::sdk::resource::Resource resource_;
-  std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> tracer_provider_;
-  std::shared_ptr<opentelemetry::context::propagation::TextMapPropagator> propagator_;
-  std::shared_ptr<opentelemetry::sdk::metrics::MeterProvider> meter_provider_;
-  std::shared_ptr<opentelemetry::sdk::logs::LoggerProvider> logger_provider_;
+  /**
+   * Uninstall the SDK, so that an instrumented application no longer makes
+   * calls to it.
+   * This method clears the global provider singletons.
+   */
+  void UnInstall();
+
+  opentelemetry::sdk::resource::Resource resource;
+  std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> tracer_provider;
+  std::shared_ptr<opentelemetry::context::propagation::TextMapPropagator> propagator;
+  std::shared_ptr<opentelemetry::sdk::metrics::MeterProvider> meter_provider;
+  std::shared_ptr<opentelemetry::sdk::logs::LoggerProvider> logger_provider;
 };
 
 }  // namespace configuration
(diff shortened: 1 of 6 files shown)


sdk/src/configuration/sdk_builder.cc
```diff
@@ -0,0 +1,1757 @@
...
+  void VisitExtension(
+      const opentelemetry::sdk::configuration::ExtensionSamplerConfiguration *model) override
+  {
+    sampler = m_sdk_builder->CreateExtensionSampler(model);
+  }
+
+  std::unique_ptr<opentelemetry::sdk::trace::Sampler> sampler;
+
+private:
+  const SdkBuilder *m_sdk_builder;
```
dbarker: nit - naming style

sdk/src/configuration/sdk_builder.cc
```diff
@@ -0,0 +1,1757 @@
...
+  void VisitExtension(
+      const opentelemetry::sdk::configuration::ExtensionSpanProcessorConfiguration *model) override
+  {
+    processor = m_sdk_builder->CreateExtensionSpanProcessor(model);
+  }
+
+  std::unique_ptr<opentelemetry::sdk::trace::SpanProcessor> processor;
+
+private:
+  const SdkBuilder *m_sdk_builder;
```
dbarker: nit - naming style

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
```diff
@@ -0,0 +1,44 @@
...
+#include "opentelemetry/sdk/trace/tracer_provider.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+class ConfiguredSdk
```
dbarker, line 28: nit - consider adding a test for the ConfiguredSdk and SdkBuilder classes to this PR. This could be just testing against one model and verify the resource and providers are generally valid and functional.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
```diff
@@ -0,0 +1,44 @@
...
+{
+public:
+  static std::unique_ptr<ConfiguredSdk> Create(
+      std::shared_ptr<Registry> registry,
+      const std::unique_ptr<opentelemetry::sdk::configuration::Configuration> &model);
+
+  ConfiguredSdk() : resource(opentelemetry::sdk::resource::Resource::GetEmpty()) {}
+
+  void Install();
```
dbarker, line 42: nit - the terms `Install` and `UnInstall` feel a bit unexpected in this context since no artifacts are installed or uninstalled. If these terms are not defined by the configuration spec, consider `Install` -> `Initialize` or `Configure` and  `UnInstall` -> `Shutdown` or `Cleanup` as options. The sdk spec uses `Shutdown` for the providers (which will be called when `ConfiguredSdk::UnInstall` is called) and the examples in otel-cpp often use `Init` and `Cleanup`.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
```diff
@@ -0,0 +1,44 @@
...
+  void UnInstall();
+
+  opentelemetry::sdk::resource::Resource resource;
```
dbarker, line 51: nit - The example (`examples/yaml/main.cc`) in the main PR doesn't seem to access these public members. Can/should they be private or protected?

sdk/src/configuration/sdk_builder.cc
```diff
@@ -0,0 +1,1757 @@
...
+#include "opentelemetry/sdk/trace/tracer_provider_factory.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+class ResourceAttributeValueSetter
```
dbarker, line 165: nit - This is a big file with many new classes defined. Consider moving the new class definitions to separate header files (perhaps in a detail sub folder) that will not be installed for public use. Rationale is to improve readability and allow unit testing of these classes in the future. Many of these class throw and we may want to test error conditions explicitly at the unit level

dbarker:
Thanks for the PR @marcalff. Just noted a few nitpicks (optional) and a request to add a test for the ConfiguredSdk class if that is reasonable.

sdk/src/configuration/sdk_builder.cc
marcalff: Fixed.

sdk/src/configuration/sdk_builder.cc
marcalff: Fixed.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
marcalff: Some items are installed, by being set into the global singletons (SetTracerProvider and friends), but I agree this terminology is not used in opentelemetry.

Fixed as `Init()` and `Cleanup()`.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
marcalff: Fixed to private.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
marcalff: I am currently implementing functional tests in the main PR, to do just that.

See functional/configuration/shelltests/kitchen-sink.test for example.

There will be many functional tests for coverage, but this part is not ready yet, it will be available in PRs to come.

sdk/src/configuration/sdk_builder.cc
marcalff: For class `ResourceAttributeValueSetter`, this is the only one that can be extracted, and tested in isolation, because it manipulates a `ResourceAttributes` and does need the sdk builder.

Every other class like `SamplerBuilder` depends on the sdk builder, and can not be tested in isolation.

I don't see how a unit test can cover `SamplerBuilder` alone, and the value this test would have.

The code that throws exceptions is not in the builders, which are very simple, but in the various CreateXXX() methods from the sdk builder class.

Testing this in isolation is possible for low level leaves in the model tree (say, create a console exporter alone), but becomes more and more difficult when the code moves up in the model tree.

To test a tracer_provider for example, the test will need to create the tracer_provider, samplers, limits, arrays of processors, and exporters.

This becomes a full functional test, so the way to test this is to implement a full functional test:

* the input is a yaml config file
* the parser is used to create a model
* the model is used to create an sdk, executing the sdk builder code

This is still work in progress (not ready for review), but you can peek at the main PR.

There is an example of test that covers an exception thrown by the sdk builder:

(main PR) functional/configuration/shelltests/sampler_jaeger.test

It tests the code from `SdkBuilder::CreateJaegerRemoteSampler()`.

dbarker: Thanks for the context. Testing the sdk builder and private classes in a set of functional test makes sense.

sdk/include/opentelemetry/sdk/configuration/configured_sdk.h
```diff
@@ -0,0 +1,56 @@
...
+        meter_provider_(meter_provider),
+        logger_provider_(logger_provider)
+  {}
+
+  void Init();
+  void Cleanup();
+
+private:
+  opentelemetry::sdk::resource::Resource resource_;
+  std::shared_ptr<opentelemetry::sdk::trace::TracerProvider> tracer_provider_;
```
dbarker: The lifetime of the providers (and when they shutdown) depends on usage of Init and Cleanup. Just thinking through some scenarios. 
1. Providers live on after ConfigureSdk is destroyed
    - User calls Create and owns a ConfiguredSdk instance
    - User calls Init but never calls Cleanup
    - User destroys the ConfiguredSdk instance
    - Providers live on in the global singletons
2. Providers shutdown with destruction of ConfiguredSdk
    - User calls Create and owns a ConfiguredSdk instance
    - User calls Init
    - User calls Cleanup
    - User destroys the ConfiguredSdk instance -> providers shut down on destruction 

Should this class also shutdown/destroy the providers in the Cleanup method, or would it make sense for an instance of this class to manage the lifetime of the providers and propagator (and state of the global singletons) through its constructor/destructor?

dbarker: @marcalff After a bit more thought and looking at how I'll integrate the yaml configuration feature into my project I think I misinterpreted the intent of the ConfiguredSdk class. 

It seems the original intent is to just provide access to the objects (resource, providers, propagator) created by the yaml parser/builder with optional utility methods (Install/Uninstall) to set/unset the singletons. Access to the created objects seems like the most important interface and will allow users to take ownership by copying the shared ptrs and explicitly call public methods on sdk providers (likely ForceFlush and Shutdown). Setting the objects into the singletons seems like it should remain optional.

If someone chooses to set objects to the singletons then they can choose to unset them or not (if they prefer to rely on the static storage scope for the lifetime). 

In that case the original naming and public access to the providers/resource/propagator does seem best.

It seems like we should revert to your original design for this class. Sorry for the back and forth.

marcalff: I also took some time to look again at this area.

Without a yaml configuration file:

* the application creates a tracer_provider, by invoking various SDK apis to build the config with code
* the application then owns the tracer_provider, and sets global singletons to use it
* the application calls ForceFlush() or Shutdown() on the SDK tracer_provider, if necessary
* the application clears the tracer provider singleton to cleanup

With a yaml configuration file:

* the application passes a `config.yaml` file to the parser
* the parser builds a configuration model
* the sdk builder reads the configuration model, and produces a tracer_provider, wrapped in a `ConfiguredSdk` object.

From this point, the rest of the flow is the same:

* the application then owns the `ConfiguredSdk::tracer_provider`, and sets global singletons to use it
* the application calls ForceFlush() or Shutdown() on the SDK tracer_provider, if necessary
* the application clears the tracer provider singleton to cleanup

So, I think it makes sense to:

* Wrap all the parts produced by the SDK builder in a class
* Provide public access to these parts, like tracer_provider, because the application needs to see (SDK) `opentelemetry::sdk::trace::tracer_provider` instead of just (API) `opentelemetry::trace::tracer_provider`, to invoke ForceFlush or Shutdown

I reverted `ConfiguredSdk` to the initial code.

This PR is primarily about the 2K lines that are the SDK builder itself, to produce the `ConfiguredSdk` final result.

If there are some changes needed to ConfiguredSdk, I propose to see that with the example code, which uses it, so that the review for the SDK builder itself is not blocked by this.

marcalff:
@dbarker Thanks for this round of comments.

All addressed, PTAL.

dbarker (approved):
Looks good. Thanks!

PR #3553, opened 2025-07-20
[SDK] Implements options for the ParentBasedSampler with default values

### LINKED ISSUE #3545: [SDK] ParentBasedSampler is missing parameters
ParentBasedSampler is mising the following parameters:
- remoteParentSampled
- remoteParentNotSampled
- localParentSampled
- localParentNotSampled

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#parentbased

The parameters might be optional, meaning the caller is not forced to provide them, but this does not mean implementing support for these parameters is optional.

mathieurobin1:
Fixes #3545 

## Changes

This PR updates the ParentBasedSampler to be fully compliant with the OpenTelemetry specification.

The previous implementation was a simplified version. This change introduces the complete logic, which distinguishes between local and remote parents, and whether they were sampled. The constructor now accepts five configurable delegate samplers with defaults as defined in the spec (`AlwaysOnSampler` and `AlwaysOffSampler`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,9 @@ Increment the:
 
 * [SDK] Update default exemplar reservoir size for exponential histograms
   [#3551](https://github.com/open-telemetry/opentelemetry-cpp/pull/3551)
-  
+
 * [SDK] Implements options for the ParentBasedSampler with default values
-  [#3549](https://github.com/open-telemetry/opentelemetry-cpp/pull/3553)
+  [#3553](https://github.com/open-telemetry/opentelemetry-cpp/pull/3553)
 
 ## [1.22 2025-07-11]
 
--- a/sdk/src/trace/samplers/parent.cc
+++ b/sdk/src/trace/samplers/parent.cc
@@ -1,7 +1,6 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-#include <map>
 #include <memory>
 #include <string>
 


sdk/include/opentelemetry/sdk/trace/samplers/parent.h
```diff
@@ -21,16 +23,36 @@ namespace trace
...
-   * @return Returns DROP always
+  explicit ParentBasedSampler(const std::shared_ptr<Sampler> &root_sampler,
+                              const std::shared_ptr<Sampler> &remote_parent_sampled_sampler =
+                                  std::make_shared<AlwaysOnSampler>(),
+                              const std::shared_ptr<Sampler> &remote_parent_nonsampled_sampler =
+                                  std::make_shared<AlwaysOffSampler>(),
+                              const std::shared_ptr<Sampler> &local_parent_sampled_sampler =
+                                  std::make_shared<AlwaysOnSampler>(),
+                              const std::shared_ptr<Sampler> &local_parent_nonsampled_sampler =
+                                  std::make_shared<AlwaysOffSampler>()) noexcept;
```
marcalff, line 51: The default values for each parameter should not be handled at this level,
in the `ParentBasedSampler` constructor.

Make all parameters mandatory.

Instead, the logic should move to the Factory::Create()

mathieurobin1: Thank you for your feedback! I certainly understand the objective of standardizing the API pattern within the SDK. My initial inclination was actually to structure it that way, but after further consideration and experimentation, I've identified a couple of potential implications with this approach:

1.  **Risk of Breaking Existing Code:**
    While the recommended path for end-users might be through the `Factory::Create()` method, it's possible that some users might prefer or already be directly instantiating ParentBasedSampler via its constructor.
    If we remove the default arguments from the constructor, any existing code that directly calls a form of `new ParentBasedSampler(myDelegateSampler)` would immediately **break**, losing a potential backward compatibility.

2.  **Impact on Testability and Test Suite Maintenance:**
    A secondary, though notable, point is the effect on the existing test suite. Removing default values from the constructor would necessitate changes in numerous tests that currently instantiate `ParentBasedSampler` directly. This would require either explicitly passing all five sampler instances in each test case (which could make tests more verbose and harder to read) or refactoring tests to use the `Factory::Create()` method (which might obscure the specific dependency being tested). 

I've outlined these points as potential implications for API consumers and our test infrastructure. If you believe these aspects are not significant or are acceptable for this change, I am absolutely fine with proceeding as you proposed and will adapt the tests accordingly.

Thank you again for your feedback on this!

lalitb (approved):

sdk/include/opentelemetry/sdk/trace/samplers/parent.h
marcalff: Ok with the change as is then, to not break existing code.

marcalff (approved):

PR #3562, opened 2025-07-27
[DOC] Document minimum required versions

### LINKED ISSUE #3560: CMake configuration on Ubuntu 20.04 fails with v1.22.0
**Describe your environment** 

Building in docker on ubuntu:20.04 (Image ID b7bab04fd9aa) with libraries from the operating system. CMake version appears to be 3.16 on Ubuntu 20.04 and since the main `CMakeLists.txt` has `cmake_minimum_required(VERSION 3.14)` this should still be fine.

Tested versions:

|Version|Result |
|-----------|----------|
|v1.20.0 |OK       |
|v1.21.0 |OK       |
|v1.22.0 |ERROR|
|main     |ERROR|

The offending commit appears to be 3eec22c9a05ca4ac7a7bae4cf272954ceb6a1d83.

**Steps to reproduce**

Save the following script as `testing-ubuntu.sh`:
```
#!/bin/sh                                                                                                                                     
#                                                                                                                                             
# Usage: ./testing-ubuntu.sh GIT_TAG                                                                                                          
#                                                                                                                                             

git_tag=$1

apt update
apt -y install git g++ protobuf-compiler libprotobuf-dev cmake libcurl4-openssl-dev

git clone -j 32 --recurse-submodules --shallow-submodules --depth=1 --branch=$git_tag https://github.com/open-telemetry/opentelemetry-cpp.git
mkdir /build && cd /build
cmake ../opentelemetry-cpp -DCMAKE_INSTALL_PREFIX=$PWD -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWITH_ASYNC_EXPORT_PREVIEW=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=Y -DBUILD_SHARED_LIBS=N -DWITH_ABI_VERSION_1=N -DWITH_ABI_VERSION_2=Y -DWITH_STL=CXX17 -DWITH_ABSEIL=N -DBUILD_TESTING=N -DWITH_BENCHMARK=N -DWITH_FUNC_TESTS=N -DWITH_EXAMPLES=N -DWITH_OTLP_GRPC=N -DWITH_OTLP_HTTP=Y -DWITH_OTLP_HTTP_COMPRESSION=N
make
```
Then run the following commands to reproduce the problem:
```
chmod +x testing-ubuntu.sh
docker run -d --rm --name=test-container ubuntu:20.04 sleep 11111
docker cp testing-ubuntu.sh test-container:/
docker exec -ti test-container /testing-ubuntu.sh v1.22.0
```

With version v1.22.0, the `cmake` step will fail with:
```
-- Using the single-header code from /opentelemetry-cpp/third_party/nlohmann-json/single_include/
CMake Error in third_party/nlohmann-json/CMakeLists.txt:
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "CXX_INCLUDE_WHAT_YOU_USE" is not allowed.


CMake Error in third_party/nlohmann-json/CMakeLists.txt:
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "CXX_CLANG_TIDY" is not allowed.
```

With version v1.21.0 and older, this issue does not appear. Reverting commit 3eec22c9a05ca4ac7a7bae4cf272954ceb6a1d83 also fixes the problem.

None of these problems appear with Ubuntu 22.04 or newer.

I expected the configuration to complete without errors or to fail with an error reporting that this version of CMake is not supported.

CMake configuration fails as a result of passing incompatible properties.

**Additional context**

Additionally, even if the configuration step is fixed, it appears that the version of `protoc` is not new enough as the build fails with the following error:
```
[  0%] [Run]: "/usr/bin/protoc" "--proto_path=/opentelemetry-cpp/third_party/opentelemetry-proto" "--cpp_out=/build/generated/third_party/opentelemetry-proto" "--experimental_allow_proto3_optional" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/common/v1/common.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/resource/v1/resource.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/trace/v1/trace.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/logs/v1/logs.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/metrics/v1/metrics.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/profiles/v1development/profiles.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/collector/trace/v1/trace_service.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/collector/logs/v1/logs_service.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto" "/opentelemetry-cpp/third_party/opentelemetry-proto/opentelemetry/proto/collector/profiles/v1development/profiles_service.proto"
Unknown flag: --experimental_allow_proto3_optional
```

I believe that there are implicit version dependencies on the tools that are used and there are missing version checks in the CMake files which fail to report them. I couldn't find any mention of required versions for the various components so I assume that it should work with what is in the OS repositories.

I'm attempting to find out what are the minimum supported versions of the various libraries that opentelemetry-cpp requires to successfully build to provide at least the HTTP protocol metrics exporting. I'm currently facing problems on Ubuntu 20.04 which, while in practice it's EOL since it's out of Maintenance & Security Support, is still claimed in `main` to be a "supported" platform (in the CI list) and thus I'd expect it to work. What I'm actually after is trying to get it to build on Rocky Linux 8 which also has an older version of protobuf in the OS repositories.

If there's a minimum version of protobuf that's required, which I believe there is as the comment next to the code that adds the `--experimental_allow_proto3_optional` flag says that it's added in version 3.13, documenting it in the dependencies and adding a check in CMake would be very much appreciated.

markus456:
The libraries that opentelemetry-cpp requires have minimum required versions which are higher than the ones that are natively available on the systems being tested. By listing the exact versions of the libraries, the dependencies are clearly stated which removes the guesswork from build problems.

This contribution is submitted by me as an individual.

Fixes #3560

## Changes


dbarker:
Thanks for the PR @markus456! The changes generally look good. Just a few CI failures on the `third_party_minimum` file link in the DocFX check and some markdown-lint errors. Please address them and I'll approve.

lalitb:
It would’ve been sufficient to just reference `third_party_minimum`. Maintaining version numbers in multiple places can lead to inconsistencies and is harder to keep in sync.

markus456:
That was what I initially was planning on doing but given that there are not that many external dependencies and that they appear to not change super frequently, I figured it'd be nicer for the end-user to just directly see it from the markdown document. If it's too much, it can of course be removed.

marcalff:
> It would’ve been sufficient to just reference `third_party_minimum`. Maintaining version numbers in multiple places can lead to inconsistencies and is harder to keep in sync.

This also helps ease of use, even if this means a (little) more maintenance on changes.

Having to follow a trail of bread crumbs just to figure out what is needed to start with opentelemetry-cpp increases the barrier of entry, and the cost of taking the wrong path (build failures as reported) right from the start can be a show stopper preventing adoption.

It is a tradeoff with maintenance, but has merits too.

lalitb:
> Having to follow a trail of bread crumbs just to figure out what is needed to start with opentelemetry-cpp increases the barrier of entry, and the cost of taking the wrong path (build failures as reported) right from the start can be a show stopper preventing adoption.

What if we keep the single source but make the path more obvious? Like adding a clear callout in INSTALL.md, something like "Quick Start Prerequisites". Or maybe auto-generate this information from canonical source.

Anyway, nothing against this PR - it's already merged and addresses a real user pain point. Just something to think about for future cases like this. For now, hopefully we can keep both places updated when versions change.

Issue #3548, opened 2025-07-16
[SDK] Implement resource detectors


marcalff:
The resource sdk spec:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#resource-detector-name

defines some well known resource detectors (container, host, process, service)

These resource detectors should be implemented in opentelemetry-cpp,
or possibly in opentelemetry-cpp-contrib in case of dependencies on a container runtime/library.

nikhilbhatia08:
Hey @marcalff!, I'm new to the repo and excited to get started. I'd love to begin contributing by working on this issue to familiarize myself with the SDK. Could you please assign it to me ? Thanks!

nikhilbhatia08:
So for this issue we need to implement the resource detectors (container, host, process, service) as their attributes definitions are defined in `/api/include/opentelemetry/semconv/incubating` and service attributes in `/api/include/opentelemetry/semconv`, so for each resource I'm planning to add them in `resource_detector.cc` for example ContainerResourceDetector will be appended in `resource_detector.cc` and so on for other resources such as host and process. If there are any changes then I would be happy to follow them. Thanks! :)

dbarker:
Where to add the new resource detectors would be a good topic for a SIG meeting. The spec seems clear that many should not be in the SDK "package". 

> Custom resource detectors related to generic platforms (e.g. Docker, Kubernetes) or vendor specific environments (e.g. EKS, AKS, GKE) MUST be implemented as packages separate from the SDK.

https://opentelemetry.io/docs/specs/otel/resource/sdk/#detecting-resource-information-from-the-environment

We could consider adding the detectors as separate libraries/targets in this repo or to the opentelemetry-cpp-contrib repo.

dbarker:
After looking into the spec more, here are some notes and a proposal for adding the four "built-in" resource detectors (container, host, process, service).  Looking for feedback here in the thread and/or live in the next SIG meeting. 

**Notes:** 
- The spec seems to make a distinction between "built-in" detectors and "Custom" detectors
- Implementing the "built-in" detectors in `opentelemetry-cpp` and "Custom" detectors in `opentelemetry-cpp-contrib` seems reasonable
- All four "built-in" detectors may require platform dependent implementations 
- Users may reasonably expect support for the tested platforms - Linux, MacOS, and Windows. 
- A PR to add a detector may not fully implement support for all of these platforms. It should be documented clearly and tested. 
[...]

PR #3572, opened 2025-07-29
[SDK] Implementation of container resource as per semconv

nikhilbhatia08:
Fixes # (issue)
Partially fixes [#3548](https://github.com/open-telemetry/opentelemetry-cpp/issues/3548)

## Changes
Added container resource detector only for collecting info about `container.id` from `/proc/self/cgroup` and added test coverage to it, but could not add test coverage to `Detect()` of ContainerResource


sdk/src/common/container.cc
```diff
@@ -0,0 +1,53 @@
...
+    std::string container_id = ExtractContainerIDFromLine(line);
+    if(!container_id.empty()) 
+    {
+      return container_id;  
+    }
+  }
+  return "";
+}
+
+std::string ExtractContainerIDFromLine(std::string &line)
```
malkia: const std::string& line?

sdk/src/resource/resource_detector.cc
```diff
@@ -21,6 +23,10 @@ namespace resource
 
 const char *OTEL_RESOURCE_ATTRIBUTES = "OTEL_RESOURCE_ATTRIBUTES";
 const char *OTEL_SERVICE_NAME        = "OTEL_SERVICE_NAME";
+/**
+ * This is the file path from where we can get container.id
+*/
+const char *C_GROUP_PATH = "/proc/self/cgroup";
```
lalitb, line 29: Follow Google style guide for const variable naming - https://google.github.io/styleguide/cppguide.html#Constant_Names

lalitb: While you are here, can you rename these const variables (new and existing) to match the google styleguide - https://google.github.io/styleguide/cppguide.html#Constant_Names

sdk/src/resource/resource_detector.cc
```diff
@@ -68,6 +74,23 @@ Resource OTELResourceDetector::Detect() noexcept
   return ResourceDetector::Create(attributes);
 }
 
+Resource ContainerResourceDetector::Detect() noexcept
```
lalitb: Can we have separate file for each resource detector. It would be easy to organize that way ?

nikhilbhatia08: Yeah Sure

nikhilbhatia08: Applied all the changes mentioned above. Thanks!

sdk/include/opentelemetry/sdk/common/container.h
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+  Matches the line with the regex to find container.id
+  @param line a single line of text, typically from the /proc/self/cgroup file
+  @return matched id or empty string
+*/
+std::string ExtractContainerIDFromLine(const std::string &line);
+}  // namespace common
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
```
lalitb: This file should also be under resource directory.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
```diff
@@ -39,6 +39,17 @@ class OTELResourceDetector : public ResourceDetector
...
+ * ContainerResourceDetector to detect resource attributes when running inside a containerized
+ * environment. This detector extracts metadata such as container ID from cgroup information and
+ * sets attributes like container.id following the OpenTelemetry semantic conventions.
+ */
+class ContainerResourceDetector : public ResourceDetector
+{
+public:
+  Resource Detect() noexcept override;
+};
+
```
lalitb: `ContainerResourceDetector `should be under separate file, and content of common/container.h should also be within that?

sdk/src/resource/resource.cc
```diff
@@ -40,9 +41,12 @@ Resource Resource::Merge(const Resource &other) const noexcept
 
 Resource Resource::Create(const ResourceAttributes &attributes, const std::string &schema_url)
 {
-  static auto otel_resource = OTELResourceDetector().Detect();
-  auto resource =
-      Resource::GetDefault().Merge(otel_resource).Merge(Resource{attributes, schema_url});
+  static auto otel_resource      = OTELResourceDetector().Detect();
+  static auto container_resource = ContainerResourceDetector().Detect();
```
dbarker: It seems unexpected to add all new resource detectors to the create method. Some resource detectors may only work on certain platforms or environments and it seems more reasonable that they should be opt-in (ie: a user will create a resource with the detector and call Merge as needed in their application).  
https://opentelemetry.io/docs/specs/otel/resource/sdk/#create

sdk/src/resource/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-add_library(opentelemetry_resources resource.cc resource_detector.cc)
+add_library(opentelemetry_resources resource.cc resource_detector.cc
```
dbarker: The spec seems to indicate that we shouldn't include the custom resource detectors in the sdk "package". Perhaps this could be a separate CMake target (one that links to `opentelemetry-cpp::resource`) in this repo or these should be implemented in `opentelemetry-cpp-contrib`. This may be worth discussion in the next SIG meeting. 

> Custom resource detectors related to generic platforms (e.g. Docker, Kubernetes) or vendor specific environments (e.g. EKS, AKS, GKE) MUST be implemented as packages separate from the SDK.

https://opentelemetry.io/docs/specs/otel/resource/sdk/#detecting-resource-information-from-the-environment

sdk/src/resource/container_detector.cc
```diff
@@ -0,0 +1,41 @@
...
+{
+
+/**
+ * This is the file path from where we can get container.id
+ */
+const char *KCGroupPath = "/proc/self/cgroup";
+
+Resource ContainerResourceDetector::Detect() noexcept
+{
+  std::string container_id = opentelemetry::sdk::resource::GetContainerIDFromCgroup(KCGroupPath);
```
dbarker: Since this only looks at "/proc/self/cgroup" it will fail to read the file on Windows. We'll need to consider how we indicate which platforms are supported by a resource detector and should log the error to the internal logger. 

See https://opentelemetry.io/docs/specs/otel/resource/sdk/#detecting-resource-information-from-the-environment

> Errors should be handled as specified in the [Error Handling principles](https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles). Note the failure to detect any resource information MUST NOT be considered an error, whereas an error that occurs during an attempt to detect resource information SHOULD be considered an error.

sdk/src/resource/CMakeLists.txt
lalitb: > Perhaps this could be a separate CMake target (one that links to opentelemetry-cpp::resource) in this repo or these should be implemented in opentelemetry-cpp-contrib.

 Agree to this -  Needs more discussion to finalize. I think most of the SIGs place it under respective contrib repos.

dbarker: I've added some more context and a proposal for how to incorporate the new resource detectors into this repo here
https://github.com/open-telemetry/opentelemetry-cpp/issues/3548#issuecomment-3156148086

@lalitb can you take a look and give some feedback?

dbarker:
Thanks for the PR @nikhilbhatia08! I think we just need to hash out how best to include these "built-in" detectors in the repo/build systems since many will bring platform dependencies. I've posted my proposal to the initial issue thread.

resource_detectors/container.cc
```diff
@@ -0,0 +1,47 @@
...
+    std::string container_id = ExtractContainerIDFromLine(line);
+    if (!container_id.empty())
+    {
+      return container_id;
+    }
+  }
+  return "";
+}
+
+std::string ExtractContainerIDFromLine(const std::string &line)
```
rafaelroquetto:
```suggestion
std::string ExtractContainerIDFromLine(std::string_view line)
```
rationale: since you aren't really storing `line`, passing `string_view` makes this function more versatile, as it's now able to seamlessly work with both `std::string` or `const char*` for instance, whilst also conveying the semantic that this is not really storing the string.

resource_detectors/container.cc
```diff
@@ -0,0 +1,47 @@
...
+{
+  static const std::regex container_id_regex(R"(^.*/(?:.*[-:])?([0-9a-f]+)(?:\.|\s*$))");
```
rafaelroquetto: Would you mind adding a comment explaining what kind of input this regexp is meant to parse? For instance:

```
# Docker cgroup v1
10:memory:/docker/3fae9b2c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef0
9:cpu,cpuacct:/docker/3fae9b2c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef0
8:cpuset:/docker/3fae9b2c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef0

# Docker cgroup v2 (unified)
0::/docker/3fae9b2c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef0

# Docker cgroup v2 (systemd driver)
0::/system.slice/docker-09ec67119d38768dbf7994d81c325e2267214428a3c2e581c81557e3650863d8.scope

# containerd (Kubernetes, cgroup v2)
0::/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod12345678-1234-5678-1234-567812345678.slice/cri-containerd-3fae9b2c6d7e8f90123456789abcdef0123456789abcdef0123456789abcdef0.scope

# CRI-O (Kubernetes, cgroup v2)
0::/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod87654321-4321-6789-4321-678987654321.slice/crio-bff4c347d3fc6a20064926fdfd1ea3c76e039c56205c6b282d3b6c8e2f13233c.scope

# Podman (rootless, host cgroupns)
0::/user.slice/user-1000.slice/user@1000.service/user.slice/libpod-aff58243c495b7409a036930a467602d82226bc1e8f7c50eb3a8a439919b505c.scope/container
```

resource_detectors/container.cc
```diff
@@ -0,0 +1,47 @@
...
+  std::smatch match;
+
+  if (std::regex_search(line, match, container_id_regex))
+  {
+    return match.str(1);
+  }
+
+  return "";
```
rafaelroquetto: As a rule of thumb, when returning empty strings, it's usually better (at least theoretically, as the compiler is smart enough to optimise things) to write `return {}` or `return std::string()` - return `""` actually invokes the `std::string` constructor that takes `const char*`, whilst the former examples invoke the default constructor, which is theoretically cheaper.

resource_detectors/container_detector.cc
```diff
@@ -0,0 +1,41 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+
+/**
+ * This is the file path from where we can get container.id
+ */
+const char *KCGroupPath = "/proc/self/cgroup";
```
rafaelroquetto:
```suggestion
constexpr const char *KCGroupPath = "/proc/self/cgroup";
```

or

```suggestion
const char KCGroupPath[] = "/proc/self/cgroup";
```

this prevents unintended assignments to `KCGroupPath`, i.e. with the present code, one can do this: https://godbolt.org/z/P36456bEE

resource_detectors/container_detector.cc
```diff
@@ -0,0 +1,41 @@
...
+{
+  std::string container_id = opentelemetry::sdk::resource::GetContainerIDFromCgroup(KCGroupPath);
+  if (container_id.empty())
+  {
+    return ResourceDetector::Create({});
+  }
+
+  ResourceAttributes attributes;
+
+  attributes[semconv::container::kContainerId] = container_id;
```
rafaelroquetto:
```suggestion
  attributes[semconv::container::kContainerId] = std::move(container_id);
```

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
```diff
@@ -0,0 +1,43 @@
...
+public:
+  Resource Detect() noexcept override;
+};
+
+/**
+ * Reads the container.id from /proc/self/cgroup file.
+ * @param file_path file path of cgroup
+ * @return container.id as string or empty string
+ */
+std::string GetContainerIDFromCgroup(const char *file_path);
```
rafaelroquetto: should this also take a `std::string_view` to make it more versatile?

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Matches the line with the regex to find container.id
+ * @param line a single line of text, typically from the /proc/self/cgroup file
+ * @return matched id or empty string
+ */
+std::string ExtractContainerIDFromLine(const std::string &line);
```
rafaelroquetto: and here to? I think there's no downside, but up to you.

sdk/src/resource/resource_detector.cc
```diff
@@ -19,8 +19,8 @@ namespace sdk
 namespace resource
 {
 
-const char *OTEL_RESOURCE_ATTRIBUTES = "OTEL_RESOURCE_ATTRIBUTES";
-const char *OTEL_SERVICE_NAME        = "OTEL_SERVICE_NAME";
+const char *KOtelResourceAttributes = "OTEL_RESOURCE_ATTRIBUTES";
+const char *KOtelServiceName        = "OTEL_SERVICE_NAME";
```
rafaelroquetto: I am not sure what the maintainers think, so it's just a suggestion, but like my last comment, I'd turn these into `constexpr const char*` instead (but I am not sure if this would violate and perceived code consistency, so ultimately it's just a suggestion)

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
nikhilbhatia08: Actually I was thinking to do this. Will change it, Thanks!

resource_detectors/container.cc
nikhilbhatia08: If I keep it as `std::string_view` then I cannot use `regex_search` and I have to convert it to `std::string` to be able to use in `regex_search`

rafaelroquetto: regex_search has a constructor that takes iterators, so I believe you can do it like this:

```
std::regex_search(line.begin(), line.end(), match, container_id_regex);
```

resource_detectors/container_detector.cc
nikhilbhatia08: Thanks for the catch! In semconv also its mentioned as `constexpr const char*`. Thanks!

resource_detectors/container.cc
nikhilbhatia08: Yes, Thanks!

nikhilbhatia08: But For this to work I have to use this `std::match_results<std::string_view::const_iterator> match;` instead of `std::smatch match;` and then return the string like this `return std::string(match[1].first, match[1].second);`

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
nikhilbhatia08: But here ifstream does not take std::string_view, so should I convert it like `ifstream cgroup_file(std::string(file_path))` or let it be ?

rafaelroquetto: Oh you are right - in this case it kind of defeats the purpose, since a std::string needs to be constructed anyway - in that case, I'd just leave it as is then  - good catch!

resource_detectors/container.cc
rafaelroquetto: I'll reply in parts to make it easier:
 > For this to work I have to use this  std::match_results<std::string_view::const_iterator> match;
 
that's right - as you probably know, `smatch` is just a type alias to `std::match_results<std::string::const_iterator>` - I don't think it's an issue to spell out the entire type, since there's no equivalent alias for string_view.

> Then return the string like this return std::string(match[1].first, match[1].second);

I could be wrong here, so take this with a grain of salt, but I believe you only need to do that if your function returns a `string_view` as well, which I don't think should be done (i.e. it should still return a `std::string`). Otherwise, `return match.str(1)` should work: https://godbolt.org/z/ehhTYnM7K

nikhilbhatia08: Sorry the type should be nostd::string_view

nikhilbhatia08: `nostd::string_view` does not have a const::iterator, any idea for any workaround ?
(edit) found a workaround `std::match_results<const char *>` for regex_search to work with `nostd::string_view`

rafaelroquetto: libc++ `string_view` iterators are simply type aliases to `value_type` (where `value_type` is `char` for `string_view`).

i.e.
```
// for string_view, value_type == char
using const_iterator = const value_type*;
using iterator = const_iterator;
```

you can  therefore write `std::match_results<const char*> match;` - or even better (although perhaps not the scope of this PR), simply the above (using `typedef` instead of `using`) to the `nostd::string_view` definition under the `size_type` typedef.

nikhilbhatia08: I would Like to go with `std::match_results<const char*> match;` and additionally I would like to add that in msvc implementation of cpp20 we should use `std::regex_search(line.data(), line.data() + line.size(), match, container_id_regex)` because it was giving me this error 

```
D:\a\opentelemetry-cpp\opentelemetry-cpp\resource_detectors\container.cc(49,12): error C2672: 'std::regex_search': no matching overloaded function found [D:\a\opentelemetry-cpp\opentelemetry-cpp\build\resource_detectors\opentelemetry_resource_detectors.vcxproj]
      C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\regex(2325,17):
      could be 'bool std::regex_search(const std::basic_string<_Elem,_StTraits,_StAlloc> &,const std::basic_regex<_Elem,_RxTraits> &,std::regex_constants::match_flag_type)'
          D:\a\opentelemetry-cpp\opentelemetry-cpp\resource_detectors\container.cc(49,12):
          'bool std::regex_search(const std::basic_string<_Elem,_StTraits,_StAlloc> &,const std::basic_regex<_Elem,_RxTraits> &,std::regex_constants::match_flag_type)': expects 3 arguments - 4 provided
```

rafaelroquetto: Sounds good to me - thank you for going the extra mile with this :)

resource_detectors/container_detector.cc
```diff
@@ -0,0 +1,42 @@
...
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+
+/**
+ * This is the file path from where we can get container.id
+ */
+constexpr const char *KCGroupPath = "/proc/self/cgroup";
```
lalitb:
```suggestion
constexpr const char *kCGroupPath = "/proc/self/cgroup";
```

nikhilbhatia08: Then I should also change the same in `resource_detector.cc`

sdk/src/resource/resource_detector.cc
```diff
@@ -19,8 +19,8 @@ namespace sdk
 namespace resource
 {
 
-const char *OTEL_RESOURCE_ATTRIBUTES = "OTEL_RESOURCE_ATTRIBUTES";
-const char *OTEL_SERVICE_NAME        = "OTEL_SERVICE_NAME";
+constexpr const char *KOtelResourceAttributes = "OTEL_RESOURCE_ATTRIBUTES";
```
lalitb:
```suggestion
constexpr const char *kOtelResourceAttributes = "OTEL_RESOURCE_ATTRIBUTES";
```

sdk/src/resource/resource_detector.cc
```diff
@@ -19,8 +19,8 @@ namespace sdk
...
+constexpr const char *KOtelServiceName        = "OTEL_SERVICE_NAME";
```
lalitb:
```suggestion
constexpr const char *kOtelServiceName        = "OTEL_SERVICE_NAME";
```

nikhilbhatia08: Fixed this with the above commit

resource_detectors/container.cc
```diff
@@ -0,0 +1,58 @@
...
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace resource
+{
+std::string GetContainerIDFromCgroup(const char *file_path)
+{
+  std::ifstream cgroup_file(file_path);
```
lalitb: check if file fails to open - returning empty string in that case?

nikhilbhatia08: Okay I will add an extra test case for it.

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
```diff
@@ -0,0 +1,44 @@
...
+class ContainerResourceDetector : public ResourceDetector
+{
+public:
+  Resource Detect() noexcept override;
+};
+
+/**
+ * Reads the container.id from /proc/self/cgroup file.
+ * @param file_path file path of cgroup
+ * @return container.id as string or empty string
```
lalitb:
```suggestion
 * @return container.id as string or an empty string if not found or on error,
```

sdk/test/resource/BUILD
```diff
@@ -9,6 +9,7 @@ cc_test(
     tags = ["test"],
     deps = [
         "//api",
+        "//resource_detectors",
```
lalitb: better to have resource detector tests under `/resource_detectors/test`, and not under `sdk/test/resource`

nikhilbhatia08: Yeah sure.

resource_detectors/container_detector.cc
lalitb: yes.

sdk/src/resource/resource_detector.cc
lalitb: I agree on changing to `constexpr const char*`.

CMakeLists.txt
```diff
@@ -634,6 +634,7 @@ if(DEFINED OPENTELEMETRY_BUILD_DLL)
 endif()
 
 add_subdirectory(api)
+add_subdirectory(resource_detectors)
```
dbarker: resource_detectors will depend on the sdk targets. Please move this after line 646. Then we'll be able to link sdk targets to the resource detector targets.

resource_detectors/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+add_library(opentelemetry_resource_detectors container.cc container_detector.cc)
+
+set_target_properties(opentelemetry_resource_detectors
+                      PROPERTIES EXPORT_NAME resource_detectors)
+set_target_version(opentelemetry_resource_detectors)
+
+target_link_libraries(opentelemetry_resource_detectors
+                      PUBLIC opentelemetry_common)
```
dbarker: Please replace `opentelemetry_common` with `opentelemetry_resources` here. The `opentelemetry_resources` target has a public link to `opentelemetry_common`.

resource_detectors/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+target_include_directories(
+  opentelemetry_resource_detectors
+  PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/resource_detectors/include>"
+  PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/sdk/include>")
```
dbarker: Once the above link to `opentelemetry_resources` is made then please remove `PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/sdk/include>"` here. The necessary sdk include directories will be propagated through the target link.

resource_detectors/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+if(OPENTELEMETRY_INSTALL)
```
dbarker: Since we don't have a test for this pkgconfig package and not all opentelemetry-cpp targets are exposed as pkgconfig packages - please remove it for now and add the target to a cmake component instead. 

lines 18-23 can be replaced with: 
```cmake
otel_add_component(
  COMPONENT
  resource_detectors
  TARGETS
  opentelemetry_resource_detectors
  FILES_DIRECTORY
  "include/opentelemetry/"
  FILES_DESTINATION
  "include/opentelemetry"
  FILES_MATCHING
  PATTERN
  "*.h")
```

nikhilbhatia08: Sure.

sdk/test/resource/CMakeLists.txt
```diff
@@ -3,8 +3,9 @@
 
 foreach(testname resource_test)
   add_executable(${testname} "${testname}.cc")
-  target_link_libraries(${testname} ${GTEST_BOTH_LIBRARIES}
-                        ${CMAKE_THREAD_LIBS_INIT} opentelemetry_resources)
+  target_link_libraries(
```
dbarker: Please move the resource detectors test to its own test cmake file (no need to copy the foreach pattern used in this file when the test is moved).

nikhilbhatia08: Sorry I did not understand this one ?

resource_detectors/container_detector.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "opentelemetry/sdk/resource/resource_detector.h"
+#include "opentelemetry/semconv/incubating/container_attributes.h"
+#include "opentelemetry/version.h"
+
+#include <string>
+#include <unordered_map>
+#include <utility>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
```
dbarker: To follow the pattern of the namespaces for other directories (for example the exporters directory uses `opentelemetry::exporter::`), using the `opentelemetry::resource_detector` namespace for the resource_detectors directory seems more expected. @lalitb what do you think?

sdk/test/resource/CMakeLists.txt
dbarker: This PR shouldn't add tests to the `opentelemetry-cpp/sdk/test/resource` directory (should not change this CMake file). Please add the tests to `opentelemetry-cpp/resource_detectors/test` and an appropriate CMake file there.

dbarker: Same request as https://github.com/open-telemetry/opentelemetry-cpp/pull/3572/files#r2263746385

nikhilbhatia08: But I was asking about this `(no need to copy the foreach pattern used in this file when the test is moved).`

dbarker: Just requesting not to copy the foreach pattern to the resource detectors test cmake file.

Instead of
```cmake
foreach(testname resource_detectors_test)
  add_executable(${testname} "${testname}.cc")
  ... 
endforeach()
```
We can just use
```cmake
add_executable(resource_detectors_test "resource_detectors_test.cc")
...
```

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
```diff
@@ -0,0 +1,44 @@
+// Copyright The OpenTelemetry Authors
```
dbarker: Please name this header to match the name of the .cc file (`container_detector.h`).

resource_detectors/include/opentelemetry/resource_detectors/container_resource_detector.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+  Resource Detect() noexcept override;
+};
+
+/**
+ * Reads the container.id from /proc/self/cgroup file.
+ * @param file_path file path of cgroup
+ * @return container.id as string or empty string
+ */
+std::string GetContainerIDFromCgroup(const char *file_path);
```
dbarker: If these methods `GetContainerIDFromCgroup` and `ExtractContainerIDFromLine` are an implementation detail and not intended for public use, consider moving the declarations/definitions to a separate header and implementation files (`container_detector_utils.<h,cc>)` and put them in a `detail` namespace. Then we can exclude installing the `container_detector_utils.h` header when calling the `otel_add_component` CMake function.   This would support testing the methods but not expose them to sdk users.

resource_detectors/container_detector.cc
nikhilbhatia08: Yeah even I was thinking the same.

resource_detectors/CMakeLists.txt
dbarker: The target includes should be the following to ensure the build and install path to the headers is set properly. 

````cmake
target_include_directories(
  opentelemetry_resource_detectors
  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
         "$<INSTALL_INTERFACE:include>")
````

File: resource_detectors/container_detector.cc [hunk 31]
```diff
@@ -0,0 +1,44 @@
...
+#include "opentelemetry/version.h"
+
+#include <string>
+#include <unordered_map>
+#include <utility>
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace resource_detector
+{
+namespace resource
```
dbarker: This extra `resource` namespace can be removed. The class should be defined in the `resource_detector` namespace.

nikhilbhatia08: Fixed.

resource_detectors/container_detector.cc
lalitb: > To follow the pattern of the namespaces for other directories (for example the exporters directory uses opentelemetry::exporter::), using the opentelemetry::resource_detector namespace for the resource_detectors directory seems more expected. @lalitb what do you think?

Yes, agree on that.

resource_detectors/CMakeLists.txt
dbarker: @nikhilbhatia08 Thanks for all the updates! Please update the `target_include_directories` command based on the comment above and then I'll approve the PR.

CMakeLists.txt
```diff
@@ -645,6 +645,7 @@ if(NOT WITH_API_ONLY)
   add_subdirectory(sdk)
   add_subdirectory(ext)
   add_subdirectory(exporters)
+  add_subdirectory(resource_detectors)
```
lalitb: I suggest including the inbuilt resource detectors under a preview flag - similar to `WITH_OTLP_RETRY_PREVIEW` - to make it clear that both the implementation and contract  may change. The flag can be removed later once both specs and implementation is stable.

lalitb (approved):
LGTM, once preview flag is added and approved by @dbarker. Thanks @nikhilbhatia08 for contribution, nice work.

resource_detectors/CMakeLists.txt
nikhilbhatia08: Yeah Sure, Thanks!

CMakeLists.txt
nikhilbhatia08: Sorry @lalitb , I don't know where and how to add similar type of flag, please can you help me implementing it, Thanks!

lalitb: Something like below in CMakeLists.txt

```cmake
option(WITH_RESOURCE_DETECTORS_PREVIEW
       "Whether to enable inbuilt resource detectors" OFF)

if(WITH_RESOURCE_DETECTORS_PREVIEW)
  add_subdirectory(resource_detectors)
endif()
```

And then user can enable it through

```console
$ cmake -DWITH_RESOURCE_DETECTORS_PREVIEW=ON .
```

nikhilbhatia08: Thank you so much for helping!

CMakeLists.txt
```diff
@@ -261,6 +261,9 @@ option(WITH_METRICS_EXEMPLAR_PREVIEW
 option(WITH_THREAD_INSTRUMENTATION_PREVIEW
        "Whether to enable thread instrumentation" OFF)
 
+option(WITH_RESOURCE_DETECTORS_PREVIEW
```
dbarker, line 264: Please add this to the preview options cmake script here 
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/test_common/cmake/preview-options.cmake

This will ensure the option is turned on and tested in CI.

nikhilbhatia08: Yeah sure.
(edit) fixed.

dbarker (approved):
Looks good! Thanks for the PR. Congrats on your first contribution! Looking forward to the next one :)

nikhilbhatia08:
Thank you so much, @dbarker, @lalitb, and @rafaelroquetto, for helping me with my first contribution! Really appreciate it :) Looking forward to contributing more.

PR #3573, opened 2025-07-29
[CONFIGURATION] File configuration - configuration example

marcalff:
Contributes to #2481

This is a partial fix, to implement a configuration example application.

## Changes

* Implement an example application, using yaml configuration
* Implement example extension points

### DIFF
--- a/examples/configuration/custom_log_record_exporter.cc
+++ b/examples/configuration/custom_log_record_exporter.cc
@@ -1,0 +1,35 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "opentelemetry/sdk/common/global_log_handler.h"
+#include "opentelemetry/sdk/logs/read_write_log_record.h"
+#include "opentelemetry/sdk/logs/recordable.h"
+
+#include "custom_log_record_exporter.h"
+
+std::unique_ptr<opentelemetry::sdk::logs::Recordable>
+CustomLogRecordExporter::MakeRecordable() noexcept
+{
+  auto recordable = std::make_unique<opentelemetry::sdk::logs::ReadWriteLogRecord>();
+  return recordable;
+}
+
+opentelemetry::sdk::common::ExportResult CustomLogRecordExporter::Export(
+    const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>>
+        & /* records */) noexcept
+{
+  OTEL_INTERNAL_LOG_ERROR("CustomLogRecordExporter::Export(): YOUR CODE HERE");
+  return opentelemetry::sdk::common::ExportResult::kSuccess;
+}
+
+bool CustomLogRecordExporter::ForceFlush(std::chrono::microseconds /* timeout */) noexcept
+{
+  OTEL_INTERNAL_LOG_ERROR("CustomLogRecordExporter::ForceFlush(): YOUR CODE HERE");
+  return false;
+}
+
+bool CustomLogRecordExporter::Shutdown(std::chrono::microseconds /* timeout */) noexcept
+{
+  OTEL_INTERNAL_LOG_ERROR("CustomLogRecordExporter::Shutdown(): YOUR CODE HERE");
+  return false;
+}
--- a/examples/configuration/custom_span_exporter_builder.cc
+++ b/examples/configuration/custom_span_exporter_builder.cc
@@ -1,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <string>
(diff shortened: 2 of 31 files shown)


examples/configuration/custom_log_record_exporter.cc
```diff
@@ -0,0 +1,35 @@
...
+{
+  auto recordable = std::make_unique<opentelemetry::sdk::logs::ReadWriteLogRecord>();
+  return recordable;
+}
+
+opentelemetry::sdk::common::ExportResult CustomLogRecordExporter::Export(
+    const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>>
+        & /* records */) noexcept
+{
+  OTEL_INTERNAL_LOG_ERROR("CustomLogRecordExporter::Export(): YOUR CODE HERE");
```
dbarker, line 21: Logging an error and returning success seems unexpected here. Consider removing the log line or returning failure.

examples/configuration/custom_span_processor.cc
```diff
@@ -0,0 +1,44 @@
...
+{
+  auto recordable = std::make_unique<opentelemetry::sdk::trace::SpanData>();
+  return recordable;
+}
+
+void CustomSpanProcessor::OnStart(
+    opentelemetry::sdk::trace::Recordable & /* span */,
+    const opentelemetry::trace::SpanContext & /* parent_context */) noexcept
+{
+  OTEL_INTERNAL_LOG_ERROR("CustomSpanProcessor::OnStart(): YOUR CODE HERE");
```
dbarker, line 24: Should these internal log lines be debug or info instead of error?

examples/configuration/custom_span_processor.h
```diff
@@ -0,0 +1,37 @@
...
+               const opentelemetry::trace::SpanContext &parent_context) noexcept override;
+
+  void OnEnd(std::unique_ptr<opentelemetry::sdk::trace::Recordable> &&span) noexcept override;
+
+  bool ForceFlush(std::chrono::microseconds timeout) noexcept override;
+
+  bool Shutdown(std::chrono::microseconds timeout) noexcept override;
+
+private:
+  std::string m_comment;
```
dbarker: nit - style

examples/configuration/custom_span_exporter_builder.cc
```diff
@@ -0,0 +1,30 @@
...
+#include "custom_span_exporter.h"
+#include "custom_span_exporter_builder.h"
+
+std::unique_ptr<opentelemetry::sdk::trace::SpanExporter> CustomSpanExporterBuilder::Build(
+    const opentelemetry::sdk::configuration::ExtensionSpanExporterConfiguration *model) const
+{
+  // Read yaml attributes
+  std::string comment = model->node->GetRequiredString("comment");
+
+  auto sdk = std::make_unique<CustomSpanExporter>(comment);
```
dbarker, line 21: The "comment" value doesn't seem to be used or stored with the CustomSpanExporter or CustomSampler. Should it be a member of these classes?

examples/configuration/custom_log_record_exporter.cc
marcalff: This is template code, to show how to implement a custom extension point (here a log record exporter), and register it with the registry, so it becomes available in the yaml configuration file.

By definition, this template code is incomplete and not functional.

Logging an error is the safest way to make sure people will notice it, so they know where in the code to provide the missing implementation for an extension point.

The most important part is "YOUR CODE HERE".

The code returns ok, acting as a no-op extension point, given it is not implemented (it does nothing).

examples/configuration/custom_span_processor.cc
marcalff: Error is more visible, pointing exactly which part of the template code needs to be provided.

What the example is about:

* show how to expose an extension point to the yaml machinery, by registering a custom builder.

What the example is not about:

* implement a fully functional extension point, with or without yaml.

examples/configuration/custom_span_processor.h
marcalff: Fixed.

examples/configuration/custom_span_exporter_builder.cc
marcalff: This is template code.

To implement a custom span exporter for example, for a different protocol, the custom span exporter will most likely need some properties defined in yaml, like an endpoint, and possibly more properties.

This code shows how to extract a property from the yaml configuration file, and use it in the code.

Using the same pattern:

```
  std::string my_endpoint = model->node->GetRequiredString("my_endpoint");
```

is typically what people will need, so the code provides an example about how to extract data.

Indeed the comment property is not used, but the point here is to show how to get the value for it, not how to use it.

PR #3591, opened 2025-08-13
[resource_detectors] implementation of process resource detector as per semconv

nikhilbhatia08:
Fixes # (issue)

## Changes
Implements process detector currently for attributes - [process.pid, process.executable.path, process.command], the next pull request will implement more of the process attributes. Added tests.

reference - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/process.md#selecting-process-attributes


nikhilbhatia08:
The tests on the latest commit fail because psapi.h is included first and then windows.h, but if we try to include windows.h first and then include psapi.h then it would fail the format test. What can be done for this any suggestions ?

lalitb:
> The tests on the latest commit fail because psapi.h is included first and then windows.h, but if we try to include windows.h first and then include psapi.h then it would fail the format test. What can be done for this any suggestions ?

What is the format error? You switch off the clang/iwyu check for that portion of the include, depending on the error you are getting.

nikhilbhatia08:
If I run format of ci then I'm getting this error

```
The following files have changes:
resource_detectors/process_detector_utils.cc
diff --git a/resource_detectors/process_detector_utils.cc b/resource_detectors/process_detector_utils.cc
index cb6e064..d20f367 100644
--- a/resource_detectors/process_detector_utils.cc
+++ b/resource_detectors/process_detector_utils.cc
@@ -7,8 +7,8 @@
 #include <string>
 
 #ifdef _MSC_VER
-#  include <windows.h>
 #  include <psapi.h>
+#  include <windows.h>
 #else
 #  include <sys/types.h>
 #  include <unistd.h>
```

So if I include psapi.h first and then windows.h, the windows.h header redefines some macros present in psapi.h, that is why windows.h should be included first, but doing so gives the above format error.

But now I got how to disable the format error for this part of code. Thanks 👍🏻 !

resource_detectors/process_detector_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+  if (!wcmd)
+    return std::string();
+
+  // Convert UTF-16 to UTF-8
+  int size_needed = WideCharToMultiByte(CP_UTF8, 0, wcmd, -1, NULL, 0, NULL, NULL);
+  if (size_needed <= 0)
+    return std::string();
+
+  std::string utf8_command(size_needed - 1, 0);  // exclude null terminator
+  WideCharToMultiByte(CP_UTF8, 0, wcmd, -1, &utf8_command[0], size_needed, NULL, NULL);
```
dbarker, line 90: Can this still fail and return zero?

resource_detectors/process_detector_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+#endif
+}
+
+std::string GetCommand(const int32_t &pid)
+{
+#ifdef _MSC_VER
+  // On Windows, GetCommandLineW only works for the CURRENT process,
+  // so we ignore `pid` and just return the current process's command line.
+  LPCWSTR wcmd = GetCommandLineW();
+  if (!wcmd)
```
dbarker, line 77: nit - style. missing braces.

resource_detectors/process_detector_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+    return std::string();
+
+  // Convert UTF-16 to UTF-8
+  int size_needed = WideCharToMultiByte(CP_UTF8, 0, wcmd, -1, NULL, 0, NULL, NULL);
+  if (size_needed <= 0)
```
dbarker, line 84: nit - style. missing braces.

resource_detectors/process_detector.cc
```diff
@@ -0,0 +1,50 @@
...
+#  define getpid _getpid
+#else
+#  include <unistd.h>
+#endif
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace resource_detector
+{
+
+opentelemetry::sdk::resource::Resource ProcessResourceDetector::Detect() noexcept
```
dbarker, line 31: Please check if the utility functions can throw exceptions. If so please handle them appropriately so Detect does not result in a call to std::terminate. Check out the otel error handling policy for guidance https://opentelemetry.io/docs/specs/otel/error-handling/.

resource_detectors/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-add_library(opentelemetry_resource_detectors container_detector_utils.cc
-                                             container_detector.cc)
+add_library(
+  opentelemetry_resource_detectors
+  container_detector_utils.cc container_detector.cc process_detector.cc
+  process_detector_utils.cc)
```
dbarker, line 7: Since we now have multiple "detail" files for the utility methods please create a detail folder in "include/opentelemetry/resource_detectors" and put them all there. Then it will be clear to SDK users that these headers should not be used. The detail folder can also then be excluded from install with cmake (through the otel_add_component function).

resource_detectors/process_detector.cc
nikhilbhatia08: But I'm using `utility` for `std::move` and it's a pure cast which can not throw any error and will not resut in std::terminate

dbarker:
Thanks for the PR! Please see comments above.

resource_detectors/process_detector.cc
dbarker: I'm thinking of the methods from `process_detector_utils.h` (or any functions called by the Detect implementation). Since Detect is marked noexcept any unhandled exceptions will result in std::terminate.

nikhilbhatia08: Then I will wrap the Detect function using a `try catch` block and then use `OTEL_INTERNAL_LOG_ERROR`, will that be ok ?

resource_detectors/process_detector.cc
```diff
@@ -0,0 +1,50 @@
...
+{
+  int32_t pid = getpid();
```
dbarker, line 33: getpid() appears deprecated on Windows. This may need OS specific handling. 
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-getpid?view=msvc-170

resource_detectors/process_detector.cc
dbarker: Wrapping each of the process_detector_utils.h method calls in try/catch seams reasonable if they can throw.

resource_detectors/process_detector.cc
nikhilbhatia08: Actually we are using` _getpid` just used a `#define getpid _getpid`

resource_detectors/process_detector_utils.cc
nikhilbhatia08: Not sure, but as it is getting checked above for the zero condition, I don't think that it will return zero below after already it being checked.

resource_detectors/process_detector.cc
dbarker: Thanks. I missed that.

resource_detectors/test/process_detector_test.cc
```diff
@@ -0,0 +1,48 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <stdint.h>
+#include <cstdio>
+#include <fstream>
+#include <string>
+
+#include "opentelemetry/resource_detectors/process_detector_utils.h"
```
dbarker: Can `GetCommand` and `GetExecutablePath` be tested?

nikhilbhatia08: Yes they can be tested, but we would need to off the test for macos. And then to test it we could normally fetch it as how we are fetching the attributes using the function

nikhilbhatia08:
For the tests added now and resource detectors tests in general we need to enable the tests `-DWITH_RESOURCE_DETECTORS_PREVIEW=ON` or else they will go untested as the actions ran in the previous commit.

resource_detectors/include/opentelemetry/resource_detectors/detail/process_detector_utils.h
```diff
@@ -0,0 +1,51 @@
...
+ * for example - /proc/<pid>/cmdline, /proc/<pid>/exe
+ */
+std::string FormFilePath(const int32_t &pid, const char *process_type);
+
+/**
+ * Retrieves the absolute file system path to the executable for a given PID.
+ *
+ * Platform-specific behavior:
+ *   - Windows: Uses OpenProcess() + GetProcessImageFileNameW().
+ *   - Linux/Unix: Reads the /proc/<pid>/exe symbolic link.
```
lalitb, line 27: nit - Add a TODO for Darwin?

nikhilbhatia08: Yeah sure.

resource_detectors/process_detector_utils.cc
```diff
@@ -0,0 +1,117 @@
...
+  std::ifstream command_line_file(command_line_path, std::ios::in | std::ios::binary);
+  std::getline(command_line_file, command, '\0');
+  return command;
+}
+
+std::string FormFilePath(const int32_t &pid, const char *process_type)
+{
+  char buff[64];
+  int len = std::snprintf(buff, sizeof(buff), "/proc/%d/%s", pid, process_type);
+  return std::string(buff, len);
```
lalitb, line 121: There's an issue here - `snprintf` returns the number of characters that would potentially be written, which can exceed the buffer length. This means we attempt to create a std::string using more characters than the buffer actually contains, leading to undefined behavior.

nikhilbhatia08: If the len variable exceeds the buffer length then we can just return the string with buffer length like 
`return std::string(buff, sizeof(buff) - 1)` so that it does not crash.

PR #3599, opened 2025-08-21
[CMAKE] upgrade cmake minimum version to 3.16

### LINKED ISSUE #3598: Upgrade CMake minimum required version to 3.16
In order to upgrade several of the project's dependencies (googletest, benchmark, grpc, protobuf) to more recent versions (See https://github.com/open-telemetry/opentelemetry-cpp/issues/3525) we'll need to upgrade the CMake minimum required version to 3.16. This issue is to track that effort. 

Seeking feedback! Please post in the thread below with any feedback/concerns about the CMake min version bump to 3.16.

dbarker:
Fixes #3598 

## Changes
- upgrade cmake min version from 3.14 to 3.16
- test against 3.16 in CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.16)
 
 # See https://cmake.org/cmake/help/latest/policy/CMP0074.html required by
 # certain version of zlib which CURL depends on.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,22 +3,6 @@
 
 cmake_minimum_required(VERSION 3.16)
 
-# See https://cmake.org/cmake/help/latest/policy/CMP0074.html required by
-# certain version of zlib which CURL depends on.
-if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
-  cmake_policy(SET CMP0074 NEW)
-endif()
-
-# Allow to use normal variable for option()
-if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13")
-  cmake_policy(SET CMP0077 NEW)
-endif()
-
-# Prefer CMAKE_MSVC_RUNTIME_LIBRARY if possible
-if(POLICY CMP0091)
-  cmake_policy(SET CMP0091 NEW)
-endif()
-
 if(POLICY CMP0092)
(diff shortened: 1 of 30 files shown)


install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.16)
```
ThomsonTan, line 4: Is the root `CMakeLists.txt` referenced in these test? If no, probably extract all the versions into a shared location.

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.16)
```
marcalff, line 4: Please check if some cmake_poliy() calls can be removed.

In my understanding, CMP0074, CMP0077 and CMP0091 can be removed, they are set as NEW due to the minimum version.

For CMP0092, to investigate.

install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
dbarker: The root CMake lists file for the project isn't referenced in the install tests. Each of these tests is an independent CMake project and needs to have the minimum defined. We could define the minimum in a separate CMake script file and include it everywhere but I'm leaning towards keeping the minimum version set explicitly in each file for readability. Using find/replace all feels reasonable to bump the version. What do you think?

CMakeLists.txt
dbarker: Thanks. I've removed the policies that are the new default with 3.16. Leaving CMP0092 as is (set to OLD). This should be addressed in the future though - generally we shouldn't be setting old policies for too long.

install/test/cmake/component_tests/exporters_elasticsearch/CMakeLists.txt
ThomsonTan: Makes sense to leave the version requirement and use find/replace for version bump.

CMakeLists.txt
ThomsonTan: Yes, I think we still need CMP0092 for now.

marcalff (approved):
LGTM, thanks for the fix.

PR #3603, opened 2025-08-22
[resource_detectors] implementation of remaining process attributes

nikhilbhatia08:
Fixes # (issue)

## Changes
implements
 - `process.command_args`
 - `process.command_line`
`process.command_args` attribute should return a list with command and full list of arguments so changed the implementation of `process.command` as it should only get the first item that is command from that list.


resource_detectors/process_detector_utils.cc
```diff
@@ -68,41 +71,64 @@ std::string GetExecutablePath(const int32_t &pid)
...
+}
 
-  // Convert UTF-16 to UTF-8
-  int size_needed = WideCharToMultiByte(CP_UTF8, 0, wcmd, -1, NULL, 0, NULL, NULL);
-  if (size_needed <= 0)
+std::vector<std::string> GetCommandWithArgs(const int32_t &pid)
+{
+#ifdef _MSC_VER
+  int argc      = 0;
+  LPWSTR *argvW = CommandLineToArgvW(GetCommandLineW(), &argc);
```
owent, line 93: As I understand it, most Windows APIs map `FunctionName` to `<FunctionName>W` when `UNICODE` is defined, and to `<FunctionName>A` otherwise. Should we just call `GetCommandLine` and use `WideCharToMultiByte` only when `UNICODE` is defined?

nikhilbhatia08: Wrote `<FunctionName>W` because windows is UNICODE first but yeah can be also done using the above mentioned method but there can be an overhead of more conversions. but yeah will code it up in the above mentioned method. Thanks!

nikhilbhatia08: Have applied the changes, but in the same file `process_detector_utils.cc` I'm using `GetProcessImageFileNameW` which also should be changed ?

nikhilbhatia08: It seems that the `<FunctionName>A` variant isn’t implemented in the official WinAPI — only the W version exists. So in this case, I think we can go with `CommandLineToArgvW`

owent: Could you please try to remove `/D "_UNICODE" /D "UNICODE"` and add `/D "_MBCS"`?
image
Or set Advanced/Character Set/**Use Multi-Byte Character Set** (Do not select ***Use Unicode Character Set***).
In my environment, without `_UNICODE` and `UNICODE`. Both `<FunctionName>A` and `<FunctionName>W` are defined, but which `<FunctionName>` use depends `UNICODE`.
image
I think we can avoid the conversion here and reduce the CPU cost when using **Use Multi-Byte Character Set** here.

nikhilbhatia08: @owent still we could only support `CommandLineToArgvW(GetCommandLineW(), &argc)` because `CommandLineToArgvW` has complex parsing capabilities which would always return correct array of args.
 
But in case of `MBCS or NON_UNICODE` we have to use `GetCommandLineA` and then we would have to create a manual parser similar to `CommandLineToArgvW` or apply conversions to `GetCommandLineA` so that it can be used with `CommandLineToArgvW` to get the correct array of args which is complex .
There are certain security remarks as well for using `GetCommandLineA` - https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinea

I dont think it is worth the time to build a custom manual parser for parsing args from `GetCommandLineA` or apply conversions as all windows versions since Windows 2000 are `UNICODE` first. So for now I think we can only support for `UNICODE` using `CommandLineToArgvW(GetCommandLineW(), &argc)`.

resource_detectors/test/process_detector_test.cc
```diff
@@ -109,40 +120,79 @@ TEST(ProcessDetectorUtilsTest, GetExecutablePathTest)
...
+
+    LocalFree(argvW);
   }
   else
   {
+    command = std::string();
+  }
+#else
+  std::string command_line_path =
+      opentelemetry::resource_detector::detail::FormFilePath(pid, "cmdline");
```
owent, line 141: It's not issue of this PR, but `/proc/%d/%s` is not available on macOS.
Could you please disable this feature or support macOS/Darwin system in another PR?

nikhilbhatia08: Yeah sure actually Darwin was a todo and we can surely implement it for darwin as well. Will soon raise a pr for macos/Darwin. Thanks!

lalitb:
Do the specs say anything about scrubbing or masking args that may contain secrets/passwords? Right now we’d be storing and exporting in plain text. Probably worth checking what other language SDKs are doing here too.

nikhilbhatia08:
@lalitb otel-js and otel-go sdk have process resource detectors and they do not have a filter, but yeah we could implement one.

lalitb:
@nikhilbhatia08  see if java is doing something - https://github.com/search?q=org%3Aopen-telemetry+process.command+scrub&type=code .  Can we add a similar pass here? At least cover the basic cases like `--password/secret=<value>` or `--password/secret <value>`. I don’t think there’s a fool-proof way to mask everything, but handling the obvious scenarios would be good.

nikhilbhatia08:
Yeah we can a filter. Actually when thinking of this we can give the user the choice of what needs to be scrubbed and what not. For example `--PASS=some_pass_key  --SOME_ARG=some_value` then when the user is using the process detector then the user can pass a list of args they want to scrub, so the user for example provides `{"PASS"}` then we can give the args list as `{"--PASS=*****", "--SOME_ARG=some_value"}` so that the user has full control over which args they want to scrub and which they want to show. Does this idea look good @lalitb ? And Can this be in another PR?

lalitb:
> Yeah we can a filter. Actually when thinking of this we can give the user the choice of what needs to be scrubbed and what not. For example `--PASS=some_pass_key --SOME_ARG=some_value` then when the user is using the process detector then the user can pass a list of args they want to scrub, so the user for example provides `{"PASS"}` then we can give the args list as `{"--PASS=*****", "--SOME_ARG=some_value"}` so that the user has full control over which args they want to scrub and which they want to show. Does this idea look good @lalitb ? And Can this be in another PR?

Yes, configuring resource keys to be scrubbed as part of this detector ctor would be right fix, as separate PR.

nikhilbhatia08:
Ready To merge :)

resource_detectors/process_detector.cc
```diff
@@ -50,16 +51,21 @@ opentelemetry::sdk::resource::Resource ProcessResourceDetector::Detect() noexcep
...
-    std::string command = opentelemetry::resource_detector::detail::GetCommand(pid);
-    if (!command.empty())
+    std::vector<std::string> command_with_args =
+        opentelemetry::resource_detector::detail::GetCommandWithArgs(pid);
+    if (!command_with_args.empty())
     {
-      attributes[semconv::process::kProcessCommand] = std::move(command);
+      std::string commandline_args =
+          opentelemetry::resource_detector::detail::ConvertCommandArgsToString(command_with_args);
+      attributes[semconv::process::kProcessCommandLine] = std::move(commandline_args);
```
dbarker: Following up on https://github.com/open-telemetry/opentelemetry-cpp/pull/3603#issuecomment-3257155573. The spec calls out that [process-command-args](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-command-args) and [process-command-line](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-command-line) should not be collected by default unless sensitive data is sanitized: 

> SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data

Since we don't have that yet, let's not collect them until a sanitization feature is added. Please comment out setting these two attributes for now with a note.

dbarker: The spec on [process-command-line](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-command-line) says `Do not set this if you have to assemble it just for monitoring; use process.command_args instead`. It seems we are assembling it just for monitoring and since the information it adds is redundant with the process-command-args value I'd propose we don't assemble and set it. What do you think @nikhilbhatia08 and @lalitb?

nikhilbhatia08: Yeah can comment this out because the next thing to be implemented is a scrub to help with sanitization. Thanks!

nikhilbhatia08:
> Yeah we can a filter. Actually when thinking of this we can give the user the choice of what needs to be scrubbed and what not. For example `--PASS=some_pass_key --SOME_ARG=some_value` then when the user is using the process detector then the user can pass a list of args they want to scrub, so the user for example provides `{"PASS"}` then we can give the args list as `{"--PASS=*****", "--SOME_ARG=some_value"}` so that the user has full control over which args they want to scrub and which they want to show. Does this idea look good @lalitb ? And Can this be in another PR?

@dbarker what do you think of this idea as a process for sanitization ?

dbarker:
> Yeah we can a filter. Actually when thinking of this we can give the user the choice of what needs to be scrubbed and what not. For example --PASS=some_pass_key --SOME_ARG=some_value then when the user is using the process detector then the user can pass a list of args they want to scrub, so the user for example provides {"PASS"} then we can give the args list as {"--PASS=*****", "--SOME_ARG=some_value"} so that the user has full control over which args they want to scrub and which they want to show. Does this idea look good @lalitb ? And Can this be in another PR?
> 
> @dbarker what do you think of this idea as a process for sanitization ?

I like the idea of implementing it in a new PR and moving the design discussion there. How about starting with a proposal in a new issue? We can review the spec and other language SDKs to see if there is any guidance.

dbarker:
Thanks for the PR @nikhilbhatia08. I am requesting two changes: 
1) comment out setting the process-command-args attribute for now until there is a santization feature, 
2) remove setting the process-command-line attribute (and utility function) since the spec implies we should not assemble it just for monitoring and should use the process-command-args attribute instead.

resource_detectors/process_detector.cc
nikhilbhatia08: Done commenting out only the attributes, with a comment

dbarker (approved):
LGTM. Thanks!

Issue #3547, opened 2025-07-16
[SDK] View should not have a unit


marcalff:
Class View has a unit member.

This member seems unused.

Note that the SDK spec only mentions units in the instrument selection criteria, not in stream configuration.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#instrument-selection-criteria

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#stream-configuration

marcalff:
Related to #2236.

A unit was added:
- in the meter selector
- in the view

but it seem, looking at the spec, that only the meter selector needs unit.

PR #3606, opened 2025-08-22
[CONFIGURATION] File configuration - functional tests

marcalff:
Contributes to #2481

This is a partial fix, to implement functional tests for file configuration.

## Changes

* Implement functional tests for file configuration.
* Fix missing include in string_util.h, for isspace.
* Code cleanup in sdk_builder.cc, now that issue #3547 is resolved.

### DIFF
--- a/api/include/opentelemetry/common/string_util.h
+++ b/api/include/opentelemetry/common/string_util.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <ctype.h>
+
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/version.h"
 
--- a/functional/configuration/run_test.sh
+++ b/functional/configuration/run_test.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+# To run tests in a local dev environment:
+# - set BUILD_DIR to the top level build directory,
+
+[ -z "${BUILD_DIR}" ] && export BUILD_DIR="${HOME}/build"
+
+export EXAMPLE_BIN_DIR="${BUILD_DIR}/examples/configuration/"
+
+# Make sure `example_yaml` is in the path
+export PATH=${PATH}:${EXAMPLE_BIN_DIR}
+
+shelltest ./shelltests
+
--- a/functional/configuration/shelltests/empty.test
+++ b/functional/configuration/shelltests/empty.test
@@ -1,0 +1,8 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+$ example_yaml --test --yaml shelltests/empty.yaml
+>
+[ERROR] [Yaml Configuration Parser] Parse failed with exception: Yaml: not a map, looking for: file_format
+FAILED TO PARSE MODEL
+>= 1
--- a/functional/configuration/shelltests/disabled_empty.test
+++ b/functional/configuration/shelltests/disabled_empty.test
@@ -1,0 +1,8 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+$ example_yaml --test --yaml shelltests/disabled_empty.yaml
+>
+MODEL PARSED
+SDK CREATED
+>= 0
--- a/functional/configuration/shelltests/disabled_empty.yaml
+++ b/functional/configuration/shelltests/disabled_empty.yaml
@@ -1,0 +1,5 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+file_format: 1.0.0
+disabled:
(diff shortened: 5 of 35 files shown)


functional/configuration/shelltests/empty.test
```diff
@@ -0,0 +1,8 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+$ example_yaml --test --yaml shelltests/empty.yaml
+>
+[ERROR] [Yaml Configuration Parser] Parse failed with exception: Yaml: not a map, looking for: file_format
```
dbarker, line 6: Are all these errors c++ exceptions? Just thinking if we should be testing the configuration module's error reporting mechanism within c++ (in a gtest).

api/include/opentelemetry/common/string_util.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <ctype.h>
```
dbarker, line 6: Is this change intended? It doesn't look like anything else in the API has changed in this PR.

functional/configuration/run_test.sh
```diff
@@ -0,0 +1,19 @@
...
+# - set BUILD_DIR to the top level build directory,
+
+[ -z "${BUILD_DIR}" ] && export BUILD_DIR="${HOME}/build"
+
+export EXAMPLE_BIN_DIR="${BUILD_DIR}/examples/configuration/"
+
+# Make sure `example_yaml` is in the path
+export PATH=${PATH}:${EXAMPLE_BIN_DIR}
+
+shelltest ./shelltests
```
dbarker, line 18: Shelltest seems like a very reasonable approach, and it looks like a new test framework for the project. Can you share a bit on testing the process output this way vs testing these configuration cases with c++ checks in gtest? Just curious.

api/include/opentelemetry/common/string_util.h
marcalff: Yes, it is intended, it is for `isspace()`.

This was found by including string_util.h alone in some code during development.

include-what-you-use missed it, because we only run IWYU on *.cc files (the tool checks the headers and code then), but this file is header only.

functional/configuration/run_test.sh
marcalff: The principle of shelltest is very simple:
- it runs a command, in this case the example_yaml program with various arguments
- it captures stdout
- it captures stderr
- it captures the command return code
and then compares this to the expected stdout / stderr / return code from the *.test file.

Comparing to gtest, for a program that takes an input file as text (the yaml config), it is much more convenient to write a *.test file, compared to implement dedicated code in a gtest function.

gtest is good for unit tests, to call a few lines of code in a controlled way, and to mock functions.

shelltest is good for functional/integration tests, to invoke a full process, especially if the test application dumps results in the output, that can be just compared with the expected result.

functional/configuration/shelltests/empty.test
marcalff: Yes, these are exceptions.

The production code:
* throw an exception
* catch the exception
* prints it with OTEL_INTERNAL_LOG_ERROR()

The example code:
* installs an error handler
* captures the data written by OTEL_INTERNAL_LOG_ERROR()
* prints the error in stdout

So overall, a C++ exception thrown will be printed in the test stdout.

This already covers the error reporting.

shelltest:
* will notice unexpected exceptions thrown, seen as extra data in the test output
* will notice missing exceptions, seen as missing data from the test output
* will check the text of the exception

Testing more exceptions path means to add more yaml and test files, without writing any new test code.

dbarker: Thanks, the context helps. This seems reasonable for now. The assumption being that a user will not want/need to catch these exceptions in their code for configuration time error handling and that they can use the internal logger (create a custom internal logger handler) if needed.

marcalff: > The assumption being that a user will not want/need to catch these exceptions in their code for configuration time error handling and that they can use the internal logger (create a custom internal logger handler) if needed.

This may change later, but the principle will still be the same, with the test application doing the catch + print, for the same result, verified by shellcheck.

functional/configuration/run_test.sh
dbarker: Thanks. If the error handling mechanism for a user would be to catch an exception from the configuration module then I'd lean more towards implementing the tests in gtest (and creating fixtures for reading the yaml files and performing common checks) in c++. Since the error handling mechanism is through the internal logger and errors are sent to stdout/stderr checking from shelltest seems reasonable.

dbarker (approved):
Looks good. Thanks!

functional/configuration/run_test.sh
marcalff: For context, there are two main ways to verify test output (that I know of):

* use predicates, like EXPECT_EQ(actual, expected), like in gtest
  * this needs a lot of code to check each property
  * this is robust to noise (random numbers, timestamps)
* print the result and use diff, like in shelltest
  * this needs only code to print
  * this is not robust to noise, random numbers, timestamps etc need to be ignored when doing the diff

dbarker: Thanks for that context. I was just thinking of the added cognitive load for contributors if we add a new testing framework to the project, and if testing at the process interface (stderr/stdout) made sense. It makes sense in this PR.

PR #3609, opened 2025-08-25
[BUILD] Use -dev versions in main branch

### LINKED ISSUE #3583: Proposal: Update release process to bump main branch to a dev version number
After each release the version on the main branch remains at the last release version. This makes it difficult to determine if a dev version (api/sdk built from unreleased main branch) is being run or the last release. 

This issue is to propose a change to the release process/scripts to bump the main branch version to differentiate it from the prior release. 

One option is to bump the version on the main branch to the next release number (minor version bump) with a "-DEV" suffix.

marcalff:
Fixes #3583

## Changes


api/include/opentelemetry/version.h
```diff
@@ -10,9 +10,9 @@
 #  define OPENTELEMETRY_ABI_VERSION_NO 1
 #endif
 
-#define OPENTELEMETRY_VERSION "1.22.0"
+#define OPENTELEMETRY_VERSION "1.23.0"
```
dbarker: Should the API and SDK version strings get the "dev"suffix?

MODULE.bazel
```diff
@@ -3,7 +3,7 @@
 
 module(
     name = "opentelemetry-cpp",
-    version = "1.22.0",
+    version = "1.23.0",
```
ThomsonTan: Need `-dev` here?

api/test/core/version_test.cc
```diff
@@ -14,5 +14,7 @@ TEST(VersionTest, Consistency)
            OPENTELEMETRY_VERSION_MINOR, OPENTELEMETRY_VERSION_PATCH);
 
   std::string actual = OPENTELEMETRY_VERSION;
-  EXPECT_EQ(actual, expected);
+  /* OPENTELEMETRY_VERSION may contain a -dev suffix */
```
ThomsonTan, line 18: Check the exact suffix?

api/include/opentelemetry/version.h
marcalff: Fixed.

MODULE.bazel
marcalff: Fixed.

api/test/core/version_test.cc
marcalff: The `-dev` convention for main is arbitrary, we can potentially have also `-rc`, `-rfc` or anything, so the `-dev` part can not be hardcoded in the unit test itself.

There is no `#define OPENTELEMETRY_VERSION_SUFFIX "-dev"` either (and I don't think we should add it), so there is nothing to compare the suffix against to validate the full version string.

This code change in the unit test is sufficient to pass with both "1.23.0" and "1.23.0-dev".

dbarker (approved):
Looks good with one minor change - the version and suffix at the top of the main CMakeLists.txt file need to be updated

PR #3629, opened 2025-09-07
[CONFIGURATION] File configuration - semantic version

marcalff:
Contributes to #2481

This is a partial fix, to implement semantic version in the yaml parser.

## Changes

* Implement semantic version checks for the file format.
  * The yaml parser checks the field file_format for major and minor version.
  * Changed every parser code from a static function to a method. This is to allow parsing depending on the format version seen, if necessary.
  * Adjusted all tests accordingly.
* Fixed minor build issues seen in the integration PR
  * Changed `IntegerAttributeValueConfiguration` to use a `long long` instead of `std::size_t`, to better fit the attributes variant.
  * Fixed `setenv`, `unsetenv` in the unit tests for Windows.


sdk/include/opentelemetry/sdk/configuration/configuration_parser.h
```diff
@@ -18,7 +99,255 @@ namespace configuration
...
+
+  std::unique_ptr<AttributesConfiguration> ParseAttributesConfiguration(
+      const std::unique_ptr<DocumentNode> &node);
+
+  std::unique_ptr<ResourceConfiguration> ParseResourceConfiguration(
+      const std::unique_ptr<DocumentNode> &node);
+
+  std::unique_ptr<Configuration> Parse(std::unique_ptr<Document> doc);
+
+private:
```
dbarker, line 348: Is the version here associated with the instance of the Document being parsed? I'm wondering if these parse methods could remain static/stateless/hidden and instead we add the version to the document and get it from the instance passed in.

marcalff: Yes, this is the version found in the document, in file_format.

Currently, all the helpers only see a node, to parse it.

To carry the version in the document only, all the helpers need to change to pass around the full document, for code to be able to inspect a version number.

Also, in the long term, having every helper public is a good thing, because eventually the parsing code for plugins will want to parse only a fraction of the document, say to parse an exporter node from a custom processor, so this change goes into the right direction.

marcalff:
> Looks good. Just a minor nitpick on making the specialized ConfigurationParser::Parse* methods const if they don't modify the file version.

Good point, I will fix that.

Issue #3546, opened 2025-07-16
[SDK] MetricFilter needs a new AttributeProcessor to handle IncludeExclude


marcalff:
Metric filtering allows to include or exclude some attributes during metric collection.

The declarative configuration defines attribute_keys filtering with:
- an include list
- an exclude list

The opentelemetry-cpp SDK needs to provide a proper subclass of AttributeProcessor to implement this behavior.

Current AttributeProcessor subclasses like FilteringAttributesProcessor only implement partially the desired filtering.

nikhilbhatia08:
Hey marcalff! I would like to pick up this issue and try it. Thanks!

marcalff:
@nikhilbhatia08 

The relevant part from the spec given in link is:


> attribute_keys: This is, at a minimum, an allow-list of attribute keys for measurements captured in the metric stream. The allow-list contains attribute keys that identify the attributes that MUST be kept, and all other attributes MUST be ignored.
> 
> Implementations MAY accept additional attribute filtering functionality for this parameter.
> 
> Users can provide attribute_keys, but it is up to their discretion. Therefore, the stream configuration parameter needs to be structured to accept attribute_keys, but MUST NOT obligate a user to provide them. If the user does not provide any value, the SDK SHOULD use the [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-advisory-parameters) advisory parameter configured on the instrument instead. If the Attributes advisory parameter is absent, all attributes MUST be kept.
> 
> Additionally, implementations SHOULD support configuring an exclude-list of attribute keys. The exclude-list contains attribute keys that identify the attributes that MUST be excluded, all other attributes MUST be kept. If an attribute key is both included and excluded, the SDK MAY fail fast in accordance with initialization [error handling principles](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles).
> 


This is to be implemented in a sub class of `opentelemetry::sdk::metrics::AttributesProcessor`.

dbarker:
The SDK classes in `sdk/metrics/view/attributes_processor.h` still need to be used in the sdk builder to close out this issue. Help still wanted!

https://github.com/open-telemetry/opentelemetry-cpp/blob/7532fca6fe151c80c8481c444de5768f9f62f4f8/sdk/src/configuration/sdk_builder.cc#L1901

iamhaseebn:
[...]

PR #3633, opened 2025-09-08
[Metrics] New Attribute Processor for Exclude list

nikhilbhatia08:
Fixes # ([issue](https://github.com/open-telemetry/opentelemetry-cpp/issues/3546#issue-3237554269))

## Changes

Created a new Attribute processor for exclude list as filter only partially exists for allowed list

This is the doc of spec - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#stream-configuration

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
+++ b/sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
@@ -101,6 +101,49 @@ private:
   std::unordered_map<std::string, bool> allowed_attribute_keys_;
 };
 
+/**
+ * FilteringExcludeAttributeProcessor filters by exclude attribute list and drops names if they are
+ * present in the exclude list
+ */
+
+class FilteringExcludeAttributesProcessor : AttributesProcessor
+{
+public:
+  FilteringExcludeAttributesProcessor(std::unordered_map<std::string, bool> &&exclude_list = {})
+      : exclude_list_(std::move(exclude_list))
+  {}
+
+  FilteringExcludeAttributesProcessor(const std::unordered_map<std::string, bool> &exclude_list = {})
+      : exclude_list_(exclude_list)
+  {}
+
+  MetricAttributes process(
+      const opentelemetry::common::KeyValueIterable &attributes) const noexcept override
+  {
+    MetricAttributes result;
+    attributes.ForEachKeyValue(
+        [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
+          if (exclude_list_.find(key.data()) == exclude_list_.end())
+          {
+            result.SetAttribute(key, value);
+            return true;
+          }
+          return true;
+        });
+
+    result.UpdateHash();
+    return result;
+  }
+
+  bool isPresent(nostd::string_view key) const noexcept override
+  {
+    return (exclude_list_.find(key.data()) == exclude_list_.end());
+  }
+
+private:
+  std::unordered_map<std::string, bool> exclude_list_;
+};
+
 }  // namespace metrics
 }  // namespace sdk
 OPENTELEMETRY_END_NAMESPACE
--- a/sdk/test/metrics/attributes_processor_test.cc
+++ b/sdk/test/metrics/attributes_processor_test.cc
@@ -51,3 +51,40 @@ TEST(AttributesProcessor, FilteringAllAttributesProcessor)
   auto filtered_attributes = attributes_processor.process(iterable);
   EXPECT_EQ(filter.size(), kNumFilterAttributes);
 }
+
+TEST(AttributesProcessor, FilteringExcludeAttributesProcessor)
+{
+  const int kNumFilterAttributes               = 3;
+  std::unordered_map<std::string, bool> filter = {
+      {"attr2", true}, {"attr4", true}, {"attr6", true}};
+  const int kNumAttributes              = 6;
+  std::string keys[kNumAttributes]      = {"attr1", "attr2", "attr3", "attr4", "attr5", "attr6"};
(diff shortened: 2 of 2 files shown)


sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -101,6 +101,50 @@ class FilteringAttributesProcessor : public AttributesProcessor
...
+      : exclude_list_(exclude_list)
+  {}
+
+  MetricAttributes process(
+      const opentelemetry::common::KeyValueIterable &attributes) const noexcept override
+  {
+    MetricAttributes result;
+    attributes.ForEachKeyValue(
+        [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
+          if (exclude_list_.find(key.data()) == exclude_list_.end())
```
lalitb: `key.data()` isn’t safe since the view may not be null-terminated. Using `std::string(key)` fixes that, but there is also an allocation on every lookup. Because `process` runs on the hot path, the better long-term fix is to make the container accept `nostd::string_view` directly with a custom hash/equality - to avoid the string copy.

Not specific to this PR, since same pattern exists in `FilteringAttributesProcessor`. For now, I’d suggest updating both processors here to use `std::string(key)` for safety, and then handle the zero-copy optimization in a follow-up PR.

nikhilbhatia08: Thanks! will raise a PR with custom hash and equality.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -80,7 +80,7 @@ class FilteringAttributesProcessor : public AttributesProcessor
     MetricAttributes result;
     attributes.ForEachKeyValue(
         [&](nostd::string_view key, opentelemetry::common::AttributeValue value) noexcept {
-          if (allowed_attribute_keys_.find(key.data()) != allowed_attribute_keys_.end())
+          if (allowed_attribute_keys_.find(std::string(key)) != allowed_attribute_keys_.end())
```
ThomsonTan, line 83: Is creating a temporary string object needed here?

nikhilbhatia08: As mentioned in the above review by @lalitb key.data() isn’t safe since the view may not be null-terminated. Using std::string(key) fixes that, but there is also an allocation on every lookup. Because process runs on the hot path, the better long-term fix is to make the container accept nostd::string_view directly with a custom hash/equality - to avoid the string copy.

Not specific to this PR, since same pattern exists in FilteringAttributesProcessor. For now, I’d suggest updating both processors here to use std::string(key) for safety, and then handle the zero-copy optimization in a follow-up PR.

And however in the next pr this would be solved

ThomsonTan: Ok, could you please file a new issue on this?
> make the container accept nostd::string_view directly with a custom hash/equality - to avoid the string copy

nikhilbhatia08: Already filed - https://github.com/open-telemetry/opentelemetry-cpp/issues/3635

sdk/test/metrics/attributes_processor_test.cc
```diff
@@ -51,3 +51,40 @@ TEST(AttributesProcessor, FilteringAllAttributesProcessor)
...
+  const int kNumAttributes                     = 6;
+  std::string keys[kNumAttributes]      = {"attr1", "attr2", "attr3", "attr4", "attr5", "attr6"};
+  int values[kNumAttributes]            = {10, 20, 30, 40, 50, 60};
+  std::map<std::string, int> attributes = {{keys[0], values[0]}, {keys[1], values[1]},
+                                           {keys[2], values[2]}, {keys[3], values[3]},
+                                           {keys[4], values[4]}, {keys[5], values[5]}};
+  FilteringExcludeAttributesProcessor attributes_processor(filter);
+  opentelemetry::common::KeyValueIterableView<std::map<std::string, int>> iterable(attributes);
+  auto filtered_attributes = attributes_processor.process(iterable);
+  EXPECT_EQ(filter.size(), kNumFilterAttributes);
```
marcalff: This test that the filter list is empty, which says nothing about filtered_attributes.

Check the filtered_attributes.size(), to make sure nothing was removed ?

sdk/test/metrics/attributes_processor_test.cc
```diff
@@ -51,3 +51,40 @@ TEST(AttributesProcessor, FilteringAllAttributesProcessor)
...
+                                           {keys[2], values[2]}, {keys[3], values[3]},
+                                           {keys[4], values[4]}, {keys[5], values[5]}};
+  FilteringExcludeAttributesProcessor attributes_processor(filter);
+  opentelemetry::common::KeyValueIterableView<std::map<std::string, int>> iterable(attributes);
+  auto filtered_attributes = attributes_processor.process(iterable);
+  for (auto &e : filtered_attributes)
+  {
+    EXPECT_TRUE(filter.find(e.first) == filter.end());
+  }
+  EXPECT_EQ(filter.size(), kNumFilterAttributes);
```
marcalff: There are 6 items initially, 3 to remove and 3 to keep.

This test checks there are 3 left, which does not prove if the proper items are kept / removed.

Add a 7th item, so we can see if the result is size 3 or 4.

nikhilbhatia08: Yeah added the other attribute and there will be 4 after filtering

PR #3643, opened 2025-09-16
[SDK] custom hash and equality for attribute processor

### LINKED ISSUE #3635: [Metrics] Attributes Processor needs custom hash and equality
currently the attributes processor `nostd::string_view` and access the value with `key.data()` but this may not be null terminated. So we need a custom hash and equality for the view container as discussed in this pr - https://github.com/open-telemetry/opentelemetry-cpp/pull/3633#discussion_r2331073832

nikhilbhatia08:
Fixes #3635

nostd::string_view and access the value with key.data() but this may not be null terminated. So we need a custom hash and equality for the view container

## Changes

### DIFF
--- a/sdk/test/metrics/attributes_processor_benchmark.cc
+++ b/sdk/test/metrics/attributes_processor_benchmark.cc
@@ -4,11 +4,9 @@
 #include <benchmark/benchmark.h>
 #include <map>
 #include <string>
-#include <unordered_map>
 #include <utility>
 
 #include "opentelemetry/common/key_value_iterable_view.h"
-#include "opentelemetry/sdk/common/custom_hash_equality.h"
 #include "opentelemetry/sdk/metrics/view/attributes_processor.h"
 
 using namespace opentelemetry::sdk::metrics;
--- a/sdk/test/metrics/sum_aggregation_test.cc
+++ b/sdk/test/metrics/sum_aggregation_test.cc
@@ -19,7 +19,6 @@
 #include "opentelemetry/nostd/shared_ptr.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/nostd/variant.h"
-#include "opentelemetry/sdk/common/custom_hash_equality.h"
 #include "opentelemetry/sdk/instrumentationscope/instrumentation_scope.h"
 #include "opentelemetry/sdk/metrics/aggregation/aggregation_config.h"
 #include "opentelemetry/sdk/metrics/data/metric_data.h"


nikhilbhatia08:
`std::unordered_map` does not support heterogeneous lookup `is_transparent` before cpp20, even with custom hash/equality older standard libraries (especially libc++ on macOS and pre cpp20) do not accept string_view keys without materializing a std::string. Any ideas on how to make this work with even before versions of cpp20 ?

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -138,11 +227,11 @@ class FilteringExcludeAttributesProcessor : public AttributesProcessor
 
   bool isPresent(nostd::string_view key) const noexcept override
   {
-    return (exclude_list_.find(std::string(key)) == exclude_list_.end());
+    return (find_hetero(exclude_list_, key) == exclude_list_.end());
   }
 
 private:
-  std::unordered_map<std::string, bool> exclude_list_;
+  std::unordered_map<std::string, bool, StringViewHash, StringViewEqual> exclude_list_;
```
nikhilbhatia08: Can we have a wrapper class so that we don't need to everytime write the whole `std::unordered_map<std::string, bool, StringViewHash, StringViewEqual>` instead we can have a `FilterMap` ?

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,91 @@ namespace sdk
...
+#endif
+
+  std::size_t operator()(const std::string &s) const noexcept
+  {
+    return std::hash<std::string>{}(s);
+  }
+
+  std::size_t operator()(opentelemetry::nostd::string_view sv) const noexcept
+  {
+#if __cplusplus >= 202002L
```
owent: Please using `OPENTELEMETRY_STL_VERSION` instead of `__cplusplus`, and C++ 17 has string_view.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,91 @@ namespace sdk
...
+    return std::hash<opentelemetry::nostd::string_view>{}(
+        opentelemetry::nostd::string_view{sv.data(), sv.size()});
+#else
+    // pre-C++20 fallback: materialize to std::string
+    return std::hash<std::string>{}(std::string{sv.data(), sv.size()});
```
owent: Can we implement a bundle version of hash algorithm for fallback?

In my understanding, std::hash use fnv1-a and here is a example:

```cpp
template <typename Ty, size_t s>
struct fnv_magic_prime_number {
  static constexpr const Ty value = 0x01000193U;
};

template <typename Ty>
struct fnv_magic_prime_number<Ty, 8> {
  static constexpr const Ty value = 0x100000001b3ULL;
};

template <typename Ty, size_t s>
struct fnv_magic_offset_basis {
  static constexpr const Ty value = 0x811C9DC5U;

  static constexpr Ty fix(Ty hval) { return hval; }
};

template <typename Ty>
struct fnv_magic_offset_basis<Ty, 8> {
  static constexpr const Ty value = 0xCBF29CE484222325ULL;

  static constexpr Ty fix(Ty hval) { return hval ^ (hval >> 32); }
};

template <typename Ty>
Ty fnv_n_buf(const void *buf, size_t len,
                                             Ty hval = fnv_magic_offset_basis<Ty, sizeof(Ty)>::value) {
  const unsigned char *bp = reinterpret_cast<const unsigned char *>(buf);
  const unsigned char *be = bp + len;
  Ty mn = fnv_magic_prime_number<Ty, sizeof(Ty)>::value;

  while (bp < be) {
    hval *= mn;
    hval ^= (Ty)*bp++;
  }

  return fnv_magic_offset_basis<Ty, sizeof(Ty)>::fix(hval);
}

template <typename Ty>
Ty fnv_n_buf_a(const void *buf, size_t len,
                                               Ty hval = fnv_magic_offset_basis<Ty, sizeof(Ty)>::value) {
  const unsigned char *bp = reinterpret_cast<const unsigned char *>(buf);
  const unsigned char *be = bp + len;
  Ty mn = fnv_magic_prime_number<Ty, sizeof(Ty)>::value;

  while (bp < be) {
    hval ^= (Ty)*bp++;
    hval *= mn;
  }

  return fnv_magic_offset_basis<Ty, sizeof(Ty)>::fix(hval);
}
```

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,91 @@ namespace sdk
...
+  {
+    return lhs == rhs;
+  }
+
+  bool operator()(const std::string &lhs, opentelemetry::nostd::string_view rhs) const noexcept
+  {
+    return lhs.size() == rhs.size() && std::memcmp(lhs.data(), rhs.data(), lhs.size()) == 0;
+  }
+
+  bool operator()(opentelemetry::nostd::string_view lhs, const std::string &rhs) const noexcept
```
owent: Is it better and simpler to use template here, and convert everything to string_view?

owent (changes requested):

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
nikhilbhatia08: Yes @owent, I was thinking about the same

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,133 @@ namespace sdk
...
+template <typename Ty>
+struct fnv_magic_offset_basis<Ty, 8>
+{
+  static constexpr const Ty value = 0xCBF29CE484222325ULL;
+
+  static constexpr Ty fix(Ty hval) { return hval ^ (hval >> 32); }
+};
+
+template <typename Ty>
+Ty fnv_n_buf(const void *buf, size_t len, Ty hval = fnv_magic_offset_basis<Ty, sizeof(Ty)>::value)
```
owent: Sorry, but the example I provided before is just a part of our production. We should change the code style to match otel-cpp here.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,133 @@ namespace sdk
...
+#if defined(OPENTELEMETRY_STL_VERSION)
+#  if OPENTELEMETRY_STL_VERSION >= 2020
+  using is_transparent = void;
+#  endif
+#endif
+
+  template <typename Lhs, typename Rhs>
+  bool operator()(const Lhs &lhs, const Rhs &rhs) const noexcept
+  {
+    opentelemetry::nostd::string_view lsv(lhs.data(), lhs.size());
```
owent: Do you think we can use `opentelemetry::nostd::string_view lsv(opentelemetry::nostd::data(lhs), opentelemetry::nostd::size(lhs));` so we can also support char array?

nikhilbhatia08: Yes @owent

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
nikhilbhatia08: Yeah this should not be exposed so this can be kept in a detail namespace ? How does this idea look @owent ?

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -18,8 +19,124 @@ namespace sdk
 {
 namespace metrics
 {
+
 using MetricAttributes = opentelemetry::sdk::metrics::FilteredOrderedAttributeMap;
 
+/**
+ * Generic FNV hash implementation
```
marcalff: The code for this new hash function should be in a separate header file.

Also, some pointers to a paper explaining the magic behind it would be welcome.

Right now the code just contains magic values, and can not be maintained.

marcalff (changes requested):
Please document where the new hash function comes from.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
nikhilbhatia08: @marcalff done with the changes, Thanks!

sdk/include/opentelemetry/sdk/common/custom_hash_equality.h
```diff
@@ -0,0 +1,157 @@
...
+};
+
+/**
+ * Cross-platform heterogeneous lookup wrapper.
+ * Falls back to std::string construction on libc++ (macOS) and pre-c++20,
+ * but uses direct lookup on libstdc++ (Linux).
+ */
+
+template <typename MapType>
+inline auto find_hetero(MapType &&map, opentelemetry::nostd::string_view key)
```
marcalff: Rename to find_heterogeneous.

Check the meaning of the previous name in English for an explanation.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -65,12 +68,19 @@ class DefaultAttributesProcessor : public AttributesProcessor
 class FilteringAttributesProcessor : public AttributesProcessor
 {
 public:
-  FilteringAttributesProcessor(std::unordered_map<std::string, bool> &&allowed_attribute_keys = {})
+  FilteringAttributesProcessor(
+      std::unordered_map<std::string,
+                         bool,
+                         opentelemetry::sdk::common::StringViewHash,
+                         opentelemetry::sdk::common::StringViewEqual> &&allowed_attribute_keys = {})
```
marcalff: Define a typedef (say FilteredNameMap) for:

 std::unordered_map<std::string,
                         bool,
                         opentelemetry::sdk::common::StringViewHash,
                         opentelemetry::sdk::common::StringViewEqual>

and use the typedef instead everywhere.

sdk/include/opentelemetry/sdk/common/custom_hash_equality.h
nikhilbhatia08: @marcalff Sorry! found out the meaning and yes it should be find_heterogenous

PR #3644, opened 2025-09-17
New circular buffer using faa instead of cas to optimize performance 

ZENOTME:
solve #3645

## Changes

I noticed that original circular buffer using by batch span processor use lock free MPSC queu based on CAS. When experiencing intense multithreading contention, Compare-And-Swap (CAS) exhibits poorer scalability compared to Fetch-And-Add (FAA). In reference of https://github.com/dbittman/waitfree-mpsc-queue, this pr attempt to implement a wait-free MPSC (Multiple Producer, Single Consumer) queue using FAA. Base on original benchmark, it indicate that this approach demonstrates better performance scalability.

```
Run on (48 X 2593.99 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 256 KiB (x24)
  L3 Unified 30720 KiB (x2)
Load Average: 7.85, 5.70, 4.48
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
BM_BaselineBuffer/1     10178537 ns        51528 ns         1000
BM_BaselineBuffer/2      7408646 ns        69828 ns         1000
BM_BaselineBuffer/4      7684772 ns       127549 ns         1000
BM_BaselineBuffer/8      7222459 ns       278660 ns         1000
BM_BaselineBuffer/16     6716972 ns       603712 ns         1215
BM_LockFreeBuffer/1      3915343 ns        53125 ns         1000
BM_LockFreeBuffer/2      4798406 ns        70581 ns         1000
BM_LockFreeBuffer/4      4562709 ns       128493 ns         1000
BM_LockFreeBuffer/8      4935221 ns       289996 ns         1000
BM_LockFreeBuffer/16     5187913 ns       618856 ns         1081
BM_OptimizedBuffer/1     4256507 ns        49970 ns         1000
BM_OptimizedBuffer/2     3398719 ns        67712 ns         1000
BM_OptimizedBuffer/4     3204749 ns       127378 ns         1000
BM_OptimizedBuffer/8     3230722 ns       296507 ns         1000
BM_OptimizedBuffer/16    3859005 ns       769220 ns         1000
```

I would like to refine this PR if it's acceptable.

### DIFF
--- a/sdk/test/common/optmize_circular_buffer.h
+++ b/sdk/test/common/optmize_circular_buffer.h
@@ -14,6 +14,7 @@
 #include "opentelemetry/sdk/common/atomic_unique_ptr.h"
 #include "opentelemetry/sdk/common/circular_buffer_range.h"
 #include "opentelemetry/version.h"
+#include <iostream>
 
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace sdk
@@ -21,14 +22,14 @@ namespace sdk
 namespace common
 {
 /*
- * A lock-free circular buffer that supports multiple concurrent producers
- * and a single consumer.
+ * An optimized lock-free circular buffer that supports multiple concurrent producers
+ * and a single consumer. Uses FAA (Fetch-And-Add) for better performance.
  */
 template <class T>
-class CircularBuffer
+class OptimizedCircularBuffer
 {
 public:
-  explicit CircularBuffer(size_t max_size)
+  explicit OptimizedCircularBuffer(size_t max_size)
       : data_{new AtomicUniquePtr<T>[max_size + 1]}, capacity_{max_size + 1}
   {}
 
@@ -39,7 +40,7 @@ public:
    */
   CircularBufferRange<const AtomicUniquePtr<T>> Peek() const noexcept
   {
-    return const_cast<CircularBuffer *>(this)->PeekImpl();
+    return const_cast<OptimizedCircularBuffer *>(this)->PeekImpl();
   }
 
   /**
@@ -55,13 +56,15 @@ public:
   template <class Callback>
   void Consume(size_t n, Callback callback) noexcept
   {
-    assert(n <= static_cast<size_t>(head_ - tail_));
+    // comsume max(n, available) elements
     auto range = PeekImpl().Take(n);
     static_assert(noexcept(callback(range)), "callback not allowed to throw");
-    tail_ += n;
-    // why not call back after tail_ += n
+    // consume elements
     callback(range);
-    // tail_ += n;
+    // free elements to let producers add new elements
+    tail_ += range.size();
+    tail_ %= capacity_;
+    count_.fetch_sub(range.size(), std::memory_order_release);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


sdk/test/common/optmize_circular_buffer.h
```diff
@@ -0,0 +1,213 @@
...
+private:
+  std::unique_ptr<AtomicUniquePtr<T>[]> data_;
+  size_t capacity_;
+  std::atomic<uint64_t> count_{0};  
+  std::atomic<uint64_t> head_{0}; 
+  uint64_t tail_{0};
+
+  CircularBufferRange<AtomicUniquePtr<T>> PeekImpl() noexcept
+  {
+    uint64_t current_count = count_.load(std::memory_order_relaxed);
```
owent, line 168: The only sync operation is processed here. Is it possible return the same CircularBufferRange for more than one thread, and cause data race?

ZENOTME: I'm sorry I can't get the point here, could you elaborate a more specific scene.

owent: When two threads call this function and finish the callings before any change, they may returns the same values. And `callback(range);` in `Consume` will be called with the same range.

ZENOTME: Yes, this circular buffer implementation is mpsc and assuming only one consumer thread calling this method. That's how it used in batch span processor.

Reneg973: Looks like `PeekImpl()` could be made const, so the const_cast could be removed too

Reneg973:
On Mac M1 I get:

```
Run on (10 X 24 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB
  L1 Instruction 128 KiB
  L2 Unified 4096 KiB (x10)
Load Average: 9.93, 6.67, 5.87
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
BM_BaselineBuffer/1      2789978 ns        32254 ns         1000
BM_BaselineBuffer/2      2949209 ns        46430 ns         1000
BM_BaselineBuffer/4      1956026 ns        72750 ns         9651
BM_BaselineBuffer/8      1690755 ns       117763 ns         5736
BM_BaselineBuffer/16     1491443 ns       212598 ns         3275
BM_LockFreeBuffer/1      1223514 ns        32825 ns        10000
BM_LockFreeBuffer/2      1369947 ns        45685 ns        10000
BM_LockFreeBuffer/4      2768243 ns        76035 ns         1000
BM_LockFreeBuffer/8      3852933 ns       135797 ns         1000
BM_LockFreeBuffer/16     4341987 ns       259862 ns         1000
BM_OptimizedBuffer/1     1471849 ns        32878 ns        10000
BM_OptimizedBuffer/2     1138345 ns        45200 ns        10000
BM_OptimizedBuffer/4     1536488 ns        73745 ns         9610
BM_OptimizedBuffer/8     1577325 ns       126947 ns         5565
BM_OptimizedBuffer/16    1833288 ns       237527 ns         2986
```

So the result is similar to SpinLockMutex: the higher the contention, the less efficient atomics are, and the better std::mutex performs.

Reneg973:
And a compiler fix for libc (MacOS) is also required, because `size_t` is of type `unsigned long` here:
```cpp
   uint64_t max_check = std::min<uint64_t>(current_count, capacity_);
```

marcalff:
Thanks for the contribution.

This PR is stale, draft, fails CI, and has merge conflicts, so now closing it.

The code change itself probably can be salvaged and evaluated, however it is provided under test instead of in the SDK production code, so some rework is needed to pass the PoC stage and drive this change all the way.

Not claiming whether FAA is better or worse compared to CAS, it is just that the PR itself is not ready for merge at this point.

PR #3649, opened 2025-09-21
[CMAKE] Fix CMake 4.x build error on MacOS runner and when building opentracing 

### LINKED ISSUE #3648: [CI] Build on MacOS broken due to CMake upgrade
Compiling opentracing-cpp now fails, because it requires a very old CMake version.

CMake was upgraded to 4.1.1 with:

* https://github.com/actions/runner-images/issues/12934

which no longer supports CMake < 3.5, causing a break.

dbarker:
Fixes #3648 

CMake 4.x removed compatibility with versions older than 3.5. 
https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features

OpenTracing is no longer being updated and sets the minimum CMake version required to 3.1. This creates a CMake configure error when using CMake 4.x. 

Older versions of many dependencies depend on minimum versions of CMake < 3.5 (those in the `install/cmake/third_party_<minimum,stable>` and `install/conan/conanfile_stable.txt`). 

## Changes

- Set `CMAKE_POLICY_VERSION_MINIMUM=3.5` when building opentracing-cpp from source using FetchContent
- Install CMake 3.31 in the MacOS CMake install test with `conanfile_stable.txt` dependency versions. 

### DIFF
--- a/.github/workflows/cmake_install.yml
+++ b/.github/workflows/cmake_install.yml
@@ -204,8 +204,9 @@ jobs:
           pip install "conan==2.15.1"
           conan profile detect --force
       - name: Install or build all dependencies with Conan
+        # CMAKE_POLICY_VERSION_MINIMUM: '3.5' is required to build opentracing with CMake 4.x
         run: |
-          conan install install/conan/conanfile_stable.txt --build=missing -of /home/runner/conan -s build_type=${BUILD_TYPE} -s compiler.cppstd=${CXX_STANDARD}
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake_install.yml
```diff
@@ -262,6 +262,7 @@ jobs:
       INSTALL_TEST_DIR: '/Users/runner/install_test'
       CXX_STANDARD: '17'
       CMAKE_TOOLCHAIN_FILE: '/Users/runner/conan/build/Debug/generators/conan_toolchain.cmake'
+      CMAKE_VERSION: 3.31.0 # building the conan "stable" versions requires CMake 3.x
```
dbarker, line 265: Downgrading is the most direct and stable solution to fix the CMake 4.x build issue. Many of the dependencies in this "stable" version set depend on CMake versions < 3.5. Later versions of the dependencies (with the exception of opentracing has a minimum of 3.1) build with CMake 4.x.

marcalff (approved):

PR #3655, opened 2025-09-23
[CONFIGURATION] File configuration - cmake build

marcalff:
Contributes to #2481

This is a partial fix, to implement the CMake build.

## Changes

* Implement the CMake build.

### DIFF
--- a/api/include/opentelemetry/plugin/detail/dynamic_load_windows.h
+++ b/api/include/opentelemetry/plugin/detail/dynamic_load_windows.h
@@ -12,10 +12,10 @@
 #include "opentelemetry/plugin/hook.h"
 #include "opentelemetry/version.h"
 
-#include <Windows.h>
+#include <windows.h>
 
-#include <WinBase.h>
 #include <errhandlingapi.h>
+#include <winbase.h>
 
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace plugin
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_properties.h
@@ -13,7 +13,7 @@
 #include <vector>
 
 #ifdef _WIN32
-#  include <Windows.h>
+#  include <windows.h>
 #endif
 
 OPENTELEMETRY_BEGIN_NAMESPACE
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_provider.h
@@ -42,7 +42,7 @@
 #include <vector>
 
 #ifdef HAVE_KRABS_TESTS
-// krabs.hpp requires this definition of min macro from Windows.h
+// krabs.hpp requires this definition of min macro from windows.h
 #  ifndef min
 #    define min(a, b) (((a) < (b)) ? (a) : (b))
 #  endif
--- a/exporters/etw/include/opentelemetry/exporters/etw/etw_random_id_generator.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/etw_random_id_generator.h
@@ -6,7 +6,7 @@
 #include "opentelemetry/version.h"
 
 #ifdef _WIN32
-#  include "Windows.h"
+#  include "windows.h"
 #endif
 
 OPENTELEMETRY_BEGIN_NAMESPACE
--- a/exporters/etw/include/opentelemetry/exporters/etw/utils.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/utils.h
@@ -17,12 +17,12 @@
 #include "opentelemetry/version.h"
 
 #ifdef _WIN32
-#  include <Windows.h>
 #  include <evntprov.h>
 #  include <wincrypt.h>
+#  include <windows.h>
 #  pragma comment(lib, "Advapi32.lib")
 #  pragma comment(lib, "Rpcrt4.lib")
-#  include <Objbase.h>
+#  include <objbase.h>
 #  pragma comment(lib, "Ole32.Lib")
 #else
 #  include <codecvt>
--- a/exporters/etw/include/opentelemetry/exporters/etw/uuid.h
+++ b/exporters/etw/include/opentelemetry/exporters/etw/uuid.h
@@ -13,7 +13,7 @@
 #include <vector>
 
 #ifdef _WIN32
-#  include "Windows.h"
+#  include "windows.h"
 #endif
 
 OPENTELEMETRY_BEGIN_NAMESPACE
--- a/exporters/otlp/src/otlp_file_client.cc
+++ b/exporters/otlp/src/otlp_file_client.cc
@@ -43,9 +43,9 @@
 #    define NOMINMAX
 #  endif
 
-#  include <Windows.h>
 #  include <direct.h>
 #  include <io.h>
+#  include <windows.h>
 
 #  ifdef UNICODE
 #    include <atlconv.h>
--- a/ext/include/opentelemetry/ext/http/server/socket_tools.h
+++ b/ext/include/opentelemetry/ext/http/server/socket_tools.h
@@ -16,7 +16,7 @@
 
 #ifdef _WIN32
 
-// #  include <Windows.h>
+// #  include <windows.h>
 
 #  include <winsock2.h>
 
--- a/sdk/src/common/fast_random_number_generator.h
+++ b/sdk/src/common/fast_random_number_generator.h
@@ -25,7 +25,7 @@ namespace common
  *
  * ...
  * #define NOMINMAX
- * #include <Windows.h>
+ * #include <windows.h>
  * ...
  *
  * See:


exporters/ostream/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+#
+# opentelemetry_exporter_ostream_span
```
ThomsonTan, line 5: This comment is unnecessary?

ci/do_ci.sh
```diff
@@ -171,6 +171,31 @@ elif [[ "$1" == "cmake.maintainer.abiv2.test" ]]; then
   eval "$MAKE_COMMAND"
   make test
   exit 0
+elif [[ "$1" == "cmake.maintainer.yaml.test" ]]; then
```
dbarker, line 174: What do you think about adding the WITH_CONFIGURATION option to the CMake cache file then it can be tested with more jobs/compilers/platforms?

We can add it to the components options or preview options. 
test_common/cmake/component-options.cmake

ci/install_ryml.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -ex
```
dbarker: I'd like to move installing ryml into the third party install CMake project (install/cmake/CMakeLists.txt) so we can install it on all platforms and use the git tag files ( third_party_release and the install/cmake/third_party_*) to manage the versions tested.  That would remove this script. Does that sound reasonable?

CMakeLists.txt
```diff
@@ -349,6 +351,18 @@ if((NOT WITH_API_ONLY) AND USE_NLOHMANN_JSON)
   include("${opentelemetry-cpp_SOURCE_DIR}/cmake/nlohmann-json.cmake")
 endif()
 
+#
+# Do we need RapidYaml ?
+#
+
+if((NOT WITH_API_ONLY) AND WITH_CONFIGURATION)
+  find_package(ryml 0.9.0 REQUIRED)
```
dbarker: Either in this PR or a follow-up we should move this to an cmake/ryml.cmake script and add FetchContent support like the other dependencies. I can help.

examples/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+if(WITH_PROMETHEUS)
+  include_directories(${CMAKE_SOURCE_DIR}/exporters/prometheus/include)
+  add_definitions(-DOTEL_HAVE_PROMETHEUS)
+  target_link_libraries(example_yaml opentelemetry_exporter_prometheus_builder)
+endif()
+
+if(DEFINED OPENTELEMETRY_BUILD_DLL)
+  target_link_libraries(example_yaml opentelemetry_cpp)
+else()
+  target_link_libraries(
```
dbarker, line 77: Please use the public CMake alias targets in the examples (ie opentelemetry-cpp::configuration) and set the scope with PRIVATE, PUBLIC, or INTERFACE.  Using the public targets will allow us to test the example against an installation of otel-cpp.

sdk/src/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+set_target_version(opentelemetry_configuration)
+
+target_include_directories(
+  opentelemetry_configuration
+  PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/sdk/include>")
+
+target_link_libraries(opentelemetry_configuration PUBLIC opentelemetry_api
+                                                         opentelemetry_common)
+
+# target_link_libraries( opentelemetry_configuration PUBLIC yaml-cpp::yaml-cpp)
```
dbarker: nit - can remove

sdk/src/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+target_link_libraries(opentelemetry_configuration PUBLIC ryml::ryml)
```
dbarker: Any chance the ryml dependency can be private to this library?

sdk/src/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+if(OPENTELEMETRY_INSTALL)
```
dbarker, line 30: This install section can be removed and handled by adding the configuration target to the SDK component at 

https://github.com/open-telemetry/opentelemetry-cpp/blob/116f4388a13a30e5fd0fed4c9583a813ab5b300f/sdk/CMakeLists.txt#L17

dbarker:
Thanks for the PR. Excited to get this building :). I left some comments  and mainly interested in getting the ryml dependency managed like the others (git tag files to define the version, FetchContent support, and install with the common third party CMake project instead of the bash script). Seems reasonable to follow up with these before the next release if it's too much change in this PR.

ci/install_ryml.sh
marcalff: Implemented, but needs review, not familiar with this part.

The only items preventing to remove install_ryml.sh at this point are the clang-tidy and iwyu build.

sdk/src/configuration/CMakeLists.txt
marcalff: Thanks, done.

CMakeLists.txt
marcalff: Done

sdk/src/configuration/CMakeLists.txt
marcalff: Fixed.

sdk/src/configuration/CMakeLists.txt
marcalff: Fixed.

ci/install_ryml.sh
marcalff: Now fixed.

ci/do_ci.sh
marcalff: I don't want to couple WITH_OTLP_HTTP and WITH_CONFIGURATION, we need to make sure every existing feature still builds without the yaml dependency.

Fixed by using the preview options.

examples/configuration/CMakeLists.txt
marcalff: Thanks, fixed.

Note that there are no CMake installation tests, this should be another PR (once I figure out how to do it).

.github/workflows/clang-tidy.yaml
```diff
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 61
+            warning_limit: 63
```
marcalff, line 20: There are 2 new warnings in the entire configuration code base.

To fix separately, this PR is about makefiles.

examples/configuration/CMakeLists.txt
dbarker: Thanks. Sounds good. The simple logs example is a good example :) 
https://github.com/open-telemetry/opentelemetry-cpp/blob/5540dae676e61700246183dc74ec74ad5cd6256e/examples/logs_simple/CMakeLists.txt#L20

ci/do_ci.sh
dbarker: Preview options sounds good. We run those with all platforms for cmake install tests, iwyu, and clang-tidy!

cmake/ryml.cmake
```diff
@@ -0,0 +1,44 @@
...
+# Import the rapidyaml target (ryml::ryml).
+# 1. Find an installed ryml package
+# 2. Use FetchContent to build ryml from a git submodule
+# 3. Use FetchContent to fetch and build ryml from GitHub
+
+# Set the FIND_ryml_VERSION variable from the git tag.
+string(REGEX REPLACE "^v([0-9]+\\.[0-9]+\\.[0-9]+)$" "\\1" FIND_ryml_VERSION "${ryml_GIT_TAG}")
+
+# TODO: use OTEL_ryml_SEARCH_MODE ?
+find_package(ryml ${FIND_ryml_VERSION} CONFIG QUIET)
```
dbarker: Looking at the ryml example for find_package they show just using the default search mode (not CONFIG). 
https://github.com/biojppm/rapidyaml/blob/master/samples/find_package/CMakeLists.txt

Following that example, let's remove CONFIG here. Removing this will also match the search mode set by `set(OTEL_ryml_SEARCH_MODE "")` in thirdparty-dependency-config.cmake. The TODO can be removed.  

We haven't been setting the version in find_package for other dependencies yet. This may be something to look into. It wall cause the cmake configuration to fail if a compatible version was not found. It seems to be up to each package how they support this check.

cmake/ryml.cmake
```diff
@@ -0,0 +1,44 @@
...
+set(ryml_PROVIDER "find_package")
+
+if(NOT ryml_FOUND)
+  set(_RYML_SUBMODULE_DIR "${opentelemetry-cpp_SOURCE_DIR}/third_party/ryml")
```
dbarker: since we aren't yet adding ryml as git submodule we can remove the `fetch_source` option and just provide the `fetch_repository` option.

exporters/otlp/CMakeLists.txt
```diff
@@ -174,6 +280,30 @@ if(WITH_OTLP_HTTP)
...
+
+  add_library(
+    opentelemetry_exporter_otlp_http_builder
+    ${OPENTELEMETRY_OTLP_TARGETS_LIB_TYPE} src/otlp_http_span_builder.cc)
+
+  set_target_properties(opentelemetry_exporter_otlp_http_builder
+                        PROPERTIES EXPORT_NAME otlp_http_exporter_builder)
+  set_target_version(opentelemetry_exporter_otlp_http_builder)
+
+  target_link_libraries(opentelemetry_exporter_otlp_http_builder
```
dbarker: nit - these two target_link_libraries(opentelemetry_exporter_otlp_http_builder ..) calls can be combined into one call. Same in a few other places in this CMake file and others.

examples/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+target_link_libraries(
+  example_yaml ${CMAKE_THREAD_LIBS_INIT}
+  PRIVATE common_metrics_foo_library
+          common_logs_foo_library
+          opentelemetry-cpp::ostream_span_exporter_builder
+          opentelemetry-cpp::ostream_metrics_exporter_builder
+          opentelemetry-cpp::ostream_log_record_exporter_builder)
+
+if(WITH_OTLP_HTTP)
+  include_directories(${CMAKE_SOURCE_DIR}/exporters/otlp/include)
```
dbarker: We should be able to remove the use the global `include_directories` and `add_definitions` methods in this file. By linking the example_yaml to the builders the required definitions and include paths should be transitive from the exporters that are linked to the builders. If they are not transitive then it points to some missing library link made by the builder targets.

examples/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+  add_definitions(-DOTEL_HAVE_OTLP_HTTP)
+  target_link_libraries(example_yaml
```
dbarker: nit - these three target_link_library calls can be one call with the three targets listed after PRIVATE. Same elsewhere in this file.

exporters/ostream/CMakeLists.txt
```diff
@@ -53,6 +82,29 @@ if(BUILD_TESTING)
...
+
+add_library(opentelemetry_exporter_ostream_metrics_builder
+            src/console_push_metric_builder.cc)
+set_target_properties(opentelemetry_exporter_ostream_metrics_builder
+                      PROPERTIES EXPORT_NAME ostream_metrics_exporter_builder)
+set_target_version(opentelemetry_exporter_ostream_metrics_builder)
+
+target_include_directories(
+  opentelemetry_exporter_ostream_metrics_builder
+  PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>")
```
dbarker: Many of the builder targets are missing the `"$<INSTALL_INTERFACE:include>"` expression in their `target_include_directories` calls. Is that intended? Not adding it would indicate that we don't intend to include these headers in the installed packages for public use.

exporters/ostream/CMakeLists.txt
```diff
@@ -66,6 +118,35 @@ target_link_libraries(opentelemetry_exporter_ostream_logs
                       PUBLIC opentelemetry_logs)
 list(APPEND OPENTELEMETRY_OSTREAM_TARGETS opentelemetry_exporter_ostream_logs)
 
+if(BUILD_TESTING)
+  add_executable(ostream_log_test test/ostream_log_test.cc)
+  target_link_libraries(ostream_log_test ${GTEST_BOTH_LIBRARIES}
```
dbarker, line 125: nit - missing scope arg `PRIVATE`

exporters/otlp/CMakeLists.txt
```diff
@@ -323,7 +590,34 @@ if(WITH_OTLP_GRPC)
...
+    "otlp_grpc_span_builder.h"
+    EXCLUDE
+    PATTERN
+    "otlp_grpc_log_record_builder.h"
+    EXCLUDE
+    PATTERN
+    "otlp_grpc_push_metric_builder.h"
+    EXCLUDE)
+
+  otel_add_component(
```
dbarker, line 604: Should there be a guard for `WITH_CONFIGURATION` around adding the builder targets and components? Is the intent that these should always be created and installed?

.github/workflows/clang-tidy.yaml
dbarker: Sounds reasonable.

examples/configuration/CMakeLists.txt
marcalff: Fixed.

exporters/otlp/CMakeLists.txt
marcalff: Fixed.

cmake/ryml.cmake
marcalff: Fixed.

cmake/ryml.cmake
marcalff: Fixed.

exporters/ostream/CMakeLists.txt
marcalff: Thanks, fixed.

exporters/ostream/CMakeLists.txt
marcalff: This is existing code that was just moved to reorder things.

I prefer to have a separate PR to review and fix all links in all libraries instead, because the same is likely to exist in other places.

Beside, I don't understand what a private link means for an executable and what it does or changes.

examples/configuration/CMakeLists.txt
marcalff: For `include_directories`, fixed.

For `add_definitions`:

* WITH_OTLP_HTTP is a cmake option, tested only in cmake files
* There is no corresponding define visible in C++
* There is no such definition in the otlp http build.

Testing ifdef WITH_OTLP_HTTP is required in the example only, when populating the registry.
This is why the makefile here adds a definition if the cmake option is set.

exporters/otlp/CMakeLists.txt
marcalff: The intent is to always create and install builders.

These builders do not depend on a ryml parser, and can be used as a better programmatic interface to create SDK objects if needed.

For example, someone can craft a XXXConfiguration node, and call a builder to get an SDK XXX object.

The builders are in a separate library, so applications not using this (the default) will not bloat with unused code in the binary.

The flag `WITH_CONFIGURATION` only guards the configuration library, and the dependency to ryml.

exporters/ostream/CMakeLists.txt
dbarker: Sounds good as a future PR to clean up the target link libraries calls. 

Adding the scope keyword PRIVATE is a modern CMake best practice that makes the intent clear by explicitly stating this dependency is an internal implementation detail. Generally all our target link library calls should explicitly set the scope.

marcalff:
> Thanks for the PR. Excited to get this building :). I left some comments and mainly interested in getting the ryml dependency managed like the others (git tag files to define the version, FetchContent support, and install with the common third party CMake project instead of the bash script). Seems reasonable to follow up with these before the next release if it's too much change in this PR.

Thanks for all the comments.

Most fixed, the cmake install tests are deferred to a follow up PR.

I propose to merge this part now, it is self contained and a good size fix already, no need to make it bigger.

exporters/ostream/CMakeLists.txt
marcalff: This improves clarity and readability, for CMakeLists that define multiple libraries.

examples/configuration/CMakeLists.txt
dbarker: Thanks for the clarification. Since the definition is new and only needed by the example_yaml target then we should prefer the modern `target_compile_definitions` method.  The `add_definitions` method will add the definition to all targets in the same file and any sub folders added after the call. In general we should avoid it in favor of the modern target based method.   

https://cmake.org/cmake/help/latest/command/target_compile_definitions.html#command:target_compile_definitions

```cmake
target_compile_definitions(example_yaml PRIVATE OTEL_HAVE_OTLP_HTTP)
```

cmake/ryml.cmake
```diff
@@ -0,0 +1,30 @@
...
+if(NOT ryml_FOUND)
+  FetchContent_Declare(
+    "ryml"
+    GIT_REPOSITORY "https://github.com/biojppm/rapidyaml.git"
+    GIT_TAG "${ryml_GIT_TAG}"
+    )
+  set(ryml_PROVIDER "fetch_repository")
+
+  set(RYML_ENABLE_TESTING OFF CACHE BOOL "" FORCE)
+  set(RYML_ENABLE_INSTALL OFF CACHE BOOL "" FORCE)
```
dbarker, line 20: For the follow up PR with install tests we should also change this to 

```cmake
set(RYML_ENABLE_INSTALL ${OPENTELEMETRY_INSTALL} CACHE BOOL "" FORCE)
```

This should allow the install to complete if ryml is fetched.

exporters/otlp/CMakeLists.txt
dbarker: Thanks. I think I see how it is working when WITH_CONFIGURATION=OFF:
- The `opentelemetry-cpp::configuration` target, when built, would expose the sdk/configuration headers and libraries to targets that link to it. 
- The builders include headers publicly from sdk/configuration
- Since builders don't link to the `opentelemetry-cpp::configuration` target, they seem to find the sdk/configuration headers through the `opentelemetry-cpp::sdk` interface (header only) target linked through the exporter targets. This is how we don't see an "include not found error" for any jobs that set `WITH_CONFIGURATION=OFF`.  

That seems okay if the configuration header files included are header only.

It may make sense as a future follow up PR to break up the `opentelemetry-cpp::configuration` target into a "configuration_common" target and "configuration_ryml" target. The common target could then be linked in to all builders and make the internal sdk dependencies clear.

dbarker (approved):
Looks good. Thanks! Noting some potential follow up PRs. 
- replace `add_definitions` with `target_compile_definitions` in the example and elsewhere
- Update ryml.cmake to install ryml if `OPENTELEMETRY_INSTALL` is ON 
- Test the new yaml example with the install  
- Consider breaking up the opentelemetry-cpp::configuration target into `opentelemetry-cpp::configuration_common` and a   `opentelemetry-cpp::configuration_ryml` target that brings in the third party dependency.

PR #3661, opened 2025-09-25
[TRACE SDK] Batch span processor options now using env variables

### LINKED ISSUE #2085: [TRACE SDK] Batch Span Processor options not using environment variables
In this code:

```
struct BatchSpanProcessorOptions
{
  /**
   * The maximum buffer/queue size. After the size is reached, spans are
   * dropped.
   */
  size_t max_queue_size = 2048;

  /* The time interval between two consecutive exports. */
  std::chrono::milliseconds schedule_delay_millis = std::chrono::milliseconds(5000);

  /**
   * The maximum batch size of every export. It must be smaller or
   * equal to max_queue_size.
   */
  size_t max_export_batch_size = 512;
};
```

the default values for each configuration option are hard coded.

The [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#batch-span-processor) requires to use the following environment variables:

* `OTEL_BSP_SCHEDULE_DELAY`
* `OTEL_BSP_EXPORT_TIMEOUT`
* `OTEL_BSP_MAX_QUEUE_SIZE`
* `OTEL_BSP_MAX_EXPORT_BATCH_SIZE`

Note that the export timeout option is also missing from `BatchSpanProcessorOptions`.

nikhilbhatia08:
Fixes #2085

Using this change we can get `max_queue_size` `schedule_delay` `export_timeout` `max_export_batch_size` using environment variables. BatchSpanProcessor still does not use `export_timeout` and tests have been added.

## Changes


marcalff (changes requested):
Some changes needed:

The header file exposes implementation details, which are compiled into the application binary.

Move every implementation details, including default values, helpers, names of environment variables, to batch_span_processor_options.cc

See existing options structures for the pattern.

nikhilbhatia08:
@marcalff Applied the above changes, Thanks! But the BatchSpanProcessor does not have the field of `export_timeout` so it would be another issue for implementing it in `BatchSpanProcessor` ?

sdk/include/opentelemetry/sdk/trace/batch_span_processor_options.h
```diff
@@ -19,6 +20,7 @@ namespace trace
  */
 struct BatchSpanProcessorOptions
 {
+  BatchSpanProcessorOptions();
```
marcalff, line 22: Add `OPENTELEMETRY_EXPORT`.

sdk/include/opentelemetry/sdk/trace/batch_span_processor_options.h
```diff
@@ -28,6 +30,11 @@ struct BatchSpanProcessorOptions
   /* The time interval between two consecutive exports. */
   std::chrono::milliseconds schedule_delay_millis = std::chrono::milliseconds(5000);
 
+  /**
+   * The maximum time allowed to to export data
+   */
+  std::chrono::milliseconds export_timeout = std::chrono::milliseconds(3000);
```
marcalff: Please add a comment that this (`export_timeout`) is not used currently by the SDK, and that this parameter is ignored.

Remove the `std::chrono::milliseconds(3000)` default value from the header file, this should be only in the implementation, and not leaked into the application binary.

Likewise, remove the initialization of `schedule_delay_millis`.

sdk/include/opentelemetry/sdk/trace/batch_span_processor_options.h
```diff
@@ -5,6 +5,7 @@
 
 #include <chrono>
 
+#include "opentelemetry/sdk/common/env_variables.h"
```
marcalff: Remove.

sdk/src/trace/batch_span_processor_options.cc
```diff
@@ -0,0 +1,69 @@
...
+
+const size_t kDefaultMaxQueueSize                           = 2084;
+const std::chrono::milliseconds kDefaultScheduleDelayMillis = std::chrono::milliseconds(5000);
+const std::chrono::milliseconds kDefaultExportTimeout       = std::chrono::milliseconds(3000);
+const size_t kDefaultMaxExportBatchSize                     = 512;
+
+inline size_t GetMaxQueueSizeFromEnv()
+{
+  std::uint32_t value;
+  if (!opentelemetry::sdk::common::GetUintEnvironmentVariable(kMaxQueueSize, value))
```
marcalff: So:
- `kMaxQueueSize` is a name
- `kDefaultMaxQueueSize` is a value

This is confusing, please rename all environment variables with an Env ~prefix~ suffix, as in `kMaxQueueSizeEnv`.

georgthegreat:
@marcalff @nikhilbhatia08, this PR breaks a bunch of existing code without providing a sensible alternative.

Prior to that we were using the following syntax using C++20 aggregate initialization:
```
 auto processor = otel::BatchSpanProcessorFactory::Create(
            std::move(exporter),
            otel::BatchSpanProcessorOptions{
                .max_queue_size = config.batch().max_queue_size(),
                .schedule_delay_millis = std::chrono::milliseconds(config.batch().schedule_delay().MilliSeconds()),
                .max_export_batch_size = config.batch().max_export_batch_size(),
            }
        );
```

It does not seem that there is a sensible rewrite of this code.
Maybe this PR can be reverted?

georgthegreat:
NB: #3687 has the same problem

lalitb:
I believe the issue comes from adding a user-defined constructor, which means `BatchSpanProcessorOptions` is no longer an aggregate type. That disables C++20 designated initializers, so patterns like `.max_queue_size = ...` stop working even though they worked earlier.
We can fix this without reverting anything by removing the constructor and using default member initializers along with a small set of static helper functions for the environment variable logic. This restores aggregate initialization while keeping the original behavior intact.

georgthegreat:
Seems like a solution to me!

PR #3667, opened 2025-09-27
[Metrics] Allow registering one callback for multiple instruments

### LINKED ISSUE #3653: A mechanism for updating multiple ObservableIntruments in one callback
I'm writing a program that uses a custom hardware device, and I'd like to get metrics off that hardware device and into our OpenTelemetry-based observability stack. The hardware reports several metrics values of interest in one operation; i.e. it basically exposes a "get a bunch of telemetry" function.

We're using `PeriodicExportingMetricReader` to send telemetry to our OpenTelemetry collector every 30 seconds. Because of this, I'd ideally like to use observable counters/gauges, so that the values collected are as fresh as possible before they're exported.

The problem is, it doesn't really make sense to register N-many observable instruments for each value exposed by our hardware; there _is_ no way to get telemetry off the device other than reading it all at once, so each callback would have to collect the entire telemetry package independently. This is rather wasteful!

* Either some way to register a single callback which will be called once per collection cycle for multiple different observable instruments (receiving them all as an array argument, perhaps), or
* A way to run arbitrary user-provided callbacks not attached to any instrument to be executed in `Meter::Collect` (such a callback could be used to update a bunch of synchronous instruments all at once), or
* Somebody to pat me on the shoulder, and say what I'm doing to hack around this at the moment is Fine Actually.

My current workaround involves registering a "dummy" Observable instrument, and inside the callback for that I poll the hardware for telemetry and then record metrics from that into synchronous counters & gauges.

**Additional context**
Happy to send a PR if we can land on a desired behavour!

KJTsanaktsidis:
Fixes #3653

## Changes

This commit adds a pair of methods `Meter::RegisterCallback` and `Meter::DeregisterCallback` which allow you to set up one callback which can record observations for multiple observable instruments.

The callback will be passed an object of type MultiObserverResult, which can be used to get an ObserverResultT for an instrument.

This allows a single expensive operation to yield multiple different observable metrics without the need to call said expensive operation multiple times.

There's a test demonstrating the usage, but the short version is

```
void callback(metrics::MultiObserverResult &result, void *state)
{
    auto instrument1 = reinterpret_cast<MyState *>(state)->instrument1;
    auto instrument2 = reinterpret_cast<MyState *>(state)->instrument2;

   result.ForInstrument<int64_t>(instrument1).Observe(32, {"some", "labels"});
   result.ForInstrument<double>(instrument2).Observe(3234.32);
}

auto instrument1 = meter->CreateInt64ObservableCounter(...);
auto instrument2 = meter->CreateDoubleObservableGauge(...);

MyState state{instrument1, instrument2};
std::array<opentelemetry::metrics::ObservableInstrument *, 2> reg_instruments{instrument1, instrument2};
meter->RegisterCallback(callback, state, reg_instruments);
```

## Design notes

* I picked the names `RegisterCallback` based on the Go SDK, which exposes a RegisterCallback method on the Meter for this purpose. Is this the right name? Is Meter scope the right place to put this?
* The callback is required to have access to a `opentelemetry::metrics::ObservableInstrument` pointer somehow through `state` in order to record a measurement. This is in line with the interface sketched out in the spec - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#multiple-instrument-callbacks
* I felt pretty silly defining the variant `ObserverResultDirect = std::variant<ObserverResultT<double>, ObserverResultT<int64_t>>` when we already have `ObserverResult = std::variant<shared_ptr<ObserverResultT<double>>, shared_ptr<ObserverResultT<int64_t>>` but it does get rid of a needless extra heap allocation in the multi instrument case so it's probably for the best?
* I don't love the fact that the `MultiObserverResult` has protected `ForInstrumentDouble` and `ForInstrumentInt64`; this is needed because you can't have a virtual template method, and I wanted the recording interface to look like `res.ForInstrument<int64_t>(instrument).Observe(value)`. We could make `ForInstrument` return a variant, but that would require a) exposing `ObserverResultDirect` as part of the API, and b) requiring the calling code to `std::get` or `std::visit` that.
* There is in fact no _actual_ reason why this interface requires you to pre-register the instruments to be observed ahead of time; the code could perfectly happily add them on the fly in calls to `ForInstrument`. However this seems to be what the spec describes, so I've stuck to the spec.

Very keen to hear what you think!


owent:
Is it possiable to drop `callbacks_` and allow constructing `MultiObservableCallbackRecord` from `ObservableCallbackRecord`? Maintaining multiple types is cumbersome for users.

KJTsanaktsidis:
> Is it possiable to drop callbacks_ and allow constructing MultiObservableCallbackRecord from ObservableCallbackRecord?

Is the ask here to essentially do...

```
class ObservableRegistry {
private:
#if OPENTELEMETRY_ABI_VERSION_NO >= 2
    std::unordered_map<uintptr_t, std::unique_ptr<MultiObservableCallbackRecord>> multi_callbacks_;
#else
  std::vector<std::unique_ptr<ObservableCallbackRecord>> callbacks_;
#endif
  std::mutex callbacks_m_;
}
```

i.e. to re-implement the existing single-observable callback functionality in terms of the multi-observable callback functionality? Yes, I can do this - let me know if I understood correctly

> Maintaining multiple types is cumbersome for users.

This I didn't fully understand, because ObservableCallbackRecord and MultiObservableCallbackRecord are implementation details of the SDK ObservableRegistry; users shouldn't be touching these classes, I thought?

(In fact: the definition of `MultiObservableCallbackRecord` could be moved out of the header and into observable_registry.cc, by type-erasing `std::unordered_map<uintptr_t, std::unique_ptr<MultiObservableCallbackRecord>> multi_callbacks_` with a shared_ptr or a unique_ptr with a deleter function instead. I think I'll do this now, it seems worthwhile not to expose types that aren't needed).

KJTsanaktsidis:
Oh, one more thought; if ObservableRegistry can be thought of as a part of the internal API/ABI of the SDK, I could probably dispense with the ABI version guards in it (since symbols from ObservableRegistry would only be used inside the SDK library itself).

this would allow unconditionally reimplementing the existing single-observable callback in terms of the multi-observable implementation.

Reckon I should do this?

lalitb:
Yes, it should be safe to do ABI breaking changes with ObservableRegistry.

owent:
> > Is it possiable to drop callbacks_ and allow constructing MultiObservableCallbackRecord from ObservableCallbackRecord?
> 
> Is the ask here to essentially do...
> 
> ```
> class ObservableRegistry {
> private:
> #if OPENTELEMETRY_ABI_VERSION_NO >= 2
>     std::unordered_map<uintptr_t, std::unique_ptr<MultiObservableCallbackRecord>> multi_callbacks_;
> #else
>   std::vector<std::unique_ptr<ObservableCallbackRecord>> callbacks_;
> #endif
>   std::mutex callbacks_m_;
> }
> ```
> 
> i.e. to re-implement the existing single-observable callback functionality in terms of the multi-observable callback functionality? Yes, I can do this - let me know if I understood correctly

Yes, that's what I want.

> > Maintaining multiple types is cumbersome for users.
> 
> This I didn't fully understand, because ObservableCallbackRecord and MultiObservableCallbackRecord are implementation details of the SDK ObservableRegistry; users shouldn't be touching these classes, I thought?

We maintain cache of registered callbacks in our project, which allow us to do lazy registration, reload , cleanup and some others things, another group of types will make us to double the cache and APIs to maintain them.

BTW: It's OK to break ABI or API in v2 now.

KJTsanaktsidis:
> Yes, that's what I want.

OK, awesome - I think I've done that, hope this looks right to you 🙏

sdk/src/metrics/state/observable_registry.cc
```diff
@@ -1,107 +1,226 @@
...
 namespace metrics
 {
+struct ObservableCallbackRecord
+{
+  ObservableCallbackRecord(opentelemetry::metrics::ObservableCallbackPtr callback,
+                           void *state,
+                           opentelemetry::metrics::ObservableInstrument *instrument)
+      : callback(callback), state(state)
+  {
+    observable_result.RegisterInstrument(static_cast<ObservableInstrument *>(instrument));
```
marcalff, line 36: Why the static_cast ?
parameter `instrument` is of the proper type already.

sdk/src/metrics/state/observable_registry.cc
```diff
@@ -1,107 +1,226 @@
...
+  }
+
+  ObservableCallbackRecord(opentelemetry::metrics::MultiObservableCallbackPtr callback,
+                           void *state,
+                           nostd::span<opentelemetry::metrics::ObservableInstrument *> instruments)
+      : callback(callback), state(state)
+  {
+    for (auto *instrument : instruments)
+    {
+      observable_result.RegisterInstrument(static_cast<ObservableInstrument *>(instrument));
```
marcalff, line 46: Likewise, do we need the static_cast ?

sdk/src/metrics/state/observable_registry.cc
```diff
@@ -1,107 +1,226 @@
...
                                      opentelemetry::metrics::ObservableInstrument *instrument)
 {
-  // TBD - Check if existing
-  std::unique_ptr<ObservableCallbackRecord> record(
-      new ObservableCallbackRecord{callback, state, instrument});
   std::lock_guard<std::mutex> lock_guard{callbacks_m_};
-  callbacks_.push_back(std::move(record));
+  auto record = std::make_unique<ObservableCallbackRecord>(callback, state, instrument);
+  auto token  = reinterpret_cast<uintptr_t>(record.get());
+  callbacks_.insert({token, std::move(record)});
```
marcalff, line 68: The scope of the `lock_guard` now covers `record` and `token`.

Can the lock_guard only protect `callbacks_` as before, it seem sufficient.

sdk/src/metrics/state/observable_registry.cc
```diff
@@ -1,107 +1,226 @@
...
+
+uintptr_t ObservableRegistry::AddCallback(
+    opentelemetry::metrics::MultiObservableCallbackPtr callback,
+    void *state,
+    nostd::span<opentelemetry::metrics::ObservableInstrument *> instruments)
+{
+  std::lock_guard<std::mutex> lock_guard{callbacks_m_};
+  auto record = std::make_unique<ObservableCallbackRecord>(callback, state, instruments);
+  auto token  = reinterpret_cast<uintptr_t>(record.get());
+  callbacks_.insert({token, std::move(record)});
```
marcalff: Likewise, can the scope of `lock_guard` be reduced to only `callbacks_` ?

sdk/src/metrics/multi_observer_result.cc
```diff
@@ -0,0 +1,154 @@
...
+}
+
+opentelemetry::metrics::ObserverResultT<double> &MultiObserverResult::ForInstrumentDouble(
+    const opentelemetry::metrics::ObservableInstrument *instrument)
+{
+  static opentelemetry::sdk::metrics::ObserverResultT<double> null_result;
+  // const_cast is appropriate here, because we're _not_ modifying the passed-in pointer;
+  // we just need to make it non-const to be able to look it up in our map.
+  auto it = observer_results_.find(
+      const_cast<opentelemetry::metrics::ObservableInstrument *>(instrument));
```
marcalff, line 110: Not sure if it is feasible, but could the `observer_results_` map use `const opentelemetry::metrics::ObservableInstrument *` instead, and register / unregister use const instruments as well ?

marcalff: No need to change all the code, I just would like to discuss if using const instruments is a) possible and b) desirable.

sdk/src/metrics/state/observable_registry.cc
marcalff: I don't expect people will add / remove callbacks all day long, so this is not critical for performances.

The concern here is if someone adds more code in the middle, like printing internal logs to debug, and get the added code under a mutex lock by accident.

sdk/test/metrics/multi_observer_test.cc
```diff
@@ -0,0 +1,216 @@
...
+    EXPECT_EQ(0, metric_data.scope_metric_data_.size());
+    return true;
+  });
+  meter->DeregisterCallback(callback_id);
+}
+
+namespace
+{
+
+struct CallbackDeregisterdWhenDestroyedState
```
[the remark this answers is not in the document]
KJTsanaktsidis: Oops, fixed.

sdk/src/metrics/state/observable_registry.cc
KJTsanaktsidis: 👍 makes sense, done.

sdk/src/metrics/multi_observer_result.cc
KJTsanaktsidis: I don't _think_ it makes sense for it to store const instruments. This pointer winds up getting used in `MultiObserverResult::StoreResults` to call non-const `ObservableInstrument::GetMetricStorage` and write to the returned storage.

We _could_ store the pointer twice; once as a const pointer in the map key, and once as a non-const pointer in the map value; the call to `StoreResults` would dig out the non-const pointer from the map value. But I think storing two pointers to the same object is more confusing than this const_cast (which is pretty trivially not mutating the pointed-to value in any way)

KJTsanaktsidis:
Thanks, sorry I wasn't able to get to this last week - made the rest of the changes suggested.

sdk/src/metrics/multi_observer_result.cc
marcalff: Thanks for the investigation. No changes then, good as is.

KJTsanaktsidis:
Thank you, a pleasure working with you.

PR #3691, opened 2025-10-12
[Metrics] Perform async Re-Aggregation when Spatial Attributes are dropped

### LINKED ISSUE #1724: Async aggregation doesn't do re-aggregation properly when the spatial dimensions are dropped
As discussed here - https://github.com/open-telemetry/opentelemetry-specification/issues/2905, Most of the languages doesn't handle re-aggregation when the the spatial dimensions are dropped, and C++ is one of them. OpenTelemetry C++ uses the last of these processed measurements as the aggregated metrics to be reported.

So, if a view is defined for async counter to remove the attributes with "version" key. And we get the measurements.

async_counter->observer(1, {"version",1})     -->  (1, {})           
async_counter->observer(2, {"version",2})     -->  (2, {})
   
   Existing behavior -> Report the last measurement (2,{}) as the aggregated metrics
   Expected behavior -> Report the sum of both the measurements (3, {}) as the aggregated metrics.

nikhilbhatia08:
Fixes #1724 

Performing Reaggregation when spatial dimensions are dropped in async counters. There are also unit tests provided to the change. 

## Changes


lalitb:
Assigned it to self. Will review during this week.

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
```diff
@@ -0,0 +1,157 @@
...
+    else if (nostd::holds_alternative<const char *>(value))
+    {
+      return std::hash<nostd::string_view>{}(nostd::string_view(nostd::get<const char *>(value)));
+    }
+    else if (nostd::holds_alternative<nostd::string_view>(value))
+    {
+      return std::hash<nostd::string_view>{}(nostd::get<nostd::string_view>(value));
+    }
+
+    return 0;  // fallback for other types
```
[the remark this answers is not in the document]
nikhilbhatia08: As for now plan is to support only two types and then extend the support for other types

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
```diff
@@ -0,0 +1,157 @@
...
+      return nostd::string_view(nostd::get<const char *>(a)) ==
+             nostd::string_view(nostd::get<const char *>(b));
+    }
+    // Compare string_view
+    else if (nostd::holds_alternative<nostd::string_view>(a))
+    {
+      return nostd::get<nostd::string_view>(a) == nostd::get<nostd::string_view>(b);
+    }
+
+    return false;
```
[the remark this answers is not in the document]
nikhilbhatia08: Same as my above comment

sdk/include/opentelemetry/sdk/metrics/observer_result.h
```diff
@@ -22,32 +21,33 @@ template <class T>
 class ObserverResultT final : public opentelemetry::metrics::ObserverResultT<T>
 {
 public:
-  explicit ObserverResultT(const AttributesProcessor *attributes_processor = nullptr)
-      : attributes_processor_(attributes_processor)
-  {}
+  explicit ObserverResultT() = default;
```
Reneg973, line 24: explicit not required for default constructor. Here it can also be removed (rule of 0)

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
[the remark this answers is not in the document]
Reneg973: you could use nostd::visit() instead of if...else if...

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
```diff
@@ -0,0 +1,157 @@
...
+};
+
+template <class T>
+using Measurements = std::unordered_map<
+    std::unordered_map<nostd::string_view, opentelemetry::common::AttributeValue>,
+    T,
+    AttributeMapHash,
+    AttributeMapEqual>;
+
+}  // namespace metrics
```
Reneg973, line 155: add another type here for use in AsyncMetricStorage:
`using MeasurementsKeyType = typename Measurements::key_type;`

sdk/include/opentelemetry/sdk/metrics/observer_result.h
```diff
@@ -22,32 +21,33 @@ template <class T>
...
 
   ~ObserverResultT() override = default;
 
   void Observe(T value) noexcept override
   {
-    data_[MetricAttributes{{}, attributes_processor_}] = value;
+    std::unordered_map<nostd::string_view, opentelemetry::common::AttributeValue> empty;
```
Reneg973, line 35: Can be replaced with new type `MeasurementsKeyType`

sdk/include/opentelemetry/sdk/metrics/observer_result.h
```diff
@@ -22,32 +21,33 @@ template <class T>
...
+    data_[empty] += value;
   }
 
   void Observe(T value, const opentelemetry::common::KeyValueIterable &attributes) noexcept override
   {
-    data_[MetricAttributes{attributes, attributes_processor_}] =
-        value;  // overwrites the previous value if present
+    std::unordered_map<nostd::string_view, opentelemetry::common::AttributeValue> attr_map;
```
Reneg973, line 41: replace with new type `MeasurementsKeyType`

sdk/src/metrics/meter.cc
```diff
@@ -592,7 +592,7 @@ std::unique_ptr<AsyncWritableMetricStorage> Meter::RegisterAsyncMetricStorage(
        ,
        exemplar_filter_type
 #endif
-  ](const View &view) {
+  ](const View &view) -> bool {
```
Reneg973, line 597: superfluous, compiler knows it

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
Reneg973: I am not sure that it's worth, adding another runtime expensive associative container just for a list of some attributes. Unordered_map gets efficient with thousands and thousands of entries. But still very slow in comparison to a `boost::flat_map` or `absl::flat_hash_map` or a simple `std::vector`.

Reneg973: with the type `std::unordered_map<nostd::string_view, opentelemetry::common::AttributeValue>` it will also lead to dangling references. It will work when the string_view is created with "abcd", because this will stay in memory, but putting in a std::string... > crash

sdk/test/metrics/sum_aggregation_test.cc
```diff
@@ -322,6 +325,126 @@ TEST(CounterToSumFilterAttributes, Double)
...
+  MeterProvider mp;
+  auto m                      = mp.GetMeter("meter2", "version2", "schema2");
+  std::string instrument_unit = "ms";
+  std::string instrument_name = "counter2";
+  std::string instrument_desc = "counter metrics2";
+
+  opentelemetry::sdk::metrics::FilterAttributeMap allowedattr;
+
+  std::unique_ptr<opentelemetry::sdk::metrics::AttributesProcessor> attrproc{
+      new opentelemetry::sdk::metrics::FilteringAttributesProcessor(allowedattr)};
```
dbarker, line 339: please replace `new` with `std::make_unique` or `std::make_shared` accordingly. Rationale - [CppCoreGuidelines#r22 and r23](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r22-use-make_shared-to-make-shared_ptrs)

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
```diff
@@ -0,0 +1,157 @@
...
+      return nostd::get<nostd::string_view>(a) == nostd::get<nostd::string_view>(b);
+    }
+
+    return false;
+  }
+};
+
+// Hash function for unordered_map of key-value pairs
+// This Custom Hash is only applied to strings and int for now
+struct AttributeMapHash
```
dbarker, line 89: Please see `sdk/common/attributemap_hash.h` -> GetHash and GetHashForAttributeValueVisitor. Consider if these can be used.

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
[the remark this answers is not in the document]
dbarker: Please see `sdk/common/attributemap_hash.h` -> GetHash and GetHashForAttributeValueVisitor

dbarker:
Thanks for the PR @nikhilbhatia08! I've left a few nitpick comments but I'm interested to dig more into the specification and requirements for this change to provide a better review. 

The original linked issue #1724 is a bit old and it references an opentelemetry-specification issue 2905 that has since been closed as not planned. 

Can you update the description of the PR with a bit more context on the use cases enabled and links to the relevant sections of otel specification? Thanks!

sdk/test/metrics/sum_aggregation_test.cc
Reneg973: > please replace `new` with `std::make_unique` or `std::make_shared` accordingly. Rationale - [CppCoreGuidelines#r22 and r23](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r22-use-make_shared-to-make-shared_ptrs)

I would agree with you, if it would be used already. But it can't be found very often in that repository. Therefore I would say better be consistent.
Replacing every new with make_unique or make_shared would be a lot of changes and should be done in its own commit

sdk/include/opentelemetry/sdk/metrics/state/measurement_attributes_map.h
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+
+    return true;
+  }
+};
+
+template <class T>
+using Measurements = std::unordered_map<
+    std::unordered_map<nostd::string_view, opentelemetry::common::AttributeValue>,
```
ThomsonTan, line 150: Is it safe to use `nostd::string_view` here as the key of `std::unordered_map`? Or better to use `std::string` which owns the string?

Reneg973: In C++ it is like always an "it depends".

As long as the type is destroyed before the strings it references to, it's fine.

lalitb:
@nikhilbhatia08  - 

I agree with the motivation here: async measurements need to be re-aggregated after a view drops spatial attributes, otherwise multiple observations can collapse onto the same attribute set incorrectly.

That said, I don’t think the current data structure change is safe. In particular, storing measurements as `unordered_map<unordered_map<string_view, AttributeValue>, T>` looks problematic:

- `string_view` does not own storage, so this can hold dangling references once the callback returns
- using an `unordered_map` as the key of another `unordered_map` is a pretty heavy hot-path choice
- the custom hash/equality only handles a subset of supported attribute value types, so it is not equivalent to the existing metric attribute machinery

I think the re-aggregation should still happen, but I would strongly prefer doing it with the existing owned `MetricAttributes`/attribute hashing path rather than introducing `measurement_attributes_map.h`.

Also, the unconditional `+=` behavior seems too broad for async instruments in general. For sum-like instruments that makes sense, but for observable gauges I would expect last-value semantics to be preserved.

The test cases in `sum_aggregation_test.cc` nicely demonstrate the expected re-aggregation behavior, and those should carry over well into a revised approach.

nikhilbhatia08:
closing this pr as its almost one year old, and I can redo this issue because I lost context on this one :)

PR #3693, opened 2025-10-14
[CONFIGURATION] File configuration - parser location

marcalff:
Contributes to #2481

This is a partial fix, to implement parser location, for better yaml error messages.

## Changes

* Implement the "Location" of a yaml node in the parser
  * Make all document nodes point to the document root
  * Make the document root own the yaml parser that is used to create the document
  * Inspect a node location by following the links up to the document root, and then invoke the rapidyaml parser.

* Implement a location property in yaml related exceptions

* Change all the parsing code to pass the current yaml node as context, in case code needs to access a node location to raise an exception.

With all these changes, when the parser fails with an invalid schema exception, the exception contains:

* the name of the yaml input file used
* the line number, column number, and offset from the document start, where the error was detected.

This helps ease of use, so error messages about invalid config.yaml files point precisely at the error location.

### DIFF
--- a/.iwyu.imp
+++ b/.iwyu.imp
@@ -14,6 +14,7 @@
   # Work around for ryml
   { "include": ["<c4/std/string.hpp>", "private", "<ryml_std.hpp>", "public"] },
 
+  { "include": ["<c4/yml/common.hpp>", "private", "<ryml.hpp>", "public"] },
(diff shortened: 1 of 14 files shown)


marcalff:
Abandoning this PR and re creating it, it has the wrong origin.

PR #3706, opened 2025-10-19
SpinLockMutex small cleanup; improved and extended benchmark

Reneg973:
Fixes # (issue)

## Changes

- `load()` not required in `try_lock()`
- shortened loop in `lock()` 
- fixed erroneous yield() in `BM_ThreadYieldSpinLockThrashing`
- added `std::mutex` to benchmark
-  removed superfluous `static` as code is already in unnamed namespace

It seems that at least on Apple M1 (lClang+libc) the std::mutex outperforms SpinlockMutex with higher number of threads.


api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -95,13 +94,9 @@ class SpinLockMutex
    */
   void lock() noexcept
   {
-    for (;;)
+    // Try once
+    while (!try_lock())
```
lalitb, line 94: This seems less efficient under contention. Calling try_lock() in every iteration causes repeated atomic exchanges and cache invalidations. The previous version avoided that by doing the initial exchange outside the spin loop.

Reneg973: Thank you for the fast review. I don't get the point. The if` (!flag_exchange...)` previously was already within the loop. I just moved it into the loop condition and replaced for with while.
```
// initial
for (;;) {
  if (!flag_.exchange(true, std::memory_order_acquire))
      {
        return;
      }
// same:
for (;flag_.exchange(true, std::memory_order_acquire);) {
// same:
while (flag_.exchange(true, std::memory_order_acquire)) {
```
But ok, there is the change in try_lock() you may have noted. The cause here is that the previously called `load()` is a bit more performant if the lock is already held, it prevents a write operation. OTOH it makes the try_lock() a non-atomic operation.

The issue with heavy contention is, we don't know how long a lock is held. In most cases it is far more than a fast_yield() or a yield(). Here it prevents other threads from taking CPU time as it blocks the core. So, a spin lock is usable most if we have just light or no contention. If it is assumed, that heavy contention is the default, std::mutex should be preferred.

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -82,8 +82,7 @@ class SpinLockMutex
    */
   bool try_lock() noexcept
   {
-    return !flag_.load(std::memory_order_relaxed) &&
-           !flag_.exchange(true, std::memory_order_acquire);
+    return !flag_.exchange(true, std::memory_order_acquire);
```
maxmarkin01: How is this more efficient? load(std::memory_order_relaxed) would avoid having a memory fence and wait for cache sync. Having said that the whole idea of SpinLock in user mode seems dubious especially in high contention scenarios.

Reneg973:
I'm confused why this CI/Format check fails! Any hint is welcome

marcalff:
@Reneg973 

Tip to format all the code (C++, CMake, Bazel):

```
[malff@malff-desktop ~]$ more ~/bin/otel_cf.sh 
#!/bin/bash
podman run --rm --privileged=true --volume ${PWD}:/otel otel/cpp_format_tools
```

See https://github.com/open-telemetry/cpp-build-tools/tree/main/cpp_format_tools

marcalff:
Thanks for the contribution.

This PR is stale, and there was no consensus on code review. Now closing.

Issue #2544, opened 2024-02-21
Static build fails without overflow check


olegrok:
I got following error:
```
In member function 'Resize',
    inlined from 'Push' at /nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:286:15,
    inlined from 'Attach' at /nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:229:20:
/nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:299:47: error: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
/nix/store/WXPVL9V9N2S0SD6IQGIK43JK9X4HXFS4-x86_64-unknown-linux-musl-gcc-13.2.0/include/c++/13.2.0/new: In member function 'Attach':
/nix/store/WXPVL9V9N2S0SD6IQGIK43JK9X4HXFS4-x86_64-unknown-linux-musl-gcc-13.2.0/include/c++/13.2.0/new:128:26: note: in a call to allocation function 'operator new []' declared here
In member function 'Resize',
    inlined from 'Push' at /nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:286:15,
    inlined from 'Attach' at /nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:229:20,
    inlined from 'Attach' at /nix/store/88WAXS933YDRH4ARHWAWF1G85H7L0F9Y-opentelemetry-static-x86_64-unknown-linux-musl-0.1.0/include/opentelemetry/context/runtime_context.h:95:54,
[...]

PR #3709, opened 2025-10-22
[API] Fix -Werror=alloc-size-larger-than= warning in runtime_context.h

marcalff:
Contributes to #2544

## Changes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -37,10 +37,10 @@ jobs:
       run: |
         sudo -E ./ci/setup_ci_environment.sh
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v4.30.8
+      uses: github/codeql-action/init@16140ae1a102900babc80a33c44059580f687047 # v4.30.9
       with:
        languages: cpp
     - name: Autobuild
-      uses: github/codeql-action/autobuild@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v4.30.8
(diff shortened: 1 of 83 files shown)


api/include/opentelemetry/context/runtime_context.h
```diff
@@ -278,6 +303,10 @@ class ThreadLocalContextStorage : public RuntimeContextStorage
       {
         return Context();
       }
+      if (size_ > max_capacity_)
+      {
+        return Context();
```
lalitb, line 308: Just curious - should it return the last valid context, instead of empty one ?

marcalff: Good question.

In my understanding, `Scope` will attach and detach contexts on creation and destruction, adding/removing from the stack.

If the code returns the last valid context, then we will have:
- every scope nested at depth max_capacity_, max_capacity_ + 1, ..., max_capacity_ + N share the same context.
- Scope at max_capacity_ + N will destroy the last valid scope on destruction, causing side effects.

So, I think `Top()` and `Pop()` should return an empty context instead, once the stack depth goes beyond the limit.

This is very theoretical, because any process with 1 million nested scopes in the call stack will never reach this point, being long dead, already killed for other reasons (stack overflow, out of memory).

api/include/opentelemetry/context/runtime_context.h
```diff
@@ -286,6 +315,10 @@ class ThreadLocalContextStorage : public RuntimeContextStorage
     void Push(const Context &context) noexcept
     {
       size_++;
+      if (size_ > max_capacity_)
+      {
+        return;
+      }
       if (size_ > capacity_)
       {
         Resize(size_ * 2);
```
Reneg973, line 324: @marcalff 
Did you see this line?
`Resize(size_ * 2);`

PR #3712, opened 2025-10-24
Context and shared ptr improvements

Reneg973:
Fixes # (issue)
none, just improvements

## Changes
non-API changes in `Context` and `nostd::shared_ptr`

- may reduce sizeof(nostd::shared_ptr) and therefore sizeof(Context), if sizeof(std::shared_ptr) == 16 instead of 32
- may create faster key-value in case of short strings
- may perform faster because of removed vtable functions


Reneg973:
excuse me, forgot to rebase on main

PR #3713, opened 2025-10-24
Context and shared ptr improvements

Reneg973:
Fixes # (issue)
non, just improvements

## Changes
non-API changes in `Context` and `nostd::shared_ptr`

- may reduce `sizeof(nostd::shared_ptr) `and therefore `sizeof(Context)`, if sizeof(std::shared_ptr) == 16 instead of 32
- prevent dynamic allocation in case of short keys (SSO)
- may perform faster because of removed vtable functions


Reneg973:
missing header `<string>` for self sufficiency in context.h, will update it later

api/include/opentelemetry/nostd/shared_ptr.h
```diff
@@ -105,20 +102,24 @@ class shared_ptr
 
   shared_ptr(unique_ptr<T> &&other) noexcept
   {
-    std::shared_ptr<T> ptr_(other.release());
+    std::shared_ptr<T> ptr_(std::move(other));
```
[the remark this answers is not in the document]
Reneg973: I don't get this!
According to [cppref](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr.html )(constructor 13) this constructor exists. Additionally nostd::unique_ptr implements move operations perfectly.
Also, if this would be true, then code like:
```
nostd::unique_ptr<T> Foo() { return new T; }
std::shared_ptr<T> ptr(Foo());
```
would not work.

PS: GPT-5 says everything is fine with this code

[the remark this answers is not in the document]
owent: As I understand it, std::shared_ptr should not recognize nostd::unique_ptr and cannot be constructed from its rvalue.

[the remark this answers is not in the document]
Reneg973: > As I understand it, std::shared_ptr should not recognize nostd::unique_ptr and cannot be constructed from its rvalue.

Perfect, I was blind and naive. Thought I saw a test for it, but didn't. After writing one I directly had compiler error.

api/include/opentelemetry/context/context.h
```diff
@@ -34,18 +34,18 @@ class Context
 
   // Creates a context object from a key and value, this will
   // hold a shared_ptr to the head of the DataList linked list
-  Context(nostd::string_view key, ContextValue value) noexcept
+  Context(nostd::string_view key, const ContextValue &value) noexcept
```
marcalff, line 37: ABI break

api/include/opentelemetry/context/context.h
```diff
@@ -57,9 +57,9 @@ class Context
   // Accepts a new iterable and then returns a new context that
   // contains the new key and value data. It attaches the
   // exisiting list to the end of the new list.
-  Context SetValue(nostd::string_view key, ContextValue value) noexcept
+  Context SetValue(nostd::string_view key, const ContextValue &value) const noexcept
```
marcalff, line 60: Valid and better, but ABI break

api/include/opentelemetry/context/context.h
```diff
@@ -92,12 +89,10 @@ class Context
   // A linked list to contain the keys and values of this context node
   struct DataList
   {
-    char *key_ = nullptr;
+    std::string key_;
```
marcalff, line 92: ABI break

api/include/opentelemetry/nostd/shared_ptr.h
```diff
@@ -47,14 +42,12 @@ class shared_ptr
 
     shared_ptr_wrapper(std::shared_ptr<T> &&ptr) noexcept : ptr_{std::move(ptr)} {}
 
-    virtual ~shared_ptr_wrapper() {}
-
-    virtual void CopyTo(PlacementBuffer &buffer) const noexcept
+    void CopyTo(PlacementBuffer &buffer) const noexcept
```
marcalff, line 45: Valid and better, but ABI break.

api/include/opentelemetry/nostd/shared_ptr.h
```diff
@@ -67,15 +60,19 @@ class shared_ptr
...
+    void Reset() noexcept { ptr_.reset(); }
 
   private:
     std::shared_ptr<T> ptr_;
   };
 
-  static_assert(sizeof(shared_ptr_wrapper) <= kMaxSize, "Placement buffer is too small");
+  struct alignas(kAlignment) PlacementBuffer
+  {
+    char data[sizeof(shared_ptr_wrapper)]{};
```
marcalff, line 73: ABI break

api/include/opentelemetry/nostd/shared_ptr.h
```diff
@@ -111,14 +108,18 @@ class shared_ptr
...
   }
 
   ~shared_ptr() { wrapper().~shared_ptr_wrapper(); }
 
   shared_ptr &operator=(shared_ptr &&other) noexcept
   {
+    if (this == &other)
+    {
+      return *this;
+    }
```
marcalff, line 122: This is valid, please post a separate PR for this fix,

marcalff (changes requested):
@Reneg973 

Thanks for the PR.

This area is much more complex that is appears on first sight.

Both the API and the ABI here must be immutable, so that:

* an library compiled with opentelemetry-cpp API 1.23.0 or older
* another library compiled with opentelemetry-cpp API 1.24.0 or newer
* all linked in an application configured with possibly yet another SDK version

all agree on what bit in the binary layout of a context or shared_ptr means, and can operate on the same data.

If we were to implement this area from scratch today, we would probably use most of the cleanup done here, but unfortunately we can not, due ABI breaking changes.

We could implement some changes only in ABI_VERSION >= 2, but this is not worth the added complexity in my opinion.

Some areas in opentelemetry-cpp are easier and less risky than others, the API is deceptive, and probably the most difficult to change, compared to contributions in the SDK itself for example.

Reneg973:
> We could implement some changes only in ABI_VERSION >= 2, but this is not worth the added complexity in my opinion.

Well, if you decide not to integrate it, you can close this PR, which is ok, even if I disagree.
The API and ABI currently add superfluous and also noticeable runtime overhead, especially for more lightweight C++ apps or on CPUs like a Raspi Zero. 
On heavily multithreaded apps: you don't use a custom private heap, which means every dynamic alloc/free blocks other threads using the global heap. Also there's a lot of shared_ptr usage. Especially the virtual inside the shared_ptr privates just adds superfluous indirections. 

>>an library compiled with opentelemetry-cpp API 1.23.0 or older
another library compiled with opentelemetry-cpp API 1.24.0 or newer

Does this mean you want to allow use cases where an object is created with API 1.23 and pushed into 1.24? Otherwise I don't see that they influence each other.

This was just a first proposal to slim down the overhead without losing functionality.

marcalff:
> > We could implement some changes only in ABI_VERSION >= 2, but this is not worth the added complexity in my opinion.
> 
> Well, if you decide not to integrate it, you can close this PR, which is ok, even if I disagree.

@Reneg973 

The problem is as follows:

For `OPENTELEMETRY_ABI_VERSION_NO == 1`, the current code is as it is, and we can not change it if this breaks the abi.

For `OPENTELEMETRY_ABI_VERSION_NO >= 2`, the code can be changed.

I agree it would improve the final product, but it also creates a significant effort in maintenance in the mean time, with potential regressions if things are different between abi 1 and 2. This is what I meant with added complexity.

> The API and ABI currently add superfluous and also noticeable runtime overhead, especially for more lightweight C++ apps or on CPUs like a Raspi Zero.

No contest here. The question is whether this is fixable without breaking existing applications, and without too much burden.

> On heavily multithreaded apps: you don't use a custom private heap, which means every dynamic alloc/free blocks other threads using the global heap.

The API code has to be header only, which excludes using a custom allocator, memory pool, or anything like it.

The SDK and exporters have more implementation freedom, but this is another area.

> Also there's a lot of shared_ptr usage.

For example, see #828

This has been known for a long time, and is not trivial to fix, due to the breaking change label.

> Especially the virtual inside the shared_ptr privates just adds superfluous indirections.

That sounds fixable in ABI_VERSION 2.

> 
> > > an library compiled with opentelemetry-cpp API 1.23.0 or older
> > > another library compiled with opentelemetry-cpp API 1.24.0 or newer
> 
> Does this mean you want to allow use cases where an object is created with API 1.23 and pushed into 1.24? Otherwise I don't see that they influence each other.
> 

An application or a library is instrumented by invoking the API only.
Say a library is compiled against API 1.23.
Calls to create a span (to instrument traces) can land either in the no-op implementation, or in the SDK.

The SDK is what is configured in the main() for the overall application. It could be SDK 1.25, which implements the API.

The problem here is that SDK 1.25 has to work not just with API 1.25, which is in the same code base in the same git label, but also with API 1.23. There is no CI build to verify this, so every time any code under api/include is touched, it needs extra review.

Long story short, if a library is instrumented (with API 1.23) and ships a binary package, linking this library in an application that uses the SDK 1.25 should not require whoever provided the library to produce and ship again another package (built with API 1.25).


> This was just a first proposal to slim down the overhead without losing functionality.

And this is appreciated.

Overall, ABI 1 and 2 are already different, and CI is already covering both, so having a few more ifdef to fix existing issues in ABI 2 only is possible.

Ultimately, this work is needed, and the sooner ABI 2 can be cleaned up and ready, the sooner we can deprecate ABI 1.

The pitfall to avoid is to start making changes, then get distracted and delayed, to end up with ABI 1 still in use, ABI 2 very different, better but still not complete, with diverging code.

Considering all the cleanup needed anyway due to recent clang-tidy findings in #3762, it looks like opentelemetry-cpp can not postpone this cleanup forever, so we might as well start it.

Do you think the PR can be adjusted to change only ABI_VERSION 2, and pass all ci ? It will be considered.

Reneg973:
Your explanations are very appreciated, thanks.

> Do you think the PR can be adjusted to change only ABI_VERSION 2, and pass all ci ? It will be considered.

ok, that's the kind of statement I waited for (getting a hint, what to do that this PR could be approved).
I'll prepare one, maybe not tomorrow, but in the next days.

marcalff:
This PR itself can not be merged due to ABI changes, as explained earlier, so now closing.

Some fragments of it have been merged already, see PR #3768.

Now, to be fair, and even if this very PR is now closed, it pointed out a lot of things that ultimately convinced us maintainers to do a full pass using clang-tidy, resulting in a massive cleanup.

A second clang-tidy pass is in progress with a more recent compiler version, with associated cleanup in progress.

PR #3714, opened 2025-10-25
[BUILD] Support build proto as shared lib on Windows

### LINKED ISSUE #3620: v1.22.0 no longer respects BUILD_SHARED_LIBS: ON
**Describe your environment**
openSUSE 15.6, gcc 7.5.0, cmake 4.0.3

We're trying to upgrade from v1.16.1 to v.1.22.0 and configuring opentelemetry-cpp v1.22.0 using the following command:

```
cmake <BUILD_ROOT>/opentelemetry-cpp-1.22.0 -DCMAKE_CXX_FLAGS="-std=c++14 -m64" \
-DCMAKE_TOOLCHAIN_FILE=build64.toolchain.cmake -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
-DWITH_ABI_VERSION=2 -DWITH_ABI_VERSION_2=ON -DWITH_ABI_VERSION_1=OFF \
-DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF -DWITH_ABSEIL=ON \
-DWITH_DEPRECATED_SDK_FACTORY=OFF -DWITH_OTLP_GRPC=OFF -DWITH_OTLP_FILE=ON \
-DWITH_OTLP_HTTP=ON -DWITH_OTLP_HTTP_COMPRESSION=ON -DWITH_EXAMPLES_HTTP=OFF \
-DWITH_OTLP_HTTP_SSL_PREVIEW=ON -DWITH_OTLP_HTTP_SSL_TLS_PREVIEW=ON \
-DOTELCPP_PROTO_PATH=<BUILD_ROOT>/opentelemetry-proto-1.7.0 \
-DCMAKE_PREFIX_PATH=\
<BUILD_ROOT>/json-3.12.0-bin;\
<BUILD_ROOT>/zlib-1.3.1-bin;\
<BUILD_ROOT>/googletest-1.14.0-bin;\
<BUILD_ROOT>/benchmark-1.8.3-bin;\
<BUILD_ROOT>/protobuf-26.1-bin;\
<BUILD_ROOT>/abseil-cpp-20240116.2-bin

```
All dependencies including protobuf and abseil are built as static libraries. Dependency versions:
```
OPENTELEMETRY_VERSION=1.22.0
   OTEL_PROTO_VERSION=1.7.0
NLOHMANN_JSON_VERSION=3.12.0
     PROTOBUF_VERSION=26.1
    ABSEILCPP_VERSION=20240116.2
        GTEST_VERSION=1.14.0
       GBENCH_VERSION=1.8.3
         ZLIB_VERSION=1.3.1

```

**Steps to reproduce**
Configure opentelemetry-cpp v1.22.0 using the above command. Then perform the "cmake --build" and "cmake --install" steps. 

What did you expect to see?
The lib directory of the install location contains only shared libraries - just like it did in v1.16.1 - because of the -DBUILD_SHARED_LIBS=ON option.

What did you see instead?
The lib directory of the install location contains a mixture of shared and static libraries, despite the -DBUILD_SHARED_LIBS=ON option.

```
$ ls
libopentelemetry_common.so                     libopentelemetry_exporter_ostream_metrics.so  libopentelemetry_exporter_otlp_file_log.a         libopentelemetry_exporter_otlp_http_log.a       libopentelemetry_metrics.so         libopentelemetry_trace.so
libopentelemetry_exporter_in_memory_metric.so  libopentelemetry_exporter_ostream_span.so     libopentelemetry_exporter_otlp_file_metric.a      libopentelemetry_exporter_otlp_http64_metric.a  libopentelemetry_otlp_recordable.a  libopentelemetry_version.so
libopentelemetry_exporter_in_memory.so         libopentelemetry_exporter_otlp_file.a           libopentelemetry_exporter_otlp_http.a           libopentelemetry_http_client_curl.so            libopentelemetry_proto.a            pkgconfig
libopentelemetry_exporter_ostream_logs.so      libopentelemetry_exporter_otlp_file_client.a    libopentelemetry_exporter_otlp_http_client.a  libopentelemetry_logs.so                        libopentelemetry_resources.so

```

**Additional context**
I'm aware of https://github.com/open-telemetry/opentelemetry-cpp/issues/3405 and https://github.com/open-telemetry/opentelemetry-cpp/pull/3435. However, we're not using grpc at all, only protobuf and only indirectly - because opentelemetry-cpp requires it - and should not be affected by the problems with the latest protobuf versions - unless I misunderstand #3405 and #3435.

### LINKED ISSUE #3665: Opentelemetry windows runtime issue
Opentelemtry is build on Windows 10 with shared libs on. When i linked to the test app and run the test app. i see 
below issue.

D:\workdir\win_shared_build\tc-monitor-cpp\build>.\bin\Release\runTests.exe                                                                                                 WARNING: All log messages before absl::InitializeLog() is called are written to STDERR                                                                                      E0000 00:00:1758866127.468952   33748 descriptor_database.cc:633] File already exists in database: opentelemetry/proto/common/v1/common.proto                               F0000 00:00:1758866127.469228   33748 descriptor.cc:2236] Check failed: GeneratedDatabase()->Add(encoded_file_descriptor, size)                                             *** Check failure stack trace: ***                                                                                                                                              @   00007FFCAF4B2499  (unknown)                                                                                                                                             @   00007FFC761C70DF  (unknown)                                                                                                                                             @   00007FFC762234C6  (unknown)                                                                                                                                             @   00007FFC762234B2  (unknown)                                                                                                                                             @   00007FFC762234B2  (unknown)                                                                                                                                             @   00007FFD477AE473  (unknown)                                                                                                                                             @   00007FFD24900F86  (unknown)                                                                                                                                             @   00007FFD249010FC  (unknown)                                                                                                                                             @   00007FFD49E29A1D  (unknown)                                                                                                                                             @   00007FFD49E7D2F7  (unknown)                                                                                                                                             @   00007FFD49E7D08A  (unknown)                                                                                                                                             @   00007FFD49E7D110  (unknown)                                                                                                                                             @   00007FFD49E7D110  (unknown)                                                                                                                                             @   00007FFD49EE3CB2  (unknown)                                                                                                                                             @   00007FFD49E85DEB  (unknown)                                                                                                                                             @   00007FFD49E85C73  (unknown)                                                                                                                                             @   00007FFD49E85C1E  (unknown)

owent:
Fixes #3665
Fixes #3620

## Changes

- Set `dllexport_decl` for proto targets and can be built as shared on windows.


owent:
My MSVC have LNK2001 unresolved external symbol for `__std_find_last_trivial_1` , `__std_search_1` , `__std_search_1` , `__std_min_8i` and many other STL's symbols, I will wait the next VS update and continue this work then.

ci/do_ci.ps1
```diff
@@ -67,7 +67,7 @@ switch ($action) {
     cmake $SRC_DIR `
       -DVCPKG_TARGET_TRIPLET=x64-windows `
       -DOPENTELEMETRY_BUILD_DLL=1 `
-     "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_DIR/scripts/buildsystems/vcpkg.cmake"
+      "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_DIR/scripts/buildsystems/vcpkg.cmake"
```
dbarker, line 99: Curious why all these format only changes are occurring. Do we have formatting applied to the powershell scripts?

cmake/opentelemetry-proto.cmake
```diff
@@ -2,17 +2,21 @@
...
+# 2 - Fetch from the opentelemetry-proto git submodule
+# (opentelemetry-cpp/third_party/opentelemetry-proto)
 #
-# 3 - Fetch from github using the git tag set in opentelemetry-cpp/third_party_release
+# 3 - Fetch from github using the git tag set in
+# opentelemetry-cpp/third_party_release
 #
 
-set(OPENTELEMETRY_PROTO_SUBMODULE "${opentelemetry-cpp_SOURCE_DIR}/third_party/opentelemetry-proto")
+set(OPENTELEMETRY_PROTO_SUBMODULE
```
dbarker: Looks like this file is also including a lot of format only changes. Can these be easily removed to so the PR only includes the intended changes?

cmake/opentelemetry-proto.cmake
```diff
@@ -323,10 +359,24 @@ if(WITH_OTLP_GRPC)
     ${LOGS_SERVICE_GRPC_PB_CPP_FILE} ${METRICS_SERVICE_GRPC_PB_CPP_FILE})
   set_target_version(opentelemetry_proto_grpc)
 
+  if(OTELCPP_PROTO_LIB_TYPE STREQUAL "SHARED_LIBRARY")
+    if(CMAKE_SYSTEM_NAME MATCHES "Windows|MinGW|WindowsStore")
+      # The codes generated by gRPC plugin do not support dll export/import
```
dbarker, line 343: gRPC does not seem well supported when built as a windows dll. See https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/BUILDING.md#windows_a-note-on-building-shared-libs-dlls



Should this be something the project tries to support and if so can it be a separate PR? It is unclear if we can/should support gRPC as a windows dll and test it in CI.

cmake/tools.cmake
```diff
@@ -222,3 +222,100 @@ function(project_build_tools_patch_default_imported_config)
     endif()
   endforeach()
 endfunction()
+
+function(project_build_tools_set_export_declaration OUTPUT_VARNAME)
```
dbarker, line 226: It is unclear what the impacts will be to platforms not currently testing in CI (including cross compilation which is not tested). Ideally we don't embed more platform specific information in these CMake files/functions.

Please consider making only the changes needed support opentelemetry-proto built as a windows dll and adding a specific test for it.

exporters/otlp/CMakeLists.txt
```diff
@@ -250,10 +250,12 @@ if(WITH_OTLP_HTTP)
...
     opentelemetry_exporter_otlp_http_client
-    PUBLIC opentelemetry_sdk opentelemetry_ext
-    # Links flags of opentelemetry_http_client_curl should be public when
-    # building internal components
-    PRIVATE opentelemetry_proto
+    PUBLIC opentelemetry_sdk
+           opentelemetry_ext
+           # Links flags of opentelemetry_http_client_curl should be public when
+           # building internal components
+           "$<BUILD_INTERFACE:opentelemetry_proto>"
```
dbarker, line 257: Can you share more about why this change is needed?

dbarker:
Thanks for the PR @owent! I've taken a quick pass and shared some feedback. It may be worth a SIG meeting discussion if the project should support windows dlls of protobuf and especially grpc given that projects warnings about unexpected behavior: 

>  you've been warned that there are some important drawbacks and some things might not work at all or will be broken in interesting ways. 

Separately we should investigate why the formatting tools are not applied consistently - is it possible the powershell scripts and .cmake files are not covered?

ci/do_ci.ps1
owent: I use the PowerShell extension in VS Code, but I don’t know if there’s a command-line alternative to PowerShell.

cmake/opentelemetry-proto.cmake
owent: cmake-format changes to CMake files were reverted, but PowerShell formatting changes were kept so all code uses a consistent style.

cmake/opentelemetry-proto.cmake
owent: I’ve reviewed the gRPC-generated code and confirmed it doesn’t contain global variables, so forcing it to build as static libraries should be safe. However, this might confuse users, create many duplicate symbols, and increase linker workload, similar to #3620.

We should discuss whether to build the gRPC code as a DLL or keep it as static libraries. I can rarely attend the SIG meeting due to the time zone, so could we discuss this on GitHub or Slack, or could someone share the outcome of the discussion with me afterward?

cmake/tools.cmake
owent: These changes apply only to users who build otel-cpp components as separate libraries rather than a single all-in-one DLL (ext/src/dll). The CI jobs in `ci/do_ci.ps1` that omit `-DOPENTELEMETRY_BUILD_DLL=1` and use `-DVCPKG_TARGET_TRIPLET=x64-windows` already cover this configuration.

The new functions also align behavior between Unix-like systems and Windows. Issue #3665 is not Windows-specific. For example, if users link against shared protobuf libraries but build otel-cpp’s proto targets as static libraries on Unix-like systems, then use those targets in multiple shared libraries, symbol registration can conflict and cause runtime crashes. If they instead build otel-cpp’s proto targets as shared libraries with `-fvisibility=hidden` on Unix-like systems, linking fails due to missing symbols.

We can use `project_build_tools_set_export_declaration` to enforce default symbol visibility on Unix-like systems and to set the appropriate export declarations on Windows.

exporters/otlp/CMakeLists.txt
owent: Some examples don’t include all required dependencies. To avoid unrelated changes in this PR, this is a temporary workaround. Can we optimize them in a separate PR?

ci/do_ci.ps1
```diff
@@ -125,6 +125,14 @@ switch ($action) {
...
       exit $exit
     }
+
+    $ALL_DLL_FILES = Get-ChildItem -Path "./*.dll" -Recurse
+    $ALL_DLL_DIRS = $(foreach ($dll_file in $ALL_DLL_FILES) {
+        $dll_file.Directory.FullName
+      }) | Sort-Object | Get-Unique
+    $env:PATH = ($ALL_DLL_DIRS -Join [IO.Path]::PathSeparator) + [IO.Path]::PathSeparator + $env:PATH
+    Write-Output "PATH=$env:PATH"
+
```
lalitb, line 160: This DLL path discovery code seems to be repeated multiple times, can we refactor it in a PowerShell function?

cmake/opentelemetry-proto.cmake
```diff
@@ -323,10 +338,24 @@ if(WITH_OTLP_GRPC)
...
+
   # Disable include-what-you-use and clang-tidy on generated code.
   set_target_properties(
-    opentelemetry_proto_grpc PROPERTIES CXX_INCLUDE_WHAT_YOU_USE ""
-                                        CXX_CLANG_TIDY "")
+    opentelemetry_proto_grpc
+    PROPERTIES CXX_INCLUDE_WHAT_YOU_USE ""
+               CXX_CLANG_TIDY ""
+               C_VISIBILITY_PRESET "default"
+               CXX_VISIBILITY_PRESET "default"
```
lalitb, line 366: Setting visibility to "default" might export more symbols than intended. Is this intended? If yes, please document as comment why do we need it.

cmake/opentelemetry-proto.cmake
```diff
@@ -323,10 +338,24 @@ if(WITH_OTLP_GRPC)
     ${LOGS_SERVICE_GRPC_PB_CPP_FILE} ${METRICS_SERVICE_GRPC_PB_CPP_FILE})
   set_target_version(opentelemetry_proto_grpc)
 
+  if(OTELCPP_PROTO_LIB_TYPE STREQUAL "SHARED_LIBRARY")
+    if(CMAKE_SYSTEM_NAME MATCHES "Windows|MinGW|WindowsStore")
+      # The codes generated by gRPC plugin do not support dll export/import
+      # declarations. To work around this, we enable WINDOWS_EXPORT_ALL_SYMBOLS
+      # property to export all symbols when building a shared library.
+      set_target_properties(opentelemetry_proto_grpc
+                            PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
lalitb, line 348: Concerned about this - this will increase the DLL size, export internal implementation details, and also potential symbol conflict - better to document these limitations as TODO to revisit in future.

cmake/tools.cmake
```diff
@@ -222,3 +222,100 @@ function(project_build_tools_patch_default_imported_config)
...
+          PARENT_SCOPE)
+    else()
+      set(${OUTPUT_VARNAME}
+          ""
+          PARENT_SCOPE)
+    endif()
+  elseif(SunPro)
+    set(${OUTPUT_VARNAME}
+        "__global"
+        PARENT_SCOPE)
```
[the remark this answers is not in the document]
lalitb: Isn't this valid comment - we are doing this check correctly at line 283?

ci/do_ci.ps1
owent: Good point, it's done.

cmake/opentelemetry-proto.cmake
owent: This is default value on Unix like system. And this is for users who change the default settings by `-fvisibility=hidden`, so it do not increase symbols than before.
Just like `WINDOWS_EXPORT_ALL_SYMBOLS` on Windows, we need something like `dllexport_decl=XXX` in protobuf, but gRPC do not support it now. we can optimize it when gRPC support this feature.

> @see https://github.com/googleapis/google-cloud-cpp/issues/5849 and https://github.com/grpc/grpc/issues/33032

cmake/opentelemetry-proto.cmake
owent: Comments have been added. We can either use the static otel-cpp libraries directly or enable `ext/src/dll`, which bundles all `.lib` files into a single DLL.  

With static otel-cpp libraries, each linked DLL/EXE gets its own copy of the symbols. Using `ext/src/dll` ensures there is only one copy of these symbols after linking, so this approach is no worse than before.

Once gRPC supports export declarations, we can further optimize this.

See also: https://github.com/googleapis/google-cloud-cpp/issues/5849 and https://github.com/grpc/grpc/issues/33032

owent:
@lalitb @ThomsonTan @dbarker 
There is another suggestion for https://github.com/open-telemetry/opentelemetry-cpp/pull/3714#discussion_r2536061020_ and https://github.com/open-telemetry/opentelemetry-cpp/pull/3714#discussion_r2536064869_ .

We currently do not support `-DCMAKE_BUILD_TYPE=ON` on Windows. A better approach would be to use a macro to define `__attribute__((__dllexport__))`, `__attribute__((__dllimport__))`, or `__attribute__((visibility("default")))` for each component and apply this macro to all exported APIs. This would allow us to build each component as a standalone DLL. By adopting this method, there would be no need to manually run `ext/src/dll/make_def.ps1` or maintain `ext/src/dll/input.src`. When building `ext/src/dll` with `-DCMAKE_BUILD_TYPE=ON`, it would only need to link all components of otel-cpp. Conversely, when `-DCMAKE_BUILD_TYPE=OFF` is used, all symbols can be exported using a single CMake function, simplifying the process of maintaining symbols in the future.

For example, we can add the following CMake utility function:

```cmake
function(project_build_tools_set_default_library_declaration DEFINITION_VARNAME)
  if(BUILD_SHARED_LIBS OR ((CMAKE_SYSTEM_NAME STREQUAL "Windows") AND DEFINED OPENTELEMETRY_BUILD_DLL))
    project_build_tools_set_shared_library_declaration(${DEFINITION_VARNAME}
                                                       ${ARGN})
  else()
    project_build_tools_set_static_library_declaration(${DEFINITION_VARNAME}
                                                       ${ARGN})
  endif()
endfunction()
```

Then use it in `opentelemetry_common` as follows:
`project_build_tools_set_default_library_declaration(OPENTELEMETRY_SDK_COMMON_API opentelemetry_common)`.

Public functions can then be declared like this:

```cpp
class GlobalLogHandler
{
public:
  OPENTELEMETRY_SDK_COMMON_API static nostd::shared_ptr<LogHandler> GetLogHandler() noexcept;

  OPENTELEMETRY_SDK_COMMON_API static void SetLogHandler(const nostd::shared_ptr<LogHandler> &eh) noexcept;

  // ...
};
```

What do you think of this approach? Should we open a new issue or discuss it further?  
It touches many files, and I can complete the changes if we agree to proceed.

ci/do_ci.ps1
```diff
@@ -151,6 +184,10 @@ switch ($action) {
     if ($exit -ne 0) {
       exit $exit
     }
+
+    $env:PATH = FindAndMergeDllPath "."
```
ThomsonTan, line 188: Is it needed to add the current dir to path? The current dir should be in the default DLL search path.

cmake/tools.cmake
```diff
@@ -222,3 +222,100 @@ function(project_build_tools_patch_default_imported_config)
     endif()
   endforeach()
 endfunction()
+
+function(project_build_tools_set_export_declaration OUTPUT_VARNAME)
+  if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang|AppleClang|Intel|XL|XLClang")
+    if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
ThomsonTan, line 228: DLL build only supports MSVC, probably mark other compilers as unsupported?

exporters/otlp/CMakeLists.txt
```diff
@@ -250,10 +250,12 @@ if(WITH_OTLP_HTTP)
...
+    PRIVATE "$<INSTALL_INTERFACE:opentelemetry_proto>"
```
ThomsonTan, line 258: Why is the opentelemetry_proto target from BUILD_INTERFACE public, but private if it is from INSTALL_INTERFACE?

cmake/tools.cmake
owent: Clang and GCC also support exporting DLL symbols. If we build the otel-cpp components as standalone libraries—rather than an all-in-one DLL, we need these declarations.

ci/do_ci.ps1
owent: On my local system, the GTest DLLs are located in build/bin/Debug, while opentelemetry_proto.dll and opentelemetry_proto_grpc.dll are in build/Debug. Other DLLs may be in additional subdirectories as well.

This script will collect all of these directories.

exporters/otlp/CMakeLists.txt
owent: Some examples and tests use the APIs from opentelemetry_proto/opentelemetry_http_client_curl, but miss the linking. Can we clean up them in another PR?

ThomsonTan (approved):
LGTM, thanks for the fix.

marcalff:
Thanks for the fix.

Looking at the patch, it contains a lot of code related to CMake scripts and Windows specific logic, which is beyond my expertise of this area, so I can not review it.

Ok to merge, based on other approvals, and the clean CI.

PR #3716, opened 2025-10-26
[SDK] Misc cleanup in attribute_utils.h

Reneg973:
Fixes # (issue)
none

## Changes
- prevent superfluous copy in construction of OwnedAttributeValue with std::string or span
- combined identical lambdas of both constructors
- use const-ref
- removed unneeded (stateless) members

Regarding the 2 constructors, IMO the pointer constructor is superfluous. Unfortunately it is being referenced already. So I'm not sure whether it's allowed to remove it.

Details:
- `const std::vector<T> copy(vals.begin(), vals.end());` this const prevents from moving
- removing stateless members reduces sizeof(class)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -60,7 +60,7 @@ jobs:
           egress-policy: audit
 
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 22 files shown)


marcalff:
@Reneg973 

See clang-format, there is an extra blank line that needs to be removed.

Will merge after that, congratulations on your first merged PR.

PR #3724, opened 2025-10-28
add validation in MetricCollector constructor and enhance related tests

Reneg973:
Fixes # (issue)
an access violation in constructor of MetricProducer when giving nullptr metric_reader.
## Changes

Need clarification first:
Is it allowed to throw exceptions in constructor at least? Currently I see a lot of classes allowing nullptr without checking early. This leads to objects having an invalid state, while member functions do nothing except spamming the log, that a required member is not set correctly.

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
+++ b/sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
@@ -43,7 +43,7 @@ class MetricCollector : public MetricProducer, public CollectorHandle
 public:
   MetricCollector(MeterContext *context,
                   std::shared_ptr<MetricReader> metric_reader,
-                  std::unique_ptr<MetricFilter> metric_filter = nullptr);
+                  std::unique_ptr<MetricFilter> metric_filter = nullptr) noexcept;
 
   ~MetricCollector() override = default;
 
--- a/sdk/src/metrics/state/metric_collector.cc
+++ b/sdk/src/metrics/state/metric_collector.cc
@@ -30,29 +30,31 @@ namespace metrics
 
 MetricCollector::MetricCollector(opentelemetry::sdk::metrics::MeterContext *context,
                                  std::shared_ptr<MetricReader> metric_reader,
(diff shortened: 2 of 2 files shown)


Reneg973:
OK, I understand you disabled exceptions. But, there is a lot of functions which may throw. Some of them even have noexcept, means it simply crashes.
Note that, under circumstances, every new could throw, every string or map or vector creation. Also, many parameters are not checked for nullptr and simply accessed. How did you plan to handle all this missing checks?
You also would need to have a custom new handler which doesn't throw. Possible, but a lot of work.

The first thing missing, if no function/even no constructor must throw, to add `noexcept` everywhere.

lalitb:
> Is it allowed to throw exceptions in constructor at least? Currently I see a lot of classes allowing nullptr without checking early. This leads to objects having an invalid state, while member functions do nothing except spamming the log, that a required member is not set correctly.

Public-facing APIs: Return real or no-op objects, shouldn't be nullptr ideally.
Internal SDK code: Assumes proper dependency injection, and nullptr shouldn't be passed in constructors. Maybe we can validate this by asserting in debug build.

But, exceptions are not thrown.

p.s btw sorry for adding copilot for review. didn't realize it's still draft.

Reneg973:
> Public-facing APIs: Return real or no-op objects, shouldn't be nullptr ideally.
> and nullptr shouldn't be passed in constructors

Can you ensure that? E.g.
```cpp
auto meter = prov->CreateMeter(...);
...->AddMeter(std::move(meter));
```
Where should the validation be? CreateMeter() could return the NoopMeter in case new fails. Should that be done? And AddMeter() could get a nullptr, aeven in Release, a check just in Debug won't be enough. Currently its not checked on output nor input.
And we don't program in an ideal system where new never fails. The chance may be low, but never 0.

Proposals to make the code safer:
- use of non-throwing new
- provide `nostd::make_unique` and `nostd::make_shared` using non-throwing new
- try/catch everything using dynamic allocations (vector/string/map...)
- provide a helper member function `bool IsValid() const noexcept` on every object which can have invalid state.

PR #3732, opened 2025-10-31
[METRICS] Add tag to AggregationConfig for aggregation type validation

### LINKED ISSUE #3731: Mismatch between AggregationConfig and AggregationType in View constructor could cause memory corruption
In the current implementation of the `View` constructor in metrics, both `AggregationType` and `AggregationConfig` are supplied by the user, which introduces a potential mismatch—for example, passing the base `AggregationConfig` when the type is `kHistogram`. Such inconsistency leads to unsafe type casting, as the system attempts to reinterpret the base configuration as a derived one, resulting in undefined behavior and possible server-side memory access violations.

```c++
View view("view_name", "description", AggregationType::kDefault, aggregation_config);
```

ThomsonTan:
Fixes #3731

## Changes

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/view/view.h
+++ b/sdk/include/opentelemetry/sdk/metrics/view/view.h
@@ -4,7 +4,6 @@
 #pragma once
 
 #include <memory>
-#include <stdexcept>
 #include <string>
 
 #include "opentelemetry/sdk/metrics/aggregation/aggregation_config.h"
--- a/sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
+++ b/sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <algorithm>
+#include <cassert>
 #include <memory>
 #include <string>
 #include <utility>
@@ -90,7 +91,8 @@ public:
 
         default:
           // Unreachable: all AggregationType enum values are handled above
-          std::abort();
+          assert(false && "Unreachable: unhandled AggregationType");
+          valid = false;
       }
 
       if (!valid)


sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
```diff
@@ -13,13 +13,23 @@ namespace sdk
...
 namespace metrics
 {
+
+enum class AggregationConfigType
+{
+  kDefault,
+  kHistogram,
+  kBase2ExponentialHistogram
+};
+
```
lalitb: Can we use the Aggression type from here - https://github.com/open-telemetry/opentelemetry-cpp/blob/b55766516d7acea524f6184103be691c996ac45e/sdk/include/opentelemetry/sdk/metrics/instruments.h#L44

sdk/include/opentelemetry/sdk/metrics/view/view.h
```diff
@@ -36,7 +37,35 @@ class View
         aggregation_type_{aggregation_type},
         aggregation_config_{aggregation_config},
         attributes_processor_{std::move(attributes_processor)}
-  {}
+  {
+    // Validate that aggregation_type and aggregation_config match
```
marcalff: Now that aggregation_config contains its proper type, can the aggregation_type parameter be removed instead ?

Reneg973:
I guess also here exceptions are not allowed.
You could also add a `noexcept` where it's missing.

sdk/include/opentelemetry/sdk/metrics/view/view.h
ThomsonTan: Probably not for now because the common `AggregationConfig` can still handle multiple aggregation types, like `kDefault`, `kSum`, etc. It is not 1:1 mapping.

ThomsonTan:
> I guess also here exceptions are not allowed. You could also add a `noexcept` where it's missing.

@Reneg973 , where do add the `noexcept`, the `View` constructor?

Reneg973:
> > I guess also here exceptions are not allowed. You could also add a `noexcept` where it's missing.
> 
> @Reneg973 , where do add the `noexcept`, the `View` constructor?

On every function, also constructor, like this:
```cpp
virtual AggregationType GetType() const noexcept { return AggregationType::kDefault; }
```

sdk/include/opentelemetry/sdk/metrics/view/view.h
```diff
@@ -36,7 +37,39 @@ class View
...
+          valid = (config_type == AggregationType::kDefault);
+          break;
+      }
+
+      if (!valid)
+      {
+#if defined(__cpp_exceptions)
+        throw std::invalid_argument("AggregationType and AggregationConfig type mismatch");
+#else
+        std::abort();
```
marcalff: So, someone defines a wrong view, and the process dies ?

ThomsonTan: If exception is not allowed, currently yes, the process dies.

sdk/include/opentelemetry/sdk/metrics/view/view.h
```diff
@@ -36,7 +37,35 @@ class View
...
+          break;
+        case AggregationType::kBase2ExponentialHistogram:
+          valid = (config_type == AggregationConfigType::kBase2ExponentialHistogram);
+          break;
+        case AggregationType::kDefault:
+        case AggregationType::kDrop:
+        case AggregationType::kLastValue:
+        case AggregationType::kSum:
+          valid = (config_type == AggregationConfigType::kDefault);
+          break;
```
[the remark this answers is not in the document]
ThomsonTan: The check is moved to `ViweRegistry::AddView`, to the deferred handling of `kDefault` logic is copied there. This is possible because `AddView` has both `AggregationConfig` and `InstrumentationSelector`.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
```diff
@@ -47,6 +49,61 @@ class ViewRegistry
...
+
+        case AggregationType::kDrop:
+        case AggregationType::kLastValue:
+        case AggregationType::kSum:
+          valid = (aggregation_config_type == AggregationType::kDefault);
+          break;
+
+        default:
+          // Unreachable: all AggregationType enum values are handled above
+          std::abort();
```
lalitb: nit - this can be a utility method say `IsDefaultAggregation(agg_type, inst_type, is_monotinic)` in default_aggregation.h, but can be done afterwards if required at other places.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
```diff
@@ -47,6 +49,61 @@ class ViewRegistry
...
 
+    // Validate parameters
+    if (!instrument_selector || !meter_selector || !view)
+    {
+#if defined(__cpp_exceptions)
+      throw std::invalid_argument("instrument_selector, meter_selector, and view cannot be null");
+#else
+      std::abort();
+#endif
+    }
```
marcalff, line 60: For every other failures in general, like illegal instrument names, the code just logs an error and ignores the instrument / view.

We should not throw exceptions or worse kill the process here, only ignore invalid configuration.

marcalff (changes requested):
For robustness, ignore invalid views / instruments / aggregations / etc.

A broken aggregation definition should not kill the application.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
ThomsonTan: Replaced `throw/abort` with internal error log and ignores it.

sdk/include/opentelemetry/sdk/metrics/view/view.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <memory>
+#include <stdexcept>
```
marcalff: Probably not needed, to remove.

sdk/include/opentelemetry/sdk/metrics/view/view_registry.h
```diff
@@ -47,6 +49,59 @@ class ViewRegistry
...
+
+        case AggregationType::kDrop:
+        case AggregationType::kLastValue:
+        case AggregationType::kSum:
+          valid = (aggregation_config_type == AggregationType::kDefault);
+          break;
+
+        default:
+          // Unreachable: all AggregationType enum values are handled above
+          std::abort();
```
marcalff: `assert(false)` instead ?

When building in release mode, should that path ever be seen, the code will fall gracefully with valid = false and an error in the log.

marcalff (commented):
LGTM, see minor cleanup.

marcalff:
@codex review

marcalff (approved):
LGTM.

marcalff:
@ThomsonTan 

The CI failure in install tests (conan) is due to a file system full on the github runner.

See:

* https://github.com/actions/runner-images/issues/13189

My take is that we still should merge this PR, the same failure exists in main already.

PR #3734, opened 2025-11-01
[CONFIGURATION] File configuration - console metric exporter

marcalff:
Follow up to https://github.com/open-telemetry/opentelemetry-cpp/issues/2481

Implement console metric exporter, per schema change:

* https://github.com/open-telemetry/opentelemetry-configuration/pull/362

## Changes

* Add new attributes to the console metric exporter

### DIFF
--- a/sdk/src/configuration/configuration_parser.cc
+++ b/sdk/src/configuration/configuration_parser.cc
@@ -44,7 +44,9 @@
 #include "opentelemetry/sdk/configuration/extension_sampler_configuration.h"
 #include "opentelemetry/sdk/configuration/extension_span_exporter_configuration.h"
 #include "opentelemetry/sdk/configuration/extension_span_processor_configuration.h"
+#include "opentelemetry/sdk/configuration/grpc_tls_configuration.h"
 #include "opentelemetry/sdk/configuration/headers_configuration.h"
+#include "opentelemetry/sdk/configuration/http_tls_configuration.h"
 #include "opentelemetry/sdk/configuration/include_exclude_configuration.h"
 #include "opentelemetry/sdk/configuration/instrument_type.h"
 #include "opentelemetry/sdk/configuration/integer_array_attribute_value_configuration.h"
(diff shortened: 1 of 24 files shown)


exporters/ostream/src/console_push_metric_builder.cc
```diff
@@ -25,10 +28,23 @@ void ConsolePushMetricBuilder::Register(opentelemetry::sdk::configuration::Regis
...
-  return OStreamMetricExporterFactory::Create();
+  sdk::metrics::AggregationTemporality aggregation_temporality;
+
+  switch (model->temporality_preference)
+  {
+    case opentelemetry::sdk::configuration::TemporalityPreference::delta:
+      aggregation_temporality = sdk::metrics::AggregationTemporality::kDelta;
+      break;
+    case opentelemetry::sdk::configuration::TemporalityPreference::cumulative:
+    case opentelemetry::sdk::configuration::TemporalityPreference::low_memory:
```
dbarker, line 41: Since the OStreamMetricExporter takes a `sdk::metrics::AggregationTemporality` which does not have a value for `kLowMemory` (that is defined for otlp exporters in `exporter::oltp::PreferredAggregationTemporality`). Should this `low_memory` value result in an configuration error/exception here?

sdk/src/configuration/configuration_parser.cc
```diff
@@ -601,11 +601,17 @@ ConfigurationParser::ParseOtlpFilePushMetricExporterConfiguration(
 
 std::unique_ptr<ConsolePushMetricExporterConfiguration>
 ConfigurationParser::ParseConsolePushMetricExporterConfiguration(
-    const std::unique_ptr<DocumentNode> & /* node */) const
+    const std::unique_ptr<DocumentNode> &node) const
 {
   auto model = std::make_unique<ConsolePushMetricExporterConfiguration>();
 
-  // FIXME-CONFIG: https://github.com/open-telemetry/opentelemetry-configuration/issues/242
+  std::string temporality_preference = node->GetString("temporality_preference", "cumulative");
```
dbarker: nit - strings `temporality_preference` and `default_histogram_aggregation` can be const

exporters/ostream/src/console_push_metric_builder.cc
marcalff: Thanks for catching that.

Looking at the spec, it turns out:
- otlp exporters have cumulative, delta and low memory temporality
- console exporters only have cumulative and delta.

I think this should be clarified at the spec first, and then in the configuration model : either reuse the same enum or define a new one.

Then opentelemetry-cpp can adjust to it.

I will print a warning in the mean time.

sdk/src/configuration/configuration_parser.cc
marcalff: Fixed, and similar in many places as well.

dbarker (approved):
Looks good. Thanks!

PR #3745, opened 2025-11-14
[SDK] Reset TraceFlags::IsSampled bit on sampler Decision::DROP

### LINKED ISSUE #3725: Child spans inherit the parent’s TraceFlags::kIsSampled bit even when the active sampler returns Decision::DROP
**Describe your environment**  
Ubuntu22.04 
OpenTelemetry-cpp v1.22.0

**Steps to reproduce**
1. Process1: Start a span, and set it to active, as parent, propogate(IsSampled=true)
2. Process2: Start a child span, with custom sampler, which returns Decision::DROP
3. Check the child span trace flag, it would be true

Per the spec, TraceFlags::kIsSampled must reflect the sampler decision. 
Dropped spans should have the sampled bit unset.

The child span always get the same `IsSampled()=true` if the parent is sampled, no matter what the custom sampler returns.

**Root Cause**
The code copies the parent’s flags (see [this line](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/trace/tracer.cc#L101)) and only sets the sampled bit when ShouldSample returns RECORD_AND_SAMPLE(see [this line](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/trace/tracer.cc#L117)). It never clears the bit for DROP/RECORD_ONLY. When a drop decision occurs, the SDK still creates a SpanContext with kIsSampled set and wraps it in a NoopSpan, so span->GetContext().IsSampled() remains true.

marcalff:
Fixes #3725

## Changes

* When a sampler returns Decision::DROP or Decision::RECORD_ONLY, clear the trace sample flag, per specs.

### DIFF
--- a/.github/workflows/cmake_install.yml
+++ b/.github/workflows/cmake_install.yml
@@ -198,6 +198,8 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           submodules: 'recursive'
+      - name: Free Disk Space
+        run: ./ci/free_disk_space.sh
       - name: Install Conan
         run: |
           python3 -m pip install pip==25.0.1
@@ -235,6 +237,8 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           submodules: 'recursive'
+      - name: Free Disk Space
+        run: ./ci/free_disk_space.sh
       - name: Install Conan
         run: |
           python3 -m pip install pip==25.0.1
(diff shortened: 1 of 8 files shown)


sdk/test/trace/tracer_test.cc
```diff
@@ -1331,3 +1373,32 @@ TEST(Tracer, SpanCleanupWithScope)
   }
   EXPECT_EQ(4, span_data->GetSpans().size());
 }
+
+TEST(Tracer, SpanSamplerDecision)
+{
+  InMemorySpanExporter *exporter              = new InMemorySpanExporter();
```
dbarker, line 1379: nit - prefer make_unique/make_shared in place of new.

PR #3762, opened 2025-11-23
[CI] upgrade to clang-tidy 20

### LINKED ISSUE #3538: Upgrade clang-tidy version used in CI
To improve test coverage we can consider upgrading clang-tidy for CI testing. The current version used in the clang-tidy workflow is 18.1.3 and the latest LLVM version available is [20.1.8](https://github.com/llvm/llvm-project/releases/tag/llvmorg-20.1.8)

dbarker:
Fixes #3538

To run clang-tidy on all header files we need to upgrade to version [20.1](https://releases.llvm.org/20.1.0/tools/clang/tools/extra/docs/clang-tidy/index.html). This version adds the `--exclude-header-filter` argument enabling analysis of all project header files excluding the `api/.../internal/absl` headers. 

## Changes

- upgrade the clang-tidy CI workflow and devcontainer to clang-tidy version 20 from version 18
- adds a python script to parse the build logs and create a markdown report on the unique warnings found

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 581
+            warning_limit: 595
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 583
+            warning_limit: 597
     env:
       CC: /usr/bin/clang-18
       CXX: /usr/bin/clang++-18


ci/do_ci.sh
```diff
@@ -341,17 +341,24 @@ elif [[ "$1" == "cmake.legacy.test" ]]; then
...
   cmake -S ${SRC_DIR} \
     -B ${BUILD_DIR} \
     -C ${SRC_DIR}/test_common/cmake/all-options-abiv2-preview.cmake  \
     "${CMAKE_OPTIONS[@]}" \
     -DWITH_OPENTRACING=OFF \
     -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" \
     -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-    -DCMAKE_CXX_CLANG_TIDY="clang-tidy;--quiet;-p;${BUILD_DIR}"
-  make -j $(nproc)
+    -DCMAKE_CXX_CLANG_TIDY="clang-tidy;--header-filter=.*/opentelemetry-cpp/.*;--exclude-header-filter=.*(internal/absl|third_party|third-party)/.*;--quiet"
```
marcalff: Is there a way to put this logic (header-filter) in the top level file .clang-tidy instead ?

It is desirable to have all the clang-tidy configuration in the same place, independent of the workflow & makefiles.

.github/workflows/clang-tidy.yaml
```diff
@@ -68,37 +67,63 @@ jobs:
...
+            -S . \
+            -B build-${{ matrix.cmake_options }} \
             -C ./test_common/cmake/${{ matrix.cmake_options }}.cmake \
             -DCMAKE_CXX_STANDARD=14 \
             -DWITH_STL=CXX14 \
             -DWITH_OPENTRACING=OFF \
             -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" \
             -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
-            -DCMAKE_CXX_CLANG_TIDY="clang-tidy;--quiet;-p;build-${{ matrix.cmake_options }}"
+            -DCMAKE_CXX_CLANG_TIDY="clang-tidy;--header-filter=.*/opentelemetry-cpp/.*;--exclude-header-filter=.*(internal/absl|third_party|third-party)/.*;--quiet"
```
marcalff: Is there a way to put this logic (header-filter) in the top level file .clang-tidy instead ?

marcalff:
@dbarker 

The generated reports (seen in CI) are very nice.

.github/workflows/clang-tidy.yaml
dbarker: There should be a way with the `HeaderFilterRegex` and `ExcludeHeaderFilterRegex` fields (see https://clang.llvm.org/extra/clang-tidy/), however I could not get these to work as expected after a few tries. I think we should use the command line options `--header-filter` and `--exclude-header-filter` for now and perhaps revisit the .clang-tidy file options later.

ci/do_ci.sh
dbarker: This should be possible, but I haven't been able to get this approach working with version 20.1.2. 

I've updated the clang-tidy github workflow to call the `ci/do_ci.sh` script so we have one place to define the header file regex.

marcalff (approved):
LGTM, great work.

PR #3767, opened 2025-11-26
fix segfault in opentelemetry::v1::context::RuntimeContext::GetCurrent () during exit() in multi-thread envs

xman1979:
Fixes # (3766)

## Changes

In a multi-threaded environment, when the process exits, this static object could get destructed while there are still running threading doing the export and therefore hitting crash.

Application could explicitly stop the threads before exit, but it is an extra constraints, and it is not always doable when opentelemetry-cpp is built as a lib.

registered in std::atexit() does not guaranteed the destruct order as well, thence we need this fix.

### DIFF
--- a/api/include/opentelemetry/context/runtime_context.h
+++ b/api/include/opentelemetry/context/runtime_context.h
@@ -178,7 +178,7 @@ private:
 
   OPENTELEMETRY_API_SINGLETON static nostd::shared_ptr<RuntimeContextStorage> &GetStorage() noexcept
   {
-    static nostd::shared_ptr<RuntimeContextStorage> context(GetDefaultStorage());
+    static thread_local nostd::shared_ptr<RuntimeContextStorage> context(GetDefaultStorage());
     return context;
   }
 };


ThomsonTan:
I think the `RuntimeContextStorage` is supposed to be global instead of thread local, and there is `ThreadLocalContextStorage` derived from the former.

xman1979:
> I think the `RuntimeContextStorage` is supposed to be global instead of thread local, and there is `ThreadLocalContextStorage` derived from the former.

yes, you are right, I see this can break SetRuntimeContextStorage(). I hope there is an better fix.

xman1979:
Drop this, for people who hits the similar issue, use this workaround to make sure to call 
opentelemetry::context::RuntimeContext::SetRuntimeContextStorage() 

```
  g_storage = nostd::shared_ptr<opentelemetry::context::ThreadLocalContextStorage>(new opentelemetry::context::ThreadLocalContextStorage());
  opentelemetry::context::RuntimeContext::SetRuntimeContextStorage(g_storage);
  ```

**Comment by Reneg973** [comment]
> In a multi-threaded environment, when the process exits, this static object could get destructed while there are still running threading doing the export and therefore hitting crash.

The question here is, HOW does the process exit? Did it forget to join all running threads before unloading?

> Drop this, for people who hits the similar issue, use this workaround to make sure to call opentelemetry::context::RuntimeContext::SetRuntimeContextStorage()
> 
> ```
>   g_storage = nostd::shared_ptr<opentelemetry::context::ThreadLocalContextStorage>(new opentelemetry::context::ThreadLocalContextStorage());
>   opentelemetry::context::RuntimeContext::SetRuntimeContextStorage(g_storage);
> ```

This won't help if the object with static memory was used. This is destroyed when the module is unloaded, but if the thread is still running -> crash

**Comment by xman1979** [comment]
> > In a multi-threaded environment, when the process exits, this static object could get destructed while there are still running threading doing the export and therefore hitting crash.
> 
> The question here is, HOW does the process exit? Did it forget to join all running threads before unloading?
> 
> > Drop this, for people who hits the similar issue, use this workaround to make sure to call opentelemetry::context::RuntimeContext::SetRuntimeContextStorage()
> > ```
> >   g_storage = nostd::shared_ptr<opentelemetry::context::ThreadLocalContextStorage>(new opentelemetry::context::ThreadLocalContextStorage());
> >   opentelemetry::context::RuntimeContext::SetRuntimeContextStorage(g_storage);
> > ```
> 
> This won't help if the object with static memory was used. This is destroyed when the module is unloaded, but if the thread is still running -> crash

The idea here is to set something global that holds the shared_ptr so the lifecycle of RuntimeContextStorage can be managed.

**Comment by Reneg973** [comment]
> The idea here is to set something global that holds the shared_ptr so the lifecycle of RuntimeContextStorage can be managed.

Yes, you extend the lifetime to be longer than your accessors (threads). This is more or less just a workaround. I just can guess your program structure, I think you're using global datasinks, which are destroyed when the module is unloaded. Try to destroy them manually at the end of main() or in a cleanup() function

**Comment by malkia** [comment]
Other libraries solves this by intentionally never destructing the object. Not ideal, but it'll work.

```
OPENTELEMETRY_API_SINGLETON
static nostd::shared_ptr<RuntimeContextStorage> &GetStorage() noexcept
{
    static auto *context = new nostd::shared_ptr<RuntimeContextStorage>(GetDefaultStorage());
    return *context;   // intentionally leaked; process teardown safe
}
```

or

```
static nostd::shared_ptr<RuntimeContextStorage> GetStorage() noexcept
{
    static nostd::shared_ptr<RuntimeContextStorage> context(
        GetDefaultStorage().release(),
        [](RuntimeContextStorage*){}  // no-op deleter
    );
    return context;
}
```

ChatGPT suggested code ^^^ - e.g. "leaky signleton", "no destructor" (of sorts) idea.

PR #3768, opened 2025-11-26
[API] fix shared_ptr self assignments

Reneg973:
see https://github.com/open-telemetry/opentelemetry-cpp/pull/3713#discussion_r2548948678

Fixes # (issue)

## Changes
prevent self assignment and self move assignment

### DIFF
--- a/.clomonitor.yml
+++ b/.clomonitor.yml
@@ -1,0 +1,6 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+# see https://github.com/cncf/clomonitor/blob/main/docs/checks.md#exemptions
+exemptions:
+  - check: artifacthub_badge
+    reason: "Artifact Hub doesn't support c++ packages"


Reneg973:
Should I remove the unit test not belonging to this PR? It increases test coverage but I'm not sure whether you want to see it in a different commit!

marcalff:
> Should I remove the unit test not belonging to this PR? It increases test coverage but I'm not sure whether you want to see it in a different commit!

Please keep the test, this is fine.

It needs to build in CI, there is a self move warning. Maybe use another variable, like:

```
ptr4 = std::move(ptr1);
```

lalitb:
> > Should I remove the unit test not belonging to this PR? It increases test coverage but I'm not sure whether you want to see it in a different commit!
> 
> Please keep the test, this is fine.
> 
> It needs to build in CI, there is a self move warning. Maybe use another variable, like:
> 
> ```
> ptr4 = std::move(ptr1);
> ```

Or if we need to test self-assignment for shared pointer. 

```cpp
ptr1 = ptr1; 
```

Reneg973:
> > It needs to build in CI, there is a self move warning. Maybe use another variable, like:

Another variable won't work, I just could use an alias (reference to ptr1), but I'm not sure it tricks all the modern compilers. It may work for now, but not anymore after compiler update. Could also be fixed by suppressing this compiler warning temporarily (if allowed).
Any tips?

marcalff:
> Any tips?

Sure, just remove the code in the unit tests that does not build, due to the self move warning.

The operator= change in the header is simple enough, it does not absolutely require unit test coverage if this brings complications.

PR #3769, opened 2025-11-26
[CI] Update markdownlint-cli version to 0.46.0

thompson-tomo:
## Changes

This bumps markdownlint-cli to the latest version which appears not to be managed via dependabot due to direct install via a command.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -4,5 +4,9 @@
     {
         "code_blocks": false,
         "tables": false
+    },
+    "MD059":
+    {
+        "prohibited_texts": []
     }
 }


.markdownlint.json
```diff
@@ -4,5 +4,9 @@
     {
         "code_blocks": false,
         "tables": false
+    },
+    "MD059":
+    {
+        "prohibited_texts": []
     }
```
marcalff, line 11: Fair enough, cleanup should happen in different PR, not related to the upgrade.

thompson-tomo: Agree, can be done seperately.

Issue #3652, opened 2025-09-23
Implement entity propagation


gaborcsardi:
New in otel 1.49.0, AFAICT: 
https://opentelemetry.io/docs/specs/otel/entities/entity-propagation/
https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/entities/0264-resource-and-entities.md

yashikabadaya:
Hi, can I take up this issue?

saphal1998:
Hi! Can I pick this up?

saphal1998:
Since `EnvEntityDetector` was merged under `resource_detectors`, I’m assuming it is intentionally opt-in rather than automatically run by default `Resource::Create()`.

Would a useful follow-up be to add documentation/examples showing how users should link `opentelemetry_resource_detectors`, call `EnvEntityDetector().Detect()`, and merge the detected resource into their SDK resource? Is this issue to add [documentation](https://opentelemetry.io/docs/specs/otel/entities/entity-propagation/#enventitydetector)

I’m unsure whether any code changes are expected beyond documentation, so guidance would be appreciated.

shashankxrm:
Hi @dbarker , I was looking into this issue and noticed that EnvEntityDetector / OTEL_ENTITIES parsing was already implemented in #3795 and is now available as an opt-in detector.

Is there still remaining work you'd like to see for #3652, or is it effectively complete? If there is remaining work, could you point me toward the intended scope? I don't want to duplicate #3795 or start on the larger Entity model work without confirmation.

dbarker:
Hi @shashankxrm,  the EnvEntity Detector is mostly complete and now that version 1.60.0  of the spec is released (this added detail on the Entity Data Model and use in the Resource SDK) we should continue the implementation.  Please see links below. 

The next step will be to add Entities to the SDK Resource library. Once entities are supported by the Resource class, we can populate them in the detectors and update the OTLP exporters to fill in the protobuf messages appropriately. I will create tracking issues for these steps soon. 

- https://opentelemetry.io/docs/specs/otel/resource/data-model/
- https://opentelemetry.io/docs/specs/otel/resource/sdk/#create
- https://opentelemetry.io/docs/specs/otel/resource/sdk/#merge-behavior-without-entities

shashankxrm:
[...]

PR #3795, opened 2025-12-19
[SDK] Implement EnvEntityDetector for OTEL_ENTITIES parsing (#3652)

Hasv07:
Related #3652
Changes:
- Add EnvEntityDetector class to parse OTEL_ENTITIES environment variable
- Parse entity format: type{id_attrs}[desc_attrs]@schema_url
- Support percent-encoding for reserved characters
- Handle duplicate entities, conflicting attributes, and malformed input per spec
- Include basic test coverage

Details:
- ```EnvEntityDetector::Detect()``` reads OTEL_ENTITIES and returns Resource with parsed attributes
- ```ParseEntities()``` splits input by semicolons and parses each entity definition
- ```ParseSingleEntity()``` extracts type, id_attrs, desc_attrs, and schema_url from entity string
- ```ParseKeyValueList()``` parses comma-separated key=value pairs
- ```PercentDecode()``` decodes percent-encoded values
- ```BuildEntityIdentityKey()``` creates stable key for duplicate detection
-    ```Error handling``` malformed entities skipped, duplicates use last occurrence, conflicts log warnings

Implements entity propagation spec:
https://opentelemetry.io/docs/specs/otel/entities/entity-propagation/

### DIFF
--- a/resource_detectors/test/env_entity_detector_test.cc
+++ b/resource_detectors/test/env_entity_detector_test.cc
@@ -4,9 +4,11 @@
 #include <gtest/gtest.h>
 #include <cstdlib>
 #include <string>
+#include <unordered_map>
 
 #include "opentelemetry/nostd/variant.h"
 #include "opentelemetry/resource_detectors/env_entity_detector.h"
+#include "opentelemetry/sdk/resource/resource.h"
 
 #if defined(_MSC_VER)
 #  include "opentelemetry/sdk/common/env_variables.h"


sdk/include/opentelemetry/sdk/resource/resource_detector.h
```diff
@@ -39,6 +39,17 @@ class OTELResourceDetector : public ResourceDetector
   Resource Detect() noexcept override;
 };
 
+/**
+ * EnvEntityDetector detects entities defined in the OTEL_ENTITIES environment
+ * variable as specified in the Entity Propagation spec:
+ * https://opentelemetry.io/docs/specs/otel/entities/entity-propagation/
+ */
+class EnvEntityDetector : public ResourceDetector
```
dbarker: Please move the new detector to the `opentelemetry-cpp/resource_detectors` folder and component. The EnvEntityDetector should have its own .h and .cc files similar to the other resource detectors.

sdk/test/resource/resource_test.cc
```diff
@@ -292,3 +292,417 @@ TEST(ResourceTest, DerivedResourceDetector)
   EXPECT_EQ(resource.GetSchemaURL(), detector.schema_url);
   EXPECT_TRUE(received_attributes.find("key") != received_attributes.end());
 }
+
```
dbarker: Please move the EnvEntityDetector tests to a new test file in the 
[resource_detectors/test](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/resource_detectors/test) folder similar to the other resource detectors.

sdk/src/resource/env_entity_detector.cc
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+  // Optional schema URL: '@...'
+  if (cursor < entity_str.size() && entity_str[cursor] == '@')
+  {
+    out.schema_url =
+        std::string{opentelemetry::common::StringUtil::Trim(entity_str.substr(cursor + 1))};
+
+    // TODO: Use a proper Schema URL validator when available.
+    if (out.schema_url.empty() || out.schema_url.find("://") == std::string::npos)
```
lalitb: I see the TODO mentions that schema validation needs work, but right now it's too loose - it accepts anything with `://` in it. We should at least check that the scheme is `http` or `https` like the spec requires.

sdk/src/resource/env_entity_detector.cc
```diff
@@ -0,0 +1,336 @@
...
+  std::string key = type + "|";
+  for (size_t i = 0; i < items.size(); ++i)
+  {
+    if (i > 0)
+    {
+      key += ",";
+    }
+    key += items[i]->first;
+    key += "=";
+    key += nostd::get<std::string>(items[i]->second);
```
lalitb: Since `id_attrs` uses `OwnedAttributeValue` which can hold int, bool, double, and other types - not just strings - using `nostd::get<std::string>` could throw if the type doesn't match. Use `nostd::get_if` or `nostd::visit` for safer type handling.

sdk/src/resource/env_entity_detector.cc
marcalff: From:
https://opentelemetry.io/docs/specs/otel/entities/entity-propagation/#validation-requirements

> Schema URL, if present, MUST be a valid URI

so this should not be restricted to `http` or `https` only.

But I agree, this probably needs revision, as one of the examples given:

```
# Multiple entities with schema URL
OTEL_ENTITIES="service{service.name=my-app,service.instance.id=instance-1}[service.version=1.0.0]@/schemas/1.21.0;host{host.id=host-123}[host.name=web-server-01]"
```

does not even use a scheme with `://`.

sdk/include/opentelemetry/sdk/resource/resource_detector.h
Hasv07: Done, thanks.

sdk/test/resource/resource_test.cc
Hasv07: Done, thanks.

sdk/src/resource/env_entity_detector.cc
Hasv07: I did it for now to validate general URIs. Stricter schema URI checks can be updated when new guidance is available. Thanks

sdk/src/resource/env_entity_detector.cc
Hasv07: Thanks for pointing this out

PR #3801, opened 2026-01-07
[EXPORTER] add OTLP HTTP metric exporter benchmark

ishan0803:
Introduce a Google Benchmark-based performance benchmark for the OTLP HTTP metric exporter, covering both binary (protobuf) and JSON encodings.

Key changes:
- Add `otlp_http_metric_exporter_benchmark` target under `WITH_BENCHMARK && WITH_OTLP_HTTP`.
- Benchmark `OtlpHttpMetricExporter::Export()` across increasing numbers of metric data points (1 → 10,000).
- Measure export cost for both binary and JSON content types.
- Disable retries and logging to reduce noise and isolate exporter overhead.

This benchmark provides a baseline for evaluating export performance and future optimizations in the OTLP HTTP metric exporter.

Fixes # (issue)

## Changes

### DIFF
--- a/exporters/otlp/benchmark/otlp_http_metric_exporter_benchmark.cc
+++ b/exporters/otlp/benchmark/otlp_http_metric_exporter_benchmark.cc
@@ -1,0 +1,128 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <benchmark/benchmark.h>
+
+#include <chrono>
+#include <cstddef>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "opentelemetry/exporters/otlp/otlp_http.h"
+#include "opentelemetry/exporters/otlp/otlp_http_metric_exporter.h"
+#include "opentelemetry/exporters/otlp/otlp_http_metric_exporter_options.h"
+
+#include "opentelemetry/nostd/string_view.h"
+
+#include "opentelemetry/sdk/common/attribute_utils.h"
+#include "opentelemetry/sdk/common/global_log_handler.h"
+
+#include "opentelemetry/sdk/instrumentationscope/instrumentation_scope.h"
+
+#include "opentelemetry/sdk/metrics/data/metric_data.h"
+#include "opentelemetry/sdk/metrics/data/point_data.h"
+#include "opentelemetry/sdk/metrics/export/metric_producer.h"
+#include "opentelemetry/sdk/metrics/instruments.h"
+#include "opentelemetry/sdk/resource/resource.h"
+
+using InstrumentationScope = opentelemetry::sdk::instrumentationscope::InstrumentationScope;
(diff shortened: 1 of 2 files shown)


ishan0803:
@marcalff I’ve fixed the formatting / IWYU issues flagged by CI.
Could you please rerun CI when you get a chance?

exporters/otlp/benchmark/otlp_http_metric_exporter_benchmark.cc
```diff
@@ -0,0 +1,128 @@
...
+  scope_metrics.metric_data_.push_back(metric);
+  rm.scope_metric_data_.push_back(scope_metrics);
+  return rm;
+}
+
+// Benchmark Export() using Binary encoding
+static void BM_OtlpHttpMetricExporter_Export_Binary(benchmark::State &state)
+{
+  OtlpHttpMetricExporterOptions opts;
+  opts.url                       = "http://localhost:4318/v1/metrics";
```
lalitb, line 73: Seems this benchmark attempts real HTTP calls that fail immediately with connection refused. Even a failed connection attempt adds ~10-100μs of overhead, which becomes significant when serialization itself takes only nanoseconds (especially for small payloads).

See how [gRPC benchmark](https://github.com/open-telemetry/opentelemetry-cpp/blob/705914b8d238d107e58c40f8385a39e753a3c6d1/exporters/otlp/test/otlp_grpc_exporter_benchmark.cc#L50)  avoid this be mocking the network layer entirely. Something we can try here ?

ishan0803: Agreed, even a failed HTTP attempt adds non-trivial overhead for smaller payload sizes, where serialization itself is very cheap.

Unlike gRPC, the HTTP metric exporter doesn’t currently expose a public way to inject a no-send transport. The only existing injection path goes through test-only peers, and I avoided reusing that here since this is a benchmark target and not built with test permissions.

Longer term, possible options could be (1) making the HTTP client injectable via a public API, (2) reusing the existing test peer, or (3) introducing a dedicated benchmark peer. But all of these require changes to exporter headers or test/benchmark boundaries and felt out of scope for this PR. For now, this benchmark measures `Export()` through the public API with retries and logging disabled.

lalitb: Thanks for clarifying. Agree with the constraints. For future, reusing the existing test peer infrastructure might be the most practical option, but that's not necessary for this PR to be valuable. I’m okay with landing this as a first baseline and revisiting injection options in a follow-up.

marcalff:
Not tested, but if I understand correctly, one can:

* deploy locally a peer (opentelemetry-collector for example) on port 4318
* use this benchmark to send traffic to the peer

so this is useful already in a dev environment, independently of CI, with zero code changes.

Ok to merge.

marcalff (approved):

Issue #3806, opened 2026-01-09
Question: should invalid OTEL_BLRP_SCHEDULE_DELAY values be covered by tests?


MPSAJID:
Hi maintainers,

I’m new to contributing to OpenTelemetry C++ and have been reading through the `BatchLogRecordProcessorOptions` env var handling while getting familiar with the codebase.

I noticed that invalid values for `OTEL_BLRP_SCHEDULE_DELAY` (e.g. non-parsable strings) are logged with a warning and fall back to the default value (1000ms). This behavior makes sense, but I didn’t see a test explicitly covering the invalid-value case.

Before opening a PR, I wanted to ask: would a small test documenting this fallback behavior be welcome, or is this something you prefer to leave implicit via the shared env parsing helper?

Happy to add a test if it’s useful, or leave it as-is if this is intentional.

Thanks!

marcalff:
@MPSAJID 

Adding more unit tests to increase coverage is always good, it reduces the risk to have undetected behavior changes.

A small test for `OTEL_BLRP_SCHEDULE_DELAY` will be welcome, feel free to contribute a PR.

If searching for ideas of other things to do, please take a look at #3068.

MPSAJID:
Thanks for the confirmation!

I’ve opened a PR adding a unit test to cover the invalid-value fallback behavior for OTEL_BLRP_SCHEDULE_DELAY:
https://github.com/open-telemetry/opentelemetry-cpp/pull/3811

Happy to make any adjustments if needed.

marcalff:
Fixed by #3812, closing.

TheMenisFactor:
What are you talking about man??

PR #3811, opened 2026-01-16
[TEST] add test for invalid OTEL_BLRP_SCHEDULE_DELAY env value

MPSAJID:
This PR adds a unit test covering the behavior when OTEL_BLRP_SCHEDULE_DELAY
is set to an invalid value.

Invalid values are logged with a warning and the default schedule delay
(1000ms) is used. The new test documents and locks in this fallback behavior
to reduce the risk of future regressions.

Related to #3806 .

### DIFF
--- a/sdk/test/logs/batch_log_record_processor_test.cc
+++ b/sdk/test/logs/batch_log_record_processor_test.cc
@@ -396,6 +396,17 @@ TEST_F(BatchLogRecordProcessorTest, TestScheduleDelayFromEnv)
   unsetenv("OTEL_BLRP_SCHEDULE_DELAY");
 }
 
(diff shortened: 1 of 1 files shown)


lalitb:
LGTM. This covers the non-parseable string case. If the intent is to validate overall robustness of `OTEL_BLRP_SCHEDULE_DELAY`, it might also be worth adding cases for negative values, zero, overflow, malformed units, or an empty string.

lalitb (approved):

MPSAJID:
Thanks for the review!

Agreed ,those are good additional cases to cover. I’ll follow up with a separate PR to extend coverage for other invalid values.

PR #3824, opened 2026-01-21
[ETW] Fix tracer initialization order

### LINKED ISSUE #3821: ETW Exporter: Tracer::isClosed_ is always true
I believe the initialization of isClosed_ is incorrect. It remains true, which prevents the ETW provider from ever being closed.

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h#L417-L427

To reproduce:
1. Run the Geneva sample application. https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/exporters/geneva-trace/example
2. Set a breakpoint in CloseWithMicroseconds().

Observed behavior:
The breakpoint is never hit because isClosed_ is always true, so the provider is never closed.

```C++
  void CloseWithMicroseconds(uint64_t) noexcept override
  {
    // Close once only
    if (!isClosed_.exchange(true))
    {
      etwProvider().close(provHandle);
    }
  }
```

You can also set a breakpoint inside the constructor and check the value of isClosed_. On my Windows 11 machine, it's always True. The constructor initialization list (initProvHandle()) runs first, then the declaration initialization, which is causing isClosed_ always True.

```C++
public:
  /**
   * @brief Tracer constructor
   * @param parent Parent TraceProvider
   * @param providerId ProviderId - Name or GUID
   * @param encoding ETW encoding format to use.
   */
  Tracer(etw::TracerProvider &parent,
         nostd::string_view providerId     = "",
         ETWProvider::EventFormat encoding = ETWProvider::EventFormat::ETW_MANIFEST)
      : opentelemetry::trace::Tracer(),
        tracerProvider_(parent),
        provId(providerId.data(), providerId.size()),
        encoding(encoding),
        provHandle(initProvHandle())
  {
    traceId_ = GetIdGenerator(tracerProvider_).GenerateTraceId();
  }
```

beep-boopp:
## Description
Fixes #3821

The `ETWTracer` was incorrectly starting in a closed state due to member initialization order. 

`isClosed_` was declared after `provHandle`, so its default initializer (`{true}`) was running *after* `provHandle`'s constructor had already set it to `false`. This overwrote the correct state.

I moved `isClosed_` before `provHandle` to ensure it initializes first.

## Testing
Added `ConstructorInitializesToOpenState` to `etw_tracer_test.cc`. 
I added a friend declaration in the header to verify that `isClosed_` is correctly `false` immediately after the tracer is instantiated.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have added tests that prove my fix is effective

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           submodules: 'recursive'
       - name: Mount Bazel Cache
@@ -59,7 +59,7 @@ jobs:
         with:
           egress-policy: audit
 
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # main March 2025
         with:
           name: benchmark_results
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 #    name: CMake test arm64 (with modern protobuf,grpc and abseil)
 #    runs-on: actuated-arm64-4cpu-16gb
 #    steps:
-#    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+#    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 #      with:
 #        submodules: 'recursive'
 #    - name: setup
@@ -42,7 +42,7 @@ jobs:
       with:
         egress-policy: audit
 
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'recursive'
     - name: setup
@@ -65,7 +65,7 @@ jobs:
       uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
       with:
         egress-policy: audit
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'false'
     - name: checkout googletest
@@ -95,7 +95,7 @@ jobs:
       uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
       with:
         egress-policy: audit
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'recursive'
     - name: setup
@@ -129,7 +129,7 @@ jobs:
       uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
       with:
         egress-policy: audit
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'recursive'
     - name: setup
@@ -163,7 +163,7 @@ jobs:
       uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
       with:
         egress-policy: audit
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'recursive'
     - name: setup
@@ -197,7 +197,7 @@ jobs:
       uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
       with:
         egress-policy: audit
-    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         submodules: 'recursive'
     - name: setup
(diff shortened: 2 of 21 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.sourceDirectory": "/wsl.localhost/Ubuntu/home/percy/opentelemetry-cpp"
+}
```
lalitb: seems to be incorrectly checked in?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -182,11 +185,19 @@ class Tracer : public opentelemetry::trace::Tracer,
   /**
    * @brief Provider Handle
    */
-  ETWProvider::Handle &provHandle;
+  //change made in pr
```
lalitb: nit -this line seems unnecessary.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -182,11 +185,19 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+  // ETWProvider::Handle &provHandle;
 
-  opentelemetry::trace::TraceId traceId_;
+  // opentelemetry::trace::TraceId traceId_;
+
+  // std::atomic<bool> isClosed_{true};
 
-  std::atomic<bool> isClosed_{true};
+  // Fix: Declare isClosed_ before provHandle so it initializes to true 
+  // *before* initProvHandle() runs.
```
lalitb: please remove all the above comments - we don't need PR change history as part of comment.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -182,13 +185,14 @@ class Tracer : public opentelemetry::trace::Tracer,
...
+  // runs its initialization logic.
+  std::atomic<bool> isClosed_{true};
   ETWProvider::Handle &provHandle;
-
   opentelemetry::trace::TraceId traceId_;
 
-  std::atomic<bool> isClosed_{true};
-
   /**
+   * @brief ETWProvider is a singleton that aggregates all ETW writes.
```
ThomsonTan: Is this line duplicated?

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -182,13 +185,14 @@ class Tracer : public opentelemetry::trace::Tracer,
   /**
    * @brief Provider Handle
    */
```
ThomsonTan, line 189: Move the above 3 lines comment with `provHandle`?

ThomsonTan:
@beep-boopp thanks for the fix. Please also address the CI issue and comments, looks good otherwise.

beep-boopp:
> @beep-boopp thanks for the fix. Please also address the CI issue and comments, looks good otherwise.

I pushed a new commit, waiting for approval. The CI shows 'Awaiting Approval' since this is my first contribution, but the build should be green once approved. Ready to merge. Once again thanks for all the help

ThomsonTan:
> > @beep-boopp thanks for the fix. Please also address the CI issue and comments, looks good otherwise.
> 
> I pushed a new commit, waiting for approval. The CI shows 'Awaiting Approval' since this is my first contribution, but the build should be green once approved. Ready to merge. Once again thanks for all the help

I approved the CI. Seems more code format needed.

beep-boopp:
> > > @beep-boopp thanks for the fix. Please also address the CI issue and comments, looks good otherwise.
> > 
> > 
> > I pushed a new commit, waiting for approval. The CI shows 'Awaiting Approval' since this is my first contribution, but the build should be green once approved. Ready to merge. Once again thanks for all the help
> 
> I approved the CI. Seems more code format needed.

Thanks for the approval! I actually just force-pushed a clearer fix (switched to a getter instead of friend class and squashed the old commits). The formatting should be solid now too, but let me know if the CI catches anything else .

beep-boopp:
> Thanks for the patch.
> 
> The test looks good, but I disagree with the fix in etw_tracer.h
> 
> A better solution is to remove method `initProvHandle()` entirely: calling a method from the constructor was a bad idea in the first place.

Thanks for the clarification , I see where you’re coming from, and I agree that calling a helper like initProvHandle() from the constructor isn’t ideal.
I’ll refactor this to remove initProvHandle() entirely and move the initialization directly into the constructor, establishing the invariants in one place.
I’ll also ensure the update complies with the project’s clang-format rules so the CI / Format check passes and incorporate both changes together in the next commit

ThomsonTan:
> Thanks for the patch.
> 
> The test looks good, but I disagree with the fix in etw_tracer.h
> 
> A better solution is to remove method `initProvHandle()` entirely: calling a method from the constructor was a bad idea in the first place.

Replace the call `initProvHandle()` by some other function call like the inner function `etwProvider().open()` should be the same effect? I suggest the we fix the the issue with the minimum change and do the refactor later as we revisit that in broader scope on similar issues.

lalitb:
> I suggest the we fix the the issue with the minimum change and do the refactor later as we revisit that in broader scope on similar issues.

+1

beep-boopp:
> > Thanks for the patch.
> > The test looks good, but I disagree with the fix in etw_tracer.h
> > A better solution is to remove method `initProvHandle()` entirely: calling a method from the constructor was a bad idea in the first place.
> 
> Replace the call `initProvHandle()` by some other function call like the inner function `etwProvider().open()` should be the same effect? I suggest the we fix the the issue with the minimum change and do the refactor later as we revisit that in broader scope on similar issues.

I believe my latest commit already does exactly this: I removed initProvHandle() entirely and replaced the call in the constructor with a direct call to etwProvider().open(provId, encoding), keeping the change minimal.
Please let me know if the current implementation matches what you had in mind.

beep-boopp:
@lalitb  Thanks for the help with the updates.
Just to confirm the current state: I have implemented the requested change by removing initProvHandle() entirely and inlining the logic into the constructor to keep the fix minimal.
The latest build failure appears to be an infrastructure issue (HTTP 500 / Timeout during Windows dependency installation) rather than a code error. Could you please re-trigger the CI?
If you spot any other adjustments needed, just let me know.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -397,17 +404,6 @@ class Tracer : public opentelemetry::trace::Tracer,
...
-   * @brief Init a reference to etw::ProviderHandle
-   * @return Provider Handle
-   */
-  ETWProvider::Handle &initProvHandle()
-  {
-    isClosed_ = false;
-    return etwProvider().open(provId, encoding);
-  }
-
-public:
```
marcalff, line 410: Removed the `public:` declaration for the constructor ?

This breaks the build, see CI.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -179,15 +184,17 @@ class Tracer : public opentelemetry::trace::Tracer,
    */
   ETWProvider::EventFormat encoding;
 
+  // Order matters: isClosed_ must initialize to true BEFORE provHandle
+  // runs its initialization logic.
+  std::atomic<bool> isClosed_{true};
```
marcalff: There is no need to more the declaration around for `isClosed_`, please revert this change back to the original code.

beep-boopp:
> See build breaks in CI.

@marcalff Thanks for the detailed review! Good catch on the missing public: modifier, I definitely missed that while cleaning up the helper function.
I’ve restored the access modifier and reverted isClosed_ to its original position as requested. It should be good to go now.

beep-boopp:
Apologies for the extra CI noise on this PR and new commit. I noticed the Windows Bazel build failed due to an MSVC-specific namespace resolution error (C2653) in the new test case. I've updated it to use the fully qualified namespace opentelemetry::exporter::etw:: to ensure compatibility across all builders. Ready for another look whenever you have the time

PR #3825, opened 2026-01-22
[BAZEL] Add linkstatic = True to http_client_curl

aaylward:
Fixes # (issue)

## Context
When consuming opentelemetry-cpp 1.24.0 via Bazel 9 using toolchains_llvm on linux, I ran into link errors like:
```
cc_wrapper.sh failed: error executing CppLink command (from cc_test rule target //cpp/example_meerkat:example_meerkat_unit_test) external/toolchains_llvm++llvm+llvm_toolchain/bin/cc_wrapper.sh @bazel-out/k8-opt/bin/cpp/example_meerkat/example_meerkat_unit_test-0.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ld.lld: error: undefined reference: curl_global_init
>>> referenced by bazel-out/k8-opt/bin/_solib_k8/libexternal_Sopentelemetry-cpp+_Sext_Ssrc_Shttp_Sclient_Scurl_Slibhttp_Uclient_Ucurl.so (disallowed by --no-allow-shlib-undefined)
```
My project built fine on mac, and Claude suggested that the problem is that opentelemetry-cpp's `http_client_curl` isn't statically linking against curl. If I'm reading it right, the recommendation in [library-distribution.md](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/docs/library-distribution.md#recommendations) seems to suggest linking deps statically as well.

Please let me know if there's a better way to solve this problem.

## Changes

* Add `linkstatic = True` to the `http_client_curl` cc_library target

This ensures the curl HTTP client is statically linked when used as a dependency, matching (what I think is) the default CMake behavior.

## Testing?
 - [x] bazel test //ext/test/http:curl_http_test
 - [x] built my project against this patch in my [bcr fork](https://github.com/bazelbuild/bazel-central-registry/compare/main...aaylward:bazel-central-registry:otel_cpp_bazel_8_9)

## Significant changes require:

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -17,9 +17,9 @@ build --copt -DGRPC_BAZEL_BUILD
 build:windows --dynamic_mode=off
 
 # Set minimum supported C++ version
-build:macos --host_cxxopt=-std=c++14 --cxxopt=-std=c++14
-build:linux --host_cxxopt=-std=c++14 --cxxopt=-std=c++14
-build:windows --host_cxxopt=/std:c++14 --cxxopt=/std:c++14
+build:macos --host_cxxopt=-std=c++17 --cxxopt=-std=c++17
+build:linux --host_cxxopt=-std=c++17 --cxxopt=-std=c++17
+build:windows --host_cxxopt=/std:c++17 --cxxopt=/std:c++17
 
 # --config=asan : Address Sanitizer.
 common:asan --copt -DADDRESS_SANITIZER
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -2,13 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0
 matrix:
   platform: ["debian10", "macos", "ubuntu2004", "windows"]
-  bazel: ["7.x", "8.x", "9.*"]
+  bazel: ["7.x", "8.x", "9.x"]
 tasks:
   verify_targets:
     platform: ${{ platform }}
     bazel: ${{ bazel }}
     build_flags:
-    - '--cxxopt=-std=c++14'
-    - '--host_cxxopt=-std=c++14'
+    - '--cxxopt=-std=c++17'
+    - '--host_cxxopt=-std=c++17'
     build_targets:
     - '@opentelemetry-cpp//api'
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -20,6 +20,7 @@ bazel_dep(name = "prometheus-cpp", version = "1.3.0", repo_name = "com_github_ju
 bazel_dep(name = "protobuf", version = "29.0", repo_name = "com_google_protobuf")
 bazel_dep(name = "rapidyaml", version = "0.9.0")
 bazel_dep(name = "rules_cc", version = "0.2.9")
+bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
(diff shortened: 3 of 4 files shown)


malkia:
Just curious, what is your "effective" `--dynamic_mode` setting - e.g. https://bazel.build/docs/user-manual#dynamic-mode

aaylward:
@malkia  I don't have it set in my bazelrc, and I believe the default for linkstatic is [False for cc_library and cc_test](https://bazel.build/reference/be/c-cpp#cc_test). My failing targets were cc_test targets.

malkia:
Please try with `--dynamic_mode=off` - bazel, for sake of speed, preffers shared (.so/.dylib) libraries, even if these were meant as static - e.g. more sharing at binary level, faster linking time, great for tests. I believe binaries in dbg/fastbuild are also build this way.

But yes, this would probably make it slower for development / CI - so your change makes sense after all.

aaylward:
`--dynamic_mode=off` worked with 1.24.0 without my patch. What do you think? Still worth merging?

malkia:
I think it's still worth merging, because the default is to have this off - I was simply pointing out bazel's behavior that triggers it, but I'm in no way authority on this. The patch would make it work without forcing people to set this flag. So thanks for that!

aaylward:
Thanks for talking it out @malkia ! Can I leave the merge to you once someone from cpp-approvers has approved?

marcalff (changes requested):
Thanks for the patch.

Based on the BCR downstream change, should this be added in MODULE.bazel ?

```
bazel_dep(name = "rules_foreign_cc", version = "0.15.1")
```

aaylward:
@marcalff Thanks for pointing that out. I looked into that a little bit more, and it looks like an older rules_foreign_cc is being pulled in transitively via [google_benchmark 1.8.4](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/google_benchmark/1.8.4/MODULE.bazel#L8). Latest google_benchmark ([1.9.5](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/google_benchmark/1.9.5/MODULE.bazel)) doesn't depend on rules_foreign_cc, so it may be preferable to update google_benchmark rather than adding rules_foreign_cc directly.

`bazel mod graph` shows that google_benchmark@1.9.5 removes the rules_foreign_cc dep, but I haven't tried out the build with 1.9.5 yet. I can check it out later today.

aaylward:
Would bumping the google_benchmark version in MODULE.bazel require CMake changes to keep the build systems in sync?

Do you prefer adding a direct dep on rules_foreign_cc to updating google_benchmark in this change?

marcalff:
Thanks for the investigations.

> Do you prefer adding a direct dep on rules_foreign_cc

Yes.

The goal is to have a MODULE.bazel in opentelemetry-cpp that is identical to the one used in BCR, so that BCR does not need to carry extra patches.

Not an expert in the bazel subtleties, but I assume that if @keith approved this in https://github.com/bazelbuild/bazel-central-registry/pull/7261, it must be the correct way to go.

aaylward:
Sounds good to me. The patch in bcr also needed 
```
    - '--cxxopt=-std=c++17'
    - '--host_cxxopt=-std=c++17'
```
so I added that as well.

PR #3829, opened 2026-01-26
[INSTALL] Add pkg-config support for exporters

### LINKED ISSUE #3787: Exporters are missing opentelemetry_add_pkgconfig instruction
Neither of exporters' CMakeLists.txt have instruction to install its pckconfig

Similar instructor is missing
```
if(OPENTELEMETRY_INSTALL)
  opentelemetry_add_pkgconfig(
    api "OpenTelemetry API"
    "A header-only library to support instrumentation with OpenTelemetry."
    "${TARGET_DEPS}")
endif()
``` 
So it is impossible to use PkgConfig to add exporter libraries to the application.

perhapsmaple:
Fixes #3787 

## Changes

Adds pkg-config support for the following exporters exporters:
- in-memory exporter
- ostream exporter
- otlp exporter
- elasticsearch exporter
- prometheus exporter
- zipkin exporter
- etw exporter (actually required?)


exporters/etw/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+include(${PROJECT_SOURCE_DIR}/cmake/pkgconfig.cmake)
```
ThomsonTan: Is pkgconfig needed here as the ETW exporter is Windows only?

exporters/etw/CMakeLists.txt
```diff
@@ -29,6 +31,13 @@ otel_add_component(
   PATTERN
   "*.h")
 
+if(OPENTELEMETRY_INSTALL)
```
ThomsonTan: Check the installation for Windows.
```suggestion
if(OPENTELEMETRY_INSTALL AND WIN32)
```

exporters/etw/CMakeLists.txt
perhapsmaple: The only scenario where this might still be useful is when building with MinGW. I don’t believe this is something we officially support (?), so I’m okay with removing it unless you think otherwise.

owent:
Could we call `opentelemetry_add_pkgconfig()` from `otel_install_components()` so we don’t have to invoke it separately for each component?

perhapsmaple:
@owent I have taken an example with the ostream exporter:

 We have 2 components but need 6 pkgconfig files:

Components:
```
otel_add_component(
  COMPONENT exporters_ostream
  TARGETS
    opentelemetry_exporter_ostream_span      # needs .pc file
    opentelemetry_exporter_ostream_metrics   # needs .pc file
    opentelemetry_exporter_ostream_logs      # needs .pc file
  ...
)
```

```
otel_add_component(
  COMPONENT exporters_ostream_builder
  TARGETS
    opentelemetry_exporter_ostream_span_builder     -> needs .pc file
    opentelemetry_exporter_ostream_metrics_builder  -> needs .pc file
    opentelemetry_exporter_ostream_logs_builder     -> needs .pc file
  ...
)
```

Pkgconfig:
```
opentelemetry_add_pkgconfig(
  exporter_ostream_span
  "OpenTelemetry Exporter - OStream Span"
  "Exports trace spans to output stream."
  "opentelemetry_trace")  -> Span-specific dependency
```

```
opentelemetry_add_pkgconfig(
  exporter_ostream_metrics
  "OpenTelemetry Exporter - OStream Metrics"
  "Exports metrics to output stream."
  "opentelemetry_metrics")  -> Metrics-specific dependency
```

```
opentelemetry_add_pkgconfig(
  exporter_ostream_logs
  "OpenTelemetry Exporter - OStream Logs"
  "Exports log records to output stream."
  "opentelemetry_logs")  -> Logs-specific dependency
```

We would need to specify a description along with dependencies for each target when calling `otel_add_component`. From my point of view, it's better to not consolidate these functions. What do you think?

lalitb:
Thanks for the PR.

Could you also update `ci/verify_packages.sh` to validate the new exporter `pkgconfig` files? Without this, we can't verify the changes are working correctly in CI.
Since not all exporters are built in every CI configuration (depending on flags like `WITH_OTLP_GRPC`, `WITH_PROMETHEUS`, etc.), you could use `pkg-config --exists` to check if the package is present before validating it. This way it validates packages that were built while skipping those not included in the build configuration.

exporters/etw/CMakeLists.txt
lalitb: I believe we can remove the pkgconfig for ETW exporter - I don't think we would ever tested for mingw/msys2, or want to support them yet.

exporters/otlp/CMakeLists.txt
```diff
@@ -232,6 +248,48 @@ if(WITH_OTLP_GRPC)
...
+
+    opentelemetry_add_pkgconfig(
+      exporter_otlp_grpc_log_builder
+      "OpenTelemetry OTLP - gRPC Log Builder"
+      "Builder for OTLP gRPC log record exporter."
+      "opentelemetry_exporter_otlp_grpc_log opentelemetry_exporter_otlp_builder_utils"
+    )
+
+    opentelemetry_add_pkgconfig(
+      exporter_otlp_grpc_metrics "OpenTelemetry OTLP - gRPC Metric Exporter"
```
lalitb, line 281: the naming is not consistent with other exporters:
  - `gRPC`: exporter_otlp_grpc_metrics (plural)                                                                                                                                         
  - `HTTP`: exporter_otlp_http_metric (singular)                                                                                                                                        
  - `File`: exporter_otlp_file_metric (singular)

ci/verify_packages.sh
```diff
@@ -7,6 +7,38 @@ set -ex
...
+done
+
+# OTLP exporters
+validate_package "opentelemetry_otlp_recordable"
+validate_package "opentelemetry_exporter_otlp_builder_utils"
+
+for protocol in grpc http file; do
+  validate_package "opentelemetry_exporter_otlp_${protocol}"
+  for variant in client builder log log_builder metrics metric_builder; do
+    validate_package "opentelemetry_exporter_otlp_${protocol}_${variant}"
```
lalitb, line 41: This looks for `opentelemetry_exporter_otlp_http_metrics` but the actual package is `opentelemetry_exporter_otlp_http_metric`. The script is successful as "--exists" check will filter it out.

exporters/otlp/CMakeLists.txt
perhapsmaple: We have the same naming with the actual libraries themselves:

``
  add_library(
    opentelemetry_exporter_otlp_grpc_metrics
``

``
  add_library(
    opentelemetry_exporter_otlp_http_metric
``

``
  add_library(
    opentelemetry_exporter_otlp_file_metric
 ``
 
 I believe the pkgconfig name should match the library target name.

ci/verify_packages.sh
perhapsmaple: I addressed this by splitting it into two different loops as the grpc library ends with ``_metrics``, but http and file end with ``_metric``.

PR #3834, opened 2026-01-28
[API] Environment Variables as Context Propagation Carriers

### LINKED ISSUE #3817: [feature request] Support the environment variable propagation spec in cpp
### Related Problems?

With the [environment variable context propagation spec](https://opentelemetry.io/docs/specs/otel/context/env-carriers/) being added, we'd like to have its support added to CPP. This is part of a larger effort to adding support in each of the currently supported languages.

### Describe the solution you'd like:

I'd like the specification implemented in CPP.

### Considered Alternatives

n/a

### Additional Context

_No response_

perhapsmaple:
Fixes #3817 

Written with inspiration from: [Go PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8442)

## Changes

Need help from maintainers on below:
- I have followed the Go implementation of having a single class. The Python implementation on the other hand uses two separate classes: [Python PR](https://github.com/open-telemetry/opentelemetry-python/pull/4609)
- Is the raw pointer correct for this use case? From my view, the carrier shouldn't own the map. Should we use nostd::shared_ptr for safety?
- I believe Windows has a character limit? Do we need to handle this?
- Keys() only returns known OTel variables (traceparent, tracestate, baggage), not all environment variables. The Go Implementation returns all environment variables. 
- Carrier is NOT thread-safe. Do we need to handle this?


marcalff:
@perhapsmaple 

Thanks for the feature.

> Need help from maintainers on below:
> 
>     * I have followed the Go implementation of having a single class. The Python implementation on the other hand uses two separate classes: [Python PR](https://github.com/open-telemetry/opentelemetry-python/pull/4609)
> 

The spec proposes either a single interface (Propagator), or two separate interfaces (Injector, Extractor) for propagation.

Given that opentelemetry-cpp uses a single interface with a TextMapCarrier, I think keeping a single class makes sense.

>     * Is the raw pointer correct for this use case? From my view, the carrier shouldn't own the map. Should we use nostd::shared_ptr for safety?

A shared pointer is better, to avoid misuse.

> 
>     * I believe Windows has a character limit? Do we need to handle this?
> 

Every operating system will have a character limit, not only windows.

Given that the code does not do a `setenv()` itself, but produces a map of environment variables to set, I propose to leave this to the code making the actual call to fork/exec/etc, so no limits to check preemptively.


>     * Keys() only returns known OTel variables (traceparent, tracestate, baggage), not all environment variables. The Go Implementation returns all environment variables.
> 

The Carrier is supposed to be ignorant of which propagator is used, only propagators will know the keys.

Conceptually, the environment variable carrier is supposed to work with any propagator, including B3, so it should not hard code any known keys.

Notice how none of the carriers, including the `HttpTextMapCarrier`, implement the `Keys()` method, the only supported pattern is set/get with a known key name.

I propose to not implement `Keys()`, used to inspect the whole container, either.

>     * Carrier is NOT thread-safe. Do we need to handle this?

For HTTP, when two threads make an http call and need to inject keys to talk to two different services, each thread will use its own carrier to inject into its own http message.

For environment variables, the pattern should be the same, the `EnvironmentCarrier` object will be per thread, not shared, so there should be no thread safety issue.

---

This PR is looking good already.

A great addition will be an example with a parent process that forks a child, so people can see in practice how to implement the whole thing.

perhapsmaple:
Thanks for the review. I have added a basic example that should work in unix environments. Do we need an example for windows as well?

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -0,0 +1,105 @@
...
+
+    std::string env_name               = ToEnvName(key);
+    env_map_ptr_->operator[](env_name) = std::string(value);
+  }
+
+private:
+  std::shared_ptr<std::map<std::string, std::string>> env_map_ptr_;
+  mutable std::map<std::string, std::string> cache_;
+
+  // Converts a header name to an environment variable name.
```
marcalff, line 93: Also map:

* `my-key` to `MY_KEY`
* `my.complex.key` to `MY_COMPLEX_KEY`

so any character that does not fit an environment variable name should be replaced by underscore.

marcalff:
> Thanks for the review. I have added a basic example that should work in unix environments. Do we need an example for windows as well?

Thanks for the example added. It looks good as is, it shows how to use the environment carrier.

marcalff:
@lalitb 

Could you review as well, given that this change touches the API ? Thanks.

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -0,0 +1,105 @@
...
+  // e.g., "traceparent" -> "TRACEPARENT"
+  static std::string ToEnvName(nostd::string_view key)
+  {
+    std::string env_name(key);
+    std::transform(env_name.begin(), env_name.end(), env_name.begin(),
+                   [](unsigned char c) { return static_cast<char>(std::toupper(c)); });
```
ThomsonTan: nit: `std::toupper(c)` needs header file `#include <cctype>`.

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -0,0 +1,105 @@
...
+    {
+      return cache_it->second;
+    }
+
+    // Read from environment
+    const char *value = std::getenv(env_name.c_str());
+    if (value != nullptr)
+    {
+      // Cache for lifetime management (string_view requires stable storage)
+      cache_[std::string(key)] = std::string(value);
```
ThomsonTan, line 69: Get is marked as `const`, but it does mutate the internal state `cache_`. Perhaps this should be documented in the API doc?

PR #3837, opened 2026-01-29
[CONFIGURATION] File configuration - exemplar filter

marcalff:
Fixes # (issue)

## Changes

* Implement the missing `exemplar_filter` property, not implemented for C++
* Follow up on: https://github.com/open-telemetry/opentelemetry-configuration/pull/131

### DIFF
--- a/sdk/src/configuration/sdk_builder.cc
+++ b/sdk/src/configuration/sdk_builder.cc
@@ -156,6 +156,10 @@
 #include "opentelemetry/sdk/trace/tracer_provider_factory.h"
 #include "opentelemetry/version.h"
 
+#ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
+#  include "opentelemetry/sdk/configuration/exemplar_filter.h"
+#endif
+
 OPENTELEMETRY_BEGIN_NAMESPACE
 namespace sdk
 {


sdk/include/opentelemetry/sdk/configuration/meter_provider_configuration.h
```diff
@@ -16,14 +17,14 @@ namespace sdk
...
-// YAML-SCHEMA: schema/meter_provider.json
+// YAML-SCHEMA: schema/meter_provider.yaml
 // YAML-NODE: MeterProvider
 class MeterProviderConfiguration
 {
 public:
   std::vector<std::unique_ptr<MetricReaderConfiguration>> readers;
   std::vector<std::unique_ptr<ViewConfiguration>> views;
-  // FIXME: exemplar_filter
+  ExemplarFilter exemplar_filter;
```
ThomsonTan: Initialize to ExemplarFilter::trace_based to match spec default. Now it is uninitialized if the object is default-constructed.

sdk/include/opentelemetry/sdk/configuration/exemplar_filter.h
```diff
@@ -0,0 +1,27 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace configuration
+{
+
+// YAML-SCHEMA: schema/meter_provider.yaml
+// YAML-NODE: ExemplarFilter
+enum class ExemplarFilter : std::uint8_t
```
lalitb, line 18: nit - can we reuse the existing [ExemplarFilterType](https://github.com/open-telemetry/opentelemetry-cpp/blob/363407c3a0d704e17edbc4ac96540764152ca80a/sdk/include/opentelemetry/sdk/metrics/exemplar/filter_type.h#L21)

sdk/src/configuration/sdk_builder.cc
```diff
@@ -1500,6 +1500,8 @@ std::unique_ptr<opentelemetry::sdk::metrics::MeterProvider> SdkBuilder::CreateMe
     meter_context->AddMetricReader(metric_reader);
   }
 
+  // FIXME: exemplar_filter not supported in metrics SDK.
```
lalitb: The exemplar filter is supported in the metrics SDK under the `ENABLE_METRICS_EXEMPLAR_PREVIEW` guard. Could we wire up the configuration to actually apply it when that flag is enabled?

sdk/include/opentelemetry/sdk/configuration/meter_provider_configuration.h
marcalff: Thanks, fixed.

sdk/src/configuration/sdk_builder.cc
marcalff: I was expecting to find it in `MeterProviderFactory::Create()`, but this has no exemplar support.

Looks like `MeterContext::SetExemplarFilter()` can be used instead.

Fixing this.

sdk/include/opentelemetry/sdk/configuration/exemplar_filter.h
marcalff: The file configuration in yaml on one hand, and the sdk implementation on the other hand, have different types on purpose.

The main issue is that:

* the yaml part follows the configuration spec, with types and enums added or removed based on compliance to the yaml schema spec, including some items removed on purpose (zipkin, otrace or jaeger propagator, etc), or with some item still missing (tls 1.2, tls 1.3) compared to the sdk implementations.
* the sdk part follows the current implementation of the opentelemetry spec, with features sometime missing since not everything is implemented in opentelemetry-cpp (see the holes in the spec compliance matrix)

In theory, `ExemplarFilter` and `ExemplarFilterType` should be the same, and eventually will be.

In practice, they are different: case in point, `ExemplarFilterType` is conditionally defined, and yet the yaml parser still needs to parse a yaml document even when `ENABLE_METRICS_EXEMPLAR_PREVIEW` is not set.

lalitb: Yes that make sense, I realized after posting the comment.

PR #3845, opened 2026-02-05
[BUILD] Ignore deprecated warning

owent:
Fixes #3835 

## Changes

- Fixes deprecated warning scope

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

Issue #3835, opened 2026-01-28
Build failure (C4996) when including opentelemetry/logs/provider.h with opentelemetry-cpp[geneva]


lukeina2z:
## Steps to Reproduce
1. Follow this article to integrate vcpkg with MSBuild/VisualStudio.

     Tutorial: Install and use packages with MSBuild in Visual Studio
     https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-msbuild

     ```sh
     .\vcpkg.exe integrate install
     ```
2. Create a C++ console application in Visual Studio 2022.
3. In Visual Studio, open Tools → Command Line → Developer Command Prompt.
4. Run:
    ```sh
   vcpkg new --application
   vcpkg add port opentelemetry-cpp[geneva]
   ```
5. In the project Property Pages, set:   vcpkg → Use Vcpkg Manifest = Yes
6. Include the following header in your .cpp file:
    ```C++
   #include "opentelemetry/logs/provider.h"
   ```
7. Build the project.

## Expected Result
The project builds successfully with no errors.

## Actual Result
The build fails with error C4996 due to deprecated symbols:

```text
opentelemetry::v1::logs::EventLogger was declared deprecated
opentelemetry::v1::logs::EventLogger::~EventLogger was declared deprecated
opentelemetry::v1::logs::EventLoggerProvider was declared deprecated
opentelemetry::v1::logs::EventLoggerProvider::~EventLoggerProvider was declared deprecated
opentelemetry::v1::logs::Provider::GetEventProvider was declared deprecated
```
## Additional Notes
This header is included because its APIs are required to set the Geneva `TracerProvider` as the global default, specifically to call:

GetLoggerProvider()
SetLoggerProvider()

Here is my testing project for this issue: https://github.com/lukeina2z/playground/tree/main/cpp/LogProviderWarning

lalitb:
The previous fix #3592 reduced the warnings, but looks like MSVC can still emit `C4996` just from compiling the headers in ABI v1 mode, and `/WX` turns that into a hard failure. Recommended workaround is to enable `ABI v2` via the vcpkg option (or temporarily disable `C4996` / don’t treat it as error). For a permanent fix, feel free to open a follow-up PR to guard the remaining header-scope deprecated references ((e.g., the EventLogger* wrappers/noop types).

owent:
[...]

PR #3855, opened 2026-02-09
[BUILD] Revisit EventLogger deprecation

marcalff:
Fixes # (issue)

## Changes

* Removed the OPENTELEMETRY_DEPRECATED C++ annotation for `EventLogger`.
  * This was used to tell the compiler to emit a warning.
* Removed all the warning suppression code from the api
  * This was used to tell the compiler to suppress the warning just added, doing both makes no sense.
* Preserved OPENTELEMETRY_DEPRECATED only on entry points, without warning suppression pragma:
  * API opentelemetry::logs::Provider::GetEventLoggerProvider()
  * API opentelemetry::logs::Provider::SetEventLoggerProvider()
  * SDK opentelemetry::sdk::logs::EventLoggerProviderFactory
* Added warning suppression code in unit tests only
  * This allows to have test coverage of deprecated code, while keeping the build clear of noise.
* Documented the code as deprecated in doxygen
* Updated DEPRECATED.md to clarify the migration path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -432,16 +432,14 @@ if(OTELCPP_MAINTAINER_MODE)
     add_compile_options(/WX)
     add_compile_options(/W4)
 
-    # Relaxed warnings (ignored)
+    # Relaxed warnings
     add_compile_options(/wd4100)
     add_compile_options(/wd4125)
     add_compile_options(/wd4566)
     add_compile_options(/wd4127)
     add_compile_options(/wd4512)
     add_compile_options(/wd4267)
-
-    # Relaxed warnings (still reported as warnings, not errors)
-    add_compile_options(/w34996) # 'xxx' was declared deprecated
+    add_compile_options(/wd4996)
 
     # Enforced warnings
     add_compile_options(/we4265) # 'class': class has virtual functions, but


lalitb:
I'm not fully clear on the intent behind these changes. My understanding is that anyone using EventLogger should receive a deprecation warning. That was broken in #3845, and from what I can tell, this change doesn't restore that behavior either - so what's the purpose of removing the warning?

marcalff:
@lalitb 

The main problem is to decide clearly what opentelemetry-cpp wants to do with deprecation for EventLogger in particular.

# Option 1

We mark code with `OPENTELEMETRY_DEPRECATED`, causing compiler warnings.

Anyone who uses a deprecated class gets a warning.

This includes:

* unit tests in opentelemetry-cpp
* noop providers
* the `Provider` class as well, because of Get/Set EventLoggerProvider.

This creates a problem for applications, even when not using EventLogger, because the simple fact of including "provider.h" (which is required and can not be avoided) causes the build to fail with -Wall -Werr.



# Option 2

We do not mark code with `OPENTELEMETRY_DEPRECATED`, and only document that some classes are deprecated.

This is what this PR does.

---

Option 1 is what was done since the beginning with #3285, but recently issue #3835 was reported complaining (correctly) that there is no way to have a clean build, and this was "fixed" by suppressing all warnings with #3845.

If we prefer to have warnings, we need to rollback #3845.

If we prefer to not have warnings, we need this PR to clean up the code.

Leaving both `OPENTELEMETRY_DEPRECATED` and pragmas to disable deprecation warnings makes no sense.

The root cause of all this is that, when a deprecation is added, there should be a migration path for users to take and avoid using the deprecated code.

In this case, because the Get/Set EventLoggerProvider methods are inline, part of the API, and located in a mandatory header file, there is no way for user code to avoid compiling deprecated code, so there is no way to resolve deprecation warnings if present.

We can not (even with conditional compilation) opt out for these methods either, as this will break the ABI for version 1.

sdk/include/opentelemetry/sdk/logs/event_logger_provider.h
```diff
@@ -23,7 +23,7 @@ namespace sdk
 namespace logs
 {
 #if OPENTELEMETRY_ABI_VERSION_NO < 2
-class OPENTELEMETRY_EXPORT OPENTELEMETRY_DEPRECATED EventLoggerProvider final
+class OPENTELEMETRY_EXPORT EventLoggerProvider final
```
ThomsonTan, line 30: Missing @deprecated doxygen on EventLoggerProvider class.

sdk/include/opentelemetry/sdk/logs/event_logger.h
```diff
@@ -19,13 +19,14 @@ namespace sdk
...
 #if OPENTELEMETRY_ABI_VERSION_NO < 2
-class OPENTELEMETRY_DEPRECATED EventLogger final : public opentelemetry::logs::EventLogger
+class EventLogger final : public opentelemetry::logs::EventLogger
 {
 public:
   /**
    * Initialize a new Eventlogger.
    * @param delegate_logger The delegate logger instance
    * @param event_domain Event domain
+   * @deprecated
```
ThomsonTan: Should this @deprecated tag be put on the class, not the constructor?

api/include/opentelemetry/logs/noop.h
```diff
@@ -91,17 +91,6 @@ class NoopLoggerProvider final : public LoggerProvider
...
-#    pragma warning(disable : 4996)
-#  elif defined(__GNUC__) && !defined(__clang__) && !defined(__apple_build_version__)
-#    pragma GCC diagnostic push
-#    pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-#  elif defined(__clang__) || defined(__apple_build_version__)
-#    pragma clang diagnostic push
-#    pragma clang diagnostic ignored "-Wdeprecated-declarations"
-#  endif
-
 class NoopEventLogger final : public EventLogger
```
ThomsonTan, line 98: Missing @deprecated on the class?

marcalff: Fixed.

sdk/include/opentelemetry/sdk/logs/event_logger_provider.h
marcalff: Fixed.

sdk/include/opentelemetry/sdk/logs/event_logger.h
marcalff: Fixed.

lalitb:
Thanks @marcalff. I agree that having both `OPENTELEMETRY_DEPRECATED` and pragma suppression together on the same class makes no sense - that definitely needs fixing.

However, I don't think this is strictly **Option 1** vs **Option 2.** There's an **Option 3**:

  The concern with **Option 1** is that including `provider.h` causes deprecation warnings even for users who never use `EventLogger`. But that's only true if we put `OPENTELEMETRY_DEPRECATED` on the class declarations - which, combined with MSVC warning on declarations, forces warnings on everyone who includes the header.

  The fix is to put `OPENTELEMETRY_DEPRECATED` only on `GetEventLoggerProvider()` and `SetEventLoggerProvider()` methods - not on the classes. The pragma push/pop around these method declarations in the header suppresses the MSVC declaration-site warning, so just including `provider.h` won't cause any warning. But when a user actually calls `GetEventLoggerProvider()` or `SetEventLoggerProvider()` in their own code, the deprecation warning fires at the call site - on GCC/Clang and intended on MSVC via scoped pragma suppression. That's exactly the behavior we want.

> In this case, because the Get/Set EventLoggerProvider methods are inline, part of the API, and located in a mandatory header file, there is no way for user code to avoid compiling deprecated code, so there is no way to resolve deprecation warnings if present.

  This is true for MSVC (which warns at declaration), but pragma push/pop around the method declarations in the header handles exactly that - it suppresses the MSVC declaration-site warning so including the header is clean. On GCC/Clang, `OPENTELEMETRY_DEPRECATED` on a method only warns at the call site, not the declaration, so there's no issue at all. The pragma around the declaration isn't defeating the purpose here - it's specifically scoped to suppress the MSVC declaration quirk. User code that actually calls  `GetEventLoggerProvider()` is outside the pragma scope, so the deprecation warning correctly fires at the call site.

  So:

   - `EventLogger`/`EventLoggerProvider` classes: Remove `OPENTELEMETRY_DEPRECATED` and pragma push/pop  (as this PR does)
   - `Noop`/`SDK` implementations: Remove `OPENTELEMETRY_DEPRECATED`
   - `GetEventLoggerProvider()`/`SetEventLoggerProvider()` in `provider.h`: Keep `OPENTELEMETRY_DEPRECATED` on these methods, with pragma push/pop around the declarations (for MSVC)

  This gives users a clean build when they only use `LoggerProvider::GetLogger()`, and a clear deprecation warning when they actually call the deprecated Event API entry points. No ABI break, no need for conditional compilation.

 We could also enforce this in CI by adding a small compile test target that compiles with `-Werror=deprecated-declarations` (GCC/Clang). or `/we4996` (MSVC), verifying that including `provider.h` and using only `GetLoggerProvider()` compiles clean, while calling  `GetEventLoggerProvider()` produces a compile error. This would be a new addition though.

marcalff:
@lalitb 

Thanks for the comments.

I implemented:

* Preserved OPENTELEMETRY_DEPRECATED only on entry points, without warning suppression pragma:
  * API opentelemetry::logs::Provider::GetEventLoggerProvider()
  * API opentelemetry::logs::Provider::SetEventLoggerProvider()
  * SDK opentelemetry::sdk::logs::EventLoggerProviderFactory

which is basically option 3.

The claim that MSVC raises a deprecation warning at declaration site is very weird to me, so I wanted to see this for myself to confirm, and used deprecated provider set/get methods without the warning suppression code.

It turns out, from the CI logs even on windows, that no warning are generated, unless I missed something in build logs.

Starting fresh with all the related cleanup done in EventLogger classes, code like:

```cpp
  OPENTELEMETRY_DEPRECATED static nostd::shared_ptr<EventLoggerProvider>
  GetEventLoggerProvider() noexcept
  {
    std::lock_guard<common::SpinLockMutex> guard(GetLock());
    return nostd::shared_ptr<EventLoggerProvider>(GetEventProvider());
  }
```

behaves as it should.

I suspect that the following helps:

* the GetEventLoggerProvider() prototype, in particular the return type, no longer uses deprecated classes
* the GetEventLoggerProvider() implementation, inlined in the header file, no longer uses deprecated classes

so that no warnings are generated at declaration time, even on MSVC.

Please double check with a local build if possible.

ThomsonTan:
For MSVC, that’s generally true—especially for deprecation warnings (e.g., C4996), which are typically emitted at the point of use (the call site) rather than at the declaration.

> so that no warnings are generated at declaration time, even on MSVC.

api/test/logs/provider_test.cc
```diff
@@ -74,6 +74,24 @@ TEST(Provider, GetLogger)
 }
 
 #if OPENTELEMETRY_ABI_VERSION_NO < 2
+
+/*
+ * opentelemetry::logs::Provider::GetLoggerProvider() is deprecated.
+ * opentelemetry::logs::Provider::SetLoggerProvider() is deprecated.
```
ThomsonTan: Should this be `GetEventLoggerProvider` and `SetEventLoggerProvider` which are deprecated?

marcalff: Indeed. Fixed.

examples/common/logs_foo_library/CMakeLists.txt
```diff
@@ -12,24 +12,3 @@ target_include_directories(common_logs_foo_library
                            PUBLIC $<BUILD_INTERFACE:${EXAMPLES_COMMON_DIR}>)
 
 target_link_libraries(common_logs_foo_library PUBLIC opentelemetry-cpp::api)
-
```
owent, line 15: This ensures users don’t see deprecation warnings or warnings-as-errors when including `opentelemetry/logs/logger.h` or `opentelemetry/logs/logger_provider.h` without using deprecated APIs. Should we keep it here?

marcalff: logger.h and logger_provider.h no longer declare deprecated APIs, only the Set/Get on provider.h does.

This change in CMake is no longer needed.

owent: Are `GetEventLoggerProvider` and `SetEventLoggerProvider` still present in `logger_provider.h`?

These cmake  scripts also serve as a safeguard to prevent accidentally introducing deprecated macros in the future — macros that may only emit warnings on certain compilers and could therefore be silently overlooked.

marcalff: Rolled back all changes in CMakeList.txt here.

examples/common/metrics_foo_library/foo_library.cc
```diff
@@ -22,9 +22,9 @@
 #include "opentelemetry/nostd/shared_ptr.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/nostd/variant.h"
+#include "opentelemetry/semconv/container_metrics.h"
 #include "opentelemetry/semconv/http_metrics.h"
-#include "opentelemetry/semconv/incubating/container_metrics.h"
-#include "opentelemetry/semconv/incubating/system_metrics.h"
+#include "opentelemetry/semconv/system_metrics.h"
```
lalitb, line 27: nit - this change looks not related to the PR.  Is it intended. Not asking to remove unless this is changed by mistake.

lalitb:
Though I do see a separate blocker: this PR changes MSVC maintainer warning policy for C4996 (/wd4996 -> /w34996 under /WX), which causes unrelated existing deprecations (examples/grpc semconv RPC constants and sscanf  configuration parser) to fail Windows CI. Could we either 
- revert the C4996 policy change in this PR (preferred, to keep scope focused), and open a follow-up tracking issue to add a dedicated deprecation contract test 
- or include fixes for all current C4996 callsites in the same PR?

examples/common/metrics_foo_library/foo_library.cc
marcalff: The incubating headers contains a lot of deprecated code.
Since the semconv we use are all stable by now, this avoids more warnings, this is intended.

marcalff:
> Though I do see a separate blocker: this PR changes MSVC maintainer warning policy for C4996 (/wd4996 -> /w34996 under /WX), which causes unrelated existing deprecations (examples/grpc semconv RPC constants and sscanf configuration parser) to fail Windows CI. Could we either
> 
>     * revert the C4996 policy change in this PR (preferred, to keep scope focused), and open a follow-up tracking issue to add a dedicated deprecation contract test
> 
>     * or include fixes for all current C4996 callsites in the same PR?

This was an attempt at more cleanup but it failed, reverting it.

Cleaning up semconv is a separate problem, see #3890

PR #3858, opened 2026-02-10
[CONFIGURATION] File configuration - handle invalid YAML

### LINKED ISSUE #3857: [CONFIGURATION] Handle invalid yaml
This is a follow up on #3851 .

The fix in 3851 is valid and addresses immediate needs, but there is more to the story.

A follow up PR is needed to:

* add a unit test in `sdk/test/configuration/yaml_test.cc`
* install ryml error handlers explicitly, instead of relying on the default behavior, by installing `Callbacks` in ryml

See https://github.com/biojppm/rapidyaml/blob/c1d0047fd3f30b09133487ba5bb74195816b02e7/src/c4/yml/common.hpp#L510

perhapsmaple:
Fixes #3857

## Changes

### DIFF
--- a/sdk/src/configuration/ryml_document.cc
+++ b/sdk/src/configuration/ryml_document.cc
@@ -7,7 +7,6 @@
 #include <ostream>
 #include <ryml.hpp>
 #include <ryml_std.hpp>
-#include <stdexcept>
 #include <string>
 
 #include "opentelemetry/sdk/common/global_log_handler.h"


sdk/src/configuration/ryml_document.cc
```diff
@@ -21,6 +22,30 @@ namespace sdk
...
+                           ryml::Location location,
+                           void * /*user_data*/)
+{
+  std::string error_msg(msg, msg_len);
+  if (location.line != ryml::npos)
+  {
+    error_msg += " at line " + std::to_string(location.line + 1) + " col " +
+                 std::to_string(location.col + 1);
+  }
+  throw std::runtime_error(error_msg);
```
marcalff: Please throw `InvalidSchemaException` instead.

sdk/src/configuration/ryml_document.cc
```diff
@@ -6,6 +6,7 @@
 #include <ostream>
 #include <ryml.hpp>
 #include <ryml_std.hpp>
+#include <stdexcept>
```
marcalff: From the CI build logs:

```

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/src/configuration/ryml_document.cc should add these lines:
#include <stddef.h>                                              // for size_t
```

so add `#include <stddef.h>` as well.

sdk/test/configuration/yaml_test.cc
```diff
@@ -717,3 +717,25 @@ file_format: "1.0"
...
+    - simple:
+        exporter:
+          otlp_http:
+            endpoint: http://localhost:4318/v1/traces
+            certificate: /app/cert.pem
+            client_key: /app/cert.pem
+            client_certificate: /app/cert.pem
+              headers:
+                - name: header_name
+                  value: header_value
```
marcalff, line 737: In the test yaml, please add a comment to point out why and where it is malformed.

This will avoid confusion, in case someone wants to "fix" the test.

Issue #3861, opened 2026-02-12
v1.25 regression: Build break on Windows with Visual Studio 2022


lukeina2z:
Below PR, released with v1.25, breaks existing applications on Windows.

[EXPORTER] Add support for otlp exporter collection limits
https://github.com/open-telemetry/opentelemetry-cpp/pull/3816

## Steps to Reproduce
1. Follow this article to integrate vcpkg with MSBuild/VisualStudio.

     Tutorial: Install and use packages with MSBuild in Visual Studio
     https://learn.microsoft.com/en-us/vcpkg/get_started/get-started-msbuild

     ```sh
     .\vcpkg.exe integrate install
     ```
2. Create a C++ console application in Visual Studio 2022.
3. In Visual Studio, open Tools → Command Line → Developer Command Prompt.
4. Run:
    ```sh
   vcpkg new --application
   vcpkg add port opentelemetry-cpp[geneva]
   ```
   Ensure the commit hash in your vcpkg-configuration.json file includes v1.25 release.
5. In the project Property Pages, set:   vcpkg → Use Vcpkg Manifest = Yes
6. Include the following code in your .cpp file: https://github.com/lukeina2z/playground/blob/main/cpp/OTelTestAppLite/main.cpp
   You can use that project for this repro.
7. Build the project.

## Expected Result
The project builds successfully with no errors.

## Actual Result
The build fails with error C2589:
'(': illegal token on right side of '::'

## Note: The problem does not exist in v1.24. For quick testing, switch the vcpkg baseline in the config file:

```json
1.25 build break:
"baseline": "0a434205c521ca43d66921271ae2a1d051e718a5",

1.24 good build:
  "baseline": "de51e6bfa96e1245f2a969a8cde249baac89f7be",
```

<img width="1535" height="927" alt="Image" src="https://github.com/user-attachments/assets/66d87659-13e0-449e-89dc-d48287bbe8a2" />

marcalff:
Windows is known to pollute the build with min and max macros, causing the C/C++ preprocessor to trip and break legal code.

See `NOMINMAX`.

Opentelemetry-cpp can be more defensive against it, sure, but the culprit is windows.h.

marcalff:
@lukeina2z 

Thanks for reporting this, and adopting the latest releases.

Please try the following work around:

```cpp
#define NOMINMAX
#include "windows.h"
```

if the application includes "windows.h" and opentelemetry-cpp header files in the same compile unit.

[...]

PR #3865, opened 2026-02-13
[TEST] Add Windows max macro regression test for OTLP public headers

ThomsonTan:
Follow up for the comment https://github.com/open-telemetry/opentelemetry-cpp/issues/3861#issuecomment-3893597977

## Changes

* Adds a new regression test in otlp_windows_max_macro_compile_test.cc that defines a Windows-style max macro before including OTLP public headers.
* Prevents recurrence of the Windows/MSVC include-order failure reported in issue #3861; no runtime behavior changes, test-only coverage improvement.

### DIFF
--- a/.github/workflows/dependencies_image.yml
+++ b/.github/workflows/dependencies_image.yml
@@ -29,7 +29,7 @@ jobs:
       uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
     -
       name: Build Image
-      uses: docker/build-push-action@601a80b39c9405e50806ae38af30926f9d957c47 # v6.19.1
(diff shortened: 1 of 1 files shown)


marcalff:
@ThomsonTan 

What this new test will do, is prevent similar bugs to be introduced again in `otlp_http_exporter_options.h`.

It will not prevent another similar bug to appear in another header file.

Suggestion: implement a new CI check script, similar to `tools/check_copyright.sh`, that looks for known dangerous coding pattern.

A script running `grep -w "max("` on all the source code will pick up the next break.

ThomsonTan:
> script, similar to `tools/check_copyright.sh`, that looks for known dangerous coding pattern.
> 
> A script running `grep -w "max("` on all the source code will pick up the next break.

Ok. Close this PR for now as it only validates regression from very specific scenario. Will think on making it general like define it as global before including the API headers.

PR #3887, opened 2026-02-23
Move implementation from SDK header files to SDK cc

perhapsmaple:
Contributes to #1429 

## Changes

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
@@ -55,7 +55,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@fa2e9d605c4eeb9fcad4c99c224cee0c6c7f3594 # v2.16.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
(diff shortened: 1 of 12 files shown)


marcalff (commented):
Thanks for the fix.

This has a good size already, so please stabilize this PR to make it ready for review and merge. More files can be changed in a different PR, if so inclined.

To fix:

* IWYU build failures
  * look for "Warning: include-what-you-use reported diagnostics:" in the build logs, in section iwyu_tool
* Resolve merge conflicts
* clang-format

sdk/include/opentelemetry/sdk/common/empty_attributes.h
```diff
@@ -20,17 +19,8 @@ namespace sdk
  * This helps to avoid constructing a new empty container every time a call is made
  * with default attributes.
  */
-static const opentelemetry::common::KeyValueIterableView<
-    std::array<std::pair<std::string, int32_t>, 0>> &
-GetEmptyAttributes() noexcept
```
lalitb, line 25: `GetEmptyAttributes()` was previously a static free function (header-only, internal linkage). It's now an externally-linked symbol requiring linkage against `opentelemetry_common`.  While this shouldn't be an issue in practice, someone could theoretically be including this header standalone without linking the SDK -  for their own utility purposes - and that would now fail with an undefined symbol error. Please add a CHANGELOG entry noting this change.

sdk/include/opentelemetry/sdk/trace/span_data.h
ThomsonTan: Move this and `SpanDataLink::GetAttributes` to .cc for consistency, since `SpanData::GetAttributes` has been moved to .cc.

lalitb:
@perhapsmaple  - Would you be resolving the remaining comments, so it can be merged. thanks.

perhapsmaple:
@lalitb Sorry I've been a little busy the last couple of weeks. Please give me a couple of days time to fix the remaining comments

marcalff:
When moving code from x.h to x.cc, make sure to preserve the clang-tidy cleanup implemented in x.h recently, if any, when merging with a recent main.

sdk/test/common/BUILD
```diff
@@ -143,6 +143,7 @@ cc_test(
     deps = [
         "//api",
         "//sdk:headers",
+        "//sdk/src/common:empty_attributes",
```
ThomsonTan, line 159: Seems this is not added to CMakeLists.txt?

perhapsmaple: Good catch. I also found `circular_buffer_test` which is in CMake but missing from Bazel

Issue #3484, opened 2025-06-17
Preserving timestamps for prometheus exporter


nishkrishnan:
I see [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/2324) that timestamps were explicitly removed from the prometheus exporter. However, shouldn't this be a setting for the user that allows them to choose whether timestamps should be honored or not? 

In my case I want to preserve the timestamps that come from the source.  How can I achieve that currently?

owent:
The time setting is now restored in the latest version and is derived from the metrics data. Exporters should only handle I/O; if you need to modify metrics data, please implement a custom processor.

dashpole:
You can see the original issue here: https://github.com/open-telemetry/opentelemetry-cpp/issues/2316.  Unless you are re-exposing metrics from another source with significantly different timestamps, you should not add them.  I'm not opposed to having an option to enable explicit timestamps, but we should make sure we define it in the [spec here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/prometheus.md#configuration) so that languages use similar names for the option.

PR #3894, opened 2026-03-01
[EXPORTERS] Configure prometheus timestamps

### LINKED ISSUE #2316: Prometheus exporter should not set explicit timestamps
**Steps to reproduce**

I saw second-hand that the prometheus exporter adds explicit timestamps, but don't have exact reproduction steps.

```
# TYPE grpc_client_attempt_started_ counter
grpc_client_attempt_started_{grpc_method="grpc.testing.TestService/UnaryCall",grpc_target="dns:///10.52.1.55:14285"} 10 1694819876572
```

The last number on the right is the start timestamp.

Metrics should not have explicit timestamps set. From the [OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric)

> MetricPoints SHOULD NOT have explicit timestamps.

Explicit timestamps are added to prometheus metrics.

https://github.com/open-telemetry/opentelemetry-cpp/blob/0563a71ce3ea2c580f535996aeb8ce9246f2d703/exporters/prometheus/src/exporter_utils.cc#L265

**Additional context**

In prometheus, the scrape time is used as the timestamp.  Explicit timestamps can limit some functionality, such as compatibility with the [prometheus push gateway](https://github.com/prometheus/pushgateway#about-timestamps), so they are generally not recommended.

carlosroman:
Fixes #2316 and #3484.

## Changes

This PR adds the ability to configure Prometheus timestamps. It was originally removed in #2316 but #2213 added it back in. I have another PR (#3895) that actually removes explicit timestamps from metric points instead if you'd prefer that one over this one. The reason I created this one is that it has been like this for 3 years so maybe people rely on this behavior.

### DIFF
--- a/exporters/prometheus/src/exporter_options.cc
+++ b/exporters/prometheus/src/exporter_options.cc
@@ -71,8 +71,7 @@ static inline bool GetPrometheusWithoutTypeSuffix()
 
 inline bool GetPrometheusWithoutTimestamps()
 {
-  constexpr char kPrometheusWithoutTypeSuffix[] =
-      "OTEL_CPP_PROMETHEUS_EXPORTER_WITHOUT_TIMESTAMPS";
+  constexpr char kPrometheusWithoutTypeSuffix[] = "OTEL_CPP_PROMETHEUS_EXPORTER_WITHOUT_TIMESTAMPS";
   bool setting{};
   const auto exists =
       opentelemetry::sdk::common::GetBoolEnvironmentVariable(kPrometheusWithoutTypeSuffix, setting);
--- a/exporters/prometheus/src/exporter_utils.cc
+++ b/exporters/prometheus/src/exporter_utils.cc
@@ -693,7 +693,7 @@ void PrometheusExporterUtils::SetMetricBasic(
     const opentelemetry::sdk::resource::Resource *resource)
 {
   if (!without_timestamps)
-  {  
+  {
     metric.timestamp_ms = time.count() / 1000000;
   }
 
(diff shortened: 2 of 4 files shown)


lalitb:
@carlosroman  - there are merge conflicts to be resolved in this PR.

exporters/prometheus/src/exporter_options.cc
```diff
@@ -69,12 +69,23 @@ inline bool GetPrometheusWithoutTypeSuffix()
   return exists ? setting : false;
 }
 
+inline bool GetPrometheusWithoutTimestamps()
+{
+  constexpr char kPrometheusWithoutTypeSuffix[] = "OTEL_CPP_PROMETHEUS_EXPORTER_WITHOUT_TIMESTAMPS";
```
lalitb, line 74: nit - The variable is named `kPrometheusWithoutTypeSuffix` but holds the timestamps env var. Should be `kPrometheusWithoutTimestamps`?

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [EXPORTER] Prometheus: Adding ability to disable timestamps from metric points
+  [#?](https://github.com/open-telemetry/opentelemetry-cpp/pull/?)
```
lalitb: nit - PR number needs to be added?

carlosroman: Have added PR number.

marcalff:
Thanks for both PRs.

As discussed in #3895, closing this PR 3894 and merging 3895.

PR #3935, opened 2026-03-14
[EXAMPLE] Add manual asynchronous context propagation example

### LINKED ISSUE #1464: [Trace] reopen - Missing example instrumenting asynchronous operation #1463
Open this issue since I don’t think #1463 Is well resolved, as we don’t find a good example how to instrument function call with async callback, in most condition, I/O operation will be made in an asynchrous fashion, but it doesn’t seem to be well supported by the cpp lib.

jgomezselles:
Fixes #1464 

## Changes

This PR adds an example that demonstrates tracing async requests between generic clients and servers. It simulates context injections and extractions and shows how to explicitly set parent span_id's in different situations, when the scope is not available, or different threads are run in parallel.

The proposed pattern may be used for any kind of asynchronous client/server communication.

It defines auxiliary functions that can be easily adapted to other code bases together with a README documenting the flow and expected outcome.

Apart from markdown lining and clang formatting, BUILD and relevant CMakeLists have been adapted.


jgomezselles:
Hi!

First I wanted to say that I'm a big fan of this project. I've been involved with OpenTelemetry and C++ in different ways in the past,
(i.e. Product Manager of OpenTelemetry at Red Hat a couple of years ago or giving some talks about how to instrument C++ code, for example [at FOSDEM](https://www.youtube.com/watch?v=27vTx3FgB0s) last year).

When I started trying to include this lib some years ago, I have to confess struggled a bit. Since then, I really wanted to contribute with an example like this, so folks can just easily create spans in regular C++ code bases that deal with async calls. Finally I had some quality time.

Of course, I'm very open for feedback. I tried my best to clearly explain this with a simple example that doesn't need to rely on other libs or clients and servers. I hope this makes sense.

For context: I'm using this kind of implementation in an async http2 client I created for load testing and demos (repo [here](https://github.com/jgomezselles/hermes/blob/main/src/http2_client/client_impl.cpp) ).

If this is eventually approved, I'd like to also contribute to the documentation.

Thanks again, and congrats on the great work maintaining this library!

lalitb:
Thanks for the PR. 

Non-blocking: this example is correct, but it reads more like a carrier-based context propagation example than an async programming example. Since we already have an HTTP propagation example, this would be more distinctive if it either (a) were renamed/reframed around manual async context propagation, or (b) showed a more practical in-process async pattern such as passing SpanContext from the submitting thread to worker-thread child spans.

jgomezselles:
Thanks for the feedback. I'll think about it and will come back with changes.

jgomezselles:
Hi again,

Thanks again for the feedback. I've been thinking about which of the proposed options to take. I finally moved on with the first one, because I want to show something more than just an async example. It is absolutely true that it went far from just that simple async example, so I'm reframing it.

Now, I think it's more clear what I'm trying to show: how to **manually** propagate context in a couple of cases.  They include the asynchronous case, but that's a detail.

Changes added:
- Reworked some of the explanations and fixed typos
- Renamed and reframed explanations towards a manual asynchronous context propagation example
- Renamed the PR accordingly
- Applyed fixes to previously failing CI jobs (linting, uwyu, etc)
- Added Changelog

I hope this is better now. Thanks again!

examples/manual_propagation/main.cc
```diff
@@ -0,0 +1,237 @@
...
+#include "opentelemetry/trace/span_startoptions.h"
+#include "opentelemetry/trace/tracer.h"
+#include "opentelemetry/trace/tracer_provider.h"
+
+namespace trace_api = opentelemetry::trace;
+namespace trace_sdk = opentelemetry::sdk::trace;
+namespace nostd     = opentelemetry::nostd;
+namespace ctx       = opentelemetry::context;
+
+using header_map = std::map<std::string, std::string>;
```
dbarker: Please use the Google naming conventions in this file for the types and functions.

examples/manual_propagation/main.cc
```diff
@@ -0,0 +1,237 @@
...
+}
+
+nostd::shared_ptr<trace_api::Tracer> get_tracer()
+{
+  auto provider = trace_api::Provider::GetTracerProvider();
+  return provider->GetTracer("foo_library");
+}
+}  // namespace
+
+namespace utils
```
dbarker: Please fix the clang-tidy warnings from this file: 

https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/23163761181

----

opentelemetry-cpp/examples/manual_propagation/main.cc (7 warnings)

| Line | Check | Message |
|---|---|---|
| 74 | `misc-use-internal-linkage` | function 'create_span' can be made static or moved into an anonymous namespace to enforce internal linkage |
| 83 | `misc-use-internal-linkage` | function 'create_child_span' can be made static or moved into an anonymous namespace to enforce internal linkage |
| 119 | `cppcoreguidelines-avoid-const-or-ref-data-members` | member 'headers_' of type 'header_map &' (aka 'map<basic_string<char>, basic_string<char>> &') is a reference |
| 122 | `misc-use-internal-linkage` | function 'inject_trace_context' can be made static or moved into an anonymous namespace to enforce internal linkage |
| 138 | `misc-use-internal-linkage` | function 'create_child_span_from_remote' can be made static or moved into an anonymous namespace to enforce internal linkage |
| 156 | `misc-use-internal-linkage` | function 'process_answer' can be made static or moved into an anonymous namespace to enforce internal linkage |
| 171 | `misc-use-internal-linkage` | function 'process_request' can be made static or moved into an anonymous namespace to enforce internal linkage |


----

examples/manual_propagation/main.cc
```diff
@@ -0,0 +1,237 @@
...
+nostd::shared_ptr<trace_api::Span> create_span(const std::string &name, trace_api::SpanKind kind)
+{
+  trace_api::StartSpanOptions opts;
+  opts.kind = kind;
+
+  auto span = get_tracer()->StartSpan(name, opts);
+  return span;
+}
+
+nostd::shared_ptr<trace_api::Span> create_child_span(
```
dbarker: not blocking - Consider consolidating the `create_span` and `create_child_span` functions into one

examples/manual_propagation/main.cc
```diff
@@ -0,0 +1,237 @@
...
+    if (it != headers_.end())
+    {
+      return it->second;
+    }
+    return "";
+  }
+
+  void Set(nostd::string_view k, nostd::string_view v) noexcept override
+  {
+    headers_.emplace(std::string(k.data(), k.size()), std::string(v.data(), v.size()));
```
dbarker: Should this line overwrite the value if the key already exists?

examples/manual_propagation/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+An auxiliary function that `injects` the tracing information to a given carrier
+structure. In this example a simple implementation of the `TextMapCarrier` as
+an `std::map<std::string, std::string>` was used.
+
+It's important to note that the `trace_api::SetSpan(current_ctx, span)`
+function is needed for the injection to explicitly use the desired span
+context.
+
+```cpp
```
dbarker: not blocking - consider moving these notes to comments in the example main.cc and removing code blocks here.

examples/manual_propagation/README.md
```diff
@@ -0,0 +1,179 @@
...
+  auto current_ctx   = ctx::RuntimeContext::GetCurrent();
+  auto ctx_with_span = trace_api::SetSpan(current_ctx, span);
+
+  // Then we inject the span info into the headers
+  HttpTextMapCarrier carrier(headers);
+  ctx::propagation::GlobalTextMapPropagator::GetGlobalPropagator()->Inject(carrier, ctx_with_span);
+}
+```
+
+The previous examples make use of the following aliases:
```
dbarker: not blocking - consider removing this note on aliases. Duplicate code in the README and main.cc may not add much value while increasing maintenance overhead.

dbarker (commented):
Thanks for the PR! I've shared some comments and a question below.

examples/manual_propagation/main.cc
jgomezselles: Sorry I missed it. I'm applying changes in the next commit. 
BTW, I left the `get_tracer` as in the other examples to keep the same notation. Please, let me know if it would be useful to make another PR changing all occurrences, and I'll work on that.

examples/manual_propagation/main.cc
jgomezselles: Ack thanks for pointing that output out.

- Applying the anonymous namespace for all of them in the next commit.
- Regarding the header ref, I'll use `std::reference_wrapper` (same pattern than in other examples, and we actually need it to be a ref).

examples/manual_propagation/main.cc
jgomezselles: Fair enough, thanks. My main blocker was not being able to figure out how to default to a nullptr within the nostd, and I dodn't want to create a shared_ptr of a span every call.

But, I'll default it to `const nostd::shared_ptr<trace_api::Span> &parent = nostd::shared_ptr<trace_api::Span>{}` , which actually would make it IIUC.

examples/manual_propagation/main.cc
jgomezselles: It should! Thanks. I would use `insert_or_apply`, but I think it's not available for cpp14. I'll use brackets instead.

examples/manual_propagation/README.md
jgomezselles: Done, sorry for the verbosity. As you point in the next comment, the maintainability burden would not be great. Moved and simplified.

examples/manual_propagation/README.md
jgomezselles: removed. Thanks

jgomezselles:
Thanks for the review! I did my best to apply the changes. Please let me know if there are more things to improve (I'll keep an eye on the linters, etc too)

examples/manual_propagation/main.cc
```diff
@@ -0,0 +1,246 @@
...
+HeaderMap ProcessRequest(HeaderMap &remote_headers)
+{
+  // Extract remote context and create server-side child span
+  auto server_span =
+      utils::CreateChildSpanFromRemote(remote_headers, "server", trace_api::SpanKind::kServer);
+
+  // Simulating work with nested spans
+  server_span->AddEvent("Processing in server");
+  auto nested = utils::CreateSpan("nested", trace_api::SpanKind::kServer, server_span);
+  nested->AddEvent("Nested did some work");
```
lalitb: Let's not use the Span Event API here. It's planned to be deprecated - https://github.com/open-telemetry/opentelemetry-specification/blob/fe98cb6e7d579b1360eb856670ac08a2133131d9/oteps/4430-span-event-api-deprecation-plan.md

lalitb:
This is a nice, clean example, but I’m not sure it adds much on top of the existing HTTP client/server propagation example, which already covers the real-world use case.

Also, using `TextMapPropagator` + a carrier for async work within the same process feels a bit misleading. That pattern is generally meant for crossing process or transport boundaries, not for thread/async propagation. We also don’t prescribe a specific approach for async context propagation.

More broadly, I’m not sure we need a separate “generic carrier” example here. In practice, users almost always interact with propagation through concrete transports like HTTP or gRPC, and the HTTP example already demonstrates the pattern clearly.

As it stands, this feels a bit overlapping and may unintentionally suggest a pattern we don’t actually recommend.

examples/manual_propagation/main.cc
dbarker: Thanks! No worries. The other examples do use a different style at times. A new PR to apply the style guide to examples would be welcome.

examples/manual_propagation/main.cc
jgomezselles: Thanks for the info. I didn't know about it. Actually, your comment helped me being prepared for when I heard it last week at a talk in kubecon. Quite shocking, to be honest :)

jgomezselles:
Hi again,

Sorry for the delay. I was afk for a couple of weeks and it also took me a while to understand the feedback. It does make sense, so thanks. I really wanted to also add a part that took me some time to understand/make work for context propagation, and that kind of ruined the example. 

The part I mean is that context injection needs:
```cpp
  auto ctx_with_span = trace_api::SetSpan(current_ctx, span);
```
But again, I do agree with your comments about (ab)using this in the example. So let's forget about it.

I've changed everything, and now the example is called "explicit_parent" which is what I really want to show, and what I needed. It should be much more simple now: just creating 2 parent spans, and showing how to set the child for each of them explicitly, when the current context cannot be used.

Thanks for the help and guidance! I'll be waiting for feedback.

lalitb:
Thanks @jgomezselles  - This is much clearer now, and the `explicit_parent` focus is a better fit than the earlier propagation/carrier framing.

One suggestion: I would avoid using `std::async` in this example. For an explicit-parent example, the core idea is simply creating a child span from an explicitly supplied parent context, and the async machinery makes it read like an async context propagation pattern even though that is not what the example is actually demonstrating.

More importantly, passing the parent span by reference into the async task is not something we want to present as a standard approach. That may be fine for a minimal example, but it does not scale well to larger applications and could easily be read as recommended async propagation guidance.

If we want to demonstrate async propagation, I think that should be a separate example with that as the explicit goal, ideally around a custom `RuntimeContextStorage` implementation for an async runtime rather than `std::async` alone. That would keep this example tightly focused on explicit parent selection, and keep a future async example focused on actual context propagation semantics.

jgomezselles:
Hi @lalitb, thanks again for your patience and time reviewing and explaining.
I did have some troubles trying to figure out how would be the recommended way to pass the
context for async execution, and an example using the `RuntimeContextStorage` would definitely
help. Altogether in an async example.

I was using async because it demonstrated a way in which I _really_ needed it, but a simple
task scheduler has the same issue for the sake of an example. 
So I've now changed the async calls for a simple vector of functions working as task scheduler.

Thanks again!

examples/explicit_parent/main.cc
```diff
@@ -0,0 +1,130 @@
...
+  // In this example, we simulate handling of 2 independent traces that are alive at the same time.
+  // This may happen for various reasons. For example, 2 different tasks
+  // are scheduled to run within the same process, and we may want to nest
+  // spans to each of them, but cannot know which one will be active at runtime.
+  auto parent_1 = utils::CreateSpan("parent_1");
+  auto parent_2 = utils::CreateSpan("parent_2");
+
+  // To make sure that both the wanted parent-child relationship is met and
+  // the parent is kept alive, the parent span is passed.
+  std::vector<std::function<void()>> tasks = {
```
dbarker, line 108: Idea for a follow up PR if interested: One case that isn't well documented or demonstrated in the examples is explicit creation of a root span. This can be helpful/necessary to break up large traces and is called out in the spec here:  [specification/trace/api.md?plain=1#L388](https://github.com/open-telemetry/opentelemetry-specification/blob/49ef085d79fd4b9a0ff88135bb7ee6eed74fe8f3/specification/trace/api.md?plain=1#L388)

This example seems like a great place to show it. 

```cpp
 opentelemetry::context::Context context_root{opentelemetry::trace::kIsRootSpanKey, true};
    trace_api::StartSpanOptions options;
    options.parent   = context_root;
```

dbarker (approved):
Looks good. Thanks for the example. Just shared an idea below on a follow up PR if interested.

marcalff:
Thanks for the PR.

Waiting on @lalitb approval, since he is familiar with this change and commented already.

examples/explicit_parent/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The output will be a set of spans, in which you can check how the relationships
+outlined in the previous diagrams are fulfilled. For example:
+
+### Trace 1
+
+| name     | parent_span_id   | span_id          |
+| -------- | ---------------- | ---------------- |
+| parent_1 | 0000000000000000 | f026e45ec526047e |
+| nested   | f026e45ec526047e | d701525271ff5d72 |
```
ThomsonTan, line 61: nit: consider noting above the tables that `trace_id` / `span_id` will differ on each run - the shown values are illustrative.

jgomezselles: Ok, done!

examples/explicit_parent/main.cc
jgomezselles: Thank you very much for the info. Since this is taking a bit more time, I went ahead and added it directly. I hope that's ok.

marcalff:
@lalitb 

Could you confirm if this is ok to merge ? Thanks.

lalitb (approved):
Thanks.

lalitb:
@copilot resolve the merge conflicts in this pull request

marcalff:
@jgomezselles 

Thanks for the contribution.

Now merging, sorry the review took so long.

jgomezselles:
Thanks a lot @marcalff , @lalitb, @dbarker and everyone else for your time and patience during this review. I learnt a lot! I hope to get time soon to continue contributing with some of your suggestions in this thread.
Thanks for your work maintaining this project!

PR #3941, opened 2026-03-18
[CODE HEALTH] Fix clang-tidy performance warnings

dbarker:
Contributes to #2053

Fixes remaining clang-tidy performance-* warnings
-performance-unnecessary-value-param
-performance-avoid-endl
-performance-move-const-arg

## Changes

Fixes the following warnings: 

----
performance-unnecessary-value-param (9 warnings)
| File | Line | Message |
|---|---|---|
| `api/include/opentelemetry/context/context.h` | 37 | the parameter 'value' is copied for each invocation but only used as a const reference; consider making it a const reference |
| `api/include/opentelemetry/context/context.h` | 62 | parameter 'value' is passed by value and only copied once; consider moving it to avoid unnecessary copies |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 27 | the parameter 1 is copied for each invocation but only used as a const reference; consider making it a const reference |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 31 | the parameter 1 is copied for each invocation but only used as a const reference; consider making it a const reference |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 35 | the parameter 1 is copied for each invocation but only used as a const reference; consider making it a const reference |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 89 | parameter 'span_context' is passed by value and only copied once; consider moving it to avoid unnecessary copies |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 103 | parameter 'span_context' is passed by value and only copied once; consider moving it to avoid unnecessary copies |
| `api/include/opentelemetry/trace/span_context_kv_iterable_view.h` | 112 | parameter 'span_context' is passed by value and only copied once; consider moving it to avoid unnecessary copies |
| `sdk/include/opentelemetry/sdk/instrumentationscope/scope_configurator.h` | 49 | the parameter 'scope_matcher' is copied for each invocation but only used as a const reference; consider making it a const reference |
----
performance-avoid-endl (5 warnings)
| File | Line | Message |
|---|---|---|
| `examples/environment_carrier/main.cc` | 110 | do not use 'std::endl' with streams; use '\n' instead |
| `examples/environment_carrier/main.cc` | 111 | do not use 'std::endl' with streams; use '\n' instead |
| `examples/environment_carrier/main.cc` | 127 | do not use 'std::endl' with streams; use '\n' instead |
| `examples/environment_carrier/main.cc` | 128 | do not use 'std::endl' with streams; use '\n' instead |
| `examples/environment_carrier/main.cc` | 158 | do not use 'std::endl' with streams; use '\n' instead |
----
performance-move-const-arg (2 warnings)
| File | Line | Message |
|---|---|---|
| `exporters/otlp/test/otlp_grpc_metric_exporter_test.cc` | 73 | std::move of the const variable 'client' has no effect; remove std::move() or make the variable non-const |
| `exporters/otlp/test/otlp_grpc_metric_exporter_test.cc` | 81 | std::move of the const variable 'client' has no effect; remove std::move() or make the variable non-const |
----

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 220
+            warning_limit: 204
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 222
+            warning_limit: 206
     env:
       CC: /usr/bin/clang-18
       CXX: /usr/bin/clang++-18
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,9 @@ Increment the:
 * [TEST] CMake component install test for resource_detectors
   [#3940](https://github.com/open-telemetry/opentelemetry-cpp/pull/3940)
 
+* [CODE HEALTH] Fix clang-tidy performance warnings
+  [#3941](https://github.com/open-telemetry/opentelemetry-cpp/pull/3941)
+
 Important changes:
 
 * [BUILD] Revisit EventLogger deprecation
--- a/examples/environment_carrier/main.cc
+++ b/examples/environment_carrier/main.cc
@@ -128,7 +128,7 @@ void RunParent()
   std::cout << "[parent] trace_id: " << Hex(root_span->GetContext().trace_id()) << '\n';
   std::cout << "[parent] span_id:  " << Hex(root_span->GetContext().span_id()) << '\n';
   std::cout << std::flush;
-  
+
   // Inject context into a map via EnvironmentCarrier
   auto env_map = std::make_shared<std::map<std::string, std::string>>();
   propagation::EnvironmentCarrier carrier(env_map);


api/include/opentelemetry/context/context.h
```diff
@@ -125,13 +125,13 @@ class Context
 
     // Builds a data list with just a key and value, so it will just be the head
     // and returns that head.
-    DataList(nostd::string_view key, const ContextValue &value)
+    DataList(nostd::string_view key, ContextValue value)
```
marcalff, line 128: This is changing _how_ a DataList object is built, but not the memory layout of the object itself, so I think this is ok and compatible w.r.t ABI changes.

dbarker: Agreed. DataList is private and the public API for Context remains the same.  This change just allows the context value to be moved instead of copied whenever a Context is created or SetValue is called.

PR #3953, opened 2026-03-26
Enable WITH_OTLP_RETRY_PREVIEW by default

### LINKED ISSUE #3481: Enable WITH_OTLP_RETRY_PREVIEW by default
This feature was introduced by #3223, and is considered stable.

* Enable the flag by default
* Mark the flag as deprecated
* Plan for flag removal

marcalff:
Fixes #3481

## Changes

### DIFF
--- a/DEPRECATED.md
+++ b/DEPRECATED.md
@@ -63,6 +63,8 @@ necessary.
 
 Remove compilation flag WITH_OTLP_RETRY_PREVIEW in CMake.
 
+Remove ifdef ENABLE_OTLP_RETRY_PREVIEW in C++ .
+
 #### Mitigation (WITH_OTLP_RETRY_PREVIEW)
 
 Applications built with `WITH_OTLP_RETRY_PREVIEW=ON` need to:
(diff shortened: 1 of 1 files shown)


api/CMakeLists.txt
```diff
@@ -84,6 +84,8 @@ target_compile_definitions(
 if(WITH_OTLP_RETRY_PREVIEW)
   target_compile_definitions(opentelemetry_api
```
dbarker, line 85: Do we also need to set the `ENABLE_OTLP_RETRY_PREVIEW` compile definition for bazel builds now?

api/CMakeLists.txt
```diff
@@ -84,6 +84,8 @@ target_compile_definitions(
...
                              INTERFACE ENABLE_OTLP_RETRY_PREVIEW)
+else()
+  message(WARNING "WITH_OTLP_RETRY_PREVIEW=OFF is deprecated")
```
ThomsonTan, line 88: Probbably also put a date in the deprecation message? So the user will be aware of the timeline.

marcalff: This is covered in file DEPRECATED.md

I will create a pinned issue to point to all pending deprecations, and/or also mention this in release notes in general.

api/CMakeLists.txt
marcalff: @dbarker 

Good question.

The best behavior is to define `ENABLE_OTLP_RETRY_PREVIEW` as well in the bazel build, so that bazel and CMake are on par.

Unfortunately, doing so unconditionally prevents users to opt out for now, increasing the risk to have a disruptive change forced upon applications with this PR.

What is needed is a bazel option to enable or disable the feature, but we have none.

It is too late and not worth adding a bazel flag at this time, since it will cause confusion, only to remove it later.

ThomsonTan (approved):

Issue #3013, opened 2024-07-26
[API] Clarify expectations for C++ exceptions, and usage of noexcept methods.


marcalff:
Several methods in the API are flagged as `noexcept`.

This is desirable, because adding instrumentation to an application (i.e., calling opentelemetry-cpp apis) should not make the application less stable.

In particular, any failure in the opentelemetry-cpp sdk or exporters should not propagate the exception up, taking the application down.

To comply with the `noexcept` contract, methods in the SDK implementation should never raise exceptions.

According to clang-tidy reports, this is not always the case.

This part should be revisited, to clarify expectations, and enforce the SDK implementation complies.

cc @msiddhu

marcalff:
Related:

* #3012

msiddhu:
I'll work on this. Currently, there are 20 warnings flagged by `clang-tidy ` as they may raise errors. I'll make a sheet explaining why and will come up with a cleanup plan. I'm busy this week but will raise a PR by late next week.

Andrew-Brock:
Hi @msiddhu, did you get anywhere in generating the spreadsheet for this? I'm interested in the status of it because in the current state, the functions being marked as `noexcept` ironically make the API more likely to introduce instability due to it now calling `std::terminate` rather than giving me a chance to handle the exception, especially in the case of the _Observable_ metrics in [sdk::metrics::ObserverResultT::Observe](https://github.com/open-telemetry/opentelemetry-cpp/blob/4998eb178601f69481e1629e5464d0272532ec9e/sdk/include/opentelemetry/sdk/metrics/observer_result.h#L31-L40) where the exception would be off in its own thread.

I also see that in places where this is attempted to be handled, such as in https://github.com/open-telemetry/opentelemetry-cpp/pull/3012, that the exception handling can also throw exceptions such as when building up the `std::stringstream tmp_stream` or in the registered `LogHandler::Handle`, as-per the definition of the [OTEL_INTERNAL_LOG_DISPATCH](https://github.com/open-telemetry/opentelemetry-cpp/blob/4998eb178601f69481e1629e5464d0272532ec9e/sdk/include/opentelemetry/sdk/common/global_log_handler.h#L160-L162) macro.

I'm happy to submit some small PRs to address some but your comment hints that some of these may not be so trivial to fix.

Andrew-Brock:
I've come across another case where exceptions may be thrown in non-obvious ways in functions marked as `noexcept`.

[...]

PR #3964, opened 2026-04-01
[CODE HEALTH] Fix clang-tidy bugprone-exception-escape warnings in API

### LINKED ISSUE #3981: [CODE HEALTH] clang-tidy reports bugprone-exception-escape warnings in API
clang-tidy reports the following warnings in the API. These methods are all marked `noexcept` and will result in `std::terminate` being called if an exception is thrown. 

----

**opentelemetry-cpp/api/include/opentelemetry/trace/trace_state.h** (4 warnings)

| Line | Check | Message |
|---|---|---|
| 55 | `bugprone-exception-escape` | an exception may be thrown in function 'FromHeader' which should not throw exceptions |
| 118 | `bugprone-exception-escape` | an exception may be thrown in function 'Get' which should not throw exceptions |
| 138 | `bugprone-exception-escape` | an exception may be thrown in function 'Set' which should not throw exceptions |
| 172 | `bugprone-exception-escape` | an exception may be thrown in function 'Delete' which should not throw exceptions |

----

dbarker:
Fixes #3981

Contributes to #2053, #3013

## Changes

- Replaces usage of `__EXCEPTION` with `OPENTELEMETRY_HAVE_EXCEPTIONS`
- Makes trace_state.h static methods noexcept and adds required try/catch
- Add unit test coverage for cases with string_view, string_util, and trace state.
- Fixes the following warnings: 

----

**bugprone-exception-escape** (4 warnings)

| File | Line | Message |
|---|---|---|
| `opentelemetry-cpp/api/include/opentelemetry/trace/trace_state.h` | 55 | an exception may be thrown in function 'FromHeader' which should not throw exceptions |
| `opentelemetry-cpp/api/include/opentelemetry/trace/trace_state.h` | 118 | an exception may be thrown in function 'Get' which should not throw exceptions |
| `opentelemetry-cpp/api/include/opentelemetry/trace/trace_state.h` | 138 | an exception may be thrown in function 'Set' which should not throw exceptions |
| `opentelemetry-cpp/api/include/opentelemetry/trace/trace_state.h` | 172 | an exception may be thrown in function 'Delete' which should not throw exceptions |

----


api/include/opentelemetry/common/macros.h
```diff
@@ -420,6 +420,31 @@ point.
...
+//     {
+//       return false;
+//     }
+//   }
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
+#  define OPENTELEMETRY_TRY try
+#  define OPENTELEMETRY_CATCH_ALL catch (...)
+#else
+#  define OPENTELEMETRY_TRY if (true)
+#  define OPENTELEMETRY_CATCH_ALL else if (false)
```
marcalff: `else if (false)` confuses the compiler, expecting another else.

Just `else` should be sufficient.

dbarker: Fixed

marcalff (commented):
The change to replace `nostd::holds_alternative` with `nostd::get_if` is clean and a no brainer.

Please prepare a separate patch to fix all `nostd::holds_alternative` so this part can be merged out of the way.

This PR can then focus on the try/catch code alone, as it will be more tricky to resolve.

api/include/opentelemetry/common/macros.h
```diff
@@ -420,6 +420,44 @@ point.
...
+#if defined(__cpp_lib_unreachable) && (__cpp_lib_unreachable >= 202202L)
+#  include <utility>
+#  define OPENTELEMETRY_UNREACHABLE() std::unreachable()
+#elif OPENTELEMETRY_HAVE_BUILTIN(__builtin_unreachable)
+#  define OPENTELEMETRY_UNREACHABLE() __builtin_unreachable()
+#elif defined(_MSC_VER)
+#  define OPENTELEMETRY_UNREACHABLE() __assume(0)
+#else
+#  include <cstdlib>
+#  define OPENTELEMETRY_UNREACHABLE() abort()
```
dbarker: Should the fallback abort or something else?

api/include/opentelemetry/nostd/string_view.h
```diff
@@ -71,7 +71,7 @@ class string_view
   {
     if (pos > length_)
     {
-#  if __EXCEPTIONS
+#  if OPENTELEMETRY_HAVE_EXCEPTIONS
```
dbarker, line 74: The behavior seems to have been inconsistent in the past between compilers (msvc is not setting __EXCEPTIONS).  Changing to use the common macro seems reasonable and expected here, but is a change of behavior on msvc (which would now compile with this throw for string_view unless exceptions are disabled).

dbarker:
> The change to replace nostd::holds_alternative with nostd::get_if is clean and a no brainer.
> 
> Please prepare a separate patch to fix all nostd::holds_alternative so this part can be merged out of the way.
> 
> This PR can then focus on the try/catch code alone, as it will be more tricky to resolve.

Thanks @marcalff. I've split this up. The nostd::variant access changes are now in #3965. Interested in feedback on the try/catch/unreachable macros given our need to support noexcept builds and analyzers.

api/include/opentelemetry/common/string_util.h
```diff
@@ -25,7 +26,15 @@ class StringUtil
     {
       right--;
     }
-    return str.substr(left, 1 + right - left);
+    OPENTELEMETRY_TRY
+    {
+      return str.substr(left, 1 + right - left);
+    }
+    OPENTELEMETRY_CATCH_ALL
```
dbarker: Once we start to add these to the sdk we may want to expand these macros to support passing information to the logger (exception message and other context from the call site).

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -171,25 +193,35 @@ class OPENTELEMETRY_EXPORT TraceState
    */
   nostd::shared_ptr<TraceState> Delete(const nostd::string_view &key) noexcept
   {
-    if (!IsValidKey(key))
+    OPENTELEMETRY_TRY
     {
-      return TraceState::GetDefault();
+      if (!IsValidKey(key))
```
dbarker: One alternative to consider is to make IsValidKey/IsValidValue noexcept and put the try catch in those methods. That is an API impact but perhaps not an ABI change. The std::regex ctor and std::regex_match may throw.

marcalff:
@dbarker 

Thanks for working on this, this is a delicate topic in opentelemetry-cpp.

Some background information, as I understand the history, to help designing the proper fix, and some general comments below.

# 1. Otel Instrumentation should not kill the application

When an application is instrumented with otel, it makes API calls to otel-cpp.
These calls are to collect data (SDK installed and configured), or not (SDK not installed, or discarding data), and the application does not expect data returned from these API calls.

Whatever happens within otel-cpp should not put the application at risk, which is why every API entry point is designed to be `noexcept`.

# 2. The application code can not be trusted to be robust

Every API that returns a pointer (say, a span) can never return a null pointer, because we can not expect the application code to systematically check for null pointer conditions on every single line instrumented.

Constantly checking pointers would be very tedious on the application side, and unsafe, possibly causing crashes if a check is missing.

For this reason, otel-cpp will never return a null pointer, but will return a "no-op" object instead, that the instrumented code can call  without precautions.

In other words, we don't want to impose a defensive programming style on the application, when adding opentelemetry instrumentation.

# 3. Memory allocation and noexcept

Points 1 and 2 are already causing severe constraints, and I am not even sure it is possible to safely return a pointer on a noop object while not raising exceptions at the same time.

When an API returns a `nostd::unique_ptr<T>` or `nostd::shared_ptr<T>`, is needs to allocate memory, which can raise exceptions, making this hard to have an API typed as `noexcept`.

Putting aside exceptions caused by Out Of Memory, at least we should make an effort to not raise any other exceptions then.

# 4. Different build flavors

Applications can be built with, or without, exception support, and we need to accommodate both.

Given that we support many compilers/platforms as well, hiding all the details behind `OPENTELEMETRY_HAVE_EXCEPTIONS` sounds a good thing.

# 5. Places that need a try-catch block

There should be very few places handling exceptions in the API itself.

Most of the time, the API consist of virtual methods implemented by the SDK, so there is no code.

When code is actually provided in-lined in the API, this is for very specific places, like:

* The runtime context
* Propagation

so I would expect only a very few try-catch blocks in the whole API.

# 6. try-catch block style

Instead of hiding all the details behind macros for the try, and more importantly the catch block, we can have more explicit code.

For example:

```
#if OPENTELEMETRY_HAVE_EXCEPTIONS
try
#endif
{
  // Regular code
}
#if OPENTELEMETRY_HAVE_EXCEPTIONS
catch (...)
{
  // catch code, only used when building with exceptions
}
#endif
```

This will avoid confusing the compiler, clang-tidy, cpp-check, and avoid complaints that a return code path is missing, code is not reachable, etc, when building with/without exceptions.

Given (4) with very few try-catch blocks needed, I think this will work better.

# 7. What to do when an exception is seen ?

Nothing.

Do not try to log it, do not call abort, nothing.

We can not log anything from the API (the SDK may not even be present), and we definitively can not take the application down with an abort.

The worst case is that opentelemetry-cpp will behave like a noop brick if it fails internally.

We can have an assert in debug or in maintainer mode to catch failures in CI, but can not do more than that.

# 8. Internal supporting code does not need to be noexcept

Code in the actual API surface, used by the application, needs to be `noexcept` (tracer provider, tracer, span, etc).

Supporting code, like `StringUtil::Trim`, is not supposed to be used directly, it is there to implement other apis.

Because of this, I think we can relax internal helpers to not be `noexcept`, instead of adding try-catch blocks inside the helper.

dbarker:
Thanks for the feedback @marcalff. The points makes sense. I'll remove the macros in the PR and we can follow up in a SIG meeting on next steps more broadly for error handling. 

**1. Otel Instrumentation should not kill the application**

Your point is focused on instrumentation (using the API) and makes sense. The spec does a good job of highlighting this, and allows cases where the API/SDK may terminate the application on initialization (See [error handling](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles))

> 1. API methods MUST NOT throw unhandled exceptions when used incorrectly by end users.
>    The API and SDK SHOULD provide safe defaults for missing or invalid arguments.
>    For instance, a name like `empty` may be used if the user passes in `null` as the span name argument during `Span` construction.
> 2. The API or SDK MAY _fail fast_ and cause the application to fail on initialization, e.g. because of a bad user config or environment, but MUST NOT cause the application to fail later at runtime, e.g. due to dynamic config settings received from the Collector.
> 3. The SDK MUST NOT throw unhandled exceptions for errors in their own operations.
>    For example, an exporter should not throw an exception when it cannot reach the endpoint to which it sends telemetry data.

Failing fast on initialization and perhaps user configurable error response (fail loud or silently) does seem desirable (see last point below). 

**3. Memory allocation and noexcept**

Memory allocation is an area for discussion and the codebase is mixed. In most cases memory is allocated with `new` alone which may throw. Some cases use `new (std::nothrow)`, some use placement new, and some use the modern `std::make_unique/shared`. 

**4. Different build flavors**

> Applications can be built with, or without, exception support, and we need to accommodate both.

Is building without exception support a requirement just for the API or to include the SDK, exporters, and detectors? Given the dependencies the project has we may not be able to provide an exception free build of all components that meets the requirements of point 1. The bazel noexept build doesn't include all components currently. 

**7. What to do when an exception is seen ?**

> We can have an assert in debug or in maintainer mode to catch failures in CI

This makes sense and I think should be a requirement for the project (ci should fail on error that would be ignored in production, a maintainer mode flag is a great idea). 

The spec gives some guidance that extends this to users who may also want a clear signal that something is going wrong. 
> SDK implementations MUST allow end users to change the library's default error handling behavior for relevant errors.
> Application developers may want to run with strict error handling in a staging environment to catch invalid uses of the API, or malformed config.
> Note that configuring a custom error handler in this way is the only exception to the basic error handling principles outlined above.
> The mechanism by which end users set or register a custom error handler should follow language-specific conventions.

api/include/opentelemetry/common/string_util.h
dbarker, line 19: @marcalff this is a good example of a method that may be considered an implementation detail for the API but is currently public. Would removing noexcept from this method be acceptable?

api/include/opentelemetry/plugin/detail/utility.h
```diff
@@ -3,36 +3,34 @@
...
 {
 namespace detail
 {
 inline void CopyErrorMessage(const char *source, std::string &destination) noexcept
-#if __EXCEPTIONS
-try
-#endif
 {
-  if (source == nullptr)
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
```
marcalff, line 17: The `opentelemetry::plugin::detail` namespace is implementation details, not supposed to be used directly.

It should be possible to relax this method to not be noexcept, if it makes the code simpler.

Beside, but this is a separate discussion, I think the whole plugin part should be deprecated and removed from the code base, so it it ok to leave and silence clang-tidy messages in the mean time.

api/include/opentelemetry/trace/trace_state.h
```diff
@@ -117,11 +128,21 @@ class OPENTELEMETRY_EXPORT TraceState
    */
   bool Get(nostd::string_view key, std::string &value) const noexcept
   {
-    if (!IsValidKey(key))
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
+    try
+#endif
+    {
+      if (!IsValidKey(key))
```
marcalff, line 135: If is possible to change IsValidKey() to be noexecpt instead ?

dbarker: These static methods in trace_state.h are now marked noexcept and required try/catch has been added.

api/include/opentelemetry/plugin/detail/utility.h
dbarker: I've kept the changes here. The intent is to remove direct usage of `__EXCEPTIONS`  and replace with the common `OPENTELEMETRY_HAVE_EXCEPTIONS`

api/test/common/string_util_test.cc
```diff
@@ -51,3 +52,12 @@ TEST(StringUtilTest, TrimString)
...
   }
 }
+
+TEST(StringUtilTest, TrimStringOutOfRange)
+{
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
+  EXPECT_EQ(StringUtil::Trim("x", 2, 1), "");
+#else
+  EXPECT_DEATH(StringUtil::Trim("x", 2, 1), "");
+#endif
```
marcalff: Instead of DEATH when exceptions are not supported, could `StringUtil::Trim` be more defensive and check invalid parameters, and return an empty string in that case ?

This test should then becomes, in all cases:

```
EXPECT_EQ(StringUtil::Trim("x", 2, 1), "");
```

I think this will reduce complexity and improve robustness, which are both desirable.

marcalff (commented):
LGTM, see one comment

api/test/common/string_util_test.cc
dbarker: `nostd::string_view::substr` calls `std::terminate` if `OPENTELEMETRY_HAVE_EXCEPTIONS` is not set. I can add defensive checks to `StringUtil::Trim` to ensure that condition is never hit but we may just want to address this in `nostd::string_view` to prevent a lot of duplicate defensive checks in the hot path (string parsing).

dbarker: I've added defensive checks to `StringUtil::Trim`. The `std::terminate` in `nostd::string_view::substr` seems to be the same as `std::string_view` if built without exceptions.

marcalff (approved):
LGTM, thanks for the fix.

PR #3965, opened 2026-04-03
[CODE HEALTH] Cleanup nostd variant access for noexcept methods in api and sdk

dbarker:
Contributes to #2053

## Changes

- replace `nostd::holds_alternative`/`nostd::get` pattern with `nostd::get_if` in api and sdk. 
- unit test for trace/context.h
- expanded test cases for sdk/logger to cover trace metadata and sdk/tracer to cover root contexts
- fixes the following warnings
----
| File | Line | Message |
|---|---|---|
| `api/include/opentelemetry/nostd/function_ref.h` | 81 | forwarding reference parameter 'f' is never forwarded inside the function body |
| `api/include/opentelemetry/trace/context.h` | 26 | an exception may be thrown in function 'IsRootSpan' which should not throw exceptions |
| `sdk/src/logs/logger.cc` | 59 | an exception may be thrown in function 'CreateLogRecord' which should not throw exceptions |
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ For edit access, get in touch on
 
 ### Maintainers
 
+* [Doug Barker](https://github.com/dbarker)
 * [Ehsan Saei](https://github.com/esigo)
 * [Lalit Kumar Bhasin](https://github.com/lalitb), Microsoft
 * [Marc Alff](https://github.com/marcalff), Oracle
@@ -108,7 +109,6 @@ For more information about the maintainer role, see the [community repository](h
 
 ### Approvers
 
-* [Doug Barker](https://github.com/dbarker)
 * [Josh Suereth](https://github.com/jsuereth), Google
 * [Pranav Sharma](https://github.com/psx95), Google
 * [WenTao Ou](https://github.com/owent), Tencent


sdk/src/logs/logger.cc
```diff
@@ -67,37 +64,14 @@ opentelemetry::nostd::unique_ptr<opentelemetry::logs::LogRecord> Logger::CreateL
...
-          opentelemetry::nostd::get<opentelemetry::nostd::shared_ptr<opentelemetry::trace::Span>>(
-              context_value);
-      if (data)
-      {
-        recordable->SetTraceId(data->GetContext().trace_id());
-        recordable->SetTraceFlags(data->GetContext().trace_flags());
-        recordable->SetSpanId(data->GetContext().span_id());
-      }
-    }
-    else if (opentelemetry::nostd::holds_alternative<
```
dbarker, line 88: I could not find an example or test where opentelemetry::trace::kSpanKey is set to a trace::SpanContext in the runtime context and removed this branch. Does it need to be supported?

marcalff: I have not seen places in the API that sets the `opentelemetry::trace::kSpanKey` with a `trace::SpanContext` either, to this _appears_ to be dead code.

However, some work is still needed to implement the full sampling feature (ComposableSamplers) and the trace context level 2, which is likely to impact this area.

Please keep support for `trace::SpanContext`, with a TODO note to research this further.

Related:

* https://github.com/open-telemetry/opentelemetry-specification/pull/4321
* https://github.com/open-telemetry/opentelemetry-cpp/issues/2283

dbarker: Thanks Marc. I've added back the path to pull a span context directly from the runtime context for log record creation with a TODO. 

I think the two spec changes shared are supported with the nominal case where the span metadata is pulled from the span directly. The random trace id flag seems to already be set so we may be able to check that one off for spec compliance (https://github.com/open-telemetry/opentelemetry-cpp/blob/26c774b09f30518213c3a517a177c3975760f269/sdk/src/trace/tracer.cc#L119)

marcalff:
@dbarker 

Please merge this PR, to validate you now have proper permissions.

Please use "Squash and merge" and remove the history for each individual commit / merge, to only preserve the final result.

PR #3973, opened 2026-04-08
[CODE HEALTH] clean up nostd::variant access in otlp exporters

dbarker:
Fixes # (issue)

## Changes
- clean up use of `nostd::holds_alternative`/`nostd::get` if/else pattern 
   - replace with `nostd::visit` in otlp AnyValue attribute utils
   - replace with `nostd::get_if` in otlp http client and otlp metric utils
- Add tests for `OtlpPopulateAttributeUtils`
- minor updates to `OtlpMetricUtils` to reduce data point copies and use const
- update conversion of uint64 attribute values to match the standard (map to string on overflow of int). 
   - See: https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/#integer-values 


exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -57,307 +63,331 @@ inline void SetUint64Value(opentelemetry::proto::common::v1::AnyValue *proto_val
...
+};
+
+}  // namespace
+
+void OtlpPopulateAttributeUtils::PopulateAnyValue(
+    opentelemetry::proto::common::v1::AnyValue *proto_value,
+    const opentelemetry::common::AttributeValue &value,
+    bool allow_bytes) noexcept
+{
+#if defined(OPENTELEMETRY_HAVE_EXCEPTIONS)
```
ThomsonTan, line 340: Should explicitly check the value as non-zero?

dbarker:
closing this draft. Will break up into smaller PRs.

PR #3991, opened 2026-04-12
[EXPORTER] Implement non-utf8 string to bytes in OTLP

### LINKED ISSUE #3491: Need validation/debug logs for protobuf setters

owent:
Fixes #3491

## Changes

- Mapping non UTF8 string to bytes in OLTP exporters.

> https://github.com/open-telemetry/opentelemetry-specification/blame/v1.55.0/specification/common/attribute-type-mapping.md.
> Reimplement #3512 to meet the latest spec.

### DIFF
--- a/cmake/protobuf.cmake
+++ b/cmake/protobuf.cmake
@@ -108,6 +108,12 @@ message(STATUS "PROTOBUF_PROTOC_EXECUTABLE=${PROTOBUF_PROTOC_EXECUTABLE}")
 # but may need the alias target created
 if(TARGET utf8_validity AND NOT TARGET utf8_range::utf8_validity)
   add_library(utf8_range::utf8_validity ALIAS utf8_validity)
+  if(protobuf_SOURCE_DIR AND EXISTS
+                             "${protobuf_SOURCE_DIR}/third_party/utf8_range")
+    target_include_directories(
+      utf8_validity
+      PUBLIC "$<BUILD_INTERFACE:${protobuf_SOURCE_DIR}/third_party/utf8_range>")
+  endif()
 endif()
 
 # For the find_package(Protobuf) path, utf8_range can also be found
--- a/cmake/protobuf.cmake
+++ b/cmake/protobuf.cmake
@@ -1,29 +1,37 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+# Import Protobuf targets (protobuf::libprotobuf and protobuf::protoc) and set
+# PROTOBUF_PROTOC_EXECUTABLE. 1. If gRPC was fetched from github then use the
+# Protobuf submodule built with gRPC 2. Find an installed Protobuf package 3.
+# Use FetchContent to fetch and build Protobuf from GitHub
 
-# Import Protobuf targets (protobuf::libprotobuf and protobuf::protoc) and set PROTOBUF_PROTOC_EXECUTABLE.
-# 1. If gRPC was fetched from github then use the Protobuf submodule built with gRPC
-# 2. Find an installed Protobuf package
-# 3. Use FetchContent to fetch and build Protobuf from GitHub
-
-if(DEFINED gRPC_PROVIDER AND NOT gRPC_PROVIDER STREQUAL "find_package" AND TARGET libprotobuf)
+if(DEFINED gRPC_PROVIDER
+   AND NOT gRPC_PROVIDER STREQUAL "find_package"
+   AND TARGET libprotobuf)
   # gRPC was fetched and built Protobuf as a submodule
 
-  set(_Protobuf_VERSION_REGEX "\"cpp\"[ \t]*:[ \t]*\"([0-9]+\\.[0-9]+(\\.[0-9]+)?)\"")
-  set(_Protobuf_VERSION_FILE "${grpc_SOURCE_DIR}/third_party/protobuf/version.json")
+  set(_Protobuf_VERSION_REGEX
+      "\"cpp\"[ \t]*:[ \t]*\"([0-9]+\\.[0-9]+(\\.[0-9]+)?)\"")
+  set(_Protobuf_VERSION_FILE
+      "${grpc_SOURCE_DIR}/third_party/protobuf/version.json")
 
   file(READ "${_Protobuf_VERSION_FILE}" _Protobuf_VERSION_FILE_CONTENTS)
   if(_Protobuf_VERSION_FILE_CONTENTS MATCHES ${_Protobuf_VERSION_REGEX})
     set(Protobuf_VERSION "${CMAKE_MATCH_1}")
   else()
-    message(WARNING "Failed to parse Protobuf version from ${_Protobuf_VERSION_FILE} using regex ${_Protobuf_VERSION_REGEX}")
+    message(
+      WARNING
+        "Failed to parse Protobuf version from ${_Protobuf_VERSION_FILE} using regex ${_Protobuf_VERSION_REGEX}"
+    )
   endif()
   set(Protobuf_PROVIDER "grpc_submodule")
 else()
 
-  # Search for an installed Protobuf package explicitly using the CONFIG search mode first followed by the MODULE search mode.
-  # Protobuf versions < 3.22.0 may be found using the module mode and some protobuf apt packages do not support the CONFIG search.
+  # Search for an installed Protobuf package explicitly using the CONFIG search
+  # mode first followed by the MODULE search mode. Protobuf versions < 3.22.0
+  # may be found using the module mode and some protobuf apt packages do not
+  # support the CONFIG search.
 
   find_package(Protobuf CONFIG QUIET)
   set(Protobuf_PROVIDER "find_package")
@@ -36,33 +44,48 @@ else()
     FetchContent_Declare(
       "protobuf"
       GIT_REPOSITORY "https://github.com/protocolbuffers/protobuf.git"
-      GIT_TAG "${protobuf_GIT_TAG}"
-    )
+      GIT_TAG "${protobuf_GIT_TAG}")
 
-    set(protobuf_INSTALL ${OPENTELEMETRY_INSTALL} CACHE BOOL "" FORCE)
-    set(protobuf_BUILD_TESTS OFF CACHE BOOL "" FORCE)
-    set(protobuf_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+    set(protobuf_INSTALL
+        ${OPENTELEMETRY_INSTALL}
+        CACHE BOOL "" FORCE)
+    set(protobuf_BUILD_TESTS
+        OFF
+        CACHE BOOL "" FORCE)
+    set(protobuf_BUILD_EXAMPLES
+        OFF
+        CACHE BOOL "" FORCE)
 
     FetchContent_MakeAvailable(protobuf)
 
     set(Protobuf_PROVIDER "fetch_repository")
 
     # Set the Protobuf_VERSION variable from the git tag.
-    string(REGEX REPLACE "^v([0-9]+\\.[0-9]+\\.[0-9]+)$" "\\1" Protobuf_VERSION "${protobuf_GIT_TAG}")
+    string(REGEX REPLACE "^v([0-9]+\\.[0-9]+\\.[0-9]+)$" "\\1" Protobuf_VERSION
+                         "${protobuf_GIT_TAG}")
 
     if(TARGET libprotobuf)
-      set_target_properties(libprotobuf PROPERTIES POSITION_INDEPENDENT_CODE ON CXX_CLANG_TIDY "" CXX_INCLUDE_WHAT_YOU_USE "")
+      set_target_properties(
+        libprotobuf
+        PROPERTIES POSITION_INDEPENDENT_CODE ON
+                   CXX_CLANG_TIDY ""
+                   CXX_INCLUDE_WHAT_YOU_USE "")
     endif()
 
   endif()
 endif()
 
 if(NOT TARGET protobuf::libprotobuf)
-  message(FATAL_ERROR "A required protobuf target (protobuf::libprotobuf) was not imported")
+  message(
+    FATAL_ERROR
+      "A required protobuf target (protobuf::libprotobuf) was not imported")
 endif()
 
 if(PROTOBUF_PROTOC_EXECUTABLE AND NOT Protobuf_PROTOC_EXECUTABLE)
-  message(WARNING "Use of PROTOBUF_PROTOC_EXECUTABLE is deprecated. Please use Protobuf_PROTOC_EXECUTABLE instead.")
+  message(
+    WARNING
+      "Use of PROTOBUF_PROTOC_EXECUTABLE is deprecated. Please use Protobuf_PROTOC_EXECUTABLE instead."
+  )
   set(Protobuf_PROTOC_EXECUTABLE "${PROTOBUF_PROTOC_EXECUTABLE}")
 endif()
 
@@ -70,7 +93,9 @@ if(CMAKE_CROSSCOMPILING)
   find_program(Protobuf_PROTOC_EXECUTABLE protoc)
 else()
   if(NOT TARGET protobuf::protoc)
-    message(FATAL_ERROR "A required protobuf target (protobuf::protoc) was not imported")
+    message(
+      FATAL_ERROR
+        "A required protobuf target (protobuf::protoc) was not imported")
   endif()
   set(Protobuf_PROTOC_EXECUTABLE "$<TARGET_FILE:protobuf::protoc>")
 endif()
@@ -78,3 +103,23 @@ endif()
 set(PROTOBUF_PROTOC_EXECUTABLE "${Protobuf_PROTOC_EXECUTABLE}")
 
 message(STATUS "PROTOBUF_PROTOC_EXECUTABLE=${PROTOBUF_PROTOC_EXECUTABLE}")
+
+# When protobuf is fetched or built as a gRPC submodule utf8_range is compiled
+# but may need the alias target created
+if(TARGET utf8_validity AND NOT TARGET utf8_range::utf8_validity)
+  add_library(utf8_range::utf8_validity ALIAS utf8_validity)
+endif()
+
+# For the find_package(Protobuf) path, utf8_range can also be found
+if(NOT TARGET utf8_range::utf8_validity)
+  find_package(utf8_range CONFIG QUIET)
+endif()
+
+# The utf8_range target should now be available for use in otel-cpp
+if(NOT TARGET utf8_range::utf8_validity)
+  message(
+    WARNING
+      "Target (utf8_range::utf8_validity) was not found. "
+      "We will not validate UTF-8 strings in the OTLP exporter."
+      "Which may lead to record drops when sending non UTF-8 data as string.")
+endif()
--- a/cmake/protobuf.cmake
+++ b/cmake/protobuf.cmake
@@ -1,37 +1,29 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-# Import Protobuf targets (protobuf::libprotobuf and protobuf::protoc) and set
-# PROTOBUF_PROTOC_EXECUTABLE. 1. If gRPC was fetched from github then use the
-# Protobuf submodule built with gRPC 2. Find an installed Protobuf package 3.
-# Use FetchContent to fetch and build Protobuf from GitHub
 
-if(DEFINED gRPC_PROVIDER
-   AND NOT gRPC_PROVIDER STREQUAL "find_package"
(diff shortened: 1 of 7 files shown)


exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -207,24 +242,23 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
...
-    else
+    auto array_value = proto_value->mutable_array_value();
+    for (const auto &val : nostd::get<std::vector<uint8_t>>(value))
     {
-      auto array_value = proto_value->mutable_array_value();
-      for (const auto &val : nostd::get<std::vector<uint8_t>>(value))
-      {
-        array_value->add_values()->set_int_value(val);
-      }
+      array_value->add_values()->set_int_value(val);
```
dbarker, line 248: This one may need discussion. I think `std::vector<uint8_t>` and `nostd::span<uint8_t>` should be mapped to protobuf bytes values. 

The enums for the attribute variants use `kTypeSpanByte` for these giving the intent that these are byte sequences (See [AttributeType](https://github.com/open-telemetry/opentelemetry-cpp/blob/34b4bae786718467680e98997400e34f514aab0c/api/include/opentelemetry/common/attribute_value.h#L81)). 

The spec calls out byte sequences and how they are exceptions to Array Value rule: 
> 
> Byte Sequences
> Byte sequences (e.g. Go's []byte slice or raw byte content of a file) SHOULD be converted to AnyValue's [bytes_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L37) field.
> 
> ... 
> 
> Array Values
> Values that represent ordered sequences of other values (such as [arrays](https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html), [vectors](https://en.cppreference.com/w/cpp/container/vector.html), ordered [lists](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists), [slices](https://go.dev/ref/spec#Slice_types)) SHOULD be converted to AnyValue's [array_value](https://github.com/open-telemetry/opentelemetry-proto/blob/38b5b9b6e5257c6500a843f7fdacf89dd95833e8/opentelemetry/proto/common/v1/common.proto#L35) field. String values and byte sequences are an exception from this rule (see above).

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -81,28 +104,34 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
...
-      }
+      proto_value->set_bytes_value(str_value.data(), str_value.size());
+    }
+  }
+  else if (nostd::holds_alternative<nostd::span<const uint8_t>>(value))
+  {
+    auto array_value = proto_value->mutable_array_value();
+    for (const auto &val : nostd::get<nostd::span<const uint8_t>>(value))
+    {
+      array_value->add_values()->set_int_value(val);
```
dbarker, line 134: Should be bytes? - see other comments about the enum and spec.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -26,6 +32,24 @@
 
 namespace nostd = opentelemetry::nostd;
 
+#if defined(OPENTELEMETRY_MACRO_UTF8_RANGE_LEGACY_IMPLEMENTATION)
```
dbarker, line 35: Supporting the legacy protobuf (and utf8 range) seems to drive a lot of complexity in the PR. Must the project support utf8 validation on Protobuf <3.22? 

Protobuf 3.22 adds the utf8_range package into the protobuf repo. It includes the  `<utf8_validity.h>` header, `utf8_range::IsStructurallyValid` method, and is installed as cmake package along with Protobuf. 

I'd advocate we bump the minimum supported Protobuf version from 3.21 to 3.22 in the same release with this change (this may impact our use of the libprotobuf apt packages from ubuntu in ci). 

If we really must continue to support protobuf 3.21, is it reasonable to not support utf8 validation in that case, document it, and provide a build/cmake warning?

exporters/otlp/CMakeLists.txt
```diff
@@ -9,6 +9,84 @@ else()
   set(OPENTELEMETRY_OTLP_TARGETS_LIB_TYPE)
 endif()
 
+if(TARGET utf8_range::utf8_validity)
+  set(OPENTELEMETRY_OTLP_DEP_UTF8_VALIDITY utf8_range::utf8_validity)
+else()
+  # Using FetchContent to fetch utf8_range from the protobuf's sources.
+  if(NOT Protobuf_PROVIDER STREQUAL "fetch_repository")
```
dbarker, line 16: We should be able to move checking/creation of the utf8_range target and compile definition to the `cmake/protobuf.cmake` file. 

Looking at how `utf8_range` is manged in Protobuf (from 3.22 on), if Protobuf was built with FetchContent then the `utf8_validity` target will exist but needs the `utf8_range::utf8_validity` ALIAS target created.  

If an installed Protobuf package was found then we should also expect that utf8_range was installed (See [protobuf/cmake/utf8_range.cmake#L10](https://github.com/protocolbuffers/protobuf/blob/514aceb974fbd55031169b79d2bd9f7646157787/cmake/utf8_range.cmake#L10) ). It can be found with `find_package(utf8_range)` to import the target. 

Could making the `utf8_range::utf8_validity` target available be simple like putting this at the end of the `cmake/protobuf.cmake`?  

```cmake
# When protobuf is fetched or built as a gRPC submodule 
# utf8_range is compiled but may need the alias target created
if(TARGET utf8_validity AND NOT TARGET utf8_range::utf8_validity)
  add_library(utf8_range::utf8_validity ALIAS utf8_validity)
endif()

# For the find_package(Protobuf) path, utf8_range can also be found
if(NOT TARGET utf8_range::utf8_validity)
  find_package(utf8_range CONFIG QUIET)
endif()

# The utf8_range target should now be available for use in otel-cpp
if(NOT TARGET utf8_range::utf8_validity)
  message(FATAL_ERROR "A required target (utf8_range::utf8_validity) was not found. "
    "Install the utf8_range package with Protobuf")
endif()
```

dbarker (commented):
Thanks for the feature @owent! Please see my comments below.

exporters/otlp/src/otlp_populate_attribute_utils.cc
lalitb, line 134: If I am reading the specs correctly, `nostd::span<const uint8_t>` represents raw bytes here, so this should still be encoded as `bytes_value`, not as an array of integers.

exporters/otlp/src/otlp_populate_attribute_utils.cc
lalitb, line 248: Same issue here for `std::vector<uint8_t>`: this is a byte buffer, so it should stay as `bytes_value` instead of being converted to an int array.

exporters/otlp/CMakeLists.txt
lalitb: Yes, moving this into `cmake/protobuf.cmake` looks cleaner and the alias + `find_package(utf8_range CONFIG QUIET)` approach makes sense for modern protobuf setups. The only caveat is version support: as written it assumes a protobuf layout that already provides utf8_validity, so it would need either an older-version fallback or an explicit decision to require newer protobuf versions.

exporters/otlp/src/otlp_populate_attribute_utils.cc
lalitb: That makes sense to me. Most of the complexity here is coming from trying to support protobuf layouts before `utf8_validity` was available. If we agree to  raise the minimum supported Protobuf version to 3.22, that would simplify this a lot. Otherwise, I think a documented “no UTF-8 validation on older protobuf” fallback with a build warning would be preferable to carrying this much custom compatibility logic in OTLP CMake.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -81,28 +104,34 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
...
   {
-    proto_value->set_string_value(nostd::get<const char *>(value));
+    const char *str_value = nostd::get<const char *>(value);
+    if (utf8_range::IsStructurallyValid(str_value))
+    {
+      proto_value->set_string_value(str_value);
+    }
+    else
+    {
+      proto_value->set_bytes_value(str_value);
```
ThomsonTan, line 114: When `str_value` is `const char*`, `set_bytes_value(str_value)` constructs a `std::string` that stops at the first NUL byte, silently truncating data. The `nostd::string_view` branch correctly passes `(data, size)`. Consider using `set_bytes_value(str_value, strlen(str_value))` for consistency, or document the NUL-terminated assumption.

exporters/otlp/src/otlp_populate_attribute_utils.cc
ThomsonTan: Ubuntu 22.04 still ships protobuf 3.21.12 via libprotobuf-dev, but I agree that bumping the minimum to 3.22 is the cleaner path. The compat code for 3.21 adds significant complexity for a single point release, so it would be great to avoid that unless there is a strong use case to keep supporting it.

exporters/otlp/CMakeLists.txt
dbarker: Good point. Falling back to not supporting utf8 validation and old otlp mapping behavior for strings with Protobuf <3.22 seems reasonable. In that case the CMake message could be a warning and not stop configuration of the project.

exporters/otlp/src/otlp_populate_attribute_utils.cc
dbarker: It looks like Ubuntu 26.04 will also ship with 3.21 :) 

It may reasonable to still support building/testing against those packages but not support utf8 validation on old protobuf. If there is a compile definition set then the otlp utils and tests could handle it.

exporters/otlp/src/otlp_populate_attribute_utils.cc
owent: Sorry, I didn’t see the spec about the conversation and assumed it was only for exporting byte types. The conversation has now been restored.

exporters/otlp/src/otlp_populate_attribute_utils.cc
owent: Thanks. It will issue a warning and skip UTF-8 validation when using protobuf 3.21 and  utf8_range can not be found.

cmake/protobuf.cmake
```diff
@@ -1,29 +1,37 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
-
-# Import Protobuf targets (protobuf::libprotobuf and protobuf::protoc) and set PROTOBUF_PROTOC_EXECUTABLE.
-# 1. If gRPC was fetched from github then use the Protobuf submodule built with gRPC
-# 2. Find an installed Protobuf package
-# 3. Use FetchContent to fetch and build Protobuf from GitHub
-
-if(DEFINED gRPC_PROVIDER AND NOT gRPC_PROVIDER STREQUAL "find_package" AND TARGET libprotobuf)
```
dbarker, line 10: Most of the changes appear to be new formatting only. Can these format only changes be easily reverted?

exporters/otlp/CMakeLists.txt
```diff
@@ -57,6 +64,12 @@ target_link_libraries(opentelemetry_exporter_otlp_builder_utils
                       PUBLIC opentelemetry_otlp_recordable)
 
 if(OPENTELEMETRY_INSTALL)
+  if(TARGET opentelemetry_otlp_utf8_validity)
```
dbarker, line 67: Is this separate target and otlp_utf8_validity package still needed?

dbarker (approved):
Looks good. Thanks for the feature. Please see a few minor cleanup questions below.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -81,12 +85,35 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
   }
   else if (nostd::holds_alternative<const char *>(value))
   {
-    proto_value->set_string_value(nostd::get<const char *>(value));
+    const char *str_value = nostd::get<const char *>(value);
+#if defined(ENABLE_OTLP_UTF8_VALIDITY)
+    if (utf8_range::IsStructurallyValid(str_value))
```
marcalff, line 90: Is there any way to opt-out and NOT perform UTF-8 validation on every string ?

My concern is with performance overhead.

Symbol `ENABLE_OTLP_UTF8_VALIDITY` is defined if the proper protobuf library is found, but this is not something a end user can control with a CMake option like `WITH_OTLP_UTF8_VALIDITY`.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -81,12 +85,35 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
...
+    {
+      proto_value->set_string_value(str_value);
+    }
+    else
+    {
+      proto_value->set_bytes_value(str_value, strlen(str_value));
```
marcalff, line 96: Here, `strlen` means the instrumented application can send binary in a string, as long as the binary does not contains a nul character ...

A properly instrumented application will need to provide a sequence of bytes instead of a string anyway, to work properly.

But this change (which faithfully implement the specs) makes every application slower, due to the added UTF-8 validation, to try to accommodate poorly instrumented applications using the wrong types.

marcalff (commented):
Thanks for the PR.

This PR actually implements the specification, which should be a good thing in general.

In this particular case, and this is an opinion, the spec is a bad idea.

Please consider to add an option to opt-out and not add this extra UTF-8 validation, for applications that are instrumented correctly (strings are string, bytes sequences are binary), which do not need the overhead.

exporters/otlp/CMakeLists.txt
```diff
@@ -33,9 +33,16 @@ target_include_directories(
 
 set(OPENTELEMETRY_OTLP_TARGETS opentelemetry_otlp_recordable)
 
-target_link_libraries(opentelemetry_otlp_recordable PUBLIC opentelemetry_logs)
-target_link_libraries(opentelemetry_otlp_recordable
-                      PUBLIC opentelemetry_metrics)
+if(TARGET utf8_range::utf8_validity)
+  target_link_libraries(
+    opentelemetry_otlp_recordable
+    PUBLIC opentelemetry_logs opentelemetry_metrics utf8_range::utf8_validity)
```
ThomsonTan, line 39: Link to `utf8_range::utf8_validity` with `PRIVATE` should be sufficient?

ThomsonTan (commented):
Consider add a CHANGELOG entry for this feature work.

cmake/protobuf.cmake
owent: Thanks, the formatting changes are revert.

exporters/otlp/CMakeLists.txt
owent: Yes, it's removed

exporters/otlp/src/otlp_populate_attribute_utils.cc
owent: `WITH_OTLP_UTF8_VALIDITY` was added to control this and is enabled by default to ensure correct exporting.

exporters/otlp/CMakeLists.txt
owent: Thanks, it's private linked now.

exporters/otlp/src/otlp_populate_attribute_utils.cc
owent: We receive a `const char*` that we assume is null-terminated, but it may contain invalid UTF-8.

ThomsonTan: Should include `<cstring>` for strlen then?

dbarker:
Thanks for the updates @owent. 


@lalitb @ThomsonTan @marcalff can you take another review pass?

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -81,12 +85,35 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
...
-                                  nostd::get<nostd::string_view>(value).size());
+    nostd::string_view str_value = nostd::get<nostd::string_view>(value);
+#if defined(ENABLE_OTLP_UTF8_VALIDITY)
+    if (utf8_range::IsStructurallyValid({str_value.data(), str_value.size()}))
+    {
+      proto_value->set_string_value(str_value.data(), str_value.size());
+    }
+    else
+    {
+      proto_value->set_bytes_value(str_value.data(), str_value.size());
```
lalitb, line 112: Should this respect `allow_bytes`? For explicit byte arrays, this helper only emits `bytes_value` when `allow_bytes` is true, otherwise it falls back to an int array. Invalid strings now always become `bytes_value`, even for span/resource/scope attribute paths that pass `allow_bytes=false`. If that is intentional for spec compliance, we should make that policy explicit; otherwise this should probably follow the same `allow_bytes` behavior as the byte array branches.

cmake/protobuf.cmake
```diff
@@ -78,3 +78,32 @@ endif()
...
+  endif()
+
+  # The utf8_range target should now be available for use in otel-cpp
+  if(NOT TARGET utf8_range::utf8_validity)
+    message(
+      WARNING
+        "Target (utf8_range::utf8_validity) was not found. "
+        "We will not validate UTF-8 strings in the OTLP exporter."
+        "Which may lead to record drops when sending non UTF-8 data as string.")
+  endif()
```
marcalff, line 108: else ... set the compile flag `ENABLE_OTLP_UTF8_VALIDITY` ?

marcalff (commented):
Thanks for the `WITH_OTLP_UTF8_VALIDITY` CMake flag.

What seem to be missing is setting `ENABLE_OTLP_UTF8_VALIDITY` when `WITH_OTLP_UTF8_VALIDITY` is ON and the proto lib is available.

exporters/otlp/src/otlp_populate_attribute_utils.cc
owent: I think allow_bytes is only used to convert `span<const uint8_t>` to bytes in some legacy cases. For strings, we should follow the spec. What do you think?

cmake/protobuf.cmake
owent: `ENABLE_OTLP_UTF8_VALIDITY` is set as a private definition in `exporters/otlp/CMakeLists.txt`. It is used only by `opentelemetry_otlp_recordable` and does not need to be exported.

exporters/otlp/src/otlp_populate_attribute_utils.cc
marcalff: > I think allow_bytes is only used to convert `span<const uint8_t>` to bytes in some legacy cases. For strings, we should follow the spec. What do you think?

Looks reasonable to me, waiting on @lalitb to confirm.

exporters/otlp/src/otlp_populate_attribute_utils.cc
marcalff: Adding the include will complicate things in the IWYU build, which does not set `WITH_OTLP_UTF8_VALIDITY` independently of `WITH_OTLP_HTTP`, etc.

Possible fix:

```
#include <cstring> // IWYU pragma: keep
```

or

```
#ifdef ENABLE_OTLP_UTF8_VALIDITY
#include <cstring>
#endif
```

which should pass CI (untested).

Ok with the code otherwise.

marcalff (approved):
LGTM, thanks for the feature and the CMake option.

See minor comment about include `<cstring>`, with a proposed (optional) fix.

marcalff:
@owent 

Thanks for the fix, now merging.

Issue #3999, opened 2026-04-14
[CI] Cleanup CI scripts and documentation


dbarker:
The ci scripts are a bit out of date and no longer match the ./ci/README.md. To help new contributors build and test PRs locally, the ci scripts, development container, and documentation need to be updated. 

- ./ci/do_ci.sh
  - Remove legacy c++11 and targets no longer used by the CI workflows
  - Set the maintainer mode cmake option where missing 
  - Use the modern `cmake --build` and `ctest` commands to allow building with Ninja
- Update devcontainer docker file to install bazel with the same script as the CI workflows
- Update ci/README

PR #4000, opened 2026-04-14
[CI] Update ci scripts and documentation

dbarker:
Contributes to #3999

## Changes

- Clean up do_ci.sh 
  - remove legacy c++11 targets
  - remove unused targets (thirdparty tag update, ..)
  - use modern cmake build and ctest commands
  - use ninja if available
  - add missing c++14 and c++17 cmake tests and set all to use maintainer mode
- Update devcontainer to install bazel the same way as CI
- Update ci script README

### DIFF
--- a/.devcontainer/Dockerfile.dev
+++ b/.devcontainer/Dockerfile.dev
@@ -29,11 +29,8 @@ RUN update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-
 
 RUN cd /opt/ci && bash setup_ci_environment.sh
 RUN cd /opt/ci && bash install_iwyu.sh
-
-ADD https://github.com/bazelbuild/bazelisk/releases/download/v1.22.1/bazelisk-linux-amd64 /usr/local/bin
-	
-RUN git config --global core.autocrlf input \
-	&& chmod +x /usr/local/bin/bazelisk-linux-amd64
+RUN cd /opt/ci && bash install_bazelisk.sh
+RUN git config --global core.autocrlf input
 
 ENV INSTALL_PACKAGES=${INSTALL_PACKAGES}
 ENV USER_NAME=devuser
--- a/ci/do_ci.sh
+++ b/ci/do_ci.sh
@@ -87,7 +87,7 @@ fi
 CMAKE_OPTIONS+=("-DCMAKE_CXX_STANDARD_REQUIRED=ON")
 CMAKE_OPTIONS+=("-DCMAKE_CXX_EXTENSIONS=OFF")
 
-CMAKE_BUILD_ARGS=(--parallel "$(nproc)")
+CMAKE_BUILD_ARGS=(--parallel)
 
 if [[ "${OTELCPP_CMAKE_VERBOSE_BUILD:-OFF}" =~ ^(1|ON|on|TRUE|true|YES|yes)$ ]]; then
   CMAKE_BUILD_ARGS+=(--verbose)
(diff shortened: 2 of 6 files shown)


.devcontainer/Dockerfile.dev
```diff
@@ -29,11 +29,8 @@ RUN update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-
 
 RUN cd /opt/ci && bash setup_ci_environment.sh
 RUN cd /opt/ci && bash install_iwyu.sh
-
-ADD https://github.com/bazelbuild/bazelisk/releases/download/v1.22.1/bazelisk-linux-amd64 /usr/local/bin
-	
-RUN git config --global core.autocrlf input \
-	&& chmod +x /usr/local/bin/bazelisk-linux-amd64
+RUN cd /opt/ci && bash install_bazelisk.sh
```
lalitb, line 34: nit: since `install_bazelisk.sh` accepts a version argument, should we pass `v1.22.1` here to preserve the previous devcontainer Bazelisk version? Otherwise this looks like a downgrade.

lalitb (approved):
Thanks. LGTM, with nit comment/

.devcontainer/Dockerfile.dev
dbarker: Good question. My intent is to try to bring the dev environment closer to the ci environment with this change. All the ci workflows currently use the default version of the script. How about a separate PR for the upgrade?

marcalff: Script ./ci/install_bazelisk.sh does not take a version as argument, it hard code version v1.16.0.

But I can not find where install_bazelisk.sh is used from today.

ci/do_ci.sh
```diff
@@ -397,11 +377,8 @@ elif [[ "$1" == "cmake.exporter.otprotocol.test" ]]; then
         -DWITH_OTLP_GRPC_CREDENTIAL_PREVIEW=ON \
         -DWITH_OTLP_RETRY_PREVIEW=ON \
         "${SRC_DIR}"
-  grpc_cpp_plugin=`which grpc_cpp_plugin`
-  proto_make_file="CMakeFiles/opentelemetry_proto.dir/build.make"
-  sed -i "s~gRPC_CPP_PLUGIN_EXECUTABLE-NOTFOUND~$grpc_cpp_plugin~" ${proto_make_file} #fixme
-  make -j $(nproc)
-  cd exporters/otlp && make test
+  cmake --build . "${CMAKE_BUILD_ARGS[@]}"
```
ThomsonTan, line 380: This runs under the boot root now which was `exportes/otlp`, does it increase the test code to be out of `cmake.exporter.otprococol.*`?

.devcontainer/Dockerfile.dev
dbarker: it's currently only used by the bazel jobs in the ci and benchmark workflows.

ci/do_ci.sh
dbarker: With this change it does run all the tests built. Before all the tests were built but only the otlp tests were run.

ThomsonTan (approved):

.devcontainer/Dockerfile.dev
marcalff: > How about a separate PR for the upgrade?

Sounds good to me

marcalff (approved):
LGTM, thanks for the cleanup.

PR #4008, opened 2026-04-16
[SDK] Add destructor to PeriodicExportingMetricReader to fix shutdown race

### LINKED ISSUE #4007: PeriodicExportingMetricReader: missing destructor causes use-after-destroy race (ThreadSanitizer)
**Describe your environment**

- OpenTelemetry C++ SDK: current `main` (also affects all released versions — the destructor has never been present)
- Compiler: GCC 12 (gcc-toolset-12)
- OS: RHEL 9 (x86_64)
- Sanitizer: ThreadSanitizer (`-fsanitize=thread`)

**Steps to reproduce**

1. Create a `PeriodicExportingMetricReader` with any exporter
2. Call `SetMetricProducer()` to start the background thread (via `OnInitialized`)
3. Destroy the reader **without** calling `Shutdown()` first
4. Build and run with `-fsanitize=thread`

Minimal reproducer (as a gtest):

```cpp
TEST(PeriodicExportingMetricReader, DestroyWithoutShutdown)
{
  std::unique_ptr<PushMetricExporter> exporter(
      new MockPushMetricExporter(std::chrono::milliseconds{0}));
  PeriodicExportingMetricReaderOptions options;
  options.export_timeout_millis  = std::chrono::milliseconds(200);
  options.export_interval_millis = std::chrono::milliseconds(500);
  MockMetricProducer producer;
  {
    auto reader =
        std::make_shared<PeriodicExportingMetricReader>(std::move(exporter), options);
    reader->SetMetricProducer(&producer);
    std::this_thread::sleep_for(std::chrono::milliseconds(50));
    // reader goes out of scope here — no Shutdown() call.
  }
}
```

Destroying a `PeriodicExportingMetricReader` without an explicit `Shutdown()` call should be safe — the destructor should stop and join the background worker thread before member destruction, just like `BatchSpanProcessor` and `BatchLogRecordProcessor` do.

The implicit default destructor destroys `cv_m_`, `cv_`, and `worker_thread_` while the background thread is still inside `cv_.wait_for()`. ThreadSanitizer reports:

```
WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread)
    #0 pthread_cond_clockwait ...
    #1 std::__condvar::wait_until ...
```

Without TSan, `std::thread::~thread()` calls `std::terminate()` (abort) because the thread was never joined.

**Additional context**

The base class `MetricReader` has `virtual ~MetricReader() = default` and does **not** call `Shutdown()` from its destructor. The two other SDK classes with background worker threads both handle this correctly:

```cpp
// sdk/src/trace/batch_span_processor.cc:365
BatchSpanProcessor::~BatchSpanProcessor()
{
  if (synchronization_data_->is_shutdown.load() == false)
  {
    InternalShutdown();
  }
}

// sdk/src/logs/batch_log_record_processor.cc:375
BatchLogRecordProcessor::~BatchLogRecordProcessor()
{
  if (synchronization_data_->is_shutdown.load() == false)
  {
    InternalShutdown();
  }
}
```

`PeriodicExportingMetricReader` is the only one missing this pattern. The fix is straightforward — add a destructor that calls `Shutdown()` and joins the thread:

```cpp
PeriodicExportingMetricReader::~PeriodicExportingMetricReader()
{
  if (!IsShutdown())
  {
    Shutdown();
  }

  {
    cv_.notify_all();
    worker_thread_.join();
  }
}
```

I have a fix and test case ready and will submit a PR shortly.

jnillius:
Fixes #4007

## Changes

Destroying a PeriodicExportingMetricReader without calling Shutdown()
leaves the background worker thread running inside cv_.wait_for() while
the mutex and condition variable members are destroyed. This causes a
use-after-destroy race detected by ThreadSanitizer.

Add an explicit destructor that calls Shutdown() and joins the worker
thread before member destruction, matching the pattern already used by
BatchSpanProcessor and BatchLogRecordProcessor.

Add a DestroyWithoutShutdown test to verify safe destruction without
an explicit Shutdown() call.

* [x] Changes in public API reviewed — the only addition is the destructor declaration in the header; no existing API is changed

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
+++ b/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
@@ -36,6 +36,8 @@ public:
                                 const PeriodicExportingMetricReaderOptions &options,
                                 const PeriodicExportingMetricReaderRuntimeOptions &runtime_options);
 
+  ~PeriodicExportingMetricReader() override;
+
   AggregationTemporality GetAggregationTemporality(
       InstrumentType instrument_type) const noexcept override;
 
--- a/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
+++ b/sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
@@ -38,6 +38,11 @@ public:
 
   ~PeriodicExportingMetricReader() override;
 
+  PeriodicExportingMetricReader(const PeriodicExportingMetricReader &)            = delete;
+  PeriodicExportingMetricReader &operator=(const PeriodicExportingMetricReader &) = delete;
+  PeriodicExportingMetricReader(PeriodicExportingMetricReader &&)                 = delete;
+  PeriodicExportingMetricReader &operator=(PeriodicExportingMetricReader &&)      = delete;
+
   AggregationTemporality GetAggregationTemporality(
       InstrumentType instrument_type) const noexcept override;
 
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -80,12 +80,6 @@ PeriodicExportingMetricReader::~PeriodicExportingMetricReader()
   {
     Shutdown();
   }
-
-  if (worker_thread_.joinable())
-  {
-    cv_.notify_all();
-    worker_thread_.join();
-  }
 }
 
 AggregationTemporality PeriodicExportingMetricReader::GetAggregationTemporality(
--- a/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
+++ b/sdk/src/metrics/export/periodic_exporting_metric_reader.cc
@@ -74,6 +74,20 @@ PeriodicExportingMetricReader::PeriodicExportingMetricReader(
   }
 }
 
+PeriodicExportingMetricReader::~PeriodicExportingMetricReader()
+{
+  if (!IsShutdown())
+  {
+    Shutdown();
+  }
+
+  if (worker_thread_.joinable())
+  {
+    cv_.notify_all();
+    worker_thread_.join();
+  }
+}
+
 AggregationTemporality PeriodicExportingMetricReader::GetAggregationTemporality(
     InstrumentType instrument_type) const noexcept
 {
--- a/sdk/test/metrics/periodic_exporting_metric_reader_test.cc
+++ b/sdk/test/metrics/periodic_exporting_metric_reader_test.cc
@@ -124,8 +124,7 @@ TEST(PeriodicExportingMetricReader, DestroyWithoutShutdown)
   // background worker thread.  Before the destructor fix this test would fail
   // under ThreadSanitizer with:
   //   WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread)
-  std::unique_ptr<PushMetricExporter> exporter(
-      new MockPushMetricExporter(std::chrono::milliseconds{0}));
+  auto exporter = std::make_unique<MockPushMetricExporter>(std::chrono::milliseconds{0});
   PeriodicExportingMetricReaderOptions options;
   options.export_timeout_millis  = std::chrono::milliseconds(200);
   options.export_interval_millis = std::chrono::milliseconds(500);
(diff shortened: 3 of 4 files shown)


marcalff:
@jnillius 

Thanks for the bug report and fix.

Please sign the EasyCLA, and fix clang-format.

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
```diff
@@ -36,6 +36,8 @@ class PeriodicExportingMetricReader : public MetricReader
                                 const PeriodicExportingMetricReaderOptions &options,
                                 const PeriodicExportingMetricReaderRuntimeOptions &runtime_options);
 
+  ~PeriodicExportingMetricReader() override;
```
dbarker, line 39: Please fix the new clang-tidy warning

Line | Check | Message
-- | -- | --
28 | cppcoreguidelines-special-member-functions | class 'PeriodicExportingMetricReader' defines a destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator

sdk/test/metrics/periodic_exporting_metric_reader_test.cc
```diff
@@ -117,6 +117,31 @@ TEST(PeriodicExportingMetricReader, Timeout)
...
 }
 
+TEST(PeriodicExportingMetricReader, DestroyWithoutShutdown)
+{
+  // Verify that destroying a reader without calling Shutdown() does not cause
+  // use-after-destroy races on the condition variable / mutex used by the
+  // background worker thread.  Before the destructor fix this test would fail
+  // under ThreadSanitizer with:
+  //   WARNING: ThreadSanitizer: unlock of an unlocked mutex (or by a wrong thread)
+  std::unique_ptr<PushMetricExporter> exporter(
```
dbarker: nit - can use std::make_unique here

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -74,6 +74,20 @@ PeriodicExportingMetricReader::PeriodicExportingMetricReader(
...
 }
 
+PeriodicExportingMetricReader::~PeriodicExportingMetricReader()
+{
+  if (!IsShutdown())
+  {
+    Shutdown();
+  }
+
+  if (worker_thread_.joinable())
```
dbarker: This section to join the thread can be removed. Shutdown will call OnShutdown which joins the thread (and calls Shutdown on the exporter).

dbarker (commented):
Good catch! Thanks for the PR. Please see feedback below on some minor points.

ThomsonTan:
Add an entry to CHANGELOG.md?

sdk/include/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.h
jnillius: applied!

sdk/test/metrics/periodic_exporting_metric_reader_test.cc
jnillius: applied!

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
jnillius: applied!

marcalff (commented):
Please fix clang-format.

LGTM otherwise.

dbarker (approved):
Looks good. Nice work.

marcalff (approved):

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -74,6 +74,14 @@ PeriodicExportingMetricReader::PeriodicExportingMetricReader(
   }
 }
 
+PeriodicExportingMetricReader::~PeriodicExportingMetricReader()
+{
+  if (!IsShutdown())
```
ThomsonTan, line 79: Is this `IsShutdown()` check needed? `Shutdown()` is already idempotent on `MetricReader`. The only effect of this guard is to suppress the
"Cannot invoke shutdown twice!" warning. Worth a short
comment explaining that, or drop the guard to match the
`BatchSpanProcessor` dtor pattern.

dbarker: I think we should keep the check, and a comment in the `PeriodicExporingMetricReader` destructor makes sense.

Many users will be adding the `MetricReader` to the `Provider/MeterContext` which will call `Shutdown` explicitly when the provider goes out of scope. In that case Shutdown will be called twice (first by the destructor of the `MeterProvider` then by the destructor of `PeriodicExporingMetricReader`. A new warning being emitted each time the `PeriodicExporingMetricReader` is destroyed may feel like a regression.

ThomsonTan: Ok, that makes sense, let's keep the check.

ThomsonTan (approved):

Issue #1996, opened 2023-02-22
Tool to detect API/ABI non-compliant changes, and integrate with CI.


lalitb:
As discussed in [today's Feb 22, 2023](https://docs.google.com/document/d/1i1E4-_y4uJ083lCutKGDhkpi3n4_e774SBLi9hPLocw/edit#heading=h.daa6czvwmr28) community meeting:

- Need tooling to detect ABI changes in CI, instead of relying on code review only
- There was some research done earlier on one of the promising - [ABI Compliance Checker (lvc.github.io)](https://lvc.github.io/abi-compliance-checker/), but good to compare other alternatives before finalizes.

ThomsonTan:
Does this mean it is always recommended for the user to build with the same version of API and SDK?
> otel-cpp doesn't guarantee ABI compatibility for SDK code

lalitb:
> Does this mean it is always recommended for the user to build with the same version of API and SDK?

No. This actually means the SDK specific component ( e.g. Processor and Exporter Interfaces) can be modified in ABI breaking way. So people having external exporters or processors are expected to rebuild them to be compatible with new SDK version.

marcalff:
Tooling added in a dedicated repository:

* https://github.com/marcalff/opentelemetry-cpp-abi

owent:
Could we publish these reports to somewhere?

marcalff:
> Could we publish these reports to somewhere?

Not yet, but will look into this.

PR #4011, opened 2026-04-18
[API] Fix `Logger.Enabled()` 

### LINKED ISSUE #2667: [API] Support for Logger::Enabled() is incomplete
```
  nostd::shared_ptr<opentelemetry::logs::Logger> logger = ...;

  if (logger->Enabled(severity)) {
  }
```

Logger::Enabled() always return false.

This is due to:

```
  inline bool Enabled(Severity severity) const noexcept
  {
    return static_cast<uint8_t>(severity) >= OPENTELEMETRY_ATOMIC_READ_8(&minimum_severity_);
  }

  mutable uint8_t minimum_severity_{kMaxSeverity};
```

The severity is kMaxSeverity by default, meaning all events are disabled by default.

There is a method to change the logger severity:

```
  void SetMinimumSeverity(uint8_t severity_or_max) noexcept
  {
    OPENTELEMETRY_ATOMIC_WRITE_8(&minimum_severity_, severity_or_max);
  }
```

but is it protected, and never called in the code base.

Any user code that checks for Enabled() can not emit logs.

Note that EmitLogRecord() helpers never honor the Enabled() flag either.

proost:
Fixes #2667

## Changes

fix to `Logger.Enabled()` use severity. so it changes to default behavior that enabled all logs to emit.

I fix it as minimal what original issue mentioned. But the issue reported two years ago, spec is changed.

do I need to follow "DEVELOPMENT" status spec?(https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#enabled)

I think adding features to "LoggerConfig" is correct to me(default behavior can be annoyed because all logs enabled).  But this is my first time, so I don't know what is rule in this project.


marcalff:
> do I need to follow "DEVELOPMENT" status spec?(https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#enabled)

Yes.

The spec in "development" status means some changes are still possible before it reaches a stable state.

SDK like opentelemetry-cpp needs to implement these specs early (so, now), to discover possible issues with the spec, so it can be adjusted if needed **_before_** reaching the stable status.

Thanks for the contribution.

api/include/opentelemetry/logs/logger.h
```diff
@@ -482,6 +477,115 @@ class Logger
...
+  bool ShouldEmitLogRecord(std::true_type,
+                           std::false_type,
+                           const ArgumentType &...args) const noexcept
+  {
+    const Severity *severity = FindLogArgument<Severity>(args...);
+    return severity == nullptr ? true : Enabled(*severity);
+  }
+
+  template <class... ArgumentType>
+  bool ShouldEmitLogRecord(std::true_type,
```
lalitb: This routes event-id log emission through `Enabled(severity, event_id)`, which calls the `virtual EnabledImplementation(...)`. We currently avoid virtual calls in the logger enablement hot path, so this would add virtual dispatch to calls such as `Info(event_id, ...)`. Can we keep the pre-emit fast path severity-only here, or otherwise make event-id enablement opt-in without putting a virtual call on the common log emission path?

@ThomsonTan could you also take a look? You added the non virtual logger enablement check, so your input would be useful here.

proost:
@marcalff @lalitb 
I changed code to follow current api/sdk specs. If community is ok, can I divide fixing "EmitLogRecord" another PR? Because I think current change is meaningful and big enough.

sdk/include/opentelemetry/sdk/configuration/severity_number.h
```diff
@@ -17,6 +17,7 @@ namespace configuration
 // YAML-NODE: SeverityNumber
 enum class SeverityNumber : std::uint8_t
 {
+  unspecified = 0,
```
proost: current code default is trace. if this is intended, I will rollback to current code.

dbarker:
@proost Thanks for the PR! Both the LoggerConfig and YAML parsing/building features are very much needed :) 

The PR has grown a bit beyond the initial issue. To facilitate review please break out the yaml config parsing to a separate PR as follow up to this one.

lalitb:
@proost  - Thanks for fixing the earlier `ShouldEmitLogRecord(...)` case.

  I think the same concern still matters for `Enabled(severity)` itself. `Enabled()` is called from application / instrumentation code as the guard before doing expensive log work, so this is one of the most important hot paths in the logs API.

  For example:

```cpp
  if (logger->Enabled(Severity::kDebug)) {
    // allocate/build expensive debug log payload
  }
```

When debug logging is disabled, applications should pay as little as possible here. Previously this was just the cached minimum-severity check. With the current code, the disabled path can still pay for runtime-context lookup and virtual dispatch through `Logger::EnabledImplementation(...)` / `LogRecordProcessor::Enabled(...)`.

Can we preserve the severity-only fast path for `Enabled(severity)`, so disabled logging remains just the cheap cached check, and only use the context / processor filtering path when that extra filtering is explicitly configured?

@ThomsonTan  - Can you have a quick look on this PR too. Thanks.

proost:
@lalitb 
I'm sorry missing that point: [97ebc84](https://github.com/open-telemetry/opentelemetry-cpp/pull/4011/commits/97ebc8481eaf503721ed2511ec9de1b8e252508f)

api/include/opentelemetry/logs/logger.h
```diff
@@ -464,7 +491,19 @@ class Logger
   //
 
 protected:
-  // TODO: discuss with community about naming for internal methods.
+  virtual bool EnabledImplementation(const opentelemetry::context::Context & /*context*/,
+                                     Severity /*severity*/) const noexcept
```
marcalff, line 509: This is a new virtual method, so it breaks the existing ABI for the API.

To discuss, we need to figure out how to roll out this change.

marcalff (dismissed):

api/include/opentelemetry/logs/logger.h
proost: I found "OPENTELEMETRY_ABI_VERSION_NO" in the release process doc. Maybe using this flag v2?

lalitb: Thanks @marcalff for flagging this. This also reminds me of #1996 - having an ABI compatibility gate that would automatically catch this kind of change.

@proost yes, adding a new virtual method changes the vtable, so it breaks the existing API ABI. `OPENTELEMETRY_ABI_VERSION_NO` is the mechanism we use for ABI-versioned API changes, but we should not break ABI v1.

I think we should either:

- keep this new virtual API under ABI v2 only, or
- find an ABI-safe design for ABI v1, without adding a new virtual method.

For example, an ABI-safe design would need to keep the `Logger` vtable and object layout unchanged. But I think that should be discussed separately if we want ABI v1 support. As of now, I do not see a strong reason to add this to ABI v1.

proost: followsing whole spec without ABI breaking change is not easy in here. So i prefer gate using flag. But community decide to include this feature in the v1, I will consider more how to deal with.

api/include/opentelemetry/logs/logger.h
```diff
@@ -476,6 +504,21 @@ class Logger
     return false;
   }
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  virtual bool EnabledImplementation(const opentelemetry::context::Context & /*context*/,
+                                     Severity /*severity*/) const noexcept
+  {
+    return true;
```
marcalff:
```suggestion
    return false;
```

The default implementation in the API needs to be disabled, for the noop logger when no SDK is present.

Not to be confused with what the SDK does by default (when there is an sdk).

api/include/opentelemetry/logs/logger.h
```diff
@@ -476,6 +504,21 @@ class Logger
...
+  }
+
+  virtual bool EnabledImplementation(const opentelemetry::context::Context &context,
+                                     Severity severity,
+                                     const EventId & /*event_id*/) const noexcept
+  {
+    return EnabledImplementation(context, severity);
```
marcalff:
```suggestion
    return false;
```

sdk/include/opentelemetry/sdk/logs/logger_config.h
```diff
@@ -27,6 +28,31 @@ class OPENTELEMETRY_EXPORT LoggerConfig
    */
   bool IsEnabled() const noexcept;
 
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
```
marcalff: Here, we are under sdk/include, not api/include.

It is ok to break the SDK ABI (with new data member), only the API ABI must be preserved at all costs.

Please remove `OPENTELEMETRY_ABI_VERSION_NO >= 2`, and have the minimum severity and trace based flag always defined in the SDK.

proost: Oh, I got it. [5aaab3d](https://github.com/open-telemetry/opentelemetry-cpp/pull/4011/commits/5aaab3d725d2f4eefc595531915c215eb75585fd)

lalitb:
This looks good to me. Just to reiterate if my understanding is correct now:

- v1: The API is safe. No API ABI break. The SDK binary layout changes, so old SDK-built binaries may need a rebuild, but normal source code should not need changes.
 - v2: The API intentionally includes the new Enabled() methods. The SDK also changes binary layout, but normal source code should still just recompile without code changes.

lalitb (approved):

marcalff (approved):
LGTM, thanks for the fix.

PR #4021, opened 2026-04-21
[API] Deprecate opentelemetry::plugin

marcalff:
Fixes # (issue)

## Changes

* Deprecated files in API, in the `opentelemetry::plugin` namespace
  * Added warnings in the files affected, and an error in no-deprecated builds. 
* Updated the DEPRECATED.md documentation
* Commented out plugin tests and examples from CMake and Bazel
* Disabled the plugin specific CI

The code is commented out for now, for easier rollback if so required.
File removal, including tests, examples and CI, to be done in the final removal PR.


ThomsonTan:
@lalitb @esigo, could you please take a look? The plugin interface has been there from the very beginning.

esigo (approved):
LGTM
Thanks for the PR :)

DEPRECATED.md
```diff
@@ -222,6 +222,52 @@ term, after ABI version 2 is declared stable.
...
+(include-what-you-use, cpp-check, clang-tidy, CI),
+and also increases confusion.
+
+This unused code is to be removed.
+
+#### Scope (plugin)
+
+The following files are to be removed from the API:
+
+* `opentelemetry/plugin/*.h`
```
ThomsonTan, line 253: Add `OPENTELEMETRY_DEPRECATED` to the header files which is going to be removed, in case anyone includes the header files?

lalitb:
Thanks, will have a look now.

lalitb:
Good to deprecate, but as per the deprecation process, we should be adding this in each public header deprecated:

```cpp
#ifdef OPENTELEMETRY_NO_DEPRECATED_CODE
#  error "header <opentelemetry/foo.h> is deprecated."
#endif
```

DEPRECATED.md
marcalff: Fixed, added warnings to each deprecated headers.

marcalff:
> Good to deprecate, but as per the deprecation process, we should be adding this in each public header deprecated:
> 
> ```c++
> #ifdef OPENTELEMETRY_NO_DEPRECATED_CODE
> #  error "header <opentelemetry/foo.h> is deprecated."
> #endif
> ```

I don't think anyone will see this, it supposes building with OPENTELEMETRY_NO_DEPRECATED_CODE in the first place.

Added warnings in all cases for visibility, as well as the error in no-deprecated builds, for more safety.

Note that I commented out CMake and bazel then, so we don't see this ourselves.

ThomsonTan (approved):

lalitb:
> I don't think anyone will see this, it supposes building with OPENTELEMETRY_NO_DEPRECATED_CODE in the first place.


Oh yes,  My earlier comment was based on the current docs/deprecation-process.md guidance for deprecated public headers, which explicitly calls for the `OPENTELEMETRY_NO_DEPRECATED_CODE` error path. So that part was about following the documented process as written.

That said, I agree that the `OPENTELEMETRY_NO_DEPRECATED_CODE` check alone is only visible in verification builds, so adding warnings in normal builds as well is a better outcome. Keeping warnings for general visibility and the hard error for no-deprecated builds makes sense to me.

I think we should do the follow-up PR to update `docs/deprecation-process.md` so the documented header deprecation pattern matches what we now consider the preferred implementation.

lalitb (approved):

PR #4035, opened 2026-04-26
[CODE HEALTH] fix nondeterministic pointer iteration order warnings

dbarker:
Contributes to #2053 

## Changes

- clean up memory management of the otlp utils functions
    - `OtlpRecordableUtils::PopulateRequest` takes in recordables through a const ref `nostd::span`. Before the method took ownership of the recordables and destroyed them which is unexpected. This PR keeps ownership with the caller. 
    -  The OTLP proto messages owned by the recordables are not created in a protobuf Arena but the request message is. This results in a deep copy when trying to move the recordable's Span/LogRecord into the request. This PR changes to an explicit deep copy for clarity. 
    - Update the `PopulateRequest` methods to remove local temporary proto messages and only create/modify the request message Arena allocated sub messages directly.   
- add tests to cover corner cases
- fix the following warnings: 

---
File | Line | Message
-- | -- | --
opentelemetry-cpp/exporters/otlp/src/otlp_recordable_utils.cc | 86 | iteration of pointers is nondeterministic
opentelemetry-cpp/exporters/otlp/src/otlp_recordable_utils.cc | 151 | iteration of pointers is nondeterministic
---


lalitb (approved):
Thanks.

marcalff (approved):
LGTM, thanks for the fix.

PR #4039, opened 2026-04-30
fix: noexcept for Observer

### LINKED ISSUE #2053: [Code health] clang-tidy cleanup

proost:
Fixes #2053

## Changes

bad allocation, bad variant access can happen. First i try to logging when exception happens, But following the [comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/3964#issuecomment-4217428324), ignore it.

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/observer_result.h
+++ b/sdk/include/opentelemetry/sdk/metrics/observer_result.h
@@ -7,6 +7,7 @@
 
 #include "opentelemetry/common/attribute_value.h"
 #include "opentelemetry/common/key_value_iterable.h"
+#include "opentelemetry/common/macros.h"
 #include "opentelemetry/metrics/observer_result.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/metrics/state/attributes_hashmap.h"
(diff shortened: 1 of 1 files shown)


lalitb:
Non-blocking: this improves the map-insertion failure path, but I think attributed observations can still terminate before this catch runs. The `MetricAttributes` construction copies attributes through `noexcept` callbacks/helpers that can allocate while storing keys/values. If that allocation throws, terminate happens before the outer catch in `Observe()` can handle it. We should either remove/catch those lower-level noexcept paths here or track it as a follow-up.

lalitb (approved):
thanks.

dbarker (approved):
Thanks for the PR. 

Still an open question about how to provide user visibility of these errors during development. Okay to silently return for now.

malkia:
Wait - why are we doing POKEMON coding here? If there is an actual memory stomp, it'll obscure it even more. This is dangerous!

lalitb:
> Wait - why are we doing POKEMON coding here? If there is an actual memory stomp, it'll obscure it even more. This is dangerous!

Fair point. This was approved and merged because it avoids `std::terminate` from the noexcept metrics path, but I agree the current catch (...) + silent return is too broad. I think we should improve this with a narrower follow-up fix.

proost:
> Wait - why are we doing POKEMON coding here? If there is an actual memory stomp, it'll obscure it even more. This is dangerous!

Valid concern. As dbaker mentioned, I think the broader issue is how to make these failures observable without putting the application at risk. 

If narrowing the catch clauses is needed here, I can send a follow-up fix.

PR #4041, opened 2026-04-30
fix examples/configuration

rishitha957:
## Changes

cmake with WITH_CONFIGURATION flag would break build now, as target link library is not recognizing private keyword

To reproduce this issue, run `cmake -S . -B build -G Ninja -DWITH_CONFIGURATION=1 && ninja -C build`

### DIFF
--- a/examples/configuration/CMakeLists.txt
+++ b/examples/configuration/CMakeLists.txt
@@ -20,7 +20,7 @@ add_executable(
   custom_log_record_processor_builder.cc)
 
 target_link_libraries(
-  example_yaml 
+  example_yaml
   PRIVATE ${CMAKE_THREAD_LIBS_INIT}
           common_metrics_foo_library
           common_logs_foo_library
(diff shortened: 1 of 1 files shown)


examples/configuration/CMakeLists.txt
```diff
@@ -20,8 +20,9 @@ add_executable(
   custom_log_record_processor_builder.cc)
 
 target_link_libraries(
-  example_yaml ${CMAKE_THREAD_LIBS_INIT}
-  PRIVATE common_metrics_foo_library
+  example_yaml
```
ThomsonTan:
```suggestion
  example_yaml 
  example_yaml
```

lalitb: I believe there is trailing whitespace after `example_yaml`, which needs to be fixed for CI to pass.

lalitb (approved):
Thanks for the fix.

examples/configuration/CMakeLists.txt
rishitha957: oh fixed the format now, thanks! not sure why it says merge blocking still .. should I wait for the CI? (but that seems to be stuck on `action required: awaiting approval from maintainers` [actions](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/25227496129)

PR #3241, opened 2025-01-10
Add stress testing framework, with basic metrics example to demonstrate.

lalitb:
## Changes

This PR adds a basic stress testing framework to validate the scalability and reliability of the functionality under high-concurrency and long-running workloads. Unlike Google Benchmark, which focuses on micro-benchmarking and latency measurements for isolated operations, this framework tries to simulate sustained, multi-threaded workloads to test a given workload. The idea is to complement the existing benchmarks by adding stress-tests to addressing long-duration and high-concurrency use-cases. 

This is already implemented for [.Net ](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main-1.3.0/test/OpenTelemetry.Tests.Stress) and [Rust](https://github.com/open-telemetry/opentelemetry-rust/tree/main/stress), and most of the ideas are taken from there. I felt the need for this to test some optimizations I am doing for metrics, but feel to comment if this doesn't seem helpful.

Also added a basic stress-testing example for metrics to demonstrate.  Below are the results from the metrics stress test as an example:

```console
$ ./stress_metrics
Starting stress test with 16 threads...
Throughput: 5009490 it/s | Avg: 4885764 | Min: 4734280 | Max: 5132395
 
Test completed:
Total iterations: 203373637
Duration: 42 seconds
Average throughput: 4885764 iterations/sec
$
```

It’s still in the early stages and will need further enhancements but should be a good starting point. Future improvements could include adding memory and CPU usage information alongside the existing throughput, as well as refining the initial warm-up period to sustain consistent data collection.

## Implementation Details:

 **Worker Threads**:
    - The worker threads (default to number of cores) are spawned to execute the workload.
    - Each worker thread executes the workload function (func) in a loop until a global STOP flag is set. (ctrl-c)
    - Each thread maintains its own iteration count to minimize contention.
 
  **Throughput Monitoring**:
    - A separate controller thread monitors throughput by periodically summing up iteration counts across threads.
    - Throughput is calculated over a sliding window (SLIDING_WINDOW_SIZE) and displayed dynamically.

  **Final Summary**:
    - At the end of the test, the program calculates and prints the total iterations, duration, and average throughput.
   
For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4042, opened 2026-05-02
Add preview support for bound  instruments 

lalitb:
## Summary

Related specs PR - https://github.com/open-telemetry/opentelemetry-specification/pull/5050

  Adds experimental bound synchronous metric instruments for metrics ABI v2. Bound instruments let callers bind a fixed attribute set once, then record repeated measurements through a bound handle:

  ```cpp
  auto counter = meter->CreateDoubleCounter("http.server.requests");

  std::map<std::string, std::string> checkout_200_attrs = {
      {"route", "/checkout"},
      {"status", "200"},
  };

  std::map<std::string, std::string> checkout_500_attrs = {
      {"route", "/checkout"},
      {"status", "500"},
  };

  auto checkout_200 = counter->Bind(
      opentelemetry::common::KeyValueIterableView<std::map<std::string, std::string>>(
          checkout_200_attrs));

  auto checkout_500 = counter->Bind(
      opentelemetry::common::KeyValueIterableView<std::map<std::string, std::string>>(
          checkout_500_attrs));

  checkout_200->Add(1);
  checkout_500->Add(1);
```
 
Both bound handles record to the same metric instrument, `http.server.requests`, with different pre-bound attribute sets. This avoids repeated per-call attribute processing and hashmap lookup for hot paths that reuse the same attribute set.

  ## Scope

  This PR intentionally keeps the preview surface narrow:

  - Counter: BoundCounter<T> + Counter<T>::Bind(...)
  - Histogram: BoundHistogram<T> + Histogram<T>::Bind(...)
  - SDK support for bound sync metric storage
  - multi-storage fanout support
  - noop support
  - bound-vs-unbound benchmark coverage
  - Bazel and CMake feature-gate support

  Not included in this preview:

  - UpDownCounter bind support
  - Gauge bind support
  - Context-bearing bound recording overloads
  - exemplar parity for bound recordings

  Those are left as follow-ups while the spec is still being finalized.

  ## Feature Gate
  
The feature is disabled by default.

  CMake:

```
  -DWITH_ABI_VERSION_2=ON
  -DWITH_METRICS_BOUND_INSTRUMENTS_PREVIEW=ON
```
  Bazel:

```
  --//api:abi_version_no=2
  --//api:with_metrics_bound_instruments_preview=true
```
  CMake rejects enabling `WITH_METRICS_BOUND_INSTRUMENTS_PREVIEW` without ABI v2.

  Bound code is guarded by: `OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW`

  ## Implementation Notes

  - Bound handles use the same metric name, description, unit, views, and export pipeline as the original instrument.
  - Attributes are processed once at bind time, then reused for later recordings.
  - Bound and unbound recordings with the same attributes are merged into the same exported datapoint.
  - Bound and unbound recordings share the same cardinality limit and overflow behavior.
  - Delta and cumulative temporality are handled by the existing metrics collection path.
  - Bound handles are safe to call after SDK storage is gone, but those late recordings are not exported.

  ## Benchmark

  | Benchmark | Mean | Median | StdDev |
  |---|---:|---:|---:|
  | BM_UnboundFixedAttrsCounter | 262 ns | 262 ns | 1.19 ns |
  | BM_BoundFixedAttrsCounter | 14.3 ns | 14.9 ns | 0.74 ns |

 Bound recording is about `18.3x` faster by mean for repeated recordings with the same fixed attribute set.
Env: Ubuntu 24.04 · GCC 13.3 · AMD EPYC 9V45 (32C/64T) · 251 GiB RAM.

### CODE REVIEW

File: api/include/opentelemetry/version.h [hunk 1]
```diff
@@ -30,3 +30,12 @@
...
 
 // clang-format on
+
+// Experimental: bound synchronous metric instruments (Counter, Histogram).
+// Enabled when both ABI v2 and the preview flag are set. Guard all bound
+// instrument code with `#ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW`.
+// This macro affects SDK class layout/vtables and MUST match between the SDK
+// library build and consumer translation units.
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2 && defined(ENABLE_METRICS_BOUND_INSTRUMENTS_PREVIEW)
+#  define OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW 1
```
lalitb, line 38: I kept this derived preview macro in version.h so API and SDK headers use one consistent guard for the ABI-v2 + preview-flag combination. This is temporary while bound instruments are experimental; once the API is stabilized, we can remove the preview flag and keep the surface under the normal ABI rules.

lalitb:
this is ready for review now. This is preview-only and gated behind ABI v2 plus `ENABLE_METRICS_BOUND_INSTRUMENTS_PREVIEW`, so existing users should not see the new API unless they opt in. The scope is Counter/Histogram bound handles for now, with the preview flag left in place while the API/spec settles.

lalitb:
I also did a local throwaway stress test on a 64-core test VM to sanity-check the expected performance shape using the stress framework #3241 

The result matches the design expectation: bound instruments help most when the workload records repeatedly to stable attribute sets, because the hot path avoids repeated attribute processing and hashmap lookup. The benefit is small when all threads record to the same single series, since the same bound cell becomes the contention point. With many pre-bound series, the bound path scaled much better because contention was spread across many handles.

So I would not describe the largest speedup number as a general result, but the trend is encouraging and supports the usefulness of the preview API for high-concurrency, stable-attribute workloads.

For example, with 2 attributes per series and 64 threads, the local stress test showed roughly:
 - 1 series: bound was about 2x faster
 - 1,024 series: bound was about 70x faster
 - 16,384 series: bound was much faster still, but I would treat that as a stress upper bound rather than a general claim

CMakeLists.txt
```diff
@@ -487,6 +499,10 @@ message(STATUS "WITH_API_ONLY: ${WITH_API_ONLY}")
 message(STATUS "WITH_NO_DEPRECATED_CODE: ${WITH_NO_DEPRECATED_CODE}")
 message(STATUS "WITH_ABI_VERSION_1: ${WITH_ABI_VERSION_1}")
 message(STATUS "WITH_ABI_VERSION_2: ${WITH_ABI_VERSION_2}")
+message(
```
dbarker: this should move to the preview section below.

test_common/cmake/preview-options.cmake
```diff
@@ -17,3 +17,6 @@ set(WITH_RESOURCE_DETECTORS_PREVIEW ${ENABLE_PREVIEW} CACHE BOOL "" FORCE)
 set(WITH_OTLP_HTTP_COMPRESSION ${ENABLE_PREVIEW} CACHE BOOL "" FORCE)
 set(WITH_CURL_LOGGING ${ENABLE_PREVIEW} CACHE BOOL "" FORCE)
 set(WITH_CONFIGURATION ${ENABLE_PREVIEW} CACHE BOOL "" FORCE)
+if(WITH_ABI_VERSION_2)
+  set(WITH_METRICS_BOUND_INSTRUMENTS_PREVIEW ${ENABLE_PREVIEW} CACHE BOOL "" FORCE)
```
dbarker, line 21: make this a cmake dependent option?

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -174,7 +207,91 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
...
+    void RecordDouble(double value) noexcept override;
+
+  private:
+    friend class SyncMetricStorage;
+    InstrumentValueType value_type_;
+    MetricAttributes attributes_;
+    // Protected by lock_.
+    opentelemetry::common::SpinLockMutex lock_;
+    std::unique_ptr<Aggregation> current_;
+    bool dirty_;
```
dbarker: nit - can initialize the bool here.

sdk/src/metrics/state/sync_metric_storage.cc
```diff
@@ -33,16 +47,203 @@ bool SyncMetricStorage::Collect(CollectorHandle *collector,
...
+                                            create_default_aggregation_());
+  bound_entries_.emplace(std::move(key), entry);
+  return entry;
+}
+
+void SyncMetricStorage::BoundEntry::RecordLong(int64_t value) noexcept
+{
+  if (value_type_ != InstrumentValueType::kLong)
+  {
+    return;
```
dbarker, line 229: should this log an error?

sdk/src/metrics/state/sync_metric_storage.cc
```diff
@@ -33,16 +47,203 @@ bool SyncMetricStorage::Collect(CollectorHandle *collector,
...
+  std::lock_guard<opentelemetry::common::SpinLockMutex> guard(lock_);
+  current_->Aggregate(value);
+  dirty_ = true;
+}
+
+void SyncMetricStorage::BoundEntry::RecordDouble(double value) noexcept
+{
+  if (value_type_ != InstrumentValueType::kDouble)
+  {
+    return;
```
dbarker, line 243: log an error?

sdk/src/metrics/sync_instruments.cc
```diff
@@ -569,6 +570,145 @@ void DoubleHistogram::Record(double value) noexcept
...
+{
+public:
+  explicit BoundLongHistogramImpl(std::shared_ptr<BoundSyncWritableMetricStorage> storage) noexcept
+      : storage_(std::move(storage))
+  {}
+  void Record(uint64_t value) noexcept override
+  {
+    if (storage_)
+    {
+      storage_->RecordLong(static_cast<int64_t>(value));
```
dbarker: check for overflow? We use static cast in similar narrowing conversions but perhaps should check and log an error.

sdk/src/metrics/sync_instruments.cc
```diff
@@ -569,6 +570,145 @@ void DoubleHistogram::Record(double value) noexcept
...
+opentelemetry::nostd::unique_ptr<opentelemetry::metrics::BoundHistogram<uint64_t>>
+LongHistogram::Bind(const opentelemetry::common::KeyValueIterable &attributes) noexcept
+{
+  std::shared_ptr<BoundSyncWritableMetricStorage> bound;
+  if (storage_)
+  {
+    bound = storage_->Bind(attributes);
+  }
+  return opentelemetry::nostd::unique_ptr<opentelemetry::metrics::BoundHistogram<uint64_t>>{
+      new BoundLongHistogramImpl(std::move(bound))};
```
dbarker, line 756: if these are startup/initialization time allocations and we accept they may throw/terminate then consider std::make_unique.

sdk/test/metrics/CMakeLists.txt
```diff
@@ -57,6 +57,19 @@ if(OPENTELEMETRY_ABI_VERSION_NO GREATER_EQUAL 2)
     TARGET multi_observer_test
     TEST_PREFIX metrics.
     TEST_LIST multi_observer_test)
+
+  if(WITH_METRICS_BOUND_INSTRUMENTS_PREVIEW)
+    add_executable(bound_sync_instruments_test bound_sync_instruments_test.cc)
+    target_link_libraries(
+      bound_sync_instruments_test ${GTEST_BOTH_LIBRARIES} ${GMOCK_LIB}
```
dbarker: missing PRIVATE

sdk/test/metrics/measurements_benchmark.cc
```diff
@@ -212,5 +213,56 @@ void BM_MeasurementsPerThreadCounterTest(benchmark::State &state)
 }
 BENCHMARK(BM_MeasurementsPerThreadCounterTest);
 
+#ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
+// Fixed-attribute hot path: unbound vs bound. The intent is to isolate
+// per-call attribute processing and hashmap lookup overhead, so the same
+// fixed attribute set is used for every Add() call.
+namespace
+{
+std::map<std::string, std::string> MakeFixedAttributes()
```
dbarker, line 224: do bound instruments support no attributes?

dbarker (commented):
Thanks for the feature. I'm excited to try it. Please see some initial feedback/questions below.

sdk/src/metrics/sync_instruments.cc
lalitb: Good point. Though I prefer to keep the existing direct construction, as it matches the surrounding `nostd::unique_ptr` API style and there is no functional issue to fix.

sdk/test/metrics/measurements_benchmark.cc
lalitb: Yes, empty attributes are supported for `Bind()`. I kept this benchmark with a fixed non-empty attribute set because it is intended to measure the per-call attribute processing/hash lookup avoided by bound instruments.

lalitb:
@dbarker Thanks for the review. I’ve addressed the comments and pushed the updates. Please let me know if you spot anything else.

api/include/opentelemetry/metrics/sync_instruments.h
```diff
@@ -26,6 +30,62 @@ class SynchronousInstrument
...
+// Histogram only. UpDownCounter, Gauge, exemplar parity, and context-bearing
+// bound operations are follow-ups. This API is experimental and is gated
+// behind both ABI v2 and ENABLE_METRICS_BOUND_INSTRUMENTS_PREVIEW
+// (see OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW in version.h).
+/**
+ * @since ABI_VERSION 2
+ * A bound counter handle obtained via Counter<T>::Bind(...). The associated
+ * attribute set is captured at Bind time so the hot path avoids per-call
+ * attribute processing and hashmap lookup. The handle must not outlive the
+ * Counter instrument from which it was obtained.
```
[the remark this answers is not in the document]
lalitb: Fixed. The docs now match the implementation: late calls after the originating Counter/SDK storage is destroyed or shut down may be dropped and are not guaranteed to be exported.

api/include/opentelemetry/metrics/sync_instruments.h
```diff
@@ -140,6 +226,33 @@ class Histogram : public SynchronousInstrument
   }
 #endif
 
+#ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
+  /**
+   * @since ABI_VERSION 2
+   * Returns a bound histogram handle for the given attribute set. Repeated
+   * calls to BoundHistogram<T>::Record(value) avoid per-call attribute
+   * processing and hashmap lookup. The bound handle MUST NOT outlive this
+   * Histogram instrument.
```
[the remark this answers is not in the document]
lalitb: Fixed. Histogram bound handles now use the same lifetime wording as counters: late calls may be dropped and are not guaranteed to be exported.

sdk/src/metrics/sync_instruments.cc
```diff
@@ -568,6 +570,164 @@ void DoubleHistogram::Record(double value) noexcept
...
+
+bool ToInt64Value(uint64_t value, const char *operation, int64_t &converted) noexcept
+{
+  if (value > static_cast<uint64_t>(std::numeric_limits<int64_t>::max()))
+  {
+    OTEL_INTERNAL_LOG_WARN(operation << " Value not recorded - value exceeds int64_t max");
+    return false;
+  }
+  converted = static_cast<int64_t>(value);
+  return true;
```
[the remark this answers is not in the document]
lalitb: Fixed. The checked uint64_t-to-int64_t conversion is now shared by bound and unbound LongCounter/LongHistogram paths, with tests for both.

sdk/src/metrics/state/sync_metric_storage.cc
```diff
@@ -32,16 +48,209 @@ bool SyncMetricStorage::Collect(CollectorHandle *collector,
...
+      }
+      // User still holds the handle (or rebound during rotation): keep.
+      if (it->second.use_count() != 1)
+      {
+        continue;
+      }
+      // use_count == 1 means no other thread can be calling RecordLong/Double
+      // on this entry, so reading dirty_ is race-free here. Lock anyway to
+      // satisfy the documented invariant on dirty_.
+      bool entry_dirty;
```
dbarker: `entry_dirty` needs initialization (clang-tidy warning `cppcoreguidelines-init-variables`)

sdk/src/metrics/sync_instruments.cc
```diff
@@ -76,7 +106,11 @@ void LongCounter::Add(uint64_t value, const opentelemetry::context::Context &con
                            << instrument_descriptor_.name_);
     return;
   }
-  return storage_->RecordLong(static_cast<int64_t>(value), context);
+  int64_t converted;
```
dbarker: `converted` needs initialization (10 clang-tidy `cppcoreguidelines-init-variables` warnings).

sdk/test/metrics/bound_sync_instruments_test.cc
```diff
@@ -0,0 +1,1036 @@
...
+
+// 2) Bound histogram records and exports same datapoint as unbound.
+TEST(BoundSyncInstruments, BoundHistogramMatchesUnbound)
+{
+  StorageHolder holder(InstrumentType::kHistogram, InstrumentValueType::kLong);
+  M attrs = {{"k", "v"}};
+  KeyValueIterableView<M> kv(attrs);
+
+  holder->RecordLong(5, kv, opentelemetry::context::Context{});
+  auto bound = holder->Bind(kv);
```
dbarker, line 290: is the Bind method that takes an initializer list tested?

dbarker (commented):
Thanks @lalitb. Just a question on tests and notes on the clang-tidy warnings. Looks good to me once CI is passing.

sdk/test/metrics/bound_sync_instruments_test.cc
lalitb: Good point, added them now.

api/include/opentelemetry/version.h
dbarker: These preview flags definitions likely need the `OPENTELEMETRY_` prefix since they are public on the API target (could be a separate PR). See some discussion here  https://github.com/open-telemetry/opentelemetry-cpp/pull/4160#discussion_r3431285097

Should we consider a new header for feature preview flag definitions (like `detail/feature_flags.h`) that is included by version.h or standardize on including them in the version header?

lalitb: Good point, I'd prefer t have this as separate PR, preferably single migration for all such feature flag.

sdk/test/metrics/bound_sync_instruments_test.cc
```diff
@@ -0,0 +1,1104 @@
...
+}
+
+uint64_t ValueAboveInt64Max() noexcept
+{
+  return static_cast<uint64_t>((std::numeric_limits<int64_t>::max)()) + 1;
+}
+}  // namespace
+
+// 1) Bound counter records and exports same datapoint as unbound counter with same attributes.
+TEST(BoundSyncInstruments, BoundCounterMatchesUnbound)
```
dbarker, line 168: Consider adding one case for binding with empty attributes.

dbarker (approved):
Look good. Thanks! See one comment on a nice to have test case.

PR #4043, opened 2026-05-02
[SDK] Do not use MultiSpanProcessor for single processors

Hailios:
Fixes # (issue)

## Changes

do not wrap single processors in a multi span processor. that is, use it when needed, otherwise not. idea here is to save execution time by not needlessly going through the multi span processor.


sdk/src/trace/tracer_context.cc
```diff
@@ -59,9 +72,32 @@ opentelemetry::sdk::trace::IdGenerator &TracerContext::GetIdGenerator() const no
...
+  {
+    return;
+  }
+
+  if (num_processors_ == 0)
+  {
+    // this is the first processor to be added, maybe the MultiSpanProcessor is not needed
+    processor_ = std::move(processor);
+  }
+  else if (num_processors_ == 1)
```
dbarker: Please add test cases for these if/else branches

sdk/src/trace/tracer_context.cc
```diff
@@ -59,9 +72,32 @@ opentelemetry::sdk::trace::IdGenerator &TracerContext::GetIdGenerator() const no
...
+  {
+    // if there already is a processor, then make a new MultiSpanProcessor to handle both
+    std::unique_ptr<MultiSpanProcessor> multi_processor(new MultiSpanProcessor({}));
```
dbarker: can use std::make_unique

dbarker (commented):
Thanks for the PR! Supporting replacement of the MultiSpanProcessor with a user defined processor is a nice addition. Please see a few comments and request for test cases.

sdk/test/trace/tracer_provider_test.cc
```diff
@@ -318,6 +319,72 @@ TEST(TracerProvider, GetTracerAbiv2)
...
+
+  auto &span_processor = context1->GetProcessor();
+
+  // Should be the SimpleSpanProcessor processor that was created above.
+#ifdef OPENTELEMETRY_RTTI_ENABLED
+  auto processor_typeed = dynamic_cast<MultiSpanProcessor *>(&span_processor);
+#else
+  auto processor_typeed = static_cast<MultiSpanProcessor *>(&span_processor);
+#endif
+  ASSERT_NE(nullptr, processor_typeed);
```
owent, line 437: Shoule we check `processor_typeed` contains all of processor1,processor2 and processor3 here?

Hailios: yes, but how?

sdk/src/trace/tracer_context.cc
Hailios: it looks like the rest of the code using the pattern above with first calling new and putting it in a smart pointer. and the specific case of creating an empty MultiSpanProcessor is not as easy with make_unique, since one would need to type out all the letters to make an empty vector of unique pointers of processors, rather than "{}".

marcalff (commented):
Thanks for the contribution.

This is a good improvement, and will save overhead in the most frequent cases.

Suggestions for a different implementation:

Have the following attributes:

```
std::unique_ptr<SpanProcessor> processor_;
MultiSpanProcessor *multi_processor_; // yes, raw pointer.
```

`processor_` will point to a single or multi processor, we don't know looking at this member alone.

`multi_processor_` == nullptr means `processor_` is a single processor.
`multi_processor_` != nullptr means `processor_` is a multi processor, and the raw pointer can be used to add more processors to it directly.

This should avoid the need to count processors, and cast back and forth between abstract and more narrow types, which causes confusion.

Drop the processor count, not needed.

Given that this class holds a unique pointer in processors_, the raw multi_processor_ pointer is safe to use.

sdk/test/trace/tracer_provider_test.cc
owent: We can add friend class only for test, like OtlpHttpExporterTestPeer in exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_exporter.h and exporters/otlp/test/otlp_http_exporter_test.cc

Hailios:
> 
> This is a good improvement, and will save overhead in the most frequent cases.
> 
> Suggestions for a different implementation:
> 
> Have the following attributes:
> 
> ```
> std::unique_ptr<SpanProcessor> processor_;
> MultiSpanProcessor *multi_processor_; // yes, raw pointer.
> ```
> 
> `processor_` will point to a single or multi processor, we don't know looking at this member alone.
> 
> `multi_processor_` == nullptr means `processor_` is a single processor. `multi_processor_` != nullptr means `processor_` is a multi processor, and the raw pointer can be used to add more processors to it directly.
> 
> This should avoid the need to count processors, and cast back and forth between abstract and more narrow types, which causes confusion.
> 
> Drop the processor count, not needed.
> 
> Given that this class holds a unique pointer in processors_, the raw multi_processor_ pointer is safe to use.

I updated the implementation to follow this advice

sdk/src/trace/tracer_context.cc
```diff
@@ -59,9 +72,31 @@ opentelemetry::sdk::trace::IdGenerator &TracerContext::GetIdGenerator() const no
...
+  {
+    // a processor exists, but it's not a MultiSpanProcessor. make a new MultiSpanProcessor
+    multi_processor_ = new MultiSpanProcessor({});
+    std::unique_ptr<MultiSpanProcessor> multi_processor(multi_processor_);
+    multi_processor->AddProcessor(std::move(processor_));
+    multi_processor->AddProcessor(std::move(processor));
 
-  auto multi_processor = static_cast<MultiSpanProcessor *>(processor_.get());
-  multi_processor->AddProcessor(std::move(processor));
+    processor_ = std::move(multi_processor);
```
lalitb, line 94: I think  this break spans that were already started before `AddProcessor()` runs? Those spans keep the recordable created by the old single processor, but after this line `Span::End()` will call `MultiSpanProcessor::OnEnd()`, which expects a `MultiRecordable`. Maybe promotion needs to preserve the old processor path for existing spans.

lalitb (changes requested):
As mentioned [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4043#discussion_r3245065206). We need to preserver the behavior as documented here - 

https://github.com/open-telemetry/opentelemetry-cpp/blob/76f087271bd97138caa9058aed454f9580f32415/sdk/include/opentelemetry/sdk/trace/tracer_provider.h#L107-L110

sdk/src/trace/tracer_context.cc
Hailios: hmm. yes I agree that it breaks currently live spans in they way you describe. unfortunate. do you have a suggestion on how to get around this?

lalitb: Maybe the cleanest fix is for `Span` to remember the processor used when it was started.

That should not require shared ownership or another heap allocation per span. It could just store a raw `SpanProcessor*` captured at start, then use that same processor for `OnEnd()`. The span already keeps the tracer/context alive, so the processor lifetime should still be covered.

lalitb: @Hailios  - Just checking in on this. I think the PR still needs a code change here before it can be approved.

The main invariant is that a span should use the same `SpanProcessor` for `MakeRecordable()`, `OnStart()`, and `OnEnd()`. Otherwise spans that were already in flight before `AddProcessor()` can still end through the new `MultiSpanProcessor`.

Hailios: Hi, I'm in a busy period, I'm aware of the wish to have that work.

Hailios: @lalitb - could you have a look at the current patch set, I've added a pointer to the Span class to keep track of its SpanProcessor?

sdk/include/opentelemetry/sdk/trace/tracer_context.h
```diff
@@ -129,6 +132,8 @@ class TracerContext
   std::unique_ptr<IdGenerator> id_generator_;
   std::unique_ptr<SpanProcessor> processor_;
   std::unique_ptr<instrumentationscope::ScopeConfigurator<TracerConfig>> tracer_configurator_;
+  // shares the pointer with processor_ if it is a MultiSpanProcessor, null otherwise
+  MultiSpanProcessor *multi_processor_;
```
dbarker, line 136: just missing initialization here triggering cppcoreguidelines-use-default-member-init

sdk/test/trace/tracer_provider_test.cc
```diff
@@ -19,6 +19,7 @@
 #include "opentelemetry/sdk/resource/resource.h"
 #include "opentelemetry/sdk/trace/exporter.h"
 #include "opentelemetry/sdk/trace/id_generator.h"
+#include "opentelemetry/sdk/trace/multi_span_processor.h"
 #include "opentelemetry/sdk/trace/processor.h"
 #include "opentelemetry/sdk/trace/random_id_generator.h"
 #include "opentelemetry/sdk/trace/sampler.h"
```
dbarker, line 25:
```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/trace/tracer_provider_test.cc should add these lines:

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/sdk/test/trace/tracer_provider_test.cc should remove these lines:
- #include "opentelemetry/common/macros.h"  // lines 13-13
```

sdk/src/trace/tracer_context.cc
```diff
@@ -60,9 +73,31 @@ opentelemetry::sdk::trace::IdGenerator &TracerContext::GetIdGenerator() const no
...
+  {
+    // this is the first processor to be added
+    processor_ = std::move(processor);
+  }
+  else if (multi_processor_ == nullptr)
+  {
+    // a processor exists, but it's not a MultiSpanProcessor. make a new MultiSpanProcessor
+    multi_processor_ = new MultiSpanProcessor({});
+    std::unique_ptr<MultiSpanProcessor> multi_processor(multi_processor_);
+    multi_processor->AddProcessor(std::move(processor_));
```
dbarker, line 91: This may be a race condition with the GetProcessor method called by spans.

sdk/src/trace/tracer_context.cc
```diff
@@ -60,9 +73,31 @@ opentelemetry::sdk::trace::IdGenerator &TracerContext::GetIdGenerator() const no
...
+    multi_processor->AddProcessor(std::move(processor));
 
-  auto multi_processor = static_cast<MultiSpanProcessor *>(processor_.get());
-  multi_processor->AddProcessor(std::move(processor));
+    processor_ = std::move(multi_processor);
```
dbarker, line 94: possible race condition here.

sdk/src/trace/tracer_context.cc
```diff
@@ -33,9 +33,22 @@ TracerContext::TracerContext(std::vector<std::unique_ptr<SpanProcessor>> &&proce
...
       id_generator_(std::move(id_generator)),
-      processor_(std::unique_ptr<SpanProcessor>(new MultiSpanProcessor(std::move(processors)))),
-      tracer_configurator_(std::move(tracer_configurator))
-{}
+      tracer_configurator_(std::move(tracer_configurator)),
+      multi_processor_(nullptr)
+{
+  if (processors.empty())
+  {
+    processor_ = std::unique_ptr<SpanProcessor>(new MultiSpanProcessor(std::move(processors)));
```
dbarker, line 41: This creates an unexpected case where you create a tracer provider with no processors then you add one with AddProcessor and you get a multi-processor decorator.

sdk/src/trace/tracer_context.cc
dbarker: The use case for adding processors after initial configuration (and while spans are active) is optional in the spec. It doesn't require the SDK support configuration updates to add processors, but does require that if the capability is supported then it must apply to all existing and new tracers. This must be thread safe and ideally not interrupt span creation/processing. 

The spec supports creating a TracerProvider with no processors and just requires that its tracers' Enable method return false if there are no processors configured. 

An alternative approach to supporting processor updates is to limit adding processors to the configuration phase. That phase could be from TracerProvider construction until the first tracer is created with GetTracer. Once a tracer is created AddProcessor can log a warning and do nothing.  This covers the initial configuration use case well since the thread configuring the provider has control over when it becomes available for other threads/libs to get tracers from (either by sharing it explicitly or by setting it in the global singleton). The benefit of this approach is that we can also make the provider AddProcessor method thread safe with the existing mutex.   

What do you think @lalitb ?

dbarker (commented):
Thanks for the PR @Hailios. Supporting the single processor case is a nice performance improvement that we should take advantage of. 

There are some potential race conditions with this current approach though. 

I think we should consider if supporting provider updates to add processors after the initial configuration phase is needed. More in comments below.

Issue #3998, opened 2026-04-14
[CI] Update ci and dev container to use llvm-22 for clang-tidy and iwyu


dbarker:
IWYU recently released [0.26](https://github.com/include-what-you-use/include-what-you-use/releases/tag/0.26) with support for llvm-22.

Currently CI runs iwyu with `llvm-18` and clang-tidy with `llvm-20`. Both can be upgraded to `llvm-22`. 

This upgrade should also be applied to the development container to support running the same CI iwyu and clang-tidy tests locally. 

See also: https://releases.llvm.org/22.1.0/tools/clang/tools/extra/docs/ReleaseNotes.html

PR #4051, opened 2026-05-04
[CI] Use fixed versions in iwyu CI & dev container

### LINKED ISSUE #4037: [CI] Create script to run IWYU locally in dev container
The IWYU Github action runs in a specific configuration of IWYU version, CMake configuration, and compiler flags. As a contributor it can be challenging to reproduce that exact environment to catch and fix IWYU warnings locally to ensure CI succeeds.  Building with different CXX standards, IWYU version, and opentelemetry-cpp CMake configuration can produce warnings that differ from CI. 

This issue is to create scripts that can be used to reproduce the CI IWYU warnings check locally

proost:
Fixes #4037 

## Changes

1. Use fixed versions in IWYU CI.
2. Unify versions with dev container for IWYU CI.

problem is that bump to 20 from 18, more warnings happen. so how about increase "warning_limit" temporarily and clean up after ?


ci/do_ci.sh
```diff
@@ -366,6 +366,35 @@ elif [[ "$1" == "cmake.clang_tidy.test" ]]; then
...
+    -DCMAKE_CXX_STANDARD=14 \
+    -DWITH_STL=CXX14 \
+    -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" \
+    -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE="include-what-you-use;-w;-Xiwyu;--mapping_file=${SRC_DIR}/.iwyu.imp;"
+  if command -v ninja >/dev/null 2>&1; then
+    IWYU_KEEP_GOING=(-k 0)
+  else
+    IWYU_KEEP_GOING=(-k)
+  fi
+  cmake --build "${BUILD_DIR}" --parallel -- "${IWYU_KEEP_GOING[@]}" 2>&1 | tee "$LOG_FILE"
```
ThomsonTan: Small consistency issue: this hard-codes `--parallel`, so `OTELCPP_CMAKE_BUILD_ARGS` no longer works for this target even though the CI README documents it as the build-args override. Can we route this through `${CMAKE_BUILD_ARGS[@]}` and keep the native `-k` args after `--`?

.devcontainer/Dockerfile.dev
```diff
@@ -19,13 +19,18 @@ RUN apt update && apt install -y wget \
     ninja-build \
     llvm-20-dev \
     libclang-20-dev \
+    clang-20 \
```
dbarker: given that we will have new warnings to fix let's jump all the way to llvm-22 for both iwyu and clang-tidy. Then update the warning limits in those github workflow jobs. Cleanup can follow in future PRs.  See
#3998

proost: [317baa3](https://github.com/open-telemetry/opentelemetry-cpp/pull/4051/commits/317baa3db4fab25635ee7bda6185f7005b2340da) 

OK, then bump clang-tidy also

ci/do_ci.sh
proost: [0db2cf1](https://github.com/open-telemetry/opentelemetry-cpp/pull/4051/commits/0db2cf17f8c3afce12d0af01d2e469a1774a2fda) 

Oh, thanks to point out.

.devcontainer/Dockerfile.dev
proost: Increase warning limit both clang tidy and IWYU: [281457f](https://github.com/open-telemetry/opentelemetry-cpp/pull/4051/commits/281457fa7187c976555d30fd76cafb68cd270187)

dbarker (approved):
Appreciate the CI updates. Thanks!

PR #4060, opened 2026-05-07
[SDK] Fix overflow attribute name to match the spec (otel.metric.overflow)

cijothomas:
## Changes

Fix the cardinality-limit overflow attribute key so it matches the
[Metrics SDK specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits).

The spec defines the synthetic overflow data point's attribute as
**`otel.metric.overflow`** (singular *metric*). The C++ SDK has been
exporting **`otel.metrics.overflow`** (plural *metrics*) since the
cardinality-limit feature was first added in #2255 (merged Nov 2023).

All other language SDKs that the spec compliance matrix lists as
implementing cardinality limits use the spec-correct singular name:

| SDK   | Attribute name           |
|-------|--------------------------|
| Go    | `otel.metric.overflow`   |
| Java  | `otel.metric.overflow`   |
| JS    | `otel.metric.overflow`   |
| .NET  | `otel.metric.overflow`   |
| C++   | `otel.metrics.overflow` (this PR fixes) |

## Why it matters

The whole point of the spec-defined overflow attribute is that backends,
dashboards, and alerting tools can detect cardinality overflow the same
way regardless of which language SDK produced the metric. Today a query
like `{otel.metric.overflow="true"}` silently misses overflow data
emitted by C++ applications.

## Compatibility note

The constant `kAttributesLimitOverflowKey` symbol is unchanged, so
in-process consumers using the symbol pick up the new value
automatically. Existing dashboards/alerts/queries that filtered on the
old (incorrect) string `otel.metrics.overflow` must be updated to
`otel.metric.overflow`. This is called out in the CHANGELOG under
*Important changes*.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] No new unit tests required — existing tests in
  `sum_aggregation_test.cc` and `cardinality_limit_test.cc` use the
  `kAttributesLimitOverflowKey` symbol, so they continue to pass with
  the corrected value.
- [x] Public API not changed (only the value of an internal const string).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,19 @@ Important changes:
   * namespace opentelemetry::plugin is deprecated
   * See file DEPRECATED.md for details.
 
+* [SDK] Fix cardinality-limit overflow attribute name to match the
+  specification
+  [#4060](https://github.com/open-telemetry/opentelemetry-cpp/pull/4060)
+
+  * The synthetic overflow data point attribute is now exported as
+    `otel.metric.overflow` (singular) per the
+    [Metrics SDK specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits).
+  * Previously the SDK exported `otel.metrics.overflow` (plural), which
+    diverged from the spec and from every other language SDK
+    (Go, Java, JS, .NET).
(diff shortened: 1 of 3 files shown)


marcalff (approved):
LGTM, thanks for the fix.

PR #4063, opened 2026-05-07
[TEST] Assert overflow attribute value is boolean true

cijothomas:
Follow-up to #4060.

The spec says the overflow data point has exactly one attribute:
`otel.metric.overflow` -> `bool true`. Existing tests only checked the
key, so a regression in the value or its type would not be caught.
This PR adds the missing `bool true` assertions and pins the
precomputed `kOverflowAttributes` set against the spec.

Tests-only change.


lalitb (approved):
Thanks @cijothomas - looking forward to more contributions from you ;)

ThomsonTan (approved):

PR #4065, opened 2026-05-07
[SDK] Support TracerConfigurator updates

dbarker:
Add support for updating the TracerConfigurator at runtime and provide an example.

The TracerProvider configuration includes the TracerConfigurator. SDKs may support updating the TracerConfigurator and resulting TracerConfig objects at runtime provided it is applied to previously created tracers. This feature allows dynamic enabling of tracers to turn tracing on or off without restarting the application.  

https://opentelemetry.io/docs/specs/otel/trace/sdk/#configuration

> The TracerProvider MAY provide methods to update the configuration. If configuration is updated (e.g., adding a SpanProcessor), the updated configuration MUST also apply to all already returned Tracers (i.e. it MUST NOT matter whether a Tracer was obtained from the TracerProvider before or after the configuration change). 

## Changes
- Add SDK `TracerProvider::UpdateTracerConfigurator` method and tests
- Add mutex for the SDK `Tracer::tracer_config_`  that will support dynamic updates including growth beyond the enabled bool. 
- Add example showing runtime changes to the tracer configurator to enable/disable tracers. 


sdk/src/trace/tracer_provider.cc
```diff
@@ -129,6 +129,26 @@ void TracerProvider::AddProcessor(std::unique_ptr<SpanProcessor> processor) noex
...
+        tracer_configurator) noexcept
+{
+  // The only way to set the TracerConfig of a tracer is on Tracer construction in
+  // TracerProvider::GetTracer or through Tracer::UpdateTracerConfig (which is private and only
+  // accessed by TracerProvider).
+  //
+  // Lock the provider mutex while updating the TracerConfiguartor in the context and setting the
+  // new TracerConfig of all existing tracers.
+  const std::lock_guard<std::mutex> guard(lock_);
+  context_->SetTracerConfigurator(std::move(tracer_configurator));
```
lalitb: One thing to guard here: if `tracer_configurator` is null, we store it and then immediately call `GetTracerConfigurator().ComputeConfig(...)` below, which will dereference a null pointer. Can we either reject/null-check it here or document that this input must be non-null?

sdk/src/trace/tracer.cc
```diff
@@ -58,7 +59,12 @@ nostd::shared_ptr<opentelemetry::trace::Span> Tracer::StartSpan(
     const opentelemetry::trace::SpanContextKeyValueIterable &links,
     const opentelemetry::trace::StartSpanOptions &options) noexcept
 {
-  if (!tracer_config_.IsEnabled())
+  // Check if the tracer is enabled using the API Tracer::Enabled() accessor if available.
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  if (!Enabled())
+#else
+  if (!tracer_config_.load()->IsEnabled())
```
lalitb: I think this does add some hot-path overhead. In ABI v1, `StartSpan` now does `tracer_config_.load() -> IsEnabled()` on every call instead of reading the inline config directly, and `Tracer` also moves from inline `TracerConfig` storage to a shared_ptr-backed config. Could we keep a cached atomic enabled flag for the StartSpan gate and leave the shared config for the update path only?

dbarker: Good idea. Updated with a local atomic bool in the sdk tracer for ABIv1

sdk/src/trace/tracer_provider.cc
dbarker: added a check and log message

sdk/src/trace/tracer_provider.cc
```diff
@@ -129,6 +129,26 @@ void TracerProvider::AddProcessor(std::unique_ptr<SpanProcessor> processor) noex
...
+  for (auto &tracer : tracers_)
+  {
+    auto new_config =
+        context_->GetTracerConfigurator().ComputeConfig(tracer->GetInstrumentationScope());
```
lalitb, line 156: Can we avoid calling `ComputeConfig()` while holding the provider lock? Since matchers can be user callbacks, this can deadlock if the callback calls back into the provider, e.g. `GetTracer()`. Maybe snapshot the tracers under the lock, then compute configs after releasing it.

sdk/src/trace/tracer_provider.cc
```diff
@@ -128,6 +128,30 @@ void TracerProvider::AddProcessor(std::unique_ptr<SpanProcessor> processor) noex
...
+    current_tracers = tracers_;
+  }
+
+  // The only way to set the TracerConfig of a tracer is on Tracer construction in
+  // TracerProvider::GetTracer or through Tracer::UpdateTracerConfig (which is private and only
+  // accessed by TracerProvider).
+  for (auto &tracer : current_tracers)
+  {
+    TracerConfig new_config =
+        context_->GetTracerConfigurator().ComputeConfig(tracer->GetInstrumentationScope());
```
lalitb, line 156: Thanks, the earlier provider-lock reentrancy issue is fixed. One remaining concurrency issue: the tracers are snapshotted, but the configurator itself is still read through `context_->GetTracerConfigurator()` after the lock is released. A concurrent `UpdateTracerConfigurator()` can replace that unique_ptr while this thread is computing configs. Can we also give  the configurator a stable lifetime for the out-of-lock compute step?

dbarker: good catch. updated to create a copy of the tracer configurator provided to the update method.

lalitb: Thanks, the copy fixes the configurator lifetime issue. I think concurrent updates can still race, though. One call can install config A, release the lock, then another call can install config B and update the tracers. If the first call resumes after that, it can update the existing tracers back to config A, while new tracers use config B from the context. Can we make the full update ordered, while still keeping `ComputeConfig()` outside the provider lock?

sdk/include/opentelemetry/sdk/trace/tracer.h
```diff
@@ -103,11 +105,26 @@ class Tracer final : public opentelemetry::trace::Tracer,
...
+
   // order of declaration is important here - instrumentation scope should destroy after
   // tracer-context.
   std::shared_ptr<InstrumentationScope> instrumentation_scope_;
   std::shared_ptr<TracerContext> context_;
+  mutable std::mutex tracer_config_mutex_;
   TracerConfig tracer_config_;
+#if OPENTELEMETRY_ABI_VERSION_NO < 2
+  std::atomic<bool> is_enabled_;
+#endif
```
marcalff, line 139: First, the check on ABI_VERSION can be removed, this is in the SDK.

Second, the initialization MUST use {value}, see PR #2244:

    [SDK] Valgrind errors on std::atomic variables (#2244)

dbarker: > First, the check on ABI_VERSION can be removed, this is in the SDK.

Thanks. This `sdk::trace::Tracer::is_enabled_` atomic was added for only ABIv1 since the API `Tracer::enabled_` exists only for ABIv2. I think we should keep it guarded here in the SDK Tracer since it is redundant with the API Tracer boolean in ABIv2. When the project moves to ABIv2 as the default and ABIv1 is deprecated this SDK `Tracer::is_enabled_` atomic should be removed. Does that sound reasonable? 

> Second, the initialization MUST use {value}

Good catch. I'll default initialize it in the header. CI is may not be warning on this since it is set in the constructor member initializer list.

sdk/src/trace/tracer_provider.cc
dbarker: @lalitb Thanks for the feedback. Given the added complexity of trying to support this use case (calling `TracerProvider::GetTracer` from the `ScopeConfigurator::ComputeConfig` method) I'm reconsidering if it is a valid case worth supporting.  Interested in your thoughts here. 

From the spec: 
> 
> A TracerConfigurator is a function which computes the [TracerConfig](https://opentelemetry.io/docs/specs/otel/trace/sdk/#tracerconfig) for a [Tracer](https://opentelemetry.io/docs/specs/otel/trace/sdk/#tracer).
> 
> The function MUST accept the following parameter:
> 
> tracer_scope: The [InstrumentationScope](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/) of the Tracer.
> The function MUST return the relevant TracerConfig, or some signal indicating that the [default TracerConfig](https://opentelemetry.io/docs/specs/otel/trace/sdk/#tracerconfig) should be used. This signal MAY be nil, null, empty, or an instance of the default TracerConfig depending on what is idiomatic in the language.
> 
> This function is called when a Tracer is first created, and for each outstanding Tracer when a TracerProvider’s TracerConfigurator is updated (if updating is supported). Therefore, it is important that it returns quickly.


Given that the `ScopeConfigurator::ComputeConfig` function is called when a tracer is created or when the `TracerConfigurator` is updated and must return quickly seems to imply that getting/creating a tracer in this function (and any side effects) would be unexpected. Doing so would call the `TracerConfigurator::ComputeConfig` function again if `GetTracer` creates a tracer.

I'd like to document this as an unsupported use case with comments to `UpdateTracerConfigurator`  (note that calling `GetTracer` from `ScopeConfigurator::ComputeConfig` is not supported and would result in deadlock), revert `UpdateTracerConfigurator` back to calling `ComputeConfig` under the `TracerProvider::lock_` mutex, and remove the current copy outside of lock workaround.  

If the use case is later needed we could add the synchronization complexity then. What do you think?

lalitb: Yes agree with this.

lalitb (approved):

sdk/src/trace/tracer.cc
```diff
@@ -197,6 +210,22 @@ void Tracer::CloseWithMicroseconds(uint64_t timeout) noexcept
...
+  {
+    std::lock_guard<std::mutex> lock(tracer_config_mutex_);
+    tracer_config_ = config;
+  }
+
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  UpdateEnabled(enabled);
+#else
+  is_enabled_.store(enabled, std::memory_order_relaxed);
+#endif
```
marcalff: Ok, so UpdateEnabled() is a method that exists only in the API, starting with ABI version 2.

How about, in ABI version 1, defining the very same method with the same implementation, only in the SDK, instead of implementing a different attribute name (is_enabled) and type (std::atomic) ?

This code then becomes:

```
UpdateEnabled(enabled);
```

in all cases.

Likewise, define method `Enabled()` and member `enabled_` in the SDK in ABI v1.

Basically, the pattern is the opposite of `ForceFlushWithMicroseconds()` which was demoted from the API to the SDK.

Here Enabled() can be in the SDK in ABI v1, promoted to the API in ABI v2.

marcalff (commented):
See suggestion for Enabled().

examples/tracer_configurator/main.cc
```diff
@@ -0,0 +1,149 @@
...
+// 1. "my_library" (example instrumented user code),
+// 2. "external_library_foo" (example instrumented third-party dependency).
+//
+// The example simulates a debugging workflow where only the tracer configuration is changed at
+// runtime through the provider:
+//
+//   Stage 1 – Start with all tracing disabled (low-overhead production default).
+//   Stage 2 – An issue is reported. Enable only the user library traces to get an initial signal.
+//   Stage 3 – The issue involves external libraries too. Enable all traces for full visibility.
+//   Stage 4 – Investigation complete. Disable all tracing again.
```
marcalff: Looks very nice.

Could you paste the test output in a comment, for reviewers ?

sdk/src/trace/tracer.cc
dbarker: Good idea. I'll try this way.

dbarker: updated to add Enabled and UpdateEnabled to the sdk Tracer for ABIv1.

dbarker: updated to add Enabled and UpdateEnabled to the sdk Tracer for ABIv1.

examples/tracer_configurator/main.cc
dbarker: Added a readme to the example and refactored the example to make the use case a bit more clear that existing tracers are updated.

sdk/src/trace/tracer_provider.cc
marcalff: Thanks for raising the question.

ComputeConfig() will invoke ScopeConfigurator, which are user provided callbacks.

However, these callbacks operate only on the configuration itself, they never see a tracer or tracer provider, so there is no risk to loop back into the provider and have a deadlock.

sdk/src/trace/tracer_provider.cc
marcalff: Regarding:

> Given that the ScopeConfigurator::ComputeConfig function is called when a tracer is created or when the TracerConfigurator is updated and must return quickly seems to imply that getting/creating a tracer in this function (and any side effects) would be unexpected. 

I don't see how the ScopeConfigurator callback can even have access to a tracer or tracer provider in the first place, these callbacks only operate on configuration, and are not aware of who/what own the configuration (the tracer and tracer provider).

ComputeConfig is given an InstrumentationScope, not a tracer, so there is no way to loop back to a tracer.

So, in my understanding, this is not an unsupported use case, it is an impossible use case instead.

Code in the patch looks ok, even locking the provider lock for the whole re configuration.

marcalff (approved):
Excellent work.

Thanks in particular for the examples, and thorough unit tests.

PR #4069, opened 2026-05-08
Fix: preserve delta start timestamp in fast path

### LINKED ISSUE #4038: Delta temporality fast path in TemporalMetricStorage::buildMetrics sets start_ts to SDK start time instead of previous collection time
When using delta aggregation temporality with a single metric reader, TemporalMetricStorage::buildMetrics takes a fast path that incorrectly sets start_ts to sdk_start_ts on every export. This means every exported delta data point has the same start_time_unix_nano (the SDK initialization time) rather than the end time of the previous collection interval.

**Location**:
temporal_metric_storage.cc, around line 55:

```
// Fast path for single collector with delta temporality
if (collectors.size() == 1 && aggregation_temporarily == AggregationTemporality::kDelta)
{
    ...
    metric_data.start_ts = sdk_start_ts;  // BUG: always SDK boot time
    metric_data.end_ts   = collection_ts;
    ...
}
```
**Expected behavior**:
Each delta export should have start_ts set to the end_ts of the previous collection (i.e., the previous collection_ts). The first export can use sdk_start_ts, but subsequent exports should advance start_ts.

**Actual behavior**:
Every delta export has start_ts = sdk_start_ts regardless of how many collections have occurred. Only end_ts advances.

**Impact**:
Backends that use start_time_unix_nano for data point placement (e.g., New Relic maps it to the timestamp field) will see all delta data points collapsed to the same timestamp, making timeseries queries unusable.

**The slow path handles this correctly**:
When collectors.size() > 1 or temporality is cumulative, the code falls through to the slow path which correctly tracks last_collection_ts per collector:

```
last_collection_ts = last_reported_metrics_[collector].collection_ts;
...
metric_data.start_ts = last_collection_ts;
```
**Suggested fix**:
Track the last collection timestamp in the fast path the same way the slow path does. For example, add a member variable (e.g., last_delta_collection_ts_) that is initialized to sdk_start_ts and updated to collection_ts after each fast-path export:

```
if (collectors.size() == 1 && aggregation_temporarily == AggregationTemporality::kDelta)
{
    if (delta_metrics->Size() == 0)
    {
        return true;
    }
    MetricData metric_data;
    metric_data.instrument_descriptor   = instrument_descriptor_;
    metric_data.aggregation_temporality = AggregationTemporality::kDelta;
    metric_data.start_ts                = last_delta_collection_ts_;  // FIX
    metric_data.end_ts                  = collection_ts;
    last_delta_collection_ts_           = collection_ts;              // FIX

    delta_metrics->GetAllEntries(...);
    return callback(metric_data);
}
```
**Workaround**:
Register a second (no-op) metric reader on the MeterProvider so that collectors.size() > 1, forcing the slow path which correctly tracks per-interval timestamps.

fmutlu68:
Fixes #4038

## Changes

- Track the last delta collection timestamp in the fast path
- Use the previous collection time as the start timestamp for delta metrics


sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -54,6 +54,11 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   // no other reader configured to collect those data.
   if (collectors.size() == 1 && aggregation_temporarily == AggregationTemporality::kDelta)
   {
+    if (!has_last_delta_collection_ts_)
+    {
+      last_delta_collection_ts_     = sdk_start_ts;
+      has_last_delta_collection_ts_ = true;
+    }
     // If no metrics, early return
     if (delta_metrics->Size() == 0)
```
ThomsonTan, line 63: On the empty-collection path, the function early-returns when `delta_metrics->Size() == 0` *before* updating `last_delta_collection_ts_`. If a cycle at `t2` is empty and the next cycle at `t3` has data, the emitted point will carry `start_ts = t1` (the last non-empty collection) rather than `t2`, even though the delta map only contains activity from `(t2, t3]`. Successive delta windows then overlap instead of abutting, which violates the OTel data model's requirement that `start_ts(n) == end_ts(n-1)` and causes rate computations (`value / (end - start)`) to under-report. The slow path doesn't have this issue because `last_reported_metrics_[collector].collection_ts` is updated on every call, including empty ones.

ThomsonTan:
Thanks @fmutlu68 for the fix. Please add a unit test that calls `buildMetrics` for a single delta collector across at least three collection cycles (with one empty cycle in the middle) and asserts that each `MetricData::start_ts` equals the previous `end_ts`. Without it, the fast path can silently regress again.

fmutlu68:
@ThomsonTan Thank you for your detailed feedback. I have created a unit test which contains 3 cycles which you mentioned. Could you check it ?

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -93,6 +100,12 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   if (present == unreported_metrics_.end())
   {
     // no unreported metrics for the collector, return.
+    auto reported = last_reported_metrics_.find(collector);
+    if (reported != last_reported_metrics_.end())
+    {
+      // To save the time for next cycle
+      reported->second.collection_ts = collection_ts;
```
lalitb: One thing I want to double check before approving: this new slow-path update may be unreachable. Once a collector has reported, we create an entry in `unreported_metrics_` for it, and later we only move `present->second`, we never erase the map key. So `present == end()` may not really happen together with `reported != end()`. Can you confirm whether this branch is actually expected to run?

fmutlu68:
Could anyone have a look at here ? @ThomsonTan @lalitb It passes all checks

lalitb (approved):
LGTM. Thanks.

Issue #3846, opened 2026-02-05
Issue: ETW Exporter Tracer Uses a Single Trace ID for All Spans


lukeina2z:
The ETW Exporter provides its own tracer implementation, which inherits from the OpenTelemetry Tracer interface. The relevant code is here:
https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h#L162C28-L163

### Expected Behavior (OpenTelemetry C++ SDK)
Using the standard OpenTelemetry SDK, two spans created with the default tracer correctly receive different Trace IDs. Here is the example code used to verify this behavior:

```C++
  void TestOTelTracer()
  {
    auto ostream_exporter = opentelemetry::exporter::trace::OStreamSpanExporterFactory::Create();
    auto ostream_processor = opentelemetry::sdk::trace::SimpleSpanProcessorFactory::Create(std::move(ostream_exporter));

    std::vector<std::unique_ptr<opentelemetry::sdk::trace::SpanProcessor>> processors;
    processors.push_back(std::move(ostream_processor));

    auto resource_attributes = opentelemetry::sdk::resource::ResourceAttributes{ {"service.name", "Test OTel"} };
    auto resource_ptr = opentelemetry::sdk::resource::Resource::Create(resource_attributes);

    std::unique_ptr<opentelemetry::sdk::trace::TracerContext> context =
      opentelemetry::sdk::trace::TracerContextFactory::Create(std::move(processors), resource_ptr);

    auto traceProvider = opentelemetry::sdk::trace::TracerProviderFactory::Create(std::move(context));
    auto tracer = traceProvider->GetTracer("OTel-Tracer-Foo");

    {
      auto spanFoo = tracer->StartSpan("Span-Foo");
    }

    {
      auto spanBar = tracer->StartSpan("Span-Bar");
    }
  }
```
As shown in the screenshot, spans Foo and Bar have distinct Trace IDs, which is the correct behavior.

<img width="1244" height="709" alt="Image" src="https://github.com/user-attachments/assets/2dcbd124-d532-49be-af79-a4b2cc4685e0" />

### Actual Behavior (ETW Exporter Tracer)
Using the ETW Exporter Tracer, based on the sample provided here:
https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/main/exporters/geneva-trace/example/main.cc
[...]

PR #4070, opened 2026-05-08
Fix: ETW span ownership lifetime

### LINKED ISSUE #3849: ETW Exporter: Crash When Ending ETW Exporter Span
The same code works correctly when using the OpenTelemetry Tracer, Span, and OTLP Exporter. However, the application crashes when switching to the ETW Exporter.

### Code to Reproduce (ETW Exporter)
```C++
  void TestEtwTracer()
  {
    tracer = traceProvider->GetTracer("Geneva-Tracer-Bar");
    spanFoo->End();
    spanBar->End();
  }
```

### Crash Screenshot

Image

### Code to Reproduce (OTLP Exporter – Works as Expected)
```C++
  void TestOTelTracer()
  {

    tracer = traceProvider->GetTracer("OTel-Tracer-Bar");
    spanFoo->End();
    spanBar->End();
  }
```

### Screenshot (No Crash)

Image

fmutlu68:
Fixes #3849

## Changes

- Remove per-span shared_ptr ownership in ETW StartSpan() hot path
- Add TracerProvider cache keyed by (provider name + encoding)
- Parse encoding from GetTracer(args) to select provider format
- Add regression test for tracer reassignment lifetime
- Document ETW tracer reuse model and hot-path guidance

## Testing

- ./ci/do_ci.sh cmake.c++17.test


fmutlu68:
Could you please have a look here ? also @ThomsonTan It passes all tests

lalitb:
ETW exporter is a bit different from other exporters: the tracer name maps to the ETW provider name/GUID, so creating multiple tracers for the same name is not the intended hot-path pattern. Applications should normally create one tracer per ETW provider name and reuse it for the process/component lifetime.

Because of that, I’d prefer not to fix this by adding `shared_from_this()`/`shared_ptr` ownership work to every span creation. `StartSpan()` is the high-throughput path for ETW. Could we instead document the intended ETW tracer reuse model and/or have `TracerProvider` retain created tracers so the lifetime protection is handled at setup/`GetTracer()` time rather than per span?

Also @ThomsonTan suggestions would be helpful, as he has more experience on this.

fmutlu68:
> ETW exporter is a bit different from other exporters: the tracer name maps to the ETW provider name/GUID, so creating multiple tracers for the same name is not the intended hot-path pattern. Applications should normally create one tracer per ETW provider name and reuse it for the process/component lifetime.
> 
> Because of that, I’d prefer not to fix this by adding `shared_from_this()`/`shared_ptr` ownership work to every span creation. `StartSpan()` is the high-throughput path for ETW. Could we instead document the intended ETW tracer reuse model and/or have `TracerProvider` retain created tracers so the lifetime protection is handled at setup/`GetTracer()` time rather than per span?
> 
> Also @ThomsonTan suggestions would be helpful, as he has more experience on this.

Thanks for the thorough explanation, I completely agree regarding the hot-path overhead; atomic reference counting in StartSpan() is not ideal for high-throughput ETW logging. The alternative approach makes perfect sense. I will revert the shared_from_this()/shared_ptr changes on the Tracer/Span and instead have the TracerProvider retain ownership of the created Tracers (e.g., via an internal map) to ensure lifetime protection at setup/GetTracer() time. I'll push the updated implementation shortly.

lukeina2z:
Can we cache these tracers in ETW TraceProvider. This would match the behavior of common TraceProvider in OTel. Here is the PR https://github.com/open-telemetry/opentelemetry-cpp/pull/4100. @fmutlu68 @lalitb @ThomsonTan

lukeina2z:
> Applications should normally create one tracer per ETW provider name and reuse it for the process/component lifetime.

The problem is the single tracer will always use an identical trace ID. Even you are handling requests from multiple clients. https://github.com/open-telemetry/opentelemetry-cpp/issues/3846

lalitb:
> Can we cache these tracers in ETW TraceProvider. This would match the behavior of common TraceProvider in OTel. Here is the PR #4100. @fmutlu68 @lalitb @ThomsonTan

Yes, this is what I suggested [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4070#issuecomment-4505459499) -  ` Could we instead document the intended ETW tracer reuse model and/or have TracerProvider retain created tracers so the lifetime protection is handled at setup/GetTracer() time rather than per span?`

lalitb:
> > Applications should normally create one tracer per ETW provider name and reuse it for the process/component lifetime.
> 
> The problem is the single tracer will always use an identical trace ID. Even you are handling requests from multiple clients. #3846

Yes, Agree. I believe tracer caching will solve the lifetime issue.

My main concern is that applications should avoid calling `GetTracer()->StartSpan()` in the hot path. Once `GetTracer()` is backed by a tracer cache, it will likely involve a mutex/map lookup, so putting it on the span creation path would add unnecessary synchronization overhead. For ETW, the intended usage should be to acquire the tracer once per provider name and reuse it, then call `StartSpan()` on that cached tracer.

fmutlu68:
@lalitb @ThomsonTan @lukeina2z I updated the PR based on the feedback around avoiding per-span ownership work in the ETW hot path.

Detailed summary of the changes:

- Reverted the `shared_from_this()` / per-span `shared_ptr` ownership approach from `StartSpan()`.
- Added `TracerProvider`-level tracer retention/cache so lifetime protection happens at `GetTracer()` time instead of span creation time.
- The tracer cache is keyed by `(provider name, encoding)` to avoid mixing tracers created for the same ETW provider name with different encodings.
- Added parsing for `GetTracer(args)` encoding values (`ETW`, `TLD`, `MSGPACK`, `XML`) so the cache key reflects the effective encoding.
- Added a regression test for the original #3849 repro where a span survives reassignment of the caller-held tracer.
- Updated the ETW README to document the intended reuse model: acquire one tracer per provider name/encoding during setup and reuse it, rather than calling `GetTracer()->StartSpan()` in hot paths.

I intentionally kept #3846 out of scope for this PR. This update focuses only on fixing the tracer/span lifetime issue without adding refcount overhead to `StartSpan()`.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1174,13 +1229,56 @@ class TracerProvider : public opentelemetry::trace::TracerProvider
 #endif
       ) noexcept override
   {
-    UNREFERENCED_PARAMETER(args);
     UNREFERENCED_PARAMETER(schema_url);
-    ETWProvider::EventFormat evtFmt = config_.encoding;
-    std::shared_ptr<opentelemetry::trace::Tracer> tracer{new (std::nothrow)
-                                                             Tracer(*this, name, evtFmt)};
-    return nostd::shared_ptr<opentelemetry::trace::Tracer>{tracer};
+    ETWProvider::EventFormat evtFmt = ParseEncodingArg(args, config_.encoding);
```
lalitb: As per the OTel spec (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer), the second argument to `GetTracer()` is the instrumentation `version`. Earlier ETW ignored this argument, so the mismatch did not matter much, but this PR now uses it to select the encoding. Is that intentional?

fmutlu68: Good catch. I have pushed an update that removes encoding selection from the args parameter entirely. args is now treated as the spec-defined instrumentation version (and simply marked unused by ETW, as before).
Encoding is now controlled exclusively via TelemetryProviderOptions["encoding"] passed to the TracerProvider constructor that flows into config_.encoding. The tracer cache key remains (name, config_.encoding) so different encodings still get separate tracers — just without abusing the args parameter.
Changes in the latest commit:

- Removed ParseEncodingArg() and TrimWhitespace() helpers (along with unused <algorithm> / <cctype> includes)
- GetTracer() now reads encoding from config_.encoding only
- Updated doc comments and README to reflect this

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1106,6 +1107,8 @@ class TracerProvider : public opentelemetry::trace::TracerProvider
         id_generator_{std::move(id_generator)},
         tail_sampler_{std::move(tail_sampler)}
   {
+    (void)Tracer::etwProvider().is_registered(std::string{});
```
ThomsonTan, line 1110: Still wondering if this is dead code.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1177,10 +1183,54 @@ class TracerProvider : public opentelemetry::trace::TracerProvider
...
     UNREFERENCED_PARAMETER(schema_url);
     ETWProvider::EventFormat evtFmt = config_.encoding;
-    std::shared_ptr<opentelemetry::trace::Tracer> tracer{new (std::nothrow)
-                                                             Tracer(*this, name, evtFmt)};
-    return nostd::shared_ptr<opentelemetry::trace::Tracer>{tracer};
+    TracerCacheKey key{name.data(), name.size(), evtFmt};
+
+    std::lock_guard<std::mutex> lock(tracers_mu_);
+    auto it = tracers_.find(key);
+    if (it != tracers_.end())
```
ThomsonTan, line 1190: The found tracer could be closed, which should be recreated even with cache hit. Or there will be problems in the below code.

```c++
auto t = tp.GetTracer("Foo");
t->Close(std::chrono::microseconds(0));
t = tp.GetTracer("Foo");
t->StartSpan("X");
```

ThomsonTan:
This introduced observable behavior (cached tracer instance), please add an entry to the CHANGELOG.

ThomsonTan (changes requested):

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
lalitb: I believe this issue still exists, and is blocker for the PR.

lalitb:
I’m not sure we should solve this with a tracer cache yet. It fixes the case where the caller overwrites the tracer
  variable, but it still does not make the lifetime model fully safe: if the `TracerProvider` is destroyed while a span is
  still alive, the cached tracer is destroyed too, and the span still holds a `Tracer&`. 

Since ETW writes synchronously and does not hand spans to a batch processor/exporter pipeline, maybe the simpler contract is acceptable here: document that the ETW provider/tracer must outlive any spans, and keep the deeper ownership change separate.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1177,10 +1183,54 @@ class TracerProvider : public opentelemetry::trace::TracerProvider
     UNREFERENCED_PARAMETER(args);
     UNREFERENCED_PARAMETER(schema_url);
     ETWProvider::EventFormat evtFmt = config_.encoding;
-    std::shared_ptr<opentelemetry::trace::Tracer> tracer{new (std::nothrow)
-                                                             Tracer(*this, name, evtFmt)};
-    return nostd::shared_ptr<opentelemetry::trace::Tracer>{tracer};
+    TracerCacheKey key{name.data(), name.size(), evtFmt};
```
ThomsonTan, line 1186: This cache key collapses distinct instrumentation scopes. Calls with the same `name` but different `version`, `schema_url`, or ABI v2 attributes all return the same tracer, while the SDK treats those identities as distinct.

Please compare the full instrumentation scope identity, preferably via `InstrumentationScope::equal()`, and add equality/inequality tests covering version, schema URL, and attributes.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -1177,10 +1183,54 @@ class TracerProvider : public opentelemetry::trace::TracerProvider
...
+  {
+    size_t operator()(const TracerCacheKey &key) const noexcept
+    {
+      size_t name_hash = std::hash<std::string>{}(key.name);
+      size_t fmt_hash  = static_cast<size_t>(key.encoding);
+      return name_hash ^ (fmt_hash + 0x9e3779b97f4a7c15ULL + (name_hash << 6) + (name_hash >> 2));
+    }
+  };
+
+  std::mutex tracers_mu_;
```
ThomsonTan, line 1232: Adding this mutex makes the public ETW `TracerProvider` neither move-constructible nor move-assignable, so existing code that moves this concrete provider no longer compiles.

Please preserve source compatibility by placing the mutex and cache behind movable, address-stable state.

PR #4071, opened 2026-05-08
[EXPORTER] Add custom HTTP client

### LINKED ISSUE #2084: Allow adding custom HTTP client
As of now, otel-cpp use Curl as the default HTTP Client, and is set [here ](https://github.com/open-telemetry/opentelemetry-cpp/blob/9bcfcb99ffb7e03f9f8d502ccc3c7991bc35fcef/CMakeLists.txt#L374) for CMake. It should be instead possible for users to bring their own HTTP Client library. Proposed solution could be:
  - Add cmake option (say WITH_CUSTOM_HTTP_CLIENT). And then WITH_HTTP_CLIENT_CURL  shouldn't be set if the new option is enabled.
  - User need to provide following as part of their application code - 
      -  Implementation for  `opentelemetry::ext::http::client::HttpClientFactory::Create()` and `opentelemetry::ext::http::client::HttpClientFactory::CreateSync()`. These two methods should return instances for custom implementation for HttpClient (async http client) and HttpClientSync (sync http client) as mentioned in step below.
      - Custom implementation for `ext::http::client::HttpClientSync` and `:ext::http::client::HttpClient` classes.
  - (Open Issue) `opentelemetry_http_client_curl` shouldn't be directly linked with other targets, probably some way to ensure that custom client library get's used at all the places.
  
This will ensure that only one definition/symbol for HttpClient and HttpClientSync is exposed (either from Custom http client, or the Curl http client) and would be invoked at run-time.

Also changes would be required in bazel build. 

We can use this issue to finalize solution before implementation.

Related issue : #1145 #1146

ltowarek:
Fixes #2084

## Changes

Adds a way to provide a custom HTTP client instead of the default `libcurl`-based client.
                                                                                                                                                              
The main motivation are embedded and cross-compiled targets where libcurl is unavailable. A working example for ESP32S3 using ESP-IDF's `esp_http_client` as the transport backend is at https://github.com/ltowarek/esp-opentelemetry-cpp/pull/30/.


owent:
Could you please add a test to use a simple custom http client?

lalitb:
Thanks @ltowarek. Couple of review comments as of now - 
1.   As @owent  mentioned, could we also add a minimal example for this option, even if the transport is just a no-op custom `HttpClient` / `HttpClientSync`? The important part is showing where the user provides `HttpClientFactory::Create()` and `CreateSync()`, and how their target is linked with the OTLP HTTP exporter.

2. Also, I believe there are still a few CMake paths that bring in curl directly. For example, Zipkin, Elasticsearch, HTTP examples, and the W3C tracecontext HTTP test server still link `opentelemetry_http_client_curl` / `CURL::libcurl`, while `WITH_CUSTOM_HTTP_CLIENT=ON` disables the curl client target globally. Can we either make this option explicitly incompatible with those curl-only paths, or update them to follow the same custom-client path?

exporters/otlp/CMakeLists.txt
```diff
@@ -318,13 +318,18 @@ if(WITH_OTLP_HTTP)
...
-           # Links flags of opentelemetry_http_client_curl should be public when
-           # building internal components
            "$<BUILD_INTERFACE:opentelemetry_proto>"
     PRIVATE "$<INSTALL_INTERFACE:opentelemetry_proto>"
-            "$<BUILD_INTERFACE:opentelemetry_http_client_curl>"
-            "$<BUILD_INTERFACE:nlohmann_json::nlohmann_json>"
-            "$<INSTALL_INTERFACE:opentelemetry_http_client_curl>")
+            "$<BUILD_INTERFACE:nlohmann_json::nlohmann_json>")
+
+  if(NOT WITH_CUSTOM_HTTP_CLIENT)
```
dbarker: can `WITH_HTTP_CLIENT_CURL=OFF` be used instead of creating a new CMake option? This existing option determines if the build includes the http curl client in `ext`.

dbarker (commented):
Thanks for the PR. Please see the question below.

exporters/otlp/CMakeLists.txt
ltowarek: Good idea. I've used [cmake_dependent_option](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#command:cmake_dependent_option), so a user can override the auto-detected value.

ltowarek:
Thanks for the feedback! I think I've covered everything in the new patch set:
* Added `opentelemetry_http_client` interface library to easily inject different implementations.
* Added tests based on `http_client::nosend::HttpClient`.
* Added an example app.
* Used `WITH_HTTP_CLIENT_CURL=OFF` instead of `WITH_CUSTOM_HTTP_CLIENT`.

marcalff (dismissed):
Thanks for the contribution.

Please consider to use:

* an abstract class `HttpClientFactory`, with virtual methods for `Create` and `CreateSync`
* A child class `HttpCurlClientFactory`, which implement this for CURL
* a HttpCurlClientFactory::singleton

Everywhere the static method HttpClientFactory::Create() is used today, pass a HttpClientFactory instance instead, taken as input parameter.

The user of an HTTP client can then decide which implementation to use:

* pass HttpCurlClientFactory::singleton to use CURL
* pass an instance of MyOwnHttpClientFactory for alternate implementations.

It should be possible to mix CURL and non CURL HTTP clients in the same binary, for example:

* send traces using HTTP + CURL to endpoint A
* send metrics using HTTP + alternate to endpoint B

ltowarek:
> 
> Please consider to use:
> 
>     * an abstract class `HttpClientFactory`, with virtual methods for `Create` and `CreateSync`
> 
>     * A child class `HttpCurlClientFactory`, which implement this for CURL
> 
>     * a HttpCurlClientFactory::singleton
> 
> 
> Everywhere the static method HttpClientFactory::Create() is used today, pass a HttpClientFactory instance instead, taken as input parameter.
> 
> The user of an HTTP client can then decide which implementation to use:
> 
>     * pass HttpCurlClientFactory::singleton to use CURL
> 
>     * pass an instance of MyOwnHttpClientFactory for alternate implementations.
> 
> 
> It should be possible to mix CURL and non CURL HTTP clients in the same binary, for example:
> 
>     * send traces using HTTP + CURL to endpoint A
> 
>     * send metrics using HTTP + alternate to endpoint B

@marcalff, it looks like a good idea.

My main goal with the current PR is build-time configuration - to enable `opentelemetry-cpp` for embedded systems as they simply can't build
 `libcurl`. The binary for such systems will always contain only one HTTP client.

I think runtime changes should be a part of a separate PR which I'm open to working on.

exporters/zipkin/CMakeLists.txt
```diff
@@ -23,9 +23,15 @@ set_target_version(opentelemetry_exporter_zipkin_trace)
 
 target_link_libraries(
   opentelemetry_exporter_zipkin_trace
-  PUBLIC opentelemetry_trace opentelemetry_http_client_curl
+  PUBLIC opentelemetry_trace "$<BUILD_INTERFACE:opentelemetry_http_client>"
```
owent: Why set it BUILD_INTERFACE only? opentelemetry_exporter_zipkin_trace depends opentelemetry_http_client's headers.

ltowarek: @owent good catch, thanks! I've added an installable target with factory headers.

ltowarek: I see that the installable target doesn't work in one scenario - I need to double check it.

ltowarek: Fixed.

ltowarek:
@marcalff I've added runtime changes to this PR. Please let me know if b8eb45ddcd450454129a907f9248ad2ddc2eceeb fits the needs.

lalitb:
Thanks for adding the runtime factory path. I think this is moving in the right direction, but can we make the injection story consistent across all HTTP exporters?

Right now `OtlpHttpExporterFactory` has overloads that accept a custom `HttpClientFactory` / `HttpClient`, but the OTLP HTTP log and metric exporter factories still only expose the default-client path. That means a user can inject a custom client for traces, but not for logs/metrics through the same public factory API.

Also, after making `HttpClientFactory::Create*()` virtual instance methods, there still seem to be a few static call sites left, for example the HTTP example and `test_ext_http_curl.cc`. Those should probably go through `GetDefaultHttpClientFactory()->Create*()` or a concrete curl factory.

ltowarek:
@lalitb thanks for the feedback! I've addressed your comments in the latest commits.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -661,9 +661,19 @@ void ConvertListFieldToJson(nlohmann::json &value,
 }  // namespace
 
 OtlpHttpClient::OtlpHttpClient(OtlpHttpClientOptions &&options)
+    : OtlpHttpClient(std::move(options), ext::http::client::GetDefaultHttpClientFactory())
```
lalitb: Can we keep the custom-client path independent from the default-client path?

If a user explicitly passes a custom `HttpClientFactory` or `HttpClient`, that should not require the default curl factory symbol to exist. The default factory should only be needed when a default constructor/factory overload is actually used.

Right now the exporter object files still contain default-client constructors like this one that reference `GetDefaultHttpClientFactory()`. In a no-curl/custom-only build, that can still create a link-time dependency on a symbol the user did not intend to use. Could we guard the default-client overloads when no default backend exists, or provide a no-curl fallback that fails clearly only if the default path is used?

owent (approved):
LGTM, thanks.

exporters/otlp/src/otlp_http_client.cc
ltowarek: @lalitb I see your point, but I'm not sure what's the best way to implement it. In my opinion, there are few options:
- **Option A** (current) - define `GetDefaultHttpClientFactory` only for CURL-based build and require custom builds to also define this function.
- **Option B** - guard default constructors using `GetDefaultHttpClientFactory` with `WITH_HTTP_CLIENT_CURL`. The API will look different depending on the build flags.
- **Option C** - provide stub for `GetDefaultHttpClientFactory` if CURL is disabled. Drawback is that custom builds can't easily inject their factories. Existing code using default constructors needs to be rewritten to use factory constructors.
- **Option D** - similar to the option C, but add a build option like `WITH_HTTP_CLIENT_DEFAULT_FACTORY`. This way custom builds could inject their factory if needed at the cost of yet another build flag.
- **Option E** - create a default factory registry and allow users to add their own factory in runtime that will be then used by `GetDefaultHttpClientFactory`. Here we're extending the API even more.

lalitb: @ltowarek  -   Thanks for laying out the options. I think the cleanest shape is to treat an explicitly provided HTTP client as the primary path, and the default factory as only a fallback path.

In other words, construction should first use the caller-provided `HttpClientFactory` / `HttpClient` when present. Only the default constructors/factory overloads should call `GetDefaultHttpClientFactory()`. If the build has no default HTTP backend and the caller uses a default overload, failing clearly is fine. But a custom client-only build should not need to define or link the default factory symbol.

So I’d prefer not to settle on Option A as the final shape. Guarding/splitting the default-client overloads, or adding a no-default fallback that only fails when the default path is actually used, seems closer to that model.

ltowarek: @lalitb I've pushed a change that makes `GetDefaultHttpClientFactory()` private.

marcalff (approved):
Thanks for the contribution.

This is good work.

I suspect `GetDefaultHttpClientFactory()` may need minor adjustments later, but this will be for a different PR.

examples/custom_http_client/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+add_library(custom_http_client_stub STATIC custom_http_client.cc)
+target_link_libraries(custom_http_client_stub PUBLIC opentelemetry_ext
+                                                     opentelemetry_sdk)
+
+add_executable(example_custom_http_client main.cc)
+target_link_libraries(
+  example_custom_http_client PRIVATE custom_http_client_stub
+                                     opentelemetry-cpp::otlp_http_exporter)
```
dbarker, line 11: Please add the new example to ctest 

```
if(BUILD_TESTING)
  add_test(NAME examples.custom_http_client
           COMMAND "$<TARGET_FILE:example_custom_http_client>")
endif()
```

examples/CMakeLists.txt
```diff
@@ -38,3 +38,7 @@ endif()
 if(WITH_CONFIGURATION)
   add_subdirectory(configuration)
 endif()
+
+if(NOT WITH_HTTP_CLIENT_CURL AND WITH_OTLP_HTTP)
```
dbarker: There doesn't appear to be a CI job that would meet this condition (`-DWITH_HTTP_CLIENT_CURL=OFF -DWITH_OTLP_HTTP=ON`) and build/run the `example_custom_http_client`. 

Can this condition just be `if(WITH_OTLP_HTTP)`? That would allow the example to be built with `iwyu`, `clang-tidy`, `codeql` and many other ci jobs.

dbarker (commented):
Thanks for the feature! Please see feedback below to build/test the new example in CI. Once that is in and CI is passing this should be good to merge.

examples/custom_http_client/CMakeLists.txt
ltowarek: Done

examples/CMakeLists.txt
ltowarek: Good catch! After recent refactoring, the example works fine even when the default `curl` path is enabled.

examples/custom_http_client/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+add_library(custom_http_client_stub STATIC custom_http_client.cc)
+target_link_libraries(custom_http_client_stub PUBLIC opentelemetry_ext
```
dbarker: The examples should link to the public cmake targets (`opentelemetry-cpp::ext` and `opentelemetry-cpp::sdk`).  This is causing the install tests to fail when trying to build the example against the installed otel-cpp package.

ltowarek: Yes, I'm on it.

ltowarek: Should be fixed now - my local build works fine.

dbarker (approved):

PR #4072, opened 2026-05-08
[SDK] Fix Base2ExponentialHistogramAggregation count dropped

### LINKED ISSUE #4068: `Base2ExponentialHistogramAggregation::Merge()` silently drops bucket counts when merged index range crosses zero (count > sum_of_buckets)
## Summary

When `Base2ExponentialHistogramAggregation::Merge()` is called with two histograms whose combined positive-bucket index range crosses zero (i.e. one input has only negative indices, the other has only or also positive indices), the downscale-trigger check on [line 315-318](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.25.0/sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc#L315-L318) silently fails because of a signed→unsigned cast wraparound. The merged histogram is then built with span > `max_buckets_`, causing `MergeBuckets()` to silently drop counts on `AdaptingCircularBufferCounter::Increment` overflow.

The exported point payload then violates the invariant `count_ == zero_count_ + Σ positive_buckets_[i] + Σ negative_buckets_[i]`, and downstream collectors (Prometheus / Mimir / etc.) reject it with errors like `ErrHistogramCountMismatch`.

This was found in production on `v1.25.0` (release 2026-02-08) but the same code path is unchanged on `master` HEAD as of 2026-05-08.

## Affected versions

- Confirmed reproduced on `v1.25.0`.
- `master` HEAD (commit checked 2026-05-08) has the same code; bug is still present.
- PR #3793 (merged 2025-12-18, included in v1.25.0) addressed the off-by-one (`>` → `>=`) in the same condition but **kept the `static_cast<size_t>` form**.
- PR #3997 (merged 2026-04-16) fixed an analogous cast wraparound in `GetScaleReduction()` but did not touch the corresponding check in `Merge()`/`Diff()`.

## Buggy code

[`sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc`](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.25.0/sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc#L315-L319) inside `Merge()`:

```cpp
auto pos_min_index = (std::min)(low_res.positive_buckets_->StartIndex(),
                                high_res.positive_buckets_->StartIndex());
auto pos_max_index = (std::max)(low_res.positive_buckets_->EndIndex(),
                                high_res.positive_buckets_->EndIndex());
// (analogous neg_min_index / neg_max_index)

if (static_cast<size_t>(pos_max_index) >=
        static_cast<size_t>(pos_min_index) + result_value.max_buckets_ ||
    static_cast<size_t>(neg_max_index) >=
        static_cast<size_t>(neg_min_index) + result_value.max_buckets_)
{
  const uint32_t scale_reduction = GetScaleReduction(...);
  // ... DownscaleBuckets ...
}
```

A nearly-identical check exists [in `Diff()` around line 388-391](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.25.0/sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc#L388-L391) using `>` instead of `>=`; same wraparound problem.

After the supposedly downscale-triggering check, `Merge()` calls [`MergeBuckets()`](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.25.0/sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc#L244-L251):

```cpp
for (int i = min_index; i <= max_index; i++) {
  auto count = A.Get(i) + B.Get(i);
  if (count > 0) {
    C.Increment(i, count);   // return value ignored
  }
}
```

`AdaptingCircularBufferCounter::Increment` returns `false` when the new index would extend the populated range beyond `max_buckets_`. Its return value is ignored here, so any out-of-range counts are silently dropped while `result_value.count_` (set earlier as `low_res.count_ + high_res.count_`) still includes them.

## Why the cast fails

`pos_min_index` and `pos_max_index` are `int32_t`. Casting a negative `int32_t` to `size_t` produces a value near `2^64 - 1`. The condition `static_cast<size_t>(pos_max_index) >= static_cast<size_t>(pos_min_index) + max_buckets_` therefore evaluates against a near-`UINT64_MAX` right-hand side, and is `false` for any plausible `pos_max_index` whenever `pos_min_index < 0 ≤ pos_max_index`.

## Concrete reproduction

A long-running cumulative-temporality consumer (`opentelemetry-cpp` v1.25.0 + OTLP gRPC exporter, default `max_buckets_ = 160`, `OTEL_METRIC_EXPORT_INTERVAL_MILLIS=10000`) records a single histogram instrument fed by a mixed-latency RPC workload:

- A "fast" phase produces values in the 1.3–30 ms range → bucket index range `[-304..-173]` at scale 5
- A "slow" phase produces values in the 1.7–13 s range → bucket index range `[26..116]` at scale 5

A few dozen slow observations per export interval are enough; once the cumulative side has buckets in `[-304..-173]` and the delta side has any positive index, every subsequent `Merge()` skips downscale due to the cast wraparound and `MergeBuckets()` drops the positive part.

We instrumented the SDK with a single 35-minute run (debug build) and observed:

- **1,467** events where `MergeBuckets()`'s `C.Increment(i, count)` returned `false` and dropped `count`
- **178** post-`Merge()` invariant violations (added a check after `MergeBuckets()` calls)

Sample diagnostic logs (one of many per cycle):

```
[OTEL_DBG] MergeBuckets_fail i=116 count=13 min=-304 max=116 max_buckets=160
           A_range=[-304..-173] B_range=[114..116] C_range=[-304..-173]
[OTEL_DBG] Merge_invariant count=2323996 total=2323966 gap=30 scale=5 zero=0
           pos_sum=2323966 neg_sum=0 low_count=2323966 high_count=30
           low_scale=5 high_scale=5 max_buckets=160
           pos_range=[-304..-173] neg_range=[0..0]
```

In the diagnostic output, `A_range` is `low_res.positive_buckets_` (`[StartIndex..EndIndex]`) and `B_range` is `high_res.positive_buckets_`. The `pos_min_index` / `pos_max_index` that the buggy condition operates on are then the unioned span across the two:

```
pos_min_index = min(A.StartIndex(), B.StartIndex())
pos_max_index = max(A.EndIndex(),   B.EndIndex())
```

Most-frequent `(A_range, B_range)` pairs across the run, with the unioned `[pos_min_index..pos_max_index]` shown alongside:

| count | A_range | B_range | unioned `[pos_min_index..pos_max_index]` |
|---|---|---|---|
| 95 | `[-304..-173]` | `[26..50]`   | `[-304..50]`  (span 355) |
| 90 | `[-308..-174]` | `[114..116]` | `[-308..116]` (span 425) |
| 90 | `[-304..-173]` | `[114..116]` | `[-304..116]` (span 421) |
| 72 | `[-308..-174]` | `[26..43]`   | `[-308..43]`  (span 352) |
| ... | ... | ... | ... |

For these pairs `pos_min_index < 0 ≤ pos_max_index`, so casting them to `size_t` makes `pos_min_index` near `UINT64_MAX` and the check is always `false`. Downscale therefore never fires, even though the actual span is 350-425 — well above `max_buckets_ = 160`.

The exported OTLP payload's `count_` is then larger than `Σ buckets`. Prometheus enforces this same invariant in [`model/histogram/histogram.go`](https://github.com/prometheus/prometheus/blob/main/model/histogram/histogram.go) and rejects such payloads.

## Proposed fix

Mirror the approach used in PR #3997 for `GetScaleReduction()` — promote to `int64_t` so the subtraction is signed and well-defined for the full `int32_t` index range:

```cpp
// Merge()
if (static_cast<int64_t>(pos_max_index) - pos_min_index + 1 >
        static_cast<int64_t>(result_value.max_buckets_) ||
    static_cast<int64_t>(neg_max_index) - neg_min_index + 1 >
        static_cast<int64_t>(result_value.max_buckets_))
{
  // downscale ...
}

// Diff() — same pattern, using low_res.max_buckets_
```

`(int32_t - int32_t)` can overflow `int32_t` arithmetic; promoting one operand to `int64_t` triggers usual-arithmetic-conversions and avoids both the signed overflow and the unsigned wraparound.

Verified by replay: applying this two-line change (one in `Merge()`, one in `Diff()`) on top of v1.25.0 and re-running the same 32-minute mixed-latency workload that previously produced 282 invariant violations brings the violation count to **0**, while the post-`Merge()` invariant check we used during diagnosis no longer fires.

Defensive add-on (optional, separate concern): change `MergeBuckets()`'s `C.Increment(i, count)` to detect failure (e.g. via assert / log / fall back to additional downscale) so future regressions of the upstream check don't silently corrupt counts.

## Impact

- All cumulative-temporality consumers of `Base2ExponentialHistogramAggregation` whose recorded values span more than `max_buckets_` worth of bucket indices and cross `index = 0` (i.e. `value = 1` in the metric's unit) are affected.
- Symptom downstream: payloads rejected with `count != sum(buckets)` errors at Prometheus-compatible TSDBs (Prometheus, Mimir, etc.), and silently lost data even when accepted by lenient backends.

## Related

- PR #3793: fixed off-by-one `>` → `>=` in the same condition, did not address the cast.
- PR #3997: fixed analogous cast in `GetScaleReduction()` (`size_t` → `int64_t` arithmetic), did not touch `Merge()`/`Diff()`.

dbarker:
Fixes #4068

## Changes

- Fix bucket downscaling issues in Merge and Diff caused by improper casts to size_t and not handling empty buckets or negative buckets. 
- Check the result of AdaptingCircularBufferCounter::Increment for Merge and Diff. Log an error and assert when the bucket index is out of range  
-  Test for Merge and Diff across a range of values
-  End to end test of Base2ExponentialHistogramAggregation with cumulative and delta temporality


lalitb (commented):
LGTM, nice work. Could we add one Diff-specific cross-zero test too? Merge is covered, but I don't think the current Diff tests hit that same negative-to-positive union case.

lalitb (approved):

PR #4078, opened 2026-05-12
[EXPORTER] OTLP HTTP exporter reads unbounded response

### LINKED ISSUE #3958: OTLP HTTP exporter reads unbounded response bodies into memory
## Description

The OTLP HTTP exporter in opentelemetry-cpp appears to read the full HTTP response body into memory without a size limit.

**Relevant code:**

[exporters/otlp/src/otlp_http_client.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.26.0/exporters/otlp/src/otlp_http_exporter.cc)
[ext/include/opentelemetry/ext/http/client/http_client.h](https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.26.0/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h)

**What looks problematic:**

- the HTTP client exposes the full response body as std::vector<uint8_t>
- the OTLP HTTP handler then copies it again into a std::string

So a malicious or compromised OTLP/HTTP endpoint could return a very large response body and cause excessive memory allocation, potentially leading to OOM.

This looks similar to the issue being discussed in other OTel SDKs:

[open-telemetry/opentelemetry-go GHSA-w8rr-5gcm-pp58](https://github.com/open-telemetry/opentelemetry-go/security/advisories/GHSA-w8rr-5gcm-pp58)

marcalff:
Fixes #3958

## Changes

* Implemented max size limits in the CURL HTTP code
* Restrict the amount of memory consumed when parsing HTTP responses.


ext/src/http/client/curl/http_operation_curl.cc
dbarker, line 72: Can we unit test exceeding the set limits?

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
```diff
@@ -43,6 +43,14 @@ const std::chrono::milliseconds kDefaultHttpConnTimeout(5000);  // ms
 const std::string kHttpStatusRegexp = "HTTP\\/\\d\\.\\d (\\d+)\\ .*";
 const std::string kHttpHeaderRegexp = "(.*)\\: (.*)\\n*";
 
+/**
+ * Default max HTTP Response size.
+ * 4MiB
+ * @see
+ * https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#otlphttp-response
+ */
+const size_t kDefaultMaxResponseSize = 4 * 1024 * 1024;
```
dbarker, line 52: Given that this is a curl specific fix and the limits really should be applied by any http client, should we move this to a common header with comments stating the client developer should implement limit checks?

dbarker (commented):
Thanks for the fix. Please see questions below.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -136,8 +194,31 @@ size_t HttpOperation::WriteVectorBodyCallback(void *ptr, size_t size, size_t nme
...
+      return 0;
+    }
+  }
+
+  if (self->response_body_.size() > self->max_response_body_size_)
+  {
+    return 0;
+  }
+
+  if (data_size > self->max_response_body_size_ - self->response_body_.size())
```
lalitb: I think this still allows more than the intended 4 MiB total response size. In the non-raw path, headers can grow up to 4 MiB and body can also grow up to 4 MiB, so together this can still buffer about 8 MiB. If the goal is a 4 MiB response limit by default, I think headers and body need to share the same budget.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: A unit test of this method alone has little value.

What would be nice is a functional test of the curl HTTP POST as a whole, against a server that returns an arbitrary big response, but I can not figure out how to do this looking at the existing curl tests, and even less integrate it in CI.

Suggestions are welcome.

ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
marcalff: There is only one HTTP implementation for now, which is this file using CURL.

When a different HTTP implementation is added, this can be reconsidered, but I don't think adding a different header file will help with anything, it has a high probability of being ignored by an alternate implementation anyway.

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Technically this observation is correct:

- raw curl responses will have a max size of 4MiB for the entire message
- non raw responses, parsed separately as a header and a body, will have a max size of 4MiB for each part, so the total is 8MiB at most.

The whole point of this fix is to protect against outrageous responses, to avoid memory consumption of GiB or more at a time, so a limit of 8MiB instead of 4Mib will not change anything regarding robustness and stability, in my opinion.

Beside, there are also some protections and limits in CURL itself, like `CURL_MAX_HTTP_HEADER`, which is 100KiB by default.

We can not rely on this only, because CURL might be compiled with a different value, but realistically the maximum memory use will be 4.1 MiB then for the header and body instead of the 4Mib -- _guidance_ -- from the opentelemetry-proto specification.

Is the added complexity to implement a common budget for the header and body worth it ?

marcalff: Now resolved, implemented a common budget.

The positive part is that it simplifies the interface, with only one size member.

lalitb (approved):

ext/src/http/client/curl/http_operation_curl.cc
marcalff: Actually, I agree this looks better now, so this is a good change.

PR #4079, opened 2026-05-12
[API+SDK] apply filtering EmitLogRecord

### LINKED ISSUE #2667: [API] Support for Logger::Enabled() is incomplete

proost:
Fixes #2667

## Changes

prev: https://github.com/open-telemetry/opentelemetry-cpp/pull/4011

fix: `Note that EmitLogRecord() helpers never honor the Enabled() flag either.` 

A huge gap exists between API/SDK spec and C++ client for `EmitLogRecord` method.

1. missing context parameter
2. missing exception parameter
3. support severity filtering in the "SDK".
4. We currently handle ObservedTimestamp and Timestamp using same type. [Spec tells both optional parameter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#emit-a-logrecord) how to divide which one is ObservedTimestamp and other one is Timestamp?

1 is easy, although ABI level breaking change is inevitable.
2 is quite confusing to me. Is just another attributes or `std::exception`? When i read [data-model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#errors-and-exceptions), It feels like one of attributes.
3 needs decision. Because In SDK level interface, we use "LogRecord" type. But we can't extract severity from "LogRecord" easily. Or do we have to overload "EmitLogRecord" with severity? 
4 needs a design. 

So i change very narrower code. Filtering in the severity filtering in the API level and trace based filtering in the sdk level. 


api/include/opentelemetry/logs/logger.h
dbarker, line 111: Per the spec this method should optionally accept a full Context 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-a-logrecord

If that were the case then any calls to get the current runtime context should be avoided.

sdk/src/logs/logger.cc
dbarker: The check above to logger_config_.IsEnabled() is not really complete and may be the wrong place to check if the logger is Enabled. The spec defines multiple conditions that can disable the logger. 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#enabled

> Enabled MUST return false when either:
> 
> - there are no registered [LogRecordProcessors](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#logrecordprocessor).
> Status: [Development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md) 
> - Logger is disabled ([LoggerConfig.enabled](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfig) is false).
> Status: [Development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md) 
> - the provided severity is specified (i.e. not 0) and is less than the configured minimum_severity in the [LoggerConfig](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfig).
> Status: [Development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md) 
> - trace_based is true in the [LoggerConfig](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#loggerconfig) and the current context is associated with an unsampled trace.
> - all registered LogRecordProcessors implement [Enabled](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#enabled-1), and a call to Enabled on each of them returns false.


The current implementation only checks for the LoggerConfig.enabled flag (skipping the other conditions). 

Also creating a log record now (when enabled) will always implicitly get the current context, even if the user has explicitly provided the context (trace id, span id, trace flags) on emit. If a user is able to and provides a full context on emit then getting the current context should be avoided. 

This may mean we need to update the Logger API EmitLogRecord API to accept a full context and only get the current context in the emit method if one is not provided. 

The Enabled check may then need to live in the EmitLogRecord method and guard creating a log record. If a user decides to call CreateLogRecord directly then perhaps it is their responsibility to call the Enabled method if needed.

sdk/src/logs/logger.cc
```diff
@@ -156,15 +156,18 @@ void Logger::EmitLogRecord(
     return;
   }
 
+  if (!IsAllowedByTraceBasedFiltering(context::RuntimeContext::GetCurrent(), logger_config_))
```
dbarker: Getting the runtime context here duplicates the call in CreateLogRecord. In general if the user provides a full context on emit then all other calls that take a context should use the one explicitly provided and avoid getting the current.

dbarker (commented):
Thanks for the PR. We definitely need some discussion here (thanks for starting it) on how to align with the spec. As you noted there are a few gaps in how the full Context is handled and Enabled is checked when creating and emitting a log record.  Generally it seems there should be a path to explicitly provide the full Context and avoid any implicit calls to get the current context from `RuntimeContext`.

api/include/opentelemetry/logs/logger.h
proost: [a0958a4](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/a0958a46a7fbc308b60b4eba554b8811bd64adfa)

Ok, now i added.

sdk/src/logs/logger.cc
proost: OK, here is what i understand. We align spec through API level using current "Enabled". 
[47d1e37](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/47d1e37f37cb378c2e0ed1040189fbbb42ed21a0)

Is this correct?

sdk/src/logs/logger.cc
proost: https://github.com/open-telemetry/opentelemetry-cpp/pull/4079#discussion_r3240070939

owent (approved):
LGTM, thanks.

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +142,42 @@ class Logger
...
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+    const opentelemetry::context::Context *context_ptr = detail::FindContextInArgs(args...);
+#endif  // OPENTELEMETRY_ABI_VERSION_NO >= 2
+
+    const Severity arg_severity = detail::FindSeverityInArgs(args...);
+    if (arg_severity != Severity::kInvalid)
+    {
+      const EventId *event_id_ptr = detail::FindEventIdInArgs(args...);
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+      const bool is_enabled =
```
lalitb: Can we avoid making every enabled log call pay for the full `Enabled(...)` chain? The disabled and below-min-severity paths stay cheap, but the normal enabled path now pays virtual/context/processor cost before record creation. It may be better to cache whether full filtering is actually needed, and only call the full chain when trace-based filtering or processor-level filtering is configured.

proost: Thx, [5e67680](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/5e676808e1d841b9f5f085f75806d8b36c2222f9) 

how about go through trace based filtering using current option?

lalitb: Thanks, this helps with the trace-based case, but I think the original concern is only partially addressed. We still need the cache to account for processor-level filtering too, otherwise the normal non-trace-based path can skip processor `Enabled(...)` checks.

proost: Ok, now i understand your intention.
[dc306d5](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/dc306d58f01085c8a6dfdb2560ebb0d03d61a2ee)

Later, i will add [logger configuration.](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/logger_provider.yaml#L185). this option might be needed in the configuration. But this is not in the spec. is ok?

sdk/src/logs/logger.cc
dbarker: Thanks. Looks reasonable. There still seems to be a gap in handling user provided Context in parts (as SpanContext or TraceID + SpanID + TraceFlags): 

1. User calls EmitLogRecord(args...) with SpanContext -> Enabled called with Current Runtime Context instead of a Context with the provided SpanContext?
2. User calls EmitLogRecord(args...) with TraceID, SpanID, and TraceFlags -> Enabled is checked against Current Runtime Context instead of a Context from the provided trace parts?

sdk/src/logs/logger.cc
```diff
@@ -143,6 +146,56 @@ opentelemetry::nostd::unique_ptr<opentelemetry::logs::LogRecord> Logger::CreateL
...
+  if (!logger_config_.IsEnabled())
+  {
+    return kNoopLogger.CreateLogRecord();
+  }
+
+  auto recordable = context_->GetProcessor().MakeRecordable();
+
+  recordable->SetObservedTimestamp(std::chrono::system_clock::now());
+
+  // Get the span metadata from the supplied context
```
dbarker: Please move the common SpanContext extract and set code to a function that both CreateLogRecord methods can call.

api/test/logs/logger_test.cc
```diff
@@ -461,3 +477,97 @@ TEST(Logger, EnabledWithExplicitContextUsesContextAwareImplementation)
...
+                       test_context);
+
+  EXPECT_EQ(logger.enabled_with_event_id_calls_, 1u);
+  EXPECT_EQ(logger.create_log_record_context_calls_, 1u);
+  EXPECT_EQ(logger.emit_log_record_calls_, 1u);
+  EXPECT_TRUE(logger.last_enabled_context_has_test_key_);
+  EXPECT_TRUE(logger.last_enabled_context_test_key_value_);
+}
+
+TEST(Logger, EmitLogRecordWithContextInArgsShortCircuitsWhenEnabledImplementationReturnsFalse)
```
dbarker, line 587: Please add tests for calling EmitLogRecord with SpanContext and trace parts (TraceID + SpanID + TraceFlags)

sdk/src/logs/logger.cc
proost: [d62c8e4](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/d62c8e4eb0fd951fa354f417b1900d699ca23a70)

sdk/src/logs/logger.cc
proost: [d4578ad](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/d4578ad2cf9203abc38402c9533e04cec4b793fc)

We should cover this too. Thank you.

api/test/logs/logger_test.cc
proost: [d4578ad](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/d4578ad2cf9203abc38402c9533e04cec4b793fc)

sdk/src/logs/logger.cc
dbarker: Thanks for the update. This change appears to result in the right behavior. If a user provides trace args independently as `SpanContext` or `TraceID+SpanID+TraceFlags` then those args are used to determine if the record should be created and to set the record's traceid, spanid, and traceflags. 

The downside for current users is the extra dynamic memory allocations to reconstruct a full Context from the args (allocating a DefaultSpan just to later get the trace data from it). We should avoid this cost for those users if possible. 

One option: Instead of constructing a full Context object when a user provides `SpanContext` or `TraceID+SpanID+TraceFlags` we pass just a SpanContext (construct if needed) through to the sdk methods. We could do this with a nostd::variant<SpanContext, context::Context> similar to `StartSpanOptions` or define new struct. That would require updating the ABIv2 interface for CreateLogRecord, EnabledImplementation, and IsAllowedByTraceBasedFiltering to take this object referencing either a full Context or SpanContext. 

What do you think? @lalitb any thoughts on this approach?

lalitb: I agree in principle. The behavior looks correct after the latest update, but the extra allocation for `DefaultSpan` is unfortunate, especially because we only need the `SpanContext` for trace_based filtering and stamping the record.

A lightweight wrapper/variant for “full Context vs explicit SpanContext” seems reasonable to me. Since this is ABIv2-only surface, I think it is better to get that shape right now. The important part is preserving the current precedence: explicit `Context` first, explicit trace args next, and `RuntimeContext` only when nothing explicit was provided.

api/include/opentelemetry/logs/logger.h
lalitb: I would not add a new logger configuration option just for this. The spec defines the observable Enabled behavior, but the C++ SDK can still implement it efficiently.

For performance, I think this should stay as an internal cache/capability decision: full filtering is needed when trace_based is enabled or when a processor actually implements Enabled filtering. Otherwise the normal enabled path should avoid paying the full virtual/context/processor chain.

One thing to watch is that the cache must stay correct if processors are added after a logger is created.

api/include/opentelemetry/logs/logger.h
```diff
@@ -535,6 +656,13 @@ class Logger
   // compatible for OpenTelemetry C++ API.
   //
   mutable uint8_t minimum_severity_{kMaxSeverity};
+
+  //
+  // extended_enabled_required_ defaults to 1 (full chain required) so subclasses that override
+  // EnabledImplementation keep their existing semantics until they explicitly opt out via
+  // SetExtendedEnabledRequired(false). Same atomicity rationale as minimum_severity_.
+  //
+  mutable uint8_t extended_enabled_required_{1};
```
lalitb: Do we need this to default to false for API compatibility? Existing custom Logger implementations may rely on the base Info/Warn/Error/EmitLogRecord(args...) helpers without overriding `EnabledImplementation(...)`. With this defaulting to true, those calls can now hit the base `EnabledImplementation`, which returns false, and logs that used to be emitted may get dropped. The SDK Logger can still opt in to the full filtering path when it needs it.

proost: [e717aab](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/e717aab9a1d918a4af487fdb85e160b4b79b2c48)

api/include/opentelemetry/logs/logger.h
proost: OK, I integrate Context And SpanContext both allowed. [8a18dc7](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/8a18dc7954a5c10358c64ce24cc5fcdc4df6a11d)

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +151,101 @@ class Logger
...
+            extended_enabled = EnabledImplementation(context_or_span, arg_severity, *event_id_ptr);
+          }
+          else
+          {
+            extended_enabled = EnabledImplementation(context_or_span, arg_severity);
+          }
+        }
+        else
+        {
+          if (event_id_ptr)
```
dbarker: In the case where the `context_or_span` is not set the runtime context will be retrieved twice with `opentelemetry::context::RuntimeContext::GetCurrent()`: Once in `EnabledImplementation` and once in `CreateLogRecord`. Please update this to retrieve the context once if possible and pass into `EnabledImplementation` and `CreateLogRecord`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -50,6 +56,23 @@ class Logger
...
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  /**
+   * Create a Log Record object using either a Context or a SpanContext.
+   *
+   * @param context_or_span Variant carrying either a full Context or just a
+   *                        SpanContext. Avoids allocating a Context purely to
+   *                        propagate trace identity.
+   * @return nostd::unique_ptr<LogRecord>
+   */
+  virtual nostd::unique_ptr<LogRecord> CreateLogRecord(
```
dbarker, line 68: Running `./ci/do_ci.sh cmake.maintainer.abiv2.test` with GCC reports several new `-Werror=overloaded-virtual=` warnings. This is currently a gap in CI. Please take a look. 

```
FAILED: api/test/logs/CMakeFiles/logger_benchmark.dir/logger_benchmark.cc.o 
/usr/bin/c++ -DENABLE_METRICS_EXEMPLAR_PREVIEW -DENABLE_OTLP_COMPRESSION_PREVIEW -DENABLE_OTLP_GRPC_CREDENTIAL_PREVIEW -DENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW -DENABLE_OTLP_RETRY_PREVIEW -DENABLE_THREAD_INSTRUMENTATION_PREVIEW -DOPENTELEMETRY_ABI_VERSION_NO=2 -DOPENTELEMETRY_NO_DEPRECATED_CODE -I/workspaces/opentelemetry-cpp/api/include -isystem /home/devuser/third-party/install/include -g -std=c++17 -Wno-error=deprecated-declarations -Wall -Werror -Wextra -Wextra-semi -Woverloaded-virtual -Wsuggest-override -Wold-style-cast -Wcast-qual -Wformat-security -Wlogical-op -Wmissing-include-dirs -Wstringop-truncation -Wundef -Wvla -MD -MT api/test/logs/CMakeFiles/logger_benchmark.dir/logger_benchmark.cc.o -MF api/test/logs/CMakeFiles/logger_benchmark.dir/logger_benchmark.cc.o.d -o api/test/logs/CMakeFiles/logger_benchmark.dir/logger_benchmark.cc.o -c /workspaces/opentelemetry-cpp/api/test/logs/logger_benchmark.cc
In file included from /workspaces/opentelemetry-cpp/api/test/logs/logger_benchmark.cc:17:
/workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/logger.h:68:40: error: 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> opentelemetry::v2::logs::Logger::CreateLogRecord(const absl::otel_v1::variant<opentelemetry::v2::trace::SpanContext, opentelemetry::v2::context::Context>&)' was hidden [-Werror=overloaded-virtual=]
   68 |   virtual nostd::unique_ptr<LogRecord> CreateLogRecord(
      |                                        ^~~~~~~~~~~~~~~
In file included from /workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/provider.h:10,
                 from /workspaces/opentelemetry-cpp/api/test/logs/logger_benchmark.cc:19:
/workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/noop.h:37:32: note:   by 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> opentelemetry::v2::logs::NoopLogger::CreateLogRecord()'
   37 |   nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override
      |  

[102/679] Building CXX object api/test/logs/CMakeFiles/logs_api_logger_test.dir/logger_test.cc.o
FAILED: api/test/logs/CMakeFiles/logs_api_logger_test.dir/logger_test.cc.o 
/usr/bin/c++ -DENABLE_METRICS_EXEMPLAR_PREVIEW -DENABLE_OTLP_COMPRESSION_PREVIEW -DENABLE_OTLP_GRPC_CREDENTIAL_PREVIEW -DENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW -DENABLE_OTLP_RETRY_PREVIEW -DENABLE_THREAD_INSTRUMENTATION_PREVIEW -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -DOPENTELEMETRY_ABI_VERSION_NO=2 -DOPENTELEMETRY_NO_DEPRECATED_CODE -I/workspaces/opentelemetry-cpp/api/include -isystem /home/devuser/third-party/install/include -g -std=c++17 -Wno-error=deprecated-declarations -Wall -Werror -Wextra -Wextra-semi -Woverloaded-virtual -Wsuggest-override -Wold-style-cast -Wcast-qual -Wformat-security -Wlogical-op -Wmissing-include-dirs -Wstringop-truncation -Wundef -Wvla -MD -MT api/test/logs/CMakeFiles/logs_api_logger_test.dir/logger_test.cc.o -MF api/test/logs/CMakeFiles/logs_api_logger_test.dir/logger_test.cc.o.d -o api/test/logs/CMakeFiles/logs_api_logger_test.dir/logger_test.cc.o -c /workspaces/opentelemetry-cpp/api/test/logs/logger_test.cc
In file included from /workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/event_logger.h:7,
                 from /workspaces/opentelemetry-cpp/api/test/logs/logger_test.cc:16:
/workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/logger.h:68:40: error: 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> opentelemetry::v2::logs::Logger::CreateLogRecord(const absl::otel_v1::variant<opentelemetry::v2::trace::SpanContext, opentelemetry::v2::context::Context>&)' was hidden [-Werror=overloaded-virtual=]
   68 |   virtual nostd::unique_ptr<LogRecord> CreateLogRecord(
      |                                        ^~~~~~~~~~~~~~~
In file included from /workspaces/opentelemetry-cpp/api/test/logs/logger_test.cc:21:
/workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/noop.h:37:32: note:   by 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> opentelemetry::v2::logs::NoopLogger::CreateLogRecord()'
   37 |   nostd::unique_ptr<LogRecord> CreateLogRecord() noexcept override
      |                                ^~~~~~~~~~~~~~~
/workspaces/opentelemetry-cpp/api/include/opentelemetry/logs/logger.h:68:40: error: 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> opentelemetry::v2::logs::Logger::CreateLogRecord(const absl::otel_v1::variant<opentelemetry::v2::trace::SpanContext, opentelemetry::v2::context::Context>&)' was hidden [-Werror=overloaded-virtual=]
   68 |   virtual nostd::unique_ptr<LogRecord> CreateLogRecord(
      |                                        ^~~~~~~~~~~~~~~
/workspaces/opentelemetry-cpp/api/test/logs/logger_test.cc:268:53: note:   by 'virtual opentelemetry::v2::nostd::unique_ptr<opentelemetry::v2::logs::LogRecord> TestLogger::CreateLogRecord()'
  268 |   nostd::unique_ptr<opentelemetry::logs::LogRecord> CreateLogRecord() noexcept override
```

sdk/test/logs/logger_sdk_test.cc
```diff
@@ -306,21 +309,30 @@ class EnablementAwareProcessor final : public LogRecordProcessor
   bool Shutdown(std::chrono::microseconds /* timeout */) noexcept override { return true; }
 
 protected:
-  bool EnabledImplementation(const context::Context &context,
-                             const InstrumentationScope &instrumentation_scope,
-                             logs_api::Severity severity,
-                             nostd::string_view event_name) const noexcept override
+  bool EnabledImplementation(
```
dbarker, line 315: Please add a few test cases to verify the log record trace id, span id, and trace flags are set when an explicit Context or SpanContext is passed to EmitLogRecord. The API tests cover the input arguments but do not verify the data is set on the log record.

dbarker (commented):
Thanks for the updates. This is looking good. Please see a few minor points below.

api/include/opentelemetry/logs/logger.h
proost: [c6d5684](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/c6d568428d0c6303fb4e392b1adb47265c7cccd7)

api/include/opentelemetry/logs/logger.h
proost: Thank you! 

[7ca49f2](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/7ca49f2a775ccb3d2b50db04a8c26cdaa48027ea)

It would be good if those options were included in the CI.

sdk/test/logs/logger_sdk_test.cc
proost: ce5d3e0

api/include/opentelemetry/logs/logger.h
```diff
@@ -535,6 +653,17 @@ class Logger
...
+  //
+  // Controls whether the EmitLogRecord(args...) template calls the
+  // EnabledImplementation virtual in addition to the cheap atomic
+  // minimum_severity_ check. When 0, the template trusts the atomic check
+  // as the complete enabled decision and skips the virtual dispatch. When
+  // 1, the template also consults EnabledImplementation, which lets a
+  // Logger subclass apply richer filtering (trace-based, processor-level
+  // Enabled, custom predicates).
+  //
+  mutable uint8_t extended_enabled_required_{0};
```
dbarker, line 663: This is an ABI breaking change. Please guard the new member on ABI v2 along with locations where used.

proost: [956f820](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/956f820c9d3c12faea40a702d35eddb7e5cdd17f)

Sorry to miss that. you are right. It should be gated.

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +151,84 @@ class Logger
...
+            event_id_ptr ? EnabledImplementation(context_or_span, arg_severity, *event_id_ptr)
+                         : EnabledImplementation(context_or_span, arg_severity);
+        if (!extended_enabled)
+        {
+          return;
+        }
+      }
+#else
+      const bool extended_enabled =
+          event_id_ptr ? EnabledImplementation(arg_severity, *event_id_ptr)
```
lalitb: This feels inconsistent between ABI versions. In v2, the richer `EnabledImplementation(...)` path is optional through `ExtendedEnabledRequired()`, so the common severity-only path stays cheap. But in v1, where we cannot add that cache without changing layout, the richer path is now always on by default. I think v1 should preserve the existing severity-only behavior and leave the richer filtering path to v2.

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +151,84 @@ class Logger
...
+   *  @c SpanId [+ @c TraceFlags]) are in args, the filter chain uses
+   *  @c Enabled(context_or_span, severity, ...) and the record is created
+   *  via @c CreateLogRecord(context_or_span), so the filter evaluates
+   *  against the trace this record is for instead of the implicit runtime
+   *  context.
    */
   template <class... ArgumentType>
   void EmitLogRecord(ArgumentType &&...args)
   {
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
```
lalitb, line 167: Can we move the cheap severity check before building `context_or_span`? In the no-explicit-context case we currently call `RuntimeContext::GetCurrent()` before checking `Enabled(arg_severity)`, so below-minimum-severity logs still pay the context lookup even though they return immediately.

proost: [8cae295](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/8cae2952550b26a666648d954f22765348e83356)

api/include/opentelemetry/logs/logger.h
proost: [8cae295](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/8cae2952550b26a666648d954f22765348e83356)

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+  return FindSeverityInArgs(std::forward<Rest>(rest)...);
+}
+
+inline const EventId *FindEventIdInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const EventId *FindEventIdInArgs(const EventId &event_id, Rest &&.../*rest*/) noexcept
```
dbarker: nit - please add OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND here to event_id so CI will catch if there is misuse.

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+inline const opentelemetry::context::Context *FindContextInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const opentelemetry::context::Context *FindContextInArgs(
+    const opentelemetry::context::Context &context,
```
dbarker: nit - use OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+  return FindContextInArgs(std::forward<Rest>(rest)...);
+}
+
+inline const trace::SpanContext *FindSpanContextInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const trace::SpanContext *FindSpanContextInArgs(const trace::SpanContext &span_context,
```
dbarker: nit - use OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+  return FindSpanContextInArgs(std::forward<Rest>(rest)...);
+}
+
+inline const trace::TraceId *FindTraceIdInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const trace::TraceId *FindTraceIdInArgs(const trace::TraceId &trace_id,
```
dbarker: nit - use OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+  return FindTraceIdInArgs(std::forward<Rest>(rest)...);
+}
+
+inline const trace::SpanId *FindSpanIdInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const trace::SpanId *FindSpanIdInArgs(const trace::SpanId &span_id,
```
dbarker: nit - use OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND

api/include/opentelemetry/logs/logger_type_traits.h
```diff
@@ -198,6 +217,160 @@ struct LogRecordHasType<ValueType, TargetType, ArgumentType...>
...
+  return FindSpanIdInArgs(std::forward<Rest>(rest)...);
+}
+
+inline const trace::TraceFlags *FindTraceFlagsInArgs() noexcept
+{
+  return nullptr;
+}
+
+template <class... Rest>
+inline const trace::TraceFlags *FindTraceFlagsInArgs(const trace::TraceFlags &trace_flags,
```
dbarker: nit - use OPENTELEMETRY_ATTRIBUTE_LIFETIME_BOUND

sdk/test/logs/logger_sdk_test.cc
```diff
@@ -653,6 +667,95 @@ TEST(LoggerSDK, LoggerTraceBasedConfigAllowsSampledExplicitContextWithNamedEvent
...
+
+  logger->EmitLogRecord(logs_api::Severity::kInfo, nostd::string_view{"msg"}, explicit_span_ctx);
+
+  EXPECT_EQ(shared_recordable->GetTraceId(), explicit_span_ctx.trace_id());
+  EXPECT_EQ(shared_recordable->GetSpanId(), explicit_span_ctx.span_id());
+  EXPECT_EQ(shared_recordable->GetTraceFlags(), explicit_span_ctx.trace_flags());
+  EXPECT_NE(shared_recordable->GetTraceId(), runtime_span->GetContext().trace_id());
+}
+
+TEST(LoggerSDK, EmitLogRecordWithExplicitTracePartsStampsTraceFields)
```
dbarker: nit - optional clean up to use a common test fixture to set up the logger and tracer providers.

CHANGELOG.md
```diff
@@ -45,6 +45,30 @@ Increment the:
 * [CODE HEALTH] Move sdk_builder.cc builders into anonymous namespace
   [#4122](https://github.com/open-telemetry/opentelemetry-cpp/pull/4122)
 
+* [API] `Logger::EmitLogRecord(...)` templates now apply the `Enabled` filter
+  chain when a `Severity` is in args.
+  [#2667](https://github.com/open-telemetry/opentelemetry-cpp/issues/2667)
+
+* [API/SDK] (ABI v2) Add `Logger::CreateLogRecord(const Context &)` virtual
```
dbarker: nit - update to `Logger::CreateLogRecord( const nostd::variant<trace::SpanContext, context::Context> &)`

dbarker (approved):
Thanks for the spec compliance fixes here. Nice work. Please see followup nitpicks and optional test cleanup below.

api/include/opentelemetry/logs/logger_type_traits.h
proost: [5081253](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/5081253aa71b2b5c608c0b6c5f8522eb8376694c)

sdk/test/logs/logger_sdk_test.cc
proost: [b293548](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/b293548a8b42b6dd6af26b6ecb9e3d23482a7e53)

CHANGELOG.md
proost: Thank you! I missed.

[af777f4](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/af777f491e27e074fdd14abd39b598ff8ea2a4b2)

dbarker (approved):

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +151,72 @@ class Logger
...
+   *  @c Enabled(context_or_span, severity, ...) and the record is created
+   *  via @c CreateLogRecord(context_or_span), so the filter evaluates
+   *  against the trace this record is for instead of the implicit runtime
+   *  context.
    */
   template <class... ArgumentType>
   void EmitLogRecord(ArgumentType &&...args)
   {
+    const Severity arg_severity = detail::FindSeverityInArgs(args...);
+    if (arg_severity != Severity::kInvalid && !Enabled(arg_severity))
```
lalitb, line 169: Can we keep this `EmitLogRecord(...)` behavior change ABI v2 only? Fixing `Logger::Enabled()` itself makes sense for v1 too, but this check also changes v1 direct emit behavior: below-minimum-severity logs that used to pass through will now be dropped. Since the issue is marked as an ABI v2/breaking-change fix, I think v1 should keep the old emit behavior and only v2 should make the helpers honor `Enabled()`.

api/include/opentelemetry/logs/logger.h
```diff
@@ -120,11 +151,72 @@ class Logger
...
+      else
+      {
+        context_or_span = opentelemetry::context::RuntimeContext::GetCurrent();
+      }
+    }
+
+    if (arg_severity != Severity::kInvalid)
+    {
+      const EventId *event_id_ptr = detail::FindEventIdInArgs(args...);
+      if (ExtendedEnabledRequired())
```
lalitb, line 205: non-blocking: I think this cache can still be stale for the supported setup flow where a default `LoggerProvider` creates a logger before `AddProcessor()` is called. `SetExtendedEnabledRequired(...)` is computed in the logger constructor, so if a filtering processor is attached later, that existing logger can keep skipping the processor `Enabled(...)` chain.

lalitb (changes requested):
Requesting changes only to avoid an accidental merge while we settle the v1 behavior question. I left the specific concern [inline](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079#discussion_r3370293274); otherwise this is looking very close.

api/include/opentelemetry/logs/logger.h
proost: OK, i agree. [2c36af1](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/2c36af156cedc9c7415f45a0f59c5ea4031b77dc)

dbarker: @lalitb Thanks for flagging the behavior change (with this PR log records will be dropped if their severity is less than the minimum severity). This is intended to comply with the spec. 

> Minimum severity: If the log record’s [SeverityNumber](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber) is specified (i.e. not 0) and is less than the configured minimum_severity, the log record MUST be dropped.

It seems we should generally apply spec compliant behavior changes independent of the ABI compatibility requirements and make sure to document them in the CHANGELOG.  In this case the default `LoggerConfig` will correctly set the `minimum_severity` to `0` per the spec and all user logs will pass the severity level filter. The majority of users likely use the default since min severity was not used to drop logs before.  

I'd advocate we revert  [2c36af1](https://github.com/open-telemetry/opentelemetry-cpp/pull/4079/commits/2c36af156cedc9c7415f45a0f59c5ea4031b77dc) and keep the spec compliant behavior change independent of ABI version as long as ABI compatibility with v1 is maintained. What do you think?

lalitb: @dbarker I agree the spec says logs below the configured minimum severity should be dropped, and fixing `Logger::Enabled()` itself is fine.

My concern is applying this helper/API behavior change to ABI v1. This is bigger than an SDK-level behavior change for applications, because the log helpers like `Debug(...)`, `Info(...)`, and `Log(...)` are the main path used by instrumentation libraries, and they all go through `EmitLogRecord(...)`.

Changing that behavior in v1 can affect instrumentation libraries and all their users. I don’t think asking instrumentation authors to update their libraries to get back the old v1 behavior is a good path. I’d prefer to keep v1 behavior stable for instrumentation libraries and make this spec-compliant helper behavior part of ABI v2.

lalitb (approved):
LGTM from my side on the current code shape.

api/include/opentelemetry/logs/logger.h
dbarker: Is the concern for the instrumentation libraries and their users mitigated largely by setting the default `LoggerConfig::minimum_severity_` to 0 ([`opentelemetry::logs::Severity::kInvalid`](https://github.com/open-telemetry/opentelemetry-cpp/blob/ceee850cf9568f2fd34286c666de8948c122ddb6/sdk/include/opentelemetry/sdk/logs/logger_config.h#L78))? 

Those instrumentation libraries would have to change the default to see a behavior change (logs dropped) and not respecting the minimum severity value they set feels more like a bug.  
 
There is a case where a Logger derived from API logger that does not change the [API  `Logger::minimum_severity_` default](https://github.com/open-telemetry/opentelemetry-cpp/blob/ceee850cf9568f2fd34286c666de8948c122ddb6/api/include/opentelemetry/logs/logger.h#L537) would drop all logs based on the severity filter. This seems to have been intended to be the default for the noop logger and could be set there instead, while allowing the API logger to set the spec default `0`. 

My concern is applying spec compliance behavior changes inconsistently on ABI version, which will be hard to track and maintain. In some cases spec compliance may need to go through a preview flag and opt in period. I don't feel like this is such a case, but it could be.

lalitb: @dbarker -  The default `minimum_severity = 0` does mitigate the default SDK case, and I agree that `Logger::Enabled()` itself should follow the configured minimum severity.

The case I’m worried about is when an application configures a non-default minimum severity, for example `Warn`, and uses multiple instrumentation libraries. Since these helpers are header/template code, one instrumentation library may have been built with the older v1 helper behavior and still emit `Debug(...)`/`Info(...)`, while another built with newer v1 headers would start dropping those records before record creation.

So the behavior can differ across instrumentation libraries in the same application depending on which v1 helper behavior they were built with. That feels broader than an SDK-only behavior change, even if the new behavior is spec-correct.

dbarker: Thanks. The example of multiple releases used in one application helps. Would a preview flag be the right path then? `WITH_LOG_SEVERITY_FILTER_PREVIEW=ON`

The preview flag can communicate the new behavior and provide a few releases for instrumentation library maintainers to make changes (or opt out) if needed. This way ABI v1 can support log severity based filtering.

lalitb: @dbarker A preview flag sounds reasonable for ABI v1.

One thing I’m still thinking about is that this helper code is compiled into instrumentation libraries. If an application uses prebuilt/shared instrumentation libraries, it may not know which ones were built with the flag enabled.

 So the flag helps with rollout, but we should document this clearly for both instrumentation library maintainers and application owners. ABI v2 can still make the spec-compliant behavior the default.

PR #4085, opened 2026-05-15
[SDK] MeterProvider: do not warn in destructor after explicit Shutdown

### LINKED ISSUE #3511: Cannot Shutdown() MeterProvider manually without a warning from destructor
Sometimes it makes sense to shut down a meter provider manually. However, the destructor of `MeterProvider` also calls `Shutdown()`, which will emit a warning if the `MeterProvider` was `Shutdown()` already manually [2]. So effectively every manual call to `MeterProvider` will create a warning message.

Ideally, if the meter provider was already shut down manually, the destructor would not emit a warning, I think.

[1] https://github.com/open-telemetry/opentelemetry-cpp/blob/0994256b75125f4cb5342a7ce61cf6cbbf405d5e/sdk/src/metrics/meter_provider.cc#L166
[2] https://github.com/open-telemetry/opentelemetry-cpp/blob/0994256b75125f4cb5342a7ce61cf6cbbf405d5e/sdk/src/metrics/meter_context.cc#L178

jaehyeong-cho:
Fixes #3511

## Changes

* Add test ensuring duplicate MeterProvider shutdown emits no log
* Modify MeterProvider to call context shutdown only once

It seems natural for `MeterProvider` to shut its context down from the destructor, and since `MeterContext` is a shared object, I think it is meaningful to warn about duplicate `Shutdown` calls.
However, I think it may not be a good experience that any user who calls the public `MeterProvider::Shutdown` API will always receive a warning log afterwards, because this isn't really a misuse case.
Therefore, with this PR, I would like to propose limiting `MeterProvider` to invoking the context shutdown at most once, so that no warning is emitted in the normal usage path.


lalitb (approved):
thanks.

sdk/src/metrics/meter_provider.cc
```diff
@@ -144,6 +144,13 @@ void MeterProvider::SetExemplarFilter(metrics::ExemplarFilterType exemplar_filte
  */
 bool MeterProvider::Shutdown(std::chrono::microseconds timeout) noexcept
 {
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  // Shutdown only once
+  if (shutdown_latch_.test_and_set(std::memory_order_acquire))
```
ThomsonTan, line 148: Wondering why is this only fixed for ABI v2? The same issue applies to v1 ABI and the same fix could be applied there too.

jaehyeong-cho: Thanks for the review. I thought adding a member variable to a public class is ABI-breaking, so I tried to follow [abi-version-policy.md](../blob/main/docs/abi-version-policy.md) and limited the change to ABI v2 only. If we saw this as a bug, I would try applying it to ABI v1 as well, but I thought it is not that serious.

lalitb: I think this can be fixed for ABI v1 too. The ABI guarantee in the policy is for the OpenTelemetry API interfaces; SDK classes are not guaranteed ABI-stable in the same way.

jaehyeong-cho: Thanks for clearly explaining the policy. Then I'll remove the ABI guard.

jaehyeong-cho: Addressed in c16f810.

PR #4101, opened 2026-05-24
[EXPORTER ETW] Generate a fresh trace id for every root span

### LINKED ISSUE #3846: Issue: ETW Exporter Tracer Uses a Single Trace ID for All Spans

lukeina2z:
Fixes #3846.

etw::Tracer was caching a single TraceId in a member (traceId_) generated once in its constructor, and using that same id as the fallback trace id for every root span it created. So if you did something as ordinary as:

    auto tracer  = provider->GetTracer("Geneva-Tracer-Foo");
    auto spanFoo = tracer->StartSpan("Span-Foo");   // root
    auto spanBar = tracer->StartSpan("Span-Bar");   // also root

both Foo and Bar would come out the back end stamped with the same env_dt_traceId, and the Geneva/Jarvis side would collapse two unrelated requests into one logical trace. The issue has a screenshot of exactly that: Span-Foo and Span-Bar emitted with the same trace id of 8507dcdc651af34fb8f21152a98e2977.

The OTel spec says each root span starts a new trace, and that is also what sdk::trace::Tracer does -- when there is no valid parent it calls generator.GenerateTraceId() on the spot, no cache anywhere (see sdk/src/trace/tracer.cc, the "if (parent_context.IsValid()) ... else trace_id = generator.GenerateTraceId()" block). The ETW exporter was the odd one out here.

Fix is one line in StartSpan(): when parentContext is invalid, call GetIdGenerator(tracerProvider_).GenerateTraceId() instead of returning the cached traceId_. This is the same call shape already used on the next line to mint the SpanId, so the access path is proven and thread-safe.

A nice side benefit: trace-id-ratio samplers were getting the same trace id for every root span on a Tracer and therefore making the same accept/reject decision for all of them, which defeated the point of probabilistic sampling. With this change the sampler sees a fresh id per root span and behaves the way the SDK does.

I deliberately did not remove the now-unused traceId_ member, its init in the ctor, or the public trace_id() accessor on Tracer. The accessor is part of the exporter's public surface and removing it would be a source break for anyone calling tracer.trace_id(). Happy to follow up with a cleanup PR if maintainers prefer.

Fixes # (issue)

## Changes


lalitb:
Thanks for the PR. The intention here is right: different root spans should not all get the same trace id.

Just to clarify, the current ETW behavior was intentional for performance reasons. ETW is a high-throughput exporter and does not always follow the regular SDK behavior when that adds cost to the hot path.

With this change, every root `StartSpan()` will generate a new trace id, which means an extra GUID (or custom ID) generation during span creation. We need to confirm that cost is acceptable for ETW scenarios for existing users, or make this optional so existing users are not affected by default?

ThomsonTan:
Add an entry to CHANGELOG as it is user-facing.

exporters/etw/include/opentelemetry/exporters/etw/etw_tracer.h
```diff
@@ -483,7 +483,16 @@ class Tracer : public opentelemetry::trace::Tracer,
         parentContext = spanContext;
       }
     }
-    auto traceId = parentContext.IsValid() ? parentContext.trace_id() : traceId_;
+    // Generate a brand-new trace id from the IdGenerator on every root
+    // span. Do NOT fall back to the Tracer-scoped cached traceId_,
+    // because that collapses every root span produced by this Tracer
+    // into a single logical trace -- breaking per-request correlation
+    // in Geneva/Jarvis and disagreeing with the OTel spec and the
```
ThomsonTan, line 490: Remove the mention of `Jarvis` as it is not public available?

dbarker:
@lukeina2z Are you still interested in contributing this change? Please see feedback from Tom and Lalit above.

PR #1606, opened 2022-09-09
[WIP] Add OTLP Metric Exporter Factory and `OtlpGrpcClient`

owent:
Signed-off-by: owent <admin@owent.net>

Fixes #1603 
Fixes #1605 

## Changes

+ Add `OtlpGrpcMetricExporterFactory` and `OtlpHttpMetricExporterFactory`.
+ Fix compatibility of OTLP log unit tests for MSVC.
+ Add `OtlpGrpcClient`

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [x] Changes in public API reviewed

PR #4104, opened 2026-05-25
[EXPORTER] Handle OTLP partial success response

### LINKED ISSUE #1577: Handle partial success responses from OTLP export services
### Problem Statement

Following the [specification change to add partial success responses](https://github.com/open-telemetry/opentelemetry-specification/pull/2696), each OTel SDK is encouraged to handle the resulting error message string in an appropriate way, considering existing norms. This requires OTLP v0.19.

### Proposed Solution

For each of the `Export*ServiceRequest` methods used by OTLP Trace and Metrics exporters, construct an error and call the appropriate handler with the error message string and the number of spans/points(/logs) dropped. Future OTel specifications may call for optional treatment of the number of dropped items, but presently that is just additional information to include in the handled error.

RaviTriv:
Fixes #1577

## Changes
PR to [handle partial success responses from OTLP collector](https://github.com/open-telemetry/opentelemetry-proto/blob/main/docs/specification.md#partial-success). As per the spec it states explicitly to not retry, the proposed solution is to log the rejection count and message.


exporters/otlp/test/otlp_grpc_exporter_test.cc
```diff
@@ -117,6 +122,71 @@ class OtlpMockTraceServiceStub : public proto::collector::trace::v1::MockTraceSe
...
+  ScopedTestLogHandler(ScopedTestLogHandler &&)                 = delete;
+  ScopedTestLogHandler &operator=(ScopedTestLogHandler &&)      = delete;
+
+  std::vector<Entry> Drain() { return handler_->Drain(); }
+
+private:
+  LogHandlerImpl *handler_{nullptr};
+  opentelemetry::nostd::shared_ptr<sdk::common::internal_log::LogHandler> previous_handler_;
+  sdk::common::internal_log::LogLevel previous_level_;
+};
```
RaviTriv: Wondering if its appropriate to move this to `test_common/`?

I based this off a nice implementation in `sdk/test/metrics/meter_provider_sdk_test.cc` and kept names fully scoped. So at the moment about 7 copies of this.

exporters/otlp/src/otlp_grpc_exporter.cc
```diff
@@ -174,15 +183,22 @@ sdk::common::ExportResult OtlpGrpcExporter::Export(
     proto::collector::trace::v1::ExportTraceServiceResponse *response =
         google::protobuf::Arena::Create<proto::collector::trace::v1::ExportTraceServiceResponse>(
             arena.get());
-    grpc::Status status = OtlpGrpcClient::DelegateExport(
-        trace_service_stub_.get(), std::move(context), std::move(arena), request, response);
+    grpc::Status status = trace_service_stub_->Export(context.get(), *request, response);
```
owent: We should only call gRPC's APIs only in otlp_grpc_client.cc to avoid symbol conflicts.

> More details can be found at #3435 #1891 #1520 #1603 #1606

exporters/otlp/src/otlp_grpc_log_record_exporter.cc
```diff
@@ -175,15 +185,23 @@ opentelemetry::sdk::common::ExportResult OtlpGrpcLogRecordExporter::Export(
     proto::collector::logs::v1::ExportLogsServiceResponse *response =
         google::protobuf::Arena::Create<proto::collector::logs::v1::ExportLogsServiceResponse>(
             arena.get());
-    grpc::Status status = OtlpGrpcClient::DelegateExport(
-        log_service_stub_.get(), std::move(context), std::move(arena), request, response);
+    grpc::Status status = log_service_stub_->Export(context.get(), *request, response);
```
owent: The same as otlp_grpc_exporter.cc‎:186

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -748,7 +763,16 @@ sdk::common::ExportResult OtlpHttpClient::Export(
     std::function<bool(opentelemetry::sdk::common::ExportResult)> &&result_callback,
     std::size_t max_running_requests) noexcept
 {
-  auto session = createSession(message, std::move(result_callback));
+  return Export(message, nullptr, std::move(result_callback), max_running_requests);
+}
+
+sdk::common::ExportResult OtlpHttpClient::Export(
+    const google::protobuf::Message &message,
+    google::protobuf::Message *response,
```
owent, line 768: I noticed the callback currently captures shared_ptr of response but the method signature also passes pointer to response separately. It might be cleaner to package these together—perhaps as a CompletionContext—so the lifecycle is managed in one place.

And we should use arena to create response to reduce memory segment.

exporters/otlp/src/otlp_http_metric_exporter.cc
```diff
@@ -183,34 +180,40 @@ opentelemetry::sdk::common::ExportResult OtlpHttpMetricExporter::Export(
           &arena);
   OtlpMetricUtils::PopulateRequest(data, service_request);
   std::size_t metric_count = data.scope_metric_data_.size();
-#ifdef ENABLE_ASYNC_EXPORT
-  http_client_->Export(*service_request, [metric_count](
-                                             opentelemetry::sdk::common::ExportResult result) {
+
+  std::shared_ptr<proto::collector::metrics::v1::ExportMetricsServiceResponse> response =
```
owent: We should use arena to create response to reduce memory segment. Maybe we can move arena into session to hold the lifetime, just like it in otlp_grpc_client

exporters/otlp/src/otlp_grpc_exporter.cc
RaviTriv: Thanks for taking a look!

Got it. Updated `DelegateExport` to take an optional callback that runs before the arena destructs, and passed a callback that logs partial_success / success.

exporters/otlp/src/otlp_http_client.cc
RaviTriv: Good idea, updated to move this into an arena in HttpSessionData.

exporters/otlp/test/otlp_grpc_exporter_test.cc
RaviTriv: Moved this to into `test_common/include/opentelemetry/test_common/sdk/common/scoped_test_log_handler.h`

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -332,7 +346,7 @@ class ResponseHandler : public http_client::EventHandler
 
       if (result_callback_)
       {
-        result_callback_(result);
+        result_callback_(result, response_);
```
lalitb: I think this ordering is risky. `response_` is owned by the arena stored in the session data, but `ReleaseSession()` moves that session data to the cleanup list before the callback reads `response_`. In async HTTP export, another thread could clean up the session and destroy the arena before this callback uses the response. Can we run the callback while the session still owns the arena, or otherwise keep the arena alive through the callback?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -123,6 +128,15 @@ class ResponseHandler : public http_client::EventHandler
       }
     }
 
+    // On 2xx, parse the body into the caller-provided typed response
+    if (response_ != nullptr && result == sdk::common::ExportResult::kSuccess)
+    {
+      if (!response_->ParseFromString(body_))
```
lalitb: This only handles binary protobuf responses. For OTLP/HTTP JSON mode, the server should send the response as JSON too, so `ParseFromString()` will fail and the partial_success response will be missed. Can we parse based on `options_.content_type`, and add a JSON response test as well?

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -123,6 +135,22 @@ class ResponseHandler : public http_client::EventHandler
       }
     }
 
+    // On 2xx, parse the body into the caller-provided typed response
+    if (response_ != nullptr && result == sdk::common::ExportResult::kSuccess)
+    {
+      if (content_type_ == HttpRequestContentType::kJson)
+      {
+        if (!google::protobuf::util::JsonStringToMessage(body_, response_).ok())
```
lalitb, line 146: Can we avoid treating this as a success if parsing the response body fails? Right now we only log the parse error, so the exporter may still continue with an empty response and miss `partial_success`.

exporters/otlp/src/otlp_http_client.cc
RaviTriv: Good catch, reordered so that the callback runs before the `ReleaseSession()` to ensure that the `response_` is valid when the callback runs.

exporters/otlp/src/otlp_http_client.cc
RaviTriv: Thanks for pointing this out! Updated to parse based on the the `content_type` and added corresponding tests

exporters/otlp/src/otlp_http_client.cc
RaviTriv: Yep, updated to set the result as `kFailure` on parse fail

exporters/otlp/test/otlp_http_metric_exporter_test.cc
```diff
@@ -21,9 +22,10 @@
 #include "opentelemetry/exporters/otlp/otlp_metric_utils.h"
 #include "opentelemetry/exporters/otlp/otlp_preferred_temporality.h"
 #include "opentelemetry/ext/http/client/http_client.h"
+#include "opentelemetry/nostd/shared_ptr.h"
 #include "opentelemetry/nostd/string_view.h"
-#include "opentelemetry/nostd/unique_ptr.h"
 #include "opentelemetry/sdk/common/exporter_utils.h"
+#include "opentelemetry/sdk/common/global_log_handler.h"
 #include "opentelemetry/sdk/common/thread_instrumentation.h"
```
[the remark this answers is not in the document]
RaviTriv: Removed unused `opentelemetry/nostd/shared_ptr.h`, need `opentelemetry/sdk/common/global_log_handler.h` for IWYU.

Edit: Included `memory` instead of  `opentelemetry/nostd/shared_ptr.h` as per IWYU

exporters/otlp/src/otlp_http_client.cc
lalitb, line 826: Can we return the final callback result here instead of always returning success? `OnResponse()` now changes the result to `kFailure` when parsing the response body fails, but this path still returns `kSuccess` as long as the wait completed. That means sync HTTP export can log a parse failure but still report success to the caller

RaviTriv: good catch! definitely, updated the sync path to return `kFailure` if parse fails.

exporters/otlp/src/otlp_http_exporter.cc
```diff
@@ -168,43 +173,53 @@ opentelemetry::sdk::common::ExportResult OtlpHttpExporter::Export(
   // When in batch mode, it's easy to export a large number of spans at once, we can alloc a lager
   // block to reduce memory fragments.
   arena_options.max_block_size = 65536;
-  google::protobuf::Arena arena{arena_options};
+  // Ownership transfers into HttpSessionData until the request completes
+  std::unique_ptr<google::protobuf::Arena> arena{new google::protobuf::Arena{arena_options}};
```
owent: We can use std::make_unique_ptr here.

exporters/otlp/src/otlp_http_log_record_exporter.cc
```diff
@@ -171,39 +176,52 @@ opentelemetry::sdk::common::ExportResult OtlpHttpLogRecordExporter::Export(
   // When in batch mode, it's easy to export a large number of spans at once, we can alloc a lager
   // block to reduce memory fragments.
   arena_options.max_block_size = 65536;
-  google::protobuf::Arena arena{arena_options};
+  // Ownership transfers into HttpSessionData until the request completes
+  std::unique_ptr<google::protobuf::Arena> arena{new google::protobuf::Arena{arena_options}};
```
owent: The same as before, we can use std::make_unique_ptr here.

exporters/otlp/src/otlp_http_client.cc
owent: Do you think we can remove the old OtlpHttpClient::Export and OtlpHttpClient::createSession API which only accept ExportResult? I think it's more clear to only keep the new callback.

exporters/otlp/src/otlp_http_exporter.cc
RaviTriv: Good catch, updated here and the other usages to use `std::make_unique_ptr`

exporters/otlp/src/otlp_http_client.cc
RaviTriv: yep, updated to remove all the other overloads since they don't include the response.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_http_client.h
```diff
@@ -173,35 +174,21 @@ class OtlpHttpClient
...
-  sdk::common::ExportResult Export(
-      const google::protobuf::Message &message,
-      std::function<bool(opentelemetry::sdk::common::ExportResult)> &&result_callback) noexcept;
-
-  /**
-   * Async export
+   * Async export with typed response
    * @param message message to export, it should be ExportTraceServiceRequest,
    * ExportMetricsServiceRequest or ExportLogsServiceRequest
+   * @param arena protobuf arena that owns response
```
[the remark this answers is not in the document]
RaviTriv: At the moment not applying any changes here. Discussing with owent in [thread](https://github.com/open-telemetry/opentelemetry-cpp/pull/4104#discussion_r3377751538) above.
- Not a public API that is being consumed by external users, so very unlikely to break anyone.
- Updated the doc to make sync and async paths more clear.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -715,47 +758,31 @@ OtlpHttpClient::OtlpHttpClient(OtlpHttpClientOptions &&options,
...
+  auto result_capturing_callback = [session_result, cb = std::move(result_callback)](
+                                       sdk::common::ExportResult result,
+                                       google::protobuf::Message *resp) {
+    *session_result = result;
+    return cb(result, resp);
+  };
+
+  auto session =
+      createSession(message, std::move(arena), response, std::move(result_capturing_callback));
+  if (auto *result = opentelemetry::nostd::get_if<sdk::common::ExportResult>(&session))
```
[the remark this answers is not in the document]
RaviTriv: Updated

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -124,6 +138,24 @@ class ResponseHandler : public http_client::EventHandler
       }
     }
 
+    // On 2xx with a non-empty body, parse it into the caller-provided typed response
+    if (response_ != nullptr && result == sdk::common::ExportResult::kSuccess && !body_.empty())
+    {
+      if (content_type_ == HttpRequestContentType::kJson)
+      {
```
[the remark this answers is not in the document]
RaviTriv: Updated to log after successfully parsing response body

owent (approved):
LGTM. thanks.

marcalff:
LGTM, but the real expert in this area is @owent , trusting his review.

@lalitb Since you looked at this PR already, do you have any remaining comments ?

Ok to merge ?

PR #4132, opened 2026-06-06
feat: log record attribute value length limit & attribute count limit

### LINKED ISSUE #4126: Implement LogRecord limits
**Implement LogRecord limits**

See https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits
The limits include: 
- `attribute_count_limit` (default 128)
- `attribute_value_length_limit` (default unlimited)

The limits are optional but desired to limit resource usage in constrained environments. 

**Current state**

Log record attribute limits are not fully supported yet. They are included in the yaml configuration (see [`LogRecordLimitsConfiguration`](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/configuration/log_record_limits_configuration.h)), but never passed to the LoggerProvider or LogRecord/Recordables. 

**Proposed changes**

1. Add a `LogRecordLimits` struct to `LoggerProvider` with spec defaults. This may be passed to recordables and used by const pointer/ref. 
2. Enforce limits in the recordable (ReadWriteLogRecord and OtlpLogRecordable) SetAttribute methods.
3. Set the `dropped_attributes_count` in the  OTLP message on export. 
   ([spec field](https://github.com/open-telemetry/opentelemetry-proto/blob/82dd303497e1bfbc86f1f8d584696195a02ea085/opentelemetry/proto/logs/v1/logs.proto#L179)).

This change should be profiled against the current state (no log record limits in the recording path).

proost:
Fix: #4126 

## Changes

Adding new feature `OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT` following spec.

I referred below:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#logrecord-limits
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#logrecord-limits
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-limits

Following spec, I changed SDK only.

I plan to apply decorator way. Before setter called, I tried to enforce this feature to log record. Problem is this is inevitable to current SDK. If have any suggestions, please give me some feedback.

value length limit benchmark:

```
-------------------------------------------------------------------------------------------------------
Benchmark                                             Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------------
BM_SetStringAttributeNoLimit                        129 ns          125 ns      5517767 bytes_per_second=15.2404Gi/s
BM_SetStringAttributeLimitNotHit                    123 ns          122 ns      5487501 bytes_per_second=15.6172Gi/s
BM_SetStringAttributeSmallLimitHit                 64.3 ns         64.3 ns     10916371 bytes_per_second=29.6844Gi/s
BM_SetStringAttributeLargeLimitHit                 93.3 ns         93.3 ns      7411721 bytes_per_second=20.4478Gi/s
BM_SetStringArrayAttributeShortArrayLimitHit        130 ns          130 ns      5384464 items_per_second=30.7945M/s
BM_SetStringArrayAttributeLongArrayLimitHit         909 ns          909 ns       774217 items_per_second=70.4199M/s
```

count limit benchmark:

baseline:
``` 
-------------------------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------
BM_SetDistinctAttributesUnderLimit/8          439 ns          439 ns      1592935 items_per_second=18.2267M/s
BM_SetDistinctAttributesUnderLimit/16         909 ns          909 ns       759657 items_per_second=17.6055M/s
BM_SetDistinctAttributesUnderLimit/32        2483 ns         2483 ns       266050 items_per_second=12.8874M/s
BM_SetDistinctAttributesUnderLimit/64        5378 ns         5378 ns       130136 items_per_second=11.9003M/s
BM_SetDistinctAttributesUnderLimit/128      11627 ns        11626 ns        60203 items_per_second=11.0098M/s
```

comparision:
``` 
BM_SetDistinctAttributesUnderLimit/8          761 ns          761 ns       917233 items_per_second=10.519M/s
BM_SetDistinctAttributesUnderLimit/16        1721 ns         1721 ns       413115 items_per_second=9.29961M/s
BM_SetDistinctAttributesUnderLimit/32        4365 ns         4364 ns       161415 items_per_second=7.33292M/s
BM_SetDistinctAttributesUnderLimit/64        9522 ns         9519 ns        73502 items_per_second=6.72351M/s
BM_SetDistinctAttributesUnderLimit/128      28387 ns        28386 ns        24260 items_per_second=4.50921M/s
BM_SetDistinctAttributesOverLimit/8           828 ns          828 ns       882998 items_per_second=9.65847M/s
BM_SetDistinctAttributesOverLimit/16          908 ns          908 ns       755289 items_per_second=17.6222M/s
BM_SetDistinctAttributesOverLimit/32         1548 ns         1543 ns       505904 items_per_second=20.7436M/s
BM_SetDistinctAttributesOverLimit/64         2292 ns         2292 ns       304527 items_per_second=27.9197M/s
BM_SetDistinctAttributesOverLimit/128        4194 ns         4193 ns       166171 items_per_second=30.5246M/s
```


.github/workflows/clang-tidy.yaml
```diff
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 346
+            warning_limit: 347
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 350
+            warning_limit: 351
```
owent: Could you please resolve these warnings but not just ignore them?
And could you please also implement maximum allowed log record attribute count in <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#logrecord-limits> ?

proost: Because of SetAttribute. SetAttribute itself is noexcept. problem is It allocate memory. so bad_alloc can happen. current otel doesn't standardize how to handle it. Is ok to swallow it?

lalitb: I agree with @owent - we should not just increase the warning limit. `SetAttribute()` is `noexcept`, so if an allocation throws here, it would terminate the process.
 If allocation fails, I think it is better to drop this attribute/value than crash. I would also avoid logging from this catch block, because the internal log path formats strings and may allocate again.

sdk/include/opentelemetry/sdk/logs/log_record_limits.h
```diff
@@ -0,0 +1,36 @@
...
+
+}  // namespace log_record_limits_env
+
+/**
+ * Configuration for limits applied to LogRecord attributes.
+ */
+struct OPENTELEMETRY_EXPORT LogRecordLimits
+{
+  std::size_t attribute_value_length_limit =
+      log_record_limits_env::GetAttributeValueLengthLimitFromEnv();
```
ThomsonTan: `LogRecordLimits::attribute_value_length_limit` has a default member initializer that calls `GetAttributeValueLengthLimitFromEnv()`, and `Recordable` now holds a default-constructed `LogRecordLimits limits_`. So every `MakeRecordable()` (one per emitted record, plus one per processor via `MultiRecordable`) performs a `getenv` + string parse even when the feature is unused, and the value is immediately overwritten by `Logger::CreateLogRecord()` via `SetLogRecordLimits()`. `LoggerConfig::Default()`/`Disabled()` also lost their static caching, so every `GetLogger()` re-reads the env too.

Suggest: default the field to `(std::numeric_limits<std::size_t>::max)()` (no env read), and read the env once in `LoggerConfig::Default()` (or at provider construction). That keeps the env-driven behavior while removing the per-record syscall (which takes a global lock on some CRTs).

sdk/src/configuration/sdk_builder.cc
```diff
@@ -191,6 +194,41 @@ using common::WildcardMatch;
...
+  }
+
+  return limits;
+}
+
+opentelemetry::sdk::logs::LoggerConfig CreateLoggerConfig(
+    const LoggerConfigConfiguration &model,
+    const opentelemetry::sdk::logs::LogRecordLimits &log_record_limits)
+{
+  return opentelemetry::sdk::logs::LoggerConfig::Create(
```
lalitb, line 272: This seems to ignore some parsed logger config values. The parser reads `minimum_severity` and `trace_based`, but here we always pass `Severity::kInvalid` and `false`. Is it intended?

sdk/include/opentelemetry/sdk/logs/recordable.h
```diff
@@ -43,6 +74,119 @@ class Recordable : public opentelemetry::logs::LogRecord
...
+        return value;
+      }
+
+      const auto value_length = std::strlen(value);
+      if (value_length <= limit_)
+      {
+        return value;
+      }
+
+      return opentelemetry::nostd::string_view(value, limit_);
```
ThomsonTan: Truncation uses a raw byte count (`string_view(value.data(), limit_)`), which can split a multi-byte UTF-8 sequence and produce invalid UTF-8. OTLP serializes attributes into protobuf `string` fields, which require valid UTF-8; strict collectors/parsers may reject or warn.

sdk/src/configuration/sdk_builder.cc
proost: Sorry, I worked on stale branch. I will update

sdk/include/opentelemetry/sdk/logs/recordable.h
proost: Thank you i didn't considered utf-8 string. When i read spec, string value can be invalid utf-8 string. Who is resposibility to utf-8 validation? sdk or user?

owent: I think what @ThomsonTan means is: when users set their attributes as valid UTF-8 strings, we should not truncate them into invalid UTF-8 strings. For OTLP, invalid UTF-8 strings will be exported as bytes, which changes the field's type and name.

.github/workflows/clang-tidy.yaml
proost: I dropped attribute: [9ce93b3](https://github.com/open-telemetry/opentelemetry-cpp/pull/4132/commits/9ce93b3a7da67b8b7294317dc545d6e4c0484944)

And also adding count limit in this PR too: [81746d5](https://github.com/open-telemetry/opentelemetry-cpp/pull/4132/commits/81746d540a07f014aee74e94a950d86cfe388749)

sdk/include/opentelemetry/sdk/logs/recordable.h
proost: [0b38823](https://github.com/open-telemetry/opentelemetry-cpp/pull/4132/commits/0b38823155cdc6417854f15616175a964ed278df)

Thank you, I covered utf8 string

.github/workflows/clang-tidy.yaml
proost: When i read spec, duplication of attribute may allowed. Do i have to include the feature in this PR?

sdk/include/opentelemetry/sdk/logs/recordable.h
```diff
@@ -179,14 +227,42 @@ class Recordable : public opentelemetry::logs::LogRecord
...
       }
 
-      return opentelemetry::nostd::string_view(value.data(), limit_);
+      return opentelemetry::nostd::string_view(value.data(), Utf8PrefixLength(value, limit_));
+    }
+
+    // Byte length of the longest prefix of `value` holding at most `limit` UTF-8 code points,
+    // never splitting a multi-byte sequence. Malformed lead/continuation bytes are counted as
+    // one byte each (we do not validate or repair user-supplied invalid UTF-8).
+    static std::size_t Utf8PrefixLength(opentelemetry::nostd::string_view value,
```
owent, line 239: In my understanding, if the original input is not a valid UTF-8 string (raw bytes set by user), there is no need to truncate it as UTF-8 string.
What do you think about it?

sdk/include/opentelemetry/sdk/logs/recordable.h
```diff
@@ -29,7 +52,60 @@ namespace logs
...
+  void SetLogRecordLimits(const LogRecordLimits &limits) noexcept { limits_ = limits; }
+
+  void SetAttribute(opentelemetry::nostd::string_view key,
+                    const opentelemetry::common::AttributeValue &value) noexcept final
+  {
+#if OPENTELEMETRY_HAVE_EXCEPTIONS
+    try
+    {
+#endif
+      if (attribute_keys_.size() < limits_.attribute_count_limit)
```
lalitb, line 67: I am worried about putting this limit/count tracking in the base `Recordable` hot path. With the default count limit, every log record now maintains attribute count state, and every attribute copies the key into a `std::string` and stores/checks it in an `unordered_set` before the exporter recordable gets it.

Since this is in `Recordable::SetAttribute()`, all exporters pay this cost, not only `ReadWriteLogRecord` or OTLP. I am okay with adding this where it is needed, but putting it in the base `Recordable` means every exporter pays extra heap allocation and key-tracking cost even if that exporter already has its own storage model or can enforce the limit more cheaply.

Can we move this out of the base `Recordable` path and let each recordable/exporter enforce the limit in the way that fits its representation?

lalitb (changes requested):
As mentioned [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4132#discussion_r3409765486).

sdk/include/opentelemetry/sdk/logs/recordable.h
proost: [3d68878](https://github.com/open-telemetry/opentelemetry-cpp/pull/4132/commits/3d688781b3e974edc3624c946c4244a3ca14c21c)

Agree. 

But if you mean more stricter check, I think we need third party like simdutf or utfcpp to check valid utf8 string. Do you think we need strict validation?
Why i didn't is spec discourage to include third party library.

sdk/include/opentelemetry/sdk/logs/recordable.h
proost: Thanks for review. 
I designed like this due to enforcement. 
I agree your concern. I redesign spec-compatible and more performance aware way And i request review again, when i finish.

proost:
@thc1006 will continue the feature in here:
https://github.com/open-telemetry/opentelemetry-cpp/pull/4157

Thanks for all the feedbacks!

Issue #4028, opened 2026-04-22
[SDK] Implement ComposableSampler


marcalff:
ComposableSampler are a recent addition in the specifications.

Implement support in the SDK.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#compositesampler

mateenali66:
interested in picking this up, quick scoping question first.

#3966 adds the config-side DTOs under `sdk/configuration/` and wires `sdk_builder.cc` to instantiate composable variants. reading #4028 as the trace-side piece: actual sampler classes under `sdk/src/trace/samplers/` (ComposableAlwaysOn/Off, ComposableParentThreshold, ComposableProbability, ComposableRuleBased) plus the composition semantics from the spec. does that split match what you had in mind?

also: one PR covering all variants + composition, or split per variant? whichever is easier to review.

marcalff:
@mateenali66 

> interested in picking this up, quick scoping question first.
> 

Thanks for volunteering.

> [#3966](https://github.com/open-telemetry/opentelemetry-cpp/pull/3966) adds the config-side DTOs under `sdk/configuration/` and wires `sdk_builder.cc` to instantiate composable variants. reading [#4028](https://github.com/open-telemetry/opentelemetry-cpp/issues/4028) as the trace-side piece: actual sampler classes under `sdk/src/trace/samplers/` (ComposableAlwaysOn/Off, ComposableParentThreshold, ComposableProbability, ComposableRuleBased) plus the composition semantics from the spec. does that split match what you had in mind?
> 

Yes.

#3966 is now merged, so the next part is to implement the SDK logic itself, under `sdk/src/trace/samplers/`.

> also: one PR covering all variants + composition, or split per variant? whichever is easier to review.

One PR is more practical.

Expect slow/no replies next week, but I will be available for review/guidance after that.

mateenali66:
starting to sketch this. before committing to the rest of the variants and tests, wanted to float the interface shapes so you can redline anything early.

### proposed types

```cpp
struct SamplingIntent
{
  uint64_t threshold_value = 0;   // 56-bit value in low bits; R >= T keeps
  bool has_threshold       = false;  // false = drop unconditionally
  bool threshold_reliable  = false;  // true = write "ot"/"th:" back

  AttributesProvider  attributes_provider{};   // invoked only on keep
  TraceStateProvider  trace_state_provider{};  // applied before CompositeSampler writes "th:"
};

[...]

PR #3966, opened 2026-04-03
[CONFIGURATION] File configuration - composable samplers

DCchoudhury15:
Fixes #3914

## Changes

This PR implements the necessary infrastructure and test coverage to fully support composable samplers. Specifically, the changes include:

* **Parser Logic:** Implemented the YAML parsing logic to properly extract composable sampler configurations.
* **SDK Builder:** Added SDK builder visitor instantiation to correctly wire the new sampler variants.
* **Test Safety:** Replaced dangerous casts with the safe Visitor pattern in the testing suite.
* **Test Coverage:** Expanded unit test coverage to ensure all composable sampler variants are accurately parsed and instantiated.

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4133, opened 2026-06-06
[SDK] Add ComposableSampler interface and CompositeSampler

mateenali66:
implements the trace-side consistent probability sampling logic for #4028,
under sdk/src/trace/samplers/. #3966 landed the config-side DTOs; this is the
SDK logic they map to.

what's here:
- `ComposableSampler` interface + `SamplingIntent` (56-bit threshold,
  threshold_reliable, optional attributes/tracestate providers)
- `CompositeSampler` adapts a ComposableSampler into a `Sampler`: parses the
  incoming "ot" tracestate, derives the 56-bit randomness from the rv sub-key
  or the low 56 bits of the trace id, applies the intent, and writes the
  effective threshold back to the th sub-key (preserving explicit rv)
- variants: `ComposableAlwaysOn`, `ComposableAlwaysOff`,
  `ComposableTraceIdRatioBased` (spec's ComposableProbability),
  `ComposableParentThreshold`, `ComposableRuleBased` (+ a `Predicate` interface)
- `CompositeSamplerFactory`, and a private `ot_trace_state` encoder/decoder
- unit tests for every variant, the th/rv encoding, parent threshold
  propagation and rule evaluation (11 tests)

the threshold math and tracestate codec follow the consistent probability
sampling spec and the java contrib consistent56 reference: keep when R >= T,
T = (1 - p) * 2^56, th encoded MSB-first with trailing zero nibbles trimmed,
rv as 14 hex digits. ComposableTraceIdRatioBased deliberately does not reuse
the existing TraceIdRatioBasedSampler, whose 64-bit acceptance-threshold math
is not consistent-probability compatible.

builds and passes locally via cmake/ninja; the existing always_on/off, parent
and trace_id_ratio sampler tests still pass.

not in this PR: wiring the file-config builder (sdk_builder.cc) to these
classes. that needs a recursive ComposableSampler builder plus predicate
construction from the rule attribute/pattern config, so it is cleaner as a
focused follow-up. the FIXME-SDK markers in sdk_builder.cc stay until then.

interface notes for @marcalff (happy to adjust): threshold as uint64_t + bool
(avoids <optional> for ABI); private ot codec under samplers/; shared_ptr
delegates so samplers compose; an abstract Predicate base class rather than
std::function for rule predicates (matches the rest of the SDK and avoids
heap-allocated callbacks on the hot path).

Part of #4028


marcalff:
> not in this PR: wiring the file-config builder (sdk_builder.cc) to these classes. that needs a recursive ComposableSampler builder plus predicate construction from the rule attribute/pattern config, so it is cleaner as a focused follow-up. the FIXME-SDK markers in sdk_builder.cc stay until then.

Yes, makes sense.

marcalff (approved):
LGTM.

This is a stand alone building block, next step is to use it from sdk_builder.

Taking the PR as is.

Thanks for the contribution, this is very good work.

PR #4135, opened 2026-06-07
[SDK] Implement the ProbabilitySampler

### LINKED ISSUE #4127: Implement Probability Sampler
This issue is to complete the Probability Sampler implementation and related tests. 
See https://opentelemetry.io/docs/specs/otel/trace/sdk/#probabilitysampler

The Probability Sampler will replace the soon deprecated TraceIdRatio sampler and can be implemented now that the Trace Context Level 2 is merged (https://github.com/open-telemetry/opentelemetry-cpp/pull/4012).

ayush-that:
Fixes #4127

ref: https://opentelemetry.io/docs/specs/otel/trace/sdk/#probabilitysampler

## Changes

adds the ProbabilitySampler from the spec. it turns the configured ratio into a 56 bit threshold, takes the randomness value from the rv key in the ot tracestate if there is one, otherwise from the last 7 bytes of the trace id, and samples the span when the randomness is greater than or equal to the threshold. sampled spans get the threshold written into the tracestate as the th key. the file configuration for the composable probability sampler now uses this sampler instead of TraceIdRatioBasedSampler.

the threshold keeps full 56 bit precision (same as opentelemetry-java-contrib). composable samplers are a separate effort (#4028), and the deprecated TraceIdRatioBased sampler is not changed.


ayush-that:
here is a demo against this branch to show the behaviour end to end. it shows the threshold boundary at R >= T, the th encoding, rv taking precedence over the trace id, th replacement, and a foreign tracestate entry surviving untouched.

output:

```
[==========] 14 tests from 1 test suite ran. (922 ms total)
[  PASSED  ] 14 tests.
ratio 1.0    -> SAMPLE, tracestate: ot=th:0
ratio 0.5, R=0x80000000000000 -> SAMPLE, tracestate: ot=th:8
ratio 0.5, R=0x7fffffffffffff -> DROP
ratio 0.5, 100000 random trace ids -> sampled 49958 (0.500)
ratio 0.5, rv=ff.. trace id low bits 0 -> SAMPLE, tracestate: ot=rv:ffffffffffffff;th:8
ratio 0.25, parent 'ot=th:8,congo=..' -> SAMPLE, tracestate: ot=th:c,congo=t61rcWkgMzE
```

<details>
<summary>demo.cc</summary>

```cpp
#include <cstdint>
#include <cstdio>
#include <map>
#include <string>
#include <utility>
#include <vector>

#include "opentelemetry/common/key_value_iterable_view.h"
#include "opentelemetry/sdk/trace/samplers/probability.h"
#include "opentelemetry/trace/span_context.h"
#include "opentelemetry/trace/span_context_kv_iterable_view.h"
#include "opentelemetry/trace/trace_state.h"
#include "src/common/random.h"

namespace trace_api = opentelemetry::trace;
using opentelemetry::sdk::common::Random;
using opentelemetry::sdk::trace::Decision;
using opentelemetry::sdk::trace::ProbabilitySampler;

namespace {
trace_api::TraceId TraceIdWithLow56(uint64_t v) {
  uint8_t buf[16] = {0};
  for (int i = 0; i < 7; ++i)
    buf[15 - i] = static_cast<uint8_t>(v >> (8 * i));
  return trace_api::TraceId(buf);
}

opentelemetry::sdk::trace::SamplingResult
Sample(ProbabilitySampler &s, const trace_api::SpanContext &ctx,
       trace_api::TraceId tid) {
  std::map<std::string, int> attrs;
  std::vector<
      std::pair<trace_api::SpanContext, std::map<std::string, std::string>>>
      links;
  opentelemetry::common::KeyValueIterableView<decltype(attrs)> attrs_view{
      attrs};
  trace_api::SpanContextKeyValueIterableView<decltype(links)> links_view{links};
  return s.ShouldSample(ctx, tid, "demo", trace_api::SpanKind::kInternal,
                        attrs_view, links_view);
}

const char *Header(const opentelemetry::sdk::trace::SamplingResult &r) {
  static std::string h;
  h = r.trace_state ? r.trace_state->ToHeader() : "(parent tracestate kept)";
  return h.c_str();
}
} // namespace

int main() {
  // ratio 1.0: always sampled, threshold 0 emitted as ot=th:0
  ProbabilitySampler all(1.0);
  auto r =
      Sample(all, trace_api::SpanContext::GetInvalid(), TraceIdWithLow56(0));
  printf("ratio 1.0    -> %s, tracestate: %s\n",
         r.decision == Decision::RECORD_AND_SAMPLE ? "SAMPLE" : "DROP",
         Header(r));

  // ratio 0.5: threshold is 2^55. R = 2^55 samples, R = 2^55-1 drops.
  ProbabilitySampler half(0.5);
  r = Sample(half, trace_api::SpanContext::GetInvalid(),
             TraceIdWithLow56(0x80000000000000));
  printf("ratio 0.5, R=0x80000000000000 -> %s, tracestate: %s\n",
         r.decision == Decision::RECORD_AND_SAMPLE ? "SAMPLE" : "DROP",
         Header(r));
  r = Sample(half, trace_api::SpanContext::GetInvalid(),
             TraceIdWithLow56(0x7fffffffffffff));
  printf("ratio 0.5, R=0x7fffffffffffff -> %s\n",
         r.decision == Decision::RECORD_AND_SAMPLE ? "SAMPLE" : "DROP");

  // 100000 random trace ids at ratio 0.5
  int sampled = 0;
  for (int i = 0; i < 100000; ++i) {
    uint8_t buf[16];
    Random::GenerateRandomBuffer(buf);
    if (Sample(half, trace_api::SpanContext::GetInvalid(),
               trace_api::TraceId(buf))
            .decision == Decision::RECORD_AND_SAMPLE)
      ++sampled;
  }
  printf("ratio 0.5, 100000 random trace ids -> sampled %d (%.3f)\n", sampled,
         sampled / 100000.0);

  // explicit rv in the parent tracestate wins over trace id randomness
  uint8_t id[16] = {1};
  trace_api::SpanContext rv_high(
      trace_api::TraceId(id), trace_api::SpanId(), trace_api::TraceFlags{0},
      false, trace_api::TraceState::FromHeader("ot=rv:ffffffffffffff"));
  r = Sample(half, rv_high, TraceIdWithLow56(0));
  printf("ratio 0.5, rv=ff.. trace id low bits 0 -> %s, tracestate: %s\n",
         r.decision == Decision::RECORD_AND_SAMPLE ? "SAMPLE" : "DROP",
         Header(r));

  // existing th is replaced, other entries survive
  ProbabilitySampler quarter(0.25);
  trace_api::SpanContext with_th(
      trace_api::TraceId(id), trace_api::SpanId(), trace_api::TraceFlags{0},
      false, trace_api::TraceState::FromHeader("ot=th:8,congo=t61rcWkgMzE"));
  r = Sample(quarter, with_th, TraceIdWithLow56(0xffffffffffffff));
  printf("ratio 0.25, parent 'ot=th:8,congo=..' -> %s, tracestate: %s\n",
         r.decision == Decision::RECORD_AND_SAMPLE ? "SAMPLE" : "DROP",
         Header(r));

  return 0;
}
```

</details>

ayush-that:
about last force push-
1. fixed the iwyu failures.
2. rebased on latest main.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,228 @@
...
+{
+constexpr char kOtTraceStateKey[] = "ot";
+
+// Rejection threshold for a ratio of zero, 2^56: drops every span.
+constexpr uint64_t kMaxThreshold = 1ULL << 56;
+
+/**
+ * Converts a ratio in [0, 1] to a 56-bit rejection threshold.
+ */
+uint64_t CalculateRejectionThreshold(double ratio) noexcept
```
lalitb: Can we reuse [CalculateThreshold](https://github.com/open-telemetry/opentelemetry-cpp/blob/17c4723884add23479baabb2fa9a57a004b41b68/sdk/src/trace/samplers/ot_trace_state.h#L34), most of the logic seems same?

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,228 @@
...
+  for (int i = 13; i >= 0; --i)
+  {
+    hex[i] = kHex[threshold & 0xf];
+    threshold >>= 4;
+  }
+  hex.erase(hex.find_last_not_of('0') + 1);
+  return hex;
+}
+
+uint64_t RandomnessFromTraceId(const trace_api::TraceId &trace_id) noexcept
```
lalitb: Same as [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4135/changes#r3409715008).

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,228 @@
...
+        OTEL_INTERNAL_LOG_WARN(
+            "ProbabilitySampler presumes TraceID randomness, but the W3C random trace flag is "
+            "not set. Upgrade the caller to W3C Trace Context Level 2.");
+      }
+    }
+    randomness = RandomnessFromTraceId(trace_id);
+  }
+
+  if (randomness < threshold_)
+    return {Decision::DROP, nullptr, {}};
```
lalitb: If this sampler drops the span, it returns no updated tracestate, so the child span falls back to the parent tracestate and can keep an old `ot=th` value. Since this is a new independent probability decision, that threshold is no longer valid. Can you check the logic here.

sdk/src/trace/samplers/probability.cc
ayush-that: Okay. I'm reusing CalculateThreshold and deleting the duplicate.

sdk/src/trace/samplers/probability.cc
ayush-that: Okay. Reusing GetRandomnessFromTraceId.

sdk/src/trace/samplers/probability.cc
ayush-that: Fixed it to strip th on the drop path and return the recomputed tracestate, with rv and the other sub keys preserved. Added a test.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,211 @@
...
+
+  bool drop = threshold_ == kMaxThreshold;
+  if (!drop)
+  {
+    uint64_t randomness = 0;
+    if (!ExplicitRandomness(ot_value, &randomness))
+    {
+      if (parent_context.IsValid() && !parent_context.trace_flags().IsRandom())
+      {
+        static std::atomic<bool> warned{false};
```
ThomsonTan, line 96: Non-blocking: probably sample on the warning instead of just showing the first warning.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,211 @@
...
+    ot_state.has_threshold  = false;
+    std::string dropped_ot  = ot_state.Serialize();
+    auto trace_state        = parent_trace_state->Delete(kOtTraceStateKey);
+    if (!dropped_ot.empty())
+      trace_state = trace_state->Set(kOtTraceStateKey, dropped_ot);
+    return {Decision::DROP, nullptr, trace_state};
+  }
+
+  std::string new_ot_value = SetThresholdSubKey(ot_value, EncodeThreshold(threshold_));
+  if (!trace_api::TraceState::IsValidValue(new_ot_value))
```
ThomsonTan: Same stale-`th` class of issue you fixed for the drop path can still occur on the sample path: when `IsValidValue(new_ot_value)` is false or the tracestate is already full, we return an empty trace_state, so the child inherits the parent's (possibly stale) `ot=th`. Should we strip the parent `th` (reusing the drop-path `OtelTraceState` logic) before returning RECORD_AND_SAMPLE in those branches?

ayush-that: Yes. Reused the drop-path strip on the invalid-ot branch. The full branch is !has_ot, so nothing stale there. Added a test.

sdk/src/trace/samplers/probability.cc
ayush-that: Kept it once-per-process to stay off the per-span hot path.

sdk/src/configuration/sdk_builder.cc
dbarker, line 490: There needs to be a new `VisitProbability` method for the probability sampler along with a new `ProbabilitySamplerConfiguration` class and `ParseProbabilitySamplerConfiguration` method with tests.  

The `VisitComposableProbability` sampler method must create a composite sampler with the `ComposableProbabilitySampler` (configured by ComposableProbabilitySamplerConfiguration). 

This will match the others (see `VisitComposableAlwaysOn`, `VisitAlwaysOn`). 

Please see the schema to see how they differentiate between the probability sampler and composite sampler with composable probability :) 
  https://github.com/open-telemetry/opentelemetry-configuration/blob/cb8fbe1d57abeb8a27bf941e664cab646f5e078a/schema/tracer_provider.yaml#L132

dbarker (commented):
Thanks for the PR! To help move this through please just include the SDK ProbabilitySampler and factory implementation with tests in this PR and remove the sdk/configuration changes. To fully support YAML configuration more classes and tests will be needed and fit nicely into a follow up PR. See more details below.

sdk/src/configuration/sdk_builder.cc
ayush-that: makes sense @dbarker, i've removed the sdk_builder.cc change so this pr is just the sdk sampler, factory and tests. i'll do the yaml configuration support in a follow up pr.

sdk/include/opentelemetry/sdk/trace/samplers/probability.h
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * @return Returns either RECORD_AND_SAMPLE or DROP based on the comparison
+   * of the span's randomness value against the configured rejection
+   * threshold. The parent SampledFlag is ignored.
+   */
+  SamplingResult ShouldSample(
+      const opentelemetry::trace::SpanContext &parent_context,
+      opentelemetry::trace::TraceId trace_id,
+      nostd::string_view /*name*/,
```
dbarker: nit - these argument names do not need to be commented out.

sdk/include/opentelemetry/sdk/trace/samplers/probability_factory.h
```diff
@@ -0,0 +1,31 @@
...
+{
+
+/**
+ * Factory class for ProbabilitySampler.
+ */
+class ProbabilitySamplerFactory
+{
+public:
+  /**
+   * Create a ProbabilitySampler.
```
dbarker, line 24: nit - missing @param comment for ratio. Please include the valid range and default value.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,122 @@
...
+
+SamplingResult ProbabilitySampler::ShouldSample(
+    const trace_api::SpanContext &parent_context,
+    trace_api::TraceId trace_id,
+    nostd::string_view /*name*/,
+    trace_api::SpanKind /*span_kind*/,
+    const opentelemetry::common::KeyValueIterable & /*attributes*/,
+    const trace_api::SpanContextKeyValueIterable & /*links*/) noexcept
+{
+  auto parent_trace_state = parent_context.trace_state();
```
dbarker: We can check if the trace state is valid and if it is empty ( parent_context.trace_state()->Empty()) to short cut the parsing to follow.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,122 @@
...
+  if (!parent_trace_state)
+  {
+    parent_trace_state = trace_api::TraceState::GetDefault();
```
dbarker: Consider not setting parent_trace_state to a default when null and instead guard the Get and Parse accordingly. Allocating a default trace state makes a shared pointer copy of a global static object and seems unnecessary here.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,122 @@
...
+    ot_state.has_threshold = false;
+  }
+  else
+  {
+    ot_state.has_threshold = true;
+    ot_state.threshold     = threshold_;
+  }
+
+  std::string new_ot_value = ot_state.Serialize();
+  auto trace_state         = parent_trace_state->Delete(kOtTraceStateKey);
```
dbarker: Same as above. parent_trace_state can stay null/empty and a guard can be added here around Delete.

sdk/src/trace/samplers/probability.cc
```diff
@@ -0,0 +1,122 @@
...
+#include "opentelemetry/version.h"
+
+#include "ot_trace_state.h"
+
+namespace trace_api = opentelemetry::trace;
+
+namespace
+{
+// Clamps ratio to [0, 1]. NaN and negatives map to 0 (never-sample).
+double ClampProbability(double ratio) noexcept
```
dbarker: If the user passes a value outside the valid range please log a warning and set the spec default. Clamping without warning would be unexpected. 

The configuration spec gives a default value of 1.0 for ratio and range of 0-1.0. 

Please see the schema here: 
https://github.com/open-telemetry/opentelemetry-configuration/blob/df064c07150d5e19b41d169a295830a42c0fe8a9/schema/tracer_provider.yaml#L219

dbarker (commented):
Thanks for the updates! Please see my review feedback below requesting some minor changes.

sdk/include/opentelemetry/sdk/trace/samplers/probability_factory.h
ayush-that: Added the param doc with the valid range and the 1.0 default

sdk/src/trace/samplers/probability.cc
ayush-that: Done. parsing is now gated on the trace state being valid and non empty.

sdk/src/trace/samplers/probability.cc
ayush-that: I removed the default. The parse path only runs when the parent state is valid and nonempty.

sdk/src/trace/samplers/probability.cc
ayush-that: Done. Delete only runs when the parent had entries, null state stays null unless we need to set the new ot value.

sdk/src/trace/samplers/probability.cc
ayush-that: Done. Now logs a warning and falls back to 1.0 per the config spec default.

sdk/src/trace/samplers/ot_trace_state.cc
```diff
@@ -172,33 +176,48 @@ OtelTraceState OtelTraceState::Parse(const std::string &ot_value) noexcept
...
     }
+  }
+  if (!rest.empty())
+  {
     if (!out.empty())
     {
       out.push_back(';');
     }
-    out.append(pair);
+    out.append(rest);
```
dbarker, line 228: Can out exceed the 256 char limit now?

sdk/src/trace/samplers/ot_trace_state.cc
```diff
@@ -172,33 +176,48 @@ OtelTraceState OtelTraceState::Parse(const std::string &ot_value) noexcept
...
-  for (const auto &pair : other_subkeys)
+
+  std::string out;
+  if (has_threshold && threshold < kMaxThreshold)
   {
-    std::size_t extra = out.empty() ? pair.size() : pair.size() + 1;
-    if (out.size() + extra > 256)
+    out.append("th:");
+    AppendThresholdHex(out, threshold);
+    if (out.size() + (rest.empty() ? 0 : rest.size() + 1) > 256)
```
dbarker: Please define the 256 char limit as a constexpr and break out the additional size calculation from the if statement. Consider adding a comment on the intent.

sdk/test/trace/probability_sampler_test.cc
```diff
@@ -0,0 +1,495 @@
...
+}
+
+TEST(ProbabilitySampler, ProbabilitySamplerHalf)
+{
+  double ratio       = 0.5;
+  int iterations     = 100000;
+  int expected_count = static_cast<int>(iterations * ratio);
+  int variance       = static_cast<int>(iterations * 0.01);
+
+  trace_api::SpanContext c(true, true);
```
dbarker, line 451: does explicit randomness `rv` need to be set in the span context's trace state for these two tests?

dbarker (commented):
Thanks for the updates. Please see a few comments and questions below.

sdk/src/trace/samplers/ot_trace_state.cc
ayush-that: done, and it covers the parse guard too.

sdk/src/trace/samplers/ot_trace_state.cc
ayush-that: no. the guard checks the final size before adding th, so th is dropped instead of overflowing.
ComposableSampler.ThresholdOmittedOverSizeLimit covers it.

sdk/test/trace/probability_sampler_test.cc
ayush-that: no, it would break them. a fixed rv would override the fresh trace id each iteration generates and make all 100k decisions identical. ExplicitRandomnessTakesPrecedence covers that case.

dbarker (approved):
Looks good to me. Thanks!

marcalff (approved):
LGTM, thanks for the feature.

PR #4136, opened 2026-06-07
[CI] iwyu clang tidy install thirdparty script to unify dev container

proost:
## Changes

I'm using dev container. But IWYU warnings are different from CI. I tried many things such as rebuild container without cache, sync exactly version IWYU, clang. But nothing works. 

~So i'm suspecting third party. I might be wrong. I ask for your understanding to use CI resources.~

My hypothesis is right. 

here is why. Current CI install third party package using apt-get configured by os. So version of installed third party packages are different from install_thirdparty script. 

So I unify CI and dev container envs so developer can identity wrong format early before sending PR.


.github/workflows/clang-tidy.yaml
```diff
@@ -62,15 +52,32 @@ jobs:
           echo "Using clang-tidy version: $(clang-tidy --version)"
           echo "clang-tidy installed at: $(which clang-tidy)"
 
-      - name: Install rapidyaml
+      - name: Cache pinned third-party
```
dbarker: Please remove caching for now. There will need to be another effort and approach generally to speed up CI builds.

.github/workflows/iwyu.yml
```diff
@@ -43,23 +43,31 @@ jobs:
...
-            libgtest-dev \
-            libbenchmark-dev \
             llvm-22-dev \
             libclang-22-dev \
             clang-22 \
             cmake
           sudo update-alternatives --install /usr/bin/llvm-config llvm-config /usr/bin/llvm-config-22 200
 
-      - name: Install rapidyaml
+      - name: Cache pinned third-party
```
dbarker: Please remove caching.

.github/workflows/iwyu.yml
```diff
@@ -43,23 +43,31 @@ jobs:
...
+
+      - name: Install pinned third-party
+        if: steps.cache_thirdparty.outputs.cache-hit != 'true'
+        env:
+          # benchmark v1.9.4 trips clang-22's new -Wc2y-extensions on __COUNTER__
+          # and its CMakeLists uses -Werror -pedantic; silence the warning.
+          CXXFLAGS: -Wno-c2y-extensions
         run: |
-          sudo -E ./ci/install_thirdparty.sh --install-dir /usr/local --tags-file third_party_release --packages "ryml"
+          sudo -E ./ci/install_thirdparty.sh --install-dir /usr/local --tags-file third_party_release \
```
dbarker: this can just be `sudo -E ./ci/install_thirdparty.sh --install-dir /usr/local --tags-file third_party_release` with no `--packages` arg to install all required third party packages.

.github/workflows/clang-tidy.yaml
```diff
@@ -62,15 +52,32 @@ jobs:
...
+
+      - name: Install pinned third-party
+        if: steps.cache_thirdparty.outputs.cache-hit != 'true'
+        env:
+          # benchmark v1.9.4 trips clang-22's new -Wc2y-extensions on __COUNTER__
+          # and its CMakeLists uses -Werror -pedantic; silence the warning.
+          CXXFLAGS: -Wno-c2y-extensions
         run: |
-          sudo -E ./ci/install_thirdparty.sh --install-dir /usr/local --tags-file third_party_release --packages "ryml"
+          sudo -E ./ci/install_thirdparty.sh --install-dir /usr/local --tags-file third_party_release \
```
dbarker: can remove the `--packages` arg to install all required third party packages.

.github/workflows/clang-tidy.yaml
```diff
@@ -62,15 +52,32 @@ jobs:
...
+            --packages "ryml;zlib;googletest;benchmark;protobuf;grpc;abseil"
+
+      - name: Refresh ld.so.cache
```
dbarker: this step can be removed. The `install_thirdparty.sh` script runs `ldconfig` when the packages are installed to `/usr/local`

.github/workflows/iwyu.yml
```diff
@@ -43,23 +43,31 @@ jobs:
...
+            --packages "ryml;zlib;googletest;benchmark;protobuf;grpc;abseil"
+
+      - name: Refresh ld.so.cache
```
dbarker: this step can be removed

dbarker (commented):
Thanks for fix! Please see comments below.

.github/workflows/clang-tidy.yaml
proost: [2fa7273](https://github.com/open-telemetry/opentelemetry-cpp/pull/4136/commits/2fa727390ee6743785d93496e6bc3325796eaec0)

.github/workflows/iwyu.yml
proost: [2fa7273](https://github.com/open-telemetry/opentelemetry-cpp/pull/4136/commits/2fa727390ee6743785d93496e6bc3325796eaec0)

.github/workflows/iwyu.yml
proost: [2fa7273](https://github.com/open-telemetry/opentelemetry-cpp/pull/4136/commits/2fa727390ee6743785d93496e6bc3325796eaec0)

.github/workflows/clang-tidy.yaml
proost: [2fa7273](https://github.com/open-telemetry/opentelemetry-cpp/pull/4136/commits/2fa727390ee6743785d93496e6bc3325796eaec0)

.github/workflows/clang-tidy.yaml
proost: Oh, you are right. Thank you
[2fa7273](https://github.com/open-telemetry/opentelemetry-cpp/pull/4136/commits/2fa727390ee6743785d93496e6bc3325796eaec0)

dbarker (approved):
Looks good. Thanks for the CI fix.

marcalff:
> here is why. Current CI install third party package using apt-get configured by os. So version of installed third party packages are different from install_thirdparty script.

This is really weird, it seems IWYU is not so predictable after all.

The whole point of IWYU for file foo.cc is to include all headers that code in file.cc uses directly, and in my understanding that list is not supposed to depend on the content of other headers like third party.

> So I unify CI and dev container envs so developer can identity wrong format early before sending PR.

Thanks for this, this is the most important property.

marcalff (approved):
LGTM, thanks for the fix.

PR #4139, opened 2026-06-07
[METRICS] Use instrument creation time as delta start_ts for first interval

### LINKED ISSUE #4062: Delta temporality uses MeterProvider start time instead of per-instrument creation time for first collection interval
**Bug**:
The OpenTelemetry spec states:

For delta aggregations, the start timestamp MUST equal the previous collection interval's timestamp, or the creation time of the instrument if this is the first collection interval for the instrument.

The C++ SDK uses ```MeterContext::GetSDKStartTime()``` (the MeterProvider creation time) as ```start_ts``` for the first delta collection interval of all instruments, rather than the creation time of each individual instrument. This affects both the fast path and slow path in ```TemporalMetricStorage::buildMetrics```.

**Call chain**:
MeterContext constructor:
```
sdk_start_ts_{std::chrono::system_clock::now()}   // captured once

Meter::Collect():
  metric_storage.second->Collect(collector, ctx->GetCollectors(), ctx->GetSDKStartTime(), ...)

SyncMetricStorage::Collect():
  temporal_metric_storage_.buildMetrics(collector, collectors, sdk_start_ts, ...)

TemporalMetricStorage::buildMetrics():
  // slow path, first interval:
  opentelemetry::common::SystemTimestamp last_collection_ts = sdk_start_ts;  // should be instrument creation time
```
All instruments receive the same ```sdk_start_ts``` regardless of when they were created.

**Expected behavior**:
If an instrument is created at T=30s and the MeterProvider was created at T=0, the first delta data point for that instrument should have ```start_time_unix_nano = T=30s```.

**Actual behavior**:
The first delta data point has ```start_time_unix_nano = T=0``` (MeterProvider creation time) regardless of when the instrument was created.

**Impact**:
Backends that use ```start_time_unix_nano``` for data point placement (e.g., New Relic maps it to the ```timestamp``` field) will show the first data point at the wrong time in charts. For instruments created well after SDK initialization, the first delta interval spans an artificially long window.

**Reference implementation**:
The Java SDK handles this correctly in ```DeltaSynchronousMetricStorage```:
```
// Constructor stores per-instrument creation time
this.instrumentCreationEpochNanos = clock.now();

// Collection uses it as the fallback for the first interval
long startEpochNanos =
    registeredReader.getLastCollectEpochNanosOrDefault(instrumentCreationEpochNanos);
```
Source: [DefaultSynchronousMetricStorage.java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/main/java/io/opentelemetry/sdk/metrics/internal/state/DefaultSynchronousMetricStorage.java)

**Suggested fix**:
Store a creation timestamp in ```TemporalMetricStorage``` (or ```SyncMetricStorage```) initialized to ```std::chrono::system_clock::now()``` at construction time. Use this instead of the passed-in ```sdk_start_ts``` for the first collection interval's ```start_ts```.

**Related**: #4038 (fast path always using ```sdk_start_ts```, even on subsequent exports)

pranitaurlam:
Fixes #4062

## Summary

For delta aggregations, the OpenTelemetry spec requires:

> The start timestamp MUST equal the previous collection interval's timestamp, or **the creation time of the instrument** if this is the first collection interval.

The bug: `TemporalMetricStorage::buildMetrics` was using `sdk_start_ts` (MeterProvider creation time) as the `start_ts` for the **first** delta collection interval in both the fast path and slow path. This meant all instruments shared the same (wrong) start time regardless of when they were individually created.

**Fix**: Store `instrument_creation_ts_` (captured via `std::chrono::system_clock::now()` in the `TemporalMetricStorage` constructor) and use it instead of `sdk_start_ts` for the first delta interval in both code paths:

- **Fast path** (single collector, delta): `last_delta_collection_ts_` initialised to `instrument_creation_ts_` instead of `sdk_start_ts`
- **Slow path**: default `last_collection_ts` set to `instrument_creation_ts_` instead of `sdk_start_ts`

This matches the Java SDK behaviour in `DefaultSynchronousMetricStorage`, which stores `instrumentCreationEpochNanos` at construction and uses it as the fallback for the first collection.

## Test plan

- [ ] Existing `sync_metric_storage_*` tests pass
- [ ] Manual verification: instrument created at T=30s after MeterProvider init shows `start_time_unix_nano ≈ T=30s` (not T=0) on first delta export

cc @marcalff @lalitb @pixerit @ThomsonTan

### DIFF
--- a/sdk/include/opentelemetry/sdk/metrics/state/temporal_metric_storage.h
+++ b/sdk/include/opentelemetry/sdk/metrics/state/temporal_metric_storage.h
@@ -56,6 +56,7 @@ private:
   // Lock while building metrics
   mutable opentelemetry::common::SpinLockMutex lock_;
   const AggregationConfig *aggregation_config_;
+  opentelemetry::common::SystemTimestamp instrument_creation_ts_;
   opentelemetry::common::SystemTimestamp last_delta_collection_ts_;
   bool has_last_delta_collection_ts_ = false;
 };
--- a/sdk/src/metrics/state/temporal_metric_storage.cc
+++ b/sdk/src/metrics/state/temporal_metric_storage.cc
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#include <chrono>
 #include <list>
 #include <memory>
 #include <mutex>
@@ -34,7 +35,8 @@ TemporalMetricStorage::TemporalMetricStorage(InstrumentDescriptor instrument_des
                                              const AggregationConfig *aggregation_config)
     : instrument_descriptor_(std::move(instrument_descriptor)),
       aggregation_type_(aggregation_type),
-      aggregation_config_(aggregation_config)
+      aggregation_config_(aggregation_config),
+      instrument_creation_ts_(std::chrono::system_clock::now())
 {}
 
 bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
@@ -45,7 +47,7 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
                                          nostd::function_ref<bool(MetricData)> callback) noexcept
 {
   std::lock_guard<opentelemetry::common::SpinLockMutex> guard(lock_);
-  opentelemetry::common::SystemTimestamp last_collection_ts = sdk_start_ts;
+  opentelemetry::common::SystemTimestamp last_collection_ts = instrument_creation_ts_;
   AggregationTemporality aggregation_temporarily =
       collector->GetAggregationTemporality(instrument_descriptor_.type_);
 
@@ -56,7 +58,7 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
   {
     if (!has_last_delta_collection_ts_)
     {
-      last_delta_collection_ts_     = sdk_start_ts;
+      last_delta_collection_ts_     = instrument_creation_ts_;
       has_last_delta_collection_ts_ = true;
     }
     // If no metrics, early return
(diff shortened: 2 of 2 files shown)


pranitaurlam:
Hi @marcalff @lalitb @ThomsonTan @pixerit 

I've opened PR #4139 to fix issue #4062 — delta temporality using MeterProvider start time instead of per-instrument creation time for the first collection interval.

The fix is minimal (6 lines across 2 files): TemporalMetricStorage now captures its own creation timestamp at construction and uses it as start_ts for the first delta interval, matching the behaviour of the Java SDK.

Could one of you please approve the workflow runs so CI can execute? Any feedback is welcome.

Thank you!

sdk/include/opentelemetry/sdk/metrics/state/temporal_metric_storage.h
```diff
@@ -56,6 +56,7 @@ class TemporalMetricStorage
   // Lock while building metrics
   mutable opentelemetry::common::SpinLockMutex lock_;
   const AggregationConfig *aggregation_config_;
+  opentelemetry::common::SystemTimestamp instrument_creation_ts_;
```
pixerit, line 59: const?

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -45,7 +47,7 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
                                          nostd::function_ref<bool(MetricData)> callback) noexcept
 {
   std::lock_guard<opentelemetry::common::SpinLockMutex> guard(lock_);
-  opentelemetry::common::SystemTimestamp last_collection_ts = sdk_start_ts;
+  opentelemetry::common::SystemTimestamp last_collection_ts = instrument_creation_ts_;
```
lalitb, line 50: Can we scope this to `delta` temporality only? This initializes the slow-path start timestamp from the instrument creation time for all temporalities, which appears to change `cumulative` `start_ts` behavior as well.

lalitb:
@pranitaurlam  -  Thanks for working on this and for helping move the issue forward.

Another PR, #4144, has now been approved and covers the same issue with the delta start timestamp fix for both the fast and slow paths, along with targeted tests. Since that PR is further along, I think we can close this one to avoid duplicate work.

Really appreciate the contribution and the intent here.

marcalff:
Thanks for the contribution.

Closing, due to duplication with #4144.

PR #4141, opened 2026-06-07
[CONTEXT] Fix env carrier key normalization to be spec compliant

### LINKED ISSUE #4061: Make environment variable propagation spec compliant

Towards: https://github.com/open-telemetry/opentelemetry-specification/issues/5040

The current implementation of environment variable propagation carrier is not specification compliant: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/env-carriers.md

At quick glance:
- the name normalization is different
- the env var names read from the OS are not normalized when getting the value
- caching could be done only once

**Additional context**

Follows https://github.com/open-telemetry/opentelemetry-cpp/issues/3817

OTel .NET, Go, Java implementation should be already specification compliant.

pranitaurlam:
Fixes #4061

## Changes

- **ASCII-only normalization**: Replace `std::isalnum` (locale-dependent) with explicit ASCII range checks (`a-z`, `A-Z`, `0-9`)
- **Digit prefix**: Prepend `_` when normalized key starts with a digit, per the spec
- **Normalized cache key**: `Get()` now caches by the normalized key, so `Get("x-b3-traceid")` and `Get("X_B3_TRACEID")` hit the same cache entry and map to the same env var `X_B3_TRACEID`
- **Remove unused includes**: Drop `<algorithm>` and `<cctype>` since `std::transform`/`std::isalnum` are no longer used
- **Tests**: Add cases for hyphen/dot replacement, digit prefix, normalized cache key lookup

## Spec reference

https://opentelemetry.io/docs/specs/otel/context/env-carriers/#key-name-normalization

CC @marcalff @lalitb @ThomsonTan


pranitaurlam:
Hi @marcalff @lalitb @ThomsonTan 
could you please review and approve PR #4141? It fixes environment variable and carrier key normalization to be spec compliment (issue #4061). Thanks

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -48,58 +46,69 @@ class EnvironmentCarrier : public TextMapCarrier
...
-    auto cache_it = cache_.find(std::string(key));
+    // Cache is keyed by normalized name for consistent lookup
+    auto cache_it = cache_.find(env_name);
     if (cache_it != cache_.end())
     {
       return cache_it->second;
     }
 
-    // Read from environment
     const char *value = std::getenv(env_name.c_str());
```
lalitb, line 61: This only looks up the normalized requested name. The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/env-carriers.md) says `Get` should also normalize key names already present in the carrier before matching, so this would not match an existing env key like `x-b3-traceid`. Should we snapshot/list the environment and match normalized names there?

marcalff: @lalitb

This is correct, the spec just changed last week, with:

* https://github.com/open-telemetry/opentelemetry-specification/pull/5102

Given that is is already merged and that other languages implementations are already following it, we don't have much choice but to follow as well.

@pranitaurlam 

Edit: changes on hold, see later comments.

~To implement:~

Once a call to std::getenv() fails, the fallback is to iterate over every environment variables, normalize the key, and if the normalized key matches, get the value.

Please see how different SIG did this (see spec PR 5102), for example in java:

```
String normalizedKey = EnvironmentSetter.normalizeKey(key);
    // first, perform an optimistic lookup for an exact match on the normalized key
    String value = carrier.get(normalizedKey);
    if (value != null) {
      return value;
    }
    // next, iterate over the carrier normalizing each entry and evaluating for a match
    // if memory allocation becomes an issue, can implement using iterative normalization, comparing
    // an entry character by character to the normalized key, normalizing along the way.
    for (Map.Entry<String, String> entry : carrier.entrySet()) {
      if (EnvironmentSetter.normalizeKey(entry.getKey()).equals(normalizedKey)) {
        return entry.getValue();
      }
    }
```

See `man environ`.

marcalff (commented):
Thanks for the fix.

The code provided looks good.

A fallback case from the spec, to iterate on all env vars if a lookup fails, is missing and needs to be implemented.

See comments for details and suggestions.

marcalff:
@lalitb @pranitaurlam 

Some updates on this ...

I think the spec itself can be simplified, and discussed this with @pellared in:

* https://github.com/open-telemetry/opentelemetry-specification/issues/5143
* https://github.com/open-telemetry/opentelemetry-specification/pull/5144

If this is accepted in the spec, we have nothing more to do.
If this is rejected in the spec, we will adjust and comply then.

Accordingly, I suggest to merge this PR as is, and only adjust if needed, no need to implement code that is about to be removed.

Approving the PR then.

@lalitb please confirm if this sounds reasonable to you.

@pranitaurlam No change then, thanks for the fix.

marcalff (approved):
Thanks for the fix.

pellared (approved):

lalitb:
> @lalitb please confirm if this sounds reasonable to you.

Thanks @marcalff for digging more.  This sounds reasonable to me.  I agree we should avoid adding env enumeration now if the spec is likely moving away from that requirement. We can follow up if the spec clarification goes the other way.

PR #4149, opened 2026-06-11
[DOCS] clarify logger format handling

### LINKED ISSUE #2628: Is formating for log messages actually implemented?
I am trying to use formatting for log messages as suggested in https://github.com/open-telemetry/opentelemetry-cpp/blob/e86ceba79dc5999c7b3655fd006847a360b351a0/api/include/opentelemetry/logs/logger.h#L287. It does not seem to work.
Q: Is it actually implemented?
Q: Do you have an example code for me? Something like:
   ```
 std::map<std::string, std::string> labelsTrace = {{"userId", "myUserId"}};
    logger->Trace(
		  "traceLogMessage of user {userId}";
		  labelsTrace);
```

### PR DESCRIPTION
**Pull request by SyedAsadRazaDevops**
## Description

Clarifies the logger API documentation for the `format` parameter. The current wording links to message templates, which can imply that placeholders are interpolated from attributes. This change documents the current behavior: the format string is stored as the log record body, and structured data should be passed through attributes.

Fixes #2628

## Type of change

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Verification

- [x] `git diff --check`
- [x] `scripts/assert-cncf-github-identity.ps1 -RepoPath upstreams\open-telemetry\opentelemetry-cpp`

No runtime behavior changed.

### CODE REVIEW

File: api/include/opentelemetry/logs/logger.h [hunk 1]
```diff
@@ -430,8 +430,11 @@ class Logger
    *
    * @param severity severity of the log
    * @param event_id event identifier of the log
-   * @param format an utf-8 string following https://messagetemplates.org/
+   * @param format a UTF-8 string stored as the log record body
```
lalitb, line 433: I think the new note about no SDK-side interpolation is useful, but should we avoid removing the message-template intent entirely? My understanding is that `format` was kept as a user-facing API concept so custom exporters or upstream agents could interpret it as a message template, even though the OpenTelemetry spec does not require the SDK itself to format it. Maybe this should say that the SDK stores the string as the log record body and does not interpolate attributes, while preserving that exporters/agents may choose to interpret it as a message template.
cc @ThomsonTan  - I believe he added the original comment.

ThomsonTan: The original intention is recommending the user to pass in message-template. It is not required for SDK or exporter (we don't do that) to do the interpolation, but everything are there for the interpolation so it could be done later, like in the storage backend. Let's add clarification on this instead of revoing the intention.

dbarker: The spec has some guidance on the ergonomic API to emit events now and this is new since the C++ event API was implemented. 
https://opentelemetry.io/docs/specs/otel/logs/api/#ergonomic-api

> Ergonomic API
> Status: [Development](https://opentelemetry.io/docs/specs/otel/document-status/)
> 
> Languages MAY additionally provide a more ergonomic and convenient logging API.
> 
> The ergonomic API SHOULD make it more convenient to emit event records following the [event semantics](https://opentelemetry.io/docs/specs/semconv/general/events/).
> 
> The design of the ergonomic API SHOULD be idiomatic for its language.

I think the change to the comment in this PR is a good one as it clarifies what member of the log record the arg sets.

If we follow the event semantics then we should also rename the `format` arg to `body` and leave it to the caller to use as they wish (which may be a format string following https://messagetemplates.org/).

dbarker (approved):
Looks good to me. Thanks!

PR #4152, opened 2026-06-12
[METRICS] Support synchronous gauge with delta temporality

### LINKED ISSUE #4146: Support synchronous gauge instrument with delta aggregation temporality in C++ SDK

Yes. We use OpenTelemetry C++ as thirdparty library. At the moment, the OpenTelemetry C++ SDK does not support synchronous gauge instruments with delta aggregation temporality. When a `MetricCollector` configured for delta temporality encounters a gauge, it currently logs an internal error (see [metric_collector.cc#L48](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/state/metric_collector.cc#L48)) and fallback to cumulative temporality behavior. This blocks our use cases where consumers need a point-in-time gauge reading scoped to each collection window — a capability already supported in the OpenTelemetry Java SDK.

1. Remove the `OTEL_INTERNAL_LOG_ERROR` in [metric_collector.cc#L48](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/src/metrics/state/metric_collector.cc#L48) that unconditionally rejects gauge instruments when delta temporality is requested.
2. Emit only the most recent sample within the current collection window. For a synchronous gauge under delta temporality, the correct semantic is: export the last recorded value whose `sample_ts_` falls strictly within the current delta window `(previous_end_ts, current_end_ts]`. Samples from prior windows must not be re-emitted.

My workaround is to implement a custom `PushMetricExporter` that post-processes `ResourceMetrics` before forwarding to the downstream exporter:

1. Maintain a `std::unordered_map<std::string, SystemTimestamp>` to track the end timestamp of each metric's most recent collection window.
2. In `Export()`, for each gauge `MetricData`, compute the delta window start as the stored end timestamp from the previous call (falling back to `start_ts` on the first export).
3. Discard any `PointDataAttributes` whose `sample_ts_` does not exceed the delta window start — these are stale readings already emitted in a prior window.
4. Among the remaining in-window points, retain only the one with the highest `sample_ts_` (last-value semantics).
5. Re-label `aggregation_temporality` as `kDelta` and update `start_ts` / `end_ts` accordingly before forwarding.
6. Forward modified `ResourceMetrics` to downstream exporter

I'd like to know if my solution is appropriate and if I understand the logic correctly. I'm willing to implement the fix in the SDK if it's okay for the community. I would welcome any guidance on the preferred location and design before opening a PR.

**Additional context**

pranitaurlam:
Fixes #4146

## Summary

Removes the block in `MetricCollector::GetAggregationTemporality` that logged an error and silently fell back to cumulative when delta was requested for a synchronous gauge instrument.

The existing delta fast-path in `TemporalMetricStorage::buildMetrics` already provides the correct window-scoped last-value semantics for gauge: because `SyncMetricStorage::Collect` swaps and resets the live `attributes_hashmap_` on every collection, `delta_metrics` contains only values recorded since the previous collection. Stale attribute sets that were not updated in the current window are simply absent from the snapshot and are never re-emitted.

**Changes:**
- `metric_collector.cc`: remove the gauge+delta error/fallback; return the reader-requested temporality directly.
- `temporal_metric_storage.cc`: update fast-path comment to mention gauge and its window-scoped last-value behavior.
- `sync_metric_storage_gauge_test.cc`:
  - Add `kDelta` to both `INSTANTIATE_TEST_SUITE_P` suites and assert values for both temporalities.
  - Add `GaugeDeltaWindowTest.NoRecordWindowEmitsNothing`: record → collect → no record → collect emits nothing.
  - Add `GaugeDeltaWindowTest.StaleAttributeNotReemitted`: record A+B → collect → record A only → collect emits only A.
  - Add `GaugeDeltaWindowTest.RecordAfterEmptyWindowEmitsNewValue`: empty window followed by a new recording is exported correctly.

## Test plan

- [x] Existing cumulative gauge tests still pass with both temporalities.
- [x] `NoRecordWindowEmitsNothing`: second collection with no new recordings emits 0 points.
- [x] `StaleAttributeNotReemitted`: attribute not recorded in the current window is not re-exported.
- [x] `RecordAfterEmptyWindowEmitsNewValue`: new recording after an empty window is exported correctly.

### DIFF
--- a/sdk/test/metrics/sync_metric_storage_gauge_test.cc
+++ b/sdk/test/metrics/sync_metric_storage_gauge_test.cc
@@ -93,21 +93,15 @@ TEST_P(WritableMetricStorageTestFixture, LongGaugeLastValueAggregation)
           if (opentelemetry::nostd::get<std::string>(
                   data_attr.attributes.find("Room.id")->second) == "Rack A")
           {
-            if (temporality == AggregationTemporality::kCumulative)
-            {
-              EXPECT_EQ(opentelemetry::nostd::get<int64_t>(lastvalue_data.value_),
-                        bg_noise_level_2_roomA);
-            }
+            EXPECT_EQ(opentelemetry::nostd::get<int64_t>(lastvalue_data.value_),
+                      bg_noise_level_2_roomA);
             count_attributes++;
           }
           else if (opentelemetry::nostd::get<std::string>(
                        data_attr.attributes.find("Room.id")->second) == "Rack B")
           {
-            if (temporality == AggregationTemporality::kCumulative)
-            {
-              EXPECT_EQ(opentelemetry::nostd::get<int64_t>(lastvalue_data.value_),
-                        bg_noise_level_2_roomB);
-            }
+            EXPECT_EQ(opentelemetry::nostd::get<int64_t>(lastvalue_data.value_),
+                      bg_noise_level_2_roomB);
             count_attributes++;
           }
         }
@@ -121,7 +115,8 @@ TEST_P(WritableMetricStorageTestFixture, LongGaugeLastValueAggregation)
 
 INSTANTIATE_TEST_SUITE_P(WritableMetricStorageTestLong,
                          WritableMetricStorageTestFixture,
-                         ::testing::Values(AggregationTemporality::kCumulative));
+                         ::testing::Values(AggregationTemporality::kCumulative,
+                                           AggregationTemporality::kDelta));
 
 TEST_P(WritableMetricStorageTestFixture, DoubleGaugeLastValueAggregation)
 {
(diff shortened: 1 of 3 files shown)


pranitaurlam:
Hi @lalitb @marcalff @ThomsonTan — I've opened #4152 to address this.

The fix removes the error/fallback in MetricCollector::GetAggregationTemporality and relies on the existing delta fast-path in TemporalMetricStorage, which already provides the correct window-scoped last-value semantics (the live attributes_hashmap_ is swapped and reset on every collection, so stale attribute sets are never in the snapshot).

Tests added for the scenarios you mentioned: record → collect → no record → collect (emits nothing), stale attribute not re-emitted, and recovery after an empty window. Happy to adjust based on your feedback.

sdk/test/metrics/sync_metric_storage_gauge_test.cc
```diff
@@ -204,4 +194,171 @@ TEST_P(WritableMetricStorageTestFixture, DoubleGaugeLastValueAggregation)
...
+  opentelemetry::sdk::metrics::SyncMetricStorage storage(
+      instr_desc, AggregationType::kLastValue, proc,
+#  ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
+      ExemplarFilterType::kAlwaysOff, ExemplarReservoir::GetNoExemplarReservoir(),
+#  endif
+      nullptr);
+
+  std::shared_ptr<CollectorHandle> collector(
+      new MockCollectorHandle(AggregationTemporality::kDelta));
+  std::vector<std::shared_ptr<CollectorHandle>> collectors{collector};
```
lalitb, line 222: Can we also add a multi-reader case here? These tests cover the single-reader delta path, but with multiple readers the storage uses the shared unreported_metrics_ path. It would be good to verify `record -> collect -> no record -> collect` for a delta gauge there too, so stale gauge values are not emitted again.

niqiangpro-cell: I added tests for multi reader case in this PR https://github.com/open-telemetry/opentelemetry-cpp/pull/4159. It verifies the expected bahaviour for multi reader path as well. What I understood is that instead of merging delta_metrics with last reported metrics like what is done for cumulative temporality, we only need to export delta_metrics as is, i.e., not emit stale values.

lalitb:
@pranitaurlam  -   Since #4159 already carries the production-code change and adds the multi-reader coverage, can we narrow this PR to only the single-reader delta gauge tests that are not covered there? That would mean removing the `MetricCollector::GetAggregationTemporality()` change from this PR and keeping only the additional single-reader coverage, so the implementation change lands in one place.

lalitb:
closing for inactivity. Feel free to reopen if needed.

PR #4157, opened 2026-06-14
[SDK] LogRecord attribute limits enforcement

### LINKED ISSUE #4126: Implement LogRecord limits

thc1006:
Fixes #4126.

Implements the LogRecord attribute count and value length limits described
by the logs SDK spec
(https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits).
Limits flow from `LoggerProvider` through `LoggerContext` to each `LogRecord`
created by `Logger::CreateLogRecord`. The infrastructure existed in the
declarative configuration tier (`LogRecordLimitsConfiguration`) but was
never wired into the runtime pipeline.

## What changed

New header `sdk/include/opentelemetry/sdk/logs/log_record_limits.h`:

* `LogRecordLimits` struct with spec defaults
  `attribute_count_limit = 128` and
  `attribute_value_length_limit = SIZE_MAX` (unlimited).

SDK Recordable hierarchy:

* `Recordable` gains a non-pure virtual `SetLogRecordLimits` with a no-op
  default body. Existing implementations that do not enforce limits
  inherit the no-op and compile unchanged. The virtual is appended at
  the end of the vtable to keep the change additive.
* `ReadableLogRecord` gains a non-pure virtual `GetDroppedAttributesCount`
  returning zero by default.
* `ReadWriteLogRecord` overrides both. `SetAttribute` checks the count
  limit before inserting and truncates string / array-of-string values
  whose length exceeds the configured limit. Truncation is byte level,
  mirroring the existing Span attribute behavior.
* `MultiRecordable` propagates `SetLogRecordLimits` to every wrapped
  recordable.

OTLP exporter:

* `OtlpLogRecordable` overrides `SetLogRecordLimits`. `SetAttribute`
  drops attributes beyond the count limit and increments the proto
  `LogRecord.dropped_attributes_count` field. Strings and string-array
  values are truncated after `OtlpPopulateAttributeUtils::PopulateAttribute`
  populates the proto `AnyValue`.

Wiring:

* `LoggerContext` owns a `LogRecordLimits` value and exposes
  `GetLogRecordLimits()`. A new fourth parameter is appended to the
  existing constructor with a default-constructed default, so existing
  call sites compile unchanged.
* `Logger::CreateLogRecord` (both ABI v1 and ABI v2 variants) calls
  `recordable->SetLogRecordLimits(context_->GetLogRecordLimits())`
  immediately after `MakeRecordable`, before any user attribute write.
* A new `LoggerProviderFactory::Create` overload accepts
  `LogRecordLimits` and builds a `LoggerContext` with those limits
  internally.

Declarative configuration:

* `SdkBuilder::CreateLoggerProvider` now maps
  `LogRecordLimitsConfiguration` from the parsed
  `LoggerProviderConfiguration` to the runtime `LogRecordLimits` and
  passes them to the new factory overload. The existing FIXME-SDK
  comment about wiring limits is removed.

## Tests

`sdk/test/logs/log_record_limits_test.cc` (new):

* Default behavior: 200 attributes accepted, zero dropped when no
  limits object is supplied.
* Count enforcement: attributes beyond the limit are dropped and
  counted. Replacing an existing key while at the limit must not drop.
* Length truncation for strings and string arrays.
* Type selectivity: only string and array-of-string are truncated;
  int / double / bool pass through.
* Combined count + length case.
* Logger-level wiring test: a `TrackingRecordable` produced by a
  `TrackingProcessor` confirms that the limits configured on
  `LoggerProvider` reach the recordable returned by
  `Logger::CreateLogRecord`.

`exporters/otlp/test/otlp_log_recordable_test.cc` augmented with four
cases that verify the proto `dropped_attributes_count` field and the
string / array truncation logic.

## Verification

Built and tested in the `otelcpp-dev` container:

* Release (gcc, abiv1): `log_record_limits_test` 9/9,
  `otlp_log_recordable_test` 23/23, `logger_sdk_test` 9/9,
  `logger_provider_sdk_test` 9/9, `simple_log_record_processor_test`
  10/10, `batch_log_record_processor_test` 13/13. No regression.
* ABI v2 (clang): `log_record_limits_test` 9/9.
* `-fno-rtti` (Bazel nortti CI mirror): `log_record_limits_test` 9/9,
  `otlp_log_recordable_test` 23/23.
* clang-format fixed point verified for all touched files.
* IWYU (abiv1-preview): include-list reconciled with diagnostics
  from a local clang-22 run.
* `git diff --check` whitespace: clean.

## Known limitations

* `ElasticSearchRecordable` is not modified in this PR. It inherits
  the no-op default and does not enforce limits. A follow-up PR can
  add enforcement if desired.
* Truncation is byte level. Strings whose configured limit falls inside
  a multibyte UTF-8 sequence will be truncated mid-sequence, matching
  the existing Span attribute behavior in `OtlpRecordable`.
* The default branch in the YAML parser
  (`ParseLogRecordLimitsConfiguration`) keeps the existing
  `attribute_value_length_limit = 4096` fallback for callers that
  opt into the `limits:` block without specifying a length. Aligning
  that fallback with the spec default (unlimited) is left to a follow-up.


proost:
@thc1006 
I'm working on the feature in here(https://github.com/open-telemetry/opentelemetry-cpp/pull/4132).
Do you want to implement this feature?

thc1006:
@proost Hey, thanks for the ping. I went through #4132 and the review thread before pushing #4157. There's a fair bit of substantive work in yours that mine doesn't have: the OTEL_LOGRECORD_* env var integration, the benchmark file, ElasticSearchRecordable enforcement, and the yaml count_limit plumbing.

The reason I went ahead with #4157 even after seeing yours is lalitb's comment from yesterday (https://github.com/open-telemetry/opentelemetry-cpp/pull/4132#discussion_r3409765486), which was asking to move enforcement out of the base Recordable hot path and let each recordable handle it the way that fits. That's the shape #4157 takes. Recordable just gains a `SetLogRecordLimits` virtual with a no-op default body, and the actual enforcement lives in `ReadWriteLogRecord` and `OtlpLogRecordable`. Non-enforcing exporters inherit the no-op so they pay nothing.

If the maintainers think #4157 is the better base, I'd like to layer the env vars, benchmark, and ES enforcement on top in follow-ups with attribution to your work. If they prefer the #4132 redesign route, happy to close #4157 and review the new version there.

@marcalff @lalitb would appreciate your call on which to drive forward.

proost:
@thc1006 

Thanks for the explanation.

I'm always open to feedback, and if there were concerns about the direction of #4132, I would have been happy to discuss and adjust it. I do wish those concerns had been raised directly on the PR earlier, as it would have helped avoid duplicated work.

That said, I don't have a strong preference on whose implementation lands. If you would like to continue driving this work, I'm happy to step aside and close #4132 so we don't spend effort maintaining two competing PRs.

So would you like to continue driving this work?

thc1006:
@proost You're right and that one's on me. The honest reason I missed #4132: I was searching open issues by `help wanted` label and recent merged PRs for landscape, not by `fixes #4126` against open PRs. #4132 has no labels and the title in the open-PR list truncates around "attribute count l...", so I read it as a narrower length-only fix and moved on. That was a diligence gap on my part, not a deliberate decision to bypass your work. Sorry about that.

If you're comfortable stepping aside, yes I'd like to continue with #4157.

After #4157 lands I'd open these as follow-ups:

1. `OTEL_LOGRECORD_ATTRIBUTE_VALUE_LENGTH_LIMIT` and `OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` env-var integration through `LogRecordLimits` and the `SdkBuilder` path, modeled on your #4132 work and adding you as a `Co-authored-by` on the commit.
2. The benchmark file from #4132, ported onto the per-recordable enforcement so the numbers reflect the new architecture, attribution in the commit body.
3. `ElasticSearchRecordable` enforcement following the same `SetLogRecordLimits` pattern.
4. YAML `attribute_count_limit` parsing along the lines #4132 did.

If you'd rather author any of those follow-ups yourself and have me review, that works equally well. Whatever keeps the work landing.

proost:
Thanks for the clarification and the apology.

I’m happy for you to continue driving this work, so I’ll close #4132 to avoid duplicate effort.

thc1006:
@proost Thanks for the clean handoff. The reviewer feedback from owent, lalitb, and ThomsonTan on #4132 still applies here, and I'll reference the specific discussions when each follow-up goes up. I'll ping you on each so you can shape the attribution however works for you.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_log_recordable.h
```diff
@@ -114,6 +123,7 @@ class OtlpLogRecordable final : public opentelemetry::sdk::logs::Recordable
   const opentelemetry::sdk::resource::Resource *resource_ = nullptr;
   const opentelemetry::sdk::instrumentationscope::InstrumentationScope *instrumentation_scope_ =
       nullptr;
+  const opentelemetry::sdk::logs::LogRecordLimits *limits_ = nullptr;
```
lalitb: I think we should avoid storing a pointer to the limits object here. In the normal logger path, `CreateLogRecord()` sets this from `LoggerContext`, but the returned `LogRecord` is an independent `unique_ptr`. User code can keep that record and call `SetAttribute()` after the logger/provider/context is gone, which would leave `limits_` dangling.

Can we instead copy `LogRecordLimits` into the recordable instead of storing `&limits`? The struct is tiny, and it avoids adding a lifetime requirement to `LogRecord`.

sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
```diff
@@ -200,6 +214,9 @@ class ReadWriteLogRecord final : public ReadableLogRecord
   int64_t event_id_{0};
   std::string event_name_;
 
+  const LogRecordLimits *limits_{nullptr};
```
lalitb: same as [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4157#discussion_r3429631211).

exporters/otlp/src/otlp_log_recordable.cc
```diff
@@ -233,10 +238,59 @@ void OtlpLogRecordable::SetTraceFlags(const opentelemetry::trace::TraceFlags &tr
...
+{
+  if (value == nullptr)
+  {
+    return;
+  }
+  if (value->value_case() == proto::common::v1::AnyValue::kStringValue)
+  {
+    if (value->string_value().size() > max_length)
+    {
+      value->mutable_string_value()->resize(max_length);
```
lalitb: This truncates protobuf `string_value` by raw byte count after `PopulateAttribute()` has already decided it is a string. For valid UTF-8 input with a multi-byte character, this can split the character and leave an invalid UTF-8 protobuf string.

Can we preserve the earlier UTF-8 concern from #4132 here, or add a regression test with a multi-byte UTF-8 string? The current OTLP tests only cover ASCII truncation.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_log_recordable.h
thc1006: Agreed. Switched to by-value storage in `ba8f9047`. The default-constructed value carries the spec defaults (count=128, length=unlimited), so a fresh recordable now enforces the spec count cap from construction — happy to revisit that semantic if you'd prefer a no-op-when-unset sentinel instead.

sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
thc1006: Same fix landed in `ba8f9047` for `ReadWriteLogRecord`.

exporters/otlp/src/otlp_log_recordable.cc
thc1006: Ported the two-tier truncation contract you and @ThomsonTan / @owent settled on in #4132 (UTF-8 code-point safe prefix for strings, byte truncation for raw bytes), adapted to byte-length budget to match `attribute_value_length_limit`'s existing unit. While in the same truncation paths I also applied the cap to raw bytes attributes (`vector<uint8_t>` SDK side, `bytes_value` OTLP side), which were previously passing through any size. Added multi-byte, malformed-input, and bytes regression tests on both sides. Commit `ba8f9047`, with co-author attribution to @proost.

sdk/src/logs/read_write_log_record.cc
```diff
@@ -151,12 +155,128 @@ const opentelemetry::trace::TraceFlags &ReadWriteLogRecord::GetTraceFlags() cons
...
+
+void TruncateStringValue(opentelemetry::sdk::common::OwnedAttributeValue &value,
+                         std::size_t max_length) noexcept
+{
+  if (nostd::holds_alternative<std::string>(value))
+  {
+    auto &s = nostd::get<std::string>(value);
+    if (s.size() > max_length)
+    {
+      s.resize(Utf8SafePrefixLength(s, max_length));
```
owent: In my understanding ,we can use std::string to hold a bytes value here. Should we truncate it as bytes but not a UTF-8 string in this situation?

thc1006: You're right — switching SDK-side `TruncateStringValue` to plain byte truncation in `1612edcb`. The `OwnedAttributeValue::std::string` variant can legitimately carry raw bytes when constructed from a non-UTF-8 source, so forcing UTF-8 boundary semantics there over-truncates that case. OTLP keeps `Utf8SafePrefixLength` for its `kStringValue` path because protobuf strings have a wire-format requirement; the two recordables are independent storage paths so OTLP wire validity is still preserved via OtlpLogRecordable's own truncation. Dropped the SDK-side UTF-8 regression tests since the SDK no longer makes that contract.

exporters/otlp/src/otlp_log_recordable.cc
thc1006: Follow-up after @owent's r3434178971: narrowing the UTF-8-aware truncation to the OTLP path only. `OtlpLogRecordable::TruncateProtoStringValue` keeps `Utf8SafePrefixLength` (your protobuf wire-format concern is fully preserved), but SDK-side `ReadWriteLogRecord::TruncateStringValue` reverts to byte cut because the in-memory `std::string` variant can carry raw bytes from non-UTF-8 sources. Each recordable's truncation now matches its own consumer's wire-format requirement. Updated in `1612edcb`; happy to flip SDK back to UTF-8-aware if you'd prefer the symmetric form.

exporters/otlp/src/otlp_log_recordable.cc
```diff
@@ -233,10 +238,113 @@ void OtlpLogRecordable::SetTraceFlags(const opentelemetry::trace::TraceFlags &tr
...
+namespace
+{
+
+// Byte length of the longest prefix of `value` that fits within `max_bytes`
+// without splitting a well-formed UTF-8 multi-byte sequence. A lead byte's
+// declared length is only honored when its continuation bytes are present and
+// in range (0x80-0xBF); otherwise the lead is treated as a one-byte unit, so
+// malformed input degrades to plain byte truncation. This keeps the resulting
+// protobuf `string_value` valid UTF-8 when the input was valid UTF-8.
+std::size_t Utf8SafePrefixLength(const std::string &value, std::size_t max_bytes) noexcept
```
dbarker: please move these common utilities to `otlp_populate_attribute_utils.<h, cc>` and add tests.  They will also need to be used for enforcing the span limits attribute length in a follow up PR.

sdk/src/logs/read_write_log_record.cc
```diff
@@ -151,12 +155,78 @@ const opentelemetry::trace::TraceFlags &ReadWriteLogRecord::GetTraceFlags() cons
...
 
+namespace
+{
+
+// Byte-length truncation. The SDK in-memory std::string variant may carry
+// raw bytes when constructed from a non-UTF-8 source, so forcing UTF-8
+// boundary semantics here would over-truncate that case. Exporters that
+// require a valid-UTF-8 wire format (OTLP protobuf, ES JSON) apply their
+// own UTF-8-aware truncation at the recordable boundary.
+void TruncateStringValue(opentelemetry::sdk::common::OwnedAttributeValue &value,
```
dbarker: This will also be needed to enforce attribute length limits for spans as well. Consider moving to sdk/common/attribute_utils.<h, cc>

sdk/src/logs/read_write_log_record.cc
```diff
@@ -151,12 +155,78 @@ const opentelemetry::trace::TraceFlags &ReadWriteLogRecord::GetTraceFlags() cons
...
+
+}  // namespace
+
 void ReadWriteLogRecord::SetAttribute(nostd::string_view key,
                                       const opentelemetry::common::AttributeValue &value) noexcept
 {
   std::string safe_key(key);
+
+  if (attributes_map_.size() >= limits_.attribute_count_limit &&
+      attributes_map_.find(safe_key) == attributes_map_.end())
```
dbarker: at the limit this may result in two lookups for an existing attribute (here and line 213 below). Can this just be one lookup for an existing key?

dbarker (commented):
Thanks for the PR. Please see comments below.

exporters/otlp/src/otlp_log_recordable.cc
thc1006: Done in `c4532cd1`. `Utf8SafePrefixLength` and `TruncateProtoAttributeValue` (renamed from `TruncateProtoStringValue` to reflect it covers string_value, bytes_value, and array_value branches) now live as `OtlpPopulateAttributeUtils` static methods. New `exporters/otlp/test/otlp_populate_attribute_utils_test.cc` adds 14 direct unit tests — UTF-8 boundary preservation, malformed-input fallback, truncated-tail fallback, invalid lead bytes, bytes_value, array recursion, null safety. The future SpanLimits PR can call both helpers from the OTLP trace recordable.

sdk/src/logs/read_write_log_record.cc
thc1006: Done in `c4532cd1`. The SDK byte-length truncation is now `sdk::common::TruncateAttributeValueByteLength` in `sdk/include/opentelemetry/sdk/common/attribute_utils.h` (renamed since it covers string, string-array, AND bytes variants). Five new unit tests in `sdk/test/common/attribute_utils_test.cc` covering all three variants plus shorter-than-budget and non-string types. The future SpanLimits PR can reuse this from ReadWriteSpanData. / / Implementation note: I put it `inline` in the existing header rather than splitting to a new `.cc`, since `attribute_utils.h` is currently header-only (no `sdk/src/common/attribute_utils.cc` exists) and the helper is ~15 LOC, matching the precedent of `sdk/common/custom_hash_equality.h`. Happy to split to a new `.cc` if you'd prefer the standard pattern.

sdk/src/logs/read_write_log_record.cc
thc1006: Done in `c4532cd1`. `SetAttribute` now does `.find()` once, then conditional `.emplace()` only when inserting a new key. Existing-key replacement and new-key insertion each cost a single hash lookup. The end-to-end behavior contract (count cap drops new keys, replace existing key doesn't increment dropped count) is still verified by `CountLimitDropsExcessAttributes` and `CountLimitReplaceExistingKeyDoesNotDrop`.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -68,6 +68,48 @@ enum OwnedAttributeType : std::uint8_t
...
+ * std::string variant may legitimately carry raw bytes when constructed
+ * from a non-UTF-8 source, so forcing UTF-8 boundary semantics here would
+ * over-truncate that case. Exporters that require a valid-UTF-8 wire format
+ * (OTLP protobuf, ES JSON) apply their own UTF-8-aware truncation at the
+ * recordable boundary.
+ */
+inline void TruncateAttributeValueByteLength(OwnedAttributeValue &value,
+                                             std::size_t max_length) noexcept
+{
+  if (nostd::holds_alternative<std::string>(value))
```
dbarker: please use `nostd::get_if` since this is a noexcpt method and `nostd::get` throws.

thc1006: Done in `31449913`. Rewrote with `nostd::get_if` (returns nullptr if the alternative doesn't hold) so the noexcept contract is statically honored without any potentially-throwing call. Bonus: the same commit also restores the Bazel link of the new `otlp_populate_attribute_utils_test` target (missing `//sdk/src/metrics` transitive dep), which should resolve the 8 Bazel CI failures triggered on the previous push.

thc1006:
@dbarker the only failing CI on `8c965e7e` is **DocFX check** and it's a transient infrastructure failure unrelated to this PR — the GitHub Releases download for docfx.zip returned HTTP 500:

```
Attempt to get headers for https://github.com/dotnet/docfx/releases/download/v2.58.5/docfx.zip failed.
The remote server returned an error: (500) Internal Server Error.
Chocolatey installed 0/1 packages. 1 packages failed.
docfx (exited 404)
```

The job failed in 40 seconds during the docfx download step before any of the PR's content was even read. Would you mind re-running just this job (I don't have admin rights to do it myself)? All other 31 jobs on this head are SUCCESS / 2 SKIPPED, no other failures.

sdk/src/logs/read_write_log_record.cc
```diff
@@ -155,8 +158,38 @@ void ReadWriteLogRecord::SetAttribute(nostd::string_view key,
...
+    it = attributes_map_
+             .emplace(std::move(safe_key), opentelemetry::sdk::common::OwnedAttributeValue{})
+             .first;
+  }
+
   opentelemetry::sdk::common::AttributeConverter converter;
-  attributes_map_[safe_key] = nostd::visit(converter, value);
+  it->second = nostd::visit(converter, value);
+
+  if (limits_.attribute_value_length_limit != (std::numeric_limits<std::size_t>::max)())
```
dbarker: Is there a reason to truncate values after converting from the non-owning (AttributeValue) to the owning (AttributeValue) type? 

Can we apply the limit and truncation within the existing AttributeConverter on allocation of the owning type?

sdk/src/logs/read_write_log_record.cc
```diff
@@ -155,8 +158,38 @@ void ReadWriteLogRecord::SetAttribute(nostd::string_view key,
...
+
+  auto it = attributes_map_.find(safe_key);
+  if (it == attributes_map_.end())
+  {
+    if (attributes_map_.size() >= limits_.attribute_count_limit)
+    {
+      ++dropped_attributes_count_;
+      return;
+    }
+    it = attributes_map_
```
dbarker: Consider eliminating the two step default OwnedAttributeValue creation and update assignment. The map element can be constructed with the truncated value directly (if truncated in the converter).

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -431,6 +431,95 @@ void OtlpPopulateAttributeUtils::PopulateAttribute(
...
+      value->mutable_bytes_value()->resize(max_length);
+    }
+    return;
+  }
+  if (value->value_case() == proto::common::v1::AnyValue::kArrayValue)
+  {
+    auto *array = value->mutable_array_value();
+    for (int i = 0; i < array->values_size(); ++i)
+    {
+      TruncateProtoAttributeValue(array->mutable_values(i), max_length);
```
dbarker: Since the source AttributeValue is a flat variant recursion is not needed and should be avoided here.  

Related to the comment in ReadWriteLogRecord, consider removing this method and performing truncation on creation of the owned value in the PopulateAnyValue method. The limit can be passed in to these methods optionally with the default of max std::size_t.

dbarker (commented):
Thanks for the cleanup. Please see comments below. The main question is: Can the truncation be applied on creation of the owned type instead of after in order to prevent allocating the overflowing value?

sdk/src/logs/read_write_log_record.cc
thc1006: Done in `ba057d45`. `AttributeConverter` now has an `explicit AttributeConverter(std::size_t max_length)` constructor, and the string/bytes/string-array overloads apply the byte-length cap during the OwnedAttributeValue construction. The default-constructed converter is unchanged so the other callers (`instrumentation_scope.h`, `ReadWriteLogRecord::SetBody`) keep their no-truncation behavior. Helper `TruncateAttributeValueByteLength` is deleted; its three branches now live in the converter.

sdk/src/logs/read_write_log_record.cc
thc1006: Done in `ba057d45`. `ReadWriteLogRecord::SetAttribute` is now a single `find` + conditional `emplace`, with the limit-aware converter producing the truncated value in one step. Removed the previous default-construct-then-assign sequence and the now-redundant post-truncate branch.

exporters/otlp/src/otlp_populate_attribute_utils.cc
thc1006: Done in `ba057d45`. All four `PopulateAttribute` / `PopulateAnyValue` overloads now take a trailing `std::size_t max_length = numeric_limits<size_t>::max()` parameter (the default preserves existing call-site behavior for the other callers in metrics/recordable/resource paths). `TruncateProtoAttributeValue` is deleted along with its `NOLINT(misc-no-recursion)` block — the flat-AttributeValue input means the recursive array descent was indeed dead code. String branches inside `PopulateAnyValue` now use `Utf8SafePrefixLength` in place, and the bytes branch byte-cuts to `max_length`. Also added a `Utf8SafePrefixLength(const char*, std::size_t, std::size_t)` overload so call sites with a `string_view` don't have to round-trip through a temporary `std::string`. The `ENABLE_OTLP_UTF8_VALIDITY` path's proto field type (string_value vs bytes_value) is still decided by the original input's validity, not the post-truncation prefix, preserving the prior semantics.

sdk/src/logs/logger.cc
```diff
@@ -162,6 +162,7 @@ opentelemetry::nostd::unique_ptr<opentelemetry::logs::LogRecord> Logger::CreateL
 
   auto recordable = context_->GetProcessor().MakeRecordable();
 
+  recordable->SetLogRecordLimits(context_->GetLogRecordLimits());
```
lalitb: Can we avoid adding this unconditional virtual call on every `CreateLogRecord()`? This is now paid by all SDK log paths, including custom recordables that do not enforce limits.

One cleaner option would be passing `LogRecordLimits` through `MakeRecordable(...)`, but that looks API-breaking for existing exporters. A safer incremental option may be to compute/store a config flag once and only call `SetLogRecordLimits()` for exporters/recordables that actually enforce limits, like OTLP and debug/ostream.

That would keep the extra limit work out of unrelated SDK hot paths.

thc1006: Thanks for catching this. v7 moved truncation into the recordable (stateful AttributeConverter), so the `SetLogRecordLimits` setter is what carries that state, and you are right that paying for the call on the common path is not justified when the recordable ignores it.

The "compute/store a config flag once" path lines up well: a `RecordableEnforcesLogRecordLimits()` virtual on `LogRecordProcessor` (and `LogRecordExporter`, defaulting `false`), cached by `Logger` at construction, and used to gate the existing call:

```cpp
if (recordable_enforces_limits_) {
  recordable->SetLogRecordLimits(context_->GetLogRecordLimits());
}
```

OTLP and ostream exporters override to return `true`. `MultiLogRecordProcessor` returns true if any child does. Custom processors keep the default and pay only a bool load plus a predicted branch. The new virtuals append to the existing vtable, so subclass layout is unchanged.

If that lines up with what you had in mind, I will push v8 with this gate.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -81,11 +101,30 @@ struct AttributeConverter
...
+      v.resize(max_length_);
+    }
+    return OwnedAttributeValue(std::move(v));
+  }
+  OwnedAttributeValue operator()(const char *v)
+  {
+    std::string s(v);
+    if (s.size() > max_length_)
+    {
+      s.resize(max_length_);
```
owent: I suggest keeping UTF-8-aware truncation here. The reasoning:

1. For UTF-8 strings: If a user stores a UTF-8 string in std::string, they would expect the truncated result to remain valid UTF-8. A naive byte-length truncation risks splitting a multi-byte code point, which could cause the OTLP exporter to fall back to bytes_value instead of the intended string_value.
2. For raw bytes: If a user stores raw binary data in std::string, they would expect simple byte-length truncation without any encoding semantics.

thc1006: Thanks, agreed. The v7 docstring defends byte cut on the grounds that `std::string` may carry raw bytes from a non-UTF-8 source, but the OTel convention that `std::string` means UTF-8 (with `nostd::span<const uint8_t>` for raw bytes) is the stronger signal here.

The v8 plan:

1. Move `Utf8SafePrefixLength` into `sdk/include/opentelemetry/sdk/common/attribute_utils.h` alongside `AttributeConverter` (inline, no new file; the OTLP-side wrapper forwards into it).
2. Route the `AttributeConverter` `std::string` / `nostd::string_view` / `const char*` overloads through it, plus the per-element cut in the `nostd::span<const nostd::string_view>` array overload; `nostd::span<const uint8_t>` keeps byte cut as you describe.
3. The existing 8 OTLP-side `Utf8SafePrefixLength` tests stay; `sdk/test/common/attribute_utils_test.cc` gets equivalent SDK-side coverage.
4. Docstring updates to match.

This composes with the gating ask in r3468037047; both land in the same v8 push as separate commits.

sdk/src/logs/logger.cc
lalitb: Yes, that direction makes sense to me. One small refinement: since `LogRecordLimits` and the processor/exporter chain are provider/context-level configuration, can we cache this capability on `LoggerContext` rather than on each `Logger`?

That would still avoid the unconditional `SetLogRecordLimits()` virtual call in `CreateLogRecord()`, but it also avoids adding per-logger state for a provider-level setting. The hot path would just read the context-level flag and call `SetLogRecordLimits()` only when some processor/exporter actually enforces limits.


For example:

```cpp
  LoggerContext {
    LogRecordLimits limits_;
    bool recordable_enforces_limits_;
  }
```

Then each logger can gate the setup with:

```cpp
  if (context_->RecordableEnforcesLogRecordLimits())
  {
    recordable->SetLogRecordLimits(context_->GetLogRecordLimits());
  }
```

thc1006: Agreed, that's a cleaner placement. `LoggerContext` is where the processor chain and `LogRecordLimits` already live, so the capability flag belongs there. I'll add the cached `recordable_enforces_limits_` member and the `RecordableEnforcesLogRecordLimits()` getter on `LoggerContext`, populated at construction from the processor chain. Both gating call sites in `Logger::CreateLogRecord()` will route through the context getter as you showed.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -81,11 +101,30 @@ struct AttributeConverter
...
+  {
+    if (v.size() > max_length_)
+    {
+      v.resize(max_length_);
+    }
+    return OwnedAttributeValue(std::move(v));
+  }
+  OwnedAttributeValue operator()(const char *v)
+  {
+    std::string s(v);
```
dbarker: this const char* path can delegate to the nostd::string_view operator to avoid allocating the full string if it overflows.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -101,48 +104,54 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
   }
   else if (nostd::holds_alternative<const char *>(value))
   {
-    const char *str_value = nostd::get<const char *>(value);
+    const char *str_value      = nostd::get<const char *>(value);
+    const std::size_t str_len  = strlen(str_value);
```
dbarker: nit - std::strlen

sdk/include/opentelemetry/sdk/common/attribute_utils.h
```diff
@@ -81,11 +101,30 @@ struct AttributeConverter
...
+    if (s.size() > max_length_)
+    {
+      s.resize(max_length_);
+    }
+    return OwnedAttributeValue(std::move(s));
+  }
+  OwnedAttributeValue operator()(nostd::span<const uint8_t> v)
+  {
+    const auto end = v.begin() + static_cast<std::ptrdiff_t>((std::min)(v.size(), max_length_));
```
dbarker: is the cast required here?

sdk/include/opentelemetry/sdk/common/attribute_utils.h
thc1006: Good catch. Will delegate:

```cpp
OwnedAttributeValue operator()(const char *v) { return (*this)(nostd::string_view(v)); }
```

This also naturally inherits the UTF-8-aware path once `Utf8SafePrefixLength` is wired into the string_view overload (per r3471819117 / r3471966844). Bundled into v8.

exporters/otlp/src/otlp_populate_attribute_utils.cc
thc1006: Will fix to `std::strlen` in v8.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
thc1006: Right, the cast is only there to silence the signed/unsigned arithmetic warning. Using `v.subspan(0, ...)` avoids it cleanly:

```cpp
OwnedAttributeValue operator()(nostd::span<const uint8_t> v)
{
  const auto truncated = v.subspan(0, (std::min)(v.size(), max_length_));
  return OwnedAttributeValue(std::vector<uint8_t>(truncated.begin(), truncated.end()));
}
```

Will refactor in v8.

ThomsonTan:
The introduced limit changes the current default behavior, a default-constructed `ReadWriteLogRecord` / `OtlpLogRecordable` now enforces the 128 attribute cap even when no limits were configured and the logger wiring path was never used. Suggest iither make "no explicit SetLogRecordLimits = unlimited" the recordable default and let the wiring layer inject 128, or keep the spec default but call this out in CHANGELOG as a breaking change.

thc1006:
BTW the `v+num` is Number of fixed commit (revisions)

thc1006:
Good catch, you're right. A default-constructed ReadWriteLogRecord / OtlpLogRecordable starts with limits_{} = {128, unlimited}, so it caps at 128 even when no provider wired it up. That goes against the intent that these limits are provider-level.

I'll go with option 1: the recordable defaults to no limits, and the wiring injects the configured ones. This lines up with the RecordableEnforcesLogRecordLimits gating in the latest revision: LoggerContext keeps the spec default ({128, unlimited}) and the Logger injects it through SetLogRecordLimits only when a processor enforces. The result:

- a bare recordable, or one behind a non-enforcing processor: no limits, matching the pre-PR behavior.
- a recordable created through the standard provider path: the spec default of 128 (or whatever the provider was configured with) applies.

Concretely I'll default limits_ to unbounded in both recordables, keep the LoggerContext default at the spec values, and update the DefaultRecordEnforcesSpecCountCap test to match. The CHANGELOG will describe the provider-path enforcement as the new behavior.

Happy to do option 2 instead (keep the spec default on the recordable and just document it) if you'd rather, but option 1 looks closer to the provider-level intent.

sdk/include/opentelemetry/sdk/common/attribute_utils.h
thc1006: Pushed in v8. One correction to my suggestion above: nostd::span has no subspan(), so the bytes overload uses the two-iterator vector constructor instead, which is equally free of the cast:

```cpp
const std::size_t kept = (std::min)(v.size(), max_length_);
return OwnedAttributeValue(std::vector<uint8_t>(v.data(), v.data() + kept));
```

sdk/test/logs/log_record_limits_test.cc
```diff
@@ -0,0 +1,331 @@
...
+  record.SetAttribute("k2", static_cast<int64_t>(2));
+  // Replace existing key while at limit; must not increment dropped count.
+  record.SetAttribute("k1", static_cast<int64_t>(99));
+
+  ASSERT_EQ(record.GetAttributes().size(), 2u);
+  ASSERT_EQ(record.GetDroppedAttributesCount(), 0u);
+  ASSERT_EQ(nostd::get<int64_t>(record.GetAttributes().at("k1")), 99);
+}
+
+TEST(LogRecordLimits, LengthLimitTruncatesString)
```
dbarker, line 89: Please also test truncation for the `const char*`  and span paths.

sdk/include/opentelemetry/sdk/logs/recordable.h
```diff
@@ -43,6 +46,19 @@ class Recordable : public opentelemetry::logs::LogRecord
   virtual void SetInstrumentationScope(
       const opentelemetry::sdk::instrumentationscope::InstrumentationScope
           &instrumentation_scope) noexcept = 0;
+
+  /**
+   * Apply attribute count and value length limits to this log. The default
+   * implementation is a no-op; concrete recordables that wish to enforce
+   * limits override this and store the supplied limits before any
+   * SetAttribute call is observed. The referenced object must outlive the
```
dbarker: please update the comment now that recordable should copy the limits

sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
```diff
@@ -153,13 +154,26 @@ class ReadWriteLogRecord final : public ReadableLogRecord
   void SetAttribute(nostd::string_view key,
                     const opentelemetry::common::AttributeValue &value) noexcept override;
 
+  /**
+   * Apply attribute count and value length limits. Must be called before any
+   * SetAttribute call to take effect. The referenced limits object must
```
dbarker: please update the comment to account for limits being copied to recordables.

sdk/include/opentelemetry/sdk/logs/exporter.h
```diff
@@ -68,6 +68,19 @@ class OPENTELEMETRY_EXPORT LogRecordExporter
...
       std::chrono::microseconds timeout = (std::chrono::microseconds::max)()) noexcept = 0;
+
+  /**
+   * Returns true when the recordable produced by MakeRecordable() enforces
+   * LogRecord attribute limits (count and value length). The default returns
+   * false so the SDK can skip the per-record SetLogRecordLimits() call for
+   * recordables that ignore limits. Exporters whose recordable applies the
+   * limits (OTLP, ostream) override this to return true.
+   *
+   * This virtual is appended at the end of the LogRecordExporter vtable to keep
```
dbarker: this second paragraph in the comment can be removed as an implementation detail.

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -101,48 +103,54 @@ void OtlpPopulateAttributeUtils::PopulateAnyValue(
   }
   else if (nostd::holds_alternative<const char *>(value))
   {
-    const char *str_value = nostd::get<const char *>(value);
+    const char *str_value      = nostd::get<const char *>(value);
+    const std::size_t str_len  = std::strlen(str_value);
```
dbarker, line 107: are there worthwhile short cuts that can be taken if the max_length limit is not set (or is the default max size_t) ?

dbarker (commented):
Thanks for the revision. Please see feedback below. Given the changes please review the code comments and update if needed.

sdk/test/logs/log_record_limits_test.cc
thc1006: Added a const char* truncation test (LengthLimitTruncatesConstCharPtr). The span paths are already covered: LengthLimitTruncatesEachStringInArray for the string array and LengthLimitTruncatesBytesAttribute for the bytes span.

sdk/include/opentelemetry/sdk/logs/recordable.h
thc1006: Updated: the comment now says implementations copy the supplied limits, so the caller does not need to keep the object alive.

sdk/include/opentelemetry/sdk/logs/read_write_log_record.h
thc1006: Updated to say the limits are copied into the recordable.

sdk/include/opentelemetry/sdk/logs/exporter.h
thc1006: Removed. I also removed the same paragraph from the LogRecordProcessor version in processor.h for consistency.

exporters/otlp/src/otlp_populate_attribute_utils.cc
thc1006: Good point. Added an early return in Utf8SafePrefixLength: when max_bytes >= size (which includes the default unbounded SIZE_MAX), the whole value fits so there is nothing to truncate and the per-byte scan is skipped. It lives in the shared helper, so both this OTLP path and the in-memory AttributeConverter benefit.

sdk/include/opentelemetry/sdk/logs/recordable.h
```diff
@@ -43,6 +46,19 @@ class Recordable : public opentelemetry::logs::LogRecord
...
           &instrumentation_scope) noexcept = 0;
+
+  /**
+   * Apply attribute count and value length limits to this log. The default
+   * implementation is a no-op; concrete recordables that wish to enforce
+   * limits override this and copy the supplied limits before any
+   * SetAttribute call is observed, so the caller does not need to keep the
+   * supplied object alive.
+   *
+   * This virtual is appended at the end of the Recordable vtable to keep
```
dbarker: this second paragraph is an implementation detail and can be removed.

dbarker (approved):
Looks good to me. Thanks. Approved, but we need to hold on merging until benchmarks are added and compared with main. 

Please run the `otlp_recordable_benchmark` to verify changes to the attribute utils did not introduce a regression for the otlp span recording path. There is no benchmark for log recordables now. At least the oltp log recordable benchmark needs to be added in a separate PR.

sdk/include/opentelemetry/sdk/logs/recordable.h
thc1006: Removed, thanks. That matches the paragraphs already dropped from the exporter and processor headers.

thc1006:
Heads up: the red `W3C Distributed Tracing Validation V1` check is a flaky harness timeout, not the change. The job log shows `w3c_tracecontext_http_test_server` left as an orphan process and force-terminated at the 5-minute mark with no test output, while V2 (the same validation, newer spec) passed in the same run. The diff in the latest commit is a comment-only removal in `recordable.h`, which cannot affect trace-context propagation; the prior commit `0ac2cf3f` was 68/0 green. A re-run should clear it whenever you next look at this.

owent (approved):
LGTM on my end. Given the scope, let's get one more reviewer to double-check before merging.

lalitb (approved):
LGTM from my side. I agree with @dbarker we should at least compare the existing `otlp_recordable_benchmark` against main for this PR, since the shared attribute utils changed. A dedicated `otlp_log_recordable_benchmark` can be a follow-up PR.

thc1006:
Ran `otlp_recordable_benchmark` against main before merge, per your note.

Built it from both `main` (a56f0e23) and this PR in Release with identical flags. The benchmark source is byte-identical on both refs, so the only variable is the attribute-utils change. I confirmed the two binaries differ, then ran them interleaved and order-balanced across 20 paired rounds, reporting the paired per-round delta with a 95% CI (pairing cancels per-round environment drift):

| benchmark | PR vs main (95% CI) |
|---|---|
| Nominal (6 attrs) | -0.8% ± 2.0% |
| MaxAttributes (128 attrs) | +0.7% ± 2.2% |
| PopulateRequest, 512 spans | -0.2% ± 1.9% |

Every interval spans zero, so no measurable regression on the span recording path. `MaxAttributes`, which exercises the attribute conversion hardest, is flat. This matches the code: `OtlpRecordable::SetAttribute` uses the default unbounded `max_length`, so `Utf8SafePrefixLength` takes its `if (max_bytes >= size) return size;` early-out and inlines away (no call sites in the binary), leaving one size comparison per attribute.

I'll add the dedicated `otlp_log_recordable_benchmark` you mentioned as a separate follow-up PR.

(Local Release build in the otelcpp-dev image.)

dbarker (approved):

PR #4159, opened 2026-06-15
[METRICS] Enable synchronous gauge delta temporality and add tests for multi reader path

### LINKED ISSUE #4146: Support synchronous gauge instrument with delta aggregation temporality in C++ SDK

niqiangpro-cell:
Fixes #4146 

## Changes

Delta temporality for synchronous gauge instruments was previously unsupported and silently fell back to cumulative. This PR removes that restriction and verifies correct behaviour on the multi-reader slow path (`unreported_metrics_`).

- `metric_collector.cc`
Removed the guard that rejected `kDelta` for `InstrumentType::kGauge` and force-overrode it to `kCumulative`. Delta temporality for synchronous gauge is now fully supported.

- `temporal_metric_storage.cc`
Clarified the fast-path comment: it applies to all instrument types (not just counter/updown-counter/histogram), and documents the window-scoped last-value semantics for synchronous gauge — only values recorded since the previous collection are emitted; stale values are not re-exported.

- `sync_metric_storage_gauge_test.cc`
  - Added `WritableMetricStorageDeltaMultiReaderTestFixture`, a dedicated fixture for multi-reader delta tests (prevents cross-instantiation with the existing cumulative fixture).
  - Added `LongGaugeLastValueAggregationDeltaTemporalityMultiReader`(parameterized, int64_t).
  - Added `DoubleGaugeLastValueAggregationDeltaTemporalityMultiReader`(parameterized, double).

Both new tests exercise the slow path (`collectors.size() > 1`) and verify:

1. Each collector independently tracks its own measurement intervals.
2. First interval `start_ts` equals instrument creation time (spec [#4062]).
3. Subsequent `start_ts` advances to the previous `end_ts`.
4. Empty intervals (`record` → `collect` → `no record` → `collect`) do not invoke the callback — stale gauge values are not re-emitted, for both `collector_a` and `collector_b` independently.
5. A late-joining collector (`collector_b`) receives the merged last-value from all prior intervals with `start_ts` = instrument creation time.


sdk/test/metrics/sync_metric_storage_gauge_test.cc
```diff
@@ -205,3 +209,360 @@ TEST_P(WritableMetricStorageTestFixture, DoubleGaugeLastValueAggregation)
...
+  EXPECT_GE(metric_a_cycle1.start_ts.time_since_epoch(), before_creation.time_since_epoch());
+  EXPECT_LE(metric_a_cycle1.start_ts.time_since_epoch(), after_creation.time_since_epoch());
+  EXPECT_GT(metric_a_cycle1.start_ts.time_since_epoch(), sdk_start_ts.time_since_epoch());
+  EXPECT_EQ(metric_a_cycle1.end_ts, collection1_ts);
+
+  // --- Empty interval for collector_a (no new records) ---
+  // The callback must NOT be invoked; the interval timestamp must be updated.
+  auto collection2_ts = std::chrono::system_clock::now();
+  count_a             = 0;
+  storage.Collect(collector_a.get(), collectors, sdk_start_ts, collection2_ts,
```
lalitb, line 295: Could we also assert that the callback is not invoked for the empty delta intervals? Right now the test only checks that zero points were observed, so an empty `MetricData` callback would still pass. Same applies to the other empty interval checks in this file.

lalitb (approved):
LGTM, with nit comment. Also the PR title should be updated because this PR is not only adding tests. It also enables sync gauge delta temporality by removing the fallback to cumulative.

niqiangpro-cell:
Hi, @marcalff @ThomsonTan When you have a moment, could you please review this PR?
I believe it's ready, but happy to make any changes if needed.
Thanks a lot for your time!

niqiangpro-cell:
> LGTM, with nit comment. Also the PR title should be updated because this PR is not only adding tests. It also enables sync gauge delta temporality by removing the fallback to cumulative.

Thank you for your review. Do you know if this change could make it into the next release cycle?

PR #4160, opened 2026-06-15
[API/SDK] add `WITH_LOG_FILTERING_PREVIEW` option

dbarker:
Follow up to https://github.com/open-telemetry/opentelemetry-cpp/pull/4079#discussion_r3380409904

## Changes

- adds `WITH_LOG_FILTERING_PREVIEW` option to guard the new log record filtering behavior 
- updates tests to cover the option 
- fixes some existing clang-tidy warnings in the impacted tests. 


api/include/opentelemetry/logs/logger.h
```diff
@@ -86,11 +86,10 @@ class Logger
...
    *       @c Enabled(severity, ...) before invoking this overload if they want
-   *       the LoggerConfig filtering rules (minimum severity, trace-based,
-   *       processor.Enabled) to be honored. In ABI v2 builds, the no-record
-   *       overload @c EmitLogRecord(args...) below does call the filter chain
-   *       automatically when @c Severity is present in @p args; in ABI v1
-   *       builds neither overload applies the filter chain.
+   *       filtering to be honored. The no-record overload @c EmitLogRecord(args...)
+   *       below applies filtering automatically when @c ENABLE_LOG_FILTERING_PREVIEW
+   *       is defined: severity, trace based, and processor based filtering are supported in ABI v2,
+   *       only severity filtering is supported ABI v1.
```
proost: nit; `is supported in ABI v1.`

proost (approved):
Thank you!

lalitb (approved):
Thanks.

api/include/opentelemetry/logs/logger.h
marcalff:
```suggestion
   *       only severity filtering is supported in ABI v1.
```

api/include/opentelemetry/logs/logger.h
```diff
@@ -154,23 +153,32 @@ class Logger
...
+   *  With the flag: a LogRecord severity filter is applied in both ABI versions.
+   *  With the flag and ABI v2: if @c ExtendedEnabledRequired() is set, a full
+   *  context-aware filter is also applied (event_id, trace-based,
+   *  processor.Enabled).
    */
   template <class... ArgumentType>
   void EmitLogRecord(ArgumentType &&...args)
   {
-#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+#if defined(ENABLE_LOG_FILTERING_PREVIEW)
```
marcalff, line 170: This symbol, `ENABLE_LOG_FILTERING_PREVIEW`, must be defined when building the user application, because logger.h is part of the API, header only.

For this feature to work:

* Any instrumented application, previously compiled against opentelemetry-cpp headers from v1.27.0 or older, must be compiled again with the new headers.

* Somehow the application build must define `ENABLE_LOG_FILTERING_PREVIEW`.

Other `ENABLE_XXX` flags are used only in the SDK, so naming does not matter much, and are (lazily) defined in a central place like api/CMakeList.txt, even when they belong to the sdk or exporters.

In this case, the flag should be renamed to `OPENTELEMETRY_ENABLE_LOG_FILTERING_PREVIEW` to avoid polluting the user build, and how to do a build in the user application should be documented with clarity.

Open questions, for my understanding:

* why do we need this flag, what does it achieve ?
* given that this is a preview flag, it will be enabled and removed in the long term, so the target behavior is to build with the flag. Why would a user application want to build without the flag for now ?
* In other words, can we remove the flag and consider LOG_FILTERING always enabled ? What are the negative effects of doing this ?
* is the LOG_FILTERING flag totally decoupled from OPENTELEMETRY_ABI_VERSION_NO, that we need all combinations and have a new flag ?

marcalff (commented):
Thanks for the patch.

See questions, for my own understanding, I don't grasp the rationale behind it.

api/include/opentelemetry/logs/logger.h
dbarker: Thanks Marc, Please see this discussion for some background: https://github.com/open-telemetry/opentelemetry-cpp/pull/4079#discussion_r3370293274

> In this case, the flag should be renamed to OPENTELEMETRY_ENABLE_LOG_FILTERING_PREVIEW to avoid polluting the user build, and how to do a build in the user application should be documented with clarity.

Sounds reasonable on the name change. Would a change log entry be enough for the documentation? 

> why do we need this flag, what does it achieve ?

Adding the preview flag allows ABIv1 to eventually gain the minimum severity log record filtering spec compliance after some number of releases.  
 
Without the the flag, the log filtering behavior is lumped into the ABI v2 guard. It becomes an undocumented behavior change that isn't strictly defined by an ABI breaking change. It seems unexpected for users when they switch to ABI v2 and ideally we leave the ABI guards for ABI breaking changes.

> given that this is a preview flag, it will be enabled and removed in the long term, so the target behavior is to build with the flag. Why would a user application want to build without the flag for now ?

In the thread above, the concern raised was that multiple instrumentation libraries in an application may include different versions of otel-cpp. If they set the minimum severity for a logger then libraries including/linked to 1.28+ version will filter logs per the spec and libraries including/linked to lower versions would not.   

> In other words, can we remove the flag and consider LOG_FILTERING always enabled ? What are the negative effects of doing this ?

This also seems reasonable. The default for minimum severity set in the SDK is 0 per the spec. This would not filter any log records. If a user/library sets the minimum severity other than 0 then logs may be filtered. 

> is the LOG_FILTERING flag totally decoupled from OPENTELEMETRY_ABI_VERSION_NO, that we need all combinations and have a new flag ?

It is partially decoupled. Minimum severity log record filtering is ABI v1 and v2 compatible. Trace based and processor/context based filtering requires ABIv2. The preview flag covers all filtering in this PR for simplicity.

marcalff: Thanks for the details.

> In the thread above, the concern raised was that multiple instrumentation libraries in an application may include different versions of otel-cpp.

With a compile flag in the API headers like `ENABLE_LOG_FILTERING_PREVIEW`, the concern with multiple libraries compiled differently still stands, some will filter with v1.28.0, some will not.

> If they set the minimum severity for a logger then libraries including/linked to 1.28+ version will filter logs per the spec and libraries including/linked to lower versions would not.

Then, we should add the check also in the SDK.

The early filtering in the API is an optimization, to allow the API code to bail out early, but this is not the final check.

Code compiled with API v1.28.0 will honor filtering early.
Code compiled with API v1.27.0 or older will not honor filtering ... in the API.

In all cases, the SDK will have the final word if it checks for filtering again, this will make instrumentation compiled with v1.27.0 or older compliant, which is what we want.

dbarker: > With a compile flag in the API headers like ENABLE_LOG_FILTERING_PREVIEW, the concern with multiple libraries compiled differently still stands, some will filter with v1.28.0, some will not.

The preview flag doesn't solve the multiple version problem (an old version may emit logs that a new version would filter). It would only communicate early that this behavior will change and allow instrumentation library maintainers to opt in. After a few releases go by then the flag is set to default on and ultimately removed. 

> Then, we should add the check also in the SDK.

Agreed. This would be better. The SDK has access to the LogRecord object in EmitLogRecord. There isn't an API to get the severity from LogRecord to perform the check though. We would need to add a GetSeverity virtual method to the recordable and then override it in all the recordable implementations.

api/CMakeLists.txt
```diff
@@ -110,6 +110,11 @@ if(WITH_THREAD_INSTRUMENTATION_PREVIEW)
                              INTERFACE ENABLE_THREAD_INSTRUMENTATION_PREVIEW)
 endif()
 
+if(WITH_LOG_FILTERING_PREVIEW)
+  target_compile_definitions(opentelemetry_api
```
dbarker, line 119: > In this case, the flag should be renamed to OPENTELEMETRY_ENABLE_LOG_FILTERING_PREVIEW to avoid polluting the user build, and how to do a build in the user application should be documented with clarity.

@marcalff since all these preview flags are set with `INTERFACE` they will transit publicly to any targets that link to the `opentelemetry-cpp::api` target. If we apply a prefix it likely should be applied to all public definitions. Could this be a follow up PR?

api/include/opentelemetry/logs/logger.h
marcalff: Thanks for all the details.

Approved.

marcalff (approved):
LGTM

api/include/opentelemetry/logs/logger.h
dbarker: Thanks. I'll add a changelog entry.

PR #4165, opened 2026-06-16
[BENCHMARK] add span otlp recordable benchmark

dbarker:
Fixes # (issue)

## Changes

- benchmark the `OtlpRecordable` recording path (Span Start->End) and the export path (populating the OTLP request message). 

### DIFF
--- a/exporters/otlp/test/otlp_recordable_benchmark.cc
+++ b/exporters/otlp/test/otlp_recordable_benchmark.cc
@@ -1,0 +1,306 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <benchmark/benchmark.h>
+
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "opentelemetry/common/attribute_value.h"
+#include "opentelemetry/common/timestamp.h"
+#include "opentelemetry/exporters/otlp/otlp_recordable.h"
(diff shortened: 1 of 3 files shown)


exporters/otlp/test/otlp_recordable_benchmark.cc
```diff
@@ -0,0 +1,306 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
```
lalitb, line 2: nit: We do not usually do this today, but it would be useful to include the benchmark results along with the hardware/environment used to run them. That gives us a baseline to compare against when future OTLP recordable changes are made. This is common practice we do in [otel-rust](https://github.com/open-telemetry/opentelemetry-rust/blob/88821497a893ff6dd4dd916621a2224394ebb0a4/opentelemetry-sdk/benches/log_enabled.rs#L8).

lalitb (approved):
thanks.

marcalff (approved):
LGTM, thanks for the benchmarks.

exporters/otlp/test/otlp_recordable_benchmark.cc
dbarker: Thanks. I will add this.

dbarker: done

lalitb: Thanks , looks good.

PR #4170, opened 2026-06-19
[CONFIGURATION] Apply default sampler when none is specified

### LINKED ISSUE #4034: [CONFIGURATION] default sampler in config
**Steps to reproduce**

Using the yaml configuration parsing, when a sampler is not specified in the yaml configuration, the sampler is not set on the tracer provider.

I expected the default sampler (parent based always on) to be set as defined in the configuration schema https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#tracerprovider

The sampler is null

RaviTriv:
Fixes #4034 

## Changes

PR to apply default sampler when one is not set in the trace provider configuration. Updated the parser to apply `ParentBasedSamplerConfiguration` with `AlwaysOnSamplerConfiguration` as per the [spec](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#tracerprovider-) when no sampler is specified.


sdk/test/configuration/yaml_trace_test.cc
```diff
@@ -522,7 +522,16 @@ file_format: "1.0-trace"
   auto config = DoParse(yaml);
   ASSERT_NE(config, nullptr);
   ASSERT_NE(config->tracer_provider, nullptr);
-  ASSERT_EQ(config->tracer_provider->sampler, nullptr);
+  ASSERT_NE(config->tracer_provider->sampler, nullptr);
+  auto *sampler = config->tracer_provider->sampler.get();
+  auto *parent =
+      reinterpret_cast<opentelemetry::sdk::configuration::ParentBasedSamplerConfiguration *>(
+          sampler);
+  ASSERT_NE(parent->root, nullptr);
```
ThomsonTan, line 620: The new test verifies the sampler is `ParentBased` with a non-null `root`, but it does not check that `root` is specifically an `AlwaysOnSamplerConfiguration`. Since the spec default is `ParentBased(root=AlwaysOn)`, the test would still pass if the default root were accidentally changed to `AlwaysOff`. Suggest asserting the root type too so the default is fully pinned down.

RaviTriv: Good catch, updated to assert that root is `AlwaysOn`

ThomsonTan (approved):

lalitb:
I think there is one remaining default-sampler edge case. This PR handles the case where `sampler` is omitted entirely, but an explicit:

  ```yaml
  sampler:
    parent_based:
  ```

still leaves `ParentBasedSamplerConfiguration::root` as `nullptr`.

The config spec says `ParentBasedSampler.root` defaults to `always_on` when [omitted](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#parentbasedsampler). Since `SdkBuilder::CreateParentBasedSampler()` later calls `CreateSampler(model->root)`, should we also apply the `always_on` default when `parent_based.root` is omitted?

RaviTriv:
> I think there is one remaining default-sampler edge case. This PR handles the case where `sampler` is omitted entirely, but an explicit:
> 
> ```yaml
> sampler:
>   parent_based:
> ```
> 
> still leaves `ParentBasedSamplerConfiguration::root` as `nullptr`.
> 
> The config spec says `ParentBasedSampler.root` defaults to `always_on` when [omitted](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#parentbasedsampler). Since `SdkBuilder::CreateParentBasedSampler()` later calls `CreateSampler(model->root)`, should we also apply the `always_on` default when `parent_based.root` is omitted?

Good catch! Updated the parser to apply the `always_on` default when `parent_based.root` is omitted.

sdk/test/configuration/yaml_trace_test.cc
```diff
@@ -522,7 +607,21 @@ file_format: "1.0-trace"
   auto config = DoParse(yaml);
   ASSERT_NE(config, nullptr);
   ASSERT_NE(config->tracer_provider, nullptr);
-  ASSERT_EQ(config->tracer_provider->sampler, nullptr);
+  ASSERT_NE(config->tracer_provider->sampler, nullptr);
+  auto *sampler = config->tracer_provider->sampler.get();
+  auto *parent =
```
marcalff, line 617: Before casting `sampler` as a parent based sampler, please add a check with `TestSamplerVisitor` to validate the sampler type.

marcalff (commented):
LGTM, see a minor comment on the tests.

sdk/test/configuration/yaml_trace_test.cc
RaviTriv: good idea, updated to add check for sampler type `kParentBased` before the cast.

marcalff (approved):
LGTM, thanks for the fix

PR #4171, opened 2026-06-19
[SEMANTIC CONVENTIONS] semconv generate events

proost:
## Changes

I found that ["exception"](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md#exception-event) is missing.

["exception" itself is on the event not the attribute
](https://github.com/open-telemetry/semantic-conventions/blob/main/model/exceptions/events.yaml).

So i add new template.


marcalff (approved):
LGTM, thanks for the feature.

Nicely done, integrates well with the existing code generation.

PR #4175, opened 2026-06-23
[TEST] fix metrics simple example 

### LINKED ISSUE #3458: Segfault in metrics_simple example on Windows
The metrics_simple example application sometimes segfaults in Windows. 

A reproduction was observed in ci testing #3449 
- https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/15358056328/job/43220935557?pr=3449#step:6:2576

### LINKED ISSUE #3740: metrics_simple example does nothing on Windows
**Describe your environment**
Same as in https://github.com/open-telemetry/opentelemetry-cpp/issues/3673
Windows 11, VS2022, cmake-4.1.1, ninja 1.13.1
opentelemetry-cpp dependencies are built as shared libraries

**Steps to reproduce**
Build opentelemetry 1.23.0 on Windows as a single shared dll (using OPENTELEMETRY_BUILD_DLL cmake option)as described in  https://github.com/open-telemetry/opentelemetry-cpp/issues/3673, including patching ext/src/dll/input.src, and replacing "`-DWITH_EXAMPLES=OFF -DWITH_EXAMPLES_HTTP=OFF`" with  "`-DWITH_EXAMPLES=ON -DWITH_EXAMPLES_HTTP=ON`". The build will have built the metrics_simple example and then will fail due to failure to link other examples (because of the same underlying issues as in https://github.com/open-telemetry/opentelemetry-cpp/issues/3673).

Then put the build opentelemetry-cpp.dll (along with it's dependencies such as protobuf/abseil DLLs) on PATH and run the built <BUILD_ROOT>\opentelemetry-cpp-1.23.0\build\examples\metrics_simple\metrics_ostream_example.exe with or without any parameters such as "counter".

Exported metrics displayed on stdout, just like it happens on Linux.

The exe appears to hang for 10 seconds then it exits. Nothing is displayed on stdout. 

**Additional context**
Modifying the example to set up an OtlpHttpMetricExporter using http://localhost:56633/v1/metric as the url and install a custom  LogHandler via opentelemetry::sdk::common::internal_log::GlobalLogHandler::SetLogHandler shows there is no attempt to connect to the non-existent OTLP HTTP metric exporter url (unlike when attempting to use a non-existent OTLP Http trace exporter url for traces).

dbarker:
Fixes #3458 (maybe also fixes #3740)

## Changes

- Fix race conditions on the static double observable counter in the example foo metrics library.
- Ensure each observable counter test uses a unique instance of the MeasurementFetcher object (passed by the state ptr).
- Enable the example for ci testing 
- Fix race conditions on rand() and clang-tidy warnings in that example: 

----

**examples/common/metrics_foo_library/foo_library.cc** (8 warnings)

| Line | Check | Message |
|---|---|---|
| 45 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 46 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 58 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 81 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 109 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 126 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 162 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |
| 190 | `misc-predictable-rand` | rand() has limited randomness; use C++11 random library instead |

----


marcalff (approved):
LGTM, thanks for the fix.

examples/metrics_simple/CMakeLists.txt
```diff
@@ -20,7 +20,4 @@ endif()
 if(BUILD_TESTING)
   add_test(NAME examples.metrics_simple
            COMMAND "$<TARGET_FILE:metrics_ostream_example>")
-  # Disable the metrics_simple example test due to sporadic segfaults on Windows
```
lalitb, line 23: Can we also replace the shared `rand()` calls with a per thread RNG before enabling this test? The PR fixes the observable counter state, but the default test still runs several examples in parallel, and they all use `rand()`, so Windows CI may still be flaky.

lalitb (approved):
LGTM - trusting rand issue would be fixed as well. Thanks.

examples/metrics_simple/CMakeLists.txt
dbarker: Good catch. That also fixes some clang-tidy warnings :)

ThomsonTan (approved):

examples/common/metrics_foo_library/foo_library.cc
```diff
@@ -32,8 +33,11 @@ namespace metrics_api = opentelemetry::metrics;
 namespace
 {
 
-static opentelemetry::nostd::shared_ptr<metrics_api::ObservableInstrument>
-    double_observable_counter;
+int random_int()
```
ThomsonTan, line 36: nit: probably name it as `thread_safe_random_int` which acts like document for the user reading this example. Or add a note in comments to highlight this requirement.

PR #4180, opened 2026-06-24
[DOC] update ABI policy docs to clarify breaking changes

dbarker:
Fixes # (issue)

Updates the ABI policy documents to expand on allowed changes and to cover API header implementation changes that impact observable behavior (clarification based on #4174). 

## Changes
- Update the policy to highlight the case where an application links to multiple instrumented libraries (where each may include API headers from different otel-cpp versions). 
- Add note and link to the stdlib doc for when building with c++ standard types. 
- Provide examples of API header changes that may be allowed in a stable version and what must target an experimental version
- Provide example and guidance on observable behavior changes (changes to method implementations) and requires these to be targeted to the experimental ABI version due to potential ODR violations.  
- Updates the ABI version policy to point to CMake and Bazel files as the source of truth for the current stable and experimental ABI version


docs/abi-policy.md
```diff
@@ -20,12 +26,28 @@ ABI stability is not provided by the interfaces the SDK provides as
 implementation hooks to vendor implementors, like exporters, processors,
 aggregators and propagators.
 
-These are some of the rules for where ABI stability is required.
+The following sections describe common cases where ABI stability is impacted
+and provide examples of how to apply changes.
 
-Note: This assumes export maps are used to properly control symbol resolution.
+ABI breaking changes must target the experimental ABI version, guarded
```
lalitb: nit - Can we scope this sentence to the ABI-stable API surface? Maybe say **_"For ABI-stable API surfaces, breaking changes…”_**
`

docs/abi-version-policy.md
```diff
@@ -285,29 +285,38 @@ This status is meant as a preview, until the ABI is declared STABLE.
...
 
 In the REMOVED status,
 the given ABI is no longer available.
 
+The build files are the source of truth for which ABI version is currently
+experimental versus stable. In CMake, the experimental version is the
+`WITH_ABI_VERSION_*` option labeled `EXPERIMENTAL` in `CMakeLists.txt`, and it
+is off by default; any version without that label is stable. The Bazel
+equivalent is the `//api:abi_version_no` flag in `api/BUILD`, which defaults to
+the stable version.
```
lalitb: The CMake part works because the option text labels v2 as `EXPERIMENTAL`, but `api/BUILD` does not currently label `abi_version_no=2` as experimental. If build files are the source of truth, can we either add that marker to the Bazel flag/config_setting or avoid saying Bazel carries the same stable/experimental source-of-truth information?

docs/abi-policy.md
```diff
@@ -1,13 +1,19 @@
...
 
 Many parts of the C++ standard library don't have well-defined ABIs. To ensure
-that a plugin compiled against one version of the C++ standard library can work
-with an application or library compiled against a different version of the C++
-standard library, we limit a portion of the OpenTelemetry API to use ABI stable
-types.
+that a component compiled against one version of the C++ standard library can
+work with an application or library compiled against a different version of the
+C++ standard library, we limit a portion of the OpenTelemetry API to use ABI
+stable types by default. The `OPENTELEMETRY_STL_VERSION` build option replaces
```
lalitb: small nit - `OPENTELEMETRY_STL_VERSION` is the macro/Bazel define, while CMake exposes this as `WITH_STL`.

docs/abi-policy.md
```diff
@@ -58,3 +80,107 @@ private:
...
+#if OPENTELEMETRY_ABI_VERSION_NO >= 2
+  // OK: new data member added only under the experimental ABI guard.
+  bool new_member_{};
+#endif
+};
+```
+
+## Definitions in API headers
+
+Entities defined in API headers (inline functions, templates, and
```
lalitb: nit - Entities defined in ABI-stable API headers

lalitb (approved):
LGTM overall. Thanks for tightening the ABI guidance.

docs/abi-version-policy.md
dbarker: Thanks for the feedback. Updated these cases, please take a look.

docs/abi-policy.md
```diff
@@ -58,3 +92,107 @@ private:
...
+    }
+    nostd::unique_ptr<LogRecord> log_record = CreateLogRecord(/* context_or_span */);
+  #else
+    // Stable ABI behavior
+    nostd::unique_ptr<LogRecord> log_record = CreateLogRecord();
+  #endif
+
+    ...
+  }
+}
```
ThomsonTan: Nit: the closing brace of this `class Logger` example is missing the trailing semicolon (`};`).

dbarker: good eye. fixed.

marcalff (approved):

PR #4186, opened 2026-06-26
[OTLP/HTTP] Honor Retry-After header when retrying exports 

### LINKED ISSUE #4172: [OTLP/HTTP] Honor Retry-After header when retrying exports
The OTLP/HTTP retry path currently uses the configured retry policy backoff, but it does not honor the `Retry-After` response header.

This is out of line with the OTLP/HTTP spec clarification in open-telemetry/opentelemetry-proto#806: `Retry-After` can be either an HTTP-date or a delay in seconds, and clients SHOULD honor the value when present.

I verified this against current `main`:

- `exporters/otlp/src/otlp_http_client.cc` passes `options_.retry_policy` to the HTTP request via `request->SetRetryPolicy(...)`.
- `ext/include/opentelemetry/ext/http/client/http_client.h` models `RetryPolicy` as max attempts plus exponential backoff parameters.
- `ext/src/http/client/curl/http_operation_curl.cc` checks retryable status codes `429`, `502`, `503`, and `504` under `ENABLE_OTLP_RETRY_PREVIEW`.
- `HttpOperation::NextRetryTime()` computes the next retry time from the retry policy using exponential backoff and jitter.
- The retry decision/timing path does not parse response headers, so neither `Retry-After: <delay-seconds>` nor `Retry-After: <HTTP-date>` affects the next retry time.

### Expected behavior

When an OTLP/HTTP response is retryable and includes `Retry-After`, the exporter should honor the header value before retrying.

The parser should support both forms defined by RFC 7231 section 7.1.3:

- delay-seconds, for example `Retry-After: 5`
- HTTP-date, for example `Retry-After: Wed, 21 Oct 2015 07:28:00 GMT`

### References

- open-telemetry/opentelemetry-proto#806
- https://github.com/open-telemetry/opentelemetry-proto/pull/806#issuecomment-4616366444
- https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3

DCchoudhury15:
Fixes #4172
## Changes

The OTLP/HTTP retry logic wasn't respecting the `Retry-After` response header. If a server returned a `429` or `503` along with `Retry-After`, the exporter would ignore that value and always use its configured exponential backoff instead.

This PR updates the retry behavior to honor the `Retry-After` header as defined in RFC 7231 §7.1.3. Both supported formats are handled:

* A delay in seconds (for example, `Retry-After: 5`)
* An HTTP-date (for example, `Retry-After: Wed, 21 Oct 2015 07:28:00 GMT`)

One implementation detail worth noting is that `NextRetryTime()` is called from the `doRetrySessions` polling loop after `ReleaseResponse()` has already cleared the response headers. Because of that, the parsed `Retry-After` value is cached in `retry_after_time_point_` before `ReleaseResponse()` is called, and `NextRetryTime()` uses the cached value instead of trying to parse the headers again.

Unit tests are not included yet. I can add them directly to this PR
if the maintainers would like the plan would be to extend the existing
test server fixture with a `/retry-after/` endpoint that sets the header,
and add two test cases to `ext/test/http/curl_http_test.cc` following
the pattern of the existing `ExponentialBackoffRetry` test: one for
`Retry-After: <delay-seconds>` and one for `Retry-After: <HTTP-date>`.
Just let me know and I will push them here.

**Known limitation:** The HTTP-date parser currently supports only the IMF-fixdate format. The two obsolete date formats (RFC 850 and `asctime`) are not supported. Since RFC 7231 specifies IMF-fixdate as the required format for senders, this should not be an issue in practice.


ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1513,6 +1585,29 @@ void HttpOperation::PerformCurlMessage(CURLcode code)
 
   if (IsRetryable())
   {
+    auto headers = GetResponseHeaders();
```
owent: Should we only parse "Retry-After" when a retry is actually needed? Parsing the header and paying CPU cost on every response seems unnecessary if there's no failure.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,69 @@
...
+
+  int day  = tm->tm_mday;
+  int days = 365 * year + year / 4 - year / 100 + year / 400 + 367 * month / 12 - 30 + day - 719530;
+
+  return static_cast<std::time_t>(days) * 86400 + tm->tm_hour * 3600 + tm->tm_min * 60 + tm->tm_sec;
+}
+
+bool ParseRetryAfterDelay(std::string value, std::chrono::seconds &delay)
+{
+  value.erase(0, value.find_first_not_of(" \t\r\n"));
```
owent: We can use `common::StringUtil::Trim` to trim string.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,69 @@
...
+  value.erase(value.find_last_not_of(" \t\r\n") + 1);
+
+  if (value.empty())
+  {
+    return false;
+  }
+
+  if (std::all_of(value.begin(), value.end(), [](unsigned char c) { return std::isdigit(c); }))
+  {
+    try
```
owent: Could you please use the codes just like `GetTimeoutFromString` in `sdk/src/common/env_variables.cc` to parse duration? The common codes can be moved into `api/include/opentelemetry/common/timestamp.h` .

And exception can be disabled by `-fno-exception` or `/EH` . We should use `OPENTELEMETRY_HAVE_EXCEPTIONS` to check if exception is enabled.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,69 @@
...
+    {
+      return false;
+    }
+  }
+  return false;
+}
+
+bool ParseRetryAfterDate(std::string value, std::chrono::system_clock::time_point &date)
+{
+  value.erase(0, value.find_first_not_of(" \t\r\n"));
```
owent: We can use `common::StringUtil::Trim` to trim string.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,69 @@
...
+  value.erase(value.find_last_not_of(" \t\r\n") + 1);
+
+  std::tm tm = {};
+  std::istringstream ss(value);
+
+  ss >> std::get_time(&tm, "%a, %d %b %Y %H:%M:%S");
```
owent: According to <https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1> seems support much more formats and timezone setting.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,69 @@
...
+  if (!ss.fail())
+  {
+    std::time_t epoch = PortableTimegm(&tm);
```
owent: Why not use `std::mktime` ?

ext/src/http/client/curl/http_operation_curl.cc
DCchoudhury15: You're right that `GetResponseHeaders()` is not ideal here since it builds a full `std::multimap` just to look up a single header.

That said, this block is already inside `if (IsRetryable())`, so it only executes for retryable responses. Successful responses never reach this path.

I'll update it to scan `response_headers_` directly and stop as soon as it finds the `retry-after` header, which avoids the unnecessary allocation and parsing overhead.

ext/src/http/client/curl/http_operation_curl.cc
DCchoudhury15: Good catch. The `try`/`catch` around `std::stoull` is very likely what's triggering the Bazel `noexcept` CI failure.

I'll replace it with a manual digit-by-digit parsing loop that performs explicit overflow checks, following the same approach used by `GetTimeoutFromString` in `sdk/src/common/env_variables.cc`. Where appropriate, I'll keep the exception-related handling guarded with `OPENTELEMETRY_HAVE_EXCEPTIONS`.

I'll also move the shared parsing logic into `api/include/opentelemetry/common/timestamp.h` as you suggested so it can be reused across different call sites.

ext/src/http/client/curl/http_operation_curl.cc
DCchoudhury15: Thanks for catching this. I went back and checked §7.1.1.1 again, and you're right. The RFC defines three valid `HTTP-date` formats:

* `IMF-fixdate`: `Sun, 06 Nov 1994 08:49:37 GMT`
* `RFC 850`: `Sunday, 06-Nov-94 08:49:37 GMT`
* `asctime`: `Sun Nov  6 08:49:37 1994`

Right now I'm only handling the first one, so I'll add support for the other two as well since the RFC requires recipients to accept all three.

I'll also implement the RFC 850 two-digit year handling. If the parsed year ends up looking more than 50 years in the future, I'll map it back to the most recent matching year in the past, as required by the spec.

For the timezone part, I don't think any extra handling is needed. All three formats are defined as UTC the first two explicitly use `GMT`, and the `asctime` format is also specified to be interpreted as UTC. So treating the parsed `std::tm` as UTC with `PortableTimegm` should be the right thing to do here.

ext/src/http/client/curl/http_operation_curl.cc
DCchoudhury15: That's exactly why I avoided `mktime()` here. It interprets the input as **local time**, whereas `HTTP-date` is always defined in UTC, so the result would depend on the machine's timezone setting and would be incorrect anywhere outside UTC.

The usual alternatives have their own drawbacks. Temporarily setting `TZ=UTC` and calling `tzset()` mutates process-wide state, which isn't something we want in this code path since `DoRetrySessions` runs on a background thread. The other option is relying on platform-specific APIs, which brings back the portability issues between POSIX and Windows that we're trying to avoid.

The integer-based conversion avoids all of those problems. It doesn't depend on any OS APIs or global state, requires no platform-specific code, and always produces the correct UTC Unix timestamp regardless of the host's timezone.

ext/src/http/client/curl/http_operation_curl.cc
owent: You're right — all valid timezones in HTTP-date are UTC/GMT, so timezone handling isn't needed. Only the datetime format variations matter.

pellared:
@DCchoudhury15, there are build failures...

api/include/opentelemetry/common/timestamp.h
```diff
@@ -202,5 +210,103 @@ class DurationUtil
...
+    value = StringUtil::Trim(value);
+    if (value.empty())
+    {
+      return false;
+    }
+
+    std::chrono::seconds::rep result = 0;
+    for (const char c : value)
+    {
+      if (!std::isdigit(static_cast<unsigned char>(c)))
```
owent: `std::isdigit` accepts an int as parameter, and this may cause conversation between signed and unsigned integer in some compilers. The CI doesn't appear to be warning about this. Could you please use `if (!std::isdigit(static_cast<int>(static_cast<unsigned char>(c))))` here?

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,55 @@
...
+    }
+
+    static const char kRetryAfterHeader[] = "retry-after:";
+    static const size_t kRetryAfterLen    = sizeof(kRetryAfterHeader) - 1;
+
+    size_t line_len = static_cast<size_t>(line_end - line);
+    if (line_len > kRetryAfterLen)
+    {
+      bool match = true;
+      for (size_t i = 0; i < kRetryAfterLen && match; ++i)
```
owent: You can use `strncmp` instead.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -35,6 +38,55 @@
...
+      {
+        match = (std::tolower(static_cast<unsigned char>(line[i])) == kRetryAfterHeader[i]);
+      }
+
+      if (match)
+      {
+        value = std::string(line + kRetryAfterLen, line_end);
```
owent: Using a `nostd::string_view` to output value will have a better performance.

api/include/opentelemetry/common/timestamp.h
```diff
@@ -202,5 +210,103 @@ class DurationUtil
...
+      }
+      tm.tm_year = full_year - 1900;
+      date       = std::chrono::system_clock::from_time_t(PortableTimegm(&tm));
+      return true;
+    }
+
+    ss.clear();
+    ss.str(str);
+    tm = {};
+    ss >> std::get_time(&tm, "%a %b %d %H:%M:%S %Y");
```
owent: Do we need `ss.imbue(std::locale::classic());` before `std::get_time` ? Or it may return a name not in English.

api/include/opentelemetry/common/timestamp.h
```diff
@@ -202,5 +210,103 @@ class DurationUtil
...
+    ss.clear();
+    ss.str(str);
+    tm = {};
+    ss >> std::get_time(&tm, "%A, %d-%b-%y %H:%M:%S");
+    if (!ss.fail())
+    {
+      std::time_t now_t = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+      int current_year  = 1970 + static_cast<int>(now_t / 31556952);
+      int full_year     = 1900 + tm.tm_year;
+      if (full_year - current_year > 50)
```
owent: The two-digit year handling relies on std::get_time's %y mapping 69–99 to 19xx and 00–68 to 20xx. This split point comes from POSIX strptime and isn't actually guaranteed by the C++ standard, so it's a non-portable assumption — could you add a comment documenting it?
Also, the RFC 7231 rule (section 7.1.1.1) is that a two-digit year more than 50 years in the future MUST be interpreted as the most recent past year with the same last two digits. The current full_year - current_year > 50 check is equivalent for valid inputs, but normalizing via (full_year - current_year + 50) / 100 would express the intent more directly and be robust to any %y mapping.

DCchoudhury15: Thanks @owent! These all make sense to me. I had a couple of follow-up thoughts/questions on each:

- **`std::isdigit` cast:** I agree. Since `std::isdigit` takes an `int`, would it be better to use `static_cast<int>(static_cast<unsigned char>(c))` here? That seems to be the portable way to avoid issues on platforms where `char` is signed.

- **`strncmp`:** Makes sense. The current loop is essentially doing a string comparison manually, so replacing it with `strncmp` should make the code a bit simpler and let the standard library handle it.

- **`nostd::string_view` output parameter:** Good point. Since the value is already stored in `raw_headers`, would returning a `string_view` into that buffer be preferable? That would avoid creating an unnecessary `std::string`, and it looks like the downstream parsing code already accepts `string_view`.

- **`imbue(std::locale::classic())`:** Thanks for pointing this out. If I understand correctly, `std::get_time` uses the stream's locale when parsing month and day names, so on systems using a non-English locale it could fail to parse a valid HTTP date. Using the classic locale would keep the behavior consistent everywhere.

- **Two-digit year handling:** That's a good point as well. I hadn't realized the `69–99` / `00–68` mapping comes from POSIX rather than the C++ standard. Would it make sense to document that assumption and normalize the parsed year using the RFC 7231 rule instead? Something like `(full_year - current_year + 50) / 100` seems to express the intent directly and wouldn't depend on how a particular `%y` implementation maps two-digit years.

owent: - **`imbue(std::locale::classic()):`** Yes, that's what I mean.
- **Two-digit year handling:** It's not in SPEC and it's the behaviour of most implement of `std::get_time` to follow RFC 7231. I think it may change in the future.

owent (approved):
LGTM, thanks.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1513,6 +1573,27 @@ void HttpOperation::PerformCurlMessage(CURLcode code)
...
   {
+
+    nostd::string_view retry_after;
+    if (FindRetryAfterValue(response_headers_, retry_after))
+    {
+      std::chrono::seconds delay;
+
+      if (opentelemetry::common::HttpUtil::ParseDelaySeconds(retry_after, delay))
+      {
+        retry_after_time_point_ = std::chrono::system_clock::now() + delay;
```
marcalff: Do not call `now()` again, reuse `last_attempt_time_` instead.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1513,6 +1573,27 @@ void HttpOperation::PerformCurlMessage(CURLcode code)
...
+      }
+      else
+      {
+        std::chrono::system_clock::time_point date;
+        if (opentelemetry::common::HttpUtil::ParseHttpDate(retry_after, date))
+        {
+          auto now                = std::chrono::system_clock::now();
+          retry_after_time_point_ = (date > now) ? date : now;
```
marcalff: Do not call `now()` again, reuse `last_attempt_time_` instead.

api/include/opentelemetry/common/timestamp.h
```diff
@@ -202,5 +211,105 @@ class DurationUtil
   }
 };
 
+/**
+ * @brief Parsing helpers for RFC 7231 value formats (HTTP-date, section
+ * 7.1.1.1, and delay-seconds, section 7.1.3) that appear in headers such
+ * as Retry-After, Date, Last-Modified, and Expires.
+ */
+class HttpUtil
```
marcalff: This new utility class looks good.

But this is really related to HTTP.

It could be located:
* with the HTTP related code
* eventually, with SDK headers

but this should not be exposed as widely as api/include: there are no reasons for any application instrumentation to depend on it.

Please move this helper as file `http_time_util.h`, in the same location as `http_operation_curl.h`.
Rename class `HttpUtil` to `HttpTimeUtil`.
Move the implementation to http_time_util.cc, same location as http_operation_curl.cc

marcalff (changes requested):
Thanks for the patch.

This looks good overall, but see some code changes in comments.

api/include/opentelemetry/common/timestamp.h
DCchoudhury15: yes i will do so

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -560,6 +614,11 @@ bool HttpOperation::IsRetryable()
 
 std::chrono::system_clock::time_point HttpOperation::NextRetryTime()
 {
+  if (retry_after_time_point_ != std::chrono::system_clock::time_point{})
+  {
+    return retry_after_time_point_;
```
marcalff, line 619: Please check the following comment in `HttpClient::doRetrySessions()`

```
  // Assumptions:
  // - This is a FIFO list so older sessions, pushed at the back, always end up at the front
  // - Locking not required because only the background thread would be pushing to this container
  // - Retry policy is not changed once HTTP client is initialized, so same settings for everyone
```

Sessions are no longer sorted by retry time.

Now that the next retry time is controlled by a -- remote -- server, is there a vulnerability here that a server claiming a retry time in 2 years from now will block the list of pending sessions ?

What mechanism detects sessions with a retry time in 2 years from now (i.e., that exceeds the maximum expected retry delay), to make sure they are closed and can not be leveraged to perform an attack ?

marcalff (dismissed):

marcalff:
Related:

* https://github.com/open-telemetry/opentelemetry-proto/pull/806

marcalff (approved):
LGTM, thanks for the feature.

Issue #4163, opened 2026-06-15
Upgrade CMake third party dependencies


dbarker:
Upgrade third party dependencies 
- gRPC -> [1.81.1](https://github.com/grpc/grpc/releases/tag/v1.81.1)
- protobuf -> [33.6](https://github.com/protocolbuffers/protobuf/releases/tag/v33.6)
- ryml -> [0.15.2](https://github.com/biojppm/rapidyaml/releases/tag/v0.15.2)
- curl -> [8.21.0](https://github.com/curl/curl/releases/tag/curl-8_21_0)

PR #4187, opened 2026-06-26
[CMAKE] Upgrade curl to 8.21.0

dbarker:
Contributes to #4163 

## Changes

- upgrade curl to [8.21.0](https://github.com/curl/curl/releases/tag/curl-8_21_0)
- curl 8.21.0 has a cmake minimum of 3.18 and the otel-cpp minimum is 3.16. This only impacts the FetchContent (and add_subdirectory) use case. Many other dependencies are still at 3.16 (grpc, protobuf, googletest) so bumping the project CMake minimum is not appropriate right now. Instead this PR: 

  - Updates the FetchContent test is updated to install CMake 3.18 
  - Updates the INSTALL.md guide accordingly. 
  
  
### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       CXX_STANDARD: '17'
-      CMAKE_VERSION: '3.16.0'
+      CMAKE_VERSION: '3.18.0' # CMake version 3.18+ is required by curl 8.21.0+
       BUILD_TYPE: 'Release'
     steps:
     - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       CXX_STANDARD: '17'
-      CMAKE_VERSION: '3.16.0'
+      CMAKE_VERSION: '3.18.0' # CMake version 3.18+ is required by curl 8.21.0+
```
ThomsonTan, line 61: Does the cmake in the CMakeLists.txt under repo root need to be updated too? It is still 3.16.

```cmake
cmake_minimum_required(VERSION 3.16)
```

dbarker: I don't think we should right now because curl is optional and many of the other dependencies are still on 3.16 (protobuf, grpc, googletest, ...).

marcalff (approved):

ThomsonTan (approved):

PR #4188, opened 2026-06-28
[Metrics SDK] Implement configurable cardinality limit

### LINKED ISSUE #3292: [Metrics SDK] Make cardinality limit configurable
As per the specs, there are 3 ways to [configure](https://github.com/open-telemetry/opentelemetry-specification/blob/0c853fe71aad74fa8d26c76c3899e87e594a9f5f/specification/metrics/sdk.md#configuration-1) cardinality limit:

> The cardinality limit for an aggregation is defined in one of three ways:
> 
> - A [view](https://github.com/open-telemetry/opentelemetry-specification/blob/0c853fe71aad74fa8d26c76c3899e87e594a9f5f/specification/metrics/sdk.md#view) with criteria matching the instrument an aggregation is created for has an aggregation_cardinality_limit value defined for the stream, that value SHOULD be used.
> - If there is no matching view, but the MetricReader defines a default cardinality limit value based on the instrument an aggregation is created for, that value SHOULD be used.
> - If none of the previous values are defined, the default value of 2000 SHOULD be used.

The implementation currently imposes the limit to be [2000](https://github.com/open-telemetry/opentelemetry-cpp/blob/edfeabe4cefbec2ba3697e41664e76f8bfcee52c/sdk/include/opentelemetry/sdk/metrics/state/attributes_hashmap.h#L29). And the measurement beyond this would be aggregated as part of overflow attribute.

om7057:
Fixes #3292

## Changes

Implements configurable aggregation cardinality limits at two priority levels
as specified by the OpenTelemetry Metrics SDK specification:

- **View-level**: Cardinality limit defined in a View's `AggregationConfig` (enforced today)
- **MetricReader-level**: Per-instrument-type limits stored on the reader via `CardinalityLimits`
  struct (parsed and accessible via `GetCardinalityLimit()`; per-collector enforcement
  is a follow-up, see TODO in `SetCardinalityLimits()`)
- **SDK default**: `kDefaultCardinalityLimit = 2000` used when neither above is set

## Implementation Details

**`cardinality_limits.h`** (new):
- `CardinalityLimits` struct with per-instrument-type fields, all defaulting to `kDefaultCardinalityLimit`
- Namespace-level `constexpr kDefaultCardinalityLimit = 2000`

**`MetricReader` changes:**
- Added `GetCardinalityLimit(InstrumentType)` method
- Added `SetCardinalityLimits(const CardinalityLimits &)` method

**`CollectorHandle` / `MetricCollector` changes:**
- Added `GetCardinalityLimit()` virtual method to `CollectorHandle` (default returns `kDefaultCardinalityLimit`)
- `MetricCollector` delegates to its underlying `MetricReader`

**Configuration changes:**
- `sdk_builder.cc`: Wires `cardinality_limits` from YAML config via `ToCardinalityLimits()` helper
- `cardinality_limits_configuration.h`: Initialized all fields (`default_limit = 2000`, others `= 0`)
- Removed "cardinality limits not supported" warnings

## Checklist

- [x] `CHANGELOG.md` updated for non-trivial changes
- [x] Unit tests have been added
- [x] Changes in public API reviewed


om7057:
Hi @dbarker 
I've implemented configurable cardinality limits for the Metrics SDK to resolve issue #3292 and bring the C++ implementation into compliance with the OpenTelemetry specification.

**Summary of Changes:**
- Added three-level cardinality limit configuration (View > Reader > SDK default)
- Wired existing YAML config infrastructure to actually apply the limits
- Ensured proper lifetime management for reader-level configs

This is my first contribution to OpenTelemetry C++. I'd appreciate guidance on:
1. Whether the approach aligns with the project's design principles
2. If additional tests beyond the existing YAML parsing tests are needed
3. If I should update the CHANGELOG.md

Looking forward to your feedback!

cc: @lalitb

marcalff (dismissed):
Thanks for the contribution.

There is something wrong the way clang-format was applied, causing spurious changes in many files.

This:

- is most likely incorrect, and will fail CI on clang-format
- makes it difficult to review the real changes from this patch.

When formatting code, a precise version of clang-format must be used. Applying a different version will cause this. The best is to use the dev container to perform formatting, as it will use the exact same version as the github CI.

In any cases, adding a file in a commit that was not changed voluntarily (i.e., with a real fix) is a red flag, and should not happen.

Please rework the patch to remove the noise, so it can be reviewed.

om7057:
@marcalff Thank you for the feedback! I've reworked the PR to remove all spurious formatting changes.

Current state
Files modified:

metric_reader.h
 - Added CardinalityLimitOptions
metric_reader.cc
 - Implemented getter/setter
metric_collector.h
 - Added interface method
metric_collector.cc
 - Implemented delegation
meter_context.h
 - Added reader query method
meter_context.cc
 - Implemented max-across-readers logic
sync_metric_storage.h
 - Added lifetime management
async_metric_storage.h
 - Added lifetime management
meter.cc
 - Applied reader-level fallback
sdk_builder.cc
 - Wired YAML config to implementation

Regarding CI: the only failing check is W3C Distributed Tracing Validation V1 (test_tracestate_key_illegal_vendor_format and test_tracestate_key_length_limit). This failure is unrelated to this PR — the changes here are entirely in the Metrics SDK and do not touch any trace context code. The W3C V1 test suite checks out the latest [w3c/trace-context]HEAD at runtime, and these two tests appear to be failing intermittently across PRs.

The PR is now clean and ready for review. Please let me know if you need any clarifications on the implementation approach!

sdk/src/metrics/meter.cc
```diff
@@ -542,14 +544,34 @@ std::unique_ptr<SyncWritableMetricStorage> Meter::RegisterSyncMetricStorage(
...
+          const AggregationConfig *effective_config = view.GetAggregationConfig();
+          std::shared_ptr<AggregationConfig> reader_config;
+
+          // If the view does not specify a cardinality limit, use the MetricReader-level default
+          if (!effective_config)
+          {
+            auto ctx_ptr = meter_context_.lock();
+            if (ctx_ptr)
+            {
+              size_t reader_limit = ctx_ptr->GetReaderCardinalityLimit(instrument_descriptor.type_);
```
lalitb: I think this breaks per-reader semantics. Example: `reader A` has limit 10, `reader B` has limit 1000. The storage uses 1000, so `reader A` can export far more series than configured. This is also a heap/memory regression for low-limit readers.

lalitb: One possible fix is to avoid resolving reader-level limits into a single storage-level config. The view-level limit can stay on the shared stream/storage, but the MetricReader fallback should be applied per collector/reader, since each reader may have a different configured limit. So instead of taking the max across readers, the collection path likely needs to use the current `CollectorHandle`'s `GetCardinalityLimit(instrument_type)` when no view-level limit exists.

om7057: Good catch @lalitb, thanks! You're right, resolving the reader-level fallback into shared storage breaks the per-reader semantics you described.

I've removed that logic here and reverted the storage constructors. For now, this PR only enforces view-level `aggregation_cardinality_limit` (plus the existing SDK default of 2000). Reader-level limits are still parsed, but not enforced.

I agree the right place to apply the `MetricReader` fallback is in the collection path on a per-`CollectorHandle` basis when no view-level limit exists. Since that's a more involved change, I'd rather tackle it in a follow-up PR.

Also fixed the IWYU warning in the latest commit.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -16,6 +17,22 @@ namespace sdk
 {
 namespace metrics
 {
+/**
+ * Configuration options for cardinality limits per instrument type.
+ * A value of 0 means use the SDK default (2000).
+ */
+struct CardinalityLimitOptions
```
dbarker: nit - consider `CardinalityLimits` for the name to follow the `LogRecordLimits` and `SpanLimits` struct names.

sdk/src/metrics/state/metric_collector.cc
```diff
@@ -45,6 +46,11 @@ AggregationTemporality MetricCollector::GetAggregationTemporality(
   return aggregation_temporality;
 }
 
+size_t MetricCollector::GetCardinalityLimit(InstrumentType instrument_type) noexcept
```
dbarker: nit - missing `std::`

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -16,6 +17,22 @@ namespace sdk
...
+{
+  size_t default_limit              = 0;  // 0 means use SDK default (2000)
```
dbarker: can we define and use the default here for each? Setting to zero seems unexpected.

sdk/src/metrics/metric_reader.cc
```diff
@@ -84,6 +90,45 @@ bool MetricReader::IsShutdown() const noexcept
...
+  switch (instrument_type)
+  {
+    case InstrumentType::kCounter:
+      return cardinality_limit_options_.counter != 0 ? cardinality_limit_options_.counter
+                                                     : cardinality_limit_options_.default_limit;
+    case InstrumentType::kHistogram:
+      return cardinality_limit_options_.histogram != 0 ? cardinality_limit_options_.histogram
+                                                       : cardinality_limit_options_.default_limit;
+    case InstrumentType::kUpDownCounter:
+      return cardinality_limit_options_.up_down_counter != 0
```
dbarker: can these just return the value already set to a default or overridden before being set to the reader to avoid the zero value special case?

sdk/test/metrics/metric_collector_test.cc
```diff
@@ -402,6 +402,95 @@ TEST_F(MetricCollectorTest, CollectWithMetricFilterTestAttributesTest2)
   }
 }
 
+TEST_F(MetricCollectorTest, CardinalityLimitDelegation)
```
dbarker, line 413: Are there tests to verify streams are dropped when the cardinally limit set through the reader is reached?

dbarker (commented):
Thanks for the contribution. Please see some minor feedback and questions below.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Sure. Renamed to CardinalityLimits to match SpanLimits / LogRecordLimits convention. Updated across all files.

sdk/src/metrics/state/metric_collector.cc
om7057: Fixed. Changed to std::size_t in metric_collector.cc and consistently throughout the new struct and method signatures.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Yes. All fields in CardinalityLimits now default to kAggregationCardinalityLimit (2000), the constant already used throughout the SDK. This also lets us remove the zero-check branching in GetCardinalityLimit() entirely; it's now a plain switch that returns the field directly.

sdk/test/metrics/metric_collector_test.cc
om7057: Added ReaderCardinalityLimitEnforcedOnCollection in metric_collector_test.cc. It sets a cardinality limit of 5 via a View-level AggregationConfig, records 10 unique attribute sets through the meter, then asserts via Produce() that total data points don't exceed the limit and the otel.metric.overflow sentinel point is present

sdk/src/metrics/metric_reader.cc
om7057: Covered by the same change as above. Since all CardinalityLimits fields now default to kAggregationCardinalityLimit (2000), there's no zero to check for. GetCardinalityLimit() is now a plain switch that just returns the field directly with no ternary or branching logic.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [METRICS SDK] Implement configurable cardinality limit at three priority levels
```
ThomsonTan: Should this be flagged as a breaking change?

sdk/src/metrics/metric_reader.cc
dbarker: I can see the CardinaltiyLimitConfiguration class uses 0 to indicate a value is not set. I missed that before.  The conversion from the declaritive config struct to the new limits struct needs to account for this. 

A user may set their own default (say 1500) and change the instrument level limits to any value (which may be the spec default of 2000), this needs to be tested.

CHANGELOG.md
om7057: Hi @ThomsonTan 
This change is purely additive. It doesn't modify any existing APIs or change default behavior, and the new virtual methods all provide default implementations, so existing code continues to work unchanged unless it explicitly opts into the new functionality.

I've also updated the CHANGELOG entry to make the compatibility story explicit.

sdk/src/metrics/metric_reader.cc
om7057: Good catch, thanks!

You're right, `CardinalityLimitsConfiguration` uses `0` to represent "unset", but we were propagating those values directly into `CardinalityLimits`, where the intended default is `2000`.

I've updated `sdk_builder.cc` so the defaults are resolved during construction: if a configuration value is `0`, we fall back to the default limit; otherwise we use the configured value.

This also keeps `SetCardinalityLimits` as a straightforward setter with no special handling for sentinel values, which means an explicit value of `2000` is treated just like any other valid configuration.

I also added a test covering that exact case to make sure it behaves as expected.

dbarker: Thanks that sounds reasonable. The CardinalityLimitsConfiguration class does not initialize the limits so this can be UB if they are accessed uninitialized. I've patched it in #4240, but feel free to fix in this PR.

sdk/include/opentelemetry/sdk/configuration/cardinality_limits_configuration.h
```diff
@@ -18,15 +18,15 @@ namespace configuration
 class CardinalityLimitsConfiguration
 {
 public:
-  std::size_t default_limit;
+  std::size_t default_limit{0};
```
dbarker: This default_limit should be initialized to the spec default of 2000. The rest are correctly initialized to 0.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -16,6 +17,23 @@ namespace sdk
 {
 namespace metrics
 {
+/**
+ * Cardinality limits per instrument type for a MetricReader.
+ * Each field defaults to kAggregationCardinalityLimit (2000).
+ * Set a field to override the limit for that instrument type.
+ */
+struct CardinalityLimits
```
dbarker: Please move the CardinalityLimits struct to its own header file. This way other components can access the const spec default value without including the reader.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -16,6 +17,23 @@ namespace sdk
...
+{
+  std::size_t default_limit              = 2000;
```
dbarker: The spec default value should be defined once. Consider a constexpr kDefaultLimit in this struct that can be used to initialize the configurable member limits.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -1498,7 +1515,21 @@ std::unique_ptr<opentelemetry::sdk::metrics::MetricReader> SdkBuilder::CreatePul
 
   if (model->cardinality_limits != nullptr)
   {
-    OTEL_INTERNAL_LOG_WARN("cardinality limits not supported, ignoring");
+    opentelemetry::sdk::metrics::CardinalityLimits cardinality_limits;
```
dbarker: this is a duplicate from periodic reader section. Please create a function in the anonamous namespace to perform the conversion from the config cardinality limits model to the sdk limits struct.

sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
```diff
@@ -36,6 +37,19 @@ class CollectorHandle
...
+   * Get the cardinality limit for a given instrument type.
+   * Returns kAggregationCardinalityLimit (2000) by default.
+   *
+   * @param instrument_type The instrument type to get the limit for
+   * @return The cardinality limit for the requested instrument type
+   */
+  virtual std::size_t GetCardinalityLimit(InstrumentType instrument_type) noexcept
+  {
+    (void)instrument_type;
+    return 2000;  // kAggregationCardinalityLimit
```
dbarker: this should return a constexpr spec default defined in the sdk CardinalityLimits struct.

sdk/include/opentelemetry/sdk/metrics/meter_context.h
```diff
@@ -108,6 +110,15 @@ class MeterContext : public std::enable_shared_from_this<MeterContext>
...
   nostd::span<std::shared_ptr<CollectorHandle>> GetCollectors() noexcept;
 
+  /**
+   * Get the cardinality limit for a given instrument type from configured readers.
+   * Returns the maximum limit across all readers, or 0 if no reader specifies a limit.
+   *
+   * @param instrument_type The instrument type to get the limit for
+   * @return The maximum cardinality limit, or 0 if not configured
+   */
+  std::size_t GetReaderCardinalityLimit(InstrumentType instrument_type) noexcept;
```
dbarker: This method appears to be only used in tests and returning the max limit of all collectors may not be useful outside of those specific tests. If the GetCollectors method above can be used to access the value from each reader and that meets the testing use case then please remove this method.

sdk/test/metrics/metric_collector_test.cc
```diff
@@ -402,6 +408,175 @@ TEST_F(MetricCollectorTest, CollectWithMetricFilterTestAttributesTest2)
...
+  EXPECT_EQ(collector->GetCardinalityLimit(InstrumentType::kUpDownCounter), 1000);
+  EXPECT_EQ(collector->GetCardinalityLimit(InstrumentType::kObservableCounter), 1000);
+}
+
+TEST_F(MetricCollectorTest, MeterContextMaxCardinalityLimit)
+{
+  auto context = std::shared_ptr<MeterContext>(new MeterContext(ViewRegistryFactory::Create()));
+
+  // Test with no readers — should return 0
+  EXPECT_EQ(context->GetReaderCardinalityLimit(InstrumentType::kCounter), 0);
```
dbarker: why zero? Shouldn't this be the spec default 2000 ?

sdk/test/metrics/metric_collector_test.cc
```diff
@@ -402,6 +408,175 @@ TEST_F(MetricCollectorTest, CollectWithMetricFilterTestAttributesTest2)
...
+  limits2.histogram = 200;   // lower than reader1
+  reader2->SetCardinalityLimits(limits2);
+  AddMetricReaderToMeterContext(context, reader2);
+
+  // Should return the max across both readers
+  EXPECT_EQ(context->GetReaderCardinalityLimit(InstrumentType::kCounter), 1000);   // max(100, 1000)
+  EXPECT_EQ(context->GetReaderCardinalityLimit(InstrumentType::kHistogram), 500);  // max(500, 200)
+}
+
+TEST_F(MetricCollectorTest, MeterContextCardinalityLimitWithMultipleReaders)
```
dbarker, line 475: This `MeterContextCardinalityLimitWithMultipleReaders` and `MeterContextMaxCardinalityLimit` seem to cover the same case of setting cardinality with multiple readers. Is getting the max limit for the set of readers a spec requirement or enabling a particular use case?

dbarker (commented):
Thanks for the updates. Please see feedback below. Approval to follow.

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ Increment the:
...
```
dbarker: Please update the changelog entry based on https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#discussion_r3500154409
As is this PR does not fully support reader level limits being applied as a fallback from the views.

sdk/include/opentelemetry/sdk/configuration/cardinality_limits_configuration.h
om7057: Sure, CardinalityLimitsConfiguration::default_limit is now {2000}. The per-instrument fields remain {0}.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Yes moved to 
cardinality_limits.h with a constexpr kDefaultCardinalityLimit = 2000 static member. metric_reader.h and metric_collector.h now include it directly. Other components can access the spec default constant without including the reader.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Yes now CardinalityLimits::kDefaultCardinalityLimit is a static constexpr std::size_t in cardinality_limits.h. All struct fields are initialized to it, so the literal 2000 appears exactly once.

sdk/src/configuration/sdk_builder.cc
om7057: Added ToCardinalityLimits(const CardinalityLimitsConfiguration &) in the anonymous namespace of sdk_builder.cc. Both CreatePeriodicMetricReader and CreatePullMetricReader now call it instead of repeating the block.

sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
om7057: CollectorHandle::GetCardinalityLimit() now returns CardinalityLimits::kDefaultCardinalityLimit instead of the magic literal 2000.

sdk/include/opentelemetry/sdk/metrics/meter_context.h
om7057: Removed from meter_context.h and meter_context.cc. The tests now use GetCollectors() directly to inspect per-reader limits, which is the right access pattern.

sdk/test/metrics/metric_collector_test.cc
om7057: Fair point. The "no readers → 0" expectation was testing the now-removed GetReaderCardinalityLimit method, which aggregated across readers and returned 0 when none were configured. That method is gone. The test has been rewritten to use GetCollectors(), with no readers, GetCollectors().size() == 0, which is the correct check.

sdk/test/metrics/metric_collector_test.cc
om7057: You're right that the "max across readers" semantics aren't a spec requirement, getting the max was an implementation choice that turned out to be incorrect (per @lalitb's earlier feedback). Both tests have been rewritten to use GetCollectors() directly. MeterContextMaxCardinalityLimit now verifies per-reader limits are stored and retrievable via individual collectors; MeterContextCardinalityLimitWithMultipleReaders covers the multi-reader scenario with distinct per-instrument overrides. They test different configurations rather than the same aggregation logic.

CHANGELOG.md
om7057: Updated. The entry now reads:

"Implements configurable cardinality limits at view-level and reader-level via CardinalityLimits struct and SetCardinalityLimits()/GetCardinalityLimit() on MetricReader. Backwards-compatible addition; existing behavior is unchanged."

This accurately reflects what the PR delivers: the reader-level limit is parsed and stored on the reader and is accessible per-collector via GetCardinalityLimit(). Applying the reader limit as a per-collector fallback during the collection path (when no view-level limit exists) is the follow-up work noted by @lalitb and is intentionally out of scope for this PR.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -45,6 +47,23 @@ class MetricReader
...
+   *
+   * @param instrument_type The instrument type to get the limit for
+   * @return The cardinality limit for the requested instrument type
+   */
+  std::size_t GetCardinalityLimit(InstrumentType instrument_type) const noexcept;
+
+  /**
+   * Set per-instrument-type cardinality limits for this reader.
+   *
+   * @param limits The cardinality limits to apply
```
dbarker, line 68: Since this reader level limits won't be enforced yet please add a TODO comment here noting that.

sdk/src/metrics/metric_reader.cc
```diff
@@ -84,6 +85,34 @@ bool MetricReader::IsShutdown() const noexcept
...
+    case InstrumentType::kGauge:
+      return cardinality_limits_.gauge;
+    default:
+      return cardinality_limits_.default_limit;
+  }
+}
+
+void MetricReader::SetCardinalityLimits(const CardinalityLimits &limits) noexcept
+{
+  cardinality_limits_ = limits;
```
dbarker, line 113: Consider logging a warning message here noting that reader level cardinality limits are not enforced yet.

sdk/test/metrics/metric_collector_test.cc
```diff
@@ -402,6 +408,190 @@ TEST_F(MetricCollectorTest, CollectWithMetricFilterTestAttributesTest2)
...
+  // reader3: counter=200, histogram=400, others=kDefaultCardinalityLimit (2000)
+  EXPECT_EQ(collectors[2]->GetCardinalityLimit(InstrumentType::kCounter), 200);
+  EXPECT_EQ(collectors[2]->GetCardinalityLimit(InstrumentType::kHistogram), 400);
+  EXPECT_EQ(collectors[2]->GetCardinalityLimit(InstrumentType::kUpDownCounter),
+            kDefaultCardinalityLimit);
+}
+
+// Verify that when a reader-level cardinality limit is set, streams exceeding
+// the limit are dropped and the overflow attribute point is produced instead.
+TEST_F(MetricCollectorTest, ReaderCardinalityLimitEnforcedOnCollection)
```
dbarker: Consider renaming this test since it covers setting cardinality limits via a view and not the reader.

dbarker (approved):
Thanks for the PR! Approved with some feedback below.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Added a TODO in the doc comment: "Reader-level limits are stored but not yet enforced as a per-collector fallback during the collection path. Enforcement will be added in a follow-up. View-level limits (via AggregationConfig) are enforced today."

sdk/src/metrics/metric_reader.cc
om7057: SetCardinalityLimits now logs an OTEL_INTERNAL_LOG_WARN noting that the limits are stored but not yet enforced during collection, and directs users to the view-level AggregationConfig for today's enforcement path.

sdk/test/metrics/metric_collector_test.cc
om7057: Renamed to ViewCardinalityLimitEnforcedOnCollection since the test sets the limit via a View's AggregationConfig, not via the reader. Updated the comment to match.

om7057:
> 
> There is something wrong the way clang-format was applied, causing spurious changes in many files.
> 
> This:
> 
> * is most likely incorrect, and will fail CI on clang-format
> * makes it difficult to review the real changes from this patch.
> 
> When formatting code, a precise version of clang-format must be used. Applying a different version will cause this. The best is to use the dev container to perform formatting, as it will use the exact same version as the github CI.
> 
> In any cases, adding a file in a commit that was not changed voluntarily (i.e., with a real fix) is a red flag, and should not happen.
> 
> Please rework the patch to remove the noise, so it can be reviewed.

@marcalff please approve the PR if all the relevant changes are being addressed according to you.

om7057:
> Thanks for the PR! Approved with some feedback below.

Thank you for the approval. 
Is the PR ready for merge?

dbarker:
> Is the PR ready for merge?

The open change request might need to be cleared first.  

@ThomsonTan and @lalitb please also take a look since you have open comments.

om7057:
> > Is the PR ready for merge?
> 
> The open change request might need to be cleared first.
> 
> @ThomsonTan and @lalitb please also take a look since you have open comments.

I've resolved the open change request, @dbarker.
So I request @ThomsonTan, @lalitb & @marcalff to go through the PR once and approve if all their changes and comments have been addressed.

ThomsonTan:
@om7057 thanks for the update. Could you please also resolve the merge conflict?

om7057:
> @om7057 thanks for the update. Could you please also resolve the merge conflict?

@ThomsonTan Sure, I've resolved it. Please check and approve if it looks good.

sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
```diff
@@ -36,6 +38,19 @@ class CollectorHandle
...
+   * Returns kAggregationCardinalityLimit (2000) by default.
+   *
+   * @param instrument_type The instrument type to get the limit for
+   * @return The cardinality limit for the requested instrument type
+   */
+  virtual std::size_t GetCardinalityLimit(InstrumentType instrument_type) noexcept
+  {
+    (void)instrument_type;
+    return kDefaultCardinalityLimit;
+  }
```
marcalff: Could this be a pure virtual method instead ?

Also, make it const ?

marcalff (commented):
LGTM, see minor comments.

sdk/include/opentelemetry/sdk/metrics/state/metric_collector.h
om7057: Sure, I believe both can be implemented.
GetCardinalityLimit is now a pure virtual const method on CollectorHandle. MetricCollector overrides it as const noexcept override, and MockCollectorHandle in common.h provides the implementation returning kDefaultCardinalityLimit. The default non-virtual body is gone.
Please review the changes @marcalff and if you feel PR is ready to merge from your end, I'll appreciate the approval.

om7057:
Hi @marcalff @lalitb @dbarker, all the CI checks have been passed, and the PR is approved as well. If you feel it's ready, please consider merging.
Thank you.

om7057:
@marcalff All the CI checks have been passed after taking the latest merge of main, could you please proceed with the PR merge.
Thank you!

marcalff (approved):

om7057:
> LGTM, thanks for the contribution.

Thank you so much for the feedback & merge.
I really appreciate the guidance @marcalff @dbarker @ThomsonTan @lalitb !

sdk/src/configuration/sdk_builder.cc
```diff
@@ -1700,6 +1736,20 @@ void SdkBuilder::AddView(
...
+  // Apply aggregation_cardinality_limit from the view stream configuration
+  if (stream->aggregation_cardinality_limit != 0)
+  {
+    if (sdk_aggregation_config)
+    {
+      sdk_aggregation_config->cardinality_limit_ = stream->aggregation_cardinality_limit;
+    }
+    else
+    {
+      sdk_aggregation_config = std::make_shared<opentelemetry::sdk::metrics::AggregationConfig>(
```
lalitb, line 1748: I think this breaks for histograms. When only `aggregation_cardinality_limit` is set, we create a regular `AggregationConfig `here. But the view registry expects  a histogram config for histogram instruments, so it rejects the whole view. Can we create the config based on the instrument’s default aggregation type?

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
```diff
@@ -45,6 +48,27 @@ class MetricReader
...
+   *
+   * @param instrument_type The instrument type to get the limit for
+   * @return The cardinality limit for the requested instrument type
+   */
+  std::size_t GetCardinalityLimit(InstrumentType instrument_type) const noexcept;
+
+  /**
+   * Set per-instrument-type cardinality limits for this reader.
+   *
+   * TODO: Reader-level limits are stored but not yet enforced as a per-collector
```
lalitb, line 64: It looks like reader-level limits are only stored, but never actually applied during collection. So if someone configures a reader limit of 10, the reader can still export more than 10 series. Should we really expose this configuration before it is enforced?

lalitb:
@om7057  - Since this PR is merged, could you please send a follow-up PR before the next release to fix the histogram case and add a test for a histogram view that only sets `aggregation_cardinality_limit`? Right now that view can be rejected and the configured limit won’t be applied - as mentioned [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#discussion_r3654993146).

om7057:
> @om7057 - Since this PR is merged, could you please send a follow-up PR before the next release to fix the histogram case and add a test for a histogram view that only sets `aggregation_cardinality_limit`? Right now that view can be rejected and the configured limit won’t be applied - as mentioned [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#discussion_r3654993146).

Sure: follow-up PR is up: #4314. It fixes the histogram-view rejection and adds SdkBuilder.AddViewHistogramCardinalityLimitOnly (fails against the pre-fix code with the exact type-mismatch rejection you saw) plus an AddViewCounterCardinalityLimitOnly regression guard for the non-histogram path. Reader-level enforcement (comment above) is still open as a separate, larger follow-up: let me know if you'd like that prioritized before the next release too.

sdk/include/opentelemetry/sdk/metrics/metric_reader.h
om7057: Yes, I do agree on this one: reader-level limits are currently just plumbing with no enforcement, so a configured limit of 10 doesn't actually cap anything yet. I called this out as a known follow-up in the original PR description; I'd like to tackle enforcement in a separate PR since it touches the collection path per CollectorHandle, which is a bigger change than the histogram fix. Left the TODO in place for now to track it; happy to prioritize it if you'd rather not ship the config surface unenforced in the meantime.

sdk/src/configuration/sdk_builder.cc
om7057: Apologies for missing histograms. I just confirmed that this rejects the view for any histogram/base2-exponential-histogram instrument when only aggregation_cardinality_limit is set. Fixed in #4314: AddView() now resolves the instrument's effective aggregation type (falling back to the instrument-derived default, same as ViewRegistry does) and builds the matching AggregationConfig subtype before applying the limit. Added a test that reproduces the rejection against the old code and verifies the fix.

Issue #4196, opened 2026-06-30
[CODE HEALTH] clang-tidy `misc-use-internal-linkage` warnings part 2


dbarker:
This issue covers resolving the misc-use-internal-linkage warnings: 

<details><summary><b>misc-use-internal-linkage (95 warnings)</b><i> - Click to expand </i></summary>
----

**misc-use-internal-linkage** (95 warnings)

| File | Line | Message |
|---|---|---|
| `opentelemetry-cpp/api/test/baggage/propagation/baggage_propagator_test.cc` | 23 | class 'BaggageCarrierTest' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/context/propagation/composite_propagator_test.cc` | 41 | class 'TextMapCarrierTest' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/context/propagation/composite_propagator_test.cc` | 61 | class 'CompositePropagatorTest' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/context/propagation/environment_carrier_test.cc` | 59 | class 'EnvironmentCarrierTest' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 264 | class 'TestLogger' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 460 | class 'TestProvider' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/logs/provider_test.cc` | 31 | class 'TestProvider' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/nostd/shared_ptr_test.cc` | 16 | class 'A' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/nostd/shared_ptr_test.cc` | 32 | class 'B' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/nostd/shared_ptr_test.cc` | 38 | class 'C' can be moved into an anonymous namespace to enforce internal linkage |
| `opentelemetry-cpp/api/test/nostd/shared_ptr_test.cc` | 50 | class 'D' can be moved into an anonymous namespace to enforce internal linkage |
[...]

PR #4199, opened 2026-06-30
[CODE HEALTH] Move trace and baggage propagation test classes into anonymous namespace

mateenali66:
Moves the `TextMapCarrierTest` helper in the b3, http_text_format and jaeger trace propagation tests, and `BaggageCarrierTest` in the baggage propagation test, into anonymous namespaces, resolving their `misc-use-internal-linkage` clang-tidy warnings. Test-only, no behavior change.

Lowered the clang-tidy `warning_limit` by 4 for both ABI presets accordingly.

Part of #4196.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 328
+            warning_limit: 324
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 338
+            warning_limit: 334
     env:
       CC: /usr/bin/clang-22
       CXX: /usr/bin/clang++-22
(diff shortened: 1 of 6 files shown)


thc1006:
These look clean. One cross-PR heads-up since #4199 and #4200 overlap on the ratchet file: both lower `warning_limit` in `clang-tidy.yaml` from the same base (328 abiv1, 338 abiv2) while touching disjoint test files (#4199 removes 4 warnings, #4200 removes 3). So whichever merges first, the other will hit a merge conflict on those two lines and need a quick rebase. After both land the combined value should be 321 and 331 (328 and 338, each minus 7). Flagging it so the second one is not accidentally left at its single-PR number (324/334 or 325/335), which would leave the ceiling looser than the actual count.

dbarker (approved):
Thanks!

marcalff (approved):
LGTM

PR #4211, opened 2026-07-03
[CMAKE] Remove environment variable support from the add external components script

### LINKED ISSUE #3409: Harden the build external components feature
Discussed in the SIG meeting on 5/14/25
- The feature to build external components provided by [opentelemetry-build-external-component.cmake](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/cmake/opentelemetry-build-external-component.cmake) supports environment variables and fetching from URLs now. 
   - if the `OPENTELEMETRY_EXTERNAL_COMPONENT_PATH` or `OPENTELEMETRY_EXTERNAL_COMPONENT_URL` environment variables are set then no cmake changes are needed to include an external component in the build.
- We discussed removing the environment variables so the feature requires a cmake parameter `OPENTELEMETRY_EXTERNAL_COMPONENT_PATH` to be set in order to function.

dbarker:
Fixes #3409 

## Changes

- Removes OPENTELEMETRY_EXTERNAL_COMPONENT_PATH and OPENTELEMETRY_EXTERNAL_COMPONENT_URL environment variable support from the cmake/opentelemetry-build-external-component.cmake script


owent (approved):

cmake/opentelemetry-build-external-component.cmake
```diff
@@ -13,37 +13,16 @@ endfunction()
...
-# component as env variable. Multiple paths can be set by separating them with.
-#  - OPENTELEMETRY_EXTERNAL_COMPONENT_URL Setting github-repo of external component
-#  as env variable
+# component. Multiple paths can be set by separating them with.
 
-# Add custom vendor component from local path, or GitHub repo
-# Prefer CMake option, then env variable, then URL.
+# Add custom vendor component from local path
+# CMake option.
 if(OPENTELEMETRY_EXTERNAL_COMPONENT_PATH)
```
dbarker, line 20: @ThomsonTan can you take a review pass? 

`OPENTELEMETRY_EXTERNAL_COMPONENT_PATH` CMake variable is used by the `opentelemetry-cpp` vckpg port, but the environment variables for `OPENTELEMETRY_EXTERNAL_COMPONENT_PATH` and `OPENTELEMETRY_EXTERNAL_COMPONENT_URL` appear to be unused.

cmake/opentelemetry-build-external-component.cmake
```diff
@@ -13,37 +13,16 @@ endfunction()
...
   endforeach()
-elseif(DEFINED ENV{OPENTELEMETRY_EXTERNAL_COMPONENT_PATH})
-  # Add custom component path to build tree and consolidate binary artifacts in
-  # current project binary output directory.
-  set(OPENTELEMETRY_EXTERNAL_COMPONENT_PATH_VAR $ENV{OPENTELEMETRY_EXTERNAL_COMPONENT_PATH})
-  foreach(DIR IN LISTS OPENTELEMETRY_EXTERNAL_COMPONENT_PATH_VAR)
-    get_directory_name_in_path(${DIR} EXTERNAL_EXPORTER_DIR_NAME)
-    add_subdirectory(${DIR} ${PROJECT_BINARY_DIR}/external/${EXTERNAL_EXPORTER_DIR_NAME})
-  endforeach()
-elseif(DEFINED $ENV{OPENTELEMETRY_EXTERNAL_COMPONENT_URL})
```
lalitb, line 37: Seems we removed the whole `OPENTELEMETRY_EXTERNAL_COMPONENT_URL` path, and there is no `-DOPENTELEMETRY_EXTERNAL_COMPONENT_URL=...` replacement. Is this intentional ?

cmake/opentelemetry-build-external-component.cmake
lalitb: Will leave it to @ThomsonTan  - he knows the vcpkg better :)

cmake/opentelemetry-build-external-component.cmake
dbarker: It is intentional. The URL path had a bug. The if(DEFINED $ENV ...) should be ENV without the $ . It currently evaluates if the value of the environment variable is a defined CMake variable and would not execute the fetch. 

The use case for this URL path is also not clear. The otel-cpp CMake build now fully supports using FetchContent to build an external component in tree and is a tested use case in CI. Rather than trying to support this approach of otel-cpp fetching an external component,  we should promote the external component fetching otel-cpp.  

I'm generally concerned about the maintenance of this approach to build external components from a CMake build maintenance perspective. It tightly couples external project CMake files to internal otel-cpp CMake scripts.

cmake/opentelemetry-build-external-component.cmake
ThomsonTan: The OPENTELEMETRY_EXTERNAL_COMPONENT_PATH variable is also used by the contrib CI/build. The env var and URL were added to help similar scenario, but only the CMake variable is required now. Good to me to remove others.

cmake/opentelemetry-build-external-component.cmake
```diff
@@ -13,37 +13,16 @@ endfunction()
 # Enable building external components through otel-cpp build
 # The config options are
 #  - OPENTELEMETRY_EXTERNAL_COMPONENT_PATH - Setting local paths of the external
-# component as env variable. Multiple paths can be set by separating them with.
-#  - OPENTELEMETRY_EXTERNAL_COMPONENT_URL Setting github-repo of external component
-#  as env variable
+# component. Multiple paths can be set by separating them with.
```
ThomsonTan:
```suggestion
# component. Multiple paths can be set by separating them with semicolons.
```

ThomsonTan (approved):
Looks good to me with a nit.

PR #4216, opened 2026-07-06
[CODE HEALTH] Fix clang-tidy bugprone-unchecked-string-to-number-conversion warnings

thc1006:
## Changes

Resolves the `bugprone-unchecked-string-to-number-conversion` warnings (part of #2053) by replacing unchecked `atoi`/`sscanf` calls with `strtol`, as the check [recommends](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unchecked-string-to-number-conversion.html).

- 8 `atoi` call sites switched to `strtol` (with an explicit cast that preserves the original result type, so no narrowing-conversion warning is introduced). `<cstdlib>` is added to the two ext/http headers that now use `strtol` directly.
- 1 `sscanf` site in `configuration_parser.cc` is annotated with `NOLINTNEXTLINE`, since its return count is already validated on the next line (a `strtol` rewrite would be redundant there).

No behavior change: `atoi(s)` and `strtol(s, nullptr, 10)` parse identically for valid input.

### Verification
- Affected example/test/sdk targets compile and link (clang-22); a local clang-tidy run confirms the 9 warnings are resolved with no new warnings introduced.
- clang-format clean on all changed files.
- The `warning_limit` in `.github/workflows/clang-tidy.yaml` is lowered from 309/319 to 300/310 (the current CI baseline minus the 9 resolved).

### DIFF
--- a/examples/grpc/client.cc
+++ b/examples/grpc/client.cc
@@ -9,7 +9,7 @@
 #include <grpcpp/support/status.h>
 
 #include <stdint.h>
-#include <stdlib.h>
+#include <cstdlib>
 #include <iostream>
 #include <string>
 #include <utility>
@@ -124,7 +124,7 @@ int main(int argc, char **argv)
   uint16_t port{default_port};
   if (argc > 1)
   {
-    port = static_cast<uint16_t>(atoi(argv[1]));
+    port = static_cast<uint16_t>(std::strtol(argv[1], nullptr, 10));
   }
 
   RunClient(port);
--- a/ext/include/opentelemetry/ext/http/server/http_server.h
+++ b/ext/include/opentelemetry/ext/http/server/http_server.h
@@ -3,7 +3,9 @@
 
 #pragma once
 
+#include <cerrno>
 #include <cstdint>
+#include <cstdlib>
 #include <functional>
 #include <list>
 #include <map>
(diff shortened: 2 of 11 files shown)


marcalff (approved):
LGTM, thanks for the fix.

examples/grpc/client.cc
```diff
@@ -124,7 +124,7 @@ int main(int argc, char **argv)
   uint16_t port{default_port};
   if (argc > 1)
   {
-    port = static_cast<uint16_t>(atoi(argv[1]));
+    port = static_cast<uint16_t>(strtol(argv[1], nullptr, 10));
```
dbarker: please add the std:: namespace to strtol

dbarker (commented):
Thanks for the PR! Please see comment below for a minor change.

thc1006:
Done. Applied it to all the strtol calls in the PR (not just this line) so they stay consistent, and switched those files from `<stdlib.h>` to `<cstdlib>` since that is the header that declares `std::strtol`.

ext/include/opentelemetry/ext/http/server/socket_tools.h
```diff
@@ -196,7 +197,7 @@ struct SocketAddr
     char const *colon  = strchr(addr, ':');
     if (colon)
     {
-      inet4.sin_port = htons(atoi(colon + 1));
+      inet4.sin_port = htons(static_cast<uint16_t>(std::strtol(colon + 1, nullptr, 10)));
```
dbarker: consider adding a check for a valid port (errno, end ptr, returned value in range) before casting to uint16_t

ext/include/opentelemetry/ext/http/server/http_server.h
```diff
@@ -433,7 +434,8 @@ class HttpServer : private SocketTools::Reactor::SocketCallback
         auto const contentLength = conn.request.headers.find("Content-Length");
         if (contentLength != conn.request.headers.end())
         {
-          conn.contentLength = atoi(contentLength->second.c_str());
+          conn.contentLength =
```
dbarker: consider adding a check for a valid return from std::strtol before casting and setting content length. Can be a follow up PR.

dbarker (approved):
Thanks for the cleanup. Approved, with some optional suggestions for this PR since the old behavior is retained.

ext/include/opentelemetry/ext/http/server/socket_tools.h
thc1006: Added, thanks. It now checks `errno`, the end pointer, and the 0-65535 range, and falls back to 0 if any of those fail.

ext/include/opentelemetry/ext/http/server/http_server.h
thc1006: Done here rather than a follow-up. It now checks `errno` and the end pointer and rejects a negative value with 400 Bad Request. I left trailing characters after a valid number tolerated, since the header value parser does not strip trailing whitespace and I did not want to reject something like `100 `.

owent (approved):

PR #4215, opened 2026-07-06
[CODE HEALTH] Fix clang-tidy misc-override-with-different-visibility warnings

thc1006:
Contributes to #4195

## Changes

Fixes a subset of #4195 by aligning the visibility of virtual function overrides with their base-class declaration, which is the fix the `misc-override-with-different-visibility` check asks for (per the [check documentation](https://clang.llvm.org/extra/clang-tidy/checks/misc/override-with-different-visibility.html)).

This PR resolves 31 of the 36 reported warnings across 13 files. The remaining 5 warnings (in `ext/include/opentelemetry/ext/http/server/http_server.h` and `socket_tools.h`) arise from `private`/`protected` inheritance of `SocketTools::Reactor::SocketCallback` and need a different, more careful change; they are deferred to a follow-up PR to keep this one reviewable.

### Fix pattern
- Base `public` -> derived override made `public` (widening; e.g. test mocks of `GetLogger`/`GetTracer`, `Predicate::Match`, `Fields`)
- Base `protected` -> derived override made `protected` (e.g. `sdk::logs::Logger::EnabledImplementation`, gtest `SetUp`/`TearDown`)
- Base `private` (NVI) -> derived override made `private` (e.g. `MetricReader::OnForceFlush`/`OnShutDown`/`OnInitialized`)

No behavior change: only access specifiers on overrides are adjusted; `.cc` definitions do not restate visibility.

### Verification
- Built all affected test and example targets locally (clang-22, ABI v1 and v2 presets); all compile and link.
- clang-format clean on all changed files.
- The `warning_limit` in `.github/workflows/clang-tidy.yaml` is lowered accordingly (abiv1: 309 -> 280, abiv2: 319 -> 288), derived from the current CI baseline minus the 29 (abiv1) / 31 (abiv2) warnings resolved here.

PR #4200, opened 2026-06-30
[CODE HEALTH] Move context propagation test classes into anonymous namespace

mateenali66:
Moves the `TextMapCarrierTest`, `CompositePropagatorTest` and `EnvironmentCarrierTest` classes in the context propagation tests into an anonymous namespace, resolving their `misc-use-internal-linkage` clang-tidy warnings. Test-only, no behavior change.

Lowered the clang-tidy `warning_limit` by 3 for both ABI presets accordingly.

Part of #4196.

PR #4217, opened 2026-07-06
[CODE HEALTH] Move api/test and sdk/test classes into anonymous namespace

thc1006:
## Changes

Part of #4196 (and #2053). Wraps file-local test helper classes in an anonymous namespace to enforce internal linkage, resolving the `misc-use-internal-linkage` warnings across 20 files under `api/test` and `sdk/test`. This follows the same approach as #4199 and #4200.

To keep the review manageable, this is the first batch. The remaining files (under `exporters`, `ext`, and `functional`, plus a few that need manual handling due to existing or nested namespaces) will follow in later batches. The files that overlap with my other open code-health PRs (#4215, #4216) are also deferred to avoid conflicts.

Four metrics tests that share the `WritableMetricStorageTestFixture` fixture (async_metric_storage_test and the sync_metric_storage counter/gauge/up_down_counter tests) are also deferred. They are linked together into `//sdk/test/metrics:all_tests`, and moving same-named fixtures into per-file anonymous namespaces makes gtest treat them as a redefinition of one test suite. They will be handled in a follow-up that renames the fixtures or shares one via a header.

### Verification
- Compiles clean (clang-22, abiv2 preset); a local clang-tidy run confirms the 31 warnings on these 20 files are resolved with no new warnings, and the code still builds.
- clang-format clean on all changed files.
- `warning_limit` lowered from 309/319 to 278/288 (the current baseline minus the 31 resolved here).

Note: this touches the same `warning_limit` line as #4215 and #4216. Whichever of these merges first, I will rebase the others accordingly (the final limit after all three is lower).

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 309
+            warning_limit: 278
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 319
+            warning_limit: 288
     env:
       CC: /usr/bin/clang-22
       CXX: /usr/bin/clang++-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [CODE HEALTH] Move api/test and sdk/test classes into anonymous namespace
+  [#4217](https://github.com/open-telemetry/opentelemetry-cpp/pull/4217)
+
 * [CMAKE] Fix and test WITH_API_ONLY option
   [#4201](https://github.com/open-telemetry/opentelemetry-cpp/pull/4201)
 
--- a/api/test/nostd/shared_ptr_test.cc
+++ b/api/test/nostd/shared_ptr_test.cc
@@ -13,6 +13,9 @@
 
 using opentelemetry::nostd::shared_ptr;
 
+namespace
+{
+
 class A
 {
 public:
@@ -237,3 +240,5 @@ TEST(SharedPtrTest, Sort)
 {
   SharedPtrTest_Sort();
 }
+
+}  // namespace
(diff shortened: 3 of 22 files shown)


marcalff:
CI failure, to investigate:

```
FAIL: //sdk/test/metrics:all_tests (Aborted) (see /home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/_main/bazel-out/k8-fastbuild/testlogs/sdk/test/metrics/all_tests/test.log)
INFO: From Testing //sdk/test/metrics:all_tests:
==================== Test output for //sdk/test/metrics:all_tests:

[ ERROR ] external/googletest+/googletest/src/gtest.cc:2838:: sdk/test/metrics/sync_metric_storage_counter_test.cc:43: Attempted redefinition of test suite WritableMetricStorageTestFixture.
All tests in the same test suite must use the same test fixture
class.  However, in test suite WritableMetricStorageTestFixture, you tried
to define a test using a fixture class different from the one
used earlier. This can happen if the two fixture classes are
from different namespaces and have the same name. You should
probably rename one of the classes to put the tests into different
test suites.```

thc1006:
Thanks for flagging this, fixed and pushed.

Those four files (async_metric_storage_test and the sync_metric_storage counter/gauge/up_down_counter ones) all declare `WritableMetricStorageTestFixture` at namespace scope. That was fine while they shared a single external-linkage class, but once each moved into its own anonymous namespace they became four different classes sharing one suite name, and gtest aborts on that when `//sdk/test/metrics:all_tests` links them together.

Renaming as the gtest message suggests would work, but a test-suite rename feels out of place in a plain internal-linkage cleanup, so I've put those four back to their state on main and left them out of this batch. That takes the PR down to 20 files with the limit at 278/288. I'll pick the four up separately, either renaming the fixtures or hoisting a shared one into a header, whichever you'd rather see.

I didn't catch this locally because I was verifying with the CMake clang-tidy build, which links each test into its own binary and never exercises the combined target. I've added a duplicate-fixture-name check across the batch so it won't slip through again.

marcalff (approved):
LGTM, thanks for the fix.

dbarker (approved):

owent (approved):

PR #4226, opened 2026-07-08
[OTLP EXPORTER] Support setting byte arrays in all attribute collections and reject null keys

dbarker:
Fixes # (issue)

Fixes a few non-compliant areas of the otlp attribute spec:

1. keys must not be null/empty
2. byte arrays must be supported by all attribute collections

See:

https://opentelemetry.io/docs/specs/otel/common/#attribute

https://opentelemetry.io/docs/specs/otel/common/#attribute-collections

> The attribute key MUST be a non-null and non-empty string.
> Case sensitivity of keys is preserved. Keys that differ in casing are treated as distinct keys.
> The attribute value MUST be one of types defined in [AnyValue](https://opentelemetry.io/docs/specs/otel/common/#anyvalue).

All attribute  collections then must accept a byte array.
[Resources](https://opentelemetry.io/docs/specs/otel/resource/sdk/), [Instrumentation Scopes](https://opentelemetry.io/docs/specs/otel/common/instrumentation-scope/), [Metric points](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#metric-points), [Spans](https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes), Span [Events](https://opentelemetry.io/docs/specs/otel/trace/api/#add-events), Span [Links](https://opentelemetry.io/docs/specs/otel/trace/api/#link) and [Log Records](https://opentelemetry.io/docs/specs/otel/logs/data-model/), contain a collection of attributes.

## Changes
- Reject empty keys when setting attributes
- In PopulateAttribute/PopulateAnyValue methods parameters:
  - remove `allow_bytes`
  - add a strongly typed AttributeValueLengthMax object
- Add tests for the modified methods
- Update internal usage to use the new parameters
- Behavior changes from v1.28
    - Attributes with empty keys are rejected
    - Byte sequences are now valid attribute values for instruments, spans, resources, and instrumentation scopes. 


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
```diff
@@ -65,35 +78,59 @@ class OtlpPopulateAttributeUtils
...
-      bool allow_bytes,
-      std::size_t max_length = (std::numeric_limits<std::size_t>::max)()) noexcept;
+      AttributeValueMaxLength max_length = AttributeValueMaxLength{}) noexcept;
+
+  // Deprecated PopulateAnyValue and PopulateAttribute overloads. `allow_bytes` is no longer
+  // meaningful because all AnyValue types support bytes. Prefer the AttributeValueMaxLength
+  // overloads above.
+  OPENTELEMETRY_DEPRECATED_MESSAGE(
+      "allow_bytes is unused; use the AttributeValueMaxLength overload instead")
+  static void PopulateAnyValue(opentelemetry::proto::common::v1::AnyValue *proto_value,
```
dbarker: Decided to deprecate these methods that take `bool allow_bytes` since `allow_bytes` is no longer needed, but we still need the `max_length` arg which will be new in version 1.28.

exporters/otlp/test/otlp_recordable_test.cc
```diff
@@ -779,6 +775,58 @@ TEST(OtlpRecordable, PopulateRequestSameScope)
...
+  const auto &proto_span = rec.span();
+  // Only one attribute; the second call must overwrite, not append.
+  ASSERT_EQ(proto_span.attributes_size(), 1);
+  EXPECT_EQ(proto_span.attributes(0).key(), "key");
+  EXPECT_EQ(proto_span.attributes(0).value().int_value(), 2);
+}
+
+// Test that updating a duplicate key does not consume an attribute slot or increment
+// the dropped-attributes counter, even when the attribute limit is already reached.
+TEST(OtlpRecordable, DISABLED_SetAttributeDeduplicateDoesNotIncrementDropped)
```
dbarker, line 804: These tests for duplicate attribute keys are added but disabled since OtlpRecordable will duplicate keys now.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
```diff
@@ -65,35 +78,59 @@ class OtlpPopulateAttributeUtils
...
-   * are unaffected.
+   * (span<const uint8_t>) are truncated at the raw byte boundary. Non-string
+   * alternatives are unaffected.
    */
   static void PopulateAnyValue(
       opentelemetry::proto::common::v1::AnyValue *proto_value,
       const opentelemetry::common::AttributeValue &value,
-      bool allow_bytes,
-      std::size_t max_length = (std::numeric_limits<std::size_t>::max)()) noexcept;
+      AttributeValueMaxLength max_length = AttributeValueMaxLength{}) noexcept;
```
owent: Can we use a more generic option class so that adding more options in the future won't require modifying all these functions?

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -58,158 +65,105 @@ inline void SetUint64Value(opentelemetry::proto::common::v1::AnyValue *proto_val
...
-  else if (nostd::holds_alternative<double>(value))
-  {
-    proto_value->set_double_value(nostd::get<double>(value));
-  }
-  else if (nostd::holds_alternative<const char *>(value))
+
+  void operator()(bool value) const { proto_value_->set_bool_value(value); }
+  void operator()(int32_t value) const { proto_value_->set_int_value(value); }
+  void operator()(int64_t value) const { proto_value_->set_int_value(value); }
+  void operator()(uint32_t value) const { proto_value_->set_int_value(value); }
```
owent: Just wondering — why doesn't this code trigger an unsigned/signed integer conversion warning?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
dbarker: Good idea. Will rename the type to be generic converter options.

exporters/otlp/src/otlp_populate_attribute_utils.cc
dbarker: Good question. I updated each overload to use static_cast to int64 where needed and use explicit types in the for loops.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
dbarker: updated

dbarker:
Hi @owent Thanks for initial feedback. Can you take another review pass?

exporters/otlp/src/otlp_populate_attribute_utils.cc
```diff
@@ -36,180 +42,135 @@ namespace exporter
...
-  {
-    proto_value->set_int_value(nostd::get<int64_t>(value));
-  }
-  else if (nostd::holds_alternative<unsigned int>(value))
+
+  void operator()(bool value) const { proto_value_->set_bool_value(value); }
+  void operator()(std::int32_t value) const
   {
-    proto_value->set_int_value(nostd::get<unsigned int>(value));
+    proto_value_->set_int_value(static_cast<std::int64_t>(value));
```
owent: I may be missing something, but implicit int32 → int64 should be safe and warning-free on MSVC, Clang, and GCC. Those usually only warn for narrowing or sign mismatches. Have you observed warnings with this pattern?

owent (approved):
There's still an unnecessary conversion, but it's harmless. The rest LGTM. Thanks.

exporters/otlp/src/otlp_populate_attribute_utils.cc
dbarker: I haven't observed any warnings with this conversion prior to adding static_cast. It is not a required cast.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
```diff
@@ -59,41 +85,58 @@ class OtlpPopulateAttributeUtils
...
-      std::size_t max_length = (std::numeric_limits<std::size_t>::max)()) noexcept;
+      AttributeConverterOptions options = AttributeConverterOptions{}) noexcept;
+
+  /**
+   *  The following overloads are deprecated and will be removed in a future release. Use the
+   *overloads that take an AttributeConverterOptions instead. The deprecated overloads call the new
+   *overloads with default AttributeConverterOptions, the `allow_bytes` parameter is unused.
+   **/
+  OPENTELEMETRY_DEPRECATED_MESSAGE(
+      "allow_bytes is unused; use the AttributeConverterOptions overload instead")
```
marcalff: These helpers use a protobuf parameter, so they are to support the OTLP implementation itself, and are not supposed or intended to be used by an application.

Please remove all outdated helpers instead of deprecating them.

marcalff:
do-not-merge label: to merge after the release of opentelemetry-cpp 1.28.0

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
dbarker: Thanks. I've removed the deprecated methods.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_populate_attribute_utils.h
marcalff: Now removed, resolved.

marcalff (approved):
LGTM, thanks for the fix.

PR #4231, opened 2026-07-09
[CODE HEALTH] Move otlp, zipkin and prometheus test classes into anonymous namespace

mateenali66:
Finishes the misc-use-internal-linkage cleanup from #4196, covering the otlp, zipkin and prometheus exporter test files thc1006 handed off after #4225.

Moves the flagged test fixtures and helper classes into anonymous namespaces to give them internal linkage. The friend-declared exporter TestPeer classes are not flagged and are left in place, since moving them would break the friend relationship with the production exporters.

Verified on an Ubuntu 24.04 + clang-tidy-22 build (the same preset the CI job uses): all 19 misc-use-internal-linkage warnings in these files are cleared, the affected test targets compile and link, and clang-format is clean. Lowered the warning_limit by 19 for both ABI presets.

Part of #4196.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 235
+            warning_limit: 216
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 245
+            warning_limit: 226
     env:
       CC: /usr/bin/clang-22
       CXX: /usr/bin/clang++-22
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 216
+            warning_limit: 137
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 226
+            warning_limit: 145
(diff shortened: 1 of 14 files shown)


marcalff (approved):
LGTM, thanks for the cleanup.

.github/workflows/clang-tidy.yaml
```diff
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 235
+            warning_limit: 216
```
dbarker: this can drop to 137
See https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/29032207655?pr=4231

.github/workflows/clang-tidy.yaml
```diff
@@ -17,9 +17,9 @@ jobs:
...
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 245
+            warning_limit: 226
```
dbarker: this can drop to 145
See https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/29032207655?pr=4231

dbarker (approved):
Thanks! Please see minor comments below on the warning counts.

.github/workflows/clang-tidy.yaml
marcalff: Yes, some cleanup PR did not lower the limits.
I was planning to tight this up in the release PR, but here works too.

PR #4240, opened 2026-07-11
[SDK] Support independent programmatic configuration

### LINKED ISSUE #4134: Support Programmatic Configuration without YAML dependencies
## Background

The configuration spec requires programmatic configuration of the SDK
in-memory that is independent of the file format (YAML, JSON, etc.).

See https://opentelemetry.io/docs/specs/otel/configuration/#programmatic

> The SDK MUST provide a programmatic interface for all configuration. This interface SHOULD be written in the language of the SDK itself. All other configuration mechanisms SHOULD be built on top of this interface.

This interface exists in `sdk/<include,src>/configuration` and available through the  `opentelemetry_configuration` CMake target.  This target also includes the YAML parsing components 
(`ryml_document.cc`, `ryml_document_node.cc`, `yaml_configuration_parser.cc`, and `ryml` dependency).  

Currently there isn't a way to only use the programmatic configuration without bringing in the `ryml` dependency. 

## Proposal

Allow users to link to the programmatic configuration components independent from the YAML configuration components. 

1. Split `opentelemetry_configuration` into two CMake targets:
   - ~~`opentelemetry_configuration`~~ `opentelemetry_core`: programmatic configuration components only. `sdk_builder.cc`,
     `configured_sdk.cc`, `registry.cc`, and component builders. 
   - ~~`opentelemetry_configuration_yaml`~~ `opentelemetry_configuration`: yaml configuration components. `yaml_configuration_parser.cc`,
     `ryml_document.cc`, `ryml_document_node.cc`, and the `ryml::ryml`
     dependency.

2. Build ~~`opentelemetry_configuration`~~ `opentelemetry_configuration_core` (programmatic configuration) unconditionally as part of the SDK.

3. Repurpose the existing `WITH_CONFIGURATION` CMake option to gate only the
   YAML configuration target (`opentelemetry_configuration_yaml`) and the `ryml` dependency.

4. Optional: Remove `doc_` from the `Configuration` class and use a default constructor. Currently `Configuration` holds a `std::unique_ptr<Document>` to keep the `ryml` parse tree alive
   for the duration of `ConfigurationParser::Parse()`.

## Example Programmatic Configuration 

```cpp
namespace cfg = opentelemetry::sdk::configuration;

auto model         = std::make_unique<cfg::Configuration>();

auto tp = std::make_unique<cfg::TracerProviderConfiguration>();

auto otlp_grpc_exporter      = std::make_unique<cfg::OtlpGrpcSpanExporterConfiguration>();
otlp_grpc_exporter->endpoint = "http://localhost:4317";

auto processor               = std::make_unique<cfg::SimpleSpanProcessorConfiguration>();
processor->exporter          = std::move(otlp_grpc_exporter);

tp->processors.push_back(std::move(processor));  

model->tracer_provider = std::move(tp);

auto registry = std::make_shared<cfg::Registry>();
opentelemetry::exporter::otlp::OtlpGrpcSpanBuilder::Register(registry.get()); 

auto sdk = cfg::ConfiguredSdk::Create(registry, model);

sdk->Install();
```

dbarker:
Fixes #4134  and follow up from #https://github.com/open-telemetry/opentelemetry-cpp/pull/3655#pullrequestreview-3277362332  

Break out the programmatic configuration to its own library so it can be used without the ryml dependency. 

[edit - crossed out items that were broken out to separate PRs]
## Changes
- ~~Creates the programmatic configuration target (`configuration_core`)~~
- ~~Adds configuration_core to the CMake install tests~~
- Adds deprecated names to components and deprecates the *builder components since they only depend on the SDK now and do not bring in dependencies. 
- ~~Adds programmatic_configuration_test.cc~~
 - ~~Fix issues found with the new test~~
   - ~~LoggerConfigurator supporting min severity and trace based fields~~
   - ~~Initialize members of config classes to the spec default constants.~~ 
   - ~~Set the default parent based root sampler to always_on~~
   - ~~Set the default sampler to parent based always on~~
   - ~~Set all batch processor options that the user configured~~
   - ~~Handle default metric stream aggregation~~ 
- ~~Fix the `cppcoreguidelines-pro-type-member-init` warnings in configuration and sdk files used by the tests.~~ 
- ~~Update INSTALL doc for the configuration targets~~      


sdk/include/opentelemetry/sdk/configuration/base2_exponential_bucket_histogram_aggregation_configuration.h
```diff
@@ -23,6 +23,7 @@ class Base2ExponentialBucketHistogramAggregationConfiguration : public Aggregati
     visitor->VisitBase2ExponentialBucketHistogram(this);
   }
 
+  // Zero values indicate the field is not set.
   std::size_t max_scale{0};
```
dbarker: This is a case where zero is a valid value for the option so it is not a great way to indicate "not set". Instead of using zero in all these config structs to indicate not set, we may consider either setting the spec defaults (which the SDK/exporter options structs also do) or bring in an std::optional/nostd::optional type.

dbarker: The spec seems clear on the in memory configuration model. 

https://opentelemetry.io/docs/specs/otel/configuration/sdk/#in-memory-configuration-model

> If a property is present and the value is null, Create MUST use the nullBehavior, or defaultBehavior if nullBehavior is not set.

Changes:
- Updated the config files to define and initialize to the default values provided by the [configuration schema](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/) as `defaultBehavior`  . 
- Updated the parser to use the config class defaults. 
- Added `TODOs` where the current implementation defaults differed from the schema.

dbarker:
This PR will be broken up into multiple PRs (will edit with PR links).
1. Tests for SdkBuilder, ConfiguredSdk, and integration tests for programmatic configuration along with fixes to the SdkBuilder they exposed. 
   - https://github.com/open-telemetry/opentelemetry-cpp/pull/4243
3. Move spec defaults from the `configuration_parser.cc` to each configuration struct.
    - https://github.com/open-telemetry/opentelemetry-cpp/pull/4244
4. Add CMake support for deprecating components and update the policy
   - https://github.com/open-telemetry/opentelemetry-cpp/pull/4272
5. CMake separation of configuration_core and configuration
   - https://github.com/open-telemetry/opentelemetry-cpp/pull/4311
6. Clean up transitive includes and linking for configuration_core
   - https://github.com/open-telemetry/opentelemetry-cpp/pull/4335
7. CMake component cleanup and deprecation of the "builder" components since the sdk will now include the builder interfaces. 
   - https://github.com/open-telemetry/opentelemetry-cpp/pull/4428

dbarker:
The last PR associated with this draft has been created. Closing this.

PR #4224, opened 2026-07-08
[SDK] Support LoggerConfigurator updates and example

dbarker:
Fixes # (issue)

Adds the ability to update a LoggerConfigurator (and existing loggers' LoggerConfig) at runtime. Following the pattern used in #4065.

This implements the optional requirement to allow updating the LoggerProvider's LoggerConfigurator : https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerconfigurator

## Changes

- Adds UpdateLoggerConfigurator to LoggerProvider to support setting new logger configs at runtime
- Adds test and an example

For significant contributions please make sure you have completed the following items:

* [ ] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4131, opened 2026-06-06
[CONFIGURATION] Implementing minimum_severity and trace_based attributes for LoggerConfig

om7057:
Fixes #4130 

## Description

Implements the missing `minimum_severity` and `trace_based` attributes for `LoggerConfig` in the declarative configuration spec.

Previously, `LoggerConfigConfiguration` only had an `enabled` boolean field. This PR adds:
- **`minimum_severity`** (type: `SeverityNumber`, default: `trace`) — allows filtering log records by severity level
- **`trace_based`** (type: `bool`, default: `false`) — enables/disables trace-based log filtering

These attributes align with the OpenTelemetry specification for logger configuration and enable users to control logging behavior via YAML configuration.

**Example YAML usage:**
```yaml
logger_provider:
  logger_configurator/development:
    default_config:
      enabled: true
      minimum_severity: warn
      trace_based: false
    loggers:
      - name: my.noisy.logger
        config:
          minimum_severity: error
          trace_based: true
```          
          
For significant contributions, please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4241, opened 2026-07-12
[SDK] Fix yaml configuration of the logger configurator

dbarker:
Fixes # (issue)

The yaml configuration of the LoggerConfigurator only respects the enabled property and ignores minimum severity and trace based. This PR fixes the SDK builder to configure all properties of the LoggerConfig. 

## Changes

- Update the CreateLoggerConfigurator method of SdkBuilder to use the LoggerConfig::Create method. 
- Add an SdkBuilder test to verify the default and scope specific logger configs have all properties set as configured. 

### DIFF
--- a/sdk/src/configuration/sdk_builder.cc
+++ b/sdk/src/configuration/sdk_builder.cc
@@ -16,6 +16,7 @@
 #include "opentelemetry/common/kv_properties.h"
 #include "opentelemetry/context/propagation/composite_propagator.h"
 #include "opentelemetry/context/propagation/text_map_propagator.h"
+#include "opentelemetry/logs/severity.h"
 #include "opentelemetry/nostd/span.h"
 #include "opentelemetry/nostd/string_view.h"
 #include "opentelemetry/sdk/common/global_log_handler.h"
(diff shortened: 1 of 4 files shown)


dbarker:
@marcalff this is a follow up to https://github.com/open-telemetry/opentelemetry-cpp/pull/4131, which added the yaml side of the LoggerConfigurator parsing. This PR is required for the LoggerConfigurator to be created correctly based on the yaml.  Since 4131 is new for the 1.28 release I think this PR fix should also be included. Please take a look.

marcalff (approved):
LGTM.

Please also remove:

- LoggerConfig::Enabled()
- LoggerConfig::Disabled()

so users are forced to call LoggerConfig::Create() with all properties.

dbarker:
> Please also remove:
> 
> LoggerConfig::Enabled()
> LoggerConfig::Disabled()
> so users are forced to call LoggerConfig::Create() with all properties.

This will impact tests and require more changes.  Let's address it in a follow up PR and not block the release.

marcalff:
> LGTM.
> 
> Please also remove:
> 
>     * LoggerConfig::Enabled()
> 
>     * LoggerConfig::Disabled()
> 
> 
> so users are forced to call LoggerConfig::Create() with all properties.

Enabled() really should go: if we keep it, there will be confusion forever with the minimum severity and the trace based flag ignored.


About Disabled(), it might be ok.

PR #4224 uses

```
  const logs_sdk::LoggerConfig disabled_config = logs_sdk::LoggerConfig::Disabled();
```

which makes sense.

@dbarker 

Any thoughts on whether to keep the Disabled() helper alone or not ?

marcalff:
Ok, merging now then.

Thanks for the fix.

dbarker:
> Any thoughts on whether to keep the Disabled() helper alone or not ?

I think Disabled is okay to keep and don't mind Enabled too much since both set the spec default values for min severity and trace based, which are hard to keep track of if not looking at the spec. They are easier to read too :) 

```cpp
 // Disabled configs
 const logs_sdk::LoggerConfig disabled_config = logs_sdk::LoggerConfig::Disabled();
 const logs_sdk::LoggerConfig disabled_config_explicit = logs_sdk::LoggerConfig::Create(false, logs_api::kInvalid, false);
 
 // Enabled configs
 const logs_sdk::LoggerConfig enabled_config = logs_sdk::LoggerConfig::Enabled();
 const logs_sdk::LoggerConfig enabled_config_explicit = logs_sdk::LoggerConfig::Create(true, logs_api::kInvalid, false);
 
```
I think this bug just came up naturally as the spec evolved and new properties were added, yaml parsing was updated but the sdk building was missed. Easy to do, but the configuration side should be good now.

PR #4243, opened 2026-07-13
[CONFIGURATION] Programmatic configuration use case tests and fixes

dbarker:
Contributes to #4134

Adds unit and integration tests for the programmatic configuration use case and fixes issues found.

## Changes

- Fixes missing initialization of SdkBuilder and ConfiguredSdk members. 
- Fixes issues with SdkBuilder for the programmatic configuration use case
   - Set the default sampler to parent based always on when the sampler config is null
   - Set the default parent based root sampler to always_on when configured root sampler is null
   - Set the default metric stream aggregation when the aggregation config is null
   - Set export_timeout for batch processors. (export timeout is unused in the SDK but should be configured to avoid being set by the environment variable) 
- SdkBuilder test for CreateParentBasedSampler has been added
- ConfiguredSdk tests for the core interface (Global Logger/Install/Uninstall) have been added
- Programmatic Configuration integration tests have been added to verify the configured SDK has the expected components and exports expected signal data


marcalff (approved):
LGTM.

Thanks for all the tests.

sdk/test/configuration/configured_sdk_test.cc
```diff
@@ -0,0 +1,313 @@
...
+//                      For integration testing to determine if the SDK is configured correctly, see
+//                      the programmatic_configuration_test.cc file
+
+TEST_F(ConfiguredSdkTest, ConfiguredSdkDefaultLogLevel)
+{
+  // default log level is info
+  auto model      = std::make_unique<config_sdk::Configuration>();
+  model->resource = std::make_unique<config_sdk::ResourceConfiguration>();
+  CreateSdk(model);
+  sdk_->Install();
```
marcalff, line 275: Not sure if TearDown() is invoked for each test or just for the serie of tests.

It would be cleaner to invoke sdk->UnInstall() explicitly in each test, so that the start and end state is clean, avoiding side effects from test A leaking to test B.

Possibly related: test `MeterProviderWithDefaults` times out, to investigate.

dbarker: TearDown() runs on each case defined by a TEST_F. Doing an Uninstall on TearDown and setting the noop providers on SetUp clears out the global state so these tests can all run in one executable. 

The timeout issue with the `MeterProviderWithDefaults` case seems to be a race condition in the PeriodicExportingMetricReader creation/startup sequence. The test can create the reader and call ForceFlush/Shutdown before the reader thread has started waiting causing the main thread to block. 

I will disable this test for now, but would like to add the ability to test programmatic configuration without instantiating the actual `PeriodicExportingMetricReader` (which spawns a thread). 

I'd like to add builder function support to the Registry (Get*Builder/Set*Builder) for all the SDK components (similar to how this is done for the otlp/ostream exporters) and use them in the SdkBuilder for creating these built-in components. This would allow use to sub in mock builders for testing programmatic configuration without needing to introspect the sdk components (which cannot be done now without adding more friends or accessors to those components). 

Would adding builder support for all sdk components be a reasonable change/addition to the configuration library?

dbarker: The tests that used the periodic metric reader are disabled now. To continue investigation and enabling these tests in followup issues/prs.

marcalff: @dbarker 

> I'd like to add builder function support to the Registry (Get_Builder/Set_Builder) for all the SDK components (similar to how this is done for the otlp/ostream exporters) and use them in the SdkBuilder for creating these built-in components. This would allow use to sub in mock builders for testing programmatic configuration without needing to introspect the sdk components (which cannot be done now without adding more friends or accessors to those components).
> 
> Would adding builder support for all sdk components be a reasonable change/addition to the configuration library?

Yes.

This is not only helping testing, it is also required for the normal use cases in production.

My overall idea is as follows:

* An application, in main(), configures the registry to cherry pick what component is allowed. This is done at compile time.
* A user, in config.yaml, selects which components are used at runtime, but only in the subset of what is available in the registry.

This way, application designers have a say about what is or is not in scope, instead of allowing every single yaml construct. An application may want traces only but no metrics for example, and only allowing a few exporters.

To do in a separate PR.

PR #4244, opened 2026-07-13
[CONFIGURATION] Define and set spec default values in configuration objects

dbarker:
Contributes to https://github.com/open-telemetry/opentelemetry-cpp/issues/4134

The configuration spec defines default values through `defaultBehavior` annotations and the in-memory configuration model must set the defaults. 

https://opentelemetry.io/docs/specs/otel/configuration/sdk/#in-memory-configuration-model

> If a property is present and the value is null, Create MUST use the nullBehavior, or defaultBehavior if nullBehavior is not set.

These values are currently only applied in the yaml parser and the configuration c++ objects for the in-memory model are left uninitialized. This PR moves the spec defaults from the parser into constants in each configuration object and initializes all members of the objects. 

See the schemas for the `defaultBehavior` annotations:
- https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/logger_provider.yaml
- https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml
- https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/meter_provider.yaml

## Changes

- Adds the spec defaults parameter values as constants to the configuration classes 
- Initialize all arithmetic members of the configuration classes to the spec defaults
    - Adds TODO comments where the current default is not aligned with the spec. 
- Use the configuration class constants in the parser.  
   

### DIFF
--- a/sdk/include/opentelemetry/sdk/configuration/attribute_limits_configuration.h
+++ b/sdk/include/opentelemetry/sdk/configuration/attribute_limits_configuration.h
@@ -18,8 +18,12 @@ namespace configuration
 class AttributeLimitsConfiguration
 {
 public:
-  std::size_t attribute_value_length_limit;
-  std::size_t attribute_count_limit;
+  // TODO: spec default is no limit, using 4096 to preserve original behavior
+  static constexpr std::size_t kDefaultAttributeValueLengthLimit = 4096;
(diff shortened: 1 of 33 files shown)


marcalff (approved):
LGTM, thanks for the fix.

PR #4245, opened 2026-07-14
[METRICS SDK] Use std::mutex for SyncMetricStorage record-path lock

abhinandan-codes:
## Problem

On Windows, `SpinLockMutex::lock()` falls through to
`std::this_thread::sleep_for(std::chrono::milliseconds(1))` as its final
back-off step. Windows rounds sleep durations up to the current system timer
resolution, which defaults to **~15.6 ms**. So a thread that loses the lock race
does not sleep ~1 ms — it parks for a **full timer quantum (~15.6 ms)**.

Under sustained contention on a single instrument (for example a hot histogram
recorded from many threads via `SyncMetricStorage`, which holds a per-stream
`SpinLockMutex`), this produces a large tail latency: the median `Record()` is a
few microseconds but **p99 jumps to ~15 ms**.

## Root cause

`sleep_for(1ms)` is not 1 ms on Windows. The intended sub-millisecond
incremental back-off becomes a ~15.6 ms stall whenever a waiter reaches the
sleep phase.

## Fix

Exploring options — see comment below for the two approaches and benchmarks; open to maintainer preference.

## Measurements

Windows 11, 6 threads hammering one histogram at ~2000 rec/s, default 15.6 ms
timer:

| build | p50 | p99 | records >= 5 ms |
|---|---|---|---|
| SpinLock (`sleep_for`) | few us | ~15 ms | ~52% |

## Notes

- Header-only change in the API; no ABI break (same size class, same public
  surface).
- `timeBeginPeriod(1)` mitigates the symptom globally but is a process-wide side
  effect, not a library-level fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@ Increment the:
 
 ## [Unreleased]
 
+* [METRICS SDK] Fix Windows metrics tail latency on the synchronous record path:
+  `SyncMetricStorage::attribute_hashmap_lock_` now uses `std::mutex` instead of
+  `common::SpinLockMutex`. The spin lock's final `sleep_for(1ms)` back-off is
+  rounded up to the Windows timer quantum (~15.6 ms), so a waiter contending
+  on a hot instrument could stall for a full quantum, inflating tail latency.
+  `std::mutex` blocks the waiter instead. Only this one contended lock changes;
+  the API `SpinLockMutex` and all other SDK locks are unchanged.
+  [#4245](https://github.com/open-telemetry/opentelemetry-cpp/pull/4245)
+
 * [OTLP/HTTP] Honor `Retry-After` response header when retrying exports,
   supporting both delay-seconds and HTTP-date formats per RFC 7231 §7.1.3.
   [#4172](https://github.com/open-telemetry/opentelemetry-cpp/issues/4172)
--- a/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
+++ b/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
@@ -103,7 +103,7 @@ public:
     }
 #endif
     static MetricAttributes attr = MetricAttributes{};
-    std::lock_guard<opentelemetry::common::SpinLockMutex> guard(attribute_hashmap_lock_);
+    std::lock_guard<std::mutex> guard(attribute_hashmap_lock_);
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
     MetricAttributes resolved = ResolveCardinality(attr);
     attributes_hashmap_->GetOrSetDefault(std::move(resolved), create_default_aggregation_)
@@ -131,7 +131,7 @@ public:
 #endif
 
     MetricAttributes attr{attributes, attributes_processor_.get()};
-    std::lock_guard<opentelemetry::common::SpinLockMutex> guard(attribute_hashmap_lock_);
+    std::lock_guard<std::mutex> guard(attribute_hashmap_lock_);
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
     // Resolve via the unified cardinality policy so unbound and bound paths
     // share one combined limit (see ResolveCardinality()).
@@ -161,7 +161,7 @@ public:
     }
 #endif
     static MetricAttributes attr = MetricAttributes{};
-    std::lock_guard<opentelemetry::common::SpinLockMutex> guard(attribute_hashmap_lock_);
+    std::lock_guard<std::mutex> guard(attribute_hashmap_lock_);
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
     MetricAttributes resolved = ResolveCardinality(attr);
     attributes_hashmap_->GetOrSetDefault(std::move(resolved), create_default_aggregation_)
@@ -188,7 +188,7 @@ public:
     }
 #endif
     MetricAttributes attr{attributes, attributes_processor_.get()};
-    std::lock_guard<opentelemetry::common::SpinLockMutex> guard(attribute_hashmap_lock_);
+    std::lock_guard<std::mutex> guard(attribute_hashmap_lock_);
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
     MetricAttributes resolved = ResolveCardinality(std::move(attr));
     // cppcheck-suppress accessMoved
@@ -300,7 +300,7 @@ private:
   nostd::shared_ptr<ExemplarReservoir> exemplar_reservoir_;
 #endif
   TemporalMetricStorage temporal_metric_storage_;
-  opentelemetry::common::SpinLockMutex attribute_hashmap_lock_;
+  std::mutex attribute_hashmap_lock_;
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
   // NOTE: ENABLE_METRICS_BOUND_INSTRUMENTS_PREVIEW changes the layout and
   // vtable of SyncMetricStorage (these conditional members and the virtual
--- a/sdk/src/metrics/state/sync_metric_storage.cc
+++ b/sdk/src/metrics/state/sync_metric_storage.cc
@@ -5,7 +5,6 @@
 #include <mutex>
 #include <utility>
 
-#include "opentelemetry/common/spin_lock_mutex.h"
 #include "opentelemetry/common/timestamp.h"
 #include "opentelemetry/nostd/function_ref.h"
 #include "opentelemetry/nostd/span.h"
@@ -23,6 +22,7 @@
 #  include <unordered_set>
 #  include <vector>
 
+#  include "opentelemetry/common/spin_lock_mutex.h"
 #  include "opentelemetry/sdk/common/global_log_handler.h"
 #  include "opentelemetry/sdk/metrics/aggregation/aggregation.h"
 #  include "opentelemetry/sdk/metrics/data/exemplar_data.h"
@@ -51,7 +51,7 @@ bool SyncMetricStorage::Collect(CollectorHandle *collector,
   std::vector<std::shared_ptr<BoundEntry>> entry_snapshot;
 #endif
   {
-    std::lock_guard<opentelemetry::common::SpinLockMutex> guard(attribute_hashmap_lock_);
+    std::lock_guard<std::mutex> guard(attribute_hashmap_lock_);
     delta_metrics = std::move(attributes_hashmap_);
     attributes_hashmap_.reset(new AttributesHashMap(aggregation_config_->cardinality_limit_));
 #ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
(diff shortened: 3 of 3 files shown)


api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -79,6 +81,26 @@ class SpinLockMutex
...
+   * The SRWLOCK spins briefly and then parks the waiting thread until the lock is
+   * released, so a contended waiter does not sleep for a full timer quantum.
+   */
+  void lock() noexcept { AcquireSRWLockExclusive(&mutex_); }
+
+  /** Releases the lock held by the execution agent. Throws no exceptions. */
+  void unlock() noexcept { ReleaseSRWLockExclusive(&mutex_); }
+
+private:
+  SRWLOCK mutex_ = SRWLOCK_INIT;
```
dbarker: This will be an ABI break.
Please see https://github.com/open-telemetry/opentelemetry-cpp/blob/4e20f690352ce6c2793efcb6448e51363a2cad18/api/include/opentelemetry/trace/provider.h#L52

api/include/opentelemetry/common/spin_lock_mutex.h
dbarker, line 123: The 1ms sleep (which may sleep longer) seems undesirable for any platform. Is there a common approach we can apply that is platform agnostic and doesn't break ABI for the  class?

dbarker (commented):
Thanks for the PR and for pointing out the 1ms sleep issue. Please see some initial feedback and question below.

marcalff (dismissed):
Thanks for the patch.

As described by @dbarker, this change breaks the ABI: it does not "backs" the spin lock with a read write lock, it just replaces the spin lock by a read write lock instead.

Given how multiple libraries can be compiled with different versions of this header file, and how each instrumentation will need a lock to use tracer / meter / logger providers, this is a major issue.

A better way would be to change the implementation of the lock() method alone, when waiting for the spin lock to be available.

malkia:
SRWLocks are not re-entrant - not sure if we fall into situation where this might happen. This also calls for AppVerif test on Windows through `cuzz` and other verifiers.

abhinandan-codes:
**Context: reworking this PR from the SRWLOCK approach to a parking back-off, and a question for the SDK team on which direction you'd prefer.**


**Option 1 — park the waiter in `SpinLockMutex` (this PR).**
Keep the existing atomic spin lock and its fast-spin/yield phases; only replace the final `sleep_for(1ms)` with parking the waiter, woken immediately on `unlock()`:
- C++20: `std::atomic<bool>::wait()` / `notify_one()`.
- Pre-C++20 MSVC (Win8+): `WaitOnAddress` / `WakeByAddressSingle`.
- Otherwise: the previous `sleep_for` fallback (unchanged off-Windows behavior).

The `std::atomic<bool>` member is unchanged, so **the public API layout/ABI is preserved**. It's a header-only change and fixes the lock for *every* `SpinLockMutex` user (trace/logs/metrics), not just metrics.

**Option 2 — an SDK-local, CMake-selectable lock for the metrics hot path.**
Leave the API `SpinLockMutex` untouched and introduce `sdk::common::SyncLock`, selected at build time (`OPENTELEMETRY_SDK_METRICS_LOCK` = `spinlock` default / `srwlock` / `std_mutex`), swapping the metrics-SDK lock callsites. The default build is a no-op; a non-default value changes SDK metrics class layout, so the macro must be set consistently across the SDK build and consumers (propagated as a PUBLIC/INTERFACE compile definition). This keeps the change opt-in and confined to metrics, at the cost of an ABI/build-consistency contract.

Separately wanted to know your testing infrastructure setup as option 1 may need testing it across C++ versions and platforms


**Which path would you prefer** — parking the API `SpinLockMutex` (Option 1), the CMake-selectable SDK lock (Option 2), or another approach you have in mind? Happy to adjust this PR accordingly. This is a critical bug blocking our move to OpenTelemetry, so any steer on the preferred direction would be a big help. Thanks!

lalitb:
@abhinandan-codes  - The Windows issue is real, and replacing the 1 ms sleep with parking is a good idea. My concern is that the measured problem is in the metrics hot path, but this PR changes the shared `SpinLockMutex` used across metrics, tracing, logging, providers, and other platforms.

Because this lock is defined in a header, `lock()` and `unlock()` now get different definitions depending on the C++ standard and `_WIN32_WINNT`. If linked code uses different settings, that violates the ODR. Some code may park and expect a notification, while other code may only clear the flag without notifying. The program may still link, but a waiting thread could get stuck. Keeping the same class size avoids a layout ABI break, but it does not make these implementations compatible.

There is also a concrete compatibility issue: a C++20 build targeting macOS 10.15 can select `atomic::wait()`, while `libc++` supports that operation only from macOS 11.

The change also performs a notification on every unlock, including uncontended paths outside metrics, but the PR only provides measurements for the Windows metrics case. There are no correctness tests for multiple waiters or repeated lock handoff either.

I suggest keeping `common::SpinLockMutex` unchanged and using a fixed SDK-internal lock for the affected metrics path. A `std::mutex`, which uses an efficient Windows lock underneath, or a dedicated internal parking lock would solve the measured problem without changing synchronization behavior across the whole project. I would also avoid a CMake option that changes the lock type, because that can create another mixed-layout problem.

Overall: the problem and general direction are good, but I think the current implementation is too broad and has compatibility risks. I’m requesting changes and would prefer the SDK-internal metrics approach.

lalitb (changes requested):
As discussed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4245#issuecomment-5087761325)

api/include/opentelemetry/common/spin_lock_mutex.h
```diff
@@ -20,30 +20,56 @@
...
+//
+// NOTE: the selection depends on the C++ standard in effect. As with the rest of
+// this header, all translation units linked together must be compiled with the
+// same C++ standard / STL, otherwise a parking lock() could be paired with a
+// non-notifying unlock().
+#if defined(__cpp_lib_atomic_wait) && __cpp_lib_atomic_wait >= 201907L
+#  define OTEL_SPINLOCK_PARK_STD 1
+#elif defined(_MSC_VER) && defined(_WIN32_WINNT) && (_WIN32_WINNT >= 0x0602)
+#  define OTEL_SPINLOCK_PARK_WIN32 1
+#  pragma comment(lib, "Synchronization.lib")
```
ThomsonTan: The `#pragma comment(lib, "Synchronization.lib")` makes every
translation unit that includes this public header acquire a link-time and
load-time dependency, invisible to CMake and Bazel.

marcalff:
> I suggest keeping `common::SpinLockMutex` unchanged and using a fixed SDK-internal lock for the affected metrics path. A `std::mutex`, which uses an efficient Windows lock underneath, or a dedicated internal parking lock would solve the measured problem without changing synchronization behavior across the whole project. I would also avoid a CMake option that changes the lock type, because that can create another mixed-layout problem.
> 
> Overall: the problem and general direction are good, but I think the current implementation is too broad and has compatibility risks. I’m requesting changes and would prefer the SDK-internal metrics approach.

Thanks @lalitb for the in depth analysis.

I agree with lalit here, using a different lock instead of a spin lock where there is contention for metrics is less intrusive compared to changing the spin lock implementation itself for every uses.

A spin lock is supposed to be held for a very short time, not causing contention. If there is contention, it could be because the lock is held for too long, so that a spin lock is not appropriate in the user case found with metrics.

ThomsonTan:
> > I suggest keeping `common::SpinLockMutex` unchanged and using a fixed SDK-internal lock for the affected metrics path. A `std::mutex`, which uses an efficient Windows lock underneath, or a dedicated internal parking lock would solve the measured problem without changing synchronization behavior across the whole project. I would also avoid a CMake option that changes the lock type, because that can create another mixed-layout problem.
> > Overall: the problem and general direction are good, but I think the current implementation is too broad and has compatibility risks. I’m requesting changes and would prefer the SDK-internal metrics approach.
> 
> Thanks @lalitb for the in depth analysis.
> 
> I agree with lalit here, using a different lock instead of a spin lock where there is contention for metrics is less intrusive compared to changing the spin lock implementation itself for every uses.
> 
> A spin lock is supposed to be held for a very short time, not causing contention. If there is contention, it could be because the lock is held for too long, so that a spin lock is not appropriate in the user case found with metrics.

I think we can scope it down to a single member. The measured hot path is entirely behind one lock:

https://github.com/open-telemetry/opentelemetry-cpp/blob/5d7db843ea08d6efa7ce7e3f9383a9e6c1f5c157/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h#L303

All four `RecordLong()` / `RecordDouble()` overloads serialize on `attribute_hashmap_lock_`, so switching just this one member to `std::mutex` (plus the four `lock_guard` instantiations) covers the "hot histogram from many threads" case in this thread.

dbarker:
> > > I suggest keeping `common::SpinLockMutex` unchanged and using a fixed SDK-internal lock for the affected metrics path. A `std::mutex`, which uses an efficient Windows lock underneath, or a dedicated internal parking lock would solve the measured problem without changing synchronization behavior across the whole project. I would also avoid a CMake option that changes the lock type, because that can create another mixed-layout problem.
> > > Overall: the problem and general direction are good, but I think the current implementation is too broad and has compatibility risks. I’m requesting changes and would prefer the SDK-internal metrics approach.
> > 
> > 
> > Thanks @lalitb for the in depth analysis.
> > I agree with lalit here, using a different lock instead of a spin lock where there is contention for metrics is less intrusive compared to changing the spin lock implementation itself for every uses.
> > A spin lock is supposed to be held for a very short time, not causing contention. If there is contention, it could be because the lock is held for too long, so that a spin lock is not appropriate in the user case found with metrics.
> 
> I think we can scope it down to a single member. The measured hot path is entirely behind one lock:
> 
> https://github.com/open-telemetry/opentelemetry-cpp/blob/5d7db843ea08d6efa7ce7e3f9383a9e6c1f5c157/sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h#L303
> 
> All four `RecordLong()` / `RecordDouble()` overloads serialize on `attribute_hashmap_lock_`, so switching just this one member to `std::mutex` (plus the four `lock_guard` instantiations) covers the "hot histogram from many threads" case in this thread.

I agree, the metrics recording path should use a std::mutex. The complexity of the base2 histogram aggregation and last value aggregation (it has a system call to read the clock) may trigger the 1ms sleep of the spinlock.  

Please take a look at https://github.com/open-telemetry/opentelemetry-cpp/issues/4317 which documents the current SpinLockMutex usage and proposes which cases to switch to std::mutex.

api/include/opentelemetry/common/spin_lock_mutex.h
abhinandan-codes: removed it

api/include/opentelemetry/common/spin_lock_mutex.h
abhinandan-codes: as discussed below, going ahead with std mutex change across metric hot path

api/include/opentelemetry/common/spin_lock_mutex.h
abhinandan-codes: Makes sense, added std mutex approach, leaving spin lock as is

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
@abhinandan-codes 

Thanks for your work and patience on this issue.

Please check CI (format, IWYU) and merge conflicts

lalitb (approved):
LGTM, Thanks @abhinandan-codes. Check the Ci, and also rebase to resolve the merge conflict.

ThomsonTan (approved):

dbarker (approved):
Thanks again for highlighting this issue! Switching to `std::mutex` in the sync metric storage (and elsewhere in the SDK) is necessary for synchronization consistency on all platforms and deployments.

marcalff:
@abhinandan-codes 

Please move this line:

```
#include "opentelemetry/common/spin_lock_mutex.h"
```

under section:

```
#ifdef OPENTELEMETRY_HAVE_METRICS_BOUND_INSTRUMENTS_PREVIEW
```

from sync_metric_storage.cc, to fix IWYU.

PR #4248, opened 2026-07-15
[SDK] Tracer::StartSpan benchmark and optimizations

dbarker:
Fixes # (issue)
The `Tracer::StartSpan` function currently performs some unnecessary dynamic memory allocation (allocating a SpanContext and potentially a DefaultSpan) and copying of global static shared pointers by constructing temporary span contexts (which call TraceState::GetDefault()). 

This latency can be removed without breaking changes. The added benchmark showed a latency reduction of ~11-23% for the range of span creation scenarios (tracer disabled, span with no parent, implicit parent, and explicit parent). For non-sampled spans the latency can be reduced by ~50%. 

## Changes
- Adds a benchmark for Tracer::StartSpan. See [baseline vs optimized benchmark results](https://github.com/open-telemetry/opentelemetry-cpp/commit/9746077ea226751d5d8813ad2bd5d4c07ca5dc5f#diff-1109ff16993b53ecab287b8530b870c918d5276f348cf58e9fd463d9613277c6)
- Implements several optimizations to reduce dynamic memory allocation and shared_ptr copies (9746077ea226751d5d8813ad2bd5d4c07ca5dc5f)
   - Use the `DefaultSpan` as a non-recording span with a stack allocated SpanContext 
   - Allocate one non recording span per tracer on construction to serve as the fallback noop span
   - Updates the SDK span to take a stack allocated SpanContext and the SamplingResult
   - Set the SamplingResult attributes in the SDK Span constructor to avoid locking the span mutex unecessarily
   - Changes the default value of the SpanStartOptions::parent to an empty Context instead of the more expensive SpanContext.
   - use std::make_shared to allocate the span when building with exceptions
   - clean up StartSpan to not instantiate and copy SpanContext if not needed.
   - add GetSpanContext helper to avoid dynamically allocating a DefaultSpan object to get an invalid SpanContext (trace::GetSpan is no longer called).
- Update tracer test cases to check for a non-recording span using the Span::IsRecording method
- Sets sampler added attributes to spans on construction (this is also a correctness fix so these attributes are available to the processor OnStart call). 
   


sdk/test/trace/tracer_benchmark.cc
```diff
@@ -0,0 +1,195 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// clang-format off
+//
+// ~/build/sdk/test/trace/tracer_benchmark --benchmark_repetitions=5 --benchmark_display_aggregates_only=true
```
dbarker: The optimizations in this PR result in a 11-23%  reduction in latency for StartSpan compared to main. 

| Benchmark | Baseline (mean ± stddev) | Optimized (mean ± stddev) | Difference % |
|---|---|---|---|
| `BM_StartSpanTracerDisabled` | 6.85 ± 0.025 ns | 5.24 ± 0.046 ns | **-23.5%** |
| `BM_StartSpan` | 187 ± 1.22 ns | 152 ± 2.74 ns | **-18.7%** |
| `BM_StartSpanWithScope` | 255 ± 1.09 ns | 218 ± 1.75 ns | **-14.5%** |
| `BM_StartSpanWithImplicitParent` | 172 ± 0.45 ns | 153 ± 0.80 ns | **-11.0%** |
| `BM_StartSpanWithExplicitParentContext` | 173 ± 0.34 ns | 154 ± 0.90 ns | **-11.0%** |
| `BM_StartSpanWithExplicitRootContext` | 187 ± 0.68 ns | 148 ± 0.69 ns | **-20.9%** |

sdk/test/trace/sampler_benchmark.cc
```diff
@@ -1,6 +1,57 @@
...
+// BM_AlwaysOnSamplerShouldSample_cv                    0.52 %          0.51 %             5
+// BM_ParentBasedSamplerShouldSample_mean               7.11 ns         7.10 ns            5
+// BM_ParentBasedSamplerShouldSample_median             7.08 ns         7.08 ns            5
+// BM_ParentBasedSamplerShouldSample_stddev            0.086 ns        0.087 ns            5
+// BM_ParentBasedSamplerShouldSample_cv                 1.22 %          1.22 %             5
+// BM_TraceIdRatioBasedSamplerShouldSample_mean         4.01 ns         4.00 ns            5
+// BM_TraceIdRatioBasedSamplerShouldSample_median       4.00 ns         4.00 ns            5
+// BM_TraceIdRatioBasedSamplerShouldSample_stddev      0.032 ns        0.032 ns            5
+// BM_TraceIdRatioBasedSamplerShouldSample_cv           0.80 %          0.80 %             5
+// BM_SpanCreation_mean                                  205 ns          205 ns            5
```
dbarker: Comparison with main and this PR on span sampler benchmark. The NoopSpanCreation case is where the tracer is enabled but the sampler decides to not sample the span (AlwaysOffSampler). 

| Benchmark | Baseline (mean ± stddev) | Optimized (mean ± stddev) | Difference % |
|---|---|---|---|
| `BM_SpanCreation` | 235 ± 0.62 ns | 205 ± 1.00 ns | **-12.8%** |
| `BM_NoopSpanCreation` | 59.3 ± 0.42 ns | 27.7 ± 0.14 ns | **-53.3%** |

api/include/opentelemetry/trace/context.h
```diff
@@ -27,13 +27,14 @@ inline nostd::shared_ptr<Span> GetSpan(const context::Context &context) noexcept
 // stored directly in the context, and returns an invalid SpanContext if neither is present.
 inline SpanContext GetSpanContext(const context::Context &context) noexcept
 {
-  if (!context.HasKey(kSpanKey))
+  const context::ContextValue context_value = context.GetValue(kSpanKey);
```
dbarker, line 30: removed `context.HasKey` call since it just calls `context.GetValue(key)` and checks for the `nostd::monostate`. The span key lookup and context value copy can be done once.

api/include/opentelemetry/trace/span_startoptions.h
```diff
@@ -59,7 +59,7 @@ struct StartSpanOptions
   // - If the `parent` field is not set, the newly created Span will inherit the
   // parent of the currently active Span (if any) in the current context.
   //
-  nostd::variant<SpanContext, context::Context> parent = SpanContext::GetInvalid();
+  nostd::variant<SpanContext, context::Context> parent = context::Context{};
```
dbarker, line 62: For most callers the value of parent is never changed. For callers that set a parent the initial value is often default constructed then replaced.  To keep this change ABI safe we don't add a `nostd::monostate` (which is really what we want) and instead construct the cheapest object by default. The `SpanContext` is a complex object with a `shared_ptr<TraceState>` that is initialized to a global `TraceState` default object. The full `Context` is simpler and only has a `shared_ptr<DataList>` member that is default initialized to null.

sdk/src/trace/tracer.cc
```diff
@@ -55,126 +114,121 @@ nostd::shared_ptr<opentelemetry::trace::Span> Tracer::StartSpan(
...
+  // if the attributes is not nullptr, add attributes to the span.
+  if (sampling_result.attributes)
+  {
+    for (auto &kv : *sampling_result.attributes)
     {
-      for (auto &kv : *sampling_result.attributes)
-      {
-        span->SetAttribute(kv.first, kv.second);
-      }
+      span->SetAttribute(kv.first, kv.second);
```
proost: very tiny improvement; how about adding new ctor to the span for SamplingResult? It can reduce acquiring lock.

dbarker: Good catch. Adding the sampling result attributes through the SDK Span SetAttribute method locks the span mutex unnecessarily. The SDK Span is not public so I've changed the constructor to include the sampling result. It shows a small improvement on the new benchmark.

proost (approved):
LGTM

marcalff (approved):
LGTM

marcalff:
@lalitb @ThomsonTan 

LGTM, but would prefer a second review.
The context in API is a sensitive area, could you review as well ?

sdk/src/trace/tracer.cc
```diff
@@ -37,11 +39,68 @@ namespace sdk
...
+
+}  // namespace
+
 Tracer::Tracer(std::shared_ptr<TracerContext> context,
                std::unique_ptr<InstrumentationScope> instrumentation_scope) noexcept
     : instrumentation_scope_{std::move(instrumentation_scope)},
       context_{std::move(context)},
-      tracer_config_(context_->GetTracerConfigurator().ComputeConfig(*instrumentation_scope_))
+      tracer_config_(context_->GetTracerConfigurator().ComputeConfig(*instrumentation_scope_)),
+      noop_span_{std::make_shared<opentelemetry::trace::DefaultSpan>(
```
ThomsonTan: The constructor now adds an unconditional allocation to a `noexcept` path. If `make_shared` cannot allocate the object/control block, `bad_alloc` escaping this constructor calls `std::terminate()`. This is also repeated for every `Tracer`, although the fallback state is identical.

dbarker: The `bac_alloc` exception case is a tricky one and much larger issue throughout the codebase.
 
This PR reduces the risk of `StartSpan` terminating or causing a segfault. 
  - SpanContext allocation has been removed (bad_alloc could escape before)
  - Failure of the Span nothrow allocation is handled (the span pointer may be null)
  - The noop tracer is no longer dynamically allocated in the hot path 

Terminating on construction of the Tracer is a slight improvement in this regard, but there is much more dynamic allocation to address in TraceState and Samplers (may allocate attributes in the result). 

The noop span is being allocated per tracer is to avoid all tracers/spans in the application hitting the global noop span's atomic reference counter. 

Since this PR reduces the risk of termination in StartSpan I'd like to merge it and follow up with an error handling policy so we can decide how best to address the dynamic allocation and exception escapes. 

What do you think?

marcalff: Regarding memory allocation, I think the constraints set by opentelemetry-cpp in general are not achievable.

Taking `Tracer::StartSpan()` for example:

- (a) it returns a `nostd::shared_ptr<Span>`
- (b) it it marked as `noexcept`
- (c) otel-cpp never returns a `nullptr` but returns a no-op object instead.

I can see (a) + (b) working, but the code will return a `nullptr` on OOM conditions and not raise an exception, breaking (c).

I can see (a) + (c) working, the but code will raise an exception on OOM conditions, breaking (b).

In general, I don't see how (a) + (b) + (c) is possible, either something has to give, or we need to find a very clever code pattern to make this happen.

In this case, the code optimization was to stop allocating a per span noop tracer all the time, to reuse a per tracer noop span instead.

While it can still fail this reduces the number of memory allocations and therefore the probability of failure, making the code more robust in degraded conditions: once a tracer is built, it can no longer fail on OOM when creating a noop span in StartSpan(), where it could fail for every call to StartSpan before.

Before this change, the code was:

```
    // create no-op span with valid span-context.

    auto noop_span = nostd::shared_ptr<opentelemetry::trace::Span>{
        new (std::nothrow)
            opentelemetry::trace::NoopSpan(this->shared_from_this(), std::move(span_context))};
    return noop_span;
```

If I understand correctly, this allocates memory, and on failures returns a `nullptr`, breaking (c), crashing the application anyway the moment the span is used.

Bad alloc is indeed a tricky case, that needs to be resolved in the entire code base, not with local fixes in different places, so it should be addressed in a dedicated PR.

marcalff:
Related:

* #4361

dbarker:
Hi @ThomsonTan and @lalitb, please review when you have chance. Thanks!

PR #4249, opened 2026-07-15
Update OtlpGrpcClientOptions to populate options for gRPC client sharing

### LINKED ISSUE #4239: Update OtlpGrpcClientOptions to populate options for gRPC client sharing
**Problem:**

`OtlpGrpcClientOptions` has only a `= default` constructor  and it initializes all options to null or c++ default values. The exporter specific options classes read environment variables and set the otel-spec defined defaults. This makes it challenging to construct a standalone `OtlpGrpcClient` with spec-compliant defaults when sharing the client among multiple exporters. 

**Desired use case:**

Construct a gRPC client and share it with all three signal exporters. 

```cpp
OtlpGrpcClientOptions client_opts;  // this should read env vars or set spec default values
auto shared_client = OtlpGrpcClientFactory::Create(client_opts);

// use a new constructor of signal options classes copy from the shared client options
// signal specific constructors override options that are not fixed at client creation
OtlpGrpcExporterOptions          trace_opts(client_opts);         
OtlpGrpcMetricExporterOptions    metric_opts(client_opts);  
OtlpGrpcLogRecordExporterOptions log_opts(client_opts);   

auto trace_exporter  = OtlpGrpcExporterFactory::Create(trace_opts, shared_client);
auto metric_exporter = OtlpGrpcMetricExporterFactory::Create(metric_opts, shared_client);
auto log_exporter    = OtlpGrpcLogRecordExporterFactory::Create(log_opts, shared_client);
```

**Proposed change:**

- Add helpers in `otlp_environment.h/cc`, that read the OTLP signal independent environment variables (`OTEL_EXPORTER_OTLP_ENDPOINT`, `OTEL_EXPORTER_OTLP_TIMEOUT`, ...) and fall back to spec defaults (`endpoint=http://localhost:4317`, ...) for all options needed by the grpc client.

- Add a default constructor  to  `OtlpGrpcClientOptions` in a .cc file that uses those generic helpers to populate the options
- Add a void* constructor to `OtlpGrpcClientOptions`  that skips populating the options for use with the void* constructors of signal specific options classes. 
- Add constructors for each  OtlpGrpcClientOptions derived class that takes a OtlpGrpcClientOptions instance and copies options fixed on client creation then overrides signal-specific options as required. See the table below for which options may be overridden when a shared client is used. 

|gRPC Client Option | Exporter Options can override with shared gRPC client? |
|---|---|
| `endpoint` | No |
| SSL fields (7) | No | 
| `credentials` | No | 
| `user_agent` | No | 
| `channel_arguments` | No |
| `max_threads` | No | 
| `compression` | No | 
| Retry policy fields (4) | No | 
| `timeout` | **Yes** | 
| `metadata` | **Yes** | 
| `max_concurrent_requests` | **Yes** |

ML-dev-crypto:
Fixes #4239

`OtlpGrpcClientOptions` previously only had a `= default` constructor,
leaving every option at null/zero regardless of environment variables
or otel-spec defaults. This made it impossible to build a standalone
`OtlpGrpcClient` with spec-compliant defaults for sharing across the
trace/metric/log exporters, as described in the issue.

Changes:
- Added signal-independent (no TRACES_/METRICS_/LOGS_ prefix) env-var
  accessors to `otlp_environment.h`, e.g. `GetOtlpDefaultGrpcClientEndpoint()`.
  These are `inline`, defined in the header rather than
  `otlp_environment.cc`, so that `otlp_grpc_client` (a lean transport-only
  target) doesn't need to depend on `otlp_recordable` and its
  sdk/logs, sdk/metrics, sdk/trace dependencies just to read an env var.
- `OtlpGrpcClientOptions` now has a real default constructor (in a new
  `otlp_grpc_client_options.cc`) that populates from those accessors,
  and a `void*` constructor that skips populating, for parity with the
  existing per-signal options classes.
- `OtlpGrpcExporterOptions` / `OtlpGrpcMetricExporterOptions` /
  `OtlpGrpcLogRecordExporterOptions` each gained a constructor taking a
  `const OtlpGrpcClientOptions&`, copying the client-fixed fields
  (endpoint, SSL/mTLS fields, credentials, user_agent, channel_arguments,
  max_threads, compression, retry policy) and overriding only the fields
  that may still differ per signal (timeout, metadata,
  max_concurrent_requests) — per the table in the issue.
- Build files updated for the new source file and one new Bazel dep
  (`//sdk/src/common:env_variables`); no new library targets, no change
  to the existing dependency graph shape.
- New unit test `otlp_grpc_client_options_test.cc` covering: default
  env-var population, the `void*` no-defaults path (including the
  derived classes' `void*` constructors), fixed-field copying from a
  shared client, and signal-specific env-var precedence over the
  shared client's generic value.

Tested: compiled and unit-tested with both GCC/libstdc++ and
MSVC/STL; `clang-format` applied.


exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -24,8 +24,14 @@ namespace otlp
 
 struct OtlpGrpcClientOptions
 {
-  virtual ~OtlpGrpcClientOptions()                                = default;
-  OtlpGrpcClientOptions()                                         = default;
+  virtual ~OtlpGrpcClientOptions();
```
owent, line 34: In my understanding, if we implement public struct or function in `.cc` files. We need to use `OPENTELEMETRY_EXPORT` to export them. Or the symbols will be hidden and cause link error in some situation.

exporters/otlp/src/otlp_grpc_client_options.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "opentelemetry/exporters/otlp/otlp_environment.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace otlp
+{
+
+OtlpGrpcClientOptions::OtlpGrpcClientOptions()
```
owent, line 14: Should members be initialized in a member initializer of the constructor to avoid [cppcoreguidelines-prefer-member-initializer](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/prefer-member-initializer.html) of clang-tidy?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
```diff
@@ -168,6 +174,226 @@ float GetOtlpDefaultTracesRetryBackoffMultiplier();
...
+  std::string value;
+  opentelemetry::sdk::common::GetStringEnvironmentVariable("OTEL_EXPORTER_OTLP_CERTIFICATE_STRING",
+                                                           value);
+  return value;
+}
+
+inline std::string GetOtlpDefaultGrpcClientSslClientKeyPath()
+{
+  std::string value;
+  opentelemetry::sdk::common::GetStringEnvironmentVariable("OTEL_EXPORTER_OTLP_CLIENT_KEY", value);
```
owent: I'd suggest moving these environment variable definitions to a shared location so both this file and otlp_environment.cc can reference the same constants. This will make maintenance easier going forward.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
```diff
@@ -168,6 +174,226 @@ float GetOtlpDefaultTracesRetryBackoffMultiplier();
...
+  }
+
+  opentelemetry::common::KeyValueStringTokenizer tokenizer{raw_value};
+  opentelemetry::nostd::string_view header_key;
+  opentelemetry::nostd::string_view header_value;
+  bool header_valid = true;
+
+  std::unordered_set<std::string> remove_cache;
+
+  while (tokenizer.next(header_valid, header_key, header_value))
```
owent: Why not export and use `DumpOtlpHeaders` here?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
```diff
@@ -168,6 +174,226 @@ float GetOtlpDefaultTracesRetryBackoffMultiplier();
 float GetOtlpDefaultMetricsRetryBackoffMultiplier();
 float GetOtlpDefaultLogsRetryBackoffMultiplier();
 
+/**
+ * Signal-independent accessors, for use by a gRPC client shared across
+ * multiple signal exporters. These read only the generic OTEL_EXPORTER_OTLP_*
+ * environment variables and fall back to the same spec defaults used above.
+ */
+
+inline std::string GetOtlpDefaultGrpcClientEndpoint()
```
owent: We should move these implementations into otlp_environment.cc to reduce symbol size.
Also, the inline keyword is merely a hint to the compiler, not a mandatory directive. Different compilers may make different inlining decisions under different optimization flags, which can lead to inconsistent symbol generation.
For example, in Release builds the compiler might fully inline these functions and omit external symbols. Later, when a Debug build links against this library, the linker may not resolve those symbols, resulting in undefined references at link time.

exporters/otlp/src/otlp_grpc_exporter_options.cc
owent: Should we use OtlpGrpcClientOptions(nullptr) here and in constructor of metrics and logs options to avoid unnecessary environment parsing?

exporters/otlp/src/otlp_grpc_exporter_options.cc
```diff
@@ -55,6 +55,17 @@ OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(void *)
...
 }
 
+OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  timeout  = GetOtlpDefaultTracesTimeout();
+  metadata = GetOtlpDefaultTracesHeaders();
+
+#ifdef ENABLE_ASYNC_EXPORT
+  max_concurrent_requests = 64;
```
owent: Why reset `max_concurrent_requests` here?

exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc
```diff
@@ -55,6 +56,18 @@ OtlpGrpcLogRecordExporterOptions::OtlpGrpcLogRecordExporterOptions(void *)
...
 
+OtlpGrpcLogRecordExporterOptions::OtlpGrpcLogRecordExporterOptions(
+    const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  timeout  = GetOtlpDefaultLogsTimeout();
+  metadata = GetOtlpDefaultLogsHeaders();
+
+#ifdef ENABLE_ASYNC_EXPORT
+  max_concurrent_requests = 64;
```
owent: Just as above, why reset `max_concurrent_requests` here?

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
ML-dev-crypto: You're right, I missed that. I'll add OPENTELEMETRY_EXPORT.

exporters/otlp/src/otlp_grpc_client_options.cc
ML-dev-crypto: Good point. I'll switch these to use a member initializer list to satisfy cppcoreguidelines-prefer-member-initializer.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
ML-dev-crypto: Agreed. The generic OTLP environment variable names are duplicated now, so I'll move them to a shared location that both the header and implementation can reference.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
ML-dev-crypto: Makes sense. I duplicated the parsing logic because DumpOtlpHeaders isn't currently reusable, but I agree it's better to avoid the duplication. I'll refactor this to reuse the existing implementation.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_environment.h
ML-dev-crypto: Good point. I'll move these helper implementations out of the header and keep them alongside the existing environment accessors while making sure the dependency structure still works.

exporters/otlp/src/otlp_grpc_exporter_options.cc
ML-dev-crypto: Agreed.I'll update this to reuse the base constructor and only override the signal-specific fields.

exporters/otlp/src/otlp_grpc_exporter_options.cc
ML-dev-crypto: Thanks for catching that,The intention was to preserve the shared client's configured value, so I'll update this constructor to copy max_concurrent_requests instead of resetting it.

exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc
ML-dev-crypto: Thanks! I'll make the same change here as well.

ML-dev-crypto:
Thanks for the thorough review, @owent! I'll address all of these comments in a follow-up commit, including the export annotation, constructor initialization cleanup, shared environment variable definitions, removal of the duplicated header parsing logic, moving the new environment helpers out of the header, reusing the base constructor for shared client initialization, and preserving the shared client's `max_concurrent_requests` value. I'll push everything together.

exporters/otlp/src/otlp_environment.cc
```diff
@@ -30,6 +30,27 @@ namespace exporter
 namespace otlp
 {
 
+constexpr char kEnvOtlpEndpoint[]                = "OTEL_EXPORTER_OTLP_ENDPOINT";
```
owent: Before C++17, a constexpr variable declared at namespace scope may have linkage symbol, so it emits an exported symbol. Please declare these as static constexpr char to give them internal linkage and avoid leaking symbols.

exporters/otlp/src/otlp_environment.cc
```diff
@@ -1312,11 +1272,175 @@ float GetOtlpDefaultMetricsRetryBackoffMultiplier()
...
+  }
+
+  return kDefault;
+}
+
+bool GetOtlpDefaultGrpcClientIsInsecure()
+{
+  std::string endpoint = GetOtlpDefaultGrpcClientEndpoint();
+
+  if (endpoint.substr(0, 6) == "https:")
```
owent: Just a suggestion, use 'nostd::string_view' here can slightly improve performance.

owent (changes requested):
Just a few minor non-blocking issues left — otherwise LGTM. Could you please resolve the conflicts?

ML-dev-crypto:
@owent Thanks! Addressed the remaining comments, rebased onto the latest main, resolved the merge conflict, and pushed the updates.

ML-dev-crypto:
@owent 
After rebasing, the CI is failing with linker errors because the generic `GetOtlpDefaultGrpcClient*()` helpers are currently defined in `otlp_environment.cc`, which isn't linked into the `otlp_grpc_client` target. I looked into a few possible approaches (moving the helpers, introducing a shared utility, etc.), but each would require changes to the target structure. Could you please take a look and let me know which direction you would recommend?

owent:
> @owent After rebasing, the CI is failing with linker errors because the generic `GetOtlpDefaultGrpcClient*()` helpers are currently defined in `otlp_environment.cc`, which isn't linked into the `otlp_grpc_client` target. I looked into a few possible approaches (moving the helpers, introducing a shared utility, etc.), but each would require changes to the target structure. Could you please take a look and let me know which direction you would recommend?

Could you please try to add `OPENTELEMETRY_EXPORT` for these function both in .h and .cc files?

ML-dev-crypto:
Thanks for the suggestion! I've updated the PR by adding `OPENTELEMETRY_EXPORT` to the relevant declarations and definitions in both the header and source files, and pushed the changes.

exporters/otlp/src/otlp_grpc_exporter_options.cc
```diff
@@ -55,6 +48,17 @@ OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(void *)
...
 }
 
+OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  timeout  = GetOtlpDefaultTracesTimeout();
+  metadata = GetOtlpDefaultTracesHeaders();
+
+#ifdef ENABLE_ASYNC_EXPORT
+  max_concurrent_requests = client_options.max_concurrent_requests;
```
lalitb: One shared-client edge case: the gRPC client only uses this when its async data is first created. With one client shared by trace/metric/log exporters, the first exporter wins and later per-signal values are ignored.

exporters/otlp/src/otlp_grpc_exporter_options.cc
```diff
@@ -55,6 +48,17 @@ OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(void *)
 #endif
 }
 
+OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  timeout  = GetOtlpDefaultTracesTimeout();
```
lalitb: I think this re-reads env vars instead of falling back to `client_options.timeout`. If the shared client timeout was set programmatically and no trace-specific env var exists, this drops that value and goes back to the default.

exporters/otlp/src/otlp_grpc_client_options.cc
```diff
@@ -0,0 +1,54 @@
...
+    : endpoint(GetOtlpDefaultGrpcClientEndpoint()),
```
lalitb, line 15: I think this still has a standalone client link issue. These helpers are defined in `otlp_environment.cc`, but this new file is compiled into the gRPC client target, which does not link the recordable target where `otlp_environment.cc` lives. `OPENTELEMETRY_EXPORT` fixes visibility, but not the missing target dependency.

exporters/otlp/src/otlp_grpc_exporter_options.cc
ML-dev-crypto: Thanks for catching this! I've updated the implementation to preserve `client_options.timeout` unless a signal-specific timeout override is present, and added a regression test covering the programmatic shared-client timeout case.

exporters/otlp/src/otlp_grpc_exporter_options.cc
ML-dev-crypto: Thanks for pointing this out. I agree this is a limitation of the shared async client rather than the option-copying logic in this PR. I'd prefer to keep this PR focused on the current fixes, but I'm happy to address it in a follow-up PR .

exporters/otlp/src/otlp_grpc_client_options.cc
ML-dev-crypto: Thanks for pointing this out. I looked into a few possible approaches, but they all seem to require changes to the target structure. Would you prefer moving these generic helpers into the gRPC client target, introducing a shared utility target, or another approach? Could you help point me in the right direction?

dbarker: Hi @ML-dev-crypto, introducing a new shared utility target with `otlp_environment.cc` sounds like a good approach.

exporters/otlp/src/otlp_grpc_exporter_options.cc
```diff
@@ -55,6 +48,21 @@ OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(void *)
...
 
+OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  std::chrono::system_clock::duration signal_timeout;
+  if (GetOtlpDefaultTracesTimeoutOverride(signal_timeout))
+  {
+    timeout = signal_timeout;
+  }
+  metadata = GetOtlpDefaultTracesHeaders();
```
ThomsonTan, line 52: The base-class copy on line 52 already preserves `client_options.metadata`, but line 59 immediately discards it and rebuilds the map only from OTLP header environment variables.

With no header environment variables:

```cpp
OtlpGrpcClientOptions client(nullptr);
client.metadata.emplace("authorization", "credential");
OtlpGrpcExporterOptions trace(client);
```

`trace.metadata` becomes empty, so programmatic authentication is silently lost.

Could we preserve the copied map and overlay only `OTEL_EXPORTER_OTLP_TRACES_HEADERS` onto it, with signal-specific values replacing same-name client values? That matches the existing generic-then-signal header precedence. Please apply the same fix to metrics and logs, and add regression tests for preservation without a signal variable plus merge/override behavior when one is present.

owent:
About the undefined reference symbols. The symbols should be declared as visibility=defaut on Unix like system and `__declspec(dllexport)/__attribute__((__dllexport__))` for the target which implement it and declare as `__declspec(dllimport)/__attribute__((__dllimport__))` for other targets on Windows.
Here is a solution:

- Using `project_build_tools_set_shared_library_declaration()` and `project_build_tools_set_static_library_declaration()` to declare symbol rule in cmake.

```cmake
if ("${OPENTELEMETRY_OTLP_TARGETS_LIB_TYPE}" STREQUAL "STATIC" OR NOT BUILD_SHARED_LIBS)
  project_build_tools_set_shared_library_declaration(OPENTELEMETRY_OTLP_RECORDABLE_API
                                                     opentelemetry_otlp_recordable)
else()
  project_build_tools_set_static_library_declaration(OPENTELEMETRY_OTLP_RECORDABLE_API
                                                     opentelemetry_otlp_recordable)
endif()

if ("${OPENTELEMETRY_OTLP_GRPC_CLIENT_LIB_TYPE}" STREQUAL "STATIC" OR NOT BUILD_SHARED_LIBS)
  project_build_tools_set_shared_library_declaration(OPENTELEMETRY_OTLP_GRPC_CLIENT_API
                                                     opentelemetry_exporter_otlp_grpc_client)
else()
  project_build_tools_set_static_library_declaration(OPENTELEMETRY_OTLP_GRPC_CLIENT_API
                                                     opentelemetry_exporter_otlp_grpc_client)
endif()
```

- Declared macro in .h files for bazel(which will always be built as static library)

In `otlp_environment.h`

```cpp
#ifndef OPENTELEMETRY_OTLP_RECORDABLE_API
#  if defined(__clang__)
#    define OPENTELEMETRY_OTLP_RECORDABLE_API __attribute__((visibility("default")))
#  elif defined(__GNUC__)
#    define OPENTELEMETRY_OTLP_RECORDABLE_API __attribute__((visibility("default")))
#  else
#    define OPENTELEMETRY_OTLP_RECORDABLE_API
#  endif
#endif
```

And in `otlp_grpc_client_options.h`

```cpp
#ifndef OPENTELEMETRY_OTLP_GRPC_CLIENT_API
#  if defined(__clang__)
#    define OPENTELEMETRY_OTLP_GRPC_CLIENT_API __attribute__((visibility("default")))
#  elif defined(__GNUC__)
#    define OPENTELEMETRY_OTLP_GRPC_CLIENT_API __attribute__((visibility("default")))
#  else
#    define OPENTELEMETRY_OTLP_GRPC_CLIENT_API
#  endif
#endif
```

- At last, do not use `OPENTELEMETRY_EXPORT` any more, use `OPENTELEMETRY_OTLP_RECORDABLE_API` to declare public functions in `opentelemetry_otlp_recordable` and use use `OPENTELEMETRY_OTLP_GRPC_CLIENT_API` to declare public functions in `opentelemetry_exporter_otlp_grpc_client`.

exporters/otlp/src/otlp_grpc_client_options.cc
dbarker: I've opened PR https://github.com/open-telemetry/opentelemetry-cpp/pull/4333  that creates the otlp_common target and should resolve the linking issues. Once merged we can update this PR.

dbarker: #4333 is merged and addresses the linking issues by moving the otlp_environment.cc to a common target that otlp grcp client links to. That change is now used by this PR's branch.  Resolving this.

ML-dev-crypto:
@owent Should the project_build_tools_set_*_library_declaration() calls target opentelemetry_otlp_common instead? And should I keep OPENTELEMETRY_OTLP_RECORDABLE_API as the macro name, or rename it to  OPENTELEMETRY_OTLP_COMMON_API to match the target?

exporters/otlp/src/otlp_grpc_exporter_options.cc
ML-dev-crypto: Thanks for pointing this out. I’ll preserve the metadata copied from client_options and overlay the signal-specific headers on top, with signal-specific values taking precedence for the same key. I’ll apply the same fix to traces, metrics, and logs and add the regression tests.

owent:
> @owent Should the project_build_tools_set_*_library_declaration() calls target opentelemetry_otlp_common instead? And should I keep OPENTELEMETRY_OTLP_RECORDABLE_API as the macro name, or rename it to OPENTELEMETRY_OTLP_COMMON_API to match the target?

The new target (opentelemetry_otlp_common) is used and I think the macro name should be `OPENTELEMETRY_OTLP_COMMON_API` now.

> BTW: I raise a PR in otel-cpp-contrib repo to use the similar way to export symbols. Which may be helpful: <https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/646>

exporters/otlp/src/otlp_grpc_client_options.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "opentelemetry/exporters/otlp/otlp_environment.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace otlp
+{
+
+OtlpGrpcClientOptions::OtlpGrpcClientOptions()
```
dbarker, line 14: clang-tidy warnings from this file: 

Line | Check | Message
-- | -- | --
22 | cppcoreguidelines-use-default-member-init,modernize-use-default-member-init | member initializer for 'max_threads' is redundant
42 | cppcoreguidelines-prefer-member-initializer | 'use_ssl_credentials' should be initialized in a member initializer of the constructor
43 | cppcoreguidelines-prefer-member-initializer | 'max_threads' should be initialized in a member initializer of the constructor
46 | cppcoreguidelines-prefer-member-initializer | 'max_concurrent_requests' should be initialized in a member initializer of the constructor

exporters/otlp/test/otlp_grpc_client_options_test.cc
```diff
@@ -0,0 +1,153 @@
...
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace exporter
+{
+namespace otlp
+{
+
+namespace
+{
+class ScopedEnvVar
```
dbarker, line 23: clang-tidy warning for the special member functions
```sh
`cppcoreguidelines-special-member-functions`	class 'ScopedEnvVar' defines a non-default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator
```

exporters/otlp/src/otlp_grpc_metric_exporter_options.cc
```diff
@@ -55,6 +57,23 @@ OtlpGrpcMetricExporterOptions::OtlpGrpcMetricExporterOptions(void *)
 #endif
 }
 
+OtlpGrpcMetricExporterOptions::OtlpGrpcMetricExporterOptions(
+    const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options),
```
dbarker: clang-tidy warnings from this file: 

Line | Check | Message
-- | -- | --
50 | cppcoreguidelines-use-default-member-init,modernize-use-default-member-init | member initializer for 'aggregation_temporality' is redundant
63 | cppcoreguidelines-use-default-member-init,modernize-use-default-member-init | member initializer for 'aggregation_temporality' is redundant

exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc
dbarker, line 7: IWYU warning for this file. 

```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc should add these lines:
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc should remove these lines:

The full include-list for /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_log_record_exporter_options.cc:
#include <chrono>
#include <string>
#include "opentelemetry/exporters/otlp/otlp_environment.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_options.h"
---

```

exporters/otlp/src/otlp_grpc_metric_exporter_options.cc
dbarker, line 10: IWYU warning for this file:

```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_metric_exporter_options.cc should add these lines:
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
#include "opentelemetry/exporters/otlp/otlp_preferred_temporality.h"

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_metric_exporter_options.cc should remove these lines:

The full include-list for /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_metric_exporter_options.cc:
#include <chrono>
#include <string>
#include "opentelemetry/exporters/otlp/otlp_environment.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_metric_exporter_options.h"
#include "opentelemetry/exporters/otlp/otlp_preferred_temporality.h"
---
```

exporters/otlp/test/otlp_grpc_client_options_test.cc
```diff
@@ -0,0 +1,153 @@
...
+
+#include <gtest/gtest.h>
+#include <chrono>
+#include <cstdlib>
+#include <string>
+
+#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
+#include "opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h"
+#include "opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_options.h"
+#include "opentelemetry/exporters/otlp/otlp_grpc_metric_exporter_options.h"
```
dbarker, line 12: IWYU warning for this file:

```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/test/otlp_grpc_client_options_test.cc should add these lines:

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/test/otlp_grpc_client_options_test.cc should remove these lines:

The full include-list for /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/test/otlp_grpc_client_options_test.cc:
#include <gtest/gtest.h>
#include <chrono>
#include <cstdlib>
#include <string>
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_log_record_exporter_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_metric_exporter_options.h"
---
```

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
dbarker, line 12: IWYU warning for this file:

```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h should add these lines:
#include <cstddef>                                          // for size_t
#include <cstdint>                                          // for uint32_t

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h should remove these lines:
- #include <memory>  // lines 10-10
- namespace grpc { class ChannelCredentials; }  // lines 15-15

The full include-list for /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h:
#include <chrono>                                           // for duration
#include <cstddef>                                          // for size_t
#include <cstdint>                                          // for uint32_t
#include <string>                                           // for basic_string
#include "opentelemetry/exporters/otlp/otlp_environment.h"  // for OtlpHeaders
#include "opentelemetry/version.h"                          // for OPENTELEM...
namespace grpc { class ChannelArguments; }  // lines 16-16
---
```

exporters/otlp/src/otlp_grpc_exporter_options.cc
dbarker, line 10: IWYU warning for this file: 

```sh

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_exporter_options.cc should add these lines:
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"

/home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_exporter_options.cc should remove these lines:

The full include-list for /home/runner/work/opentelemetry-cpp/opentelemetry-cpp/exporters/otlp/src/otlp_grpc_exporter_options.cc:
#include <chrono>
#include <string>
#include "opentelemetry/exporters/otlp/otlp_environment.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_client_options.h"
#include "opentelemetry/exporters/otlp/otlp_grpc_exporter_options.h"
---

```

dbarker (commented):
Hi @ML-dev-crypto , Could you please resolve the CI issues? Once passing I'll approve. 

Please see notes below on the fixes needed.

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -22,10 +22,16 @@ namespace exporter
 namespace otlp
 {
 
-struct OtlpGrpcClientOptions
+struct OPENTELEMETRY_EXPORT OtlpGrpcClientOptions
```
dbarker, line 32: RE: https://github.com/open-telemetry/opentelemetry-cpp/pull/4249#issuecomment-5231760424

@owent since this class now uses the same export declaration as the otlp exporter options classes that derive from it, would it be reasonable to defer any changes to those definitions to a follow up PR?

owent: The export/import declaration should follow the target. What I mean is:

- Public functions in otlp_environment.cc should be declared as OPENTELEMETRY_OTLP_COMMON_API, since that file is built into the opentelemetry_otlp_common target.
- Public functions in otlp_grpc_client.cc should be declared as OPENTELEMETRY_OTLP_GRPC_CLIENT_API.

As for OtlpGrpcClientOptions, it's fine to leave it without an export declaration — it's a header-only class with no source file. Every target that references it will compile its own copy. This only adds a bit of linker load; there's no visibility issue.

dbarker: > The export/import declaration should follow the target.

That makes sense to me. Since this would be a new policy for the otlp exporter folder (and likely needed for the SDK classes as well), is it reasonable to apply the new policy in follow-up PRs? 

Properly setting the export/import declarations (for all platforms and on the intended public facing classes) is something we need to address broadly throughout the project.

> As for OtlpGrpcClientOptions, it's fine to leave it without an export declaration — it's a header-only class with no source file.

This PR adds the .cc file for `OtlpGrpcClientOptions` and may have prompted this discussion.

owent: > That makes sense to me. Since this would be a new policy for the otlp exporter folder (and likely needed for the SDK classes as well), is it reasonable to apply the new policy in follow-up PRs?

My suggestion is to apply it to all components. If we export symbols this way consistently across all components, `ext/src/dll/input.src` would no longer be needed, and we could keep a uniform symbol-exporting rule across all platforms, compilers, and optimization settings — by setting the default visibility to hidden on Unix-like systems. It would also reduce linker load and speed up linking. (For example, some compiler versions may inline a symbol in Release builds but keep it "imported" in Debug builds, which leads to link errors. This issue has occurred with certain versions of protobuf.)

> This PR adds the .cc file for OtlpGrpcClientOptions and may have prompted this discussion.

Sorry, I missed that. Then it should use something like `OPENTELEMETRY_OTLP_GRPC_CLIENT_API` to export its symbols.

dbarker: Applying the symbol visibility attributes to all components sounds like a nice improvement.  

The grpc client class doesn't have symbol visibility attributes now so the change must be broader than this PR should take on. 

Let's merge this once CI is passing and address the visibility attributes in a new PR that fixes the grpc client as well.

dbarker: @owent, This PR should pass CI now. Please take a review pass (your review is still requesting changes).

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
```diff
@@ -35,7 +48,7 @@ struct OtlpGrpcClientOptions
   std::string endpoint;
 
   /** Use SSL. */
-  bool use_ssl_credentials{};
+  bool use_ssl_credentials{true};
```
owent, line 51: Not a problem, but I think it would more clear to make all member initialization in one place(Maybe in `OtlpGrpcClientOptions::OtlpGrpcClientOptions` of `otlp_grpc_client_options.cc`).

exporters/otlp/src/otlp_environment.cc
```diff
@@ -1312,11 +1293,175 @@ float GetOtlpDefaultMetricsRetryBackoffMultiplier()
...
+OPENTELEMETRY_EXPORT bool GetOtlpDefaultGrpcClientIsInsecure()
+{
+  std::string endpoint = GetOtlpDefaultGrpcClientEndpoint();
+
+  if (opentelemetry::nostd::string_view(endpoint).substr(0, 6) == "https:")
+  {
+    return false;
+  }
+
+  if (endpoint.substr(0, 5) == "http:")
```
owent, line 1330: It still allocate new std::string, please use `opentelemetry::nostd::string_view(endpoint).substr(0, 5)` here.

exporters/otlp/src/otlp_grpc_exporter_options.cc
```diff
@@ -28,31 +29,27 @@ OtlpGrpcExporterOptions::OtlpGrpcExporterOptions()
...
-#endif
+OtlpGrpcExporterOptions::OtlpGrpcExporterOptions(const OtlpGrpcClientOptions &client_options)
+    : OtlpGrpcClientOptions(client_options)
+{
+  std::chrono::system_clock::duration signal_timeout;
+  if (GetOtlpDefaultTracesTimeoutOverride(signal_timeout))
+  {
+    timeout = signal_timeout;
+  }
+  metadata = GetOtlpDefaultTracesHeaders();
```
owent, line 52: metadata in client_options should be restore.

owent (changes requested):

exporters/otlp/include/opentelemetry/exporters/otlp/otlp_grpc_client_options.h
owent: @dbarker The PR looks close to ready. I noticed a few remaining issues, though most are just nits. The one I'd consider blocking is here: https://github.com/open-telemetry/opentelemetry-cpp/pull/4249#discussion_r3669585881 — it looks like the metadata fix discussed there didn't make it into the current head commit.

PR #4191, opened 2026-06-30
[API/SDK] feat: exemplar filters

proost:
Fixes
* https://github.com/open-telemetry/opentelemetry-cpp/issues/4178
* https://github.com/open-telemetry/opentelemetry-cpp/issues/2526

## Changes

Complete exemplar filter.

This changes accompany breaking change. When I read [versioning policy](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/Versioning.md), flagged feature is unstable, so API breaking change is accepted, right?

I tried to abstract filte logic into `FilteredExemplarReservoir` to avoid filtering logic scattered to metric storage.

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4254, opened 2026-07-16
[API] reduce allocation in no span case

proost:
## Changes

Comment: https://github.com/open-telemetry/opentelemetry-cpp/pull/4191#discussion_r3589490320

Reducing allocation in the no span case. Because a new default span `return nostd::shared_ptr<Span>(new DefaultSpan(SpanContext::GetInvalid()));` is created when we get context form `GetSpan`. 


api/include/opentelemetry/trace/propagation/http_trace_context.h
```diff
@@ -41,7 +41,7 @@ class HttpTraceContext : public context::propagation::TextMapPropagator
   void Inject(context::propagation::TextMapCarrier &carrier,
               const context::Context &context) noexcept override
   {
-    SpanContext span_context = trace::GetSpan(context)->GetContext();
+    SpanContext span_context = trace::GetSpanContext(context);
```
dbarker, line 44: Not an ABI break. No observable behavior change.

dbarker (approved):
Looks good to me.  Thanks for the PR!

It needs additional review and approval as this is an expansion of the API.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
Do not merge:

Release 1.28.0 needs to be published first, and this PR can be merged after.

This is to limit last minute changes to API just when releasing.

api/include/opentelemetry/trace/context.h
```diff
@@ -23,6 +23,33 @@ inline nostd::shared_ptr<Span> GetSpan(const context::Context &context) noexcept
...
+  if (!context.HasKey(kSpanKey))
+  {
+    return SpanContext::GetInvalid();
+  }
+
+  const context::ContextValue context_value = context.GetValue(kSpanKey);
+
+  // Get the span metadata from the active span in the context.
+  if (const nostd::shared_ptr<Span> *maybe_span =
+          nostd::get_if<nostd::shared_ptr<Span>>(&context_value))
```
ThomsonTan, line 39: Could the active variant alternative contain an empty `nostd::shared_ptr<Span>` here? `nostd::get_if` would still return a non-null pointer to the stored `shared_ptr`, so this condition would succeed even though `*maybe_span` is null, and the next line would dereference it.

The same issue applies to `maybe_span_context` below. Should both branches also check that the stored `shared_ptr` is non-null and return `SpanContext::GetInvalid()` otherwise?

proost: You are right, Thank you! 
[7d7b79a](https://github.com/open-telemetry/opentelemetry-cpp/pull/4254/commits/7d7b79a386ed866b2f647458c14d1cbabc5bf816)

marcalff: On one hand, I am not sure if it is possible to add an empty `nostd::shared_ptr<Span>` into a context, so this extra check might not be needed.

On the other hand, now that this logic is in the API, being defensive and extra cautious is better ... especially because it needs to be robust to any version of the API/SDK.

So, this is a good change.

dbarker: It is possible for the context to contain a  span key and null span ptr value by passing a null span  to SetSpan or when constructing a Scope object (neither check). Given the `trace::GetSpan(context)->GetContext()` pattern used before it seems it has not been a problem in practice but adding a check here shouldn't hurt.

PR #4232, opened 2026-07-09
[SDK] Span limits configuration

dbarker:
Fixes #4046 

Add support for programmatic and yaml configuration of span limits through the TracerProvider. Apply span limits on the SpanData recordable and update the OtlpRecordable to use the configured limits (deprecating the limits set in otlp exporter options classes). 

## Changes

- Adds SpanLimits to the TracerProvider
- Updates the sdk builder to configure the span limits and test
- Adds span limit support in SpanData and OtlpRecordable with tests
- Adds tests to the TracerProvider, Tracer, SpanData, and OtlpRecordable
- (TODO) Deprecates the OtlpGrpcExporterOptions and OltpHttpExporterOptions span limit fields 

For significant contributions please make sure you have completed the following items:

* [x] `CHANGELOG.md` updated for non-trivial changes
* [x] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4258, opened 2026-07-16
[SDK] Read span limits from environment variables

### LINKED ISSUE #4257: Read span limits from environment variables (OTEL_SPAN_EVENT_COUNT_LIMIT etc.)

The SDK does not read the [span-limit environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#span-limits) (`OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`, `OTEL_SPAN_EVENT_COUNT_LIMIT`, `OTEL_SPAN_LINK_COUNT_LIMIT`, `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`, `OTEL_LINK_ATTRIBUTE_COUNT_LIMIT`, and the general `OTEL_ATTRIBUTE_*` fallbacks). Since #4232 the SDK has full `SpanLimits` support via the `TracerProvider` (programmatic and declarative configuration), but deployments that configure limits through environment variables — the common pattern in Kubernetes, and what the other SDKs (Java/Go/Python/…) support — silently get no limits.

Concrete motivation: gRPC's C++ OpenTelemetry plugin adds span events per message and per TCP write for the lifetime of a streaming RPC. Long-lived streaming spans grow unboundedly, and one oversized OTLP export batch is rejected by a collector's default 4 MiB gRPC receive limit (`RESOURCE_EXHAUSTED: grpc: received message larger than max`), dropping every span in the batch. Operators reaching for `OTEL_SPAN_EVENT_COUNT_LIMIT` today find it is a no-op ([spec compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables)).

Mirror the existing `batch_span_processor_options_env` pattern (`OTEL_BSP_*`): a `span_limits_env::GetSpanLimitsFromEnv()` helper that builds a `SpanLimits` from the environment, used as the default by the `TracerProviderFactory::Create` overloads that currently default to `SpanLimits::NoLimits()`.

To preserve the backward-compatible behavior chosen in #4232 (no limits unless explicitly configured):
- a field whose environment variable is unset stays at its `NoLimits()` value — with no environment variables set, behavior is unchanged;
- a set variable applies to its field, with the span-specific variables taking precedence over the general `OTEL_ATTRIBUTE_*` ones per the spec;
- explicitly passed `SpanLimits` (programmatic or declarative configuration) continue to take precedence over the environment, matching the spec's configuration-precedence order.

I have a working implementation following the repo conventions and can open a PR.

- Applying the spec defaults (128) whenever any variable is set: more spec-literal, but surprising — setting only `OTEL_SPAN_EVENT_COUNT_LIMIT` would suddenly cap attributes too, and #4232 deliberately kept `NoLimits()` as the SDK default.
- Reading the environment inside `SpanLimits`' default member initializers (like `BatchSpanProcessorOptions`): conflicts with `SpanLimits{}` already meaning "spec defaults" and would make the YAML/declarative path environment-dependent.
- Doing nothing: every consumer needing env-based limits must parse the variables themselves and thread them through, which the spec intended the SDK to do.

**Additional context**

Follow-up to #4046 / #4232. Once merged, the C++ column of the spec compliance matrix env-variables section can flip for the span-limit rows (separate PR to the specification repo).

HrMathematiker:
Fixes #4257

## Changes

Adds `span_limits_env::GetSpanLimitsFromEnv()`, which builds a `SpanLimits` from the [environment variables defined in the specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#span-limits), and uses it as the default in the `TracerProviderFactory::Create` overloads that previously defaulted to `SpanLimits::NoLimits()`. The implementation mirrors the existing `batch_span_processor_options_env` pattern (`OTEL_BSP_*`).

Semantics, chosen to preserve the backward-compatible default established in #4232:

- A field whose environment variable is unset keeps its `SpanLimits::NoLimits()` value — with no variables set, `GetSpanLimitsFromEnv() == SpanLimits::NoLimits()` and SDK behavior is unchanged.
- Supported variables: `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT`, `OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT`, `OTEL_SPAN_EVENT_COUNT_LIMIT`, `OTEL_SPAN_LINK_COUNT_LIMIT`, `OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT`, `OTEL_LINK_ATTRIBUTE_COUNT_LIMIT`, plus the general `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` / `OTEL_ATTRIBUTE_COUNT_LIMIT`, with the span-specific variables taking precedence per the spec.
- Explicitly passed `SpanLimits` (programmatic, or declarative configuration through the sdk builder) continue to take precedence over the environment, matching the spec's configuration-precedence order.

Follow-up once merged: flip the C++ cells of the span-limit env-variable rows in the [spec compliance matrix](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables) (separate PR to the specification repo).

### DIFF
--- a/sdk/src/trace/span_limits.cc
+++ b/sdk/src/trace/span_limits.cc
@@ -51,9 +51,9 @@ void UpdateFromEnv(const char *env_var_name, std::size_t &limit)
 
 }  // namespace
 
-SpanLimits GetSpanLimitsFromEnv()
+SpanLimits GetSpanLimitsFromEnv(const SpanLimits &defaults)
 {
-  SpanLimits limits = SpanLimits::NoLimits();
+  SpanLimits limits = defaults;
 
   // General attribute limits first; the span-specific variables below take precedence.
   UpdateFromEnv(kAttributeValueLengthLimitEnv, limits.attribute_value_length_limit);
--- a/sdk/src/trace/span_limits.cc
+++ b/sdk/src/trace/span_limits.cc
@@ -1,0 +1,75 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include <cstddef>
+#include <cstdint>
+
+#include "opentelemetry/sdk/common/env_variables.h"
+#include "opentelemetry/sdk/trace/span_limits.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace trace
+{
+namespace span_limits_env
+{
+
+// Environment variable names, see
+// https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#span-limits
+static constexpr const char *kAttributeValueLengthLimitEnv = "OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT";
+static constexpr const char *kAttributeCountLimitEnv       = "OTEL_ATTRIBUTE_COUNT_LIMIT";
+static constexpr const char *kSpanAttributeValueLengthLimitEnv =
+    "OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT";
+static constexpr const char *kSpanAttributeCountLimitEnv  = "OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT";
+static constexpr const char *kSpanEventCountLimitEnv      = "OTEL_SPAN_EVENT_COUNT_LIMIT";
+static constexpr const char *kSpanLinkCountLimitEnv       = "OTEL_SPAN_LINK_COUNT_LIMIT";
+static constexpr const char *kEventAttributeCountLimitEnv = "OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT";
+static constexpr const char *kLinkAttributeCountLimitEnv  = "OTEL_LINK_ATTRIBUTE_COUNT_LIMIT";
+
+namespace
+{
+
+void UpdateFromEnv(const char *env_var_name, std::uint32_t &limit)
+{
+  std::uint32_t value{};
+  if (opentelemetry::sdk::common::GetUintEnvironmentVariable(env_var_name, value))
+  {
+    limit = value;
+  }
+}
+
+void UpdateFromEnv(const char *env_var_name, std::size_t &limit)
+{
+  std::uint32_t value{};
+  if (opentelemetry::sdk::common::GetUintEnvironmentVariable(env_var_name, value))
+  {
+    limit = static_cast<std::size_t>(value);
+  }
+}
+
+}  // namespace
+
+SpanLimits GetSpanLimitsFromEnv()
+{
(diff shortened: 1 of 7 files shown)


ThomsonTan:
@HrMathematiker thanks for the contribution. Please take a look on the missing CLA authorization.

marcalff:
@HrMathematiker 

Thanks for the contribution, this part is really needed in opentelemetry-cpp.

Please clear the EasyCLA, merging can not happen without it.

HrMathematiker:
@marcalff the CLI is authorized now :) 
Could you please review the MR?

owent (approved):

sdk/src/trace/span_limits.cc
```diff
@@ -0,0 +1,75 @@
...
+  {
+    limit = static_cast<std::size_t>(value);
+  }
+}
+
+}  // namespace
+
+SpanLimits GetSpanLimitsFromEnv()
+{
+  SpanLimits limits = SpanLimits::NoLimits();
```
dbarker: this can move to the function args and default to NoLimits, then users can set their own fallback defaults if the environment variables are not set.

dbarker (approved):
Looks good to me. Thanks for the contribution. Just optional feedback below.

sdk/src/trace/span_limits.cc
HrMathematiker: Applied, thanks

sdk/src/trace/span_limits.cc
```diff
@@ -0,0 +1,75 @@
...
+static constexpr const char *kEventAttributeCountLimitEnv = "OTEL_EVENT_ATTRIBUTE_COUNT_LIMIT";
+static constexpr const char *kLinkAttributeCountLimitEnv  = "OTEL_LINK_ATTRIBUTE_COUNT_LIMIT";
+
+namespace
+{
+
+void UpdateFromEnv(const char *env_var_name, std::uint32_t &limit)
+{
+  std::uint32_t value{};
+  if (opentelemetry::sdk::common::GetUintEnvironmentVariable(env_var_name, value))
```
lalitb, line 37: Could we also update the Bazel deps for this new env parsing code? CMake includes the new source file, but Bazel will pick it up too and needs the matching `env_variables` dependency.

lalitb (approved):
LGTM with comment on bazel build. Thanks

sdk/src/trace/span_limits.cc
HrMathematiker: Done, thanks

PR #4264, opened 2026-07-17
[API] make EnvironmentCarrier spec-compliant (#4190)

### LINKED ISSUE #4190: Make environment variable propagation spec compliant - continuation

Towards: https://github.com/open-telemetry/opentelemetry-specification/issues/5040

The current implementation of environment variable propagation carrier is not specification compliant: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/env-carriers.md

At quick glance:
- (normative) empty name normalized to `_`: https://github.com/open-telemetry/opentelemetry-specification/pull/5163
- (non-normative) caching should be rather removed as it introduces overhead: https://github.com/open-telemetry/opentelemetry-specification/pull/5179 
- (normative) document [operational guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/env-carriers.md#operational-guidance), including initialization-time extraction, child process environment handling, and security considerations.

**Additional context**

Follows https://github.com/open-telemetry/opentelemetry-cpp/pull/4141

om7057:
## Summary

Fixes #4190.

Aligns `EnvironmentCarrier` with the OpenTelemetry environment carriers specification.

## Changes

This PR brings `EnvironmentCarrier` into compliance with the OpenTelemetry environment carriers specification by making the following changes:

1. **Empty key normalization (normative, spec#5163)**
   - `NormalizeKey("")` now returns `"_"` as required by the specification.
   - This implements the normative requirement:
     > "An empty environment variable name is non-normalized and normalizes to `_`."

2. **Operational guidance documentation (non-normative)**
   - Added documentation covering:
     - Initialization-time extraction
     - Child process environment handling
     - Security considerations

## Tests

Updated unit tests to verify the new behavior:

- **`NormalizeKeyEmpty`**
  - Verifies that `Set("", value)` stores the value under the `"_"` key.

- **`GetEmptyKeyNormalizesToUnderscore`**
  - Verifies that `Get("")` retrieves the value from the `"_"` environment variable.

---

### Checklist

- [ ] `CHANGELOG.md` updated (if required for this change)
- [x] Unit tests added
- [x] Public API changes reviewed

### DIFF
--- a/api/include/opentelemetry/context/propagation/environment_carrier.h
+++ b/api/include/opentelemetry/context/propagation/environment_carrier.h
@@ -27,12 +27,35 @@ namespace propagation
 // The carrier supports two usage scenarios:
 //
 // 1. Extract (default constructor): Reads context from environment variables.
-//    Get() reads from TRACEPARENT, TRACESTATE, BAGGAGE environment variables.
-//    Set() is a no-op. Values are cached on first access for lifetime management.
+//    Get() reads directly from the process environment on every call. Callers should
+//    create a carrier and call Extract() once during process initialization and retain
+//    the resulting Context for the lifetime of the process or task. Keys are normalized
+//    to environment variable names before lookup (e.g. "traceparent" -> "TRACEPARENT").
+//    Set() is a no-op.
 //
 // 2. Inject (shared_ptr constructor): Writes context to a provided map.
-//    Set() writes to the provided std::map. Keys are automatically converted
-//    from lowercase header names to uppercase environment variable names.
+//    Set() writes to the provided std::map. Keys are automatically converted from
+//    lowercase header names to uppercase environment variable names. The resulting map
+//    is then typically merged into the environment of a child process before spawning.
+//
+// Operational guidance (non-normative):
+//
+// - Initialization-time extraction: Extract context once at program startup and thread
+//   the resulting Context through the rest of the application. Do not repeatedly call
+//   Get() on a long-lived carrier; environment variables are intended to be set by the
+//   parent before this process starts and should not change during process lifetime.
+//
+// - Child process environment handling: To propagate context to a child process, create
+//   a carrier with an env_map, call Inject() to populate the map, then merge the map
+//   into the child process environment (e.g., via execve/posix_spawn env, or equivalent
+//   platform APIs) before spawning. Create separate maps for child processes with
+//   different contexts.
+//
+// - Security considerations: Environment variables are visible to all code within the
+//   process and may be visible to other processes or users with sufficient permissions.
+//   Do not propagate sensitive information (credentials, secrets) via environment
+//   variable carriers. In multi-tenant environments, consider the extra exposure risk
+//   of environment variables being accessible to co-located processes.
 
 class EnvironmentCarrier : public TextMapCarrier
 {
@@ -45,8 +68,16 @@ public:
       : env_map_ptr_(std::move(env_map))
   {}
 
-  // Returns the value associated with the passed key.
-  // The key is normalized per spec before lookup and caching.
+  // Returns the value of the environment variable corresponding to the normalized form of key.
+  // The key is normalized per spec before lookup (e.g. "traceparent" -> "TRACEPARENT").
+  // Returns an empty string_view if the variable is not set.
+  //
+  // The returned string_view is valid for the lifetime of this carrier object.
+  //
+  // Note: values are cached on first access. This is safe because environment variables used
+  // for context propagation are set by the parent process before this process starts and are
+  // not expected to change during process lifetime. Caching also ensures that the returned
+  // string_view remains valid for the lifetime of the carrier, regardless of when it is read.
   nostd::string_view Get(nostd::string_view key) const noexcept override
   {
     std::string env_name = NormalizeKey(key);
@@ -84,13 +115,23 @@ private:
   mutable std::map<std::string, std::string> cache_;
 
   // Normalizes a key to an environment variable name per the OTel spec:
+  //   - empty key is normalized to "_"
   //   - ASCII letters are uppercased
   //   - characters that are not ASCII letters, digits, or '_' are replaced with '_'
   //   - a leading '_' is prepended if the first character is a digit
-  // e.g., "traceparent" -> "TRACEPARENT", "x-b3-traceid" -> "X_B3_TRACEID",
-  //        "1bad" -> "_1BAD"
+  // Examples:
+  //   ""            -> "_"
+  //   "traceparent" -> "TRACEPARENT"
+  //   "x-b3-traceid"-> "X_B3_TRACEID"
+  //   "1bad"        -> "_1BAD"
   static std::string NormalizeKey(nostd::string_view key)
   {
+    // Spec: empty key normalizes to "_"
+    if (key.empty())
+    {
+      return "_";
+    }
+
     std::string result(key);
     for (auto &c : result)
     {
@@ -104,7 +145,7 @@ private:
         c = '_';
       }
     }
-    if (!result.empty() && result[0] >= '0' && result[0] <= '9')
+    if (result[0] >= '0' && result[0] <= '9')
     {
       result.insert(result.begin(), '_');
     }
--- a/api/test/context/propagation/environment_carrier_test.cc
+++ b/api/test/context/propagation/environment_carrier_test.cc
@@ -103,7 +103,8 @@ TEST_F(EnvironmentCarrierTest, GetCachesValues)
   auto value1 = carrier.Get("traceparent");
   EXPECT_EQ(value1, "00-4bf92f3577b34da6a3ce929d0e0e4736-0102030405060708-01");
 
-  // Change environment - cached value should be returned
+  // Cached value is returned even after the environment variable changes.
+  // This ensures the returned string_view remains valid for the lifetime of the carrier.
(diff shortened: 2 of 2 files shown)


om7057:
CC @pellared @pranitaurlam: This PR addresses the three items outlined in #4190:

1. Empty key normalizes to `_` (normative, spec#5163)
2. Removed caching semantics from `Get()` (non-normative, spec#5179)
3. Added operational guidance doc comments (non-normative)

api/test/context/propagation/environment_carrier_test.cc
```diff
@@ -103,8 +103,9 @@ TEST_F(EnvironmentCarrierTest, GetCachesValues)
   auto value1 = carrier.Get("traceparent");
   EXPECT_EQ(value1, "00-4bf92f3577b34da6a3ce929d0e0e4736-0102030405060708-01");
 
-  // Change environment - cached value should be returned
-  test_setenv("TRACEPARENT", "changed-value");
+  // Without caching, a second Get reflects the current environment state.
+  // (Per the spec, the carrier is intended to be used for a single Extract() at
+  // initialization time; repeated Get() calls on a carrier are not a typical usage pattern.)
```
lalitb: I think this test should change `TRACEPARENT` before the second `Get()`. As written, it would still pass with the old caching behavior, so it does not really cover the “no stale cache” change.

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -45,24 +68,28 @@ class EnvironmentCarrier : public TextMapCarrier
       : env_map_ptr_(std::move(env_map))
   {}
 
-  // Returns the value associated with the passed key.
-  // The key is normalized per spec before lookup and caching.
+  // Returns the value of the environment variable corresponding to the normalized form of key.
+  // The key is normalized per spec before lookup (e.g. "traceparent" -> "TRACEPARENT").
+  // Returns an empty string_view if the variable is not set.
+  //
+  // The returned string_view is valid for the lifetime of this carrier object.
```
lalitb, line 75: This is not entirly true. Maybe we can say "it is valid until the next `Get()` for the same key.

api/test/context/propagation/environment_carrier_test.cc
om7057: Yes, I verified that the test wasn't actually proving the no-caching behavior since the env var never changed between calls. Updated the test to set TRACEPARENT to a different value before the second Get(), so it now fails with the old caching implementation and only passes with the new behavior.

api/include/opentelemetry/context/propagation/environment_carrier.h
om7057: Agreed, that was an overstatement. Updated the comment to say "valid until the next Get() call for the same key", which is the accurate bound since value_store_ gets overwritten on each call for the same key.

om7057:
@lalitb I've resolved the comments, please go through the changes once again.
Thank you!

api/include/opentelemetry/context/propagation/environment_carrier.h
marcalff: In a multi threaded process, this is very dangerous:

- Thread 1 calls get, and is given a string view.
- Something calls setenv() to change a value for a given key
- Thread 2 calls get, which refreshes the value
- Now thread 1 runs on fumes and crashes, as the underlying memory storage for the string view is gone.

marcalff:
@om7057 

Per spec PR #5179:

> Language implementations remain free to use internal storage when required by their API shape.

I propose to keep the cache here, for the sake of providing a safe storage to support the `std::string_view` returned by 
`Get()`.

In practice, caching the result of `getenv()`, or not caching it and calling `getenv()` all the time, should not make any difference: propagation using environment variables is used when a parent process forks a child process, I don't see how and why anyone would want to call `setenv()` after the child process is started.

What matters is that the code can not be abused to cause a crash, so there must be stable memory to back up the `std::string_view` returned, in all cases, hence: we cache and never call `getenv()` again.

api/include/opentelemetry/context/propagation/environment_carrier.h
```diff
@@ -81,16 +108,28 @@ class EnvironmentCarrier : public TextMapCarrier
 
 private:
   std::shared_ptr<std::map<std::string, std::string>> env_map_ptr_;
-  mutable std::map<std::string, std::string> cache_;
+  // Backing storage for string_views returned by Get(). Maps normalized key -> last-read value.
+  // Values are refreshed on every Get() call so that the current environment is always reflected.
+  mutable std::map<std::string, std::string> value_store_;
```
marcalff: Please revert, this is not safe because the returned string_view given by Get() must point to valid memory at all time.

marcalff (changes requested):
Thanks for the patch.

Ok with the normalization of "" -> "_".

Please keep the caching logic as is.

om7057:
> @om7057
> 
> Per spec PR #5179:
> 
> > Language implementations remain free to use internal storage when required by their API shape.
> 
> I propose to keep the cache here, for the sake of providing a safe storage to support the `std::string_view` returned by `Get()`.
> 
> In practice, caching the result of `getenv()`, or not caching it and calling `getenv()` all the time, should not make any difference: propagation using environment variables is used when a parent process forks a child process, I don't see how and why anyone would want to call `setenv()` after the child process is started.
> 
> What matters is that the code can not be abused to cause a crash, so there must be stable memory to back up the `std::string_view` returned, in all cases, hence: we cache and never call `getenv()` again.

Agreed. Reverted to the original cache_ with first-read-wins semantics. The spec PR #5179 explicitly allows internal storage for API shape reasons, and the cache is exactly the right tool here, stable memory for the returned string_view, with no risk of it being invalidated.

api/include/opentelemetry/context/propagation/environment_carrier.h
om7057: Reverted. cache_ is restored with the original first-read-wins semantics.

api/include/opentelemetry/context/propagation/environment_carrier.h
om7057: @marcalff Ok, I understood that overwriting value_store_ on every call would allow Thread 2's Get() to invalidate a string_view still held by Thread 1, causing a crash. The first-read-wins cache_ avoids this entirely since cached strings are never modified or destroyed until the carrier itself is destroyed. Reverted accordingly.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
@om7057 

Approved.

Please update the change set comments for this PR, to reflect the final code.

om7057:
> @om7057
> 
> Approved.
> 
> Please update the change set comments for this PR, to reflect the final code.

@marcalff Done; rewrote the commit history for this branch in a clean commit. Thanks for the review and approval!

om7057:
@lalitb @marcalff All the checks have been passed after the approval; kindly merge if you feel it's ready
Thank you.

ThomsonTan:
This changes a public carrier's normalization behavior and documentation, probably add an entry to the CHANGELOG.

marcalff:
> > @om7057
> > Approved.
> > Please update the change set comments for this PR, to reflect the final code.
> 
> @marcalff Done; rewrote the commit history for this branch in a clean commit. Thanks for the review and approval!

Some clarifications here:

we merge using a smash commit anyway, so multiple commits are not visible, rewriting the commit history is not required.

What I meant to fix is this comment:

> Remove caching semantics (non-normative, spec#5179): Get() previously cached the first getenv() result and served stale values on subsequent calls for the same key. The carrier now calls getenv() on every Get() invocation so the current environment state is always reflected. A value_store_ map is retained only to own the memory backing the returned string_view (required by the TextMapCarrier contract), not to serve cached data.

in the PR description itself.

marcalff:
> This changes a public carrier's normalization behavior and documentation, probably add an entry to the CHANGELOG.

I will add an entry when making the next release, every single commit merged gets an entry in CHANGELOG upon release.

ThomsonTan (approved):

om7057:
> > > @om7057
> > > Approved.
> > > Please update the change set comments for this PR, to reflect the final code.
> > 
> > 
> > @marcalff Done; rewrote the commit history for this branch in a clean commit. Thanks for the review and approval!
> 
> Some clarifications here:
> 
> we merge using a smash commit anyway, so multiple commits are not visible, rewriting the commit history is not required.
> 
> What I meant to fix is this comment:
> 
> > Remove caching semantics (non-normative, spec#5179): Get() previously cached the first getenv() result and served stale values on subsequent calls for the same key. The carrier now calls getenv() on every Get() invocation so the current environment state is always reflected. A value_store_ map is retained only to own the memory backing the returned string_view (required by the TextMapCarrier contract), not to serve cached data.
> 
> in the PR description itself.

@marcalff 
Thanks for clarifying! Updated the PR description to remove the stale caching paragraph. It now reflects the final code.
Feel free to merge whenever you're ready!

om7057:
@ThomsonTan Thanks for the approval!

om7057:
@marcalff, @lalitb, @ThomsonTan, just a small reminder regarding this PR: please check if it's ready to merge.

om7057:
Thank you so much, @marcalff, for merging the PR.
Happy to contribute to the spec-compliant part of the repo.
Looking forward to contribute further in otel-ecosystem.

PR #4266, opened 2026-07-17
fix(ext): enable TCP keepalive for curl HTTP client connections

### LINKED ISSUE #3333: Crash in http exporter, for tcp half open connections
**Steps to reproduce**
Sudden bring down of otel collector in a high load setup.

Clients handling export failures

Hang on export, 30+ seconds. We can see one connection stuck in half open, one stuck in connect SYN_SENT... note that server is gone for some time at this point.

netstat -nap | grep 4318
tcp        0      0 192.168.67.174:51510    192.168.10.101:4318     ESTABLISHED 23/smfcc
tcp        0      1 192.168.67.174:43912    192.168.10.101:4318     SYN_SENT    -

**Additional context**
We suspect this crash is due to otel not providing way to set keep alives on client connections (CURLOPT_TCP_KEEPALIVE)... we've set timeout (CURLOPT_TIMEOUT) to 5s, and ideally for that CURLOPT_TCP_KEEPALIVE should be set to 1s, but otel doesn't have an option to set that. 

Request that a config for TCP_KEEPALIVE be provided.

Babul422:
Fixes #3333

## Description

This PR fixes #3333 by enabling TCP keepalive on libcurl HTTP client connections.

### Root Cause

When a remote endpoint silently disappears without closing the TCP connection, the client can remain connected to a half-open socket for an extended period. Since TCP keepalive is currently disabled, the exporter may remain blocked waiting on the connection.

### Solution

Enable `CURLOPT_TCP_KEEPALIVE` during `HttpOperation::Setup()`. This allows libcurl and the underlying operating system to detect stale TCP connections using TCP keepalive mechanisms.

### Testing

<ul>
<li>Successfully built the project.</li>
<li>curl_http_test passes.</li>
<li>otlp_http_exporter_test passes.</li>
</ul>

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1162,6 +1162,12 @@ CURLcode HttpOperation::Setup()
     return rc;
   }
 
+  rc = SetCurlLongOption(CURLOPT_TCP_KEEPALIVE, 1L);
+  if (rc != CURLE_OK)
+  {
+    return rc;
+  }
+
   if (reuse_connection_)
   {
     rc = SetCurlLongOption(CURLOPT_FRESH_CONNECT, 0L);
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1219,6 +1219,12 @@ CURLcode HttpOperation::Setup()
     return rc;
   }
 
+  rc = SetCurlLongOption(CURLOPT_TCP_KEEPALIVE, 1L);
+  if (rc != CURLE_OK)
+  {
+    return rc;
+  }
+
   if (reuse_connection_)
   {
     rc = SetCurlLongOption(CURLOPT_FRESH_CONNECT, 0L);
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -425,10 +425,7 @@ HttpOperation::HttpOperation(opentelemetry::ext::http::client::Method method,
                              std::chrono::milliseconds http_conn_timeout,
                              bool reuse_connection,
                              bool is_log_enabled,
-                             const RetryPolicy &retry_policy,
-                             bool keepalive_allow,
-                             std::chrono::seconds keepalive_idle,
-                             std::chrono::seconds keepalive_intvl)
+                             const RetryPolicy &retry_policy)
     :  // Optional connection params
       is_raw_response_(is_raw_response),
       reuse_connection_(reuse_connection),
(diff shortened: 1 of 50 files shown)


marcalff (approved):
LGTM, thanks for the fix.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1162,6 +1162,12 @@ CURLcode HttpOperation::Setup()
     return rc;
   }
 
+  rc = SetCurlLongOption(CURLOPT_TCP_KEEPALIVE, 1L);
```
owent: Would it be possible to expose this as a configuration option for HttpClient, OtlpHttpClient, and all HTTP-based exporters? That way users can opt-in rather than having it unconditionally enabled.
Also, I noticed CURLOPT_TCP_KEEPIDLE and CURLOPT_TCP_KEEPINTVL are being set without CURLOPT_TCP_KEEPALIVE. These two options are normally only effective when keep-alive is enabled, so we should probably set all three together.

Babul422: Hi @owent,

Thanks for the review.
My understanding is that the preferred direction is to make TCP keepalive opt-in (disabled by default) and expose it through HttpClient, OtlpHttpClient, and the HTTP-based exporters.

I was planning to expose a single boolean configuration (for CURLOPT_TCP_KEEPALIVE) while enabling the related libcurl behavior, rather than exposing additional public configuration for CURLOPT_TCP_KEEPIDLE and CURLOPT_TCP_KEEPINTVL unless you think those should also be configurable.

Does that match what you had in mind?

Thanks!

owent: I think that will be helpful to let CURLOPT_TCP_KEEPIDLE and CURLOPT_TCP_KEEPINTVL also be configurable.

lalitb: +1. I think these should be configurable together. If we only enable `CURLOPT_TCP_KEEPALIVE`, the timing still depends on libcurl/OS defaults, so it may not help users who need faster detection for half-open connections like in #3333.

lalitb (changes requested):
As discussed [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4266#discussion_r3610783872) - The keepalive option and timing knobs should be configurable together, otherwise this may not solve #3333 for users who need faster half-open detection.

lalitb:
Thanks for working on this. Enabling TCP keepalive can be useful, but I don’t think we can say yet that it fixes #3333.

TCP keepalive helps when a connection was established and the other side disappears without closing it. It does not help a connection still being created, such as the `SYN_SENT` connection shown in the issue. We also already set a timeout for the complete HTTP request, so we should first understand why a five-second timeout can still lead to a much longer wait.

I’m also unsure about adding this setting to the common Request interface. Keepalive controls the underlying TCP connection, which may be reused for many requests. It is therefore better treated as a curl/client setting than as a property of one HTTP request. Changing the common virtual interface may also affect external custom HTTP clients.

I suggest keeping this PR in draft for now. Let’s first add a small test that reproduces the long wait, identify whether the problem is in curl timeout handling or our session/shutdown logic, and discuss the right place for transport settings as part of #4448.

Issue #2526, opened 2024-02-08
Update exemplar and remove ExemplarFilter interface


ThomsonTan:
This comes from the spec change https://github.com/open-telemetry/opentelemetry-specification/pull/3820

PR #4267, opened 2026-07-18
[SDK] refactor: exemplar filters

### LINKED ISSUE #4178: Is support for exemplars still experimental/preview?
Hello, 

We are interested in using Exemplars to correlate metrics with trace/span IDs. As far as I can see, this is the latest PR merged related to the feature:

https://github.com/open-telemetry/opentelemetry-cpp/pull/1710

This PR is from 2022 and claims that the feature is still in "preview" mode and gated behind a flag. The latest commit I could find related to Exemplars is this one from 2025:

https://github.com/open-telemetry/opentelemetry-cpp/commit/7b3dbd0ed525f74cde6ea158eed56f4ea69580cd

Even there, the flag is still explicitly used.

Are Exemplars mature enough to be used in production software? If not, what are the current drawbacks of using them?

Thanks.

proost:
Related to https://github.com/open-telemetry/opentelemetry-cpp/issues/4178,  https://github.com/open-telemetry/opentelemetry-cpp/issues/2526

## Changes

Breaking Change is accompanied, But Because this is preview version, I think acceptable.

I will send another PR for configuration from envs.


lalitb:
@proost  - I think this won't yet close #4178  - We can't declare exemplar stable till we have evaluated their performance on hot-path of metrics.

sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir_utils.h
```diff
@@ -62,18 +65,21 @@ static inline nostd::shared_ptr<ExemplarReservoir> GetExemplarReservoir(
...
+                  histogram_agg_config->boundaries_),
+              GetMapAndResetCellMethod(instrument_descriptor))));
     }
   }
 
-  return nostd::shared_ptr<ExemplarReservoir>(new SimpleFixedSizeExemplarReservoir(
-      GetSimpleFixedReservoirDefaultSize(agg_type, agg_config),
-      SimpleFixedSizeExemplarReservoir::GetSimpleFixedSizeCellSelector(),
-      GetMapAndResetCellMethod(instrument_descriptor)));
+  return ExemplarReservoir::GetSimpleFilteredExemplarReservoir(
```
lalitb: For `AlwaysOff`, can we return `GetNoExemplarReservoir()` directly? Otherwise every metric recording still goes through the exemplar filter path just to drop the sample, so `AlwaysOff` is not really the same as disabling exemplar sampling

sdk/include/opentelemetry/sdk/metrics/exemplar/filtered_exemplar_reservoir.h
```diff
@@ -0,0 +1,111 @@
...
+  {}
+
+  void OfferMeasurement(int64_t value,
+                        const MetricAttributes &attributes,
+                        const opentelemetry::context::Context &context,
+                        const opentelemetry::common::SystemTimestamp &timestamp) noexcept override
+  {
+    if (should_sample_(context))
+    {
+      reservoir_->OfferMeasurement(value, attributes, context, timestamp);
```
lalitb: I think `AlwaysOn` still does not fully work here. This forwards the measurement, but the real reservoir drops it later if there is no valid span context. Since trace/span ids are optional for exemplars, `AlwaysOn + Context{}` should still produce an exemplar without trace ids. Can we add a test that collects from the real reservoir and checks the returned exemplar is non-null?

sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir_utils.h
proost: For the performance, you are right. But according to [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#exemplarfilter), looks like current implementation is correct.

Here is what i understand:
```
ExemplarFilter decide whether measurement is eligible for the examplar. ExamplarReservoir decide wheter examplar to be stored or not.
```

If i missunderstand the spec, please let me know where I got it wrong. (spec is bit confused to me.)

sdk/include/opentelemetry/sdk/metrics/exemplar/filtered_exemplar_reservoir.h
proost: Same point for the above https://github.com/open-telemetry/opentelemetry-cpp/pull/4267#discussion_r3631366150

sdk/test/metrics/exemplar/filtered_exemplar_reservoir_test.cc
```diff
@@ -0,0 +1,125 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
```
dbarker: The bazel valgrind error is due to a test timeout. It seems the additional tests may take longer. The timeout may need to be increased. 

```
[----------] 1 test from CounterToSumFilterAttributes (30 ms total)

[----------] 1 test from CounterToSumFilterAttributesWithCardinalityLimit
[ RUN      ] CounterToSumFilterAttributesWithCardinalityLimit.Double
-- Test timed out at 2026-07-22 15:36:51 UTC --
```

sdk/include/opentelemetry/sdk/metrics/state/async_metric_storage.h
```diff
@@ -66,11 +63,8 @@ class AsyncMetricStorage : public MetricStorage, public AsyncWritableMetricStora
     for (auto &measurement : measurements)
     {
 #ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
-      if (exemplar_filter_type_ == ExemplarFilterType::kAlwaysOn)
-      {
-        exemplar_reservoir_->OfferMeasurement(measurement.second, {}, {},
-                                              std::chrono::system_clock::now());
-      }
+      exemplar_reservoir_->OfferMeasurement(measurement.second, {}, {},
```
dbarker: As Lalit mentioned we need to have a near zero cost path when the filter is AlwaysOff. The system call to the clock can be expensive and should be avoided in that case. We need a shortcut here. The spec supports this :) 

> If Exemplar sampling is off, the SDK MUST NOT have overhead related to exemplar sampling.

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -97,10 +81,7 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
       return;
     }
 #ifdef ENABLE_METRICS_EXEMPLAR_PREVIEW
-    if (EnableExamplarFilter(exemplar_filter_type_, context))
-    {
-      exemplar_reservoir_->OfferMeasurement(value, {}, context, std::chrono::system_clock::now());
-    }
+    exemplar_reservoir_->OfferMeasurement(value, {}, context, std::chrono::system_clock::now());
```
dbarker: Same as above. A shortcut in the AlwaysOff case is needed. The system call to the clock should be made only if an exemplar is being sampled. If trace based and the span is not sampled there should be no call to the clock.

sdk/include/opentelemetry/sdk/metrics/exemplar/filtered_exemplar_reservoir.h
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  void OfferMeasurement(double value,
+                        const MetricAttributes &attributes,
+                        const opentelemetry::context::Context &context,
+                        const opentelemetry::common::SystemTimestamp &timestamp) noexcept override
```
dbarker: Please double check if this timestamp is ever used and remove this argument if not. 

[reservoir_cell.h#L136](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir_cell.h#L136) gets a new time stamp.

dbarker (commented):
Thanks for the PR! Please see some initial feedback below.

sdk/include/opentelemetry/sdk/metrics/exemplar/filtered_exemplar_reservoir.h
dbarker: Lalit is correct. In the AlwaysOn case the exemplar should not be dropped in the case of no active span.  A test for this would be helpful.

sdk/include/opentelemetry/sdk/metrics/exemplar/filtered_exemplar_reservoir.h
proost: Nice catch! You are right. ts doesn't needed. [ca5c78e](https://github.com/open-telemetry/opentelemetry-cpp/pull/4267/commits/ca5c78ef57143f28126cbaf3d48f18328c7fe9fa)

proost:
@dbarker @lalitb 
I'm sorry to bothering you. I missunderstand [the issue](https://github.com/open-telemetry/opentelemetry-cpp/issues/2526) and spec.

So this PR change:
1. removing span context check so that alwaysOn works correctly.
2. little bit refactoring(removing useless "GetSimpleFilteredExemplarReservoir")
3. tiny performance improving that removing timestamp.

CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ Increment the:
   namespaces
   [#4302](https://github.com/open-telemetry/opentelemetry-cpp/pull/4302)
 
+* [SDK] Complete exemplar filtering: the exemplar filter(`AlwaysOn`/
```
ThomsonTan, line 188: Should also be declared in the breaking change section below?

proost: [5eddcd2](https://github.com/open-telemetry/opentelemetry-cpp/pull/4267/commits/5eddcd2aa851fcde40ba2630b9ecacaf887d53e9)

sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir_cell.h
```diff
@@ -134,11 +134,8 @@ class ReservoirCell
   {
     attributes_  = attributes;
     record_time_ = opentelemetry::common::SystemTimestamp(std::chrono::system_clock::now());
-    const auto current_ctx = opentelemetry::trace::GetSpanContext(context);
-    if (current_ctx.IsValid())
-    {
-      context_.reset(new opentelemetry::trace::SpanContext{current_ctx});
-    }
+    context_.reset(
```
dbarker: For performance and overhead, please consider if this copy construction and dynamic allocation of a shared SpanData object can be avoided. Can be a follow up PR.

dbarker (approved):
Thanks for the fix. I think this is reasonable. Please see optional feedback (can be follow up PRs) and consider the copilot feedback. We do still need benchmarking for the exemplar path to resolve the linked ticket.

sdk/include/opentelemetry/sdk/metrics/exemplar/reservoir_cell.h
proost: [7bf2585](https://github.com/open-telemetry/opentelemetry-cpp/pull/4267/commits/7bf25852a479d87490f9b76c85d78c1503ddc1f5)

Including this change, also reasonable to me. How about this?

lalitb:
@proost Could you please change `Fixes #4178` to `Related to #4178` in the PR description? This PR improves the exemplar implementation, but we still need hot-path performance evaluation before considering exemplars stable and closing that issue.

lalitb (approved):
Thanks.

PR #4268, opened 2026-07-19
[BUILD] Rename cmake options with prefix OTELCPP_ 

### LINKED ISSUE #4184: [CMAKE] Add an `OTELCPP_` prefix to CMake options
**Problem**

`opentelemetry-cpp`'s CMake options use generic names (`WITH_EXAMPLES`, `WITH_BENCHMARK`, `BUILD_TESTING`, etc.) with no project prefix. CMake's options create cached variables that are global per project.  When the library is consumed via `FetchContent` or `add_subdirectory`, these options can collide with identically-named options in the parent project.

[`BUILD_TESTING`](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html) is a special case because it is a CMake-standard variable set by `include(CTest)`. A parent project setting `BUILD_TESTING OFF` to disable `opentelemetry-cpp`'s tests will also disable their own. 

**Proposed fix**

Introduce `OTELCPP_` as the prefix for all project options (e.g. `OTELCPP_WITH_EXAMPLES`, `OTELCPP_BUILD_TESTING`). Retain the old option names as deprecated aliases that forward to the new names for a release cycle or two.

Users with existing command-line or preset-based configurations are unaffected (alias forwards the value, prints a deprecation warning). Users with a stale `CMakeCache.txt` must clear the cache and reconfigure their build.

owent:
Fixes #4184

## Changes

- Rename cmake options with prefix OTELCPP_ 


CMakeLists.txt
```diff
@@ -155,67 +170,98 @@ if(NOT WITH_STL STREQUAL "OFF")
...
+# default and do not treat it as a deprecated alias.
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+  if(DEFINED BUILD_TESTING)
+    option(OTELCPP_BUILD_TESTING "Whether to enable tests" "${BUILD_TESTING}")
+  else()
+    option(OTELCPP_BUILD_TESTING "Whether to enable tests" ON)
+  endif()
+  otelcpp_option_flag(
+    OTELCPP_WITH_BENCHMARK "Whether to build benchmark program"
+    ${OTELCPP_BUILD_TESTING})
```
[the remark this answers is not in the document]
owent: OTELCPP_WITH_BENCHMARK should follow OTELCPP_BUILD_TESTING, or benchmark may not be imported.

tools/build.cmd
```diff
@@ -92,15 +92,15 @@ REM - stl              - build with Standard Template Library
 REM ##########################################################################################
 REM Build with nostd implementation.
 REM ##########################################################################################
-set CONFIG=-DWITH_STL:BOOL=OFF %*
+set CONFIG=-DOTELCPP_WITH_STL:BOOL=OFF %*
```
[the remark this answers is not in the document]
owent: Just keep the old behaviour

tools/build.cmd
```diff
@@ -92,15 +92,15 @@ REM - stl              - build with Standard Template Library
...
 set "OUTDIR=%ROOT%\out\%BUILDTOOLS_VERSION%\nostd"
 call :build_config
 
 REM ##########################################################################################
 REM Build with STL implementation. This option does not yield benefits for vs2015 build.
 REM ##########################################################################################
 if "%BUILDTOOLS_VERSION%" neq "vs2015" (
-  set CONFIG=-DWITH_STL:BOOL=ON %*
+  set CONFIG=-DOTELCPP_WITH_STL:BOOL=ON %*
```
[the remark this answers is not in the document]
owent: Just keep the old behaviour

test_common/cmake/all-options-abiv1-preview.cmake
```diff
@@ -3,8 +3,8 @@
 
 # Enable all options with ABI V1 including preview features
 
-set(WITH_ABI_VERSION_1 ON CACHE BOOL "" FORCE)
```
dbarker, line 6: Can we add a test .cmake file with all the old options and map to their new names? This would let us verify the conversion works in a CI test that configures the project with old options and checks that all the new options are set in the cache.  These all-options files should be the source of truth for what options are supported.

tools/build.cmd
dbarker: is this file still used or can be removed?

CMakeSettings.json
dbarker, line 11: is this file still used or can be removed?

cmake/tools.cmake
```diff
@@ -319,3 +319,86 @@ function(project_build_tools_set_static_library_declaration DEFINITION_VARNAME)
     endforeach()
   endif()
 endfunction()
+
+function(_otelcpp_get_legacy_option_name __OPTION_NAME __OUTPUT_VARIABLE)
+  if(__OPTION_NAME MATCHES "^OTELCPP_WITH_")
```
dbarker: The `_otelcpp_get_legacy_option_name` function currently encodes the old to new mapping as a chain of `elseif` branches. It makes a bit more challenging to review for completeness. Since the list of options isn't too large, consider replacing the elseif statements here with a single list (in its own `cmake/otelcpp_option_map.cmake`):

```cmake
# Each entry is a pair: OLD_NAME;NEW_NAME
set(_OTELCPP_LEGACY_OPTION_MAP
  "WITH_OTLP_GRPC;OTELCPP_WITH_OTLP_GRPC"
  "BUILD_W3CTRACECONTEXT_TEST;OTELCPP_BUILD_W3CTRACECONTEXT_TEST"
  "OPENTELEMETRY_INSTALL;OTELCPP_INSTALL"
  # ...
)
```

`_otelcpp_get_legacy_option_name` would then loop over that list to find the matching new name, and CMake tests could include the same file and iterate the pairs directly. That way we have a single source of truth for the option name changes. What do you think?

dbarker (commented):
Thanks! Please see initial feedback below. We'll need to add some CI testing for the conversion since it can be high impact.

test_common/cmake/all-options-abiv1-preview.cmake
owent: Yes, sorry for missing it. I'll add the test later.

tools/build.cmd
owent: I'm not sure. It seems not relate to this PR. Can we clean up these files in another PR?

CMakeSettings.json
owent: I'm not sure. Should we rename then to vs2022 ?

cmake/tools.cmake
owent: Most of options are start with `OTELCPP_WITH_`, do you think we can just apply this map for special names that do not start with `OTELCPP_WITH_`? It will be more simple to read.

CMakeSettings.json
dbarker: We can remove this file in a separate PR and add it to .gitignore since it is user/ide specific.

Looks like the better replacement is a `CMakePresets.json` to standardize builds for development and CI. Also a good follow up PR! :)  
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

cmake/tools.cmake
dbarker: I think having one source of truth with an explicit map of all options will be easier to review and test. It is likely less error prone too because the rule is a simple old_option to new_option map. It will ensure that only valid old options are converted to new cached options.

CHANGELOG.md
```diff
@@ -43,6 +46,29 @@ Increment the:
...
+      `OTELCPP_BUILD_W3CTRACECONTEXT_TEST`
+    * `BUILD_PACKAGE` becomes `OTELCPP_BUILD_PACKAGE`
+    * `TARBALL` becomes `OTELCPP_TARBALL`
+    * `OPENTELEMETRY_BUILD_DLL` becomes `OTELCPP_BUILD_DLL`
+    * `OPENTELEMETRY_INSTALL` becomes `OTELCPP_INSTALL`
+    * `OPENTELEMETRY_SKIP_DYNAMIC_LOADING_TESTS` becomes
+      `OTELCPP_SKIP_DYNAMIC_LOADING_TESTS`
+    * `OPENTELEMETRY_EXTERNAL_COMPONENT_PATH` becomes
+      `OTELCPP_EXTERNAL_COMPONENT_PATH`
+  * The legacy option names are still accepted but deprecated: a deprecation
```
ThomsonTan: This statement overstates compatibility for `BUILD_TESTING`. That variable is explicitly excluded from deprecated-alias handling. For example, when a parent has `BUILD_TESTING=ON` and adds OpenTelemetry as a subproject, `OTELCPP_BUILD_TESTING` still defaults to `OFF` and no deprecation warning is emitted.

Could we document this exception explicitly? `BUILD_TESTING` only seeds `OTELCPP_BUILD_TESTING` for top-level builds; subprojects must set `OTELCPP_BUILD_TESTING` directly. The behavior is intentional, but the migration contract should match it.

CMakeSettings.json
owent: cmake-presets is available from 3.19. do you think we can upgrade the cmake_minimum_required to 3.19? If so , we can use json map to maintain this option mapping instead of using list or internal variable names.

dbarker: Thanks for adding the explicit option mapping. I think we should stay with 3.16 for now. The main dependencies (googletest, protobuf) are also still on 3.16 with curl setting 3.18 as the minimum.

We can use these presets though for development and testing use cases where we can install the version of CMake needed.

cmake/tools.cmake
dbarker: Thanks for adding the explicit mapping.

ci/do_ci.sh
```diff
@@ -543,12 +543,200 @@ elif [[ "$1" == "cmake.api_only.test" ]]; then
...
+  for expected_entry in \
+      "OTELCPP_WITH_API_ONLY:BOOL=ON" \
+      "OTELCPP_WITH_EXAMPLES_HTTP:BOOL=ON" \
+      "OTELCPP_INSTALL:BOOL=OFF"; do
+    grep -q "^${expected_entry}$" CMakeCache.txt || {
+      echo "ERROR: expected CMakeCache.txt entry missing: ${expected_entry}"
+      exit 1
+    }
+  done
+  exit 0
```
dbarker, line 760: Consider adding a reconfigure pass here by re-running cmake in the same build directory with a different legacy option value than the initial configure. Maybe something like:

```bash
# Pass 4: reconfigure with a changed legacy value in an existing build dir.
cmake "${CMAKE_OPTIONS[@]}" -DWITH_EXAMPLES=ON "${SRC_DIR}"   # initial
cmake "${CMAKE_OPTIONS[@]}" -DWITH_EXAMPLES=OFF "${SRC_DIR}"  # reconfigure
```

The expected outcome should ideally match the standard option behavior (the new value is applied).

CMakeLists.txt
```diff
@@ -155,67 +170,98 @@ if(NOT WITH_STL STREQUAL "OFF")
...
 endif()
 
 if(WIN32)
-  option(WITH_ETW "Whether to include the ETW Exporter in the SDK" ON)
+  otelcpp_option_flag(OTELCPP_WITH_ETW
+                      "Whether to include the ETW Exporter in the SDK" ON)
 else()
-  if(DEFINED (WITH_ETW))
-    message(FATAL_ERROR "WITH_ETW is only supported on Windows")
+  if(DEFINED OTELCPP_WITH_ETW OR DEFINED WITH_ETW)
```
dbarker: can this be `if(OTELCPP_WITH_ETW OR WITH_ETW)`? It seems the configure stage should only fail if a non windows system configures the flag on.  The before diff seems to show the check had a bug `if(DEFINED (WITH_ETW))` and may not have been working as expected.

cmake/tools.cmake
```diff
@@ -319,3 +319,127 @@ function(project_build_tools_set_static_library_declaration DEFINITION_VARNAME)
...
+set(_otelcpp_legacy_option_name_OTELCPP_BUILD_W3CTRACECONTEXT_TEST
+    "BUILD_W3CTRACECONTEXT_TEST")
+set(_otelcpp_legacy_option_name_OTELCPP_INSTALL "OPENTELEMETRY_INSTALL")
+set(_otelcpp_legacy_option_name_OTELCPP_SKIP_DYNAMIC_LOADING_TESTS
+    "OPENTELEMETRY_SKIP_DYNAMIC_LOADING_TESTS")
+set(_otelcpp_legacy_option_name_OTELCPP_BUILD_DLL "OPENTELEMETRY_BUILD_DLL")
+set(_otelcpp_legacy_option_name_OTELCPP_BUILD_PACKAGE "BUILD_PACKAGE")
+set(_otelcpp_legacy_option_name_OTELCPP_EXTERNAL_COMPONENT_PATH
+    "OPENTELEMETRY_EXTERNAL_COMPONENT_PATH")
+set(_otelcpp_legacy_option_name_OTELCPP_TARBALL "TARBALL")
```
dbarker, line 376: Consider adding all the existing options to the table and then we can check for any current options that are not accounted for. Maybe just an empty string for those that already have the OTELCPP_ prefix.

Something like: 

```cmake
set(_otelcpp_legacy_option_name_OTELCPP_MAINTAINER_MODE "")
set(_otelcpp_legacy_option_name_OTELCPP_VERSIONED_LIBS "")
```

and then adding a fatal message on a missing entry in `_otelcpp_get_legacy_option_name`:

```cmake
  else()
    message(FATAL_ERROR
      "${__OPTION_NAME} has no legacy option mapping. Add a _otelcpp_legacy_option_name_${__OPTION_NAME} entry (use \"\" for new options)")
```

dbarker (commented):
Thanks for the updates. Please see some feedback and questions below.

cmake/tools.cmake
```diff
@@ -319,3 +319,127 @@ function(project_build_tools_set_static_library_declaration DEFINITION_VARNAME)
...
+
+  set(${__OUTPUT_VARIABLE}
+      "${__RESOLVED_DEFAULT}"
+      PARENT_SCOPE)
+endfunction()
+
+function(otelcpp_option_flag __OPTION_NAME __HELP_TEXT __DEFAULT_VALUE)
+  _otelcpp_get_option_default("${__OPTION_NAME}" "${__DEFAULT_VALUE}"
+                              __RESOLVED_DEFAULT)
+  option(${__OPTION_NAME} "${__HELP_TEXT}" "${__RESOLVED_DEFAULT}")
```
lalitb, line 465: Could we consider keeping the old option names available as normal, non-cache variables during the deprecation period? Some external components may still read options such as WITH_OTLP_GRPC or WITH_STL. This would help them migrate without adding the old names back to the global cache.

CHANGELOG.md
owent: Thanks, document is added into CHANGELOG.md now

ci/do_ci.sh
owent: Thanks for the suggestion. I've been thinking about how to implement the reconfigure pass, but I ran into a limitation: when both the legacy and new options are set, we should prefer the new one. However, on reconfigure CMake does not distinguish whether the new option's value was provided by the user on the command line or simply carried over from CMakeCache.txt, so we can't tell whether applying the legacy value would override an explicit user choice. Do you have any suggestion for handling this case?

CMakeLists.txt
owent: My understanding is no. The original check was meant to reject any explicit setting of the flag on non-Windows systems, regardless of its value. Changing it to if(OTELCPP_WITH_ETW OR WITH_ETW) would only catch the ON case — a user could still set either option to OFF, which should also be rejected.

cmake/tools.cmake
owent: I understand the goal of making the table a complete registry, but I'm a bit concerned about the long-term cost: the FATAL_ERROR would require every future option to carry an empty _otelcpp_legacy_option_name_* entry, which is a permanent maintenance burden for what is essentially a temporary migration problem. Would it be acceptable to treat a missing entry as "no legacy name" instead? The completeness check you described could still be enforced once by the CI test with the all-options files, rather than at configure time for every future option.

cmake/tools.cmake
owent: Sorry but I can't get the point. The legacy options are already not cached now. If you find the old names in CMakeCache.txt, could you please try to remove this file and reconfigure otelcpp?

CMakeLists.txt
dbarker: Thanks.  Can we use the new `otelcpp_dependent_option` to fully remove use of the old `WITH_ETW` and the if statement here?

cmake/tools.cmake
dbarker: Thanks. That makes sense with a migration completeness test in ci.

ci/do_ci.sh
dbarker: Thanks for the clarification. How about ignoring the legacy option on the reconfigure step and emit a warning if its value is different from the new option cached value?

CMakeLists.txt
owent: `otelcpp_dependent_option` only define defferent default value by condition. I added `AVAILABLE_CONDITION` and `UNAVAILABLE_MESSAGE` to `otelcpp_option_flag` to remove use of the old WITH_ETW and the if statement.

ci/do_ci.sh
owent: Good point, the warning message is added.

cmake/tools.cmake
lalitb: Thanks. I was not referring to CMakeCache.txt. I meant the other direction: after `OTELCPP_WITH_OTLP_GRPC` is resolved, could we also set `WITH_OTLP_GRPC` to the same value as a normal, non-cache variable?

Currently, the old name exists only when the caller defines it. If the caller uses the new name, or relies on its default, an external component still reading `WITH_OTLP_GRPC` sees it as undefined. Mirroring the resolved value would keep such components working during deprecation without adding the old name to the cache.

owent: The legacy option will now be set to match the new option. That said, I'm wondering about the `add_subdirectory(...)` use case—non-cached variables would remain in the sub-scope and therefore wouldn't be readable by the parent project. Would this be an issue we should address?

ci/do_ci.sh
```diff
@@ -543,12 +543,200 @@ elif [[ "$1" == "cmake.api_only.test" ]]; then
...
         "${SRC_DIR}/install/test/cmake/api_only_test"
   cmake --build . "${CMAKE_BUILD_ARGS[@]}"
   ctest --output-on-failure
   exit 0
+elif [[ "$1" == "cmake.legacy_options.test" ]]; then
+  # Ensure the legacy (pre "OTELCPP_" prefix) CMake option names still drive
+  # the new OTELCPP_* options with unchanged semantics. See
+  # https://github.com/open-telemetry/opentelemetry-cpp/issues/4184
+  #
+  # Pass 1: only the legacy names are given. Every legacy name must trigger a
```
dbarker, line 578: Should `WITH_BENCHMARK` and `OPENTELEMETRY_INSTALL`  both be `OFF` in the first pass to avoid using the defaults?

CMakeLists.txt
```diff
@@ -57,86 +59,99 @@ if(DEFINED ENV{VCPKG_ROOT} AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
...
+                        "Which version of the Standard Library for C++ to use")
 
-option(WITH_GSL
-       "Whether to use Guidelines Support Library for C++ latest features" OFF)
+otelcpp_option_flag(
+  OTELCPP_WITH_GSL
+  "Whether to use Guidelines Support Library for C++ latest features" OFF)
 
-set(OPENTELEMETRY_INSTALL_default ON)
+set(OTELCPP_INSTALL_DEFAULT ON)
```
dbarker: consider  changing `OTELCPP_INSTALL_DEFAULT` -> `_OTELCPP_INSTALL_DEFAULT` to signal it is a local internal project variable instead of an option.

CHANGELOG.md
```diff
@@ -65,6 +68,37 @@ Increment the:
...
+    * `BUILD_TESTING` becomes `OTELCPP_BUILD_TESTING`. Unlike the other renamed
+      options, `BUILD_TESTING` is not treated as a deprecated alias:
+      * When opentelemetry-cpp is the top-level project, `BUILD_TESTING` sets
+        the default value of `OTELCPP_BUILD_TESTING`.
+      * When opentelemetry-cpp is a subproject, the parent project's
+        `BUILD_TESTING` no longer controls opentelemetry-cpp tests, which
+        default to `OFF`.
+      * Set `OTELCPP_BUILD_TESTING` explicitly to enable or disable
+        opentelemetry-cpp tests in either top-level or subproject builds.
+    * `BUILD_W3CTRACECONTEXT_TEST` becomes
```
dbarker: Some of these are not widely used options by users. Since this is an example list consider using the more common options (WITH_OTLP_*, WITH_CONFIGURATION, WITH_*_PREVIEW) or dropping the example list since the first bullet point also covers the name change rule.

CHANGELOG.md
```diff
@@ -65,6 +68,37 @@ Increment the:
 
 Breaking changes:
 
+* [CMAKE] Rename cmake options with prefix `OTELCPP_`
+  [#4268](https://github.com/open-telemetry/opentelemetry-cpp/pull/4268)
+  * All CMake build options were renamed with the `OTELCPP_` prefix to avoid
+    name collisions with user and third-party project variables:
+    * `WITH_*` options are renamed to `OTELCPP_WITH_*`, for example
+      `WITH_OTLP_HTTP` becomes `OTELCPP_WITH_OTLP_HTTP`
+    * `BUILD_TESTING` becomes `OTELCPP_BUILD_TESTING`. Unlike the other renamed
```
dbarker: the default for WITH_BENCHMARK also changed from ON to `${OTELCPP_BUILD_TESTING}`. This should be documented also.

api/CMakeLists.txt
```diff
@@ -11,111 +11,112 @@ set_target_properties(opentelemetry_api PROPERTIES EXPORT_NAME api)
...
+  message(WARNING "OTELCPP_WITH_OTLP_RETRY_PREVIEW=OFF is deprecated")
 endif()
 
-if(WITH_OTLP_GRPC_SSL_MTLS_PREVIEW)
+if(OTELCPP_WITH_OTLP_GRPC_SSL_MTLS_PREVIEW)
   target_compile_definitions(opentelemetry_api
                              INTERFACE ENABLE_OTLP_GRPC_SSL_MTLS_PREVIEW)
 else()
-  message(WARNING "WITH_OTLP_GRPC_SSL_MTLS_PREVIEW=OFF is deprecated")
+  message(WARNING "OTELCPP_WITH_OTLP_GRPC_SSL_MTLS_PREVIEW=OFF is deprecated")
```
dbarker, line 101: Please also add the new name to the DEPRECATED.md for the previously deprecated options. 

Consider adding a separate section in DEPRECATED.md noting general deprecation of the old names following the template.

dbarker (approved):

dbarker:
Thanks for the PR @owent ! Please see some feedback above. The main risk I see is that on the next release this change will trigger CMake deprecation warnings in many projects. This is communicated in the release notes well, but may still cause friction as CI workflows will need to be updated. Leaving the discuss tag on the PR to get feedback from others in the SIG meetings coming up.

ci/do_ci.sh
owent: - **Pass 1** set `OTELCPP_BUILD_TESTING` to `OFF` , which will make the default value of `WITH_BENCHMARK ` to `OFF`.
- `TARBALL=ON` needs `OPENTELEMETRY_INSTALL=ON`, so **Pass 1** set it to `ON` to avoid to mislead users. **Pass 3** also test `OPENTELEMETRY_INSTALL=OFF`.

CMakeLists.txt
owent: Thanks, it's done.

CHANGELOG.md
owent: Thanks, the break change is documented now.

CHANGELOG.md
owent: Thanks. Would you mind reviewing this when you have a moment? I'm happy to make any adjustments if needed.

api/CMakeLists.txt
owent: Thanks! Could you take a quick look and let me know if it looks good?

dbarker: Looks good. Thanks!

ext/src/dll/input.src
```diff
@@ -4,15 +4,20 @@
 // clang-format off
```
lalitb, line 4: Could we move the DLL export additions and the related Windows CI corrections to a separate PR? They are not part of renaming the CMake options and materially change the Windows DLL surface. Keeping them here makes it difficult to tell whether a regression comes from option migration or from the DLL changes.

cmake/tools.cmake
```diff
@@ -319,3 +319,199 @@ function(project_build_tools_set_static_library_declaration DEFINITION_VARNAME)
...
+      # so CMake does not report a recognized deprecated option as unused.
+      set(__LEGACY_VALUE "${${__LEGACY_OPTION_NAME}}")
+      message(
+        DEPRECATION
+          "CMake option ${__LEGACY_OPTION_NAME} is deprecated. Use ${__OPTION_NAME} instead."
+      )
+      # The namespaced option takes precedence when both names are provided.
+      if(NOT DEFINED ${__OPTION_NAME})
+        set(__RESOLVED_DEFAULT "${__LEGACY_VALUE}")
+      elseif(NOT "${${__OPTION_NAME}}" EQUAL "${__LEGACY_VALUE}")
```
lalitb: nit - This should use `STREQUAL`, not `EQUAL`.

CMakeLists.txt
```diff
@@ -300,74 +350,73 @@ include(GNUInstallDirs)
...
-  set(WITH_ETW OFF)
-  set(WITH_CONFIGURATION OFF)
-  set(WITH_HTTP_CLIENT_CURL OFF)
-  set(WITH_OTLP_HTTP_COMPRESSION OFF)
-  set(WITH_BENCHMARK OFF)
-  set(WITH_FUNC_TESTS OFF)
-  set(BUILD_TESTING OFF)
-  set(BUILD_W3CTRACECONTEXT_TEST OFF)
+if(OTELCPP_WITH_API_ONLY)
+  set(OTELCPP_WITH_EXAMPLES OFF)
```
lalitb: Should we also set `WITH_EXAMPLES` to `OFF` here?Earlier, the code copies `OTELCPP_WITH_EXAMPLES` into `WITH_EXAMPLES`. After this line, the new option can be `OFF` while the old option is still ON. Code using the old option would then see the wrong value.

cmake/tools.cmake
lalitb: Thanks, this scope is fine. The parent project does not need to read the old variables. They only need to be available inside `opentelemetry-cpp` and its external-component subdirectories.

I added one separate comment because the old variables can become stale when the new values are changed later by `OTELCPP_WITH_API_ONLY`.

ext/src/dll/input.src
owent: I don't known why but some CI jobs failed before, so I run make_def.ps1 to rebuild them.
Do you think we can just keep them to make all jobs passed?

cmake/tools.cmake
owent: Thanks, it's fixed now.

CMakeLists.txt
owent: Good point. All legacy options are also set now.

ext/src/dll/input.src
dbarker: Hi @owent,  Could you resolve the merge conflicts and try to revert the dll/input.src file change? Once this passes CI it should be ready merge and go into the next release.

owent: Sorry for the delay. Conflicts are resloved. @dbarker

dbarker: Thanks @owent. The last request from @lalitb seems to be reverting the changes to `ext/src/dll/input.src` for the dll. Did this PR impact the DLL build to require these changes?

marcalff (dismissed):
Thanks for the fix.

Please move changes to file `./ext/src/dll/input.src` to a separate PR.

This helps investigations in general, like when using git bisect, as this change seems unrelated.

Looks ok otherwise.

owent:
> Thanks for the fix.
> 
> Please move changes to file `./ext/src/dll/input.src` to a separate PR.
> 
> This helps investigations in general, like when using git bisect, as this change seems unrelated.
> 
> Looks ok otherwise.

The changes in `./ext/src/dll/input.sr` is reverted. And so [CI / CMake -> exporter proto (Build as DLL) (pull_request)](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/32331834618/job/96313833941?pr=4268) failed.

I raise another PR #4461 to add these symbols.

ext/src/dll/input.src
owent: The changes in `ext/src/dll/input.src`  are stills required. As mentioned by @marcalff I revert `ext/src/dll/input.src` in this PR, so [CI / CMake -> exporter proto (Build as DLL) (pull_request)](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/32331834618/job/96313833941?pr=4268) failed. And I raise another PR #4661 to add these symbols.

marcalff:
> > Thanks for the fix.
> > Please move changes to file `./ext/src/dll/input.src` to a separate PR.
> > This helps investigations in general, like when using git bisect, as this change seems unrelated.
> > Looks ok otherwise.
> 
> The changes in `./ext/src/dll/input.sr` is reverted. And so [CI / CMake -> exporter proto (Build as DLL) (pull_request)](https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/32331834618/job/96313833941?pr=4268) failed.
> 
> I raise another PR #4461 to add these symbols.

Thanks @owent.

This is my analysis of why it failed:

Before this fix, the windows DLL CI used:

```
  "cmake.exporter.otprotocol.dll.test" {
    cd "$BUILD_DIR"
    cmake $SRC_DIR `
      -DVCPKG_TARGET_TRIPLET=x64-windows `
      -DOPENTELEMETRY_BUILD_DLL=1 `
      -DWITH_OTPROTCOL=ON `
      "-DCMAKE_TOOLCHAIN_FILE=$VCPKG_DIR/scripts/buildsystems/vcpkg.cmake"
```

It turns out `WITH_OTPROTCOL=ON` does nothing, so the single DLL was built without OTLP GRPC and without OTLP HTTP, and it passed.

With this change, the single DDL is built with:

```
      -DOTELCPP_BUILD_DLL=1 `
      -DOTELCPP_WITH_OTLP_GRPC=ON `
      -DOTELCPP_WITH_OTLP_HTTP=ON `
```

which means OTLP GRPC and OTLP HTTP are now included, and it failed because of missing exports.

The missing exports are now fixed in #4461 which is an issue of its own.
This work, #4268, just exposed an existing issue while cleaning up cmake options.

sdk/src/configuration/CMakeLists.txt
marcalff:
```suggestion
endif(OTELCPP_WITH_CONFIGURATION)
```

marcalff: Fir for:

```
CMake Warning (dev) in sdk/src/configuration/CMakeLists.txt:
  A logical block opening on the line

    D:/a/opentelemetry-cpp/opentelemetry-cpp/sdk/src/configuration/CMakeLists.txt:102 (if)

  closes on the line

    D:/a/opentelemetry-cpp/opentelemetry-cpp/sdk/src/configuration/CMakeLists.txt:131 (endif)

  with mis-matching arguments.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

ext/src/dll/input.src
marcalff: I think I found the root cause for why this change was needed.

See https://github.com/open-telemetry/opentelemetry-cpp/pull/4268#issuecomment-5354418331

This is now resolved.

marcalff (approved):
LGTM.

Thanks @owent

PR #4272, opened 2026-07-21
[CMAKE] Add support for CMake component deprecation and update the policy

dbarker:
Fixes # (issue)

The project provides CMake components in the installed package for use with `find_package(opentelemetry-cpp COMPONENTS ...)`. The names of the components and targets they import are user facing and need a path for deprecation as things change. 

Some use cases for this include:
 -  merging components: 
       The config builder components can be merged with the exporter components they depend on since they bring in no new dependencies. 
  - moving components to opentelemetry-cpp-contrib.  

This PR adds support to the `otel_install_component` CMake function for renaming and merging components while deprecating old component names. Users who call `find_package` on a deprecated component name will see a deprecation warning while the old component name resolves to the new component name.    

This change is a part of #4240 and tested in that PR. 

## Changes

- update CMake install scripts in ./cmake/
    - Add DEPRECATED_NAMES and DEPRECATED fields to the `otel_install_component`  function
 - update the installed CMake template files to support resolving deprecated component names to the new component names. 
 - update the deprecation policy to include CMake components. 
 

lalitb:
Thanks @dbarker. This makes sense for #4240 and lets us merge the builder components without breaking existing users. Just wondering if we should keep this PR focused on on DEPRECATED_NAMES, add a small test here, and leave the unfinished DEPRECATED support for later when we have a real use case.

cmake/find-package-support-functions.cmake
```diff
@@ -4,6 +4,29 @@
...
+# Components without an alias are removed from the list
+#-------------------------------------------------------------------------
+function(resolve_deprecated_components requested_components_inout)
+  set(_result ${${requested_components_inout}})
+  foreach(_COMPONENT IN LISTS ${requested_components_inout})
+    if(${_COMPONENT} IN_LIST OTEL_DEPRECATED_COMPONENTS_LIST)
+      set(_replacement_var "COMPONENT_${_COMPONENT}_REPLACEMENT")
+      if(DEFINED ${_replacement_var})
+        message(DEPRECATION "opentelemetry-cpp: component `${_COMPONENT}` is deprecated, use `${${_replacement_var}}` instead.")
+        list(REMOVE_ITEM _result ${_COMPONENT})
```
lalitb, line 18: Do we also need to set `_FOUND` for the old component name? Someone using the old name may still check its `_FOUND` value, but currently we only set it for the replacement.

dbarker: Good catch. I've added setting the old component's  _FOUND variable in the cache and tested here: 
https://github.com/dbarker/opentelemetry-cpp/blob/e585b0d9ad31835e02761d7c3a3786ac608827cc/install/test/cmake/install_functions_test/post_install_tests/deprecated_name/CMakeLists.txt#L24

dbarker:
> Thanks @dbarker. This makes sense for https://github.com/open-telemetry/opentelemetry-cpp/pull/4240 and lets us merge the builder components without breaking existing users. Just wondering if we should keep this PR focused on on DEPRECATED_NAMES, add a small test here, and leave the unfinished DEPRECATED support for later when we have a real use case.

Thanks @lalitb. I've addressed the feedback. To add the small test I've moved the existing install "usage tests" into an `install_functions_test`  folder and created new ci script targets for `cmake.install_functions.test`. This test injects a mock deprecated component into the build and verifies it is resolved correctly through `find_package`. I've also added setting the deprecated components _FOUND flag. Let me know if you spot anything else.

cmake/otel-install-functions.cmake
```diff
@@ -420,6 +445,26 @@ function(otel_install_components)
...
+  if(OTEL_DEPRECATED_COMPONENTS_LIST)
+     message(STATUS "Install DEPRECATED COMPONENT definitions:")
+  endif()
+
+  set(OTEL_COMPONENTS_REPLACEMENTS_BLOCK "")
+  foreach(_DEPRECATED_COMPONENT IN LISTS OTEL_DEPRECATED_COMPONENTS_LIST)
+    get_property(_REPLACEMENT_COMPONENT DIRECTORY ${PROJECT_SOURCE_DIR}
+      PROPERTY OTEL_COMPONENT_REPLACEMENT_${_DEPRECATED_COMPONENT})
+    message(STATUS "  Deprecated COMPONENT ${_DEPRECATED_COMPONENT}")
+    if(_REPLACEMENT_COMPONENT)
```
ThomsonTan, line 480: The alias resolver is single-pass over the *original* request list: `foreach(_COMPONENT IN LISTS ${requested_components_inout})` iterates the input, and the replacement is only `list(APPEND)`-ed to `_result`, so a replacement that is itself deprecated is never resolved further. That makes this a one-hop lookup rather than a resolve-to-final.

cmake/find-package-support-functions.cmake
```diff
@@ -4,6 +4,46 @@
...
 include("${CMAKE_CURRENT_LIST_DIR}/thirdparty-dependency-definitions.cmake")
 
+#-------------------------------------------------------------------------
+# Function to resolve deprecated component names to their aliases.
+# Components without an alias are removed from the list
+#-------------------------------------------------------------------------
+function(resolve_deprecated_components requested_components_inout)
+  set(_result ${${requested_components_inout}})
+  foreach(_COMPONENT IN LISTS ${requested_components_inout})
+    if(${_COMPONENT} IN_LIST OTEL_DEPRECATED_COMPONENTS_LIST)
```
ThomsonTan: `if()` re-dereferences an unquoted left operand that names a defined variable, so `if(${_COMPONENT} IN_LIST ...)` can take the wrong branch.

install/test/cmake/install_functions_test/post_install_tests/no_components/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ if(NOT opentelemetry-cpp_FOUND)
   )
 endif()
 
+# Test components added:
+set(TEST_COMPONENT_TARGETS
```
ThomsonTan, line 26: Unused variable?

cmake/find-package-support-functions.cmake
```diff
@@ -4,6 +4,46 @@
...
+      set(_replacement_var "COMPONENT_${_COMPONENT}_REPLACEMENT")
+      if(DEFINED ${_replacement_var})
+        message(DEPRECATION "opentelemetry-cpp: component `${_COMPONENT}` is deprecated, use `${${_replacement_var}}` instead.")
+        list(REMOVE_ITEM _result ${_COMPONENT})
+        list(APPEND _result ${${_replacement_var}})
```
owent, line 19: Is it OK to change the order?

cmake/find-package-support-functions.cmake
```diff
@@ -4,6 +4,46 @@
...
+  set(${requested_components_inout} "${_result}" PARENT_SCOPE)
+endfunction()
+
+#-------------------------------------------------------------------------
+# Function to set the <package>_<component>_FOUND variable for requested
+# deprecated component names that have a replacement component.
+#-------------------------------------------------------------------------
+function(set_deprecated_components_found requested_components_in)
+  foreach(_COMPONENT IN LISTS ${requested_components_in})
+    if(${_COMPONENT} IN_LIST OTEL_DEPRECATED_COMPONENTS_LIST)
```
owent: Would it be better to use a map-based approach here to avoid the $O(N^2)$ lookup cost?

cmake/find-package-support-functions.cmake
dbarker: Yes it is okay and tested. Resolving the deprecated name to the replacement name happens before the get_dependent_components function is called and the dependent components are inserted into the list in front of the requested component.

cmake/otel-install-functions.cmake
dbarker: Thanks. The intended behavior is that a declared component (`otel_add_component(my_component)`) must declare all the deprecated component names it replaces under `DEPRECATED_NAMES`.  A deprecated name can only have one replacement name and it must be a valid component (not a deprecated name). 

So there must never be a case where deprecated `component_v1` resolves to deprecated `component_v2` that  resolves to a current `component_v3`.  

I've added checks and tests for these cases.

cmake/find-package-support-functions.cmake
dbarker: Thanks good catch. Fixed all instances to use the cmake variable name directly in these `IN_LIST` conditions.

install/test/cmake/install_functions_test/post_install_tests/no_components/CMakeLists.txt
dbarker: Thanks. Updated the test to check these targets.

cmake/find-package-support-functions.cmake
dbarker: No. The list is small and not worth creating a separate map structure now.

dbarker:
Hi @owent, @ThomsonTan @lalitb Thanks for the feedback. All has been addressed, please review when you have a moment.

cmake/otel-install-functions.cmake
```diff
@@ -317,22 +346,27 @@ endfunction()
...
-#      FILES_DIRECTORY <directory>
-#      FILES_DESTINATION <destination>
-#      FILES_MATCHING <matching>)
+#      [FILES_DIRECTORY <directory>
+#       FILES_DESTINATION <destination>
+#       FILES_MATCHING <matching>])
 #-----------------------------------------------------------------------
 function(otel_add_component)
-  set(optionArgs )
+  set(optionArgs DEPRECATED)
```
lalitb, line 367: Thanks, the main feedback is addressed and the new tests look good.

I noticed two small leftovers from the unfinished `DEPRECATED` support mentioned in my earlier comment: `DEPRECATED` is still accepted in `optionArgs`, and the documentation still shows this unimplemented option.

Could we remove these for now and add them later when the feature is implemented?

lalitb (approved):
LGTM overall. I left one small non-blocking comment about cleaning up the remaining unfinished `DEPRECATED` option and its outdated documentation.

cmake/otel-install-functions.cmake
dbarker: Thanks. I will follow up with a PR to fill in that implementation.

PR #4277, opened 2026-07-22
[CONFIGURATION] Add periodic metric reader builder support to registry

dbarker:
Followup to https://github.com/open-telemetry/opentelemetry-cpp/pull/4243#discussion_r3602430452

The configuration tests instantiated the PeriodicMetricReader and call ForceFlush. This caused tests to timeout sporadically. This PR extends the config registry and SDK builder to support setting a builder for the periodic metric reader. A synchronous mock reader is then used in configuration tests to verify config components. 

## Changes

- Add Set/Get methods for the PeriodicMetricReader to the registry and use them in the sdk builder
- Move common config test components to a common header
- Update config tests to use a simple sync metric reader in the periodic reader's registry slot. 
- Add an SDK builder test for PeriodicMetricReaderConfiguration
- Disable config tests that use the batch processors - they'll be enabled in a followup PR. 

### DIFF
--- a/sdk/test/configuration/config_test_common.h
+++ b/sdk/test/configuration/config_test_common.h
@@ -332,7 +332,7 @@ private:
 };
 
 // ---------------------------------------------------------------------------
-// Synchronous metric reader: collects and exports on ForceFlush/Shutdown in
+// Synchronous metric reader: collects and exports on ForceFlush in
 // the calling thread.
 class SyncMetricReader : public opentelemetry::sdk::metrics::MetricReader
 {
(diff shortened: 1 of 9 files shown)


sdk/test/configuration/config_test_common.h
```diff
@@ -0,0 +1,449 @@
...
+      opentelemetry::sdk::metrics::InstrumentType instrument_type) const noexcept override
+  {
+    return exporter_->GetAggregationTemporality(instrument_type);
+  }
+
+private:
+  bool CollectAndExport() noexcept
+  {
+    bool result = true;
+    Collect([this, &result](opentelemetry::sdk::metrics::ResourceMetrics &metric_data) {
```
lalitb: Should we check the return value from Collect() here? If collection fails, result stays true, so ForceFlush() may incorrectly report success.

lalitb (approved):

sdk/test/configuration/config_test_common.h
```diff
@@ -0,0 +1,449 @@
...
+  }
+
+  bool OnForceFlush(std::chrono::microseconds timeout) noexcept override
+  {
+    return CollectAndExport() && exporter_->ForceFlush(timeout);
+  }
+
+  bool OnShutDown(std::chrono::microseconds timeout) noexcept override
+  {
+    return CollectAndExport() && exporter_->Shutdown(timeout);
```
lalitb: Should we always call `exporter_->Shutdown()` here? With`&&`, a collection failure skips exporter shutdown and may leave its resources open.

sdk/test/configuration/config_test_common.h
dbarker: Thanks. Updated to return the result from collect.

sdk/test/configuration/config_test_common.h
dbarker: Thanks. That is better. Updated to rely on force flush solely for triggering a collection and shutdown only calls shutdown on the exporter. This also allowed more strict checks on the export buffer size in the tests.

lalitb (approved):

marcalff (approved):

Issue #4195, opened 2026-06-30
[CODE HEALTH] clang-tidy `misc-override-with-different-visibility` warnings


dbarker:
This issue covers resolving the `misc-override-with-different-visibility` warnings: 

<details><summary><b>misc-override-with-different-visibility (36 warnings)</b><i> - Click to expand </i></summary>
----

| File | Line | Message |
|---|---|---|
| `opentelemetry-cpp/api/include/opentelemetry/trace/propagation/http_trace_context.h` | 194 | visibility of function 'Fields' is changed from public in class 'TextMapPropagator' to private |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 266 | visibility of function 'GetName' is changed from public in class 'Logger' to private |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 270 | visibility of function 'CreateLogRecord' is changed from public in class 'Logger' to private |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 277 | visibility of function 'EmitLogRecord' is changed from public in class 'Logger' to private |
| `opentelemetry-cpp/api/test/logs/logger_test.cc` | 462 | visibility of function 'GetLogger' is changed from public in class 'LoggerProvider' to private |
| `opentelemetry-cpp/api/test/logs/provider_test.cc` | 33 | visibility of function 'GetLogger' is changed from public in class 'LoggerProvider' to private |
| `opentelemetry-cpp/api/test/trace/provider_test.cc` | 22 | visibility of function 'GetTracer' is changed from public in class 'TracerProvider' to private |
| `opentelemetry-cpp/examples/configuration/custom_pull_metric_exporter.h` | 24 | visibility of function 'OnForceFlush' is changed from private in class 'MetricReader' to public |
| `opentelemetry-cpp/examples/configuration/custom_pull_metric_exporter.h` | 26 | visibility of function 'OnShutDown' is changed from private in class 'MetricReader' to public |
| `opentelemetry-cpp/examples/configuration/custom_pull_metric_exporter.h` | 28 | visibility of function 'OnInitialized' is changed from private in class 'MetricReader' to public |
| `opentelemetry-cpp/ext/include/opentelemetry/ext/http/server/http_server.h` | 246 | visibility of function 'onSocketAcceptable' is changed from private (through private inheritance of class 'SocketTools::Reactor::SocketCallback') to protected |
[...]

PR #4280, opened 2026-07-23
[CODE HEALTH] Fix remaining misc-override-with-different-visibility warnings

thc1006:
## Changes

Completes #4195 by resolving the 5 `misc-override-with-different-visibility` warnings that #4215 deferred, all in the `ext/http` embedded server. Each override had wider visibility than its inheritance-adjusted base:

- `HttpServer` privately inherits `SocketCallback`, so its `onSocketAcceptable`, `onSocketReadable`, `onSocketWritable` and `onSocketClosed` overrides are aligned from `protected` to `private`.
- `Reactor` inherits `Thread` as `protected`, so its `onThread` override is aligned from `public` to `protected`.

### Scope and compatibility

These overrides are dispatched only through their base interfaces (`SocketCallback&` and `Thread::onThread`), so no dynamic-dispatch call path changes. The change does narrow the source-level surface a derived class could reach (for example, calling the base implementation from a subclass) in these installed headers.

I chose this direction because the private inheritance of `SocketCallback` looks deliberate (the callbacks are HttpServer's internal reactor plumbing) and HttpServer is extended through its public request-handler API (e.g. `operator[]`, `setServerName`) rather than by overriding the reactor callbacks; `FileHttpServer` is such a subclass and does not touch them. If maintainers would rather preserve that extension surface, two alternatives are: switch `HttpServer` to `protected` inheritance of `SocketCallback` (keeps the callbacks protected but widens the base-conversion surface), or `NOLINT` the four overrides. I am happy to take whichever direction you prefer.

The unrelated `sendMore()` helper keeps its `protected` visibility (an earlier revision narrowed it along with the section; that has been reverted).

### Verification

- A syntax-only build (clang-18, C++14) of `file_http_server.h`, which pulls in `http_server.h` and `socket_tools.h`, the `FileHttpServer` subclass, and an `HttpServer` instantiation, compiles cleanly.
- clang-format and markdownlint 0.46.0 clean.
- `warning_limit` lowered from 163/173 to 158/168. I will confirm from the clang-tidy report artifacts that `misc-override-with-different-visibility` reaches 0 and the totals match before treating this as ready.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 163
+            warning_limit: 158
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 173
+            warning_limit: 168
     env:
       CC: /usr/bin/clang-22
       CXX: /usr/bin/clang++-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [CODE HEALTH] Fix remaining misc-override-with-different-visibility warnings
+  [#4280](https://github.com/open-telemetry/opentelemetry-cpp/pull/4280)
+
 * [CODE HEALTH] Fix more clang tidy warnings (member initialization)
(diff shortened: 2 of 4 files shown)


dbarker (approved):
Looks good to me. Thanks.

marcalff (approved):

Issue #4287, opened 2026-07-24
[TRACKING] ext/http embedded server correctness and safety audit follow-ups


thc1006:
While fixing #4281 and #4282 I read through the rest of `ext/http` and found more than fits in one PR. This is a tracker rather than a single bug: each item below has its own root cause, its own platform surface and its own fix, and I am opening a child issue and a small PR per item as I get to them rather than filing everything at once.

Every item is a **code-inspection finding** against `main` at [`9d38caf`](https://github.com/open-telemetry/opentelemetry-cpp/commit/9d38caf20eacb2618bcd5d9536aba0efcc2990aa). The affected code path and the conditions that reach it are described per item. Runtime reproduction and regression-test status belong in the child issues, not here. Please push back on anything you consider a deliberate limitation of a test server rather than a defect, and I will drop it.

One further potentially security-sensitive finding was reported privately under the project security policy and is intentionally out of scope here.

### Memory safety

- [x] #4288: `processRequest()` uses a `Connection` after `handleConnectionClosed()` erases it. Fixed by #4289.
- [x] #4290: out-of-bounds write in `SocketAddr(char const *)` on Windows, where the copy loop uses a byte count as an element count. Fixed by #4292.
- [x] #4291: `SocketAddr(char const *)` reads uninitialized bytes because the host buffer is terminated at a fixed index. Fixed by #4292.

### Process safety

- [ ] #4293: writes can raise `SIGPIPE` and terminate the host process. PR #4294 is up. Worth landing before the fatal send error item below, since otherwise the process can die before that handling runs.

### Nonblocking write handling

[...]

PR #4289, opened 2026-07-24
[TEST] Fix use-after-free when an HTTP handler closes the connection

### LINKED ISSUE #4288: [BUG] HttpServer accesses a Connection after handleConnectionClosed() erases it
Part of #4287.

`HttpServer::processRequest()` closes the connection when a handler returns `-1` and then keeps using the object it just destroyed.

`m_connections` is a `std::map<SocketTools::Socket, Connection>` (`http_server.h:160`), and `handleConnectionClosed()` ends with `m_connections.erase(connIt)` (`:370`), which destroys the mapped `Connection`. Any `Connection &` bound to it dangles from that point.

`processRequest()` calls it without returning:

```cpp
      if (conn.response.code == -1)
      {
        LOG_TRACE("HttpServer: [%s] closing by request", conn.request.client.c_str());
        handleConnectionClosed(conn);          // conn is destroyed here
      }
    }

    if (conn.response.message.empty())         // :773, read of a destroyed object
    {
      conn.response.message = getDefaultResponseMessage(conn.response.code);
    }

    conn.response.headers["Host"]           = m_serverHost;    // :778-781, writes
    conn.response.headers["Connection"]     = (conn.keepalive ? "keep-alive" : "close");
    conn.response.headers["Date"]           = formatTimestamp(time(nullptr));
    conn.response.headers["Content-Length"] = std::to_string(conn.response.body.size());
```

Control then returns to `handleConnection()`, which reads `conn.request.protocol` and `conn.response.*` to build the status line at `:527-537`, assigns `conn.sendBuffer`, and calls `sendMore(conn)` on a socket that has already been closed.

The other two callers are fine: `onSocketReadable()` returns immediately after calling it (`:288-289`), and it is the last statement of `onSocketClosed()`.

### Reachability

`-1` is a documented part of the handler protocol rather than an internal value. From `file_http_server.h:129-131`:

```
// Two additional 'special' return codes possible here:
// 0    - proceed to next handler
// -1   - immediately terminate and close connection
```

No handler in the repository returns `-1` today, but `HttpRequestCallback::onHttpRequest` is public installed API, so any user of the embedded server can reach this.

### Suggested fix

Give `processRequest()` a `bool` return meaning "the connection is still alive" and have `handleConnection()` return immediately when it is false. The same signal is needed to close connections on fatal `send()` errors, which is another item in #4287, so it is worth introducing here.

While in that function, `handleConnectionClosed()` passes the result of `m_connections.find()` straight to `erase()` without comparing it against `end()`. Every current call site passes a connection that is in the map, but the guard is cheap.

thc1006:
Fixes #4288
Refs #4287

## The bug

`m_connections` is a `std::map<SocketTools::Socket, Connection>`, so the `m_connections.erase(connIt)` at the end of `handleConnectionClosed()` destroys the mapped `Connection`. `processRequest()` called it and did not return, so everything after that point wrote to a destroyed object: `conn.response.message`, then the `Host`, `Connection`, `Date` and `Content-Length` headers. Control then went back to `handleConnection()`, which read `conn.request.protocol` and `conn.response.*` to build the status line and called `sendMore(conn)` on an already closed socket.

`-1` is a documented part of the handler contract rather than an internal value; `file_http_server.h` states that a handler returning `-1` terminates the connection.

## Fail before, pass after

A standalone harness drives `HttpServer` with a handler that returns `-1`, then connects a raw client and sends one request. On `origin/main`:

```
==11==ERROR: AddressSanitizer: heap-use-after-free
READ of size 4 at 0x5150000001d8 thread T1
    #0 HttpServer::processRequest(Connection&)   http_server.h:775
    #1 HttpServer::handleConnection(Connection&) http_server.h:525
    #2 HttpServer::onSocketReadable(Socket)      http_server.h:293
freed by thread T1 here:
    #8 HttpServer::handleConnectionClosed(Connection&) http_server.h:370
    #9 HttpServer::processRequest(Connection&)         http_server.h:769
```

Freed by the `erase` at `:370`, read at `:775`. With this change the same harness runs clean under both ASan and UBSan, the handler is invoked once, and the client observes the connection closed with no response.

## The fix

`processRequest()` no longer destroys the connection it was handed. It reports an outcome and `handleConnection()` performs the close:

```cpp
if (processRequest(conn) == RequestOutcome::CloseConnection)
{
  conn.state = Connection::Closing;
  handleConnectionClosed(conn);
  return;
}
```

`handleConnectionClosed()` keeps the `end()` guard and now also asserts, so a broken invariant is caught in a debug build rather than silently swallowed. The guard does not make the function idempotent, since the reactor removal and socket close above it have already run, and there is a comment saying so.

The other two callers were already correct: `onSocketReadable()` returns immediately afterwards, and it is the last statement in `onSocketClosed()`.

## Test

`/close/` is now registered in `SetUp()` and returns `-1`. `HandlerRequestedCloseKeepsServerUsable` asserts that the request fails, that a handler invocation counter is exactly 1, and that the next request is still served. The counter is the part that matters: without it, an unregistered or misspelled route passes the first assertion for the wrong reason, which is exactly what happened.

`bazel test --config=asan //...` covers `//ext/test/http:curl_http_test`.

## Verification

- clang-format 18 clean on both files.
- clang-tidy 18 with the repo `.clang-tidy` on `http_server.h`: 3 warnings before and 3 after. The first version of the enum added a `performance-enum-size` warning, so it now has an explicit `std::uint8_t` base type. `warning_limit` is untouched.
- Syntax-only build of `file_http_server.h`, which pulls in the subclass and an `HttpServer` instantiation, compiles clean.

## Scope

Only the handler-requested close. The rest of #4287, including fatal `send()` errors and the Windows partial-write stall, is left for its own PRs.


owent (approved):
LGTM, thanks.

marcalff (approved):
LGTM, thanks for the analysis and fix.

marcalff:
@thc1006 

Please check CI, there is a test failure in bazel (MacOS)

```
2026-07-28T14:56:26.7866140Z FAIL: //ext/test/http:curl_http_test (Aborted) (see /Users/runner/.cache/bazel/cb80cd7005204d9de6ac6bba60573b2f/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/ext/test/http/curl_http_test/test.log)
2026-07-28T14:56:26.7989740Z [7,694 / 9,846] 34 / 166 tests, 1 failed; Testing //ext/test/http:curl_http_test; 6s darwin-sandbox ... (3 actions running)
2026-07-28T14:56:26.8091240Z INFO: From Testing //ext/test/http:curl_http_test:
2026-07-28T14:56:26.8093480Z ==================== Test output for //ext/test/http:curl_http_test:
2026-07-28T14:56:26.8180790Z Running main() from gmock_main.cc
2026-07-28T14:56:26.8230540Z [==========] Running 20 tests from 1 test suite.
2026-07-28T14:56:26.8233080Z [----------] Global test environment set-up.
2026-07-28T14:56:26.8248250Z [----------] 20 tests from BasicCurlHttpTests
2026-07-28T14:56:26.8249670Z [ RUN      ] BasicCurlHttpTests.DoNothing
2026-07-28T14:56:26.8250500Z [       OK ] BasicCurlHttpTests.DoNothing (0 ms)
2026-07-28T14:56:26.8251410Z [ RUN      ] BasicCurlHttpTests.HttpRequest
2026-07-28T14:56:26.8252140Z [       OK ] BasicCurlHttpTests.HttpRequest (504 ms)
2026-07-28T14:56:26.8252940Z [ RUN      ] BasicCurlHttpTests.HttpResponse
2026-07-28T14:56:26.8253710Z [       OK ] BasicCurlHttpTests.HttpResponse (0 ms)
2026-07-28T14:56:26.8254440Z [ RUN      ] BasicCurlHttpTests.SendGetRequest
2026-07-28T14:56:26.8255240Z [       OK ] BasicCurlHttpTests.SendGetRequest (27 ms)
2026-07-28T14:56:26.8255940Z [ RUN      ] BasicCurlHttpTests.SendPostRequest
2026-07-28T14:56:26.8256730Z [       OK ] BasicCurlHttpTests.SendPostRequest (3 ms)
2026-07-28T14:56:26.8257480Z [ RUN      ] BasicCurlHttpTests.RequestTimeout
2026-07-28T14:56:26.8258190Z [       OK ] BasicCurlHttpTests.RequestTimeout (5023 ms)
2026-07-28T14:56:26.8259000Z [ RUN      ] BasicCurlHttpTests.CurlHttpOperations
2026-07-28T14:56:26.8259790Z [Error] File: ext/src/http/client/curl/http_operation_curl.cc:1289 Unexpected HTTP method
2026-07-28T14:56:26.8260750Z [       OK ] BasicCurlHttpTests.CurlHttpOperations (502 ms)
2026-07-28T14:56:26.8261530Z [ RUN      ] BasicCurlHttpTests.RetryPolicyEnabled
2026-07-28T14:56:26.8262250Z [       OK ] BasicCurlHttpTests.RetryPolicyEnabled (4 ms)
2026-07-28T14:56:26.8288310Z [ RUN      ] BasicCurlHttpTests.RetryPolicyDisabled
2026-07-28T14:56:26.8291560Z [       OK ] BasicCurlHttpTests.RetryPolicyDisabled (4 ms)
2026-07-28T14:56:26.8314940Z [ RUN      ] BasicCurlHttpTests.ExponentialBackoffRetry
2026-07-28T14:56:26.8318290Z [       OK ] BasicCurlHttpTests.ExponentialBackoffRetry (20 ms)
2026-07-28T14:56:26.8333550Z [ RUN      ] BasicCurlHttpTests.SendGetRequestSync
2026-07-28T14:56:26.8336190Z [       OK ] BasicCurlHttpTests.SendGetRequestSync (4 ms)
2026-07-28T14:56:26.8352790Z [ RUN      ] BasicCurlHttpTests.HandlerRequestedCloseKeepsServerUsable
2026-07-28T14:56:26.8355920Z Assertion failed: (it != m_sockets.end()), function onThread, file socket_tools.h, line 817.
2026-07-28T14:56:26.8374620Z ================================================================================
```

thc1006:
Thanks for catching that. Root cause and fix are pushed.

`kevent()` and `epoll_wait()` both hand back a batch of events. Handling one of them can close a connection, and `removeSocket()` drops that socket from `m_sockets` straight away, so a later entry in the same batch can still name it. The lookup then reaches `m_sockets.end()`, and reading `it->socket` goes through that iterator. The assert stops a debug build there. Without `NDEBUG` nothing does.

A handler asking to close now happens inside a read callback, which is what makes the new test reach it. Both loops skip an event whose socket is no longer registered. The Windows path takes one event per wait rather than a batch, so I left it alone.

macOS Bazel and macOS Conan were the same test and the same assertion, so this covers both.

marcalff:
This patch changes test code only, no code change in SDK or exporters.

PR #4292, opened 2026-07-24
[TEST] Make SocketAddr string parsing safe and reject malformed addresses

### LINKED ISSUE #4290: [BUG] Out-of-bounds write in SocketAddr(char const *) on Windows
Part of #4287.

`SocketAddr(char const *)` copies the address into a `WCHAR` buffer using a byte count as an element count.

https://github.com/open-telemetry/opentelemetry-cpp/blob/9d38caf20eacb2618bcd5d9536aba0efcc2990aa/ext/include/opentelemetry/ext/http/server/socket_tools.h#L186-L194

```cpp
WCHAR buf[200];
for (int i = 0; i < sizeof(buf) && addr[i]; i++)
{
  buf[i] = addr[i];
}
buf[199] = L'\0';
```

`sizeof(buf)` is 400 on a platform with a 2 byte `WCHAR`, not 200. An address string longer than 200 characters keeps writing `buf[200]` through `buf[399]`, which is 400 bytes past the end of the array.

The loop also assigns a possibly negative `char` directly to a `WCHAR`, which sign extends for bytes above 0x7F.

### Suggested fix

Bound the loop with `sizeof(buf) / sizeof(buf[0]) - 1` and terminate at the number of characters actually copied. Note that the repository targets C++14, so `std::size` is not available. Widening should go through `unsigned char`.

The `WSAStringToAddressW` return value is also ignored, so a malformed address leaves the result at whatever `m_data` was value initialized to. That is worth handling in the same change.

I would like to work on this one, together with #4291, since both are in the same constructor.

### LINKED ISSUE #4291: [BUG] SocketAddr(char const *) reads uninitialized bytes when parsing host:port
Part of #4287.

`SocketAddr(char const *)` terminates its host buffer at a fixed index rather than at the end of what it copied.

https://github.com/open-telemetry/opentelemetry-cpp/blob/9d38caf20eacb2618bcd5d9536aba0efcc2990aa/ext/include/opentelemetry/ext/http/server/socket_tools.h#L213-L216

```cpp
char buf[16];
memcpy(buf, addr, (std::min<ptrdiff_t>)(15, colon - addr));
buf[15] = '\0';
::inet_pton(AF_INET, buf, &inet4.sin_addr);
```

`buf` is uninitialized and `memcpy` writes only the host part. For `"127.0.0.1:8800"` that is 9 bytes, so `buf[9]` through `buf[14]` keep indeterminate values while only `buf[15]` is set to `'\0'`. `inet_pton` then reads past the copied text until it finds that terminator, so its input includes indeterminate bytes.

`inet_pton` requires the whole string to be a valid dotted quad, so any non-zero garbage makes it fail. Its return value is ignored, and `m_data` is value initialized, so the observable result is a silent fall back to `0.0.0.0` rather than an arbitrary address. In practice the stack bytes are often zero and the parse succeeds, which is why this has not been noticed.

### Suggested fix

Terminate at the number of bytes actually copied instead of at index 15, and check the `inet_pton` return value.

I would like to work on this one, together with #4290, since both are in the same constructor.

thc1006:
Fixes #4290
Fixes #4291
Refs #4287

This started as the two memory-safety fixes for #4290/#4291 and grew, in response to review, into one address parser shared by both platforms with a clear valid/invalid contract. `SocketAddr(char const *)` has no in-repo caller (the server uses the `(u_long, int)` overload), so the stricter failure semantics change no existing call site.

## The memory-safety bugs (original scope)

- **POSIX** terminated the host buffer at a fixed index, so `inet_pton` read uninitialized bytes for any host shorter than 15 characters.
- **Windows** bounded the host copy loop with `sizeof(buf)`, a byte count, on a `WCHAR[200]`, so an address over 200 characters wrote past the end.

## The fix: one parser for both platforms

Rather than patch two separate paths (POSIX `strtol` + `inet_pton`, Windows `WSAStringToAddress`), both now use the same parser: `inet_pton` for the host (Winsock has provided it since Vista, via `<ws2tcpip.h>`) plus a strict decimal port. It parses into a local `sockaddr_in` and commits it into the storage with `memcpy` only on success, so a failed parse leaves the object at `AF_UNSPEC` with `port() == -1`. This removes the reasons the old Windows path was fragile (`WSAStringToAddress` returned `WSAEINVAL` unless `sin_family` was preset) by not using it at all.

It also fixes the contract issues review surfaced:

- A failed parse used to leave a usable `0.0.0.0:<port>`; it now leaves `AF_UNSPEC`, distinguishable from a real endpoint including the legitimate `:0`.
- An overlong host (`255.255.255.2559`) used to truncate into a different valid address; it is now rejected.
- The port grammar is decimal digits only with a pre-multiply overflow check, so trailing garbage (`80junk`, `80:90`), a sign or whitespace (`+80`, `-0`, ` 80`), and out-of-range values are all rejected on both platforms rather than relying on `strtol`'s locale-sensitive leniency.

## Storage layout made explicit

`port()`/`toString()` and the parser used to access the `sockaddr` storage through a `sockaddr_in` reference (an alignment/type-access issue). They now `memcpy` a local `sockaddr_in` in and out. Three `static_assert`s pin the assumptions this and the socket syscalls rely on: `sizeof(sockaddr) == sizeof(sockaddr_in)` (a compile-time restriction matching this wrapper's existing IPv4-only ABI assumption rather than a general C++/POSIX guarantee, so the length passed to `connect()`/`bind()` is exactly right), the address-family field at the same offset, and `sizeof(SocketAddr) == sizeof(sockaddr)`. The `(u_long, int)` constructor still uses a `reinterpret_cast` and truncates out-of-range ports; it has an `addListeningPort` caller, so it is tracked in #4307 rather than changed here.

## Build

`ws2_32` is declared as an interface dependency of `opentelemetry_ext` (CMake and Bazel), so consumers and the test inherit it rather than relinking it; the header's `#pragma comment(lib)` autolink is MSVC-only and MinGW/GCC ignore it. CMake propagates `ws2_32` to consumers including non-MSVC toolchains, and Bazel propagates the corresponding linker option for the repository's Windows toolchain; MinGW compilation is not currently covered by CI.

## Tests

`socket_tools_test.cc` has 17 cases. Failures are asserted through an `ExpectInvalid` helper that checks both `AF_UNSPEC` and `port() == -1`, so a corrupted family byte cannot pass, plus positive cases for a legitimate `:0`, a leading-zero port (`080` parses as 80), the longest representable host, and the truncation, trailing-garbage and sign rejections above.

## Verification

- 17 tests pass under `-fsanitize=address,undefined`. The shorthand form (`127.1`) is rejected on every platform, since `inet_pton` requires four decimal components, and is covered by a test; leading-zero components (`01.02.03.004`) are deliberately not asserted, because glibc rejects them while macOS `inet_pton` accepts them.
- clang-format, cmake-format (with the repo config) and buildifier clean; clang-tidy 18 on `socket_tools.h` is unchanged at 2 pre-existing warnings.
- Windows CMake CI runs the tests and Windows Bazel CI is build-only; a reviewer confirming a real Windows parse is still valuable, but using `inet_pton` removes the Windows-specific `WSAStringToAddress` grammar and its `sin_family` precondition.


marcalff (approved):
LGTM

marcalff:
This PR changes test code only.
The mock HTTP server used for unit tests should be more robust now.

PR #4294, opened 2026-07-24
[BUG] Stop embedded HTTP server writes from raising SIGPIPE

### LINKED ISSUE #4293: [BUG] Embedded HTTP server writes can raise SIGPIPE and terminate the process
Part of #4287.

`Socket::send()` passes a flags argument of `0`:

https://github.com/open-telemetry/opentelemetry-cpp/blob/9d38caf20eacb2618bcd5d9536aba0efcc2990aa/ext/include/opentelemetry/ext/http/server/socket_tools.h#L355-L359

There is no `MSG_NOSIGNAL` and no `SO_NOSIGPIPE` anywhere under `ext/`.

POSIX specifies that a write to a connected socket that can no longer transmit raises `SIGPIPE` as well as failing with `EPIPE`, unless the write suppresses it. The default disposition of `SIGPIPE` terminates the process, so an embedded server whose peer has gone away can take the host process down before any of the error handling in `sendMore()` runs.

Reproduced on Linux with a `socketpair()` whose peer end is closed, forking once per flag value:

```
  flags = 0             -> KILLED by signal 13 (Broken pipe)
  flags = MSG_NOSIGNAL  -> exited 0
```

### Suggested fix

Handle it inside `Socket::send()` rather than changing the host process signal disposition, since a library should not be deciding that for the program that embeds it.

- Linux and other platforms that define `MSG_NOSIGNAL`: pass it per call.
- macOS and BSD, which offer `SO_NOSIGPIPE` instead: set the socket option once, at socket creation and again on the socket returned by `accept()`.
- Windows: nothing to do, there is no `SIGPIPE`.

Worth doing before the fatal send error item in #4287, because otherwise the process can die before that error handling is ever reached.

thc1006:
Fixes #4293
Refs #4287

## The problem

`Socket::send()` passed a flags argument of `0`, and there was no `MSG_NOSIGNAL` or `SO_NOSIGPIPE` anywhere under `ext/`. POSIX raises `SIGPIPE` on a write to a connected socket that can no longer transmit, in addition to failing with `EPIPE`, and its default disposition terminates the process. The embedded server could therefore be killed before any of the error handling in `sendMore()` ran.

## The fix

- `send()` passes `MSG_NOSIGNAL` on every platform that defines it. That is Linux and the BSDs, **not** macOS, which offers `SO_NOSIGPIPE` instead. This repo's POSIX CI runs Linux and macOS, so both mechanisms are exercised but on different runners.
- `MSG_NOSIGNAL` is the guaranteed per-send protection wherever the platform defines it. Where the platform offers `SO_NOSIGPIPE` instead or as well, the option is set once per socket, in the `(af, type, proto)` constructor and on the socket handed back by `accept()`. On the BSDs, which define both, that is a second layer. On macOS, which does not define `MSG_NOSIGNAL` at all, it is the only mechanism. Either way it is best effort: a `setsockopt` failure is logged rather than failing the socket. This PR therefore does not make a hard guarantee on macOS, where `SO_NOSIGPIPE` is the only mechanism available and a failed `setsockopt` does not invalidate the socket. The header comment beside that call says so; an earlier revision of it claimed `MSG_NOSIGNAL` covered the gap, which is exactly backwards for the platform the branch exists for.
- Windows has no `SIGPIPE` and compiles both branches out.

This PR prevents process termination only. The fatal send-error connection cleanup (closing the socket, removing the reactor registration, changing state) is unchanged and remains tracked in #4287.

## Testing

`ext/test/http/socket_sigpipe_test.cc` is a fixture-based test. Fixture-member sockets are closed in `TearDown`, while raw socketpair descriptors remain protected by a local `ScopedFd` whose destructor runs on every exit path, including a fatal assertion. `TearDown` then drains any pending `SIGPIPE` and restores the previous signal mask, so a leaked descriptor, signal mask, or queued `SIGPIPE` cannot pollute other tests sharing the binary.

- `gtest_add_tests` registers each case as its own CTest process, so no case can rely on another to set up the signal environment. Each signal-behavior case therefore guards itself and skips if `SIGPIPE` is `SIG_IGN` in its own process, because POSIX does not require an ignored blocked signal to remain pending. In that environment the absence of a pending `SIGPIPE` cannot portably prove that the socket-level suppression mechanism was responsible.
- A baseline (`RawSendToClosedPeerRaisesSigPipe`) asserts that a raw `send(..., 0)` to a closed peer does raise `SIGPIPE`. This is an independent case, not a precondition for the others.
- The two suppression mechanisms are verified independently, so a regression in either is caught: `WrapperSendToClosedPeerUsesMsgNoSignal` calls the wrapper `send()` on a raw-fd socket without `suppressSigPipe()` (only `MSG_NOSIGNAL` in play), and `SuppressSigPipeGuardsRawSendToClosedPeer` calls `suppressSigPipe()` then a raw `send(..., 0)` that bypasses `MSG_NOSIGNAL` (only `SO_NOSIGPIPE` in play). Each asserts `EPIPE` with no pending `SIGPIPE`.
- Where `SO_NOSIGPIPE` is defined, two tests read it back with `getsockopt()`: one on a `(af, type, proto)`-constructed socket, and one on an `accept()`ed socket whose listener has its `SO_NOSIGPIPE` explicitly cleared to 0 first, so a value of 1 proves `accept()`'s own `suppressSigPipe()` rather than inheritance.
- The `SO_NOSIGPIPE` cases skip at runtime where the option is absent (Linux), and Windows gets an explicit skipped test.

The CMake target is POSIX-only and registered with `gtest_add_tests`, the convention used elsewhere in this tree. Every `TEST_F` declaration is present in the POSIX binary (platform-dependent cases skip at runtime rather than being compiled out), so the source-parsing `gtest_add_tests` registers only tests the binary actually contains. The Bazel target builds on all platforms; on Windows it compiles the single skipped test rather than an empty binary, and links `ws2_32` for `socket_tools.h`'s `WsaInitializer` global.

## On #4292 and merge order

This PR uses a separate `socket_sigpipe_test.cc`, so it does not depend on #4292's test source. A rebase may still be needed after either PR lands, since both touch `CHANGELOG.md`, `ext/test/http/BUILD` and `ext/test/http/CMakeLists.txt`.

## Verification

- On Linux the tests build and run (the `MSG_NOSIGNAL` and negative-control cases run, the `SO_NOSIGPIPE` cases skip); the `SO_NOSIGPIPE` branch compiles under `-Wall -Wextra -Werror` in both C++14 and C++17.
- clang-format, cmake-format (with the repo `.cmake-format.py`) and buildifier are clean; GCC maintainer mode (`-Wlogical-op -Werror`) is clean.
- `MSG_NOSIGNAL` and `SO_NOSIGPIPE` are both used behind `#ifdef`, so no platform sees an undefined identifier.


thc1006:
Closing following @dbarker's guidance on #4287, same reasoning as #4283: `socket_tools.h` is internal and the bar for it is the tests and the example. I have noted on #4287 that an unhandled SIGPIPE ends the test process rather than failing a call, in case that ever matters for the testing bar, but nothing here needs a decision. Happy to reopen.

PR #4300, opened 2026-07-24
[BUG] Make file_http_server.h compile on Windows

### LINKED ISSUE #4299: [BUG] file_http_server.h does not compile on Windows
Part of #4287.

`FileHttpServer::FileGetSuccess()` cannot compile on Windows.

https://github.com/open-telemetry/opentelemetry-cpp/blob/9d38caf20eacb2618bcd5d9536aba0efcc2990aa/ext/include/opentelemetry/ext/http/server/file_http_server.h#L52-L56

```cpp
bool FileGetSuccess(const std::string &filename, std::vector<char> &result)
{
#ifdef _WIN32
  std::replace(filename.begin(), filename.end(), '/', '\');
#endif
```

`filename` is a reference to const, so `begin()` and `end()` yield const iterators and `std::replace` cannot write through them. This is an ordinary C++ error rather than an MSVC quirk. Compiling that branch with clang 18 gives:

```
error: cannot assign to return value because function 'operator*' returns a const value
 4337 |           *__first = __new_value;
note: in instantiation of function template specialization
  'std::replace<__gnu_cxx::__normal_iterator<const char *, std::basic_string<char>>, char>'
  requested here
```

gcc 12 reports the same thing.

It has gone unnoticed because no translation unit in the repository includes `file_http_server.h`, so CI never compiles it on any platform. It is an installed header, so a downstream consumer building on Windows hits it immediately.

`<algorithm>` is also missing from this file. `std::replace` currently resolves only because `socket_tools.h` happens to include it.

### Suggested fix

Substitute the separators on a local copy, and include `<algorithm>` directly.

A compile-only target that includes every installed `ext` header on Linux, macOS and Windows would have caught this, and other installed headers with no in-repo includer are likely in the same position. Happy to propose one separately if that would be welcome.

thc1006:
Fixes #4299
Refs #4287

## The problem

`FileHttpServer::FileGetSuccess()` called `std::replace(filename.begin(), filename.end(), ...)` on a `const std::string &`, which yields const iterators, so the `_WIN32` branch could not compile. clang 18 and gcc 12 both reject it; it is plain C++, not an MSVC quirk. It survived because no translation unit in the repository includes this installed header, so CI never compiled it on any platform.

## The fix

The separator substitution runs on a local copy on Windows; the `#else` branch binds a reference so other platforms do not pay for a copy. `<sstream>` is now included directly for `std::ostringstream`, and `<algorithm>` (for the Windows-only `std::replace`) is included under `_WIN32`; both were previously resolving only because `socket_tools.h` happens to include them.

## Regression protection (added in review)

The header still had no test compiling it, so the fix could regress unnoticed. A compile-only target (`file_http_server_header_compile`, a CMake OBJECT library and a Bazel `cc_library`) now includes the header as its only translation unit. CI compiles the POSIX branch on Linux and macOS and the `_WIN32` branch on the Windows runners, so a compile-time break in the active platform branch of this header now fails the build. It is compiled, not run. It does not prove direct-include hygiene, since a symbol can still resolve through a transitive header; that is checked separately by the repository's IWYU workflow. The probe uses no symbol from the header on purpose, so its include is marked `// IWYU pragma: keep`.

Compiling the header for the first time under the maintainer configuration also surfaced pre-existing issues in it that no build had seen before: three extra semicolons after in-class function bodies (errors under `-Wextra-semi -Werror`) and two single-character string-literal searches, `find(".")` and `find_last_of(".")`, that `clang-tidy performance-faster-string-find` flags. Both are fixed here so the header clears the same gates as the rest of the tree rather than raising a limit.

## Not in this PR

The header has neighbouring robustness gaps (unchecked `tellg()`, ignored `read()` result, non-ASCII filenames, URI query handling) and a path-confinement concern already being handled privately. They are recorded in #4287 rather than folded into this scoped build fix.

## Verification

- The compile-only target builds under `-Wall -Wextra -Werror -Wextra-semi` on clang for C++14, 17 and 20, and on the repository's Windows runners for the `_WIN32` branch.
- clang-format 18 clean; the CHANGELOG entry is within the markdown line-length limit.
- Rebased onto current `main`, keeping both the #4300 and #4303 CHANGELOG entries; GitHub reports the branch mergeable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Increment the:
   namespaces
   [#4303](https://github.com/open-telemetry/opentelemetry-cpp/pull/4303)
 
+* [BUG] Make file_http_server.h compile on Windows
+  [#4300](https://github.com/open-telemetry/opentelemetry-cpp/pull/4300)
+
 * [CODE HEALTH] Move metrics storage test fixtures into anonymous namespace
   [#4286](https://github.com/open-telemetry/opentelemetry-cpp/pull/4286)
 
--- a/ext/include/opentelemetry/ext/http/server/file_http_server.h
+++ b/ext/include/opentelemetry/ext/http/server/file_http_server.h
@@ -85,7 +85,7 @@ private:
    */
   std::string GetMimeContentType(const std::string &filename)
   {
-    std::string file_ext = filename.substr(filename.find_last_of(".") + 1);
+    std::string file_ext = filename.substr(filename.find_last_of('.') + 1);
     auto file_type       = mime_types_.find(file_ext);
     return (file_type != mime_types_.end()) ? file_type->second : HTTP_SERVER_NS::CONTENT_TYPE_TEXT;
   };
@@ -104,7 +104,7 @@ private:
     }
     // If filename appears to be a directory, serve the hypothetical index.html
     // file there
-    if (name.find(".") == std::string::npos)
+    if (name.find('.') == std::string::npos)
       name += "/index.html";
 
     return name;
--- a/ext/include/opentelemetry/ext/http/server/file_http_server.h
+++ b/ext/include/opentelemetry/ext/http/server/file_http_server.h
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <fstream>
 #include <iostream>
+#include <sstream>
 #include <string>
 #include <unordered_map>
 #include <vector>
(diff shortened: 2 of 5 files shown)


owent:
This is dead code and will be removed in #4306 . Maybe this PR can be closed.

thc1006:
Superseded by #4306, which removes `file_http_server.h` entirely. That's the better call since the header is unused, so there's nothing left to fix here. Closing.

Thanks @owent for the cleaner fix and @dbarker for landing it.

PR #4302, opened 2026-07-24
[CODE HEALTH] Move SDK common and logs test helpers into anonymous namespaces

thc1006:
Part of #4196

Three `misc-use-internal-linkage` sites in the SDK tests, the second of three batches after #4301.

| file | classes |
|---|---|
| `sdk/test/common/global_log_handle_singleton_lifetime_test.cc` | `GlobalLogHandlerChecker`, `CustomLogHandler` |
| `sdk/test/logs/logger_provider_set_test.cc` | `TestProvider` |

## The one that needed care

`global_log_handle_singleton_lifetime_test.cc` is a destruction-order test: `GlobalLogHandlerChecker`'s destructor must run after the global log handler and aborts otherwise.

Review noted that the existing assertion could not actually catch a reversed order: `handle && custom_handler_destroyed` only aborts when the handler destructor has run yet a handle is still reachable, so a checker destroyed first (the case the test is for) slipped through. That is a pre-existing weakness, not something the namespace move caused, but since this PR leans on the test I tightened the predicate to `!custom_handler_destroyed || handle` so a reversed order now aborts. The correct order still passes.

It already had an anonymous namespace holding only `ConstructChecker()`, with one flagged class above it and the other below, so I extended that namespace over both instead of adding a second one.

I built and ran the test rather than reasoning about it:

```
[ RUN      ] GlobalLogHandleSingletonTest.Lifetime
GlobalLogHandlerChecker constructed
[       OK ] GlobalLogHandleSingletonTest.Lifetime (4 ms)
[  PASSED  ] 1 test.
Custom Gobal Log Handle destroyed
GlobalLogHandlerChecker destroyed and check pass.
```

The handler is destroyed first and the checker's check passes, which is the invariant the test exists for. That is what I expected, since an anonymous namespace changes linkage rather than storage duration and the function-local static is still constructed on its first call from the test body, but the test is cheap to run and reasoning about destruction order is a bad place to be confident.

`logger_provider_set_test.cc` is the ordinary case: a namespace around the test body.

## Warning limit

151 to 148 and 161 to 158. Three on each preset, since none of these are ABI gated, unlike the `TestEventLoggerProvider` site in #4301.

This branch is cut from main rather than stacked on #4301, so if that one merges first this needs a rebase and the limit recomputed to 142 and 153. Happy to do that whenever it is useful; I kept them independent so they can be reviewed in either order.

## Verification

- Both classes and `TestProvider` end up inside an anonymous namespace, checked by walking namespace depth, and both files have balanced open and close.
- Both files compile with `-DOPENTELEMETRY_ABI_VERSION_NO=1` and `=2`.
- The lifetime test builds and runs green, with the destruction order shown above.
- clang-format 18 clean.
- Neither `sdk/test/common` nor `sdk/test/logs` shares a `TEST` suite name across files, so this should not repeat the `all_tests` redefinition from #4217.

Same honest gap as #4301: I could not re-run `misc-use-internal-linkage` locally, because it only reports classes from llvm-22 onward and clang-tidy 18, 20 and 21 report functions and variables only, so a local run returns zero regardless. The site list comes from the CI log artifact for main, and CI is the authority on the resulting count.


sdk/test/common/global_log_handle_singleton_lifetime_test.cc
```diff
@@ -19,9 +22,13 @@ class GlobalLogHandlerChecker
   {
     using opentelemetry::sdk::common::internal_log::GlobalLogHandler;
     auto handle = GlobalLogHandler::GetLogHandler();
-    if (handle && custom_handler_destroyed)
+    // The checker must outlive the global handler. Fail if it is destroyed first
+    // (custom_handler_destroyed still false) or if the handler is somehow still reachable
+    // (handle non-null). The old "handle && custom_handler_destroyed" only caught the second
+    // case, so a reversed order slipped through.
+    if (!custom_handler_destroyed || handle)
```
marcalff, line 28: Comments **in the code** should be about the **code**.

Comments **about the code history** should be with the code history, aka in the PR explaining what changed and why.

Here, the comment:

> The old "handle && custom_handler_destroyed" only caught the second case, so a reversed order slipped through.

is out of place and misleading to a reader.

marcalff (changes requested):
Thanks for the fix.

See a minor change about a comment.

The change itself is minor, but taking this opportunity to illustrate what are the best practices to follow.

sdk/test/common/global_log_handle_singleton_lifetime_test.cc
thc1006: Thanks, this is a helpful distinction to keep in mind. You are right that the sentence was code history rather than something a reader of the code needs.

I have trimmed the comment to describe only what the check does, and kept the reason for the change in the commit message (ca2de05e).

marcalff (approved):
LGTM, thanks for the fix.

PR #4303, opened 2026-07-24
[CODE HEALTH] Move SDK trace and metrics test helpers into anonymous namespaces

thc1006:
Part of #4196

This is the last of three batches after #4301 and #4302, and clears the SDK test sites in #4196.

| file | classes |
|---|---|
| `sdk/test/metrics/exemplar/aligned_histogram_bucket_exemplar_reservoir_test.cc` | `AlignedHistogramBucketExemplarReservoirTestPeer` |
| `sdk/test/metrics/measurements_benchmark.cc` | `MockMetricExporter` |
| `sdk/test/trace/batch_span_processor_test.cc` | `MockSpanExporter`, `BatchSpanProcessorTestPeer`, `MockLogHandler` |
| `sdk/test/trace/tracer_provider_set_test.cc` | `TestProvider` |
| `sdk/test/trace/tracer_test.cc` | `MockSampler`, `MockDecisionSampler`, `MockIdGenerator` |

## Shape of the change

`tracer_test.cc` and `measurements_benchmark.cc` already had an anonymous namespace that stopped just short of the flagged classes, so those are extended rather than joined by a second one.

`batch_span_processor_test.cc` is the one worth a second look. It already sits inside a named namespace opened by `OPENTELEMETRY_BEGIN_NAMESPACE`, so the anonymous namespace nests inside that and closes just before `OPENTELEMETRY_END_NAMESPACE`. It covers every class and every `TEST` in the file, including the `BatchSpanProcessorOptionsEnvTest` group at the end.

`tracer_provider_set_test.cc` closes its namespace after the trailing `#endif` of the `NO_GETENV` block, so the placement holds whether or not that block is compiled.

## Warning limit

151 to 142 and 161 to 152, nine on each preset since none are ABI gated.

Cut from main rather than stacked on #4301 and #4302. If any of the three merge before this, the limit needs recomputing down by the merged batch's count. Happy to rebase whenever; they were kept independent so they can be reviewed in any order.

## Verification

- All nine classes end up inside an anonymous namespace, checked by finding the nearest namespace open and close around each rather than by eye.
- Every file compiles under `-DOPENTELEMETRY_ABI_VERSION_NO=1` and `=2`.
- `BatchSpanProcessorTestPeer` and `AlignedHistogramBucketExemplarReservoirTestPeer` are the only `TEST_F` suite names in the batch, and each appears in exactly one file, so no `all_tests` redefinition. This was the specific risk from #4217 and the main thing I wanted to rule out.
- `measurements_benchmark.cc` is a benchmark target, so it needs to link rather than only compile; the change is a namespace wrap and does not touch linkage of any symbol the benchmark harness needs.
- clang-format 18 clean.

Same honest gap as the earlier two batches: `misc-use-internal-linkage` only reports classes from llvm-22 onward, so a local clang-tidy run returns zero regardless and proves nothing. The site list is from the CI log artifact for main, and CI is the authority on the count.

### DIFF
--- a/.github/workflows/clang-tidy.yaml
+++ b/.github/workflows/clang-tidy.yaml
@@ -17,9 +17,9 @@ jobs:
       matrix:
         include:
           - cmake_options: all-options-abiv1-preview
-            warning_limit: 151
+            warning_limit: 142
           - cmake_options: all-options-abiv2-preview
-            warning_limit: 161
+            warning_limit: 152
     env:
       CC: /usr/bin/clang-22
       CXX: /usr/bin/clang++-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Increment the:
 
 ## [Unreleased]
 
+* [CODE HEALTH] Move SDK trace and metrics test helpers into anonymous
+  namespaces
+  [#4303](https://github.com/open-telemetry/opentelemetry-cpp/pull/4303)
+
 * [CODE HEALTH] Move metrics storage test fixtures into anonymous namespace
   [#4286](https://github.com/open-telemetry/opentelemetry-cpp/pull/4286)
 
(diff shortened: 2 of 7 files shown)


marcalff (approved):
LGTM

PR #4306, opened 2026-07-25
Remove file_http_server.h

### LINKED ISSUE #4299: [BUG] file_http_server.h does not compile on Windows

owent:
Fixes #4299

## Changes

- Remove file_http_server.h, it's not used for a long time.


marcalff (approved):
LGTM, this is dead code.

PR #4309, opened 2026-07-26
[CONFIGURATION] Implement EventToSpanEventBridge log record processor

### LINKED ISSUE #4155: [CONFIGURATION] File configuration - EventToSpanEventBridge
Implement the following types from the declarative configuration spec:

    type: ExperimentalEventToSpanEventBridgeLogRecordProcessor

om7057:
Fixes #4155

Implements ExperimentalEventToSpanEventBridgeLogRecordProcessor from the declarative configuration spec. Mirrors the EventToSpanEventBridge processor already in the Java SDK: if a log record has an event name and its trace/span id match the currently active span, it gets copied over as a span event on that span. Doesn't export anything itself and doesn't stop the record from continuing through the rest of the pipeline.

YAML key is event_to_span_event_bridge/development, matching the schema in open-telemetry/opentelemetry-configuration (checked opentelemetry_configuration.json and meta_schema_language_cpp.yaml, which list this type as not_implemented with no extra properties).

New processor class under sdk/include/opentelemetry/sdk/logs/event_to_span_event_bridge_processor.h + .cc, wired into the config parser, the visitor, and SdkBuilder the same way simple/batch already are.

Tests: sdk/test/logs/event_to_span_event_bridge_processor_test.cc covers bridging onto the current matching span, ignoring records with no event name, and ignoring events from a different span. sdk/test/configuration/yaml_logs_test.cc covers the YAML parsing side.

Ran the full local test suite (915 tests) before pushing, all green. CHANGELOG updated.


om7057:
@dbarker @marcalff could you please review this PR?

sdk/include/opentelemetry/sdk/logs/event_to_span_event_bridge_processor.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A log record processor that bridges log records representing events (i.e. log records with a
+ * non-empty event name) onto the current active span as span events.
+ *
+ * A log record is bridged to a span event if and only if:
+ *  - the log record has a non-empty event name,
+ *  - the log record carries a valid trace id and span id,
+ *  - the current active span (from the ambient context at the time OnEmit is called) is
+ *    recording, and
```
lalitb: How do we know the ambient span at `OnEmit()` is the span from the log's resolved context?

The Logs SDK specification requires using the resolved context: the explicitly supplied context when present, otherwise the ambient context. These can be different. For example, the log may be emitted with `span_a` explicitly while `span_b` is currently ambient. In that case this processor finds `span_b`, its IDs do not match the log, and the event is silently dropped.

The current `LogRecordProcessor::OnEmit()` API does not receive the resolved context, so I don't think `Tracer::GetCurrentSpan()` can implement the required behavior correctly.

lalitb (changes requested):
As mentioned [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4309#discussion_r3661196371) - . This violates the Logs SDK specification for explicitly supplied contexts and can silently drop valid events. The current C++ `LogRecordProcessor::OnEmit()` API does not expose that context, so the API gap
  needs to be addressed before this bridge can be implemented correctly.

om7057:
> How do we know the ambient span at `OnEmit()` is the span from the log's resolved context?
> 
> The Logs SDK specification requires using the resolved context: the explicitly supplied context when present, otherwise the ambient context. These can be different. For example, the log may be emitted with `span_a` explicitly while `span_b` is currently ambient. In that case this processor finds `span_b`, its IDs do not match the log, and the event is silently dropped.
> 
> The current `LogRecordProcessor::OnEmit()` API does not receive the resolved context, so I don't think `Tracer::GetCurrentSpan()` can implement the required behavior correctly.

@lalitb Thank you for pointing this out. The current implementation doesn't handle explicitly supplied contexts properly; it will silently drop events from logs that were emitted with a different span context than what's currently active.

Would it be acceptable to:

Add a method to the Recordable interface to store/retrieve the resolved context
Update Logger to set it when creating the recordable
Have the bridge use it to properly match against the log's context
This way we keep existing processor signatures unchanged while giving processors access to the context they need. What do you think?

sdk/include/opentelemetry/sdk/logs/event_to_span_event_bridge_processor.h
om7057: @lalitb Got it. I didn't fully consider the case where a log is explicitly created with one span context while a different span is ambient.

I worked on this by storing resolved context on the Recordable interface:

Added SetResolvedContext() and GetResolvedContext() methods to the Recordable base class
Updated ReadWriteLogRecord to store and retrieve the context
Updated Logger to set the resolved context when creating recordables
Modified EventToSpanEventBridge to use it for proper span matching
This keeps the LogRecordProcessor API stable while giving the bridge access to the log's actual context. The changes are in the latest commits on this branch.

lalitb: I don’t think `Recordable` should store the resolved context. Its purpose is to carry the log record through the processing and export pipeline, and it may be retained by a batch processor. Keeping a `Context` there can also keep the span alive longer than needed.

 The resolved context should be passed to the processor during emission instead. Since `OnEmit()` is a public SDK extension point, we should keep it for compatibility and add a separate `OnEmitWithContext(record, resolved_context)` method that forwards to `OnEmit(record)` by default. Using a separate name avoids C++ overload hiding. The logger and multi-processor can pass the resolved context through, while only the bridge needs to override the new method. The context should only be valid for the duration of the call and should not be retained.

om7057: @lalitb Thank you for the guidance. I've implemented the solution exactly as you suggested:

- Added `OnEmitWithContext(record, context)` to `LogRecordProcessor` with a default implementation that forwards to `OnEmit()` for backward compatibility
- Updated `Logger` to call `OnEmitWithContext()` with the resolved context
- Updated `MultiLogRecordProcessor` to forward the context to child processors
- Implemented `EventToSpanEventBridge` to override `OnEmitWithContext()` and use the resolved context for proper span matching

The context is passed by reference and never retained. All changes are pushed and ready for review.

om7057:
@lalitb Thank you for the guidance. I've implemented the solution exactly as you suggested:

- Added `OnEmitWithContext(record, context)` to `LogRecordProcessor` with a default implementation that forwards to `OnEmit()` for backward compatibility
- Updated `Logger` to call `OnEmitWithContext()` with the resolved context
- Updated `MultiLogRecordProcessor` to forward the context to child processors
- Implemented `EventToSpanEventBridge` to override `OnEmitWithContext()` and use the resolved context for proper span matching

The context is passed by reference and never retained. All changes are pushed and ready for review.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -60,6 +60,26 @@ class LogRecordProcessor
...
+   * with access to the resolved context (explicit if provided, else ambient).
+   * The resolved context is valid only for the duration of this call and must not be retained.
+   *
+   * The default implementation forwards to OnEmit() for backward compatibility.
+   * Processors that need access to the log's resolved context should override this method.
+   *
+   * @param record the log recordable object
+   * @param context the resolved context (SpanContext or Context variant)
+   */
+  virtual void OnEmitWithContext(
```
ThomsonTan: This changes the public SDK extension point, please document both compatibility changes explicitly. To avoid the source break, give the visitor method a default no-op implementation. For `LogRecordProcessor`, either introduce the new virtual only in a new ABI version or clearly document that downstream processors must be rebuilt.

sdk/src/logs/event_to_span_event_bridge_processor.cc
```diff
@@ -0,0 +1,309 @@
...
+
+  // Check if the resolved span context matches the log's trace/span IDs
+  if (!(resolved_span_context.trace_id() == log_trace_id) ||
+      !(resolved_span_context.span_id() == log_span_id))
+  {
+    return;
+  }
+
+  // Get the span object for the resolved context and add the event
+  auto current_span = opentelemetry::trace::Tracer::GetCurrentSpan();
```
ThomsonTan: The live span from the resolved `Context` is found and then discarded. In the explicit-context case:

1. [`GetSpan(ctx)`](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/src/logs/event_to_span_event_bridge_processor.cc#L254-L264) returns `span_a`, so the first ID check passes.
2. [`GetCurrentSpan()`](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/src/logs/event_to_span_event_bridge_processor.cc#L273-L294) returns ambient `span_b`.
3. The second ID check compares `span_b` with the log IDs for `span_a` and drops the event.

This recreates the requirement that the resolved span must also be ambient. For the `Context` alternative, retain the `Span` returned by `GetSpan(ctx)` and call `AddEvent` on that object after the recording and ID checks. Handle the bare `SpanContext` alternative separately because it does not carry a live `Span`.

Please add a regression test with explicit `span_a`, ambient `span_b`, and log IDs for `span_a`; assert that only `span_a` receives the event. The [current processor tests](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/test/logs/event_to_span_event_bridge_processor_test.cc#L50-L133) call only `OnEmit`, so this path remains untested.

sdk/test/configuration/yaml_logs_test.cc
```diff
@@ -100,6 +101,27 @@ file_format: "1.0-logs"
...
+    - event_to_span_event_bridge/development:
+)";
+
+  auto config = DoParse(yaml);
+  ASSERT_NE(config, nullptr);
+  ASSERT_NE(config->logger_provider, nullptr);
+  ASSERT_EQ(config->logger_provider->processors.size(), 1);
+  auto *processor = config->logger_provider->processors[0].get();
+  ASSERT_NE(processor, nullptr);
+  auto *bridge = reinterpret_cast<
```
ThomsonTan: `reinterpret_cast` does not check the dynamic type, so [`ASSERT_NE(bridge, nullptr)`](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/test/configuration/yaml_logs_test.cc#L117-L122) only repeats the preceding non-null check on `processor`.

For example, if the YAML key is misspelled as `event_to_span_event_brige/development`, the parser [falls through to `ParseExtensionLogRecordProcessorConfiguration`](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/src/configuration/configuration_parser.cc#L618-L633) and returns a non-null `ExtensionLogRecordProcessorConfiguration`. The cast still produces a non-null pointer, so this test passes even though no bridge configuration was parsed. The neighboring simple/batch tests use the same unsafe cast but at least inspect type-specific parsed fields; this empty configuration has no such signal.

Please make the assertion type-aware without relying on RTTI: pass a test `LogRecordProcessorConfigurationVisitor` to `processor->Accept`, record that [`VisitEventToSpanEventBridge`](https://github.com/om7057/opentelemetry-cpp/blob/ec30ecbb9b8cd1fcd5aa0879c50d5bbdc00a7043/sdk/include/opentelemetry/sdk/configuration/event_to_span_event_bridge_log_record_processor_configuration.h#L18-L24) was called, and fail if `VisitExtension` or any other callback runs. That directly proves this key selected the bridge configuration and catches the fallback case.

sdk/src/logs/logger.cc
```diff
@@ -197,8 +197,10 @@ void Logger::EmitLogRecord(
 
   auto &processor = context_->GetProcessor();
 
-  // Send the log recordable to the processor
-  processor.OnEmit(std::move(recordable));
+  // Send the log recordable to the processor with the resolved context
+  const nostd::variant<trace_api::SpanContext, context::Context> resolved_context{
+      context::RuntimeContext::GetCurrent()};
```
ThomsonTan: This adds an unconditional `RuntimeContext::GetCurrent()` call for every emitted log, even when no bridge is configured and the processor ignores the context. That adds hot-path work for all users.

The three cited lookups are not all on one path: ordinary ABI v2 logging performs two, not three. The remaining issue is the extra dispatch-time lookup, especially for ABI v1, which cannot use the explicit-`Context` path.

Please keep ordinary dispatch on `OnEmit`, let the bridge resolve ambient context only when configured, and pass an already-resolved explicit context through a separate context-aware path. Add a test proving that a processor which ignores context causes no extra `GetCurrent()` call.

sdk/src/logs/event_to_span_event_bridge_processor.cc
```diff
@@ -0,0 +1,309 @@
...
+private:
+  const std::unordered_map<std::string, opentelemetry::sdk::common::OwnedAttributeValue>
+      *attributes_;
+};
+
+}  // namespace
+
+std::unique_ptr<Recordable> EventToSpanEventBridgeProcessor::MakeRecordable() noexcept
+{
+  return std::unique_ptr<Recordable>(new ReadWriteLogRecord());
```
ThomsonTan: Adding this processor makes `MultiLogRecordProcessor` create a bridge `ReadWriteLogRecord` for every emitted log. The record is populated along with every other child recordable; body and attributes become owned copies. For a log with no event name, `OnEmit` then immediately discards that work.

Please avoid eagerly creating the bridge child recordable for non-event logs. An `Enabled` override alone is not enough: another processor may enable the log, after which `MakeRecordable` still creates recordables for every child. This needs per-child filtering or lazy recordable creation in the multi-processor path.

Please add a test with a normal processor plus the bridge and verify that an empty event name does not call the bridge's `MakeRecordable`, while a named event still does.

sdk/src/logs/event_to_span_event_bridge_processor.cc
```diff
@@ -0,0 +1,309 @@
...
+  if (event_timestamp.time_since_epoch().count() == 0)
+  {
+    event_timestamp = log_record->GetObservedTimestamp();
+  }
+
+  EventAttributesKeyValueIterable attributes(log_record->GetAttributes());
+  current_span->AddEvent(event_name, event_timestamp, attributes);
+}
+
+void EventToSpanEventBridgeProcessor::OnEmitWithContext(
```
ThomsonTan: These two entry points (`OnEmit` and `OnEmitWithContext`) duplicate the same validation and emission pipeline, and they have already drifted. Probably extract the common part to a separate function.

sdk/include/opentelemetry/sdk/logs/processor.h
om7057: I went with the "clearly document" option rather than gating on a new ABI version; let me know if you'd prefer the other.

OnEmitWithContext() and ConsumesResolvedContext() both have default implementations, so existing processors compile and behave unchanged. That, and the fact that the added virtuals change the LogRecordProcessor vtable so subclasses must be rebuilt, is now stated in the header and in the CHANGELOG breaking-changes section.

VisitEventToSpanEventBridge() was the actual source of the break; it was pure virtual. It now defaults to a no-op, with a comment explaining why it differs from its siblings.

sdk/src/logs/event_to_span_event_bridge_processor.cc
om7057: Thank you for tracing it step by step, the span from GetSpan(ctx) was found and then thrown away for the ambient one, so the second ID check could never pass in the explicit case.

The processor now adds the event to the span from the resolved Context. The bare SpanContext alternative is handled separately as you asked, since it carries no live Span and there's no way to look one up by id; it resolves through ambient with the ID check rejecting a mismatch. Added BridgesOntoExplicitContextSpanWhileDifferentSpanIsAmbient (explicit span_a, ambient span_b, ids for span_a, asserts only span_a gets the event), which fails against the previous code, plus tests for the mismatch and bare-SpanContext cases.

One gap I should flag rather than paper over: that test drives OnEmitWithContext() directly. End-to-end through Logger the explicit context still doesn't arrive, because Logger::EmitLogRecord() only receives the LogRecord, context_or_span is consumed by CreateLogRecord() and there's no route to emit time (and per the earlier discussion it shouldn't live on the Recordable). So today the SDK always hands the bridge the ambient context. Closing that needs the context-aware emit path you alluded to, which looks like an ABI-v2-guarded addition carrying context_or_span from logger.h down to the processor. Want me to add that in this PR, or land the processor-level fix and do the plumbing separately?

sdk/test/configuration/yaml_logs_test.cc
om7057: Agreed; it just repeated the non-null check above it.

It now passes a recording LogRecordProcessorConfigurationVisitor to processor->Accept() and asserts VisitEventToSpanEventBridge fired while VisitExtension/VisitBatch/VisitSimple did not. No RTTI involved. I also added a companion test using your exact misspelling (event_to_span_event_brige/development) that pins the extension fallback, so the bridge assertion can't silently pass for a non-bridge model.

sdk/src/logs/logger.cc
om7057: Got it, that shouldn't cost anything when nothing reads it. Thanks also for the correction on the lookup count.

Ordinary dispatch is back on OnEmit(). Added LogRecordProcessor::ConsumesResolvedContext() (default false); MultiLogRecordProcessor reports true if any child does, and LoggerContext caches it the same way it already caches RecordableEnforcesLogRecordLimits(). Logger only resolves the context when a processor will read it, so simple/batch pipelines add no lookup. Test installs a counting RuntimeContextStorage and asserts EmitLogRecord() adds zero GetCurrent() calls for a context-ignoring processor.

The second half passes an already-resolved explicit context through a separate path. Currently, the resolved context is always ambient.

sdk/src/logs/event_to_span_event_bridge_processor.cc
om7057: I agree the waste is real, but I couldn't find a fix that actually removes the copy, so I'd rather ask than change shared infrastructure on a guess.

The event name isn't known when MakeRecordable() runs: EmitLogRecord calls CreateLogRecord() first (logger.h:216), and the EventId argument is applied afterwards via LogRecordSetterTrait. Argument order is caller-controlled, so SetEventId can also land after body/attributes. So nothing at MakeRecordable() time can key on it, and the test as specified is only satisfiable by deferring creation.

The catch with deferring: populating the child later means retaining the fields, and the buffer has to be filled unconditionally (it's still unknown whether an event name is coming), so it moves the attribute/body copy rather than removing it. The one variant that does remove it is replaying from a sibling recordable at SetEventId time, since with a simple/batch processor present the data is already there, but that needs a non-RTTI way to discover a sibling is a ReadableLogRecord (there's a nortti CI job), so a new Recordable hook, and it degrades when no readable sibling exists.

Since this touches MultiRecordable/MultiLogRecordProcessor for every user to optimize one optional processor, which direction do you prefer: the sibling-replay hook, something else, or a tracked follow-up?

sdk/src/logs/event_to_span_event_bridge_processor.cc
om7057: Both entry points now share BridgeRecordToSpan(), which holds every precondition and the emission in one place; OnEmit() passes the ambient span and OnEmitWithContext() passes the span from the resolved context.

om7057:
Thanks for the thorough review @ThomsonTan.

Fully addressed: the compatibility documentation and the visitor source break, the type-aware YAML assertion (plus a misspelled-key guard), and the duplicated emit pipeline.

Partially addressed, with one honest caveat: the discarded-span bug is fixed in the processor and covered by a regression test that fails against the old code, and ordinary dispatch is back on OnEmit() with no added GetCurrent() for context-ignoring pipelines. But the explicit context still never reaches the processor end-to-end, because Logger::EmitLogRecord() doesn't receive it, so in practice the bridge is still handed the ambient context. I've left details and a question on the relevant threads.

Still open: the eager child recordable. I couldn't find a fix that actually avoids the copy and would like your steer before touching MultiRecordable; details on that thread.

Verified locally; please check once more that the changes have been made.

om7057:
@lalitb, can you confirm if all the changes requested are addressed in the later commits, and if so could you please approve the PR?

om7057:
@lalitb, just a small reminder: the merge conflicts have been resolved after addressing the request changes please have a look at them.

om7057:
@ThomsonTan @lalitb Everything else from the review is addressed. One thing is still open: [this question](https://github.com/open-telemetry/opentelemetry-cpp/pull/4309#discussion_r3679550377) on whether to add the ABI-v2-guarded `context_or_span` plumbing from `logger.h` down to the processor in this PR, or land the processor-level fix now and track that as a separate follow-up.

Would like to know what can be preferred!

om7057:
@lalitb small ping on this

sdk/src/logs/logger.cc
```diff
@@ -197,6 +197,17 @@ void Logger::EmitLogRecord(
 
   auto &processor = context_->GetProcessor();
 
+  // Only resolve the context when a processor will actually read it: resolving means a
+  // RuntimeContext::GetCurrent() call, which would otherwise be pure overhead on every log for
+  // the common pipelines (simple/batch) that ignore the context entirely.
+  if (context_->ConsumesResolvedContext())
+  {
+    const nostd::variant<trace_api::SpanContext, context::Context> resolved_context{
```
lalitb: @om7057  -   The explicit context is still lost here.  If the log uses `span_a` while `span_b` is ambient, this passes `span_b` to the processor and the event is silently dropped.

Please carry the already resolved context through the logger emit path and add an end-to-end test with explicit `span_a` and ambient `span_b`.

sdk/include/opentelemetry/sdk/logs/processor.h
```diff
@@ -60,6 +61,47 @@ class LogRecordProcessor
...
+
+  /**
+   * Returns true when this processor reads the resolved context passed to
+   * OnEmitWithContext(). The default returns false, so the SDK Logger keeps dispatching through
+   * OnEmit() and does not resolve the ambient context on the emit path.
+   *
+   * Only processors that override OnEmitWithContext() should return true. Composite processors
+   * return true if any child does.
+   */
+  virtual bool ConsumesResolvedContext() const noexcept { return false; }
```
dbarker: Once the context is passed through EmitLogRecord to processor OnEmit, it should be done so unconditionally. This extra virtual method here is not necessary and should be removed.

sdk/src/logs/logger.cc
om7057: Thanks for catching; the resolved context was getting lost. EmitLogRecord(args...) computed context_or_span but then called the record-only overload, which re-derived it from RuntimeContext::GetCurrent() instead of using the one already resolved. Exactly the span_a/span_b case you described.

Fixed in 7696c6b1: added Logger::EmitLogRecordWithContext() (distinctly named, default forwards to EmitLogRecord() so existing implementations are unaffected) and threaded the resolved context through to it. Added ContextAwareProcessorReceivesExplicitContextNotAmbientSpan, which reproduces your scenario and fails against the old code. Verified in a fresh ABI v2 build (my local builds were all ABI v1) plus the full ABI v1 suite

om7057:
@lalitb I've addressed the comment, please check once again!

sdk/src/logs/logger.cc
```diff
@@ -197,6 +197,17 @@ void Logger::EmitLogRecord(
...
+        context::RuntimeContext::GetCurrent()};
+    processor.OnEmitWithContext(std::move(recordable), resolved_context);
```
dbarker: Calling emit with the resolved context is a spec requirement and should occur unconditionally. `ConsumesResolvedContext` should be removed as an unnecessary API expansion of the processor interface. 

As @lalitb  points out the Context must be resolved during the API EmitLogRecord calls (it is either explicitly passed in by the user or the current context is implicitly used).  

We need to account for all supported logger API use cases and document what the expected behavior will be with the new bridge processor for each. 

Here are the uses cases that need to be supported from the API. 

```cpp
Context original_context;
Context context = trace::SetSpan(original_context, my_span); 

// Cases 1-4 conform to the spec since a full Context is resolved and 
// a span ptr can be accessed from the processor. 

// Case 1: call emit with context resolved to explicit context
logger->EmitLogRecord("with explicit context", context);

// Case 2: call emit with context resolved to current context
logger->EmitLogRecord("with implicit current context"); 

// Case 3: call create then emit with ABIv1. Context is resolved to current context OR to 
auto record_abiv1 = logger->CreateLogRecord();
logger->EmitLogRecord(std::move(record_abiv1); 

// Case 4: call create then emit with ABIv2. Context is resolved to the explicit context
auto record_abiv2 = logger->CreateLogRecord(context); 
logger->EmitLogRecord(std::move(record_abiv2); 
 

// Cases 5-7 are supported by the otel-cpp API but don't comply with the spec since a full context cannot be effectively resolved in a way that would make the underlying span ptr accessible.  

SpanContext span_context;        // a user-managed SpanContext (ids + flags only, no live Span)
TraceId trace_id;
SpanId  span_id;
TraceFlags flags;

// Case 5: explicit SpanContext
logger->EmitLogRecord("with span context", span_context);

// Case 6: explicit TraceId + SpanId (+ optional TraceFlags) 
logger->EmitLogRecord("with trace/span id parts", trace_id, span_id, flags);

// Case 7: ABI v2 create-then-emit with explicit SpanContext
auto record_w_span_context= logger->CreateLogRecord(span_context);
logger->EmitLogRecord(std::move(record_w_span_context));

```

dbarker (commented):
Thanks for digging into this! The bridge processor is a nice driver to get these interfaces aligned with the spec. 

Since this is a major API driving change for both the logger API and the logs SDK. I think it is worth breaking those changes out to a new PR with more analysis and design documentation accounting for currently supported otel-cpp use cases and spec requirements. It is possible that some of the ABIv2 logger interfaces should change to address passing context explicitly through to the processors.  

Once those interfaces are settled and merged the bridge processor can use them. 

Please see feedback below with some of the use cases that we need to support or explicitly break. Sharing links to the relevant specs:

https://opentelemetry.io/docs/specs/otel/logs/api/#emit-a-logrecord
https://opentelemetry.io/docs/specs/otel/logs/sdk/#onemit
https://opentelemetry.io/docs/specs/otel/context/

dbarker (changes requested):
Per comments above I'm requesting that this PR be broken up into:

1. A PR with Logger API and SDK changes and tests for spec compliant context passing through emit to the processors. More design and analysis is needed. 
2. A PR (this existing one is fine) with only the bridge processor and declarative config support

om7057:
Thanks @dbarker, that makes sense. I'll split it: bridge processor and declarative config stay here, and I'll open a new PR for the Logger API and SDK context passing work with the use case matrix you laid out documented against the spec.

Before I start on that PR, I'd like to get alignment on one open tension (left a comment on the ConsumesResolvedContext() thread). ThomsonTan's earlier feedback on this same PR was that resolving context should be opt-in (only when a processor needs it) to avoid a RuntimeContext::GetCurrent() call on every log for pipelines that don't care. Your comment says the opposite: it should be unconditional per spec, and ConsumesResolvedContext() should go away. I want to design the new PR around whichever of these is actually correct rather than build it twice.

sdk/include/opentelemetry/sdk/logs/processor.h
om7057: Understood on removing the gate for spec compliance, want to flag a conflict before I do since it affects the design of the follow up PR.

ThomsonTan's earlier review on this same code asked for the opposite: resolving the context costs a RuntimeContext::GetCurrent() call, and that shouldn't happen on the hot path for pipelines (simple/batch) that never read it, hence the ConsumesResolvedContext() gate.

If the spec requires the resolved context to always reach OnEmit(), the cost has to be paid unconditionally, which reopens that concern. A couple of options for the new PR to consider:
- Pay the cost always, as you're describing. Spec says so, this is likely correct.
- Or resolve lazily: pass something like a thunk or callable that resolves RuntimeContext::GetCurrent() only if a processor actually invokes it, so processors that ignore the parameter never pay for the lookup.

Which direction should the new design PR pursue?

sdk/src/logs/logger.cc
om7057: Thank you for these cases, this is exactly the kind of matrix that was missing. Replied on the other thread about the ConsumesResolvedContext tension with ThomsonTan's feedback, so I will not repeat that here.

On cases 1 to 4, agreed those are the clean spec compliant ones and should just work once the context passing is unconditional.

On cases 5 to 7, I think the honest answer is that a bare SpanContext or TraceId plus SpanId cannot resolve to a live Span, so no processor can get a Span pointer out of them no matter how we wire things. My current bridge implementation falls back to comparing against the ambient span's ids in that case, meaning the event only gets added if the ambient span happens to be the one the ids refer to. That is a real, documented limitation rather than a bug, since the API genuinely gives us nothing to look up a Span by id.

Case 7 looks like it has a second version of the same bug though. CreateLogRecord(span_context) stamps the record's trace and span ids, but nothing carries span_context forward to the later EmitLogRecord(record) call, so that call falls back to resolving the ambient context again, same as case 3. If that is correct then case 7 needs the same fix as the case this PR started from, just for the create then emit pattern instead of the single call pattern.

Also want to flag that case 3's comment cuts off after "Context is resolved to current context OR to", might be missing the rest of that sentence.

I will pull all of this into the new PR's design doc once we're aligned on the ConsumesResolvedContext direction, want to avoid designing it piecemeal across comments here.

sdk/include/opentelemetry/sdk/logs/processor.h
dbarker: It is not obvious, but the logging emit path will always get the current runtime context if an explicit Context was not provided by the caller. The Context is used to enrich the log record with span metadata on creation and for filtering when the `trace_based` config is enabled. 

Tom is right and the processor OnEmit method should not call an extra `context::RuntimeContext::GetCurrent()`  since it either was already called when the log record was created or the user passed in a Context explicitly. 

In both cases the Context exists prior to calling processor OnEmit and that method should receive it unconditionally. 

In the new PR, some refactoring of the ABIv2 path may be needed to allow the one Context to be passed all the way through the emit path to OnEmit of the processor.

om7057: Thanks for clarifying. I had missed that the context is already resolved once upstream, either during CreateLogRecord or from the explicit Context the caller passed in, so OnEmit does not need a second RuntimeContext::GetCurrent() call at all; it just needs the one that already exists threaded through. That actually satisfies the concern of processor always receives the resolved context for spec compliance.
I will take all of this into the new PR; thanks for working through it here.

om7057:
@dbarker Before I start on the new PR, I want to make sure I’m implementing in the right direction.

My understanding is:

* Resolve the context once, either from the explicit `Context` passed to `EmitLogRecord` / `CreateLogRecord`, or from `RuntimeContext::GetCurrent()` when no context is provided.
* Pass that resolved context through to `OnEmit` without doing another `GetCurrent()` downstream.
* Remove `ConsumesResolvedContext()` so processors always receive the context directly.
* Fix the create-then-emit case (case 7), where `CreateLogRecord(context_or_span)` is followed by a separate `EmitLogRecord(record)`, so the same resolved context is preserved through to `OnEmit`.

The last point is the only part I'm not sure about yet: how should the resolved context be carried from `CreateLogRecord` to the later `EmitLogRecord(record)` call?

I see a couple of options:

1. Store the resolved context on the `Recordable` / `LogRecord` itself.
2. Change the single-record `EmitLogRecord` overload to require a `Context`, which would make callers pass it again in the two-step pattern.
3. Use another mechanism you already have in mind.

Could you confirm the overall summary and which direction you'd prefer for the create-then-emit case?

dbarker:
> @dbarker Before I start on the new PR, I want to make sure I’m implementing in the right direction.
> 
> My understanding is:
> 
> * Resolve the context once, either from the explicit `Context` passed to `EmitLogRecord` / `CreateLogRecord`, or from `RuntimeContext::GetCurrent()` when no context is provided.
> * Pass that resolved context through to `OnEmit` without doing another `GetCurrent()` downstream.
> * Remove `ConsumesResolvedContext()` so processors always receive the context directly.
> * Fix the create-then-emit case (case 7), where `CreateLogRecord(context_or_span)` is followed by a separate `EmitLogRecord(record)`, so the same resolved context is preserved through to `OnEmit`.
> 

Yes this is the right direction. And the "case 7" is one to address in the new design. 

> The last point is the only part I'm not sure about yet: how should the resolved context be carried from `CreateLogRecord` to the later `EmitLogRecord(record)` call?
> 
> I see a couple of options:
> 
> 1. Store the resolved context on the `Recordable` / `LogRecord` itself.
> 2. Change the single-record `EmitLogRecord` overload to require a `Context`, which would make callers pass it again in the two-step pattern.
> 3. Use another mechanism you already have in mind.
> 
> Could you confirm the overall summary and which direction you'd prefer for the create-then-emit case?

I think you expressed the challenge and those potential options should be considered. The other option I see is to only support passing the Context to EmitLogRecord and no longer enrich the log record with span metadata in CreateLogRecord. It is not clear to me if this option can be achieved in ABIv1 with all the use cases supported (https://github.com/open-telemetry/opentelemetry-cpp/pull/4309#discussion_r3759064917) though and it needs some analysis to consider what approach will meet the spec, not degrade performance of the emit path, and not break existing use cases unnecessarily. 

Analysis and design options presented in the separate PR will help us all understand the trade-offs. A prototype in the PR can be really helpful to see what the impact is for a given design.

om7057:
@dbarker I've split out the resolved-context design/prototype into a separate PR: #4421

It analyzes the 7 use cases we discussed, removes the `ConsumesResolvedContext()` opt-in, and always passes the resolved context to `LogRecordProcessor::OnEmit`/`OnEmitWithContext` unconditionally, including a documented limitation + fix for the `CreateLogRecord(context)` then plain `EmitLogRecord(record)` case, backed by tests (including exact `RuntimeContext::GetCurrent()` call-count assertions).

I think #4421 should merge first, then I'll rebase this PR on top of it and strip out the now-superseded `ConsumesResolvedContext`/context-plumbing from this PR, updating `EventToSpanEventBridgeProcessor` to use the settled API instead. That way this PR ends up scoped to just the bridge processor and declarative config, and the context-passing change gets reviewed on its own merits without being tied to the bridge use case.

Would appreciate your review of #4421.

om7057:
@dbarker, just a small reminder on this one!

om7057:
@lalitb @dbarker just a ping on this!

sdk/src/logs/event_to_span_event_bridge_processor.cc
```diff
@@ -0,0 +1,272 @@
...
+
+  EventAttributesKeyValueIterable attributes(log_record->GetAttributes());
+  span->AddEvent(event_name, event_timestamp, attributes);
+}
+
+}  // namespace
+
+std::unique_ptr<Recordable> EventToSpanEventBridgeProcessor::MakeRecordable() noexcept
+{
+  return std::unique_ptr<Recordable>(new ReadWriteLogRecord());
```
dbarker: Dynamic allocation of a temporary recordable shows the design challenge from trying to use the current processor interface for these non-exporting use cases like the event-to-span processor.

The event-to-span processor really fits into the category of [advanced processing](https://opentelemetry.io/docs/specs/otel/logs/supplementary-guidelines/#advanced-processing) from the spec. Processors that need to form a pipeline on a single stream for mutating, filtering, and transforming data requires a new processor interface, support from the LoggerContext to replace the root processor, and standardization of SDK spec exporters to use the ReadWriteLogRecord type recordable that has a spec compliant [read-write interface](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord).  

To see the issue with using the current processor interface for these use cases let's consider a reasonable four processor pipeline. The pipeline is intended to apply a custom filter by attributes, sanitize sensitive data, bridge the log events to spans, and then export over otlp http.  The desired outcome is a filtered/sanitized stream of records to the http otlp endpoint. 

If we use the current processor interface the following four processor config fans out instead of forming a pipeline. This means four copies of the log record, four times the recording cost (body, attributes, span context) , and export of unfiltered/unsantizied records.
 
```yaml
logger_provider:
  processors:
      my_logs_filter:         # Log record copy 1 (not exported)
      my_log_record_sanitizer:    # Log record copy 2 (not filtered, is sanitized, not exported)
      event_to_span_bridge:       # Log record copy 3 (not filtered, not sanitized, exported as span events only)
      batch:                      # Log record copy 4 (not filtered, not sanitized, all exported)
          exporter:
              otlp_http: 
``` 

This is driven by the processor interface requiring transfer of record ownership on emit and the built-in multi-processor that implements the fan out behavior.

We need a new root processor that supports a pipeline of composable processors, a way to set that processor as the root in the LoggerContext, and common read-write log record for the built-in SDK exporters that would allow filtering/mutating/transforming in the pipeline.

dbarker (commented):
Thanks for updating this. I had some time to review the changes and spec around these advanced processing use cases. Processors like the bridge processor will require some new architectural components and changes to existing components to be spec compliant and maintain reasonable performance. This is in addition to the Logger API level EmitLogRecordWithContext change, which is needed. 

The architecture components / changes that seem to be required include: 
1. Standardize the SDK built-in exporters (OTLP, ostream) on a recordable with a read-write interface. Pipeline processors need to read data from the record and not duplicate it. 
2. Create a new processor interface for composable pipeline processors that can read and mutate data but do not own the recordable life-cycle. 
3. Create a new root level processor that supports pipeline processors 
4. Update the LoggerContext to support setting a new root processor.  

Please see the example and rationale below. 

To move forward with this PR in the near term please consider reducing scope to just the configuration model and YAML parser for the event-to-span bridge.  We can merge that quickly.

om7057:
> Thanks for updating this. I had some time to review the changes and spec around these advanced processing use cases. Processors like the bridge processor will require some new architectural components and changes to existing components to be spec compliant and maintain reasonable performance. This is in addition to the Logger API level EmitLogRecordWithContext change, which is needed.
> 
> The architecture components / changes that seem to be required include:
> 
> 1. Standardize the SDK built-in exporters (OTLP, ostream) on a recordable with a read-write interface. Pipeline processors need to read data from the record and not duplicate it.
> 2. Create a new processor interface for composable pipeline processors that can read and mutate data but do not own the recordable life-cycle.
> 3. Create a new root level processor that supports pipeline processors
> 4. Update the LoggerContext to support setting a new root processor.
> 
> Please see the example and rationale below.
> 
> To move forward with this PR in the near term please consider reducing scope to just the configuration model and YAML parser for the event-to-span bridge. We can merge that quickly.

Thanks for the detailed explanation; I understood it clearly now. The fan-out example makes the problem obvious: each processor owns and mutates its own copy, so a filter/sanitizer stage upstream of the bridge never actually reaches the exporter's output.

I'll reduce this PR's scope to just the configuration model and YAML parser for the event-to-span bridge (the config schema, parsing, and validation), and drop the processor implementation for now. That gets the declarative-config piece in without depending on the pipeline architecture you outlined.

A couple of questions before I make the change:

- Is there already a tracking issue for the read-write recordable + composable pipeline processor + root processor work you described? If not, would it help to open one so this bridge processor (and future pipeline-shaped processors) has something to reference once that lands?
- For the config-only scope: should the YAML schema still validate/accept the event-to-span-bridge processor block even though there's no processor behind it yet, or would you rather I gate that out until the implementation exists?

Will get the scoped-down version pushed once I hear back.

dbarker:
> A couple of questions before I make the change:
> 
>1. Is there already a tracking issue for the read-write recordable + composable pipeline processor + root processor work you described? If not, would it help to open one so this bridge processor (and future pipeline-shaped processors) has something to reference once that lands?

There is no tracking issue for this work yet. Please log an issue just to track the bridge processor SDK implementation and add a `FIXME-SDK` comment with a link to that issue in the sdk builder where the model would be used. 

Please reference comments from this PR in that issue and mark it blocked pending a design/implementation for the components needed to support these types of non-exporting processors. 

> 1. For the config-only scope: should the YAML schema still validate/accept the event-to-span-bridge processor block even though there's no processor behind it yet, or would you rather I gate that out until the implementation exists?

The parser should create the model element for the bridge processor and the SdkBuilder should detect if it is configured and emit a warning log message so the user knows it is not currently supported.

om7057:
Thanks for the direction. Done:

- Opened https://github.com/open-telemetry/opentelemetry-cpp/issues/4454 to track the SDK implementation, referencing the architecture discussion and use case matrix from this PR, marked blocked pending the read-write recordable / composable pipeline processor / root processor work.
- Added a `FIXME-SDK` comment in `SdkBuilder::CreateEventToSpanEventBridgeLogRecordProcessor` linking to that issue.
- Scoped this PR down to just the configuration model and YAML parser: the parser still creates the `EventToSpanEventBridgeLogRecordProcessorConfiguration` model element, and `SdkBuilder` now logs a warning and skips building a processor rather than throwing or crashing, so a pipeline like `[event_to_span_event_bridge, batch]` still runs `batch` normally.
- Removed the actual `EventToSpanEventBridgeProcessor` implementation and its tests, and reverted the `Logger`/`LogRecordProcessor` resolved-context plumbing (`EmitLogRecordWithContext`, `OnEmitWithContext`) since it existed only to serve that processor. It'll come back once the new pipeline architecture is in place and someone implements the processor against it.

Pushed. Let me know if you'd like anything adjusted.

sdk/include/opentelemetry/sdk/configuration/log_record_processor_configuration_visitor.h
```diff
@@ -27,8 +28,20 @@ class LogRecordProcessorConfigurationVisitor
...
 
-  virtual void VisitBatch(const BatchLogRecordProcessorConfiguration *model)         = 0;
-  virtual void VisitSimple(const SimpleLogRecordProcessorConfiguration *model)       = 0;
+  virtual void VisitBatch(const BatchLogRecordProcessorConfiguration *model)   = 0;
+  virtual void VisitSimple(const SimpleLogRecordProcessorConfiguration *model) = 0;
+
+  /**
+   * Visit an event_to_span_event_bridge/development processor configuration.
+   *
+   * Unlike its siblings this method is not pure virtual: it was added after the visitor was
```
dbarker: Please make `VisitEventToSpanEventBridge` pure virtual. 

It is okay to make breaking changes to these internal configuration classes now. This is an internal class used only by tests and the `SdkBuilder`. The main public facing classes from this directory include the models, builders, parser and ConfiguredSdk class. Configuration classes as a whole have been under the experimental `WITH_CONFIGURATION` cmake option in releases until now.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -2294,6 +2301,24 @@ SdkBuilder::CreateSimpleLogRecordProcessor(
   throw UnsupportedException(die);
 }
 
+std::unique_ptr<opentelemetry::sdk::logs::LogRecordProcessor>
+SdkBuilder::CreateEventToSpanEventBridgeLogRecordProcessor(
+    const opentelemetry::sdk::configuration::EventToSpanEventBridgeLogRecordProcessorConfiguration
+        * /* model */) const
+{
+  // FIXME-SDK: the event_to_span_event_bridge/development processor has no SDK implementation
+  // yet. It requires new pipeline processor architecture (a read-write recordable, a composable
```
dbarker: nit - please keep the comment short here and let the issue hold the implementation details on how this processor may be enabled in the future. Plans and designs will change once discussed. No need to log out the issue number since it is already in the comment and linked to the is PR.

CHANGELOG.md
```diff
@@ -288,6 +296,15 @@ Breaking changes:
   * The public configuration member `max_buckets_` was renamed to `max_size_` to
     match the configuration schema. Please adjust SDK configuration accordingly.
 
+* [CONFIGURATION] `LogRecordProcessorConfigurationVisitor` gained the virtual
```
dbarker: Please remove the breaking changes section per the discussion in https://github.com/open-telemetry/opentelemetry-cpp/pull/4309/changes#r3816289169

CHANGELOG.md
```diff
@@ -15,6 +15,14 @@ Increment the:
 
 ## [Unreleased]
 
+* [CONFIGURATION] Parse the `event_to_span_event_bridge/development` log
+  record processor configuration in declarative configuration. The SDK
+  implementation of this processor is not yet available; `SdkBuilder` logs a
+  warning and skips it when configured. See
```
dbarker: Please keep this note focused just on the changes from the PR and remove the last sentence. The PR links to the issue tracking the SDK implementation and that provides the context for those interested.

dbarker (approved):
Thanks for the discussion and updates. Approved with some minor requested changes below.

om7057:
@dbarker Thanks for approving. 

I've resolved the changes:
- `VisitEventToSpanEventBridge` is now pure virtual (both implementers already overrode it, so this was a clean change).
- Shortened the `FIXME-SDK` comment in `sdk_builder.cc` to just the tracking issue link.

om7057:
@lalitb also could check this once?

om7057:
@lalitb a small reminder on this one!

dbarker:
@lalitb This PR has been descoped to just the YAML parsing and model. The SdkBuilder will log a warning if a user configures the events to spans bridge processor to indicate that the feature is not supported. This is consistent with  how the SDK reports unsupported yaml elements from schema 1.1.0.  Please take a look since your change request is currently blocking PR.

om7057:
@lalitb just a small reminder on this one!

PR #4311, opened 2026-07-26
[BUILD] Create the configuration_core library for programmatic config

dbarker:
Contributes to #4134 and follow up from #https://github.com/open-telemetry/opentelemetry-cpp/pull/3655#pullrequestreview-3277362332

This PR creates a new library (`configuration_core`) to contain the SDK configuration components that do not require external dependencies. This library is built unconditionally with the SDK and available in the `sdk` CMake component. The YAML configuration component (and ryml dependency) remain with the `configuration` library and publicly link to the `configuration_core` library. 

## Changes
- Creates the `opentelemetry-cpp::configuration_core` CMake target for all configuration components that don't require external dependencies
   - The  `opentelemetry-cpp::configuration` CMake target now just includes the  ryml dependent components (ryml document, yaml parser). 
- Creates the configuration_core library for Bazel builds
- Updates the INSTALL guide to cover these two targets

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -242,6 +242,8 @@ build configuration.
 |                            | opentelemetry-cpp::trace                                                                         |
 |                            | opentelemetry-cpp::metrics                                                                       |
 |                            | opentelemetry-cpp::logs                                                                          |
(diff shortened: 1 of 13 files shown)


marcalff (approved):
LGTM, thanks for the fix.

PR #4313, opened 2026-07-27
[SDK] Support MeterConfigurator updates and example

### LINKED ISSUE #4256: MeterProvider support for updating the MeterConfigurator
### Feature

Add an `UpdateMeterConfigurator` method to the SDK `MeterProvider` which recomputes and sets the new `MeterConfig` (the enabled flag) for existing and new meters.

**Background**

The spec allows for updating the MeterConfigurator at runtime. 

https://opentelemetry.io/docs/specs/otel/metrics/sdk/#meterconfigurator

> A MeterConfigurator is a function which computes the [MeterConfig](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#meterconfig) for a [Meter](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#meter).
> ...
> This function is called when a Meter is first created, and for each outstanding Meter when a MeterProvider’s MeterConfigurator is updated (if updating is supported). Therefore, it is important that it returns quickly.

Enabling and disabling meters by instrumentation scope at runtime can be valuable for debugging, testing, and managing signals in production. 

**Requirements:**

 - Updating the `MeterConfigurator` must be thread safe
 - Update must create and apply a new MeterConfig for all existing Meters created by the provider
 - The updated `MeterConfigurator` must apply to new Meters created by the provider
 - `GetMeter` and `UpdateMeterConfigurator` must prevent race conditions on the meters
 - Updating the `MeterConfig` (the enabled flag) of a Meter must not block other methods of the `Meter` class
 - There should be an example of how to use the method :) 
 
See the tracer and logger configurator update PRs for common patterns and examples: 

#4065 Tracer Configurator update support
#4224  Logger Configurator update support

adityabavadekar:
Fixes #4256

Adds the ability to update a `MeterConfigurator` (and the `MeterConfig` of
existing meters) at runtime. Following the pattern used in #4065 and #4224.

This implements the optional requirement to allow updating the
MeterProvider's MeterConfigurator:
https://opentelemetry.io/docs/specs/otel/metrics/sdk/#meterconfigurator

> This function is called when a Meter is first created, and for each
> outstanding Meter when a MeterProvider's MeterConfigurator is updated (if
> updating is supported).

## Changes

- Adds `UpdateMeterConfigurator` to `MeterProvider` to support setting new
  meter configs at runtime. The update holds the same lock as `GetMeter`, so a
  meter is never returned with a config that is out of date with the provider
  configurator.
- Moves the `Meter` enabled flag to an atomic so updating it does not block
  instrument creation or `Collect`.
- Adds `MeterContext::SetMeterConfigurator`.
- Adds tests and an example.

Note: an Instrument created while its Meter is disabled will automatically start recording when the Meter is enabled. It does not need to be recreated. Measurements made while the Meter is disabled are discarded. Both behaviors are mentioned in the example README.


examples/meter_configurator/main.cc
```diff
@@ -0,0 +1,230 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// This example shows how to use MeterProvider::UpdateMeterConfigurator to enable and disable
+// meters by instrumentation scope at runtime. Updating the MeterConfigurator affects all existing
+// and future meters provided by the MeterProvider. It is safe to call concurrently with
+// MeterProvider::GetMeter and with instrument creation and recording on existing meters.
```
dbarker: existing meters -> existing instruments

examples/meter_configurator/main.cc
```diff
@@ -0,0 +1,230 @@
...
+//          investigate. Its metrics are collected and exported again.
+// Stage 4: The investigation completes and the external_library meter is disabled again.
+//
+// IMPORTANT: a Meter that is disabled when an instrument is created returns a no-op instrument,
+// and that instrument stays no-op even if the Meter is enabled later. Create instruments while
+// the Meter is enabled (as the classes below do at startup) and then toggle the meters to start
+// and stop collection. Code that must tolerate being started with a disabled Meter has to create
+// its instruments again after the Meter is enabled.
+
+#include <stdint.h>
```
dbarker: nitpick - use the c++ header `<cstdint>` and std:: namespace for integer types.

examples/meter_configurator/main.cc
```diff
@@ -0,0 +1,230 @@
...
+
+// Creates a MeterProvider with an OStreamMetricExporter and an initial ScopeConfigurator.
+std::shared_ptr<metrics_sdk::MeterProvider> CreateMeterProvider(
+    std::unique_ptr<scope_sdk::ScopeConfigurator<metrics_sdk::MeterConfig>> configurator)
+{
+  auto exporter = metrics_exporters::OStreamMetricExporterFactory::Create();
+
+  // A long export interval is used so that this example only exports when it calls ForceFlush.
+  metrics_sdk::PeriodicExportingMetricReaderOptions options;
+  options.export_interval_millis = std::chrono::milliseconds(60000);
```
dbarker: consider reducing this (maybe 10 s). I think force flush may block for as long as the export interval plus the export.

sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -140,6 +141,20 @@ class Meter final : public opentelemetry::metrics::Meter
...
+
+  /**
+   * Update this meter's MeterConfig. Called only by
+   * MeterProvider::UpdateMeterConfigurator when the provider-level MeterConfigurator is
+   * replaced at runtime.
+   */
+  void UpdateMeterConfig(MeterConfig config) noexcept;
+
+  /** Returns whether this meter is enabled by its current MeterConfig. */
+  bool IsMeterEnabled() const noexcept { return meter_enabled_.load(std::memory_order_relaxed); }
```
dbarker: Consider renaming to IsEnabled()

sdk/include/opentelemetry/sdk/metrics/meter.h
```diff
@@ -151,7 +166,9 @@ class Meter final : public opentelemetry::metrics::Meter
                                               InstrumentEqualNameCaseInsensitive>;
   MetricStorageMap storage_registry_;
   std::shared_ptr<ObservableRegistry> observable_registry_;
-  MeterConfig meter_config_;
+  // MeterConfig state is stored in an atomic variable so that instrument creation and Collect()
+  // never block on a concurrent MeterProvider::UpdateMeterConfigurator.
+  std::atomic<bool> meter_enabled_{true};
```
dbarker: As implemented the dynamic enable/disable has some quirks: 
- If the meter is disabled when the instrument is created then the instrument is a noop for the life of the process regardless if the meter is later enabled. 
- If an instrument is created from an enabled meter and then that meter is disabled, the instrument still records into storage as if it is enabled. The recording cost does not go away.  
 
This is a bit out of line with the spec: 

> If a Meter is disabled, it MUST behave equivalently to [No-op Meter](https://opentelemetry.io/docs/specs/otel/metrics/noop/#meter).
> The value of enabled MUST be used to resolve whether an instrument is [Enabled](https://opentelemetry.io/docs/specs/otel/metrics/api/#enabled). See [Instrument Enabled](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-enabled) for details.


The instrument needs to also have knowledge of the meter's enabled state. This is also a requirement if we support the sync instrument [Enabled API method](https://opentelemetry.io/docs/specs/otel/metrics/api/#enabled). I think we should support this method :). 

To support the ability to enable and disable instruments consider making this a `std::shared_ptr<std::atomic<bool>>`. This could then be passed on to each instrument (`Synchronous` base class), `SyncMetricStorage` and `BoundEntry` to allow each instrument to disable (behave as a noop instrument) and enable based on the `MeterConfig` applied at any time.  

I think it would okay to not implement full sharing of this `meter_enabled` flag and adding the sync instrument `Enabled` API method in this PR, but there should be a follow up to do so.

examples/meter_configurator/main.cc
```diff
@@ -0,0 +1,230 @@
...
+// The example simulates a typical cost-management and debugging workflow:
+//
+// Stage 1: Startup. All meters are enabled and all three scopes report metrics.
+// Stage 2: The noisy external_library metrics are not needed in steady state, so its meter is
+//          disabled. Its measurements are no longer collected or exported.
+// Stage 3: A user reports unexpected behavior, so the external_library meter is re-enabled to
+//          investigate. Its metrics are collected and exported again.
+// Stage 4: The investigation completes and the external_library meter is disabled again.
+//
+// IMPORTANT: a Meter that is disabled when an instrument is created returns a no-op instrument,
```
dbarker: Thanks for the explanation here. I think we should move towards meters/instruments that can be enabled and disabled dynamically independent of the MeterConfig on startup. More in this [comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/4313#discussion_r3670163184) .

dbarker (commented):
Thanks for the feature and tests! Please see some initial comments below.

sdk/test/metrics/meter_provider_sdk_test.cc
```diff
@@ -343,3 +356,265 @@ TEST(MeterProvider, ExplicitShutdownNotWarnOnDestructionCheck)
...
+  auto meter   = provider->GetMeter("scope.toggle");
+  auto counter = meter->CreateUInt64Counter("counter.toggle");
+
+  provider->UpdateMeterConfigurator(DisableAll());
+  counter->Add(1);
+  EXPECT_TRUE(CollectScopeNames(reader).empty());
+
+  provider->UpdateMeterConfigurator(EnableAll());
+  counter->Add(1);
+  EXPECT_EQ(CollectScopeNames(reader), (std::set<std::string>{"scope.toggle"}));
```
ThomsonTan, line 679: This assertion only proves that the scope is present; it does not prove that the collected value is correct. All six new `UpdateMeterConfigurator*` tests use [`CollectScopeNames()`](https://github.com/adityabavadekar/opentelemetry-cpp/blob/19936ab9f73dcf69511772c909d21548005333a1/sdk/test/metrics/meter_provider_sdk_test.cc#L403-L415), which discards the metrics and points:

```cpp
disable -> Add(1)  // should be ignored
enable  -> Add(1)
expected sum: 1
current sum:  2
```

Both sums produce `{"scope.toggle"}`, so this test stays green while the disabled measurement is retained and later exported. Please inspect the counter's `SumPointData` (or add a value-returning collection helper) and assert that the sum is `1` here.

sdk/src/metrics/meter.cc
```diff
@@ -101,26 +102,31 @@ Meter::Meter(
...
+  meter_enabled_.store(config.IsEnabled(), std::memory_order_relaxed);
+}
+
 opentelemetry::nostd::unique_ptr<metrics::Counter<uint64_t>> Meter::CreateUInt64Counter(
     opentelemetry::nostd::string_view name,
     opentelemetry::nostd::string_view description,
     opentelemetry::nostd::string_view unit) noexcept
 {
-  if (!meter_config_.IsEnabled())
+  if (!IsMeterEnabled())
```
ThomsonTan: This branch turns the Meter's current disabled state into a permanent object choice: `kNoopMeter` returns a [`NoopCounter`](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/api/include/opentelemetry/metrics/noop.h#L36-L50) whose `Add()` methods are empty. Re-enabling the Meter cannot replace a counter already held by library code:

```cpp
static auto counter = meter->CreateUInt64Counter("requests");  // Meter is disabled
// The host later enables this Meter.
counter->Add(1);  // Still a NoopCounter.
```

This defeats the runtime-enable-for-debugging use case in [#4256](https://github.com/open-telemetry/opentelemetry-cpp/issues/4256), especially for third-party instrumentation. The example avoids the failure only by [creating every instrument while enabled](https://github.com/adityabavadekar/opentelemetry-cpp/blob/19936ab9f73dcf69511772c909d21548005333a1/examples/meter_configurator/main.cc#L23-L27); an application cannot force a dependency to recreate its instruments.

Please return an instrument whose recording behavior observes the Meter's shared enabled state (or an equivalent revivable proxy), and add a regression test: create while disabled, enable, `Add(1)`, collect, and assert value `1`.

examples/meter_configurator/README.md
```diff
@@ -0,0 +1,249 @@
...
+
+```sh
+~/build/examples/meter_configurator/example_meter_configurator
+```
+
+**Expected output:**
+
+Metrics are exported to stdout via the `OStreamMetricExporter`. The example
+uses a long export interval and calls `ForceFlush` at the end of each
+stage, so each stage exports exactly once.
```
ThomsonTan: The long interval does not make `ForceFlush()` the only export trigger. [`OnInitialized()` starts the worker, and `DoBackgroundWork()` calls `CollectAndExportOnce()` before its first wait](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/src/metrics/export/periodic_exporting_metric_reader.cc#L88-L112). That startup collection races with Stage 1's recording and `ForceFlush()`:

```text
documented: 4 stage-triggered export batches
observed:   5 batches, with the extra batch's placement varying
```

Therefore the checked-in exact transcript is not deterministic. The [CTest target](https://github.com/adityabavadekar/opentelemetry-cpp/blob/19936ab9f73dcf69511772c909d21548005333a1/examples/meter_configurator/CMakeLists.txt#L10-L13) only runs the executable, and Bazel uses a plain [`cc_test`](https://github.com/adityabavadekar/opentelemetry-cpp/blob/19936ab9f73dcf69511772c909d21548005333a1/examples/meter_configurator/BUILD#L6-L20), so neither validates stdout.

sdk/src/metrics/meter_provider.cc
```diff
@@ -130,6 +130,31 @@ void MeterProvider::AddView(std::unique_ptr<InstrumentSelector> instrument_selec
...
+        "ignoring.");
+    return;
+  }
+
+  // Shares the lock with GetMeter so a Meter is never returned while its MeterConfig is out of
+  // date with the provider configurator.
+  const std::lock_guard<std::mutex> guard(lock_);
+  context_->SetMeterConfigurator(std::move(meter_configurator));
+
+  for (auto &meter : context_->GetMeters())
```
ThomsonTan, line 163: for (auto &meter : context_->GetMeters())[`GetMeters()` is explicitly non-thread-safe](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/include/opentelemetry/sdk/metrics/meter_context.h#L97-L103), so this loop is safe only under a cross-class invariant: every production mutation of `meters_` currently reaches `AddMeter()` or `RemoveMeter()` through a `MeterProvider` method that holds `lock_`. That invariant is not stated here, and it is not structurally enforced because [`MeterContext` exposes both mutators publicly](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/include/opentelemetry/sdk/metrics/meter_context.h#L153-L163).

```text
T1: provider lock_ -> GetMeters() -> iterate borrowed span
T2: meter_lock_    -> AddMeter()  -> meters_ may reallocate
```

`lock_` does not exclude a direct/future T2 caller. The analogous [tracer](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/src/trace/tracer_provider.cc#L152-L154) and [logger](https://github.com/open-telemetry/opentelemetry-cpp/blob/7a38ec8874ba90eeda322505eb2d038e743f1d74/sdk/src/logs/logger_provider.cc#L129-L131) loops explicitly document their ownership invariant, and their collections are provider-private.

Please either enforce this invariant by using `ForEachMeter()` or making the mutators provider-only, or at minimum add the equivalent ownership comment above this loop so future callers do not introduce a race.

sdk/test/metrics/meter_provider_sdk_test.cc
```diff
@@ -520,3 +537,401 @@ TEST(MeterProvider, ExplicitShutdownNotWarnOnDestructionCheck)
...
+  // With the final configurator enabling all scopes, every existing meter must be collectable.
+  provider->UpdateMeterConfigurator(EnableAll());
+  auto meter   = provider->GetMeter("scope.final");
+  auto counter = meter->CreateUInt64Counter("counter.final");
+  counter->Add(1);
+  EXPECT_FALSE(CollectScopeNames(reader).empty());
+}
+
+// Guards the lock order between UpdateMeterConfigurator (lock_) and collection (meter_lock_ then
+// lock_, via an observable callback calling GetMeter). Hangs rather than fails on regression.
```
dbarker, line 844: Please remove this test. Calling GetMeter from inside an observable instrument callback is not a supported use case and is causing tsan failures.

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -56,7 +57,8 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
                     ExemplarFilterType exemplar_filter_type,
                     nostd::shared_ptr<ExemplarReservoir> &&exemplar_reservoir,
 #endif
-                    const AggregationConfig *aggregation_config)
+                    const AggregationConfig *aggregation_config,
+                    std::shared_ptr<MeterEnabledState> meter_enabled_state = nullptr)
```
dbarker, line 61: The meter_enabled_state is a required argument and should never be null for a valid `SyncMetricStorage`.

sdk/src/metrics/meter.cc
```diff
@@ -532,23 +466,30 @@ std::unique_ptr<SyncWritableMetricStorage> Meter::RegisterSyncMetricStorage(
         auto storage_iter = storage_registry_.find(view_instr_desc);
         if (storage_iter != storage_registry_.end())
         {
-          WarnOnNameCaseConflict(GetInstrumentationScope(), storage_iter->first, view_instr_desc);
+          if (IsEnabled())
```
dbarker, line 469: Please revert this condition on IsEnabled. These warnings are based on instrument identity and independent of the current meter's enabled state.

sdk/src/metrics/meter.cc
```diff
@@ -605,15 +546,22 @@ std::unique_ptr<AsyncWritableMetricStorage> Meter::RegisterAsyncMetricStorage(
...
+          }
           // static_pointer_cast is okay here. If storage_registry_.find is successful
           // InstrumentEqualNameCaseInsensitive ensures that the
           // instrument type and value type are the same for the existing and new instrument.
           async_storage = std::static_pointer_cast<AsyncMetricStorage>(storage_iter->second);
         }
         else
         {
-          WarnOnDuplicateInstrument(GetInstrumentationScope(), storage_registry_, view_instr_desc);
+          if (IsEnabled())
```
dbarker, line 560: Please revert this condition on IsEnabled

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -285,6 +288,8 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
   nostd::shared_ptr<ExemplarReservoir> exemplar_reservoir_;
 #endif
   TemporalMetricStorage temporal_metric_storage_;
+  // Owning Meter's enabled flag; null (tests/benchmarks) means enabled. Gates Bind().
```
dbarker, line 291: Please remove this comment as the meter_enabled_state_ should never be null. If it is null then the `SyncMetricStorage` is in an invalid state and cannot be used.

sdk/src/metrics/sync_instruments.cc
```diff
@@ -55,6 +57,10 @@ LongCounter::LongCounter(const InstrumentDescriptor &instrument_descriptor,
 void LongCounter::Add(uint64_t value,
                       const opentelemetry::common::KeyValueIterable &attributes) noexcept
 {
+  if (!IsEnabled())
```
dbarker, line 60: Checking `IsEnabled` through the shared `MeterEnabledState` object results in acceptable performance compared with the current noop meter approach. 

Local testing with `sdk/test/metrics/sync_instruments_benchmark.cc` shows the call to Add/Record is within ~1 ns of the current implementation.

dbarker (changes requested):
Thanks for the updates! Please see some feedback and requested changes below.

dbarker:
Hi @adityabavadekar, Just checking in. Are you still interested in contributing this PR?

adityabavadekar:
Yes, I am still interested, and I have fixed the issues you mentioned. Please review.
The changes:

- Reverted the `IsEnabled()` warning conditions, at 4 sites rather than 2 since
  `WarnOnDuplicateInstrument` is also identity-based. All 4 match `main` again.
- Made `meter_enabled_state` required, updating 27 test construction sites, and removed 2 null
  checks it makes unreachable.
- Removed the deadlock test.
- Fixed a bug introduced earlier: `SyncMetricStorage::Bind` returning `nullptr` was wrapped into
  a handle with null storage, so instruments bound while disabled stayed silent after
  re-enabling. It now binds lazily on enable. `BoundCounterObservesMeterEnabledState` masked
  this by using two separate states.
- Merged `main` and fixed the iwyu diagnostics.

sdk/test/metrics/meter_provider_sdk_test.cc
adityabavadekar: I've updated the assertions:

- Added `CollectCounterSum()`, which reads `SumPointData` and returns the sum, or -1 if absent so
  missing data stays distinct from a sum of 0.
- `UpdateMeterConfiguratorReEnable` now asserts `EXPECT_EQ(1, ...)`, so it fails if a disabled
  measurement is retained.
- `UpdateMeterConfiguratorInstrumentCreatedWhileDisabledRecordsAfterEnable` uses the same helper.

sdk/src/metrics/meter_provider.cc
adityabavadekar: - `ForEachMeter` is not usable here: it acquires `meter_lock_`, which the collect path holds
  before taking `lock_`, so the inverted order deadlocks. The loop carries a comment documenting
  the ownership invariant, the minimum option you offered.
- `tracers_` and `loggers_` are provider-private, whereas `meters_` is in `MeterContext` with
  public `AddMeter`/`RemoveMeter`. Should I make those two private with `friend class
  MeterProvider`, matching `Meter::UpdateMeterConfig`? Left alone for now since it alters the
  `MeterContext` public API.

examples/meter_configurator/README.md
adityabavadekar: The `README` notes that batch count and order vary due to the reader's initial collect, so the
transcript is just a sample run.

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
adityabavadekar: Removed the default, so it is now a required argument. That meant updating 27 construction sites across 6 test files.

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
adityabavadekar: Removed. The comment now just notes the flag is shared so configurator updates reach the storage. I also dropped the null checks in `Bind` and `Synchronous::IsEnabled` that the required argument makes unreachable.

sdk/src/metrics/meter.cc
adityabavadekar: Reverted. I did the same at the two `WarnOnDuplicateInstrument` sites, since those warnings are identity-based too, so all four now match `main`.

sdk/include/opentelemetry/sdk/metrics/meter.h
adityabavadekar: Instruments now share the Meter's `MeterEnabledState`, so one created while the meter is disabled records once the meter is enabled instead of staying a no-op. The sync instrument `Enabled` API method is still outstanding and I will open a follow-up issue for it.

sdk/src/metrics/meter.cc
adityabavadekar: Fixed. `Create*` no longer returns a no-op instrument for a disabled meter. Instruments share the Meter's `MeterEnabledState` and check it per call, so one held by library code starts recording when the meter is enabled. `UpdateMeterConfiguratorInstrumentCreatedWhileDisabledRecordsAfterEnable` covers exactly that case.

sdk/src/metrics/meter.cc
```diff
@@ -93,37 +94,35 @@ namespace metrics
...
+      observable_registry_(new ObservableRegistry())
 {
   if (auto meter_context_locked_ptr = meter_context_.lock())
   {
-    meter_config_ = meter_context_locked_ptr->GetMeterConfigurator().ComputeConfig(*scope_);
+    UpdateMeterConfig(meter_context_locked_ptr->GetMeterConfigurator().ComputeConfig(*scope_));
   }
   else
   {
+    UpdateMeterConfig(MeterConfig::Default());
```
dbarker, line 110: Should this be a disabled config if the context is invalid?

sdk/src/metrics/state/sync_metric_storage.cc
```diff
@@ -185,6 +186,13 @@ bool SyncMetricStorage::Collect(CollectorHandle *collector,
 std::shared_ptr<BoundSyncWritableMetricStorage> SyncMetricStorage::Bind(
     const opentelemetry::common::KeyValueIterable &attributes) noexcept
 {
+  // Bound entries are only GC'd during Collect(), which a disabled Meter skips, so binding
```
dbarker, line 189: Please see the comment on Meter::Collect. I think this bound entry cleanup concern can be addressed in a simpler way by calling the Collect method for sync storage independent of meter enabled state.

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -56,7 +57,8 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
                     ExemplarFilterType exemplar_filter_type,
                     nostd::shared_ptr<ExemplarReservoir> &&exemplar_reservoir,
 #endif
-                    const AggregationConfig *aggregation_config)
+                    const AggregationConfig *aggregation_config,
+                    std::shared_ptr<MeterEnabledState> meter_enabled_state)
```
dbarker, line 61: Since an instrument may have many metric streams, ideally the streams (storage) should not need to track the enabled state. Instead, the instrument should perform the check before recording to the stream. This would remove the need for changes to the sync storage classes.

sdk/src/metrics/meter.cc
```diff
@@ -641,7 +575,7 @@ std::unique_ptr<AsyncWritableMetricStorage> Meter::RegisterAsyncMetricStorage(
 std::vector<MetricData> Meter::Collect(CollectorHandle *collector,
                                        opentelemetry::common::SystemTimestamp collect_ts) noexcept
 {
-  if (!meter_config_.IsEnabled())
+  if (!IsEnabled())
```
dbarker, line 578: A meter may be disabled after sync instruments have recorded data and before the data has been collected/exported. In this case the data should still be collected and exported even though the meter is currently disabled. 

The current implementation will always skip collection for async and sync instruments. Please update this to support collecting/exporting data from sync instruments even when the meter is disabled. 

If we run Collect on the sync instruments unconditionally then the BoundEntry clean up concern also seems to be addressed without requiring more complexity of adding the meter enabled state to the storage classes.

dbarker (commented):
Thanks for the updates. I've shared some more feedback below. Can we remove the meter enabled state from the storage classes?

PR #4314, opened 2026-07-27
[Metrics SDK] Fix histogram views rejected when only aggregation_cardinality_limit is set

om7057:
## Summary

Follow-up to #4188, addressing review feedback left after that PR merged
(https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#pullrequestreview and inline comments on `sdk_builder.cc` / `metric_reader.h`).

When a view's stream configuration set `aggregation_cardinality_limit` without an explicit `aggregation` block, `SdkBuilder::AddView()` always constructed a plain `AggregationConfig`. For instruments whose default aggregation is `kHistogram` or `kBase2ExponentialHistogram`, this doesn't match the type `ViewRegistry::AddView()` requires, so the whole view was silently rejected (`AggregationType and AggregationConfig type mismatch. Ignoring AddView call.`) and the configured cardinality limit never took effect.

`AddView()` now resolves the instrument's effective aggregation type (falling back to the instrument-derived default the same way `ViewRegistry::AddView()` does) and constructs the matching `AggregationConfig` subclass (`HistogramAggregationConfig` / `Base2ExponentialHistogramAggregationConfig` / plain `AggregationConfig`) before applying the limit.

- `sdk/src/configuration/sdk_builder.cc`: fix `AddView()` to pick the correct `AggregationConfig` subtype.
- `sdk/test/configuration/sdk_builder_test.cc`: adds `SdkBuilder.AddViewHistogramCardinalityLimitOnly` (reproduces and verifies the fix; fails against the pre-fix code with the exact "type mismatch" rejection) and `SdkBuilder.AddViewCounterCardinalityLimitOnly` (guards the still-correct non-histogram path).

Reader-level cardinality limit enforcement remains a separate follow-up, as noted in the existing `metric_reader.h` TODO — this PR only addresses the histogram view-rejection bug.

## Test plan

- [x] New unit tests `SdkBuilder.AddViewHistogramCardinalityLimitOnly` and `SdkBuilder.AddViewCounterCardinalityLimitOnly` pass.
- [x] Confirmed `AddViewHistogramCardinalityLimitOnly` fails against the pre-fix code with the reported `ViewRegistry::AddView` rejection.
- [x] Ran `yaml_metrics_test`, `configured_sdk_test`, `view_registry_test`, and the full `sdk_builder_test` suite — all green.

### DIFF
--- a/sdk/src/configuration/sdk_builder.cc
+++ b/sdk/src/configuration/sdk_builder.cc
@@ -146,6 +146,7 @@
 #include "opentelemetry/sdk/logs/processor.h"
 #include "opentelemetry/sdk/logs/simple_log_record_processor_factory.h"
 #include "opentelemetry/sdk/metrics/aggregation/aggregation_config.h"
+#include "opentelemetry/sdk/metrics/aggregation/default_aggregation.h"
 #include "opentelemetry/sdk/metrics/cardinality_limits.h"
 #include "opentelemetry/sdk/metrics/exemplar/filter_type.h"
 #include "opentelemetry/sdk/metrics/export/metric_producer.h"
@@ -1745,8 +1746,39 @@ void SdkBuilder::AddView(
     }
     else
     {
-      sdk_aggregation_config = std::make_shared<opentelemetry::sdk::metrics::AggregationConfig>(
-          stream->aggregation_cardinality_limit);
+      // No explicit `aggregation` block was configured, so the view falls back to the
+      // instrument's default aggregation. ViewRegistry::AddView() rejects a view whose
+      // AggregationConfig type does not match the (possibly instrument-derived) aggregation
+      // type, so the config created here must match that same default rather than always
+      // being a plain AggregationConfig (which only satisfies kSum/kLastValue/kDrop).
+      auto effective_aggregation_type = sdk_aggregation_type;
+      if (effective_aggregation_type == opentelemetry::sdk::metrics::AggregationType::kDefault)
+      {
+        bool is_monotonic{false};
+        effective_aggregation_type =
+            opentelemetry::sdk::metrics::DefaultAggregation::GetDefaultAggregationType(
+                sdk_instrument_type, is_monotonic);
+      }
+
+      switch (effective_aggregation_type)
+      {
+        case opentelemetry::sdk::metrics::AggregationType::kHistogram:
+          sdk_aggregation_config =
+              std::make_shared<opentelemetry::sdk::metrics::HistogramAggregationConfig>(
+                  stream->aggregation_cardinality_limit);
+          break;
+
+        case opentelemetry::sdk::metrics::AggregationType::kBase2ExponentialHistogram:
+          sdk_aggregation_config = std::make_shared<
+              opentelemetry::sdk::metrics::Base2ExponentialHistogramAggregationConfig>(
+              stream->aggregation_cardinality_limit);
+          break;
+
+        default:
+          sdk_aggregation_config = std::make_shared<opentelemetry::sdk::metrics::AggregationConfig>(
+              stream->aggregation_cardinality_limit);
+          break;
+      }
     }
   }
 
--- a/sdk/src/configuration/sdk_builder.cc
+++ b/sdk/src/configuration/sdk_builder.cc
@@ -1768,12 +1768,6 @@ void SdkBuilder::AddView(
                   stream->aggregation_cardinality_limit);
           break;
 
-        case opentelemetry::sdk::metrics::AggregationType::kBase2ExponentialHistogram:
-          sdk_aggregation_config = std::make_shared<
-              opentelemetry::sdk::metrics::Base2ExponentialHistogramAggregationConfig>(
-              stream->aggregation_cardinality_limit);
-          break;
-
         default:
           sdk_aggregation_config = std::make_shared<opentelemetry::sdk::metrics::AggregationConfig>(
               stream->aggregation_cardinality_limit);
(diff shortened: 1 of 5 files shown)


om7057:
cc @lalitb @marcalff follow-up to #4188 as requested [here](https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#issuecomment-...). Fixes the histogram-view rejection when only aggregation_cardinality_limit is set, plus a regression test that reproduces the exact rejection reported. Reader-level cardinality limit enforcement (the other open TODO) is intentionally out of scope here; happy to take that on in a separate PR if you'd like it prioritized before the next release.

lalitb (approved):
Thanks for fixing. LGTM, you would have to fix the CI build before it can go through.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -1745,8 +1746,33 @@ void SdkBuilder::AddView(
...
+        effective_aggregation_type =
+            opentelemetry::sdk::metrics::DefaultAggregation::GetDefaultAggregationType(
+                sdk_instrument_type, is_monotonic);
+      }
+
+      switch (effective_aggregation_type)
+      {
+        case opentelemetry::sdk::metrics::AggregationType::kHistogram:
+          sdk_aggregation_config =
+              std::make_shared<opentelemetry::sdk::metrics::HistogramAggregationConfig>(
```
ThomsonTan, line 1770: This fixes the view registration, but it appears to silently collapse the histogram to a single bucket. `HistogramAggregationConfig` default-constructs `boundaries_` as an empty vector, while the histogram constructor chooses between configured and default boundaries based only on whether the config pointer is non-null:

```cpp
if (ac) { point_data_.boundaries_ = ac->boundaries_; }  // now taken with an empty vector
else    { point_data_.boundaries_ = {0.0, 5.0, ... 10000.0}; }
point_data_.counts_ = std::vector<uint64_t>(point_data_.boundaries_.size() + 1, 0);
```

In other words, "no config" previously selected the SDK's 15 default boundaries, whereas the synthesized config now means "explicitly use zero boundaries," producing one bucket. Could we preserve the default boundaries when creating this config and add a test that asserts the resulting histogram still has 15 boundaries and 16 buckets?

sdk/test/configuration/sdk_builder_test.cc
```diff
@@ -260,3 +269,93 @@ TEST(SdkBuilder, CreatePeriodicMetricReader)
...
+      instrument_descriptor, *instrumentation_scope, [&](const metrics_sdk::View &view) {
+        matched++;
+        // The view must not be rejected: it should carry a
+        // HistogramAggregationConfig (not a plain AggregationConfig), since
+        // the instrument's default aggregation for kHistogram is kHistogram.
+        auto *aggregation_config = view.GetAggregationConfig();
+        EXPECT_NE(aggregation_config, nullptr);
+        if (aggregation_config)
+        {
+          EXPECT_EQ(aggregation_config->GetType(), metrics_sdk::AggregationType::kHistogram);
```
ThomsonTan, line 326: The new regression test proves only that the view registers with a `HistogramAggregationConfig` carrying the limit `42`; it never constructs the aggregation or inspects `boundaries_`. Running the test's exact assertions against the object `AddView()` builds shows all three pass while the histogram is already degenerate:

```
EXPECT_NE(aggregation_config, nullptr)   -> PASS
EXPECT_EQ(GetType(), kHistogram)         -> PASS
EXPECT_EQ(cardinality_limit_, 42u)       -> PASS
[not asserted] boundaries_.size()        =  0
[not asserted] bucket count (size + 1)   =  1
```

So the suite stays green (and codecov reports the changed lines fully covered) with zero boundaries and a single bucket. Please make this a behavioral assertion: call `DefaultAggregation::CreateAggregation(...)` with the returned config and verify the resulting `HistogramPointData` keeps the SDK defaults (15 boundaries, 16 bucket counts) alongside the cardinality-limit check, so the test pins what users actually receive rather than the shape of the code.

sdk/src/configuration/sdk_builder.cc
lalitb: Good catch!

sdk/test/configuration/sdk_builder_test.cc
om7057: Yes, understood. The test was checking the shape of the code, not what a user actually gets. Updated AddViewHistogramCardinalityLimitOnly to call DefaultAggregation::CreateAggregation() with the config the view ends up holding, then assert on the resulting HistogramPointData: 15 boundaries, 16 buckets. I confirmed this assertion actually catches the regression; reverting the boundaries fix locally makes it fail with boundaries_.size() == 0 / counts_.size() == 1, which is exactly the degenerate case you described.

sdk/src/configuration/sdk_builder.cc
om7057: @ThomsonTan Thank you for pointing it out. HistogramAggregationConfig default-constructs with an empty boundaries_, and the histogram aggregation constructors only fall back to the SDK defaults when the config pointer itself is nullptr, not when it's non-null-but-empty. Since this fallback path synthesizes the config rather than getting it from an explicit aggregation block, it was accidentally saying "use zero boundaries" instead of "use the defaults."

Fixed in 34846672, added HistogramAggregationConfig::DefaultBoundaries() as a single source of truth for the SDK's 15-boundary default list, and pointed the synthesized config at it. While I was in there, I also deduped the two other spots (LongHistogramAggregation/DoubleHistogramAggregation) that had this same literal hardcoded, so there will be no chance of the three copies drifting apart in the future.

om7057:
I've resolved the comments @ThomsonTan please go through them once.

om7057:
@lalitb @ThomsonTan Is this follow-up PR ready for merge?

ThomsonTan (approved):

PR #4320, opened 2026-07-31
fix(http): reject out-of-range listening ports

### LINKED ISSUE #4305: [BUG] SocketAddr integer constructor silently wraps an out-of-range port
Part of #4287. Split out of the #4292 review so it does not widen that PR.

`SocketAddr(u_long addr, int port)` narrows the port without checking its range:

https://github.com/open-telemetry/opentelemetry-cpp/blob/main/ext/include/opentelemetry/ext/http/server/socket_tools.h

```cpp
inet4.sin_port = htons(static_cast<uint16_t>(port));
```

So `port = -1` becomes 65535, `65536` becomes 0, `99999` becomes 34463. `HttpServer::addListeningPort(int port)` passes straight through to this overload, so an out-of-range port silently binds a different one.

#4292 gave the string constructor a contract where a malformed port produces an invalid `SocketAddr` (`AF_UNSPEC`, `port() == -1`). The two constructors are now inconsistent: one rejects a bad port, the other wraps it.

This is deliberately not folded into #4292 because the integer overload has a real caller and changing its failure behaviour means deciding how `addListeningPort` should report a bad port (return value, clamp, or leave the socket unbound), which is a separate design question.

### Suggested direction

Validate `port < 0 || port > 65535` in the integer overload, leave `AF_UNSPEC` on failure to match the string constructor, and have `addListeningPort` surface the failure rather than binding a wrapped port. Add tests for the boundary values.

noob339:
Fixes #4305

## Changes

* Added range validation to `SocketAddr(u_long addr, int port)` so ports below `0` or above `65535` are rejected instead of wrapping during the `uint16_t` conversion.
* Updated `HttpServer::addListeningPort(int port)` to return `-1` for an invalid port before creating or binding a socket.
* Added a unit test covering invalid values including `-1`, `65536`, and `99999`.

Tested with:
```text
bazel test //ext/test/http:curl_http_test \
  --test_filter=HttpServerTest.RejectsOutOfRangeListeningPorts
```

### DIFF
--- a/ext/include/opentelemetry/ext/http/server/socket_tools.h
+++ b/ext/include/opentelemetry/ext/http/server/socket_tools.h
@@ -174,12 +174,8 @@ struct SocketAddr
   /// <returns>SocketAddr</returns>
   SocketAddr() {}
 
-  SocketAddr(u_long addr, int port)
+  SocketAddr(u_long addr, u_int16_t port)
   {
-    if (port < 0 || port > 65535)
-    {
-      return;
-    }
 
     sockaddr_in &inet4    = reinterpret_cast<sockaddr_in &>(m_data);
     inet4.sin_family      = AF_INET;
--- a/ext/include/opentelemetry/ext/http/server/socket_tools.h
+++ b/ext/include/opentelemetry/ext/http/server/socket_tools.h
@@ -176,6 +176,11 @@ struct SocketAddr
 
   SocketAddr(u_long addr, int port)
   {
+    if (port < 0 || port > 65535)
+    {
+      return;
+    }
+
     sockaddr_in &inet4    = reinterpret_cast<sockaddr_in &>(m_data);
     inet4.sin_family      = AF_INET;
     inet4.sin_port        = htons(static_cast<uint16_t>(port));
--- a/ext/include/opentelemetry/ext/http/server/http_server.h
+++ b/ext/include/opentelemetry/ext/http/server/http_server.h
@@ -199,7 +199,7 @@ public:
 
   void setServerName(std::string const &name) { m_serverHost = name; }
 
-  int addListeningPort(int port)
+  int addListeningPort(uint16_t port)
   {
 
 
--- a/ext/include/opentelemetry/ext/http/server/http_server.h
+++ b/ext/include/opentelemetry/ext/http/server/http_server.h
@@ -201,11 +201,18 @@ public:
 
   int addListeningPort(int port)
   {
+
+
+    SocketTools::SocketAddr addr(0, port);
+    if (addr.port() == -1)
+    {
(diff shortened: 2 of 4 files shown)


thc1006:
Thanks for picking this up, and the fix reads right to me. Two things that are easier to sort out now than after review.

The description says `Fixes #4287`, which is the audit tracker holding a dozen or so separate defects. Merging with that keyword would close the whole tracker. `Fixes #4305` is the one this resolves.

#4292 is approved and touches the same part of `socket_tools.h`, so this will want a rebase once it lands. It gives the string constructor the same contract you are using here, an invalid address left as `AF_UNSPEC` with `port()` returning `-1`, so the two constructors end up consistent. It also adds `ext/test/http/socket_tools_test.cc`, which may be a more natural home for the new test than `curl_http_test.cc`.

I can answer anything about the issue if that helps.

noob339:
Thanks for catching that. I updated the description to use Fixes #4305 instead of closing the audit tracker.

I’ll rebase once #4292 lands and resolve any overlap in socket_tools.h. I’ll also move the port validation coverage into ext/test/http/socket_tools_test.cc after the rebase, since that sounds like the better location for the constructor-level tests.

noob339:
Thanks again for the guidance. I rebased onto the merged #4292 changes, moved the integer `SocketAddr` coverage into `ext/test/http/socket_tools_test.cc`, and updated the PR to reference `Fixes #4305`.

The focused Bazel test passes locally. Let me know if any other suggestions arise please!

dbarker (approved):
Thanks for the pr. This impacts internal http server used in testing and the http example. Not for production use.

ext/include/opentelemetry/ext/http/server/socket_tools.h
```diff
@@ -176,6 +176,11 @@ struct SocketAddr
 
   SocketAddr(u_long addr, int port)
   {
+    if (port < 0 || port > 65535)
+    {
+      return;
+    }
+
```
marcalff: Instead of doing all this, how about changing port to `uint16_t` ?

marcalff (changes requested):
Thanks for the contribution.

Please see comments for a simpler solution.

thc1006:
Sure, I'll do it.

Async is probably easiest if your assignment allows it: send the questions here, or wherever suits you, and I'll write back properly. If it has to be a live conversation, that works too. Tell me your time zone and roughly when, and we'll find a slot.

Good luck with the course. And thanks for #4320, by the way. A port that only fits in a wider type is exactly the sort of thing that sits unnoticed for years.

marcalff:
This PR was abandoned.

Issue fixed by #4500 instead.

Closing.

PR #4321, opened 2026-07-31
[METRICS] Add bound UpDownCounter and Gauge preview support

lalitb:
## Changes

Adds bound UpDownCounter and Gauge support to the metrics API, SDK, and no-op implementations.

Bound and unbound writes now share an aggregation, so mixing them on the same Gauge keeps the last recorded value even when both writes have identical timestamps.

Includes regression tests, benchmarks, and an ABI v2 preview CI job.

Requires ABI v2 and `OTELCPP_WITH_METRICS_BOUND_INSTRUMENTS_PREVIEW=ON`. Preview-disabled behavior is unchanged. Context and exemplar support remain follow-ups.

## Benchmarks

Five-run medians on an Apple M4 Pro, Release build, ABI v2 with bound preview enabled and exemplars disabled:

| Fixed-attribute recording | Unbound | Bound |
|---|---:|---:|
| UpDownCounter | 279.1 ns | 13.1 ns |
| Gauge | 285.6 ns | 35.3 ns |

Binding helps most with repeated recordings. Recording and collecting 1,000 Gauge series took 1.59 ms unbound versus 1.11 ms bound, about 30% less time.

Contention still slows shared-series writes. Mixed benchmark iterations contain two recordings. Some contended cases varied by about 25% between runs. These results compare bound and unbound calls on the current code, not before and after the PR.

With the preview enabled, ordinary recordings also use the shared-entry storage. Collection holds the map lock while rotating entries. The benchmarks above do not measure the cost of enabling the preview.

## Validation

All 40 benchmark cases completed. All 42 bound tests passed. Across the six targeted test binaries, 92/95 tests passed. Three existing Gauge tests failed because timestamps were equal where the assertions require a strictly later time. `git diff --check` passed.

- [x] `CHANGELOG.md` updated
- [x] Unit tests added
- [x] Public API changes reviewed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,36 @@ jobs:
       run: |
         ./ci/do_ci.sh cmake.api_only.test
 
+  cmake_bound_metrics_preview_test:
+    name: CMake bound metrics preview test
+    runs-on: ubuntu-24.04
+    env:
+      CXX_STANDARD: '17'
+    steps:
+    - name: Harden the runner (Audit all outbound calls)
+      uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      with:
+        egress-policy: audit
+
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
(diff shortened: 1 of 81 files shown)


lalitb:
This is ready for review now. thanks.

PR #4324, opened 2026-07-31
[SDK] Downscale base2 exponential histogram buckets in place

### LINKED ISSUE #3366: Base2 Exponential Histogram Aggregation Performance
Follow up from PR #3346 review [comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/3346#discussion_r2043260347)

1. Base2 Exponential Histogram Aggregation downscaling in-place. 
2. Aggregation benchmarks for throughput when multiple threads are involved.

Something along the lines of what’s done in [measurements_benchmark.cc](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/test/metrics/measurements_benchmark.cc)

ThomsonTan:
## Change Summary

Base2 exponential histogram downscaling now reuses its existing circular buffer instead of allocating and copying a replacement each time.

The implementation safely folds wrapped buckets in place and uses a fused `AdaptingIntegerArray::Fold()` operation to reduce per-bucket dispatch overhead. It preserves automatic integer-width promotion and removes the existing in-place downscaling TODO.

### Performance

MSVC Release wall time per downscale:

| Window | Before | After | Improvement |
|-------:|-------:|------:|------------:|
| 20 | 780 ns | 441 ns | 43.5% |
| 160 | 5,218 ns | 3,495 ns | 33.0% |
| 640 | 20,153 ns | 13,940 ns | 30.8% |

### Validation

- Compared against the previous implementation across 2,000 randomized buffers.
- Covered wrapped, across-zero, full-buffer, widening, boundary-index, and large-shift cases.
- Mutation tests confirmed that both traversal directions are required.
- Aggregation tests independently verified bucket placement after repeated downscaling.

Fixes #3366

## Changes


sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -89,27 +89,9 @@ void DiffBuckets(const AdaptingCircularBufferCounter &left,
...
-    const uint64_t count = buckets->Get(i);
-    if (count > 0)
-    {
-      new_buckets->Increment(i >> by, count);
-    }
-  }
-  buckets = std::move(new_buckets);
+  // Downscaling is done in place, which preserves the existing bucket size
+  // optimisation (integer width) and avoids allocating a replacement buffer.
+  buckets->Downscale(by);
```
dbarker, line 94: One thing we are currently missing is to enforce that scale never drops below -10 at all times. Enforcing the "reasonable minimum" is a spec requirement and something other SDK's do. Consider adding it this PR or a follow up.  

> The implementation MUST maintain reasonable minimum and maximum scale parameters that the automatic scale parameter will not exceed.

ThomsonTan: Agreed that the runtime-bound gap is real, but I would keep it in a follow-up because it is aggregation policy, not part of this PR's buffer-folding mechanism. One wording nuance: the [SDK specification](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#base2-exponential-bucket-histogram-aggregation) requires a "reasonable minimum", but does not prescribe `-10`; `-10` is the configured `MaxScale` minimum chosen by this implementation.

Filed issue https://github.com/open-telemetry/opentelemetry-cpp/issues/4325 for tracking.

sdk/test/metrics/circular_buffer_counter_test.cc
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <gtest/gtest.h>
+#include <stddef.h>
```
dbarker: nit - prefer the c++ header `<cstdint>` and types in std:: namespace.

dbarker (approved):
Looks good. Thanks! Minor non-blocking nitpick.

lalitb (approved):
Thanks.

sdk/test/metrics/circular_buffer_counter_test.cc
ThomsonTan: Do you mean `#include <cstddef>` as `<cstdint>` is already in the include list? Will switch to C++ header and std:: namespace.

dbarker: Thanks, yes. Not blocking. Generally good to use the c++ headers in new code as we go. IWYU often recommends the old headers so there may be a config change needed for that tool.

PR #4334, opened 2026-08-02
[CONFIGURATION] Add the probability sampler to file configuration

ayush-that:
Follow-up to #4135, requested by @dbarker in https://github.com/open-telemetry/opentelemetry-cpp/pull/4135#discussion_r3590665898.

ref: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

## Changes

Adds the probability sampler to file configuration under the schema key `probability/development`: new `ProbabilitySamplerConfiguration`, `VisitProbability`, `ParseProbabilitySamplerConfiguration`, and `SdkBuilder::CreateProbabilitySampler`. Ratio is optional and defaults to 1.0 per the schema. The parser does not range check the value, same as the other samplers, so validation stays in `ProbabilitySampler`, which warns and uses the spec default.

`VisitComposableProbability` now builds the composite sampler with `ComposableProbabilitySampler` instead of the `TraceIdRatioBasedSampler` placeholder.


sdk/src/configuration/configuration_parser.cc
```diff
@@ -1756,6 +1757,18 @@ ConfigurationParser::ParseParentBasedSamplerConfiguration(const std::unique_ptr<
...
 // NOLINTEND(misc-no-recursion)
 
+std::unique_ptr<ProbabilitySamplerConfiguration>
+ConfigurationParser::ParseProbabilitySamplerConfiguration(const std::unique_ptr<DocumentNode> &node,
+                                                          size_t /* depth */) const
+{
+  using Config = ProbabilitySamplerConfiguration;
+  auto model   = std::make_unique<ProbabilitySamplerConfiguration>();
+
+  model->ratio = node->GetDouble("ratio", Config::kDefaultRatio);
```
dbarker, line 1767: Please add range checking here. The range should be included in the config object (0-1) and throw an exception if the value is outside the range. See the Base2 histogram config for an example.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -491,7 +501,8 @@ class SamplerBuilder : public opentelemetry::sdk::configuration::SamplerConfigur
       const opentelemetry::sdk::configuration::ComposableProbabilitySamplerConfiguration *model)
       override
   {
-    sampler = opentelemetry::sdk::trace::TraceIdRatioBasedSamplerFactory::Create(model->ratio);
+    sampler = opentelemetry::sdk::trace::CompositeSamplerFactory::Create(
```
dbarker: All other composable sampler visit methods are also incorrect and need to be updated/tested. Consider leaving these alone in this PR and following up in a new PR focusing on fixing/testing composable sampler configuration.

sdk/test/configuration/yaml_trace_test.cc
```diff
@@ -927,6 +935,57 @@ file_format: "1.0-trace"
   ASSERT_EQ(ratio->ratio, 3.14);
 }
 
+TEST(YamlTrace, default_probability_sampler)
```
dbarker, line 938: once ratio value range checking is added to the parser, please add a yaml test for out-of-range values.

dbarker (commented):
Thanks for the PR! Please see initial feedback below.

marcalff (commented):
Please see previous comments from @dbarker.

LGTM otherwise.

sdk/src/configuration/sdk_builder.cc
marcalff: We need to implement composable samplers in the SDK first, then plumbing will follow.

To fix in a different PR.

sdk/src/configuration/configuration_parser.cc
ayush-that: I have added.

sdk/src/configuration/sdk_builder.cc
ayush-that: I'm reverting from this PR.
Will open a follow up.

sdk/test/configuration/yaml_trace_test.cc
ayush-that: Added small, large, and NaN tests.

dbarker (approved):
Looks good. Thank you!

marcalff (approved):
LGTM.

PR #4335, opened 2026-08-02
[CONFIGURATION]  Decouple config registry and builder headers

dbarker:
Contributes to: #4134

The Registry class manages the builders used to create SDK components for each configuration struct. This class header currently includes all the builders (which include all the sdk components to be built).  

The transitive includes from the Registry can be reduced by forward declaring the sdk builders in the registry and forward declaring the registry in each exporter builder.  

This change helps move in the direction where the builders can link to just the SDK components they build and not require linking all SDK libraries. 

## Changes

- Forward declare the builders in registry.h 
- Move the Set methods in Registry to the registry.cc file to enable forward declaring the registry. 
- Forward declare the registry and privately link configuration_core with builders outside the ./sdk/configuration 


exporters/otlp/CMakeLists.txt
```diff
@@ -61,8 +61,9 @@ target_include_directories(
   PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>"
          "$<INSTALL_INTERFACE:include>")
 
-target_link_libraries(opentelemetry_exporter_otlp_builder_utils
-                      PUBLIC opentelemetry_otlp_recordable)
+target_link_libraries(
+  opentelemetry_exporter_otlp_builder_utils
+  PUBLIC opentelemetry_otlp_recordable opentelemetry_configuration_core)
```
marcalff, line 66: ostream uses `PRIVATE opentelemetry_configuration_core`, should this be private also ?

marcalff (approved):
LGTM, see minor comment in CMake.

exporters/otlp/CMakeLists.txt
dbarker: no that had to be a public link for now. The otlp builders need the transitive config core  dependency from the utils target. I plan to follow up to break up the config core target next so we better limit the transitive SDK dependencies to the builders.

Issue #3075, opened 2024-10-01
LoggerProvider::ForceFlush() doesn't return false on failed log exports


NickPoole:
**Describe your environment**
Using OtlpGrpcLogRecordExporter or OtlpHttpLogRecordExporter log export failures are ignored at the LogRecordProcessor layer.
SimpleLogRecordProcessor has a comment that failures on LogRecordExporter::Export() should be surfaced, but BatchLogRecordProcessor just blindly calls LogRecordExporter::Export().
There's no programmatic way to ensure that logs are reaching their target destination and not being dropped from the LogRecordProcessor or LoggerProvider.

**Steps to reproduce**
Modify the https://github.com/open-telemetry/opentelemetry-cpp/blob/main/examples/otlp/http_log_main.cc example to call and report on the result of ForceFlush, then run the example against an address that's not an OTLP endpoint. ForceFlush() will report true.

**What is the expected behavior?**
I was expecting ForceFlush() to report false if the LogRecordExporter::Export() call had failed.

**What is the actual behavior?**
ForceFlush() returns true regardless of the logs failing to export.

**Additional context**
Turning up the global logging level, it is possible to observe error events, but that's not the same as being able to check for successful logging.

owent:
According to <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md#forceflush-1> we can use callback or a event to notify users the result of exporting. It may changes ABI, should we discuss how to implement it? @open-telemetry/cpp-approvers @open-telemetry/cpp-maintainers

PR #4337, opened 2026-08-02
[BUG] Stop the Elasticsearch async ForceFlush reporting success without waiting

### LINKED ISSUE #4336: [BUG] Elasticsearch async ForceFlush reports success on timeout and ignores the caller deadline
### Describe your environment

`main` at `11fa0db0`, Linux, GCC 14.2, configured with `-DWITH_ELASTICSEARCH=ON -DWITH_ASYNC_EXPORT_PREVIEW=ON`. The path is inside `#ifdef ENABLE_ASYNC_EXPORT`, so it only exists in an async build.

### Steps to reproduce

An in-test fake `HttpClient` whose `Session::SendRequest()` accepts the handler and never calls it back, so the session `Export()` starts never finishes.

```cpp
auto client = std::make_shared<FakeHttpClient>([](http_client::EventHandler &) {});
ElasticsearchExporterOptions options;
options.response_timeout_ = 2;  // seconds
ElasticsearchLogRecordExporter exporter(options, client);

auto record = exporter.MakeRecordable();
exporter.Export(nostd::span<std::unique_ptr<sdklogs::Recordable>>(&record, 1));

const auto start   = std::chrono::steady_clock::now();
const bool flushed = exporter.ForceFlush(std::chrono::milliseconds{1});
const auto elapsed = std::chrono::steady_clock::now() - start;
```

### What is the expected behavior?

`flushed == false`, returning at roughly one millisecond. That is what the method's own documentation promises in `es_log_record_exporter.h`:

```
 * @param timeout an option timeout, default to max.
 * @return return true when all data are exported, and false when timeout
```

### What is the actual behavior?

```
[probe] ForceFlush(1ms) returned true after 2000 ms
```

Both halves are wrong. It reports success while a session is still running, and it ignores the caller's deadline in favour of `response_timeout_`.

### Additional context

`exporters/elasticsearch/src/es_log_record_exporter.cc`, in `ForceFlush`:

```cpp
while (timeout_steady > std::chrono::steady_clock::duration::zero())
{
  if (finished_session_counter_.load(...) >= running_counter)
  {
    break;
  }

  std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
  if (std::cv_status::no_timeout != force_flush_cv.wait_for(
                                        lk_cv, std::chrono::seconds{options_.response_timeout_}))
  {
    break;
  }
  timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
}

return timeout_steady > std::chrono::steady_clock::duration::zero();
```

The timeout branch leaves the loop without subtracting the time it just spent, and the loop condition has already established that `timeout_steady` is positive, so the return is `true` on that path no matter what. It is not a corner case: **every** exit through the inner timeout reports success. The only way to get `false` is to be notified repeatedly without completing until the subtraction drains the budget.

The wait itself is `seconds{options_.response_timeout_}` rather than the caller's remaining time, which is where the 2000 ms comes from.

There is a spec angle beyond the header's own wording. `LogRecordProcessor` `ForceFlush` says *"If a timeout is specified, the LogRecordProcessor MUST prioritize honoring the timeout over finishing all calls"*, and the exporter's `Export` *"MUST NOT block indefinitely, there MUST be a reasonable upper limit after which the call must time out"*. A `ForceFlush` that overruns its deadline and then claims success works against both.

Two smaller things in the same area, mentioned so they are not lost rather than to widen this issue:

- The completion callback increments `finished_session_counter_` and calls `notify_all()` without holding `force_flush_cv_m`, so a completion landing between the waiter's predicate check and its park is not seen until the next wakeup. `OtlpHttpClient::ReleaseSession` publishes its counter under a lock for this reason.
- `Shutdown(timeout)` has its parameter commented out and always returns `true`. Bounding it properly needs a deadline aware `FinishAllSessions` in `ext/http`, which is an API question rather than an exporter fix, so I would rather raise that separately if you agree it is worth doing.

I have a fix and a regression test for the `ForceFlush` part and am happy to open a PR. The shape is a single `steady_clock` deadline from the caller timeout, `wait_until` with a completion predicate, and the predicate as the return value, plus publishing the counter under `force_flush_cv_m`. One decision I would rather have your view on first: should `ForceFlush` wait only for the sessions that were in flight when it was entered, which is what the current counter snapshot intends, or should it keep waiting while new exports arrive?

### LINKED ISSUE #4338: [BUG] Elasticsearch async exporter counts a session's completion zero or twice
### Describe your environment

`main` at `11fa0db0`, Linux, GCC 14.2, `-DWITH_ELASTICSEARCH=ON -DWITH_ASYNC_EXPORT_PREVIEW=ON`. `AsyncResponseHandler` only exists under `ENABLE_ASYNC_EXPORT`.

### Steps to reproduce

Two exports through a fake `HttpClient` injected via the public constructor. The first session reports two terminal events, which is what the curl path does on a setup failure (see below). The second never calls back.

```cpp
int session = 0;
auto client = std::make_shared<FakeHttpClient>([&session](http_client::EventHandler &handler) {
  if (++session == 1)
  {
    handler.OnEvent(http_client::SessionState::ConnectFailed, "");
    handler.OnEvent(http_client::SessionState::CreateFailed, "");
  }
  // session 2 never calls back
});

ExportOnce(exporter);  // session 1
ExportOnce(exporter);  // session 2, still in flight

const bool flushed = exporter.ForceFlush(std::chrono::milliseconds{20});
```

### What is the expected behavior?

`flushed == false`. One of the two sessions has not finished.

### What is the actual behavior?

```
[repro] ForceFlush with one session still running returned true
```

Immediately, in 0 ms. `finished_session_counter_` reached 2 from a single session, so `finished >= running` was already true.

I ran this on top of the fix in #4337, so this is not the timeout defect reported in #4336. The counters are monotonic and compared with `>=`, which makes the overshoot permanent: every later flush also returns one session early.

### Additional context

`AsyncResponseHandler` accounts a session's completion neither at most once nor at least once.

**More than once.** `OnResponse` and every handled terminal `OnEvent` call `result_callback_` directly, and the callback increments `finished_session_counter_`. There is no guard. Two in-tree paths produce two terminal callbacks for one session:

- `HttpOperation::SendAsync` dispatches `ConnectFailed` and returns non-`CURLE_OK` when `Setup()` fails (`ext/src/http/client/curl/http_operation_curl.cc`, the `code != CURLE_OK` branch). `Session::SendRequest` then takes its `else` branch and dispatches `CreateFailed` for the same handler (`ext/src/http/client/curl/http_client_curl.cc`). This is deterministic, not a race.
- The async completion lambda in `Session::SendRequest` uses two independent `if` blocks rather than `else if`, so an operation that was aborted and also has a response fires `OnEvent(Cancelled)` and `OnResponse`.

**Fewer than once.** `ReadError`, `WriteError` and `Destroyed` fall into `default: break`, so `result_callback_` never fires and the session is never counted as finished. The synchronous handler treats all three as terminal, `Destroyed` since #4298 and the other two in #4331, so the two handlers in the same file currently disagree.

### Suggested shape

Route every outcome through one idempotent step:

```cpp
void CompleteOnce(sdk::common::ExportResult result) noexcept
{
  bool expected = false;
  if (completed_.compare_exchange_strong(expected, true, std::memory_order_acq_rel))
  {
    result_callback_(result);
  }
}
```

with `OnResponse` and every terminal `OnEvent` going through it, and the switch enumerating all fifteen states so `-Wswitch` catches a future addition rather than a `default:` swallowing it. The guard is also what makes widening the terminal set safe, since it removes the double-completion objection to treating `ReadError`, `WriteError` and `Destroyed` as terminal here the way the synchronous handler does.

Worth covering: each of the three ignored states finishing one session as a failure, a response followed by another terminal event counting once, a cancellation followed by a late response counting once, and a duplicate terminal event counting once.

Two things I would rather flag than fold in. The two `if` blocks and the non-returning gzip failure branch in `Session::SendRequest` are shared HTTP client behaviour, so fixing them there would remove the double-fire for every consumer, but that is a separate change. And an exporter side guard is still warranted either way, because `EventHandler` is a public extension point and an injected client is free to emit whatever it likes.

I am happy to open the PR for the exporter side. It overlaps #4297 and #4337 in the same file, so I would rather land those first unless you would prefer it sooner.

thc1006:
Fixes #4336.

`ForceFlush` waited for `options_.response_timeout_` rather than the timeout the caller gave it, and the branch that ran out of time returned `true` whatever had happened:

```cpp
while (timeout_steady > std::chrono::steady_clock::duration::zero())
{
  ...
  timeout_steady -= std::chrono::steady_clock::now() - start;
}
return timeout_steady > std::chrono::steady_clock::duration::zero();
```

The loop condition has already established that the value is positive, so every flush that ran out of time reported success.

Two more ways it answered without having waited. It compared monotonic totals with no session identity, so a completion from a session started *after* the call could satisfy a waiter for one started before it. And a batch already inside `Export()` was not covered at all, because the session was registered after the request had been built.

One `steady_clock` deadline, taken at function entry so waiting for the mutex spends the caller's budget, and a wait on a predicate over the set of running session ids against a watermark. `wait_until` returns the predicate, so the return value answers the caller's question rather than being a leftover duration. Registration moves to the top of `Export()`. The serialising lock is gone rather than made timed: each call snapshots what it waits for and publishes nothing, so two callers were already safe side by side. Measured at 2921 ms for a `ForceFlush(20ms)` queued behind a 3 second one.

**This is stacked on #4502.** That pull request carries the exactly-once completion this accounting depends on, and closes #4338. GitHub cannot take a fork branch as a base, so until #4502 lands the diff here shows both; this half is 861 of the lines. Review #4502 first.

Twenty cases: sixteen for the flush, and four that check exactly-once through the flush rather than through the completion log line. Reverting each defect turns cases red: the deadline change turns three red, the counter comparison turns the two substitution cases red, and moving registration back below the request build turns `AnExportAlreadyUnderWayIsSomethingToWaitFor` red.

Not fixed here: `Shutdown(timeout)` still ignores its timeout, and `true` means every export the call snapshotted reported a terminal outcome, not that its batch reached Elasticsearch, which is [#3075](https://github.com/open-telemetry/opentelemetry-cpp/issues/3075).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ Increment the:
   [#4334](https://github.com/open-telemetry/opentelemetry-cpp/pull/4334)
 * [BUG] Stop reading past a `nostd::string_view` that is not NUL terminated
   [#4346](https://github.com/open-telemetry/opentelemetry-cpp/pull/4346)
+* [BUG] Stop the Elasticsearch async ForceFlush reporting success without
+  waiting for the sessions it was asked about
+  [#4337](https://github.com/open-telemetry/opentelemetry-cpp/pull/4337)
 
 * [OTLP/HTTP] Honor `Retry-After` response header when retrying exports,
   supporting both delay-seconds and HTTP-date formats per RFC 7231 §7.1.3.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ Increment the:
   `noexcept` function.
   [#4439](https://github.com/open-telemetry/opentelemetry-cpp/issues/4439)
 
+* [BUG] Elasticsearch: report an asynchronous export's outcome exactly once
+  [#4502](https://github.com/open-telemetry/opentelemetry-cpp/pull/4502)
+
 * [CONFIGURATION] Apply general `attribute_limits` per individual limit field.
   If a model-specific limit is set it is used, otherwise the matching general
   limit, otherwise the model-specific default. Limit fields on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,9 @@ Increment the:
   deprecated C headers (`stdint.h`, `stddef.h`, `stdlib.h`, `string.h`,
   `stdio.h`, `ctype.h`, `limits.h`, `assert.h`) with their C++ equivalents
   ([#4349](https://github.com/open-telemetry/opentelemetry-cpp/pull/4349))
+* [BUG] Stop the Elasticsearch async ForceFlush reporting success without
+  waiting for the sessions it was asked about
+  [#4337](https://github.com/open-telemetry/opentelemetry-cpp/pull/4337)
 
 * [CONFIGURATION] Add SDK component builder interfaces to the registry
   [#4358](https://github.com/open-telemetry/opentelemetry-cpp/issues/4358)
--- a/exporters/elasticsearch/CMakeLists.txt
+++ b/exporters/elasticsearch/CMakeLists.txt
@@ -55,4 +55,10 @@ if(BUILD_TESTING)
     TARGET es_log_record_exporter_test
     TEST_PREFIX exporter.
     TEST_LIST es_log_record_exporter_test)
+
+  # AnIndefiniteFlushReturnsOnceEverythingIsFinished exercises the branch that
+  # waits without a deadline, so a regression there does not fail, it stops.
+  # CTest's default bound is 25 minutes, which is a long time to spend learning
+  # that.
+  set_tests_properties(${es_log_record_exporter_test} PROPERTIES TIMEOUT 30)
 endif() # BUILD_TESTING
--- a/exporters/elasticsearch/CMakeLists.txt
+++ b/exporters/elasticsearch/CMakeLists.txt
@@ -55,4 +55,12 @@ if(OTELCPP_BUILD_TESTING)
     TARGET es_log_record_exporter_test
     TEST_PREFIX exporter.
     TEST_LIST es_log_record_exporter_test)
+
+  # AnIndefiniteFlushParksUntilTheOutcomeArrives really does park on the
+  # condition variable, so a regression there does not fail, it stalls until the
+  # exporter's own bound expires. CTest's default is 25 minutes, which is a long
+  # time to spend learning that. This is deliberately well above the bounds the
+  # cases use themselves, so that one of them expiring reports which case failed
+  # rather than being killed from the outside.
+  set_tests_properties(${es_log_record_exporter_test} PROPERTIES TIMEOUT 60)
 endif() # OTELCPP_BUILD_TESTING
--- a/exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_record_exporter.h
+++ b/exporters/elasticsearch/include/opentelemetry/exporters/elasticsearch/es_log_record_exporter.h
@@ -16,8 +16,9 @@
 
 #ifdef ENABLE_ASYNC_EXPORT
 #  include <condition_variable>
-#  include <cstddef>
+#  include <cstdint>
 #  include <mutex>
+#  include <set>
 #endif
 
 OPENTELEMETRY_BEGIN_NAMESPACE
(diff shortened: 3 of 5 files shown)


thc1006:
The one red is `Bazel asan config`, and it is not from this diff. `BasicCurlHttpTests.ElegantQuitQuick` in `//ext/test/http:curl_http_test` missed its 20 ms wall clock budget at `curl_http_test.cc:940`. That is the flake #4342 described and #4488 fixes; #4342 is closed, the flake is not.

This branch changes five files, all under `exporters/elasticsearch/`, so rather than argue it from the diff:

```
$ bazel query "somepath(//ext/test/http:curl_http_test, //exporters/elasticsearch/...)"
INFO: Empty results
```

The same job is green on `main` at `70fdb766`, the commit this is rebased onto.

I cannot re-run a job here, so it will clear on the next push or on a maintainer re-run.

PR #4351, opened 2026-08-03
[SDK] Create instrumentation scope library

dbarker:
Contributes to [#1429](https://github.com/open-telemetry/opentelemetry-cpp/issues/1429)

This PR creates CMake and Bazel targets for the instrumentation scope library. This is an important target to provide as it supports users in linking to just what they need use the `InstrumentationScope` object in their projects (e.g. building custom configurators, etc.). Also supports moving SDK code out of headers and into .cc files (#1429). 

With this change users can link directly to the library with: 

```cmake
   target_link_libraries(my_library PRIVATE opentelemetry-cpp::instrumentation_scope)
``` 

The library will also be a public transitive link provided by metrics, trace, and logs targets. 

## Changes

- moves instrumentation_scope.h impl to instrumentation_scope.cc and creates library targets with CMake and Bazel


marcalff (approved):
LGTM

PR #4353, opened 2026-08-04
[METRICS SDK] Enforce a runtime minimum scale for Base2ExponentialHistogramAggregation

### LINKED ISSUE #4325: Define and enforce a runtime minimum scale for Base2ExponentialHistogramAggregation
**Describe your environment**

- Repository revision: PR [#4324](https://github.com/open-telemetry/opentelemetry-cpp/pull/4324)
- Reproduced on Windows with MSVC 19.44 using a CMake/Ninja Release build
- The runtime scale behavior predates PR #4324; that PR only exposed it during review

**Steps to reproduce**

Configure a base2 exponential histogram at the lowest currently accepted `max_scale_` and `max_size_`, then record values at both ends of the finite `double` range:

```cpp
using namespace opentelemetry::sdk::metrics;

Base2ExponentialHistogramAggregationConfig config;
config.max_scale_ = kMaxScaleMin;  // -10
config.max_size_  = kMaxSizeMin;   // 2

Base2ExponentialHistogramAggregation aggregation{&config};
aggregation.Aggregate((std::numeric_limits<double>::denorm_min)());
aggregation.Aggregate((std::numeric_limits<double>::max)());

const auto point =
    opentelemetry::nostd::get<Base2ExponentialHistogramPointData>(aggregation.ToPoint());
EXPECT_GE(point.scale_, kMaxScaleMin);
```

The current implementation accepts `-10` as the configured maximum scale in [`aggregation_config.h`](https://github.com/open-telemetry/opentelemetry-cpp/blob/c0142c63e3df5ebc5c1723f9a28e06a561f0fb29/sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h#L59-L64), but automatic scale reduction never applies a lower bound.

The aggregation should define and enforce a reasonable runtime minimum scale for every automatic reduction path.

If `-10` is selected as the C++ SDK's runtime minimum, `point.scale_` should never become less than `-10`. When an input range still cannot fit within `max_size_` at that floor, the implementation should apply an explicit, documented policy while preserving all recorded counts.

The reproduction produces:

```console
point.scale_ == -11
```

At scale `-10`, `denorm_min()` and `max()` map to bucket indices `-2` and `0`. Their three-bucket span exceeds `max_size_ == 2`, so `GetScaleReduction()` requests one additional reduction. [`Base2ExponentialHistogramAggregation::Downscale()`](https://github.com/open-telemetry/opentelemetry-cpp/blob/c0142c63e3df5ebc5c1723f9a28e06a561f0fb29/sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc#L232-L262) then subtracts that reduction without consulting any runtime floor:

```cpp
const uint32_t scale_reduction =
    GetScaleReduction(start_index, end_index, point_data_.max_buckets_);
Downscale(scale_reduction);
// ...
point_data_.scale_ -= static_cast<int32_t>(by);
```

**Additional context**

Related:

- Reviewer discussion on PR [#4324](https://github.com/open-telemetry/opentelemetry-cpp/pull/4324#discussion_r3693625734)
- Configuration validation issue [#4250](https://github.com/open-telemetry/opentelemetry-cpp/issues/4250)

ThomsonTan:
Fixes #4325 

## Changes

This change enforces a runtime minimum scale of -10 for base2 exponential histograms, guarantees sufficient bucket capacity at that floor to preserve recorded counts, centralizes scale reduction across record, merge, and diff paths, guards the indexer against oversized shifts, and adds regression tests plus changelog documentation for the new behavior.


sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -94,6 +97,102 @@ void DownscaleBuckets(std::unique_ptr<AdaptingCircularBufferCounter> &buckets, u
   buckets->Downscale(by);
 }
 
+// Bucket buffers are never narrower than kMinBucketsAtFloor so that the whole double range stays
+// representable once the scale is pinned at kMinRuntimeScale. This only widens the degenerate
+// max_size_ == 2 configuration, and the extra slot is only reachable at the floor.
+size_t BucketCapacity(size_t max_buckets) noexcept
+{
+  return (std::max)(max_buckets, kMinBucketsAtFloor);
```
lalitb: This can create 3 buckets when the user configured `max_size = 2`. Since `max_size` is the maximum number of buckets, can we use `-11` as the runtime floor instead? At `-11`, the full double range fits in 2 buckets, so we do not need this exception.

ThomsonTan: Good catch, thanks. Changed the runtime floor to `-11` and dropped `kMinBucketsAtFloor`, so `max_size` is now a hard cap. At `-11` every finite double maps to an index in `[-1, 0]`, and the `max_size = 2` repro from #4325 now ends at scale `-11` with a two-bucket span and capacity 2.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -221,45 +326,66 @@ void Base2ExponentialHistogramAggregation::AggregateIntoBuckets(
...
+  // instead of relying on Increment() to fail.
+  uint32_t scale_reduction = 0;
+  if (!buckets->Empty())
   {
-    const int32_t start_index = (std::min)(buckets->StartIndex(), index);
-    const int32_t end_index   = (std::max)(buckets->EndIndex(), index);
-    const uint32_t scale_reduction =
-        GetScaleReduction(start_index, end_index, point_data_.max_buckets_);
-    Downscale(scale_reduction);
+    scale_reduction =
```
lalitb: This range calculation now runs for every non-first recording, even when the new index already fits. `Increment()` checks the range again below, and both checks happen while holding the spin lock. Can we keep the old fast path: try `Increment(index, 1)` first, and only calculate the downscale when it fails?

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -221,45 +326,66 @@ void Base2ExponentialHistogramAggregation::AggregateIntoBuckets(
...
+                          (std::max)(buckets->EndIndex(), index), point_data_.max_buckets_);
+  }
 
-    buckets->Increment(index >> scale_reduction, 1);
+  // Downscale() may stop short of the request at the floor, so shift the index by what was
+  // actually applied.
+  const uint32_t applied = Downscale(scale_reduction);
+  if (!buckets->Increment(index >> applied, 1))
+  {
+    OTEL_INTERNAL_LOG_ERROR(
```
lalitb, line 394: This error can repeat on every recording when externally supplied point data has bucket indexes that do not match its scale. For example, at scale -11, an existing bucket at index 100 can never accept a normal value at index -1. Since the state does not change, every call logs again while holding the spin lock. Can we log this only once, like the scale-floor warning?

sdk/test/metrics/aggregation_test.cc
```diff
@@ -1117,3 +1118,249 @@ TEST(Aggregation, Base2ExponentialHistogramAggregationRecordPathRepeatedDownscal
...
+    const Base2ExponentialHistogramIndexer collapsed(scale);
+    for (double value : values)
+    {
+      EXPECT_EQ(collapsed.ComputeIndex(value), shifted.ComputeIndex(value))
+          << "scale " << scale << ", value " << value;
+    }
+  }
+}
+
+TEST(Aggregation, Base2ExponentialHistogramAggregationCopiedPointDataKeepsFloorCapacity)
```
lalitb, line 1394: This covers a narrow physical buffer, but the point still advertises a valid `max_buckets_` value of 2. Can we also test externally supplied `max_buckets_` values 0 and 1, which exercise the new loop bound? It would also be useful to test point data whose bucket index does not match its scale, so the repeated-error case above stays covered.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Increment the:
 
 ## [Unreleased]
 
+* [METRICS SDK] Enforce a runtime minimum scale for
```
lalitb: Can we make this more specific? Histograms may now export scale -11, and this also prevents a non-terminating loop for externally supplied point data with max_buckets_ set to 0 or 1. These are useful behavior changes for users reading the changelog.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
ThomsonTan: Added flag `bucket_index_error_emitted_` to gate the log.

sdk/test/metrics/aggregation_test.cc
ThomsonTan: Added both, thanks.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
dbarker: Hi @lalitb, Is this issue resolved and PR ready to approve?

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -164,14 +263,20 @@ Base2ExponentialHistogramAggregation::Base2ExponentialHistogramAggregation(
     point_data_.negative_buckets_ =
         std::make_unique<AdaptingCircularBufferCounter>(*point_data.negative_buckets_);
   }
+
+  EnsureBucketCapacity(point_data_.positive_buckets_, BucketCapacity(point_data_.max_buckets_));
```
lalitb: Can we validate or normalize the supplied buckets here, not only resize them? When bucket indexes do not match the supplied scale, `Record` drops the bucket after count and sum were already updated, so the exported count no longer matches the buckets. `Merge` can assert in debug builds or lose counts in release builds, and `Diff` has the same problem when the result contains far-apart indexes. Since these constructors are where supplied point data enters, rejecting or normalizing it here would give `Record`, `Merge`, and `Diff` one consistent rule. Please cover all three paths with the same malformed point data.

ThomsonTan: Good catch! Validating at the constructor was the right call. Fixed.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -436,9 +564,9 @@ std::unique_ptr<Aggregation> Base2ExponentialHistogramAggregation::Diff(
       (right.zero_count_ >= left.zero_count_) ? (right.zero_count_ - left.zero_count_) : 0;
 
   result_value.positive_buckets_ =
-      std::make_unique<AdaptingCircularBufferCounter>(right.max_buckets_);
+      std::make_unique<AdaptingCircularBufferCounter>(BucketCapacity(result_value.max_buckets_));
```
lalitb, line 567: This seems to be regression arising from the latest commit. Could we handle the case where one side has no positive or negative buckets? We currently skip downscaling, but the other side's buckets may not fit the smaller result buffer. This can assert in debug builds or drop counts in release builds. Please add a test with one side empty and different bucket budgets.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
```diff
@@ -94,6 +97,142 @@ void DownscaleBuckets(std::unique_ptr<AdaptingCircularBufferCounter> &buckets, u
...
+                             int32_t max_index) noexcept
+{
+  return buckets.Empty() || (buckets.StartIndex() >= min_index && buckets.EndIndex() <= max_index);
+}
+
+void NormalizeSuppliedPointData(Base2ExponentialHistogramPointData &point_data) noexcept
+{
+  const size_t capacity = BucketCapacity(point_data.max_buckets_);
+  EnsureBucketCapacity(point_data.positive_buckets_, capacity);
+  EnsureBucketCapacity(point_data.negative_buckets_, capacity);
```
lalitb, line 153: Could we check the indexes before widening the buffers? A one-slot buffer containing `INT32_MAX` overflows the copy loop's counter before the new validation runs. Please add a test for this boundary case too.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
ThomsonTan: Good catch, thanks: fixed the regression in https://github.com/open-telemetry/opentelemetry-cpp/pull/4353/commits/bb77c835f65024d809faadde35efb648c651f090 by having `GetScaleReductionForUnion()` return 0 only when both sides are empty and otherwise compute the reduction from the populated side's span, and `DiffOneSideEmptyFitsReportedBudget` diffs an empty 2-bucket aggregation against an 8-bucket one on each sign.

sdk/src/metrics/aggregation/base2_exponential_histogram_aggregation.cc
ThomsonTan: Thanks, bb77c835f now checks the supplied indexes in `NormalizeSuppliedPointData()` before `EnsureBucketCapacity()` widens anything and gives the copy loop an `int64_t` cursor so it cannot wrap at `INT32_MAX`, with `RejectsMaxIndexBeforeWidening` covering a one-slot `INT32_MAX` counter through both constructors.

lalitb (approved):
Thanks.

Issue #4352, opened 2026-08-04
[CONFIGURATION] Registry has builders for all components and link only what you use


dbarker:
### Background:

The `configuration_core` library was created to separate the programmatic configuration interface from the YAML parsing components in completion of #4134. 

It links to all SDK signal libraries (traces, logs, and metrics). Therefore, if a user of programmatic or yaml configuration desires to only configure one signal provider they must still link all SDK libraries. 

This is driven by two things: `SdkBuilder`, which constructs all signal providers via direct `Factory::Create` calls in one TU, and `ConfiguredSdk`, whose header stores all three SDK provider types. This also limits what SDK component builders a user can register for configuration.

### Goals:

This ticket has two goals:
1. Allow linking just to the SDK components being configured and
2. Support registering builders for every configurable SDK component, including the signal providers themselves.

Together these would allow linking and registering only the builders that an application supports, providing more control over configurable options, ability to customize the SDK components, and reduce binary size.

Some use cases enabled by this include:
   - Registering a custom builder to replace a default SDK component builder
       - A user may define their own implementation of a`TracerConfigurator`, `BatchSpanProcessor`, and `TracerProvider` for example. 
   - Registering mock component builders for test scenarios
   - Intentionally not registering a component's builder to prevent its use in the application

### Example:

If the config components are decoupled from the SDK signal providers, the example below should only require the application to link the configuration library, the trace default builders, the gRPC span exporter, and users custom batch processor. No metrics or logs libraries would be linked and attempts to configure them would result in an error. 

```cpp
namespace cfg = opentelemetry::sdk::configuration;

auto otlp_grpc_exporter      = std::make_unique<cfg::OtlpGrpcSpanExporterConfiguration>();
otlp_grpc_exporter->endpoint = "http://localhost:4317";

auto processor               = std::make_unique<cfg::BatchSpanProcessorConfiguration>();
processor->exporter          = std::move(otlp_grpc_exporter);

auto tp = std::make_unique<cfg::TracerProviderConfiguration>();
tp->processors.push_back(std::move(processor));

auto model             = std::make_unique<cfg::Configuration>();
model->tracer_provider = std::move(tp);

// New Registry behavior:
//   Constructing a registry creates an empty registry that brings no
//   SDK dependencies outside of `configuration_core`. A config node 
//   with no registered builder is a error
auto registry = std::make_shared<cfg::Registry>();

// New SDK signal specific method to register all its component builders:
[...]

PR #4355, opened 2026-08-04
[CONFIGURATION] Decouple SDK includes from builder interface classes

dbarker:
Contributes to #4352

Follow up to #4335 to continue breaking dependencies between `configuration_core` and the SDK signal libraries.

The goal is to decouple the abstract builder headers from the SDK component headers. After this change, including an abstract builder header no longer pulls in the SDK headers of the component it builds. The SDK type is forward declared. The requirement for the complete type is deferred to the translation units that implement or invoke a builder, which by definition already include the concrete SDK header and link that signal's library. 

This makes the contract for `configuration_core` more explicit in that it provides the
abstract builder headers without transitively providing the SDK signal headers.

## Changes

- Replace SDK component includes with forward declarations in the abstract builder headers


marcalff (approved):
LGTM

PR #4349, opened 2026-08-03
[CODE HEALTH] Enable clang-tidy modernize-deprecated-headers and fix violations

om7057:
Fixes #4348

## Changes

Enables the `modernize-deprecated-headers` clang-tidy check, which was disabled in `.clang-tidy`, and replaces deprecated C headers with their C++ equivalents (`stdint.h` -> `cstdint`, `stddef.h` -> `cstddef`, `stdlib.h` -> `cstdlib`, `string.h` -> `cstring`, `stdio.h` -> `cstdio`, `ctype.h` -> `cctype`, `limits.h` -> `climits`, `assert.h` -> `cassert`) across `api`, `sdk`, `exporters`, `ext`, `examples`, `opentracing-shim`, `resource_detectors`, `test_common`, and `functional`.

As noted in the issue, IWYU was still recommending the deprecated headers, so `.iwyu.imp` is updated first to map `struct tm`/`struct FILE` usages to `<ctime>`/`<cstdio>` instead of `<time.h>`/`<stdio.h>`.

Two now-duplicate includes that this change surfaced were also removed, in `api/test/core/timestamp_test.cc` and `exporters/elasticsearch/src/es_log_recordable.cc`.

* [x] `CHANGELOG.md` updated for non-trivial changes
* [ ] Unit tests have been added
* [ ] Changes in public API reviewed

PR #4356, opened 2026-08-04
[CODE HEALTH] Qualify C library calls with std:: after modernize-deprecated-headers

### LINKED ISSUE #4354: [CODE HEALTH] Qualify C library calls with std:: after modernize-deprecated-headers
Follow-up to #4348 / #4349.

Switching deprecated C headers to their C++ equivalents (`stdlib.h` -> `cstdlib`, `string.h` -> `cstring`, etc.) leaves the actual library calls unqualified (e.g. `memcpy(...)` instead of `std::memcpy(...)`). Both forms compile today since most implementations still expose these names in the global namespace, but only the `std::` form is guaranteed by the standard.

This ticket tracks adding `std::` to those unqualified calls where feasible, per direction from #4349.

Priority list:

From `<cstring>`: `memcpy`, `memmove`, `memset`, `memcmp`, `strlen`, `strcmp`, `strncmp`, `strchr`, `strrchr`, `strstr`, `strncpy`, `strncat`, `strtok`, `strerror`

From `<cctype>`: `isalnum`, `isalpha`, `isdigit`, `isspace`, `tolower`, `toupper`, ...

From `<cstdlib>`: `strtol`, `strtoll`, `strtoul`, `strtod`, `malloc`/`free`, `qsort`, `bsearch`, `abs`/`labs`/`llabs`, `getenv`, `abort`, `exit`

From `<cstdio>`: `snprintf`, `vsnprintf`, `fopen`, `fclose`, `fprintf`, `fscanf`, `printf`, `FILE`

From `<ctime>`: `time`, `gmtime`, `localtime`, `strftime`, `mktime`, `tm`

As with the header migration, this should probably land as one or more directory-scoped PRs rather than a single repo-wide change.

om7057:
Fixes #4354

## Changes

Follow-up to #4349. Swapping deprecated C headers for their C++ equivalents left the underlying library calls unqualified (`memcpy`, `strcmp`, `isspace`, `strtod`, `snprintf`, `strftime`, etc.), relying on most implementations also exposing these names in the global namespace, which the standard does not guarantee.

Adds `std::` to call sites for the functions covered by `<cstring>`, `<cctype>`, `<cstdlib>`, `<cstdio>`, and `<ctime>`, plus the `FILE` and `tm` types, across `api`, `sdk`, `exporters`, `ext`, `examples`, and `functional`. Also adds a couple of `#include`s that were missing and relied on transitive inclusion.

Left untouched, per review feedback on #4349:

* Vendored code (`api/include/opentelemetry/nostd/internal/absl`, `exporters/etw/include/opentelemetry/exporters/etw/TraceLoggingDynamic.h`)
* Call sites already qualified with `::` (explicit global scope), since that looked like a deliberate choice rather than an oversight
* Platform-specific functions with no `std::` form (`gmtime_r`/`gmtime_s`, `strtok_s`/`strtok_r`)

## Test plan

- [x] Full local build (`cmake --build .`) with `BUILD_TESTING`, `WITH_EXAMPLES`, `WITH_FUNC_TESTS`, `WITH_OTLP_FILE`, `WITH_OTLP_HTTP`, `WITH_CONFIGURATION`, and `WITH_EXAMPLES_HTTP` enabled, 0 errors, 0 warnings.
- [x] `ctest` on both configs, 796/796 and 1233/1237 passed (the 4 failures were `ext.http.curl` timing tests that are flaky under parallel load in this environment; all pass at `-j1`).

### DIFF
--- a/api/include/opentelemetry/common/kv_properties.h
+++ b/api/include/opentelemetry/common/kv_properties.h
@@ -192,7 +192,7 @@ public:
     nostd::unique_ptr<const char[]> CopyStringToPointer(nostd::string_view str)
     {
       char *temp = new char[str.size() + 1];
-      memcpy(temp, str.data(), str.size());
+      std::memcpy(temp, str.data(), str.size());
       temp[str.size()] = '\0';
       return nostd::unique_ptr<const char[]>(temp);
     }
--- a/api/include/opentelemetry/common/string_util.h
+++ b/api/include/opentelemetry/common/string_util.h
@@ -27,11 +27,11 @@ public:
     {
       return nostd::string_view();
     }
-    while (left <= right && isspace(str[left]))
+    while (left <= right && std::isspace(str[left]))
     {
       left++;
     }
-    while (left <= right && isspace(str[right]))
+    while (left <= right && std::isspace(str[right]))
     {
       right--;
     }
--- a/api/include/opentelemetry/trace/span_id.h
+++ b/api/include/opentelemetry/trace/span_id.h
@@ -23,7 +23,10 @@ public:
   SpanId() noexcept = default;
 
   // Creates a SpanId with the given ID.
-  explicit SpanId(nostd::span<const uint8_t, kSize> id) noexcept { memcpy(rep_, id.data(), kSize); }
+  explicit SpanId(nostd::span<const uint8_t, kSize> id) noexcept
+  {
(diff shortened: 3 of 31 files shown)


marcalff (approved):
LGTM

dbarker (approved):

Issue #4339, opened 2026-08-02
[BUG] OTLP ForceFlush can report success with requests pending, and waits on the exporter timeout rather than the caller's


thc1006:
### Describe your environment

`main` at `11fa0db0`, read at source level. Both paths are inside `#ifdef ENABLE_ASYNC_EXPORT`.

The gRPC half below is a model rather than a run of the exporter, and here is why, because it is itself worth knowing. I did build the OTLP exporters (Debian system protobuf 3.21.12 and gRPC 1.51.1) and tried to drive `ForceFlush` directly through the existing `friend class OtlpGrpcClientTestPeer` hook. That does not compile: `OtlpGrpcClientAsyncData` is only forward declared in `otlp_grpc_client.h` and defined in `otlp_grpc_client.cc`, so no test translation unit can dereference `async_data_` even with friendship. The counters `ForceFlush` decides on are unreachable from a test, which is a fair part of why this function has no coverage today, and it is something a fix will have to answer for. The Elasticsearch equivalents are #4336 and #4338; this is the same family in the two exporters most people use, which is why it is a separate report.

### What is the expected behavior?

`ForceFlush(timeout)` returns `true` only when the requests it was waiting for have finished, and returns within the timeout it was given.

### What is the actual behavior?

**OTLP gRPC returns success when any one request completes.** `OtlpGrpcClient::ForceFlush`:

```cpp
while (timeout_steady > std::chrono::steady_clock::duration::zero() &&
       request_counter > async_data_->finished_request_counter.load(std::memory_order_acquire))
{
  std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
  if (std::cv_status::timeout !=
      async_data_->session_waker.wait_for(lock, async_data_->export_timeout))
  {
    break;
  }
  timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
}
return timeout_steady > std::chrono::steady_clock::duration::zero();
```

`std::cv_status::timeout != status` is true when the wait was **notified**, so any notification leaves the loop without re-checking the counter, and the return is the leftover duration rather than the predicate.

A model that keeps the loop and both branch conditions exactly, with two requests where one completes:

```
started=2 finished=1 ForceFlush returned true after 50 ms
```

That is a model, not the exporter. What I can state about the exporter itself is structural: the counter is only re-read by the `while` condition, and the `break` skips it.

[...]

PR #4357, opened 2026-08-04
[BUG] Stop OTLP ForceFlush returning on the first notification and overrunning its deadline

thc1006:
Towards #4339. @owent confirmed the inverted comparison and suggested bounding the wait by the caller's remaining time, which is what this does. It does not close the issue; see "What this leaves" below.

## OTLP gRPC reported the whole flush done on one completion

`OtlpGrpcClient::ForceFlush` left its wait loop on any notification:

```cpp
if (std::cv_status::timeout !=
    async_data_->session_waker.wait_for(lock, async_data_->export_timeout))
{
  break;
}
```

`std::cv_status::timeout != status` is true when the wait was **notified**, and the loop condition is the only place `finished_request_counter` is read, so the `break` skipped it. One finished request out of many ended the wait, and the function then returned the leftover duration rather than the predicate.

Rather than inverting the test I dropped the `break` and let the loop condition decide, which is what the `==` form reduces to once the wait is bounded. The bounded `wait_for` is what covers a notification lost between the check and the wait; the `break` was not doing that. The return is the counter predicate now.

The comment above it described `running_sessions_` and gc session cleanup, neither of which exists in the gRPC client. It is a verbatim copy of the `OtlpHttpClient` one, down to the missing space in `forever.We`, so it is replaced rather than kept.

## Both clients waited on the exporter's timeout, not the caller's

`ForceFlush(1ms)` could block for the full configured export timeout. The wait interval is `min(configured timeout, caller's remaining time)` now, as suggested. No notification path and no lock discipline changes.

`OtlpHttpClient::Shutdown` is where that shows up in practice:

```cpp
// Wait util all sessions are canceled.
while (cleanupGCSessions())
{
  ForceFlush(std::chrono::milliseconds{1});
}
```

Each of those calls could take `options_.timeout`, so shutdown cost a multiple of the export timeout rather than a multiple of 1 ms.

## The result is the predicate, on both clients

Both loops used to end by returning what was left of the deadline. A completion that lands between the last check and the last wait takes its notification with it, and the wait then consumes the rest of the deadline and reports a timeout for work that had already finished. Each client now returns the condition it was waiting on.

## One thing I tried and backed out

`OtlpGrpcClient` increments `finished_request_counter` before it runs `grpc_async_callback`, which is where the `ExportResult` reaches the caller, so a bounded wait can see the counter satisfied while the result is still on its way. Moving the increment after the callback closes that, and I had it that way for a while.

It is not worth it here. `DelegateAsyncExport` is public and takes the callback, so a caller whose callback calls `ForceFlush()` would wait for an increment that cannot happen until the callback returns, and the no-deadline form of that wait never ends. Trading a bounded early signal for an unbounded self-wait is the wrong way round, and neither of the two bugs this PR fixes needs the change. Where a request counts as finished belongs with the identity work #4339 still needs.

## Test

In `otlp_http_exporter_custom_client_test.cc`, `ForceFlushReturnsWithinTheCallerDeadline` holds a request open with the nosend HTTP client and asserts `ForceFlush(50ms)` returns inside its deadline against a 30 second client timeout. It drives `OtlpHttpClient::Export` directly with a non-zero request budget, so it does not depend on `ENABLE_ASYNC_EXPORT`.

| | result | elapsed |
| --- | --- | --- |
| `otlp_http_client.cc` reverted to `main`, test kept | `Expected: (elapsed) < (1000), actual: 30000 vs 1000` | 30000 ms |
| this branch | passed | 50 ms |

`ForceFlushReportsSuccessOnceTheSessionIsDone` covers the other side of the new return, that a flush with nothing left to wait for says so. Nothing was asserting the return value on that path. It passes on `main` as well, where the old `timeout_steady > 0` happened to agree, so it guards the predicate rather than reproducing a bug.

Built and run in two configurations, `WITH_ASYNC_EXPORT_PREVIEW` off and on, seven of seven passing in both, whole file 51 ms.

The file carries the `#ifndef OPENTELEMETRY_STL_VERSION` guard its neighbours in `exporters/otlp/test` already use, so like them it builds under `WITH_STL=OFF` and is empty in the `cmake.c++*.stl.test` jobs.

The gRPC half is covered through the stub. `OtlpGrpcClientAsyncData` is defined in the `.cc`, so no test can read `finished_request_counter` directly, but it does not have to: `OtlpMockTraceServiceStub` was running the completion callback inline, so every request was already finished by the time `ForceFlush` could be called and the loop never had anything to wait for. Holding the callback instead leaves the request in flight, which is all these two cases need.

In `otlp_grpc_exporter_test.cc`, `OtlpGrpcExporterFlushTestPeer` carries two cases. `ForceFlushReturnsWithinTheCallerDeadline` asks for 50 ms against the 10 second export timeout. `ForceFlushKeepsWaitingWhenAnotherRequestCompletes` leaves two requests in flight and completes one while the flush is parked, which is the notification that used to end the wait.

| | result | elapsed |
| --- | --- | --- |
| `otlp_grpc_client.cc` reverted to `main`, tests kept | `Expected: (elapsed.count()) < (5000), actual: 10000 vs 5000` | 10000 ms |
| same, second case | `Value of: flushed / Actual: true / Expected: false` | 100 ms |
| this branch | both passed | 59 ms, 1006 ms |

The second row is the bug in one line: the flush returned success at the exact moment a different request completed, with the one it was waiting for still in flight.

56 of 56 with `WITH_ASYNC_EXPORT_PREVIEW=ON`. With it off the client tracks no in-flight requests, so the two cases skip in the fixture rather than being compiled away. `gtest_add_tests` registers from the source text, so an `#if`-ed out case is still registered with CTest and then passes without running. CTest lists both, and the binary reports 54 passed, 2 skipped.

## What this leaves

`ForceFlush` still does not fully meet the contract in your point 1. Both clients snapshot a monotonic total and wait for the finished total to reach it, so a request that starts **after** the call can satisfy the snapshot while one that was pending **before** it is still running. On entry `started=10 finished=8`; two new requests start and finish; finished reaches 10 and the wait ends with the original two in flight.

Reading both clients for that follow-up turned up four more places where the boundary sits in the wrong spot, all of them on `main` and none of them touched here:

* a request is not registered until after the exporter has serialised it, so a flush
during `PopulateRequest` sees nothing;
* the HTTP registration is two steps, the `running_sessions_` insert under the lock and
`start_session_counter_` after it;
* the HTTP loop leaves on `running_sessions_.empty()`, which counts sessions started
after the call, so a flush can also wait too long;
* `Unbind` runs the caller's callback before `ReleaseSession`, deliberately, because the
response lives in an arena the session data owns.

All five want the same thing, a token issued at the exporter's `Export` entry rather than a counting total, which also settles the wrap a monotonic total has on a 32 bit target. That is a change to both hot paths, so I kept it out of one you had already scoped and wrote the set up on #4339 instead. Happy to send it as a follow-up.

## Behaviour change worth knowing

`OtlpGrpcClient::Shutdown()` defaults to a timeout of 0, documented as "no timeout is applied". With the `break` gone that call waits for outstanding requests instead of returning after the first completion, so an application that exported and exited immediately will now see the export finish rather than be cancelled.

It does not wait indefinitely under a normal configuration: a call sets `context->set_deadline(now() + options.timeout)` whenever `options.timeout` is positive, so each one completes or fails within the export timeout and increments the counter either way. A caller that wants the old return-early behaviour can pass an explicit short timeout.

## One thing I noticed but did not touch

`OtlpGrpcClientOptions::timeout` and `OtlpHttpClientOptions::timeout` both default to a zero duration, and the exporters fill in a real value. If a client is left with zero, `set_deadline` is skipped, and in `ForceFlush` the wait interval is zero as well, so the loop spins instead of sleeping. `main` does the same thing there, since `wait_for(lock, 0)` returns `cv_status::timeout` and the old `!=` test then did not break either, so this is not something the PR changes. I left it alone because the sensible re-check interval for "no configured timeout" is a policy call, and falling back to the caller's remaining time would reintroduce the overflow that `AdjustWaitForTimeout` exists to avoid. A sketch of the policy is on #4339 with the rest of the follow-up.

## Checklist

* [x] Changes in public API reviewed: no signature changes, but `ForceFlush` returns
`false` in cases where it used to return `true`, and a default `Shutdown()` now waits. Both are described above.

## CI note

`func_otlp_grpc` has a pre-existing nondeterministic double free that aborts after the tests report passing, and it reds gcc-14 and clang-18 on `main` as well. It is not related to this change.

### DIFF
--- a/exporters/otlp/src/otlp_http_client.cc
+++ b/exporters/otlp/src/otlp_http_client.cc
@@ -878,13 +878,6 @@ bool OtlpHttpClient::ForceFlush(std::chrono::microseconds timeout) noexcept
   // What the wait was for, rather than what is left of the deadline. A session that finishes
   // between the last check and the last wait takes its notification with it, and the flush it was
   // holding up has happened.
-  {
-    std::lock_guard<std::recursive_mutex> guard{session_manager_lock_};
-    if (running_sessions_.empty())
-    {
-      return true;
-    }
-  }
   return finished_session_counter_.load(std::memory_order_acquire) >= wait_counter;
 }
 
--- a/exporters/otlp/src/otlp_http_client.cc
+++ b/exporters/otlp/src/otlp_http_client.cc
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#include <algorithm>
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
@@ -857,8 +858,12 @@ bool OtlpHttpClient::ForceFlush(std::chrono::microseconds timeout) noexcept
     // When changes of running_sessions_ and notify_one/notify_all happen between predicate
     // checking and waiting, we should not wait forever.We should cleanup gc sessions here as soon
     // as possible to call FinishSession() and cleanup resources.
+    const std::chrono::steady_clock::duration wait_interval = (std::min)(
+        std::chrono::duration_cast<std::chrono::steady_clock::duration>(options_.timeout),
+        timeout_steady);
+
     std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
-    if (std::cv_status::timeout == session_waker_.wait_for(lock, options_.timeout))
+    if (std::cv_status::timeout == session_waker_.wait_for(lock, wait_interval))
     {
       cleanupGCSessions();
     }
@@ -870,7 +875,17 @@ bool OtlpHttpClient::ForceFlush(std::chrono::microseconds timeout) noexcept
     timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
   }
 
-  return timeout_steady > std::chrono::steady_clock::duration::zero();
+  // What the wait was for, rather than what is left of the deadline. A session that finishes
+  // between the last check and the last wait takes its notification with it, and the flush it was
+  // holding up has happened.
+  {
+    std::lock_guard<std::recursive_mutex> guard{session_manager_lock_};
+    if (running_sessions_.empty())
+    {
+      return true;
+    }
+  }
+  return finished_session_counter_.load(std::memory_order_acquire) >= wait_counter;
 }
 
 bool OtlpHttpClient::Shutdown(std::chrono::microseconds timeout) noexcept
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ Increment the:
   [#4334](https://github.com/open-telemetry/opentelemetry-cpp/pull/4334)
 * [BUG] Stop reading past a `nostd::string_view` that is not NUL terminated
   [#4346](https://github.com/open-telemetry/opentelemetry-cpp/pull/4346)
+* [BUG] Stop OTLP `ForceFlush` returning on the first notification and
+  overrunning the caller's deadline
+  [#4357](https://github.com/open-telemetry/opentelemetry-cpp/pull/4357)
 
 * [OTLP/HTTP] Honor `Retry-After` response header when retrying exports,
   supporting both delay-seconds and HTTP-date formats per RFC 7231 §7.1.3.
--- a/exporters/otlp/src/otlp_grpc_client.cc
+++ b/exporters/otlp/src/otlp_grpc_client.cc
@@ -757,20 +757,20 @@ bool OtlpGrpcClient::ForceFlush(
   while (timeout_steady > std::chrono::steady_clock::duration::zero() &&
          request_counter > async_data_->finished_request_counter.load(std::memory_order_acquire))
   {
-    // When changes of running_sessions_ and notify_one/notify_all happen between predicate
-    // checking and waiting, we should not wait forever.We should cleanup gc sessions here as soon
-    // as possible to call FinishSession() and cleanup resources.
+    // A completion can land between the loop condition and the wait, so the wait is bounded and the
+    // condition re-read on every wakeup. Never wait past the caller's deadline.
+    const std::chrono::steady_clock::duration wait_interval =
+        (std::min)(std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+                       async_data_->export_timeout),
+                   timeout_steady);
+
     std::chrono::steady_clock::time_point start_timepoint = std::chrono::steady_clock::now();
-    if (std::cv_status::timeout !=
-        async_data_->session_waker.wait_for(lock, async_data_->export_timeout))
-    {
-      break;
-    }
+    async_data_->session_waker.wait_for(lock, wait_interval);
 
     timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
   }
 
-  return timeout_steady > std::chrono::steady_clock::duration::zero();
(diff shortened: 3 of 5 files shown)


thc1006:
The one red job is a Docker Hub timeout in the functional test setup, before anything
is compiled:

```
#2 [internal] load metadata for docker.io/otel/opentelemetry-collector:0.123.0@sha256:c8e3...
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/otel/opentelemetry-collector/manifests/sha256:...": dial tcp
ERROR: failed to build: failed to solve: DeadlineExceeded
```

The same job passed on the previous commit of this branch,
https://github.com/open-telemetry/opentelemetry-cpp/actions/runs/30941753951, and the
two commits differ by two lines of `CHANGELOG.md` and nothing else, so no part of the
diff reaches it.

It is also not the `func_otlp_grpc` double free in the CI note above. That one aborts
after the tests report passing; this one never got to the build.

I do not have re-run rights on the repository. Happy to force-push the same tree to
retrigger it if that is easier than a re-run.

exporters/otlp/src/otlp_http_client.cc
```diff
@@ -870,7 +875,17 @@ bool OtlpHttpClient::ForceFlush(std::chrono::microseconds timeout) noexcept
     timeout_steady -= std::chrono::steady_clock::now() - start_timepoint;
   }
 
-  return timeout_steady > std::chrono::steady_clock::duration::zero();
+  // What the wait was for, rather than what is left of the deadline. A session that finishes
+  // between the last check and the last wait takes its notification with it, and the flush it was
+  // holding up has happened.
+  {
+    std::lock_guard<std::recursive_mutex> guard{session_manager_lock_};
```
owent: Is this necessary? when `running_sessions_.empty()` is `true`. `finished_session_counter_.load(std::memory_order_acquire) >= wait_counter` should always be true.

thc1006: You are right, it cannot change the answer. Removed, the return is the counter alone now.

What I checked before dropping it. `ReleaseSession` is the only place a session leaves
`running_sessions_`, and `finished_session_counter_.fetch_add` sits right after the `erase`
under the same `session_manager_lock_`, so the counter is exactly the number removed.
`addSession` is the only place one enters, and its `fetch_add` comes after the insert, so
`wait_counter` is never larger than the number inserted by then. An empty map means removed
equals inserted, which is already at least `wait_counter`. The map holds a
`shared_ptr<Session>`, so a registered session stays alive and its address cannot be reused
under `running_sessions_[session.get()]`, which is the one way the two could have drifted.

I added `ForceFlushReportsSuccessOnceTheSessionIsDone` for the invariant, since nothing was
asserting the return value on that path. It passes on `main` too, where the old
`timeout_steady > 0` happened to agree, so it guards the new predicate rather than
reproducing a bug.

The gRPC half also has direct coverage now, which the earlier revision was missing.
`OtlpMockTraceServiceStub` ran the completion callback inline, so a request was always
finished before `ForceFlush` could see it; holding the callback instead leaves one in flight.
Reverted to `main` the two cases give 10000 ms for a 50 ms deadline, and `flushed == true`
at 100 ms with a request still running.

owent (approved):
LGTM, thanks.

lalitb (approved):
LGTM - Nice focused fix.

Issue #4360, opened 2026-08-05
[BUG] The curl HTTP client can report a request failed and still send it, and can deliver two terminal events


thc1006:
Three places in `Session::SendRequest` where the shared curl client tells an exporter
something other than what it then does. This is `ext/src/http/client/curl/`, the production
client every HTTP exporter runs on, not the embedded test server whose scope was settled on
#4287.

All three are on `main` today.

## 1. A gzip failure reports the request as failed and then sends it anyway

`http_client_curl.cc`, inside the `kGzip` branch, which is compiled under
`ENABLE_OTLP_COMPRESSION_PREVIEW`, so this one reaches only builds with compression enabled:

```cpp
    if (stream != Z_OK)
    {
      if (callback)
      {
        callback->OnEvent(SessionState::CreateFailed, zs.msg ? zs.msg : "");
      }
      is_session_active_.store(false, std::memory_order_release);
    }

    deflateEnd(&zs);
```

There is no `return`. Control falls through to `curl_operation_.reset(new HttpOperation(...))`
and `SendAsync`, so the exporter has already been told the request failed while the request
goes out.

The body is the worrying part. `deflateInPlace` compresses into the caller's buffer, and
only the success path adds `Content-Encoding: gzip` and resizes to `max_size`. A failure
partway through therefore leaves a buffer that may have been written in place, at its
original length, with no encoding header, and that is what gets sent.

What an exporter does with this depends on the exporter, but the shapes are all bad: report
failure and let a retry send the batch a second time, or report failure while the server
accepts something it cannot parse.

## 2. Cancelled and OnResponse can both fire for one operation

In the `SendAsync` completion lambda:

```cpp
        if (operation.WasAborted())
        {
          callback->OnEvent(SessionState::Cancelled, "");
        }

        if (operation.GetSessionState() == SessionState::Response)
        {
          ...
          callback->OnResponse(*response);
        }
```

Two independent `if`s, so an operation aborted after a response arrived delivers both. A
handler that treats either as terminal sees its export settle twice. That is the same shape
as #4338 on the Elasticsearch side, where the fix under review adds a first-writer-wins
guard in the exporter. The guard is needed in every consumer as long as the client can do
this.

## 3. One setup failure produces two terminal events

`HttpOperation::SendAsync` dispatches `ConnectFailed` and returns non-`CURLE_OK`
(`http_operation_curl.cc`). Back in `Session::SendRequest`, `success` is false, so the else
[...]

PR #4363, opened 2026-08-05
[BUG] Report one outcome when a curl session is cancelled after the response arrives

thc1006:
Part of #4360, which lists three problems in the shared curl client's callback contract.
This is only the one with a single obvious answer. The other two need a decision I would
rather hear first, and they are still open on the issue.

## What happens

`Session::SendRequest` passes this completion callback to `SendAsync`:

```cpp
if (operation.WasAborted())
{
  callback->OnEvent(SessionState::Cancelled, "");
}

if (operation.GetSessionState() == SessionState::Response)
{
  ...
  callback->OnResponse(*response);
}
```

Two independent `if`s, and the two conditions are not exclusive. `HttpOperation::Abort`
raises `is_aborted_` and schedules the abort; it does not touch the session state. So a
transfer that has already been answered, and then cancelled, satisfies both, and the handler
is told the request was cancelled and given its response.

The window is not narrow. `PerformCurlMessage` dispatches `Response` and then calls
`Cleanup`, which is where the completion callback runs, and nothing between them is
synchronised against `Abort`. `HttpClient::CancelAllSessions` is what `Shutdown` calls, so
shutting down while a response is arriving lands in it.

## What it costs in tree

`OtlpHttpClient::ResponseHandler` already guards against being finished twice: both
`OnResponse` and the terminal `OnEvent` states take `stopping_` with a compare and exchange,
and only the winner calls `Unbind`. Because the client delivers `Cancelled` first, the
winner is the failure. `Unbind` moves the result callback out, so the response that follows
finds nothing to report and the export is recorded as failed while the server has it.

That is the reason to fix it here rather than in the consumers. A first writer guard cannot
pick the right outcome when it is handed the wrong one first.

## Why the response wins

The state only reaches `Response` when the transfer succeeded: on a non-`CURLE_OK` result
`PerformCurlMessage` dispatches `ConnectFailed` or `SendFailed` first, which moves the state
out of the `Connecting`/`Connected`/`Sending` chain the `Response` transition needs. A
`Response` state therefore means the server answered.

That also makes it the safer of the two to report. Telling a caller its request was
cancelled when the server had already accepted it invites a retry, and for an exporter that
means writing the same batch twice.

`Cancelled` is unchanged for a cancel that lands while the request is still in flight, which
is the case it was written for.

## What this leaves

The `Cancelled` arm can still double up with what `Cleanup` sends. `Cleanup` dispatches
`Cancelled` itself whenever the state is still `Created`, `Connecting`, `Connected` or
`Sending`, and then runs this callback, so an operation aborted in flight and torn down
through `Session::FinishOperation` reports `Cancelled` twice. The second case below measures
exactly that, two notifications for one request, which is why it asserts at least one rather
than one. An aborted operation whose transfer failed is the same shape, the failure state and
then `Cancelled`.

Each of those says one thing twice rather than two different things, and the consumers
collapse them: `OtlpHttpClient` through `stopping_`, the Elasticsearch synchronous handler
through `recordCompletion`, and its asynchronous handler not at all, which is #4338.

Dropping this arm entirely would close those too, but that decides which layer owns the
event rather than correcting an outcome, and it is the third item on #4360. What this change
covers is the one pairing where the two notifications disagree and the wrong one wins.

## Test

`ACancelAfterTheResponseReportsOneOutcome` cancels from inside the handler's `Response`
event. That is not a lucky interleaving: `DispatchEvent` notifies the handler and only then
stores the new state, and the completion callback runs after both, so by the time it is
reached the operation is aborted and its state is `Response`. Cancelling from the handler
also keeps `CancelSession` on the thread that owns the curl handle, which is where the
background loop already is.

The handler counts terminal notifications and the case asserts one.

| | terminal notifications |
| --- | --- |
| `http_client_curl.cc` reverted to `main`, test kept | `Expected equality of these values: 1 ... Which is: 2` |
| this branch | 1 |

`ACancelBeforeTheResponseReportsCancelled` covers the other arm, which nothing in the suite
reached, so the line that reports a cancelled request was never run and Codecov saw the patch
at 50 per cent. The server handler takes `mtx_requests` before it answers, so holding it keeps
a response from racing the cancel, and the abort is torn down through
`Session::FinishOperation` while the transfer is still in flight.

Two `Cancelled` arrive, one from `Cleanup` and one from this callback, so counting them all
would stay green whether or not the arm ran. `GetCurlErrorMessage` never yields an empty
string, so `Cleanup`'s carries a reason and the callback's never does, and the case counts the
empty one. Deleting the arm turns it red at 0 against 1. The total is deliberately not
asserted, since the doubling is the separate question above.

This one characterises rather than reproduces: it passes on `main` too, where the same arm
runs from the first `if`. What it adds is that the line cannot go uncovered again unnoticed.

gcov on the changed lines, from the suite: the response test drives `OnResponse` 24 times, the
`else if` is evaluated 7 times, and its body runs once, where before it ran not at all.

21 of 21 in `curl_http_test`, clang-format clean, and clang-tidy under the CI header filters
reports the same set as `main` for both files.

## Checklist

* [x] Changes in public API reviewed: no signatures change. A handler that used to receive
  `Cancelled` followed by `OnResponse` for one request now receives only `OnResponse`. For the
  OTLP HTTP exporter that turns a shutdown racing a response from a reported failure into the
  result the server actually returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Increment the:
 
 ## [Unreleased]
 
+* [BUG] Report one outcome per request when a curl session is cancelled after
+  the response arrives
+  ([#4363](https://github.com/open-telemetry/opentelemetry-cpp/pull/4363))
+
 * [CODE HEALTH] Enable clang-tidy `modernize-deprecated-headers` and replace
   deprecated C headers (`stdint.h`, `stddef.h`, `stdlib.h`, `string.h`,
   `stdio.h`, `ctype.h`, `limits.h`, `assert.h`) with their C++ equivalents
--- a/ext/src/http/client/curl/http_client_curl.cc
+++ b/ext/src/http/client/curl/http_client_curl.cc
@@ -207,8 +207,8 @@ void Session::SendRequest(
                         http_request_->is_log_enabled_, http_request_->retry_policy_));
   bool success =
       CURLE_OK == curl_operation_->SendAsync(this, [this, callback](HttpOperation &operation) {
-        // Both can hold at once. Cancelling only raises a flag, and the transfer may already
-        // have been answered, so the response is reported and one request finishes once.
+        // Both can hold at once: cancelling only raises a flag, and the transfer may already
+        // have been answered. A request that was answered is reported as answered.
         if (operation.GetSessionState() == opentelemetry::ext::http::client::SessionState::Response)
         {
           // we have a http response
--- a/ext/src/http/client/curl/http_client_curl.cc
+++ b/ext/src/http/client/curl/http_client_curl.cc
@@ -207,12 +207,8 @@ void Session::SendRequest(
                         http_request_->is_log_enabled_, http_request_->retry_policy_));
   bool success =
       CURLE_OK == curl_operation_->SendAsync(this, [this, callback](HttpOperation &operation) {
-        if (operation.WasAborted())
-        {
-          // Manually cancelled
-          callback->OnEvent(opentelemetry::ext::http::client::SessionState::Cancelled, "");
-        }
-
+        // Both can hold at once. Cancelling only raises a flag, and the transfer may already
+        // have been answered, so the response is reported and one request finishes once.
         if (operation.GetSessionState() == opentelemetry::ext::http::client::SessionState::Response)
         {
           // we have a http response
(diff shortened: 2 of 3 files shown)


lalitb (approved):
Thanks.

marcalff (approved):
LGTM, thanks for the fix and tests.

dbarker:
TSAN caught a race in one of the tests from PR. Please see https://github.com/open-telemetry/opentelemetry-cpp/issues/4369 for the logs. This may be an existing bug in the curl client. @thc1006 are you tracking this one?

thc1006:
Yes, tracking it. I put the detail on #4369 so it sits with the logs.

Short of it: `CancelSession` writes to the easy handle through `curl_easy_setopt` while the IO thread is driving that same handle. Any cancel off the IO thread does that, `Shutdown()` included. The new case just makes the window land often enough to get caught.

Fix is built and green locally. I asked over there which of two routes you'd rather have before I send it.

PR #4364, opened 2026-08-05
[SDK] Percent-decode values from OTEL_RESOURCE_ATTRIBUTES

### LINKED ISSUE #1536: Decode values from `OTEL_RESOURCE_ATTRIBUTES`
While using the Go SDK I noticed an inconsistency in how `OTEL_RESOURCE_ATTRIBUTES` baggage values are encoded and decoded. This was due to a difference in the version of the W3C Baggage spec used for [baggage propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#propagation) and the [`OTEL_RESOURCE_ATTRIBUTES` decoding](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#specifying-resource-information-via-an-environment-variable).

https://github.com/open-telemetry/opentelemetry-specification/pull/2670 updates the W3C Baggage spec version used by the Resource spec to require percent-decoding values from  `OTEL_RESOURCE_ATTRIBUTES`.

om7057:
## Summary

Fixes #1536.

The OpenTelemetry resource specification now requires values from `OTEL_RESOURCE_ATTRIBUTES` to be decoded using the W3C Baggage value grammar. While looking into this, I noticed that `OTELResourceDetector::Detect()` was parsing the attributes but storing values exactly as they appeared in the environment.

For example:

```text
OTEL_RESOURCE_ATTRIBUTES="key=hello%20world"
```

was producing the value `hello%20world` instead of `hello world`.

This PR adds a small percent-decoding helper in `sdk/src/resource/resource_detector.cc` and applies it when parsing attribute values.

Malformed escape sequences are preserved rather than causing the attribute to be rejected, which matches the existing parser behavior of being permissive instead of validating attribute syntax.

The change is intentionally scoped to value decoding only. Parsing, trimming, and attribute key handling remain unchanged.

## Testing

Added tests covering:

* Common percent-encoded values (`%20`, `%2C`, `%25`).
* Malformed escape sequences (such as a trailing `%` or invalid hex digits).
* Existing `OTELResourceDetector` behavior to ensure there are no regressions.

The full `resource_test` suite passes locally.

### DIFF
--- a/sdk/src/resource/resource_detector.cc
+++ b/sdk/src/resource/resource_detector.cc
@@ -35,9 +35,10 @@ unsigned char HexDigitValue(unsigned char c) noexcept
 }
 
 // Percent-decodes a value from OTEL_RESOURCE_ATTRIBUTES, per the W3C Baggage value grammar
-// the resource spec defers to (https://github.com/open-telemetry/opentelemetry-specification/pull/2670).
-// A malformed escape sequence is left in the output as-is rather than dropping the attribute,
-// since this parser has never validated attribute syntax.
+// the resource spec defers to
+// (https://github.com/open-telemetry/opentelemetry-specification/pull/2670). A malformed escape
+// sequence is left in the output as-is rather than dropping the attribute, since this parser has
+// never validated attribute syntax.
 std::string PercentDecode(const std::string &value)
 {
   std::string decoded;
@@ -48,9 +49,9 @@ std::string PercentDecode(const std::string &value)
         IsHexDigit(static_cast<unsigned char>(value[i + 1])) &&
         IsHexDigit(static_cast<unsigned char>(value[i + 2])))
     {
-      decoded.push_back(static_cast<char>((HexDigitValue(static_cast<unsigned char>(value[i + 1]))
-                                           << 4) |
-                                          HexDigitValue(static_cast<unsigned char>(value[i + 2]))));
+      decoded.push_back(
+          static_cast<char>((HexDigitValue(static_cast<unsigned char>(value[i + 1])) << 4) |
+                            HexDigitValue(static_cast<unsigned char>(value[i + 2]))));
       i += 2;
     }
     else
--- a/sdk/src/resource/resource_detector.cc
+++ b/sdk/src/resource/resource_detector.cc
@@ -8,6 +8,7 @@
 #include "opentelemetry/semconv/service_attributes.h"
 #include "opentelemetry/version.h"
 
+#include <cctype>
 #include <cstddef>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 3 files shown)


sdk/src/resource/resource_detector.cc
```diff
@@ -21,6 +22,47 @@ namespace resource
...
+// sequence is left in the output as-is rather than dropping the attribute, since this parser has
+// never validated attribute syntax.
+std::string PercentDecode(const std::string &value)
+{
+  std::string decoded;
+  decoded.reserve(value.size());
+  for (std::size_t i = 0; i < value.size(); ++i)
+  {
+    if (value[i] == '%' && i + 2 < value.size() &&
+        IsHexDigit(static_cast<unsigned char>(value[i + 1])) &&
```
marcalff: Ok, so this logic is correct, but personally I would do this:

```
unsigned char value_i_1 = static_cast<unsigned char>(value[i + 1]);
unsigned char value_i_2 = static_cast<unsigned char>(value[i + 2]);
```

and then use

```
std::isxdigit(value_i_1)
```

and

```
unsigned char unescaped_value = HexDigitValue(value_i_1) << 4 + HexDigitValue(value_i_2);
char unescaped = static_cast<char>(unescaped_value);
decoded.push_back(unescaped);
```

A matter of style, the point is to evaluate value[i+1] only once not twice, and to cast only once not twice.

om7057: Ok, got the idea. I've fixed that.

Also, I kept | instead of +. Since + has higher precedence than <<, the expression would be parsed as shifting by 4 + HexDigitValue(value_i_2) rather than shifting by 4 and combining the low nibble. Using | makes the intent explicit and correctly combines the two nibbles after the fixed 4-bit shift.

I also updated the implementation to read and cast the high/low values only once, and replaced the custom hex-digit helper with std::isxdigit.

om7057:
@marcalff addressed the twice-happening cast; please check once.

sdk/src/resource/resource_detector.cc
marcalff: > Also, I kept | instead of +.

Good call.

marcalff (approved):
LGTM, thanks for the fix.

marcalff:
@om7057 

Please do not push new commit to re trigger a failed build.
Maintainers can restart failed builds in CI if necessary.

Also, no need to merge from main every time main changes.

om7057:
> @om7057
> 
> Please do not push new commit to re trigger a failed build. Maintainers can restart failed builds in CI if necessary.
> 
> Also, no need to merge from main every time main changes.

@marcalff Understood. I was just trying to make sure that the PR remains in a mergeable state; I will avoid repeated merges moving forward.

marcalff:
Now that #4376 got in, it's your turn.

PR #4365, opened 2026-08-05
[BUG] Prevent lost condition-variable wakeups during OTLP file, periodic metric, and batch span processor shutdown

mike-nugent-mongodb:
## Changes

The issue in each changed spot is that there is a race between the "shutdown" call and the worker thread, such that the worker thread might not actually shut down. The worker will eventually shut down on the next configured periodic export, but if this could be configured to be relatively long (e.g. 1 minute). There's no workaround for this:

- `ForceFlush` will early-return after shutdown, so one can't use `ForceFlush` to un-stuck a worker thread.
- Calling `Shutdown` more than once can lead to undefined behavior if it tries to join the worker thread from two different threads - at the very least, the return value of `joinable` is undefined, so a second thread trying to check if the worker is still joinable might see `joinable` as false if the thread has been joined but is stuck.

I did not update `CHANGELOG.md` since the change seems small, but let me know if I should. I did not add unit tests as they would need to be invasive and complicated to reliably catch these. AFAIK there is no existing bug/issue documenting this.


sdk/src/metrics/export/periodic_exporting_metric_reader.cc
```diff
@@ -283,6 +283,11 @@ bool PeriodicExportingMetricReader::OnShutDown(std::chrono::microseconds timeout
 {
   if (worker_thread_.joinable())
   {
+    {
+      // Acquiring cv_m_ guarantees that the next time the worker thread checks the wait condition
+      // on cv_ (either from notify below or any other reason) it will see IsShutdown() return true.
+      std::lock_guard<std::mutex> cv_guard{cv_m_};
```
mateenali66, line 294: fence is right, the wait predicate checks `IsShutdown()` under `cv_m_`. the force flush path below has the same store-then-notify shape (the "must not wait for ever" workaround), in scope here or follow-up?

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -1496,6 +1495,13 @@ class OPENTELEMETRY_LOCAL_SYMBOL OtlpFileSystemBackend : public OtlpFileAppender
 
           {
             std::unique_lock<std::mutex> lk(concurrency_file->background_thread_waker_lock);
+            // Even though is_shutdown is atomic, the lock guarantees that either a change to
+            // is_shutdown will be observed, or background_thread_waker_cv will see the notification
+            // at shutdown.
+            if (concurrency_file->is_shutdown.load(std::memory_order_acquire))
```
mateenali66: with the early check gone, shutdown is observed only after one more flush pass, intended? fine if it's deliberate flush-on-shutdown, just checking

mateenali66 (commented):
reader and file exporter fixes look right. per the overlap discussion on #4382: dropping the `batch_span_processor.cc` hunk here would let both PRs land without conflict, the other two fixes aren't covered there at all. also needs a CHANGELOG entry

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -994,6 +994,10 @@ class OPENTELEMETRY_LOCAL_SYMBOL OtlpFileSystemBackend : public OtlpFileAppender
   {
     if (file_)
     {
+      {
+        std::lock_guard<std::mutex> waker_guard{file_->background_thread_waker_lock};
+        file_->is_shutdown.store(true, std::memory_order_release);
```
lalitb, line 999: Could we apply this same guarded store plus notify in `OtlpFileSystemBackend::Shutdown()` too? When there is nothing pending, `ForceFlush()` returns before notifying the worker, so `Shutdown()` can return while the background thread remains parked until `flush_interval`. The destructor now handles this correctly, but callers that keep the client alive after `Shutdown()` still retain the delayed worker.

sdk/src/trace/batch_span_processor.cc
```diff
@@ -376,7 +376,15 @@ bool BatchSpanProcessor::InternalShutdown(std::chrono::microseconds timeout) noe
 
   if (worker_thread_.joinable())
   {
-    synchronization_data_->is_force_wakeup_background_worker.store(true, std::memory_order_release);
+    {
```
lalitb, line 388: Could we drop the `BatchSpanProcessor` changes from this PR? #4382 contains the more complete fix for this path, including `ForceFlush`, completion notification, and regression coverage. Keeping the trace fix there avoids overlapping changes between the two PRs.

mike-nugent-mongodb: Done, I've removed `BatchSpanProcessor` changes.

exporters/otlp/src/otlp_file_client.cc
mike-nugent-mongodb: Done.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
mike-nugent-mongodb: Since it's in the same vein, I've added the change to ForceFlush as well.

exporters/otlp/src/otlp_file_client.cc
mike-nugent-mongodb: The check on `is_shutdown` occurs at almost the same time as previously, the only difference is that the check now occurs while holding `concurrency_file->background_thread_waker_lock` instead of prior to acquiring it. In both cases, if a flush is occurring while `is_shutdown` is set then the code will `break` prior to the next `wait_for`.  Does that make sense or am I misunderstanding?

While looking into this I do notice there is an issue where a `ForceFlush` may not cause an additional flush and make take the full flush interval if `file_->background_thread_waker_cv.notify_all();` happens just before the background thread blocks on `concurrency_file->background_thread_waker_cv.wait_for(lk, flush_interval);`. Fixing this would require a more involved change, so I'd prefer that to be separate. Some details on that: if you just add a variable to cause `background_thread_waker_cv.wait_for` to exit when it's set (e.g. `is_force_background_thread_wake`), the problem is that `ForceFlush` waits on `background_thread_waiter_cv` , so it could miss the notification from the background thread unless there is some additional synchronization (e.g., acquire the `waiter` lock before releasing the `waker` lock).

mike-nugent-mongodb (commented):
I've removed the `BatchSpanProcessor` and added the change to `CHANGELOG` in addition to addressing the comments.

exporters/otlp/src/otlp_file_client.cc
mateenali66: Re-read the loop against the head commit, the only thing between the old check position and the new one is the `BeforeWait()` hook, so no extra flush pass. Agreed.

That hook is the one thing the move does change. On the shutdown path `BeforeWait()` now fires and `AfterWait()` never does, because the `break` is inside the lock scope. Before this PR the shutdown break came ahead of `BeforeWait()`, so the pair stayed balanced. `thread_instrumentation.h` documents the two as bracketing a blocking wait, so an app that flips thread state in `BeforeWait` and restores it in `AfterWait` leaves it set at thread exit. Hoisting the break out of the lock scope keeps both the fence and the pairing:

```cpp
bool shutdown_requested = false;
{
  std::unique_lock<std::mutex> lk(concurrency_file->background_thread_waker_lock);
  shutdown_requested = concurrency_file->is_shutdown.load(std::memory_order_acquire);
  if (!shutdown_requested)
  {
    concurrency_file->background_thread_waker_cv.wait_for(lk, flush_interval);
  }
}
```

then `break` after the `AfterWait()` block. Agreed the `ForceFlush` / `wait_for` race is a separate change.

sdk/src/metrics/export/periodic_exporting_metric_reader.cc
mateenali66: ForceFlush change looks right, and no lock-order problem with it: this path takes `force_flush_m_` then `cv_m_` inside the predicate, and the worker's `cv_m_` lock is scoped to the do-while body so it is released before `CollectAndExportOnce` runs.

One direction is still open though. This fixes the waker side (`cv_`). The completion side is unchanged: `CollectAndExportOnce` CASes `force_flush_notified_sequence_` and calls `force_flush_cv_.notify_all()` holding nothing, so an `OnForceFlush` caller can evaluate the predicate, miss the notify, then wait. That is what the "must not wait for ever" chunked wait is compensating for, so the chunked loop has to stay as long as that store-then-notify is unguarded. Same bucket as the `wait_for` race you described, fine as a follow-up, just noting why that workaround can't come out yet.

mateenali66:
branch is conflicting against main now that #4382 has landed, needs a rebase before CI can give a clean signal here

exporters/otlp/src/otlp_file_client.cc
owent:
```cpp
  std::unique_lock<std::mutex> lk(concurrency_file->background_thread_waker_lock);
  concurrency_file->background_thread_waker_cv.wait_for(lk, flush_interval, [concurrency_file]() {
    return concurrency_file->is_shutdown.load(std::memory_order_acquire);
  });
```

Is these codes more clear? We also use conditional wait_for in other components.

exporters/otlp/src/otlp_file_client.cc
```diff
@@ -1496,6 +1498,13 @@ class OPENTELEMETRY_LOCAL_SYMBOL OtlpFileSystemBackend : public OtlpFileAppender
 
           {
             std::unique_lock<std::mutex> lk(concurrency_file->background_thread_waker_lock);
+            // Even though is_shutdown is atomic, the lock guarantees that either a change to
+            // is_shutdown will be observed, or background_thread_waker_cv will see the notification
+            // at shutdown.
+            if (concurrency_file->is_shutdown.load(std::memory_order_acquire))
+            {
+              break;
+            }
```
owent, line 1511: In my understanding, When `concurrency_file->is_shutdown` is true, `std::fflush` should still need to be called one more time.So Shutdown will still flush all pending records.

mike-nugent-mongodb: I think this is a preexisting issue, as previously `is_shutdown` did not cause an additional `fflush`, so I'd prefer to leave off fixing this.

exporters/otlp/src/otlp_file_client.cc
mike-nugent-mongodb: Good catch, I've moved the break to after `AfterWait`.

As far as using the conditional `wait_for`, I don't think it would be a lot clearer right now because `is_shutdown` has to kept outside the function, so the code would be more like 

```
  std::unique_lock<std::mutex> lk(concurrency_file->background_thread_waker_lock);
  concurrency_file->background_thread_waker_cv.wait_for(lk, flush_interval, [concurrency_file, &is_shutdown]() {
    return is_shutdown = concurrency_file->is_shutdown.load(std::memory_order_acquire);
  });
```
but let me know if this is preferred.

mike-nugent-mongodb (commented):
> branch is conflicting against main now that https://github.com/open-telemetry/opentelemetry-cpp/pull/4382 has landed, needs a rebase before CI can give a clean signal here

I have added merge commits with main, so https://github.com/open-telemetry/opentelemetry-cpp/pull/4382 changes are included. The only conflict was from CHANGELOG.

exporters/otlp/src/otlp_file_client.cc
owent: In the existing codes. `Shutdown()` will calls `ForceFlush` onece, which will wake up the background thread and calls `std::fflush`.

mike-nugent-mongodb: There is no significant behavior change here because `is_shutdown` is set prior to `wait_for`, so both in the old code and new if `Shutdown()` (and hence `ForceFlush()`) is called while the thread is in `wait_for` an additional `std::fflush` will occur. It would be a problem if `is_shutdown` was set in a potential condition of `wait_for` (as suggested in the [other comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/4365#discussion_r3821839517)) so I updated the code comment to mention this.

The new code does *slightly* expand the window for which a shutdown call can fail to `fflush`, but in a minor way: previously, `Shutdown()` called just after `BeforeWait` would cause `fflush` but now it does not (`Shutdown()` called after the  `std::unique_lock<std::mutex>` creation still causes `fflush` because `Shutdown()` now blocks on acquiring the lock).

owent: The behaviour before just follow the other OTLP exporters. Which mean when Shutdown is called, the pending records will be try to be exported once, but the new codes will break and skip `std::fflush` in the codes below( <https://github.com/open-telemetry/opentelemetry-cpp/pull/4365/changes#diff-0a8555076f1a604583ef92e84a10c80eb2806d28624516bea8d71f911cae8d62R1521> ). In my understanding it also breaks the spec ( <https://opentelemetry.io/docs/specs/otel/trace/sdk/#shutdown-1> , "Shutdown MUST include the effects of ForceFlush." ) ""

mike-nugent-mongodb: When the background writer thread is blocked on `background_thread_waker_cv.wait_for` and `Shutdown` is called, the `if` on line 1521 returns false because `is_shutdown` was set on line 1507 prior to when `Shutdown` was called and is not updated again until the next iteration of the loop. If the statement was

```
if(concurrency_file->is_shutdown.load(std::memory_order_acquire))
{
  break;
}
```

then it would be a problem.

owent: Sorry, I misunderstand before, it uses `if (is_shutdown)`, not `if (concurrency_file->is_shutdown)`. LGTM after conflicts are resolved.

owent (approved):

marcalff (approved):
LGTM, thanks for the fix.

PR #4366, opened 2026-08-05
[CONFIGURATION] Build composable samplers from file configuration

### LINKED ISSUE #4028: [SDK] Implement ComposableSampler

ayush-that:
Fixes #4028

Follows up #4334, requested by @dbarker in https://github.com/open-telemetry/opentelemetry-cpp/pull/4334#discussion_r3700394202.

Schema reference https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/tracer_provider.yaml

## Changes

The composable visit methods in `SdkBuilder` were placeholders. `always_on`, `always_off`, and `probability` built the legacy samplers, and `parent_threshold` and `rule_based` ignored their config and sampled everything. They now build the real composable samplers through a recursive visitor wrapped in `CompositeSamplerFactory::Create`, with a new predicate for `rule_based` matching (attribute values and patterns, span kinds, parent).

The composable probability `ratio` is now range checked at parse time, same as #4334, and `ComposableProbabilitySampler` guards against NaN instead of hitting `std::llround(NaN * 2^56)`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,8 @@ Increment the:
 
 ## [Unreleased]
 
-* [CONFIGURATION] Build composable samplers from file configuration. An
-  out-of-range ratio on the composable probability sampler now fails
-  configuration parsing instead of being silently clamped; a NaN ratio
-  passed programmatically to `ComposableProbabilitySampler` falls back to
-  the default 1.0 with a warning.
+* [CONFIGURATION] Add support for the composite sampler configuration
+  (programmatic and from yaml)
   ([#4366](https://github.com/open-telemetry/opentelemetry-cpp/pull/4366))
 
 * [BUG] Report one outcome per request when a curl session is cancelled after
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,16 @@ Increment the:
 
 ## [Unreleased]
 
+* [CONFIGURATION] Build composable samplers from file configuration. An
+  out-of-range ratio on the composable probability sampler now fails
+  configuration parsing instead of being silently clamped; a NaN ratio
+  passed programmatically to `ComposableProbabilitySampler` falls back to
+  the default 1.0 with a warning.
+  ([#4366](https://github.com/open-telemetry/opentelemetry-cpp/pull/4366))
+
 * [BUG] Report one outcome per request when a curl session is cancelled after
   the response arrives
   ([#4363](https://github.com/open-telemetry/opentelemetry-cpp/pull/4363))
-
 * [CODE HEALTH] Enable clang-tidy `modernize-deprecated-headers` and replace
   deprecated C headers (`stdint.h`, `stddef.h`, `stdlib.h`, `string.h`,
   `stdio.h`, `ctype.h`, `limits.h`, `assert.h`) with their C++ equivalents
--- a/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
@@ -47,7 +47,9 @@ struct RuleBasedPredicateOptions
  * RuleBasedPredicate matches a span when every active criteria group in its
  * options matches: parent kind, span kind, an attribute equal to one of the
  * given values, and an attribute matching the included/excluded wildcard
- * patterns (excluded wins).
+ * patterns (excluded wins). Non-string attributes are matched by their string
+ * representation; doubles use the shortest form that reads back as the same
+ * value, so 404.0 matches "404".
  */
 class RuleBasedPredicate : public Predicate
 {
--- a/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
@@ -1,0 +1,78 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "opentelemetry/common/key_value_iterable.h"
+#include "opentelemetry/nostd/string_view.h"
+#include "opentelemetry/sdk/trace/samplers/predicate.h"
+#include "opentelemetry/trace/span_context.h"
+#include "opentelemetry/trace/span_context_kv_iterable.h"
+#include "opentelemetry/trace/span_metadata.h"
+#include "opentelemetry/version.h"
+
+OPENTELEMETRY_BEGIN_NAMESPACE
+namespace sdk
+{
+namespace trace
+{
+
+/**
+ * Matching criteria for RuleBasedPredicate. A criteria group left at its
+ * default is inactive and matches every span.
+ */
+struct RuleBasedPredicateOptions
+{
+  bool match_values = false;
+  std::string values_key;
+  std::vector<std::string> values;
+
+  bool match_patterns = false;
+  std::string patterns_key;
+  std::vector<std::string> included;
+  std::vector<std::string> excluded;
+
+  bool match_parent_none   = false;
+  bool match_parent_remote = false;
+  bool match_parent_local  = false;
+
+  bool match_span_kind_internal = false;
+  bool match_span_kind_server   = false;
+  bool match_span_kind_client   = false;
+  bool match_span_kind_producer = false;
+  bool match_span_kind_consumer = false;
+};
+
+/**
+ * RuleBasedPredicate matches a span when every active criteria group in its
+ * options matches: parent kind, span kind, an attribute equal to one of the
+ * given values, and an attribute matching the included/excluded wildcard
+ * patterns (excluded wins).
+ */
+class RuleBasedPredicate : public Predicate
+{
+public:
+  explicit RuleBasedPredicate(RuleBasedPredicateOptions options);
+
+  bool SpanMatches(
+      const opentelemetry::trace::SpanContext &parent_context,
+      nostd::string_view name,
+      opentelemetry::trace::SpanKind span_kind,
+      const opentelemetry::common::KeyValueIterable &attributes,
+      const opentelemetry::trace::SpanContextKeyValueIterable &links) const noexcept override;
+
+  nostd::string_view GetDescription() const noexcept override;
+
+private:
+  bool ParentMatches(const opentelemetry::trace::SpanContext &parent_context) const noexcept;
+  bool SpanKindMatches(opentelemetry::trace::SpanKind span_kind) const noexcept;
+
+  RuleBasedPredicateOptions options_;
+};
+
+}  // namespace trace
+}  // namespace sdk
+OPENTELEMETRY_END_NAMESPACE
--- a/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
+++ b/sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
@@ -6,11 +6,8 @@
 #include <string>
 #include <vector>
 
-#include "opentelemetry/common/key_value_iterable.h"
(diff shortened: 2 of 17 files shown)


sdk/src/configuration/sdk_builder.cc
```diff
@@ -428,6 +450,325 @@ class ResourceAttributeValueSetter
   std::string name_;
 };
 
+class RuleBasedPredicate : public opentelemetry::sdk::trace::Predicate
```
dbarker: Please move the Predicate into the sdk trace samplers directory so it can be tested independently and used outside the configuration builder.

sdk/src/trace/samplers/composable_probability.cc
```diff
@@ -20,6 +22,11 @@ namespace trace
 
 ComposableProbabilitySampler::ComposableProbabilitySampler(double ratio)
 {
+  if (std::isnan(ratio))
+  {
+    OTEL_INTERNAL_LOG_WARN("[ComposableProbabilitySampler] ratio is NaN, using the default 1.0");
```
dbarker: Please emit the warning for any case where the input value is rejected and default is set.  The ProbabiliySampler and ComposableProbabilitySampler should have the same behavior to invalid ratio input.

CHANGELOG.md
```diff
@@ -15,10 +15,16 @@ Increment the:
 
 ## [Unreleased]
 
+* [CONFIGURATION] Build composable samplers from file configuration. An
```
dbarker: Consider rewording to focus on support for the composite sampler configuration (programmatic and from yaml)

dbarker (commented):
Thanks for the PR! Please see some initial feedback below.

CHANGELOG.md
ayush-that: Reworded

sdk/src/configuration/sdk_builder.cc
ayush-that: I have moved it to the trace samplers directory with direct unit tests.

sdk/src/trace/samplers/composable_probability.cc
ayush-that: Done

sdk/src/trace/samplers/rule_based_predicate.cc
```diff
@@ -0,0 +1,189 @@
...
+                      function_ref<bool(const std::string &)> check) noexcept
+{
+  bool matched = false;
+  attributes.ForEachKeyValue(
+      [&](nostd::string_view attr_key, opentelemetry::common::AttributeValue value) {
+        if (attr_key != key)
+        {
+          return true;
+        }
+        matched = nostd::visit(ValueMatcher(check), value);
```
dbarker: Please use the `VisitVariant` method from `sdk/common/attribute_utils.h` to handle exceptions from `nostd::visit`.

sdk/src/trace/samplers/rule_based_predicate.cc
```diff
@@ -0,0 +1,189 @@
...
+namespace nostd = opentelemetry::nostd;
+
+// Checks every string form of an attribute value; array values match if any item matches.
+class ValueMatcher
+{
+public:
+  explicit ValueMatcher(function_ref<bool(const std::string &)> check) : check_(check) {}
+
+  bool operator()(bool v) const { return check_(v ? "true" : "false"); }
+  bool operator()(int32_t v) const { return check_(std::to_string(v)); }
```
dbarker: Please update the check function to take a string_view to avoid allocating a string for each attribute in the hot path.

For numeric value conversion we can use `std::snprintf` and a stack allocated buffer .

dbarker (commented):
Thanks for the updates. Please see feedback below on the rule based predicate.

sdk/test/trace/composable_sampler_test.cc
```diff
@@ -362,14 +387,117 @@ TEST(ComposableSampler, ThresholdOmittedOverSizeLimit)
...
+  EXPECT_FALSE(Matches(pred, {{"other", "a"}}));
+  EXPECT_TRUE(Matches(pred, {{"service", static_cast<int32_t>(42)}}));
+
+  static const opentelemetry::nostd::string_view items[] = {"x", "a"};
+  EXPECT_TRUE(Matches(
+      pred,
+      {{"service", opentelemetry::nostd::span<const opentelemetry::nostd::string_view>(items)}}));
+}
+
+TEST(RuleBasedPredicate, AttributePatterns)
```
dbarker, line 543: Please add a test that covers matching for all attribute value types. Double is an interesting one because our conversion choice for representation and precision impacts this. 

> 
> Patterns to match against a single attribute. Non-string attributes are matched using their string representation:
>           for example, a pattern of "4*" would match any http.response.status_code in 400-499. For array attributes, if any
>           item matches, it is considered a match.

sdk/test/trace/composable_sampler_test.cc
```diff
@@ -362,14 +387,117 @@ TEST(ComposableSampler, ThresholdOmittedOverSizeLimit)
...
+  RuleBasedPredicateOptions options;
+  options.match_span_kind_server = true;
+  RuleBasedPredicate server_only(std::move(options));
+  EXPECT_TRUE(
+      Matches(server_only, {}, trace_api::SpanContext::GetInvalid(), trace_api::SpanKind::kServer));
+  EXPECT_FALSE(
+      Matches(server_only, {}, trace_api::SpanContext::GetInvalid(), trace_api::SpanKind::kClient));
+}
+
+TEST(RuleBasedPredicate, AttributeValues)
```
dbarker, line 472: Please cover all the attribute value types in this test.

sdk/test/trace/composable_sampler_test.cc
```diff
@@ -362,14 +387,117 @@ TEST(ComposableSampler, ThresholdOmittedOverSizeLimit)
...
+
+  RuleBasedPredicateOptions remote;
+  remote.match_parent_remote = true;
+  RuleBasedPredicate remote_only(std::move(remote));
+  EXPECT_TRUE(Matches(remote_only, {}, MakeParent(true)));
+  EXPECT_FALSE(Matches(remote_only, {}, MakeParent(false)));
+  EXPECT_FALSE(Matches(remote_only, {}));
+}
+
+TEST(RuleBasedPredicate, SpanKindMatching)
```
dbarker, line 441: Please cover all the span kind types in this test.

sdk/test/trace/composable_sampler_test.cc
ayush-that: Done

sdk/test/trace/composable_sampler_test.cc
ayush-that: Done

sdk/test/trace/composable_sampler_test.cc
ayush-that: I have covered all types now.

sdk/src/trace/samplers/rule_based_predicate.cc
ayush-that: Done.

sdk/src/trace/samplers/rule_based_predicate.cc
ayush-that: Done.

sdk/src/trace/samplers/rule_based_predicate.cc
```diff
@@ -0,0 +1,245 @@
...
+      return check_("NaN");
+    }
+    if (std::isinf(v))
+    {
+      return check_(v < 0 ? "-Infinity" : "Infinity");
+    }
+    // Doubles use the shortest of %.15g, %.16g and %.17g that reads back as the same value.
+    char buffer[kValueBufferSize];
+    int length = 0;
+    for (int precision = 15; precision <= 17; ++precision)
```
dbarker: The `double` branch runs up to three `snprintf` calls plus a `strtod` on the hot sampling path. Since this predicate is evaluated for every span and matching attribute, the cost of formatting should be fixed and predictable. It is okay to fix a precision default and document it. 

Please replace the loop with a single `snprintf` at a fixed significant-digit precision using `%.*g`. The precision can be tunable in C++ via `RuleBasedPredicateOptions` (not exposed in YAML), with a default of 6 significant digits (enough to match human-readable values like `404`, `1.5`, `200.0` reliably).

ayush-that: I have replaced the loop with a single snprintf using %.*g.
Let's hope for GitHub actions outage to resolve soon.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -428,6 +444,166 @@ class ResourceAttributeValueSetter
   std::string name_;
 };
 
+class ComposableSamplerBuilder
+    : public opentelemetry::sdk::configuration::SamplerConfigurationVisitor
+{
+public:
+  // NOLINTBEGIN(misc-no-recursion)
```
dbarker, line 451: The yaml schema supports infinite sampler nesting due to the `ExperimentalComposableRuleBasedSampler` containing an array of `ExperimentalComposableRuleBasedSamplerRule` which each may contain a `ExperimentalComposableRuleBasedSampler` (See: [schema/tracer_provider.yaml#L340](https://github.com/open-telemetry/opentelemetry-configuration/blob/52c265e8a59fbf25348be7dfd2cb5c432801358d/schema/tracer_provider.yaml#L340)). 

Without a limit to the sampler nesting depth, a misconfigured yaml file may cause excessive resource usage and latency given that the nested samplers will be run in the `Tracer::StartSpan` method on the hot path. We need some protection for this. 

Please add a comment here to clarify why the recursion is accepted,  a depth limit (10 seems reasonable), and throw a configuration error when the limit is hit.

dbarker (approved):
Nice work! Please see [this comment](https://github.com/open-telemetry/opentelemetry-cpp/pull/4366#discussion_r3736199176) for a follow up PR.  I plan to merge once CI passes.

sdk/src/configuration/sdk_builder.cc
dbarker: This can be a follow up PR and is not blocking

sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
```diff
@@ -0,0 +1,80 @@
...
+  bool match_parent_local  = false;
+
+  bool match_span_kind_internal = false;
+  bool match_span_kind_server   = false;
+  bool match_span_kind_client   = false;
+  bool match_span_kind_producer = false;
+  bool match_span_kind_consumer = false;
+
+  // Significant digits used to format double attribute values for matching.
+  int double_precision = 6;
```
dbarker, line 46: nit - this type should be std::uint32_t and ranged checked before use.  Okay for followup PR

sdk/src/configuration/sdk_builder.cc
ayush-that: Thank you! I will do a follow up PR.

sdk/include/opentelemetry/sdk/trace/samplers/rule_based_predicate.h
ayush-that: Got it.

sdk/src/configuration/sdk_builder.cc
dbarker: Thanks. Please hold on that PR. I'm integrating this PR into #4367 and may address there. The composable sampler builders will move out of the SdkBuilder in that PR so there would be conflicts.

ayush-that: Okie.

PR #4367, opened 2026-08-05
[CONFIGURATION] SDK default component builder libraries and example

dbarker:
Contributes to #4352

Create configuration builder libraries per signal and update SdkBuilder to use them. 

This brings a breaking change to how users create a default registry populated with SDK component builders. 

```cpp
// before this PR
auto registry = std::make_shared<sdk::configuration::Registry>();

// with this PR
auto registry = sdk::configuration::RegistryFactory::Create();
```
## Changes
- creates the per-signal builder libraries
   - `opentelemetry-cpp::configuration_logs_builders` (9846e83d17ccdbfca142256be4c42696cb787631)
   - `opentelemetry-cpp::configuration_metrics_builders` (f3328f32651324b1b7135c2d95f8b85cfee0280a)
   - `opentelemetry-cpp::configuration_trace_builders` (79770686e0ac03b516ead6ef38400b4eb3ef904a)
- creates the registry factory to create a fully populated registry with the default builders
   - `openetelemetry-cpp::configuration_registry_factory`   (d077f5f166c71c15e5eafec35019bba9d22d7476)
- updates the SdkBuilder to get the new builders from the registry (3c1ba5a783695c246de94d43b881ca6e1611035c)
- enables the batch processor tests in the programmatic configuration test suite (a3442f37fc99f78ec2777eb76e82c4793693feba)
- updates the sdk and configuration component install tests (d4c7101779c8e247126ca9159261d774eabd5df6)
- adds new yaml example (sdk-default.yaml) and test for builds without exporters (69b55ec789a4faa12b7271b108a3e3fb0dbc652d, 3d02037b8f013aefcf254acac75b2c7b627aba6e)
- Adds a CompositeSampler to the model and breaks out composable samplers from the SamplerConfigurationVisttor 


examples/configuration/extensions.yaml
```diff
@@ -16,29 +16,14 @@ disabled: false
...
   # Configure span processors.
   processors:
-    # Configure a batch span processor.
-    - batch:
-        # Configure exporter.
-        #
-        # Environment variable: OTEL_TRACES_EXPORTER
-        exporter:
-          # Configure exporter to be zipkin.
-          zipkin:
```
dbarker: Removed due to lack of a Zipkin builder. Also it is not critical for the extensions example.

examples/configuration/CMakeLists.txt
```diff
@@ -69,10 +69,27 @@ if(DEFINED OPENTELEMETRY_BUILD_DLL)
...
   target_link_libraries(
     example_yaml
-    PRIVATE opentelemetry-cpp::configuration opentelemetry-cpp::common
+    PRIVATE opentelemetry-cpp::configuration_registry_factory
+            opentelemetry-cpp::configuration opentelemetry-cpp::common
             opentelemetry-cpp::trace opentelemetry-cpp::logs)
 endif()
 
 if(BUILD_TESTING)
-  add_test(NAME examples.example_yaml COMMAND "$<TARGET_FILE:example_yaml>")
```
dbarker: The example is now tested here against three different yaml files covering all the default builders, extension builders, and composite sampler. These tests now fail on non-zero returns  (thanks to the `--test` arg) or if the example logs an error.

examples/configuration/custom_log_record_exporter.cc
```diff
@@ -20,18 +20,18 @@ opentelemetry::sdk::common::ExportResult CustomLogRecordExporter::Export(
     const opentelemetry::nostd::span<std::unique_ptr<opentelemetry::sdk::logs::Recordable>>
         & /* records */) noexcept
 {
-  OTEL_INTERNAL_LOG_ERROR("CustomLogRecordExporter::Export(): YOUR CODE HERE");
+  OTEL_INTERNAL_LOG_WARN("CustomLogRecordExporter::Export(): YOUR CODE HERE");
```
dbarker: These are now warn so any other error would cause the test to fail.

dbarker:
For smooth integration - this PR depends on several others and review/merge of these should be completed first: 
1. #4368
1. #4374 
1. #4366

examples/configuration/extensions.yaml
dbarker: Change moved to #4374 

Zipkin exporter has been removed from the schema: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/CHANGELOG.md#v100-rc3---2025-12-11

examples/configuration/CMakeLists.txt
```diff
@@ -61,18 +63,6 @@ if(WITH_PROMETHEUS)
                         PRIVATE opentelemetry-cpp::prometheus_exporter_builder)
 endif()
 
-if(DEFINED OPENTELEMETRY_BUILD_DLL)
```
dbarker, line 64: The DLL build is not supported or tested for OTLP or Configuration classes (most are missing the required exports). Removing this to make the example easier to understand by moving the important config targets to the top of the file.

sdk/include/opentelemetry/sdk/configuration/sampler_configuration_visitor.h
```diff
@@ -18,12 +18,7 @@ class ParentBasedSamplerConfiguration;
 class ProbabilitySamplerConfiguration;
 class TraceIdRatioBasedSamplerConfiguration;
 class ExtensionSamplerConfiguration;
-class ComposableAlwaysOffSamplerConfiguration;
```
dbarker, line 21: The `CompositeSampler` was missing from the model and forced `SamplerConfigurationVisitor` to take on knowledge of composable samplers which are a different type (and class hierarchy) from top level samplers. This forced an anti-pattern in the `SdkBuilder` where the `SamplerBuilder` visitor repeated the same implementation for each composable sampler override and the `ComposableSamplerBuilder` visitor had to override methods for the top level (non-composable) samplers to throw an error. Additionally it forced the `SdkBuilder` to provide create methods for each composable sampler. 

The fix is to make the `CompositeSampler` a formal model component that owns a composable sampler config, create a builder interface for it (added to the registry), and separate the sampler and composable sampler visitors.

sdk/include/opentelemetry/sdk/configuration/sdk_builder.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <cstddef>
```
dbarker, line 6: @marcalff

I'm happy with the content of this PR and test coverage, but realize it has grown a bit and changes a lot especially after the sampler/composable sampler visitor refactor and adding CompositeSampler to the model (see [discussion_r3744253342](https://github.com/open-telemetry/opentelemetry-cpp/pull/4367#discussion_r3744253342)).

For context, this PR is step 2 from [#4352](https://github.com/open-telemetry/opentelemetry-cpp/issues/4352#issuecomment-5180573803) and the goal is to instantiate all the sdk component builders in separate trace, logs, metrics builders libraries then use the registered builders in the SdkBuilder. This breaks many direct dependencies on the signal libraries, just leaving the provider level dependencies in the SdkBuilder that will be addressed in step 3. 

I'm interested in your feedback on the overall design and desired scope for review. 

**Questions:** 
1. Any concerns with the high level changes in the PR? 
   - Creating `CompositeSampler` as a formal model component with its own builder in the registry, and the separation of sampler and composable sampler visitors. 
   - Adding a `RegistryFactory` and changing the `Registry` to be empty on construction, which is a breaking change. 
3. Do you have a preference on scope of review? 
   - I can break this into smaller PRs if that would help. A alternative to this one PR may be:
      - PR 1: Add CompositeSampler model and builder to the registry and update tests
         - New interface and visitor classes
         - Update registry and tests
       - PR 2: Trace builders
          - Break out sdk trace builder components from SdkBuilder to the new trace_builders component
          - Use the ComposableSamplerConfigurationVisitor (and update all impacted files)
          - Update SdkBuilder to use builders from the registry for all trace components (except the provider)
          - Add the registry factory and use it in the tests and examples
       - PR 3: Logs builders
           - Break out sdk log builder components from SdkBuilder to the new logs_builders component
       - PR 4: Metrics builders
           - Break out sdk metrics builder components from SdkBuilder to the new metrics_builders component

        Note: PR 2 will still be a big one but needs to occur in one merge to avoid drift with the SdkBuilder. PRs 3 and 4 are fairly narrow after that.

Happy to adjust. I just want to make sure the direction is right before going further.

sdk/include/opentelemetry/sdk/configuration/registry_factory.h
```diff
@@ -0,0 +1,26 @@
...
+namespace sdk
+{
+namespace configuration
+{
+
+class RegistryFactory
+{
+public:
+  // Returns a Registry pre-populated with all default signal builders.
+  static std::shared_ptr<Registry> Create();
```
marcalff, line 21: nit: a Create() method should return a unique_ptr, not a shared_ptr. It is then to the caller to assign this unique_ptr to a shared_ptr if it wants it shared.

sdk/include/opentelemetry/sdk/configuration/sdk_builder.h
marcalff: @dbarker 

> **Questions:**
> 
>     1. Any concerns with the high level changes in the PR?
>        
>        * Creating `CompositeSampler` as a formal model component with its own builder in the registry, and the separation of sampler and composable sampler visitors.

This is definitively the proper design, good work here.

>        * Adding a `RegistryFactory` and changing the `Registry` to be empty on construction, which is a breaking change.
> 

An empty registry is the price to pay to cherry pick which dependencies to allow, while ignoring others.

Any helper such as "please populate all by default" will re-introduce back dependencies on every builder library, which is what we want to avoid.

The key here is to provide a code example for people to follow, for adoption, and we have one.

Providing a registry with everything in it by default is counter productive, it will point users to the wrong direction, only for them to later abandon file configuration when they discover the real link dependencies and decide they can not have that in their application.

>     2. Do you have a preference on scope of review?

Ok, so it is a big review.

Ok to merge in one PR.

marcalff (approved):
Excellent work.

See minor nit, ok to merge.

sdk/include/opentelemetry/sdk/configuration/registry_factory.h
dbarker: Thanks I'll follow up with this one separately. It will impact all the call sites to the SdkBuilder and ConfiguredSdk::Create.

dbarker: The PR to resolve this is up  :) 

https://github.com/open-telemetry/opentelemetry-cpp/pull/4487

PR #4382, opened 2026-08-07
[SDK] Fix lost wakeup in BatchSpanProcessor shutdown/force-flush notify

### LINKED ISSUE #4373: [BUG] BatchSpanProcessor: lost condvar wakeup can stall Shutdown/ForceFlush by up to one schedule_delay
**Describe your environment**

- Platform: macOS 26.5.2, Darwin Kernel Version 25.5.0, arm64
- Build system: Bazel
- opentelemetry-cpp: reproduced on current `main` and release 1.27.0

**Description**

At multiple `cv.notify_all()` sites in `batch_span_processor.cc` the notify is issued
**without holding the mutex that guards the waiter's predicate**, which leaves a lost wakeup window.

A missed wakeup causes a stall of up to one `schedule_delay`, which is quite unexpected for Shutdown
and ForceFlush.

**Steps to reproduce**

I added a stress test in my fork that tries to hammer on these points to reproduce the issue and managed
to get it to trigger on 75/100 runs. 

The test makes it observable by setting `schedule_delay` to 60 min resulting in a lost shutdown wakeup to
park the worker for ~1 h inside `Shutdown()`'s `join()`, and a separate watchdog to catch it when it's stuck.

You can checkout the test in this PR, I have for now opened it against my own fork as I have some questions below. https://github.com/denizariyan/opentelemetry-cpp/pull/1

Please note that the stress test will pass in that PR branch as the branch also includes a fix. Run in this commit to see it fail https://github.com/denizariyan/opentelemetry-cpp/pull/1/changes/9b41cc9ba7924d1e63f0b1eb0e4a2f0b775ab970

```
bazel test //sdk/test/trace:batch_span_processor_test_stress \
  --runs_per_test=100 --nocache_test_results --test_output=errors
```

- `Shutdown()` and `ForceFlush()` should return promptly, regardless of the exact timing between
the notify and the worker parking.
- The stress test linked above passes all runs.

The wakeup is occasionally missed and the operation stalls up to one `schedule_delay`.
With a 60-min `schedule_delay` in the test, the async `Shutdown()` stalls past the 1-min watchdog and
the test `std::abort()`s within a few rounds.

Pre-fix, on my machine the target fails ~75/100 runs.

```
//sdk/test/trace:batch_span_processor_test_stress                        FAILED in 75 out of 100 in 300.1s
  Stats over 100 runs: max = 300.1s, min = 6.1s, avg = 61.7s, dev = 37.7s
```

**Additional context**

The `OnEnd` preemptive-export notify is arguably fine or even better to leave best-effort for better
performance under contention as the next span re-notifies the parked worker anyways, but I'd expect the
`ForceFlush()` and `Shutdown()` paths to be prompt. Especially during process exit, a stall here is very
unpleasant.

From a brief look same behaviour exists for log and metric equivalents of the batch processor too,
but I don't have much of an experience with them so I haven't looked much into those.

I already have a working fix as far as my testing and would be happy to create a PR.

**Questions**

1. Is this lost-wakeup known / considered acceptable given it's not an un-bounded hang?
2. Is the notify without lock pattern intended even for ForceFlush and Shutdown which IMO
are not operations one would call in a hot / performance critical path?

denizariyan:
Fixes #4373  

## Changes

The condvar `notify_all()` calls in `BatchSpanProcessor` were issued without holding the
mutex that guards the waiter's predicate, leaving a lost-wakeup window: if the worker
evaluated its `wait_for` predicate as `false` but had not yet parked, the notify was
missed and it only woke on the next `schedule_delay`.

- Hold `cv_m` when notifying the worker (`ForceFlush` wakeup, `InternalShutdown`) and
  `force_flush_cv_m` when notifying the caller (`NotifyCompletion`), so a concurrent
  predicate-check-then-park cannot miss the wakeup. This makes shutdown and force-flush
  completion prompt.
- The `OnEnd` preemptive-export notify is intentionally left as is since it's sits in the per span
  code path and and a missed wakeup there only delays a threshold-triggered export by up to one
 `schedule_delay` **if traffic goes idle right after**. The next `OnEnd` re-notifies the parked worker
  otherwise.
- As adding the required machinery to catch this kind of an issue in a unit test would be pretty invasive
  and not something that was done before in this project as far as I can see, I added a stress test instead.
  In my testing the issue in both `ForceFlush` and `Shutdown` paths is easily reproduced with this test pre-fix.
  

I provided some more details in the linked issue. #4373 

  * Change is small but I assumed you would want a bug fix noted, so I did. Let me know if you rather keep it out.
  * Added a stress test instead. Reasoning in third point above
  * No changes in public API


denizariyan:
I just realized this PR and https://github.com/open-telemetry/opentelemetry-cpp/pull/4365 **partially** overlap.

https://github.com/open-telemetry/opentelemetry-cpp/pull/4365 also fixes the lost wakeup in the `Shutdown` in `BatchSpanProcessor` along with similar issues in other parts of the system but does not address the lost wakeup in the `ForceFlush` case for `BatchSpanProcessor`. And this PR is roughly the other way around where it fixes the lost wakeup for both `Shutdown` and `ForceFlush` in `BatchSpanProcessor` but does not address the issues in other parts.

Not sure how to best resolve this, maybe I could drop the `Shutdown` part of this PR but then also have to remove the test for it as it will fail on this PR or if we could merge the other PR first I could also rebase on top of it. I guess another option would be to drop the `BatchSpanProcessor` parts from the other PR but I am a believer in _first come first serve_ 😄

denizariyan:
Hey @dbarker, thanks for running the CI! I fixed the reported warnings/errors, could you give it another go when you get a chance?

sdk/src/trace/batch_span_processor.cc
```diff
@@ -98,7 +98,7 @@ void BatchSpanProcessor::OnEnd(std::unique_ptr<Recordable> &&span) noexcept
   size_t buffer_size = buffer_.size();
   if (buffer_size >= max_queue_size_ / 2 || buffer_size >= max_export_batch_size_)
   {
-    // signal the worker thread
+    // Best effort wakeup for worker thread.
```
mateenali66: this window is still open since `buffer_` isn't under `cv_m`, worst case the worker parks for the full schedule delay. worth stating that bound here?

sdk/src/trace/batch_span_processor.cc
```diff
@@ -127,6 +127,7 @@ bool BatchSpanProcessor::ForceFlush(std::chrono::microseconds timeout) noexcept
     if (synchronization_data_->force_flush_pending_sequence.load(std::memory_order_acquire) >
         synchronization_data_->force_flush_notified_sequence.load(std::memory_order_acquire))
     {
+      std::lock_guard<std::mutex> cv_lock(synchronization_data_->cv_m);
```
mateenali66, line 131: this takes `cv_m` while holding `force_flush_cv_m`. safe today because the worker releases `cv_m` before `NotifyCompletion`, but that ordering is load-bearing now, worth a comment pinning it

sdk/src/trace/batch_span_processor.cc
```diff
@@ -309,6 +313,7 @@ void BatchSpanProcessor::NotifyCompletion(
       exporter->ForceFlush(timeout);
     }
 
+    std::lock_guard<std::mutex> lock(synchronization_data->force_flush_cv_m);
```
mateenali66, line 320: with this closed the chunked wait in ForceFlush (the "must not wait for ever" workaround) is no longer needed for correctness, follow-up to simplify?

sdk/test/trace/batch_span_processor_test_stress.cc
```diff
@@ -0,0 +1,140 @@
...
+              << "m at round " << round << ". " << stall_hint << '\n';
+    std::abort();
+  }
+  return result.get();
+}
+
+template <typename Operation>
+void RunWorkerParkRace(const char *operation_name, const char *stall_hint, Operation operation)
+{
+  constexpr int kRounds    = 2000;
```
mateenali66, line 81: how long does a full run take on windows/macos CI? ~6000 thread spawns, might want a smaller CI default with the full count behind an env var

mateenali66 (commented):
fix looks right. re the #4365 overlap: since this PR covers both Shutdown and ForceFlush and carries the regression test, I'd drop the BatchSpanProcessor hunk from #4365 and keep its file exporter and periodic reader fixes, that splits cleanly with no ordering dependency

sdk/src/trace/batch_span_processor.cc
denizariyan: Yes, that is what I meant with _Best effort wakeup_ there, I'll extend the comment with the worst case outcome to make it clear.

sdk/src/trace/batch_span_processor.cc
denizariyan: Yes, it would be nice to simplify this part. I'll create an issue for it in case someone wants to take it up as I am a bit low on bandwidth in the upcoming weeks.

sdk/test/trace/batch_span_processor_test_stress.cc
denizariyan: I pulled the numbers off the last CI run (commit `e9ffb1de`). The threads are short-lived and do essentially no work, so the wall clock stays small:

| Job | ShutdownRacesWorkerPark | ForceFlushRacesWorkerPark | Total |
|---|---|---|---|
| Windows MSVC (maintainer mode) | 0.96s | 1.13s | **2.1s** |
| Windows 2022 vcpkg cxx17 | 0.79s | 0.95s | **1.7s** |
| Windows 2025 vcpkg cxx20 | 0.69s | 0.83s | **1.5s** |
| macOS 14 brew cxx17 | 0.39s | 0.44s | **0.8s** |
| macOS 14 conan cxx17 | 0.39s | 0.39s | **0.8s** |
| Linux gcc 14 (maintainer mode) | 0.43s | 0.45s | **0.9s** |

Given the number of iterations is what provides the means to catch the race and the runtime is negligible I would rather keep a single number for both CI and local runs.

For scale, the existing metrics stress test in the same CI jobs:

| Platform | Metrics stress | This test |
|---|---|---|
| Linux | 13–18s | 0.9s |

mateenali66: that settles it, sub-second is nothing, single constant is fine

sdk/src/trace/batch_span_processor.cc
owent: I think we shouldn't lock any other mutex in ForceFlush.

1. When background is waiting when call `synchronization_data_->cv.wait_for(lk, ...`, it will lock `synchronization_data_->cv_m` and cause deadlock here. And the wakeup notification can not be sent.
2. This will make ForceFlush block for more time than `timeout`.

denizariyan: Good catch! Both of these exist in the current `main` but not with this PR.

> 1. When background is waiting when call `synchronization_data_->cv.wait_for(lk, ...`, it will lock `synchronization_data_->cv_m` and cause deadlock here. And the wakeup notification can not be sent.

On `main`, `DoBackgroundWork()` declares `std::unique_lock<std::mutex> lk(cv_m)` at loop-body scope, so `cv_m` stays held across `Export()` -> `NotifyCompletion()`. If I added the `cv_m` acquisition to `ForceFlush()` alone, you would get exactly what you describe: `NotifyCompletion()` takes `force_flush_cv_m` while the worker holds `cv_m`, `ForceFlush()` takes `cv_m` while holding `force_flush_cv_m`, and that is an ABBA deadlock.

But this PR also scopes that wait here: https://github.com/open-telemetry/opentelemetry-cpp/pull/4382/changes#diff-6f1f4caf95893b12ea6cb9203fdb5a4f383eb6f2be79ae113254986ff4d42463R192-R206

```cpp
{
  std::unique_lock<std::mutex> lk(synchronization_data_->cv_m);
  synchronization_data_->cv.wait_for(lk, timeout, [this] { ... });
  synchronization_data_->is_force_wakeup_background_worker.store(false, std::memory_order_release);
}
```

With `cv_m` released before `Export()`, the only nesting left is `force_flush_cv_m` -> `cv_m` in `ForceFlush()`, and the worker never holds `cv_m` when it wants `force_flush_cv_m`. So no cycle as far as I can see.

> 2. This will make ForceFlush block for more time than `timeout`.

Similarly, with the scoping, the added blocking is bounded by the worker's `cv_m` hold time, which doesn't include the drain/export operations themselves, just atomic load&store + `buffer_.empty()` check. I think this is acceptable, WDYT?

I think now that we can guarantee there would be no lost wakeups, we could re-shape this operation to make it easier to follow and harder to break (maybe by hoisting the wakeup out of the predicate) but I would rather keep the restructuring out of this bugfix PR.

sdk/src/trace/batch_span_processor.cc
denizariyan: Extended in https://github.com/open-telemetry/opentelemetry-cpp/pull/4382/changes/217e5e19c754c6c237be5a22c49ea2dfb642d775

sdk/src/trace/batch_span_processor.cc
owent: Sorry, I may be missing something here. The main branch never tries to lock `synchronization_data_->cv_m` in `ForceFlush`, so there is no ABBA deadlock between the thread calling `ForceFlush` and the background thread.

Limiting the scope of the wait in the background thread does not solve this problem either: the deadlock only occurs while the background thread is waiting on `synchronization_data_->cv`, which still prevents `ForceFlush` from calling `synchronization_data_->cv.notify_all()` to wake it up.

In some scenarios, the timeout and `schedule_delay_millis_` are set to large values, and waiting that long is not acceptable — for example, when gracefully shutting down an application.

denizariyan: You are right about `main`, my wording was sloppy, sorry. `main` has neither lock, so there is no ABBA there. What I meant is that adding the `cv_m` acquisition to `ForceFlush()` on top of `main` without the rest of the changes in this PR would create an issue, because `main` holds `cv_m` across `Export()` -> `NotifyCompletion()`, so the scoping change has to be part of the same PR. That is a statement about why both hunks are here, not about a bug in `main`.

On the second point though, I think (please let me know if I misunderstood what you meant) your concern rests on `wait_for` in `DoBackgroundWork` holding the mutex while parked, and it does not. `cv.wait_for(lk, timeout, pred)` atomically **unlocks** `lk` and blocks, and only reacquires it when it is woken or times out. So while the background thread is parked, `cv_m` is unlocked and `ForceFlush()` acquires it immediately. The mutex is held only while the predicate is being evaluated and while `wait_for` returns, which here is one atomic load&store plus `buffer_.empty()` so that is how long the added acquisition can block for.

The stress test in this PR also proves it, `ForceFlushRacesWorkerPark` runs 2000 rounds with `schedule_delay_millis` set to **10 minutes** and a 1 minute watchdog that aborts the binary if `ForceFlush()` does not return. If taking `cv_m` deadlocked against a parked worker, the very first round would hang. It completes in around 1 second on Linux, macOS and Windows CI. Remove the lock and the same test fails, which is the lost wakeup this PR is fixing. I have also ran this test for over 1k rounds on my local machine without issues.

Your third point is the reason I would like to keep the lock rather than drop it. A large `schedule_delay_millis_` leading to a long wait when the notification is missed is the problem this PR aims to solve. Without the lock the store and the `notify_all()` can land after the worker has evaluated its predicate but before it parks, the notification is lost, and the worker then sleeps the full schedule delay. With a 10 minute delay that is a 10 minute `ForceFlush()`, and the same applies to `InternalShutdown()` on the graceful shutdown path. Taking `cv_m` is what makes the wakeup guaranteed instead of best effort.

Please let me know if this clarifies it or if I misunderstood your concern.

sdk/src/trace/batch_span_processor.cc
```diff
@@ -188,18 +190,21 @@ void BatchSpanProcessor::DoBackgroundWork()
...
-      {
-        return true;
-      }
-
-      return !buffer_.empty();
-    });
-    synchronization_data_->is_force_wakeup_background_worker.store(false,
-                                                                   std::memory_order_release);
+    // Wait for `timeout` milliseconds.
+    {
```
lalitb, line 196: The current ordering looks safe to me: the worker releases `cv_m` before `Export()` can reach `NotifyCompletion()`. Could we add a short comment here saying that `cv_m` must not be held while acquiring `force_flush_cv_m`? This scope is now what prevents an ABBA deadlock, and that is easy to miss in a future refactor.

lalitb:
I believe batch log processor still has the same lost-wakeup pattern in `ForceFlush()`, `NotifyCompletion()`, and
`InternalShutdown()`. The metric shutdown path is already covered by #4365. I’m okay keeping this PR trace-only, but could we open and link a follow-up for the log processor so we don’t lose the remaining cases?

sdk/src/trace/batch_span_processor.cc
denizariyan: Added an explanation in https://github.com/open-telemetry/opentelemetry-cpp/pull/4382/changes/32da17e2f9227e9f88935dd9affeb109e70ca3d3. WDYT?

denizariyan:
> I believe batch log processor still has the same lost-wakeup pattern in `ForceFlush()`, `NotifyCompletion()`, and `InternalShutdown()`. The metric shutdown path is already covered by #4365. I’m okay keeping this PR trace-only, but could we open and link a follow-up for the log processor so we don’t lose the remaining cases?

Yes, I also briefly mentioned this in the linked issue. Created another issue for tracking the issue on the logs side https://github.com/open-telemetry/opentelemetry-cpp/issues/4400

lalitb (approved):
Thanks.

denizariyan:
Thanks @lalitb, fyi I don't have rights to run the CI or merge. Could you do that for me?

sdk/src/trace/batch_span_processor.cc
marcalff: > I think we shouldn't lock any other mutex in ForceFlush.
> 
>     1. When background is waiting when call `synchronization_data_->cv.wait_for(lk, ...`, it will lock `synchronization_data_->cv_m` and cause deadlock here. And the wakeup notification can not be sent.

In my understanding, this is not an issue.

wait_for(), aka pthread_cond_wait(), internally releases the mutex while waiting for the condition, and re acquire the mutex once the condition is signaled, so the mutex is -- not -- held for the entire wait duration.

Acquiring the mutex lock before pthread_cond_signal / broacast is what makes delivering signals reliable.

marcalff (approved):
LGTM, thanks for the fix.

PR #4385, opened 2026-08-08
[CONFIGURATION] Apply metric view attribute filters

### LINKED ISSUE #3546: [SDK] MetricFilter needs a new AttributeProcessor to handle IncludeExclude

iamhaseebn:
Fixes #3546

## Changes

- Build the existing metric view attribute processors from attribute_keys configuration.
- Support include-only, exclude-only, and combined filters.
- Apply exclusions after inclusions, as documented.
- Preserve all attributes when attribute_keys contains neither list.

## Testing

- AppleClang maintainer-mode build with WITH_CONFIGURATION=ON
- Focused CreateAttributesProcessor regression test
- SDK builder and YAML metrics tests: 58 passed
- Full configured suite: 1,046 passed, with 4 repository-disabled tests not run

* [x] CHANGELOG.md updated for non-trivial changes
* [x] Changes in public API reviewed - no public API changes


sdk/src/configuration/sdk_builder.cc
```diff
@@ -1891,16 +1893,43 @@ std::unique_ptr<opentelemetry::sdk::metrics::AggregationConfig> SdkBuilder::Crea
...
 
-  // FIXME-SDK: https://github.com/open-telemetry/opentelemetry-cpp/issues/3546
-  // FIXME-SDK: Need a subclass of AttributesProcessor for IncludeExclude
-  OTEL_INTERNAL_LOG_WARN("IncludeExclude attribute processor not supported, ignoring");
+  FilterAttributeMap excluded;
+  if (model->excluded != nullptr)
+  {
+    for (const auto &key : model->excluded->string_array)
+    {
+      excluded.emplace(key, true);
```
mateenali66: the schema defines included/excluded as patterns (`?`/`*`, the kitchen sink example uses `foo.*`): https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema/common.yaml. exact map lookups won't match those, and `included: [foo.*]` + `excluded: [foo.bar]` can't be expressed by dropping excluded keys from the included map at build time, it needs per-attribute matching like `match(included) && !match(excluded)`. is pattern support planned here or as a follow-up?

sdk/test/configuration/sdk_builder_test.cc
```diff
@@ -723,6 +726,53 @@ TEST(SdkBuilder, CreatePeriodicMetricReader)
   EXPECT_TRUE(captured->exporter != nullptr);
 }
 
+TEST(SdkBuilder, CreateAttributesProcessor)
```
mateenali66, line 819: worth adding wildcard cases once pattern matching lands, `foo.*` included with `foo.bar` excluded is the interesting one

mateenali66 (commented):
exact-key matching works, but the config schema defines these as wildcard patterns, see inline

sdk/src/configuration/sdk_builder.cc
iamhaseebn: Good catch. I updated the processor to evaluate the configured include and exclude patterns per attribute with the existing WildcardMatch helper, so exclusions retain precedence across overlapping patterns.

sdk/test/configuration/sdk_builder_test.cc
iamhaseebn: Added regression coverage for an included foo.* pattern with foo.bar excluded, along with ? matching.

sdk/src/configuration/sdk_builder.cc
mateenali66: predicate and the include_all default look right, and reusing common::WildcardMatch keeps it consistent with the scope configurators. resolved from my side

lalitb (approved):
Thanks.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -216,6 +217,59 @@ using common::WildcardMatch;
 namespace
 {
 
+class IncludeExcludeAttributesProcessor final
```
dbarker: Please move this class to the metrics library in `metrics/view`

dbarker (commented):
Thanks for PR! Please see feedback below.

sdk/src/configuration/sdk_builder.cc
iamhaseebn: Thanks - I've moved `IncludeExcludeAttributesProcessor` into the metrics view library in 2fb8fceb.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
```diff
@@ -162,6 +163,30 @@ class FilteringExcludeAttributesProcessor : public AttributesProcessor
...
+
+  MetricAttributes process(
+      const opentelemetry::common::KeyValueIterable &attributes) const noexcept override;
+
+  bool isPresent(nostd::string_view key) const noexcept override;
+
+private:
+  static bool MatchesAny(const std::vector<std::string> &patterns, nostd::string_view key) noexcept;
+
+  bool include_all_;
```
dbarker: please default to true. The default behavior is to include all attributes.

sdk/src/configuration/sdk_builder.cc
```diff
@@ -1891,16 +1892,31 @@ std::unique_ptr<opentelemetry::sdk::metrics::AggregationConfig> SdkBuilder::Crea
...
+  }
+
+  std::vector<std::string> excluded_patterns;
+  if (model->excluded != nullptr)
+  {
+    excluded_patterns = model->excluded->string_array;
+  }
+
+  return std::make_unique<IncludeExcludeAttributesProcessor>(
+      model->included == nullptr, std::move(included_patterns), std::move(excluded_patterns));
```
dbarker: this currently would set include_all to false if the model->included vector is an empty vector. This would deviate from the schema's default behavior to include all attributes. 

Please change this to 

```cpp
return std::make_unique<IncludeExcludeAttributesProcessor>(
    included_patterns.empty(),
    std::move(included_patterns), std::move(excluded_patterns));
```

sdk/test/configuration/sdk_builder_test.cc
```diff
@@ -723,6 +726,75 @@ TEST(SdkBuilder, CreatePeriodicMetricReader)
...
+    auto processor = builder.CreateAttributesProcessor(model);
+    ASSERT_NE(processor, nullptr);
+    auto filtered = processor->process(wildcard_iterable);
+
+    EXPECT_EQ(filtered.size(), 2u);
+    EXPECT_NE(filtered.find("foo.baz"), filtered.end());
+    EXPECT_NE(filtered.find("question.x"), filtered.end());
+  }
+
+  // An exclude-only configuration retains every key that is not excluded.
```
dbarker, line 863: Please also test all attributes are retained when entering an empty `included` field: 

```yaml
attribute_keys:
  included: []
```

```yaml
attribute_keys:
  included:
```

dbarker (commented):
Thanks for the changes. Please see requested changes below to cover the case when a user enters an empty `included:` field (all attributes must be retained).

sdk/src/configuration/sdk_builder.cc
lalitb: Small nuance: the schema requires at least one item when `included` is present, so both `included: []` and an empty `included:` are technically invalid. `Include-all` is the default when the field is omitted. I’m okay with `included_patterns.empty()` as defensive handling, but I don’t think the schema requires it.

sdk/include/opentelemetry/sdk/metrics/view/attributes_processor.h
iamhaseebn: Updated - `include_all_` now defaults to `true` in 70480b53.

sdk/src/configuration/sdk_builder.cc
iamhaseebn: Updated - the builder now uses `included_patterns.empty()` as defensive handling for empty include lists in 70480b53.

sdk/test/configuration/sdk_builder_test.cc
iamhaseebn: Updated - added defensive regression coverage for both `included: []` and `included:`, confirming all attributes are retained, in 70480b53.

dbarker (approved):
Thanks for the feature!

PR #4388, opened 2026-08-10
[Metrics SDK] Enforce MetricReader-level cardinality limits as a fallback during collection

### LINKED ISSUE #4387: [Metrics SDK] Enforce MetricReader-level cardinality limit as a fallback during collection
## Background

#4188 added `MetricReader::SetCardinalityLimits()` / `GetCardinalityLimit()`, storing per-instrument-type reader-level cardinality limits, but explicitly left them unenforced (see the TODO in `MetricReader::SetCardinalityLimits()`). The initial attempt to enforce them by resolving the reader's limit directly into the shared per-view metric storage was reverted during review, since it broke per-reader semantics: with two readers sharing one instrument's storage (limit 10 and limit 1000), using the reader's limit for shared storage capacity let the low-limit reader see far more series than configured (see https://github.com/open-telemetry/opentelemetry-cpp/pull/4188#discussion_r3506330000 area).

Per the spec, cardinality limit precedence is View > Reader > SDK default. Today only View-level (`AggregationConfig::cardinality_limit_`) and the SDK default are enforced.

## Proposal

- Shared recording storage (the `AttributesHashMap` inside `SyncMetricStorage`/`AsyncMetricStorage`) should be sized at the *max* cardinality limit across all `MetricReader`s currently attached, when the view has no explicit limit of its own, so no reader loses data purely because storage was capped too low for it.
- Each reader's own (possibly stricter) limit should then be re-applied to just its own output during collection (`TemporalMetricStorage::buildMetrics`), so a stricter reader is still capped to what it configured, without affecting other readers sharing the same storage.

I'd like to work on this — will follow up with a PR shortly.

om7057:
## Summary

Fixes #4387

Follow-up to #4188 and #4314. `MetricReader::SetCardinalityLimits()` parsed and stored reader-level cardinality limits but never enforced them (see the TODO that was in `MetricReader::SetCardinalityLimits()`). The initial attempt to enforce them in #4188 resolved the reader's limit directly into shared per-view storage, which was reverted during review because it broke per-reader semantics: with two readers sharing one instrument's storage at different limits, a reader could either lose data it should have gotten, or force every other reader down to its own stricter limit.

## Changes

This enforces the View > Reader > SDK default precedence at both ends of the pipeline:

- **Recording storage sizing** (`SyncMetricStorage`/`AsyncMetricStorage`): when a view has no explicit `AggregationConfig` cardinality limit, the shared `AttributesHashMap` used for recording is now sized at the *max* cardinality limit across all `MetricReader`s currently attached, so no reader loses data purely because storage was capped too low for it. When a view does set an explicit limit, storage sizing is unchanged.
- **Per-collector re-capping** (`TemporalMetricStorage::buildMetrics`): each reader's own (possibly stricter) limit is re-applied to just its own collected output, via `AttributesHashMap`'s existing overflow mechanism (`Set()` already redirects new keys beyond capacity into the overflow point). This means a stricter reader is still capped to what it configured, without affecting other readers sharing the same storage.

`SyncMetricStorage`/`AsyncMetricStorage` gained a new constructor parameter (`recording_cardinality_limit`) for this; the original constructor signature is preserved as a delegating overload so any external caller is unaffected.

## Tests

- New regression test `MetricCollectorTest.ReaderCardinalityLimitFallbackWithMultipleReaders`: two readers (limits 3 and 50, no view-level limit) share one counter; the low-limit reader is capped with overflow present, the high-limit reader sees every recorded series with no data loss and no overflow. Confirmed this test fails against the pre-fix code (low-limit reader saw all 10 points, no overflow).
- Full existing metrics test suite (`metric_reader_test`, `metric_collector_test`, `sync_instruments_test`, `async_instruments_test`, `async_metric_storage_test`, `bound_sync_instruments_test` under `WITH_METRICS_BOUND_INSTRUMENTS_PREVIEW`) passes unchanged.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Unit tests added
- [x] Public API changes reviewed (new constructor parameter, back-compat overload preserved)


sdk/src/metrics/state/temporal_metric_storage.cc
lalitb, line 178: Could we check how overflow values are combined here?

When several attribute sets exceed the reader limit, `AttributesHashMap::Set()` sends each one to the overflow key, but it replaces the previous overflow aggregation. For example, with 10 counter series and a limit of 3, the overflow point can contain only one of the other 7 values.

This silently loses metric data. I think this path needs to merge with the existing overflow aggregation. Please also update the test to check that the total counter value is still 10, not only that an overflow point exists.

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -114,9 +114,17 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
...
-                                                : kAggregationCardinalityLimit));
+  //
+  // When no view-level limit is configured (aggregation_config_ == nullptr), fall back to this
+  // specific collector's own MetricReader-level limit rather than a flat SDK default. The
+  // underlying recording storage may be sized larger (to avoid losing data for readers with a
+  // higher limit, see SyncMetricStorage/AsyncMetricStorage), so capping merged_metrics here to
+  // this collector's own limit re-applies the View > Reader > SDK default precedence per reader:
+  // AttributesHashMap::Set() below routes anything beyond this capacity into the overflow point,
+  // without affecting what other collectors of the same storage see.
+  std::unique_ptr<AttributesHashMap> merged_metrics(new AttributesHashMap(
```
lalitb, line 164: The single-reader delta fast path above returns before reaching this new limit.

For example, one delta reader with a limit of 3 can still export up to 2000 attribute sets because the fast path returns the recording map directly.

Could we apply the reader limit in that fast path too, and add a test with one delta reader?

sdk/src/metrics/meter.cc
```diff
@@ -83,6 +87,32 @@ std::ostream &operator<<(std::ostream &os,
...
+// because the shared cap was sized for a stricter reader. Each reader's own (possibly lower)
+// limit is then re-applied to just its own output during collection; see
+// TemporalMetricStorage::buildMetrics().
+std::size_t ResolveRecordingCardinalityLimit(
+    const opentelemetry::sdk::metrics::AggregationConfig *aggregation_config,
+    opentelemetry::nostd::span<std::shared_ptr<opentelemetry::sdk::metrics::CollectorHandle>>
+        collectors,
+    opentelemetry::sdk::metrics::InstrumentType instrument_type)
+{
+  if (aggregation_config)
```
lalitb: Does a non-null `aggregation_config` always mean that the view explicitly set a cardinality limit?

A view may have an aggregation config only to set histogram boundaries. Its cardinality limit then contains the default value, but this code treats it as an explicit view-level limit and ignores the reader limit.

I think we need to distinguish “limit not set” from “limit explicitly set to 2000” before applying the View > Reader > SDK precedence.

lalitb (changes requested):
Thanks for working on this. The overall approach makes sense, but I found a few cases where metric values can be lost or the configured reader limit is not applied. I think these should be fixed before merging, so I am requesting changes.

sdk/src/metrics/state/temporal_metric_storage.cc
om7057: Thanks for tracing. `Set()`'s overflow branch did `hash_map_[GetOverflowAttributes()] = std::move(aggr)`, an unconditional overwrite, so a second (and any later) over-capacity attribute set discarded whatever value was already accumulated in the overflow bucket rather than adding to it. `GetOrSetDefault()`'s overflow path was already correct (it returns the existing overflow aggregation object so the caller accumulates onto it), so `Set()` was the odd one out.

Fixed by adding `SetOverflowMerged()`, which merges into the existing overflow entry via `Aggregation::Merge()` when one is already present, and only assigns directly when there isn't one yet. Both `Set()` overloads now route through it.

Extended `ReaderCardinalityLimitFallbackWithMultipleReaders` to assert the *total* counter value across all points (real + overflow) still equals the number of `Add(1, ...)` calls, not just that an overflow point exists. Confirmed this new assertion fails against the pre-fix code.

sdk/src/metrics/state/temporal_metric_storage.cc
om7057: You're right, and tracing it further: this was actually a symptom of a bug in `ResolveRecordingCardinalityLimit()` in `meter.cc`, not something needing a separate fix in the fast path itself.

That function started `max_limit` at `kAggregationCardinalityLimit` (2000) and took the max with each attached reader's configured limit. With a single delta reader whose own limit is 3, `max(2000, 3)` still returned 2000, so the *recording* storage itself (`SyncMetricStorage`'s `attributes_hashmap_`) was sized at 2000 regardless of the reader's stricter limit, and the fast path (which reads straight from that recording storage) faithfully reported everything in it.

Removed the floor: the max is now taken purely across attached readers' actual limits (falling back to the SDK default only when there are no readers at all yet). With that fixed, a single reader's recording storage is sized exactly at that reader's own limit, so the fast path is correctly capped without needing to touch it directly.

Added `ReaderCardinalityLimitFallbackSingleDeltaReader`: one delta reader, limit 3, 10 unique attribute sets. Confirmed this fails against the pre-fix code (reports up to 2000).

sdk/src/metrics/meter.cc
om7057: Correct, it doesn't, and this was a real gap: `SdkBuilder::AddView()` calls `CreateAggregationConfig(stream->aggregation, ...)` for the `aggregation` block (e.g. histogram boundaries), which can produce a non-null config whose `cardinality_limit_` was never touched and is just sitting at the compiled-in default (2000). My code was treating that as "the view explicitly wants 2000," which silently defeated the reader-level fallback for exactly the case you describe.

Added `AggregationConfig::cardinality_limit_explicit_` (defaults to `true`, so any existing programmatic/test caller that constructs a config directly keeps its prior behavior unchanged). `SdkBuilder::AddView()` is the only place that has the information needed to know when a config was built without `aggregation_cardinality_limit`, so it now sets this to `false` right after `CreateAggregationConfig()`, and back to `true` if `aggregation_cardinality_limit` is also set on the same stream. `meter.cc` and `temporal_metric_storage.cc` now both check `aggregation_config && aggregation_config->cardinality_limit_explicit_` instead of just null-checking.

Added two `SdkBuilder` tests: a histogram view with explicit boundaries and no cardinality limit ends up `cardinality_limit_explicit_ == false`, and the same view with `aggregation_cardinality_limit` also set ends up `true` with the configured value.

om7057:
> Thanks for working on this. The overall approach makes sense, but I found a few cases where metric values can be lost or the configured reader limit is not applied. I think these should be fixed before merging, so I am requesting changes.

@lalitb Thank you for thorough review, I've addressed the requested changes, please go through them once.

om7057:
@lalitb I've resolved the merge conflicts as well, updating with main.
Waiting for your approval!

om7057:
@lalitb small reminder on this, I've rebased the branch so no merge conflicts anymore.

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
```diff
@@ -40,7 +40,15 @@ class AggregationConfig
...
   size_t cardinality_limit_;
-  virtual ~AggregationConfig() = default;
+  // Whether cardinality_limit_ reflects an intentionally-configured value, as opposed to
+  // just the compiled-in default left untouched because this config was built for some other
+  // reason (e.g. histogram boundaries). Defaults to true so any caller that constructs an
+  // AggregationConfig directly (tests, programmatic API) keeps prior behavior: the config's
+  // cardinality_limit_ is always honored as-is. SdkBuilder::AddView() is the one place that
+  // knows when a config was synthesized without an explicit `aggregation_cardinality_limit`,
+  // and sets this to false there so a MetricReader-level fallback can apply instead.
+  bool cardinality_limit_explicit_ = true;
```
lalitb: This still treats the default limit as explicit for programmatic views.

For example, a user may create `HistogramAggregationConfig()` only to set histogram boundaries. Its default limit of 2000 is then treated as a view limit, so a reader limit of 100 is ignored.

Could we represent “not specified” for programmatic configs too, and add a test for this case?

sdk/src/metrics/meter.cc
```diff
@@ -541,14 +581,16 @@ std::unique_ptr<SyncWritableMetricStorage> Meter::RegisterSyncMetricStorage(
         else
         {
           WarnOnDuplicateInstrument(GetInstrumentationScope(), storage_registry_, view_instr_desc);
+          auto recording_cardinality_limit = ResolveRecordingCardinalityLimit(
+              view.GetAggregationConfig(), ctx->GetCollectors(), view_instr_desc.type_);
```
lalitb, line 585: This limit is calculated only when the storage is first created.

For example, if the instrument is created with a reader limit of 3 and a reader with limit 50 is added later, the existing storage remains capped at 3. The new reader cannot recover the series that were already merged into overflow.

Could we update existing storage when readers or their limits change, or make those limits immutable after instrument creation? Please also add a test where the instrument exists before the higher-limit reader is added.

lalitb:
> @lalitb small reminder on this, I've rebased the branch so no merge conflicts anymore.

Thanks for addressing the earlier findings. I found two remaining cases where the reader-level limit is not applied correctly. I think these should be fixed before merging.

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
om7057: Good catch, fixed. Added a sentinel value so the constructor can tell "no limit given" apart from "limit explicitly passed", without needing `<optional>`. Now `HistogramAggregationConfig()` correctly comes out as not-explicit, while `HistogramAggregationConfig(500)` still does. Added a test.

sdk/src/metrics/meter.cc
om7057: Confirmed, this is real since AddMetricReader can be called anytime. Resizing live storage safely felt too risky for this PR, so I went with your other option: documented it as a known limitation on AddMetricReader, and added a test pinning the behavior. Happy to open a follow-up for the dynamic-resize version if you want that tracked separately.

om7057:
> > @lalitb small reminder on this, I've rebased the branch so no merge conflicts anymore.
> 
> Thanks for addressing the earlier findings. I found two remaining cases where the reader-level limit is not applied correctly. I think these should be fixed before merging.

@lalitb thanks for the thorough review. I've pushed a fix for the two remaining findings:

Please check once more!

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
lalitb: Thanks, the constructor case is fixed now. I found a separate compatibility case where direct assignment to the public `cardinality_limit_` field leaves `cardinality_limit_explicit_` false, so I will raise that separately.

sdk/src/metrics/meter.cc
lalitb: Thanks. Looking through the existing lifecycle, late reader attachment is supported for reader-specific collection behavior, but it is not thread-safe and does not rebuild instrument storage. I am okay treating a later higher recording capacity as a documented limitation.

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
```diff
@@ -40,13 +52,21 @@ class AggregationConfig
   }
 
   size_t cardinality_limit_;
```
lalitb, line 73: This public field can still be assigned directly, but that no longer marks the limit as explicit.

For example:

```cpp
  HistogramAggregationConfig config;
  config.cardinality_limit_ = 100;
```

`cardinality_limit_explicit_` remains false, so the reader fallback can override the configured limit.  Could we keep these two values from getting out of sync, or preserve the existing direct-assignment behavior another way?

sdk/src/metrics/state/temporal_metric_storage.cc
```diff
@@ -114,9 +114,21 @@ bool TemporalMetricStorage::buildMetrics(CollectorHandle *collector,
...
+  // histogram boundaries; see cardinality_limit_explicit_), fall back to this specific
+  // collector's own MetricReader-level limit rather than a flat SDK default. The underlying
+  // recording storage may be sized larger (to avoid losing data for readers with a higher
+  // limit, see SyncMetricStorage/AsyncMetricStorage), so capping merged_metrics here to this
+  // collector's own limit re-applies the View > Reader > SDK default precedence per reader:
+  // AttributesHashMap::Set() below routes anything beyond this capacity into the overflow point,
+  // without affecting what other collectors of the same storage see.
+  const bool has_explicit_view_limit =
+      aggregation_config_ && aggregation_config_->cardinality_limit_explicit_;
+  std::unique_ptr<AttributesHashMap> merged_metrics(new AttributesHashMap(
```
lalitb, line 164: This applies the reader limit only in the normal merge path. The single-reader delta fast path above returns `delta_metrics` directly, before reaching this code.

For example, if storage was created with capacity 2000 and a delta reader with limit 3 is attached later, that fast path can still emit more than 3 attribute sets.

Could we apply the reader fallback in the delta fast path too, and add a regression test for a stricter reader attached after instrument creation?

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
```diff
@@ -278,6 +308,8 @@ class SyncMetricStorage : public MetricStorage, public SyncWritableMetricStorage
   InstrumentDescriptor instrument_descriptor_;
   // hashmap to maintain the metrics for delta collection (i.e, collection since last Collect call)
   const AggregationConfig *aggregation_config_;
+  // Capacity used to (re)size attributes_hashmap_. See the constructor comment above.
+  const std::size_t recording_cardinality_limit_;
```
lalitb, line 317: The bound-instrument admission path does not use this resolved recording limit. `ResolveCardinality()` still uses `aggregation_config_->cardinality_limit_`.

These values can now differ when the limit comes from a reader. For example, with a reader limit of 3 and no explicit view limit, bound keys are admitted using 2000 even though the recording hashmap is capped at 3.

Could we use `recording_cardinality_limit_` in `ResolveCardinality()` so bound and unbound recording use the same effective limit?

lalitb:
@om7057 -   Thanks, I checked the latest update.

The original programmatic-constructor issue is fixed. I am also okay with treating a higher limit from a late-added reader as a documented limitation, since existing instrument storage is not rebuilt dynamically.

During the recheck, I found three related cases in the new implementation: direct assignment to the public limit field, the single-reader delta fast path, and bound-instrument admission. I have left separate inline comments for them.

Cardinality-limit handling crosses recording storage, per-reader collection, overflow, and optional bound-instrument paths, so small changes can affect several code paths. I think changes in this area need careful review and focused regression coverage.

Once those are addressed, I will check it again.

sdk/include/opentelemetry/sdk/metrics/aggregation/aggregation_config.h
om7057: Fixed. Made both fields private with a `SetCardinalityLimit()` setter that sets them together; direct assignment is no longer possible.

sdk/src/metrics/state/temporal_metric_storage.cc
om7057: Fixed. The fast path now re-caps through a hashmap sized to the collector's effective limit when the recorded size exceeds it, mirroring the merge path below. Added a test that fails without this fix (10 points, no overflow, instead of capped at 3).

sdk/include/opentelemetry/sdk/metrics/state/sync_metric_storage.h
om7057: Fixed. ResolveCardinality() now reads recording_cardinality_limit_ instead of the aggregation config's own limit, so bound and unbound admission always agree.

om7057:
> @om7057 - Thanks, I checked the latest update.
> 
> The original programmatic-constructor issue is fixed. I am also okay with treating a higher limit from a late-added reader as a documented limitation, since existing instrument storage is not rebuilt dynamically.
> 
> During the recheck, I found three related cases in the new implementation: direct assignment to the public limit field, the single-reader delta fast path, and bound-instrument admission. I have left separate inline comments for them.
> 
> Cardinality-limit handling crosses recording storage, per-reader collection, overflow, and optional bound-instrument paths, so small changes can affect several code paths. I think changes in this area need careful review and focused regression coverage.
> 
> Once those are addressed, I will check it again.

@lalitb all three addressed: AggregationConfig fields are now private behind SetCardinalityLimit(), the delta fast path re-caps to the collector's limit, and bound/unbound admission both use recording_cardinality_limit_. Replied inline on each with specifics. Added a regression test for the fast-path bug that I verified fails without the fix. Ready for another look whenever you have time.

om7057:
@lalitb just a reminder!

om7057:
@lalitb just a small ping on this one!

om7057:
@lalitb All the feedback above has been addressed as of the last round of fixes on 2026-08-19, with tests added for each case. Could you take another look when you have a chance? Also merged main in to resolve the conflicts from the sdk_builder.cc split.

om7057:
@lalitb just a small ping!

Issue #4391, opened 2026-08-10
[BUG] The curl client mutates and frees easy handle resources before removing the handle from the multi handle


thc1006:
**Describe your environment** `main` at f6e48180, libcurl 8.14.1. This is a code reading report, not a crash report. I have not seen any of it misbehave at runtime, and I say where that matters below.

Three places in the curl client touch an easy handle, or free something the handle still points at, before the handle has been taken out of the multi handle, and the `CURLMcode` that would say whether the removal worked is dropped on the floor.

### Teardown resets the handle before it is removed

`HttpOperation::Cleanup()` clears `CURLOPT_PRIVATE` and resets the handle, and only then hands it over to be removed:

```cpp
      if (curl_resource_.easy_handle != nullptr)
      {
        curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_PRIVATE, NULL);   // :556
        curl_easy_reset(curl_resource_.easy_handle);                           // :557
      }
      session->GetHttpClient().ScheduleRemoveSession(...);                     // :559
```

`curl_multi_remove_handle` does not run until the IO thread gets to `doRemoveSessions`, at `http_client_curl.cc:819`. On a cancel the transfer can still be active at `:556`, and [curl_easy_setopt](https://curl.se/libcurl/c/curl_easy_setopt.html) says:

> Changing options with curl_easy_setopt() while a transfer is still in progress may cause undefined and undesired behavior.

### The header list is freed before the removal too

```cpp
      if (nullptr != removing_handle.second.headers_chunk)
      {
        curl_slist_free_all(removing_handle.second.headers_chunk);             // :816
      }

      curl_multi_remove_handle(multi_handle_, removing_handle.second.easy_handle);  // :819
      curl_easy_cleanup(removing_handle.second.easy_handle);                        // :820
```

[CURLOPT_HTTPHEADER](https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html):

> When this option is passed to curl_easy_setopt, libcurl does not copy the entire list so you must keep it around until you no longer use this handle for a transfer before you call curl_slist_free_all on the list.

At `:816` the handle is still attached, so the list is freed while it is arguably still in use for the transfer. Swapping `:816` and `:819` costs nothing.

### The removal result is not checked

`:819` ignores its `CURLMcode` and `:820` cleans the handle up regardless. If the removal ever fails, that frees a handle the multi stack still knows about. The retry path has the same shape at `:862`:

[...]

Issue #4390, opened 2026-08-10
[BUG] Cancelling from a Created or Connecting event loses the cancel and leaves the operation unfinished


thc1006:
**Describe your environment** `main` at f6e48180, Ubuntu 24.04, gcc 14, libcurl 8.14.1, CMake. The ordering below has no platform conditionals in it.

`HttpOperation::SendAsync()` hands the first event to the user before the operation is ready to be cancelled, and a cancel that lands in that window is thrown away. The request is then neither sent nor completed, and the caller blocks when it tries to finish the session.

```cpp
  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_PRIVATE, session);

  DispatchEvent(...SessionState::Connecting);                    // :1454  user handler runs here
  is_finished_.store(false, std::memory_order_release);
  is_aborted_.store(false, std::memory_order_release);           // :1456  clears the cancel
  is_cleaned_.store(false, std::memory_order_release);

  async_data_->session = session;                                // :1459  route published here
  ...
  async_data_->result_promise = std::promise<CURLcode>();        // :1462  promise created here
  ...
  session->GetHttpClient().ScheduleAddSession(...);              // :1467
```

Say the handler calls `Session::CancelSession()` from that `Connecting` event, which the public API allows. `Abort()` at `:1509` raises `is_aborted_`, then checks `async_data_->session` at `:1516`, which is still the `nullptr` written at `:1440`, so no abort is scheduled. `CleanupSession()` at `:379` does run, and it removes the session from `sessions_`.

`SendAsync` then resumes and undoes the rest of it. Line `:1456` puts `is_aborted_` back to false. `ScheduleAddSession` at `:641` erases the pending abort at `:647`. When the IO thread reaches `doAddSessions`, the session is no longer in `sessions_`, so the lookup at `:735` misses and it `continue`s at `:738` without ever adding the handle to the multi handle.

What is left is a live promise with nobody to fulfil it. `HttpOperation::Finish()` waits on it at `:516`, and `~HttpOperation` waits on it at `:491` when the state is `Connecting`.

**Steps to reproduce** Add this to `ext/test/http/curl_http_test.cc` on unmodified `main`. It uses `TerminalCountingHandler`, which is already there:

```cpp
TEST_F(BasicCurlHttpTests, CancelFromConnecting)
{
  auto session_manager = std::make_shared<http_client::curl::HttpCurlClientFactory>()->Create();
[...]

Issue #4389, opened 2026-08-10
[BUG] resetMultiHandle deadlocks on sessions_m_ while recovering from a curl multi error


thc1006:
**Describe your environment** `main` at f6e48180, Ubuntu 24.04, gcc 14, libcurl 8.14.1, CMake. Nothing here is platform conditional, so it should read the same everywhere.

`HttpClient::resetMultiHandle()` takes `sessions_m_` and then calls two things that take it again on the same thread. `sessions_m_` is a plain `std::mutex`, declared at `ext/include/opentelemetry/ext/http/client/curl/http_client_curl.h:372`, so relocking it from the owning thread is undefined behaviour, and in practice the IO thread stops there for good.

```cpp
void HttpClient::resetMultiHandle()                                 // http_client_curl.cc:884
{
  std::list<std::shared_ptr<Session>> sessions;
  std::lock_guard<std::mutex> session_lock_guard{sessions_m_};      // :887  taken here
  ...
  for (auto &session : sessions)
  {
    session->CancelSession();                                       // :901  -> CleanupSession -> :383
    session->FinishOperation();
  }

  doRemoveSessions();                                               // :905  -> :800
  ...
}
```

`Session::CancelSession()` calls `http_client_.CleanupSession(session_id_)` at `:251`, and `CleanupSession` opens with `std::lock_guard<std::mutex> lock_guard{sessions_m_}` at `:383`. `doRemoveSessions()` takes the same mutex at `:800`, on the first pass of a `do` loop, so it does not depend on there being anything left to remove. Either call is enough on its own.

**Steps to reproduce** The trigger is `curl_multi_perform` returning something other than `CURLM_OK` at `:472`, and I could not find a way to provoke that from outside the library. So rather than guess, I forced the branch once on otherwise unmodified `main`:

```cpp
static std::atomic<int> inject_once{0};
if (mc != CURLM_OK || inject_once.fetch_add(1) == 0)
{
  fputs("INJECT enter", stderr);
  self->resetMultiHandle();
  fputs("INJECT return", stderr);
}
```

`curl_http_test --gtest_filter='*SendGetRequest*'` prints `INJECT enter`, never prints `INJECT return`, finishes zero cases and exits 124 under a 90 second `timeout`. A core taken at 15 seconds has the IO thread here:

```text
Thread 3 (LWP 853585):
#1  pthread_mutex_lock () from /lib/x86_64-linux-gnu/libc.so.6
[...]

Issue #4369, opened 2026-08-05
TSAN detected race in `BasicCurlHttpTests.ACancelBeforeTheResponseReportsCancelled`


dbarker:
The following TSAN warning was detected in CI after merging main from a7da83e3798348b9da833eb828f5096944025aef

**Warning Summary:** 

````
SUMMARY: ThreadSanitizer: data race (/home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/_main/bazel-out/k8-fastbuild-tsan/bin/ext/test/http/curl_http_test+0xf8bd7) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6) in setopt_long

SUMMARY: ThreadSanitizer: data race (/home/runner/.cache/bazel/e6b8d6759295e14b76bc8cb98b604748/execroot/_main/bazel-out/k8-fastbuild-tsan/bin/ext/test/http/curl_http_test+0xf8c71) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6) in setopt_long
````
**Detail logs:** 

<details><summary>TSAN detailed job logs (click to view)</summary>

```
[ RUN      ] BasicCurlHttpTests.ACancelBeforeTheResponseReportsCancelled
==================
WARNING: ThreadSanitizer: data race (pid=9273)
  Write of size 6 at 0x728400008169 by main thread (mutexes: write M0):
    #0 setopt_long <null> (curl_http_test+0xf8bd7) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6)
    #1 Curl_vsetopt <null> (curl_http_test+0xff287) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6)
    #2 curl_easy_setopt <null> (curl_http_test+0xff834) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6)
    #3 opentelemetry::v1::ext::http::client::curl::HttpOperation::Abort() <null> (curl_http_test+0xaceb8) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6)
    #4 opentelemetry::v1::ext::http::client::curl::Session::CancelSession() <null> (curl_http_test+0x85194) (BuildId: 2ecf4c8802f9938f5574a5cb9410d448fa1062b6)
[...]

PR #4392, opened 2026-08-10
[BUG] Cancel a curl session without writing to the easy handle

### LINKED ISSUE #4375: [BUG] HttpOperation::Abort writes to the curl easy handle from the cancelling thread
Splitting this out of #4369, which is about the test that happened to expose it. dbarker asked for a separate issue to track the client side and the discussion of how to fix it.

## What happens

`HttpOperation::Abort()` writes to the curl easy handle from whichever thread cancels:

```cpp
void HttpOperation::Abort()
{
  is_aborted_.store(true, std::memory_order_release);
  if (curl_resource_.easy_handle != nullptr)
  {
    // Enable progress callback to abort from polling thread
    curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_NOPROGRESS, 0L);
    ...
```

The background thread is driving that same handle through the multi handle at the time. libcurl's rule is that an easy handle must not be used from two threads at once, and TSAN saw exactly that pairing:

```text
main thread   Session::CancelSession -> HttpOperation::Abort -> curl_easy_setopt
thread T17    curl_multi_perform -> Curl_connect -> resolve_server -> Curl_resolv
```

Full log in #4369.

## Reach

`Session::CancelSession()` is public, and `HttpClient::CancelAllSessions()` calls it for every session, which is what `OtlpHttpClient::Shutdown()` does. So an application that shuts down while a request is in flight is in this, not only a test. It is intermittent: the window is however long the IO thread spends inside the handle, so the odds go up when the cancel lands early, during name resolution or connect.

## A possible fix

`OnProgressCallback` reads nothing but `is_aborted_`, which is atomic and is already set on the line above, and the callback is registered at setup through `CURLOPT_XFERINFOFUNCTION` and `CURLOPT_XFERINFODATA`. `CURLOPT_NOPROGRESS` is the only piece still left at its default, which is why `Abort()` has to reach for the handle at all. Turning it on at setup would leave `Abort()` with the flag and the scheduled teardown:

```cpp
void HttpOperation::Abort()
{
  is_aborted_.store(true, std::memory_order_release);
  if (async_data_ && nullptr != async_data_->session)
  {
    async_data_->session->GetHttpClient().ScheduleAbortSession(
        async_data_->session->GetSessionId());
  }
}
```

The `curl_resource_.easy_handle` read goes too, since reading that member from the cancelling thread is the same kind of sharing.

The cost is a progress callback that runs during every transfer and does one atomic load. I have this built: 22 of 22 in `curl_http_test`, 19.1 s against 18.1 s on main so no slowdown, clean under `OTELCPP_MAINTAINER_MODE=ON`, and a probe confirms the callback really does fire, 459 times across the suite, so `CURLOPT_NOPROGRESS = 0` at setup is taking effect.

I have not sent it, because it is a behaviour change in a shared client and worth agreeing on first. Happy to open it if this shape looks right.

## What I could not do

Reproduce it. Twelve TSAN runs on the box, including one pointed at a slow-resolving host to widen the window, all clean. The evidence is the CI log and the code path, so whatever fix lands wants checking against the Bazel TSAN config rather than a local one.

thc1006:
Fixes #4375.

`Abort()` ran `curl_easy_setopt` on whichever thread called `CancelSession()`, while the IO thread was driving the same easy handle through `curl_multi_perform`. That is the pairing TSAN caught in #4369, and libcurl's rule is that one easy handle must not be used by two threads at the same time.

## Where the option is set now

`CURLOPT_NOPROGRESS` was the only option still left at its default, which is the whole reason `Abort()` had to reach for the handle. `SendAsync()` sets it now, right beside `CURLOPT_PRIVATE`, on the thread that owns the handle and before the handle is scheduled. Cancelling is left with the atomic flag the progress callback already reads plus the abort it schedules.

It goes there rather than in `Setup()` because `Setup()` serves the synchronous path too, and nothing on that path can be cancelled: `Abort()` runs only from `Session::CancelSession()`, and `HttpClientSync` never makes a session. Arming the callback there would have bought nothing.

@owent, this keeps the fallback you described. `ScheduleAbortSession()` already calls `wakeupBackgroundThread()`, so on 7.68 and later `curl_multi_wakeup()` gets the IO thread out of `curl_multi_poll` on its own. What the progress callback still buys is the other case, where the IO thread is inside `curl_multi_perform` rather than waiting in the poll, and arming it up front costs one atomic load per progress tick instead of a cross thread write to the handle. The callback fires 376 times across `curl_http_test`, so the option is taking effect, and the cost does not show up: over five runs of the same 22 cases, `main` came in between 17.6 s and 18.6 s and this branch between 16.1 s and 18.6 s, which is the same spread.

## The race that actually reproduces

There is a second one in the same call, and it turned out to be the one I could trigger. `AsyncData::session` is a plain `Session *` that `Abort()` reads at `:1516` while `Cleanup()` writes `nullptr` to it at `:552` on the IO thread. It is `std::atomic<Session *>` now.

There is no lifetime question sitting behind the pointer value here: `Abort()` has exactly one caller, `Session::CancelSession()` at `http_client_curl.cc:249`, so the pointer is always the session whose own method is running, and callers hold that by `shared_ptr`.

Against unmodified `main`, with only the new test case added, TSAN reports on every run:

```text
WARNING: ThreadSanitizer: data race
  Write of size 8 by thread T3:
    #0 HttpOperation::Cleanup() ext/src/http/client/curl/http_operation_curl.cc:552
    #1 HttpOperation::PerformCurlMessage(CURLcode) ext/src/http/client/curl/http_operation_curl.cc:1629
  Previous read of size 8 by main thread:
    #0 HttpOperation::Abort() ext/src/http/client/curl/http_operation_curl.cc:1516
    #1 Session::CancelSession() ext/src/http/client/curl/http_client_curl.cc:249
```

| `curl_http_test --gtest_filter='*RepeatedCallerThreadCancels*'`, 5 runs | TSAN warnings |
| --- | --- |
| `main` at f6e48180 plus the new case | 5 |
| this branch | 0 |

## The return value had to change with it

`OnProgressCallback()` returned `CURL_PROGRESSFUNC_CONTINUE` when the request had not been aborted. That branch was unreachable while `CURLOPT_NOPROGRESS` was only cleared inside `Abort()`, since by then `is_aborted_` is already up and the callback returns early. Arming the option up front makes it the normal path, and it does not mean what the name suggests:

> If your callback function returns CURL_PROGRESSFUNC_CONTINUE it makes libcurl to continue executing the default progress function.

`lib/progress.c` falls straight through to `progress_meter()`, which writes to `data->set.err`, and that is stderr unless `CURLOPT_STDERR` says otherwise:

```c
      if(rc != CURL_PROGRESSFUNC_CONTINUE) {
        if(rc) {
          failf(data, "Callback aborted");
          return CURLE_ABORTED_BY_CALLBACK;
        }
        return CURLE_OK;
      }
    ...
    if(showprogress)
      progress_meter(data);
```

With the option armed and nothing differing but the return value:

| non-aborted return | `curl_http_test` stderr | progress meter headers |
| --- | --- | --- |
| `CURL_PROGRESSFUNC_CONTINUE` | 10,305 bytes | 31 |
| `0` | 114 bytes, one expected log line | 0 |

So without this half, the fix would have printed curl progress bars to stderr for every export that ran longer than a second. A standalone libcurl program built twice from one source, differing only in that return, gives the same answer: 711 bytes of meter against none.

## Tests

`RepeatedCallerThreadCancelsAreClean` cancels from the caller thread while the IO thread owns the handle, twenty times. Nothing listens on 19937, so each attempt fails to connect and the IO thread reaches `Cleanup()` while the caller is still inside `CancelSession()`, which is the overlap the race needs. It passes on `main` too, since a data race is not a functional failure, and it is there for the sanitizer builds.

The two existing cancel cases keep asserting they cancel from the IO thread, but the comment explaining why has to go: it said cancelling reaches `curl_easy_setopt` and therefore has to run on the owning thread, and after this that is no longer true.

## What this does not fix

Three things I ran into on the way and filed rather than folding in here: #4389, where the IO thread deadlocks on `sessions_m_` recovering from a `curl_multi_perform` error, #4390, where cancelling from a `Created` or `Connecting` event loses the cancel and leaves the caller blocked forever, and #4391, where the handle is reset and its header list freed before the removal from the multi handle. All three are on this file or its neighbour and none of them changes what is here.

#4360 is also unaffected: a cancel that the progress callback turns into `CURLE_ABORTED_BY_CALLBACK` still produces `ConnectFailed` or `SendFailed` followed by `Cancelled`. That was already the case before this change, since `Abort()` armed the same callback, so the event sequence is the same as it was.

## Checks

24 of 24 in `curl_http_test`, clean under `OTELCPP_MAINTAINER_MODE=ON`, clean under `clang-format` 18.1.8. The whole suite under TSAN is 24 of 24 with zero warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ Increment the:
 * [BUG] Stop the curl IO thread deadlocking on itself while recovering from a
   multi handle error
   ([#4394](https://github.com/open-telemetry/opentelemetry-cpp/pull/4394))
+* [BUG] Cancel a curl session without writing to the easy handle from the
+  cancelling thread
+  ([#4392](https://github.com/open-telemetry/opentelemetry-cpp/pull/4392))
 * [CODE HEALTH] Enable clang-tidy `modernize-deprecated-headers` and replace
   deprecated C headers (`stdint.h`, `stddef.h`, `stdlib.h`, `string.h`,
   `stdio.h`, `ctype.h`, `limits.h`, `assert.h`) with their C++ equivalents
--- a/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
+++ b/ext/include/opentelemetry/ext/http/client/curl/http_operation_curl.h
@@ -362,7 +362,8 @@ private:
 
   struct AsyncData
   {
-    Session *session{nullptr};  // Owner Session
+    // Read by Abort() on whichever thread cancels, cleared by Cleanup() on the IO thread.
+    std::atomic<Session *> session{nullptr};  // Owner Session
 
     std::thread::id callback_thread;
     std::function<void(HttpOperation &)> callback;
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -375,12 +375,9 @@ int HttpOperation::OnProgressCallback(void *clientp,
     return -1;
   }
 
-  // CURL_PROGRESSFUNC_CONTINUE is added in 7.68.0
-#  if defined(CURL_PROGRESSFUNC_CONTINUE)
-  return CURL_PROGRESSFUNC_CONTINUE;
-#  else
+  // Not CURL_PROGRESSFUNC_CONTINUE, which asks libcurl to also run its built-in progress
+  // meter, and that meter writes to stderr.
   return 0;
-#  endif
 }
 #else
 int HttpOperation::OnProgressCallback(void *clientp,
@@ -546,11 +543,9 @@ void HttpOperation::Cleanup()
   if (async_data_)
   {
     // Just reset and move easy_handle to owner if in async mode
-    if (async_data_->session != nullptr)
+    Session *session = async_data_->session.exchange(nullptr, std::memory_order_acq_rel);
+    if (session != nullptr)
     {
-      auto session         = async_data_->session;
-      async_data_->session = nullptr;
-
       if (curl_resource_.easy_handle != nullptr)
       {
         curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_PRIVATE, NULL);
@@ -1438,7 +1433,7 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
 
   async_data_.reset(new AsyncData());
   async_data_->is_promise_running.store(false, std::memory_order_release);
-  async_data_->session = nullptr;
+  async_data_->session.store(nullptr, std::memory_order_release);
 
   ReleaseResponse();
 
@@ -1452,12 +1447,17 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   }
   curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_PRIVATE, session);
 
+  // Only a session can be cancelled, so only this path needs the progress callback live. Setting
+  // it here, on the thread that owns the handle and before the handle is scheduled, leaves
+  // Abort() with nothing to do but raise the flag.
+  curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_NOPROGRESS, 0L);
+
   DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
   is_finished_.store(false, std::memory_order_release);
   is_aborted_.store(false, std::memory_order_release);
   is_cleaned_.store(false, std::memory_order_release);
 
-  async_data_->session = session;
+  async_data_->session.store(session, std::memory_order_release);
   if (false == async_data_->is_promise_running.exchange(true, std::memory_order_acq_rel))
   {
     async_data_->result_promise = std::promise<CURLcode>();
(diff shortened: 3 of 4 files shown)


lalitb (approved):
LGTM. Nice work finding and fixing this race. Thanks!

marcalff:
@thc1006 

Please resolve merge conflicts, will merge after that.

thc1006:
Rebased onto main and it's mergeable again.

Both conflicts were with #4394, which landed in between. The test file one keeps both cases, `ResetMultiHandleWithASessionDoesNotDeadlock` from #4394 and the two cancel cases from here, and the CHANGELOG keeps both entries. I checked afterwards that #4394's own change survived rather than assuming it had: `friend class HttpClientTestPeer` is still in the header, and `sessions_m_` is still scoped to the snapshot inside `resetMultiHandle` rather than held across `CancelSession` and `doRemoveSessions`.

26 of 26 in `curl_http_test`, `clang-format` 18.1.8 clean, and `include-what-you-use` 0.26 with the repository mapping file reports nothing on the three files this touches.

The two commits also became one. The second only corrected the description of the first, so keeping both would have put the same subject line on the branch twice.

thc1006:
Conflicts resolved, the branch is on current main.

The one red job, `CMake -> exporter proto (Build as DLL)`, is not from this branch. Step 5, the environment setup, failed while bootstrapping vcpkg:

```
Downloading https://github.com/microsoft/vcpkg-tool/releases/download/2025-09-03/vcpkg.exe -> D:\a\opentelemetry-cpp\opentelemetry-cpp\tools\vcpkg\vcpkg.exe...
While calling Windows API function WinHttpReceiveResponse got error 0x00002EE2:
The operation timed out
D:\a\opentelemetry-cpp\opentelemetry-cpp\tools\vcpkg\scripts\bootstrap.ps1 : Downloading vcpkg.exe failed.
```

The three steps after it, `run cmake test (DLL build)`, `run cmake cxx20 test (DLL build)` and `run otprotocol test (DLL build)`, are all `skipped`, so nothing here was compiled in that job and the log holds no compiler output at all. The rebase touched `ext/test/http/curl_http_test.cc` and `CHANGELOG.md` and nothing else.

Re-running it needs admin rights on the repository, so that one is yours rather than mine.

Issue #4265, opened 2026-07-17
[TEST] Fix ext.http.curl.BasicCurlHttpTests.ElegantQuitQuick flaky test


dbarker:
The following test is timing based and flaky in CI. This test should be fixed or disabled. 

```
 898/1247 Test  #898: ext.http.curl.BasicCurlHttpTests.ElegantQuitQuick .........................................................................***Failed    0.05 sec
Running main() from /private/tmp/googletest-20250808-4877-y4cffa/googletest-1.17.0/googletest/src/gtest_main.cc
Note: Google Test filter = BasicCurlHttpTests.ElegantQuitQuick
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BasicCurlHttpTests
[ RUN      ] BasicCurlHttpTests.ElegantQuitQuick
/Users/runner/work/opentelemetry-cpp/opentelemetry-cpp/ext/test/http/curl_http_test.cc:681: Failure
Value of: cost < std::chrono::milliseconds{20}
  Actual: false
Expected: true
cost ms: 21 libcurl version: 0x80600

[  FAILED  ] BasicCurlHttpTests.ElegantQuitQuick (43 ms)
[----------] 1 test from BasicCurlHttpTests (43 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (43 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] BasicCurlHttpTests.ElegantQuitQuick

 1 FAILED TEST
```

owent:
Could we increase the timeout here? GitHub CI runners sometimes schedule CPUs slowly, and a slightly higher timeout might help prevent flaky failures.

dbarker:
> Could we increase the timeout here? GitHub CI runners sometimes schedule CPUs slowly, and a slightly higher timeout might help prevent flaky failures.

We could but it may be helpful to check and see if the test can be made more deterministic. I'd like to reduce the timeouts and sleeps in unit tests :)

yogarajalakshmi-s:
Hi @dbarker! Can you assign this to me?

[...]

PR #4395, opened 2026-08-10
[BUG] Complete a curl operation that is never scheduled

### LINKED ISSUE #4390: [BUG] Cancelling from a Created or Connecting event loses the cancel and leaves the operation unfinished

### LINKED ISSUE #4393: [BUG] An invalid URL creates an unregistered curl session whose operation never completes
**Describe your environment** `main` at f6e48180, Ubuntu 24.04, gcc 14, libcurl 8.14.1, CMake. Nothing platform specific.

`HttpClient::CreateSession()` gives back a session that the client does not know about when the URL fails to parse:

```cpp
std::shared_ptr<...Session> HttpClient::CreateSession(nostd::string_view url) noexcept
{
  const auto parsedUrl = common::UrlParser(std::string(url));
  if (!parsedUrl.success_)
  {
    return std::make_shared<Session>(*this);                       // :330
  }
  auto session = std::make_shared<Session>(*this, parsedUrl.scheme_, ...);
  auto session_id = ++next_session_id_;                            // :334
  session->SetId(session_id);

  std::lock_guard<std::mutex> lock_guard{sessions_m_};
  sessions_.insert({session_id, session});                         // :338
  return session;
}
```

The early return skips both the id and the registration, so the caller holds a session whose `session_id_` is still its default `0` (`http_client_curl.h:233`) and which is in nobody's map. It is otherwise a normal session: `CreateRequest()` and `SendRequest()` both work on it.

That would be harmless if the request then failed. It does not. `CURLOPT_URL` is not parsed when it is set, so `Setup()` returns `CURLE_OK` and `SendAsync` goes on to create the promise and call `ScheduleAddSession(0)`. `doAddSessions` looks `0` up in `sessions_`, misses, and `continue`s. Nothing ever runs the operation and nothing ever completes its future.

**Steps to reproduce** Against unmodified `main`, using `TerminalCountingHandler`, which is already in the test file:

```cpp
TEST_F(BasicCurlHttpTests, InvalidUrl)
{
  auto manager = std::make_shared<http_client::curl::HttpCurlClientFactory>()->Create();
  auto session = manager->CreateSession("http://127.0.0.1:not-a-port");
  auto request = session->CreateRequest();
  request->SetUri("get/");

  auto handler = std::make_shared<TerminalCountingHandler>();
  session->SendRequest(handler);
  session->FinishSession();
  manager->FinishAllSessions();
}
```

`SendRequest` returns and `FinishSession` never does. Under a 60 second `timeout` the binary exits 124 having finished zero cases. A core taken at 12 seconds:

```text
#5  std::__future_base::_State_baseV2::wait (...) at /usr/include/c++/14/future:360
#7  HttpOperation::Finish (...) at ext/src/http/client/curl/http_operation_curl.cc:516
#8  Session::FinishSession (...) at ext/src/http/client/curl/http_client_curl.cc:259
```

I only have that one URL measured, since the loop I wrote never got past it.

**What is the expected behavior?** A URL the client cannot parse ends the request with a terminal event, ideally `CreateFailed` carrying the reason, and `FinishSession()` returns.

**What is the actual behavior?** `FinishSession()` blocks forever. For an exporter this means an application that gets its endpoint wrong hangs at shutdown rather than logging a bad endpoint, and the misconfiguration is usually a string from the environment.

**Additional context** This is the same root cause as #4390, reached a different way: **an operation can be given a promise without ever being scheduled, and then nobody is left to complete it.** #4390 needs a handler that cancels from an early event; this one needs a typo. One fix covers both, and it does not have to be large: if `ScheduleAddSession` reports that the session is not registered, `SendAsync` can complete the operation itself instead of leaving the future open.

Whether an unparseable URL should also get a distinct `CreateFailed` with the reason, rather than the generic terminal event that fix would produce, is a separate and better question, and it is yours rather than mine. I am happy to write either shape.

Found while working on #4375 and #4390.

### LINKED ISSUE #4402: [BUG] FinishSession() from an OnEvent handler deadlocks the curl client's background thread
**Describe your environment**

main at `768d04de`, Debian, gcc 14.2.0, libcurl 8.14.1, plain CMake build with the defaults. No sanitizer needed, an ordinary build reproduces it every time. I haven't tried other platforms, but nothing in the mechanism looks platform specific.

**Steps to reproduce**

Drop this into `ext/test/http/curl_http_test.cc`. `CustomEventHandler` is the helper already in that file. Nothing listens on 19937, so the connection fails and the background thread dispatches `ConnectFailed`.

```cpp
class FinishFromOnEventHandler : public CustomEventHandler
{
public:
  http_client::Session *session_ = nullptr;
  std::atomic<bool> entered_{false};
  std::atomic<bool> returned_{false};

  void OnResponse(http_client::Response &) noexcept override {}

  void OnEvent(http_client::SessionState state, nostd::string_view) noexcept override
  {
    if (state == http_client::SessionState::ConnectFailed && session_ != nullptr &&
        !entered_.exchange(true))
    {
      session_->FinishSession();
      returned_.store(true, std::memory_order_release);
    }
  }
};

TEST_F(BasicCurlHttpTests, FinishSessionFromOnEvent)
{
  auto session_manager = std::make_shared<http_client::curl::HttpCurlClientFactory>()->Create();
  auto session = session_manager->CreateSession("http://127.0.0.1:19937");
  auto request = session->CreateRequest();
  request->SetUri("get/");

  auto handler      = std::make_shared<FinishFromOnEventHandler>();
  handler->session_ = session.get();

  session->SendRequest(handler);
  std::this_thread::sleep_for(std::chrono::seconds(3));

  EXPECT_TRUE(handler->entered_.load(std::memory_order_acquire)) << "handler never ran";
  EXPECT_TRUE(handler->returned_.load(std::memory_order_acquire))
      << "FinishSession from OnEvent did not return";
}
```

The handler has to be the first caller of `FinishSession()`. If the test body calls it first, `Finish()` early-returns at the `is_finished_.exchange` on line 506 and never reaches the wait below, which is what sent me down the wrong path the first time I looked at this.

`FinishSession()` returns, the same as it does when a handler calls it from `OnResponse`. That path is deliberately supported: `Cleanup()` marks the calling thread on lines 566 and 568 around `callback(*this)`, and `Finish()` consults that marker on line 514 before it waits. `BasicCurlHttpTests.FinishInAsyncCallback` covers it, through `FinishInCallbackHandler`, which calls `session_->FinishSession()` from `OnResponse` on line 158 of the same test file.

It never returns, and it takes the client's background thread with it.

```
curl_http_test.cc: Failure
FinishSession from OnEvent did not return
```

The process then has to be killed. Under `timeout 90` the run exits 124.

**Additional context**

Backtrace of the stuck thread, from a core taken on `f6e48180`. The only change to this file since then is #4399 in `NextRetryTime`, which sits well below every line referenced here.

```
#5  std::__future_base::_State_baseV2::wait ()
#6  std::__basic_future<CURLcode>::wait ()
#7  HttpOperation::Finish ()                        http_operation_curl.cc:516
#8  Session::FinishSession ()                       http_client_curl.cc:259
#9  FinishFromOnEventHandler::OnEvent ()
#10 HttpOperation::DispatchEvent ()                 http_operation_curl.cc:412
#11 HttpOperation::PerformCurlMessage (code=CURLE_COULDNT_CONNECT)
#13 HttpClient::MaybeSpawnBackgroundThread()::<lambda()>
```

`Finish()` waits on `result_future` on line 516. The one place that promise is satisfied is line 574, at the tail of `Cleanup()`, and the same background thread only reaches `Cleanup()` after `PerformCurlMessage` returns. So the thread is blocked waiting for something further down its own call stack. No race is involved, it deadlocks on every run.

What decides whether `Finish()` waits is that thread marker, and the marker only covers `callback(*this)` inside `Cleanup()`. `DispatchEvent` on line 412 calls `OnEvent` with no marker set, so every `OnEvent` the background thread dispatches is exposed: `Connected`, `Sending`, `Response`, `ConnectFailed`, `SendFailed` and `Cancelled`. Only `Created` and the first `Connecting` are dispatched on the caller's own thread and come back normally.

Five of those dispatches sit inside `WriteMemoryCallback`, `WriteVectorHeaderCallback`, `WriteVectorBodyCallback`, `ReadMemoryCallback` and `PreRequestCallback`, so in those cases the thread parks inside `curl_multi_perform` while libcurl still owns the easy handle.

One state is split down the middle, and that is what makes this awkward to discover from the outside. On an aborted request `Cancelled` reaches the handler twice: once from `Cleanup()` line 536 through `DispatchEvent`, carrying libcurl's error string as the reason, and once from the completion lambda in `Session::SendRequest`, carrying an empty reason. The second one runs inside the marker and returns fine. The first one deadlocks. Same state, different reason string, opposite outcome. (The double delivery itself is #4360, not this.)

`background_thread_` is a single `std::unique_ptr<std::thread>` per `HttpClient` and `MaybeSpawnBackgroundThread` reuses the running one, so once it is parked nothing else on that client makes progress either. For an exporter that means the block is not limited to the request whose handler called in.

The public header documents `session->FinishSession()` from the caller and says nothing either way about handlers, so I don't think this is a documented restriction users are ignoring. The library already supports the same call from the adjacent callback, and that's the part that makes the asymmetry look unintended rather than deliberate.

I haven't sent a patch because I'd be guessing at the contract. Widening the marker to cover `DispatchEvent` is the small change, but it's one slot on the operation and `DispatchEvent` runs on both threads, so it would need save and restore rather than clear to default. The other reading is that `Finish()` should never wait at all when it's already on the thread that drives the transfer. Say which shape you'd want and I'll put it together.

### LINKED ISSUE #4408: [BUG] The curl client writes HttpOperation::session_state_ from two threads without synchronization
**Describe your environment** `main` at f1e1c9cb, Linux, gcc 14.2, libcurl 8.14.1, CMake build with `-fsanitize=thread`. The race is in the source rather than in a build option, so the Bazel `--config=tsan` toolchain sees the same thing.

**Steps to reproduce**

`HttpOperation::DispatchEvent` stores the new state after the handler has returned, at `http_operation_curl.cc:415`:

```cpp
void HttpOperation::DispatchEvent(SessionState type, const std::string &reason)
{
  if (event_handle_ != nullptr)
  {
    event_handle_->OnEvent(type, reason);
  }

  session_state_ = type;
}
```

`session_state_` is a plain `SessionState` member (`http_operation_curl.h:341`, and `SessionState` is a `std::uint8_t` enum). Two threads reach that store for one operation whenever a handler cancels from an early event on a client that is already polling:

1. Send a request on a fresh client and let it finish. `MaybeSpawnBackgroundThread()` runs after `SendAsync` returns, so this first request is what puts an IO thread there at all.
2. Create a second session and call `CancelSession()` from its `Connecting` event, without leaving the handler yet.
3. The cancel wakes the IO thread, which runs `doAbortSessions` (`http_client_curl.cc:777`) into `Session::FinishOperation` (`:269`) into `HttpOperation::Cleanup`. `Cleanup` finds a non terminal state and dispatches a `Cancelled` of its own at `http_operation_curl.cc:536`, which stores `session_state_` on the IO thread.
4. The handler returns and `DispatchEvent` stores `session_state_ = Connecting` on the caller thread.

I held the handler open for 400 ms so the overlap is wide enough to be deterministic rather than something I got lucky on. Three TSAN runs out of three report it, and five out of five reach the overlap on the behavioral build.

Either one thread owns the store, or the store is synchronized. And a state the operation has already left should not replace a terminal one.

TSAN reports a data race on the store. Trimmed to the two stacks:

```
WARNING: ThreadSanitizer: data race
  Write of size 1 by main thread:
    #0 HttpOperation::DispatchEvent(...) http_operation_curl.cc:415
    #1 HttpOperation::SendAsync(...) http_operation_curl.cc:1455
    #2 Session::SendRequest(...) http_client_curl.cc:209

  Previous write of size 1 by thread T2:
    #0 HttpOperation::DispatchEvent(...) http_operation_curl.cc:415
    #1 HttpOperation::Cleanup() http_operation_curl.cc:536
    #2 Session::FinishOperation() http_client_curl.cc:269
    #3 HttpClient::doAbortSessions() http_client_curl.cc:777
```

Past the undefined behavior, the later store wins, so `GetSessionState()` ends at `Connecting` for an operation that was cancelled. Two places read that member and act on it: `Cleanup` at `:529` decides from it whether to dispatch at all, and the completion callback in `Session::SendRequest` compares it against `Response` to decide whether to deliver a response.

**Additional context**

Nothing in `ext/test/http/curl_http_test.cc` reaches this today, which is why `Bazel tsan config` is green. Every case that cancels from an early event uses a freshly created client, and there the IO thread does not exist while the event runs, so there is only one writer. That is measured rather than assumed: the same probe without the first request reports one thread inside the handler, and with it, two.

A warning for whoever writes the regression test. My first attempt counted concurrent handler entries with `acq_rel` atomics and TSAN went quiet. The IO thread's release RMW and the caller's later RMW on the same atomic synchronize with each other, which orders the two stores and hides the race the counters exist to catch. They have to be `relaxed`. I checked that with a deliberate unsynchronized counter on the same line as a control: the `acq_rel` build reported neither, the `relaxed` build reported both.

On `main` this scenario reports two races and then hangs. #4395 removes the hang and the other one, a write to `async_data_->session` in `SendAsync` racing the read of it in `Cleanup`, and leaves this one. So this is not something #4395 introduces, and it is not something #4395 fixes either.

Two shapes for a fix, and the choice does not look like mine to make:

- Make `session_state_` a `std::atomic`. Removes the undefined behavior, leaves the later writer winning, so a cancelled operation can still report `Connecting`. It also changes the layout of a type in an installed header.
- Store before dispatching and refuse to leave a terminal state. That fixes the reported value too, but it changes what a handler sees when it asks `GetSessionState()` from inside its own event, which today is the state before the one it is being told about.

Scheduling note, same as on #4360. #4392 and #4395 are open against these two files. Tell me which shape you want and I will fit it around whatever has landed rather than sending a change that collides with them. It also holds up the regression test for the polling client half of #4390, because that test turns `Bazel tsan config` red on this race.

thc1006:
Fixes #4390.
Fixes #4393.
Fixes #4408.
Fixes #4402.

An operation could be given a promise and then never handed to the IO thread, and then nobody was in a position to complete it. `FinishSession()` blocked forever.

Three ways in, and each one hangs on `main` today:

| | |
| --- | --- |
| A handler cancels from the `Created` or `Connecting` event | `SendAsync` dispatched the event and only then reset `is_aborted_`, so the cancel was thrown away, while `CleanupSession()` had already taken the session out of the client. `ScheduleAddSession` then erased the pending abort for good measure |
| The URL does not parse | `CreateSession` hands back a session it never registered and never gave an id. `CURLOPT_URL` is not checked when it is set, so `Setup()` succeeds and the request reaches `ScheduleAddSession(0)`, which nothing can find |
| `curl_multi_add_handle` rejects the handle | The result was dropped, so the operation sat there with nothing to run it |

They are one invariant with three holes in it: **an operation that never gets scheduled still has to finish.**

## What changed

The flags and the cancel route are published before the first event, so a handler cancelling from it is not overwritten by the rest of `SendAsync`. The future is published *after* the event, deliberately: a handler calling `FinishSession()` from `Connecting` would otherwise wait on a transfer that has not been scheduled yet, with `SendAsync` unable to schedule it because it is inside that handler.

`ScheduleAddSession` now reports whether the session was still registered, and `SendAsync` finishes the operation itself when it was not, or when the event cancelled it. `doAddSessions` does the same for the ones libcurl rejects, outside `sessions_m_`, since finishing one reaches the caller's handler and a handler that cancels takes that lock again.

All three say the same thing to the handler, a failed create carrying the reason, through one method rather than the same three lines in two places. The one libcurl refuses used to arrive as `Cancelled`, which the enum calls "(manually) cancelled" and which both exporters print in those words, for a request nobody cancelled and without the reason libcurl gave. Nothing behaves differently for the change: `CreateFailed` and `Cancelled` both set `need_stop` in the OTLP HTTP handler and both end the wait in the Elasticsearch one, so what moves is the message.

There is one ordering subtlety worth pointing at. Between publishing the future and scheduling the add, the IO thread can already have torn the operation down and found no future to complete, so `SendAsync` rechecks `is_cleaned_` and settles it with an `exchange`. Whichever side gets there first, the value is set exactly once.

That same window had a second defect in it, raised by @lalitb in review. `DispatchEvent` stored `session_state_` after the handler returned, so a handler cancelling from the first event on a client that is already polling had the IO thread storing a `Cancelled` of its own at the same time, and the store after the handler returned overwrote it. ThreadSanitizer reports it on this branch before the change, five runs out of five. The store now happens before the handler runs, which orders it ahead of the cancel through `sessions_m_` and `session_ids_m_`, and the member is a `std::atomic` for the paths where a third thread cancels instead of the handler. `SessionState` is a `std::uint8_t` enum, and on the toolchains this was built with `std::atomic` of it stays one byte with alignment one, so the layout does not move. The standard promises neither, so both are `static_assert`s in `http_operation_curl.cc` rather than a claim here: a toolchain where they do not hold fails the build instead of changing the layout quietly. Worth being exact about whose layout, since I was not. `ext/CMakeLists.txt` installs four headers from `ext/http/client` and `http_operation_curl.h` is not one of them, so a CMake consumer never sees this type. `//ext` takes `hdrs = glob(["include/**/*.h"])`, so a Bazel consumer does, which is why the guard is worth keeping rather than dropping. Lock freedom is a separate property and nothing here checks it, since `is_always_lock_free` is C++17 and this still builds at C++14.

## What a second review round found

Three things, all in the code this branch adds rather than in what it inherited, and each one
is held by a case that fails without the fix.

`FinishUnscheduled` cleaned up and then dispatched the terminal event. The operation holds the
caller's handler as a bare pointer and the only strong reference to it is the one
`Session::SendRequest` captured in the completion callback, so the cleanup that releases that
callback is what frees the handler, and the event that followed read it. AddressSanitizer
reports a heap-use-after-free at the `OnEvent` call reached from `FinishUnscheduled`. Nothing in
the suite caught it, because every case that reaches this path either keeps its own reference to
the handler or hands `SendAsync` a completion capturing none, so the ownership that breaks was
never assembled. The event goes first now. `AnUnscheduledSessionTellsAHandlerNothingElseHolds` is
the only case in the file that leaves the completion callback as the sole owner, which is what
every real caller of `SendRequest` arranges, and it records its own event against its own
destruction into state the case owns, so it fails without a sanitizer as well as with one.

`AsyncData::callback_thread` was a plain `std::thread::id` written on the thread running a
callback and read by `Finish()` and by the destructor on whichever thread called them. Its
accessors put a standalone `atomic_thread_fence` either side of a plain load and store, which
orders nothing: fences synchronize through atomic operations on a common object and there was
none. ThreadSanitizer stayed quiet because both accessors carried
`OPENTELEMETRY_SANITIZER_NO_THREAD`, and the fences were compiled out under that same sanitizer,
so what it was told to ignore was a bare unsynchronized access. The callback scope already
answers the question the member existed to answer and answers it without shared state, so the
member, its accessors and the friend declaration are gone. `Finish()` consulted the scope before
it ever reached the member, which made the second check unreachable as anything but true.

`FinishUnscheduled` is private now, with the client as a friend. It forces a terminal state,
cleans up, dispatches, fulfils the promise and hands the easy resource back, and the client is
the only thing that ever discovers a request nothing will run. `ext/CMakeLists.txt` does not
install this header, but `//ext` globs it, so a public method here is source API for a Bazel
consumer.

## The refusal was reported before the operation it describes was settled

A third reading of `doAddSessions` found the loop in the wrong order. It logged the refusal, then read `GetOperation()`, then finished it.

The log goes through the global handler, which is replaceable application code and, unlike everything else on this path, is not inside a callback scope for that operation. A handler answering with `Session::FinishSession()` reaches `HttpOperation::Finish()`, which is not `InsideCallbackFor(this)` and so waits on a promise only the `FinishUnscheduled` below it can fulfil, on the same thread. Reading the operation after the log is the other half: a handler calling `SendRequest()` from there swaps the operation, and the loop then finishes the new one and leaves the refused one unsettled.

`FinishUnscheduled` already dispatches its event before it cleans up, and its comment says why that matters: a handler calling `FinishSession()` from that event is inside a callback for the operation, so `Finish()` returns rather than waiting. Reporting after it is what brings the log handler under the same protection. Nothing else moved.

`ARejectedSessionIsSettledBeforeItIsReported` installs a global log handler that records whether the terminal event has arrived by the time the refusal is logged, and puts the previous handler and level back through a scope guard. With the report moved back above the settlement it fails at the ordering assertion; with the change it passes. Forty cases in the binary, all passing.

Worth naming because it nearly went in wrong: the first version of that assertion asked only whether the handler had seen any event. `SendAsync` dispatches `Created` during setup, so that is true in both orders, and the case passed with the defect deliberately put back. It asks for `CreateFailed` now.

## What this path shares with the rest of the client, and does not fix

Telling a refused session that nothing will run its request means dispatching a terminal event, and a handler is free to answer it by starting another request on the same session. `Session::curl_operation_` is a `std::unique_ptr<HttpOperation>` and `Session::SendRequest` assigns into it with `reset`, so that answer frees the operation whose member function is dispatching the event. AddressSanitizer reports a heap use after free, single threaded and on the first attempt.

This is not introduced here. The same two stacks reproduce on `main` at `58ed80d9` through `Connecting`, which is dispatched from inside `SendAsync` and has nothing to do with this change. I have put the measurement on #4448, where the question of what a callback may call is one of the open contract items, rather than opening an issue for it.

I looked at fixing it inside this pull request and decided not to. Taking ownership of the operation before finishing it removes the use after free, and it also makes `FinishSession()` on another thread return without waiting, because that wait is keyed on the session still holding an operation rather than on there being outstanding work. `FinishFromAnotherThreadWaitsForTheTerminalEvent` fails on that change, and it is right to. Both behaviours follow from the same ownership model, so choosing between them is the contract decision rather than something this change should settle on its own.

## The part I looked hardest at

This makes the completion callback run on the calling thread in those paths, where it always ran on the IO thread before. I went through every in-tree caller rather than assume:

- `OtlpHttpClient::addSession` closes its `session_manager_lock_` scope before `SendRequest`, which the comment there says is deliberate, and the lock is recursive anyway. The handler's `Cancelled` reaches `Unbind` and `ReleaseSession`, and `ReleaseSession` moves the session to `gc_sessions_` rather than calling `FinishSession`, so nothing waits.
- The Elasticsearch exporter holds no lock across either of its `SendRequest` calls, and the synchronous `waitForResponse()` waits on a predicate, which is exactly the case where the completion is recorded before the waiter arrives.
- A handler calling `FinishSession()` from inside a callback this operation dispatched is covered by a callback scope. Entering one records the operation on the calling thread, and `Finish()` returns rather than waiting on a completion that thread has not published yet. That covers the completion callback and every event, which is what closes #4402.
- This is not a new shape either way. `Session::SendRequest` already calls `callback->OnEvent(CreateFailed, "")` inline on the calling thread when `SendAsync` fails.

The behaviour change a user could notice is that a cancel from `Connecting` is now honoured. It used to be dropped and the request went out regardless.

## Tests

`CancelFromCreatedCompletes`, `CancelFromConnectingCompletes` and `InvalidUrlCompletes`. All three hang rather than fail without the change: `timeout 25` gives exit 124 and zero cases finished, and they pass in about 500 ms with it. Worth knowing for a bisect, since a regression here looks like a CI timeout rather than a red assertion.

`CancelFromConnectingWhilePollingCompletes` is the fourth, and the only one that puts an IO thread there before the event runs. A client spawns one only after `SendAsync` returns, so it warms the client with a completed request first, then cancels from `Connecting` and stays in the handler until the IO thread has entered it too. Two things follow from that. It reports the state store race without the change and none with it, five runs out of five each way. And instrumenting both tail branches of `SendAsync` and running the whole binary shows it is the only thing that reaches the `is_cleaned_` recheck, which the three cases above never touch.

What it stays in the handler *on* took a second look. It waited on `inside_events_`, which is raised on the way into an event and lowered on the way out, and the event the IO thread dispatches there is a few atomics long, so sampling it every millisecond almost never caught it at two: four runs measured 30005, 30509, 30006 and 30015 ms and all passed, because the count the assertion reads is written by the thread that creates the overlap and never goes down. Waiting on that one instead leaves as soon as the overlap has happened, and the case goes from 30006 ms to 511 ms, ten runs out of ten, with the binary down from about 53 seconds to 23.5. The shorter wait is also the one that works: against a faithful revert of what this branch changed about `session_state_`, a plain member again and the store back after the handler, ThreadSanitizer reports the race five runs out of five with it and none at all with the thirty second version, which spends that long and a great many reads of the same address between the two writes. A bound that does expire now means the overlap really did not happen, so it is recorded and checked rather than left to read as one dispatched event.

`ASessionTheMultiHandleRefusesIsFinished` is the third root cause, and nothing reached it before. `SendAsync` does the whole async setup and `Session::SendRequest` is what starts the worker, so the case builds the operation, sends it, and drives `doAddSessions` itself on one thread. What refuses the add is a seam: the client reaches `curl_multi_add_handle` through a member, and a case replaces it with one that returns the code it wants. It used to swap the client's multi handle for a null one, which works, but it works by handing libcurl a handle its own documentation says not to use after `curl_multi_init` returns null, so what was under test was libcurl's defence against invalid input rather than a refusal it promises. The manual only promises that a non-zero return means the add failed. Reverting that finish to the plain one it had fails the case twice over, on the state it reports and on the absence of a cancel.

Its overlap counters are `relaxed` on purpose. With `acq_rel` the two threads synchronize through the counter itself, which orders the very stores the case exists to keep unordered, and ThreadSanitizer goes quiet. A deliberate unsynchronized counter used as a control confirms it: the `acq_rel` build reports neither, the `relaxed` build reports both.

## Checks

All 38 cases in `curl_http_test` pass in each of eight configurations, built and run one at a time from a committed tree: the default one, AddressSanitizer, ThreadSanitizer, UndefinedBehaviorSanitizer with `-fno-sanitize-recover`, C++20, `WITH_OTLP_HTTP_COMPRESSION=ON`, `WITH_THREAD_INSTRUMENTATION_PREVIEW=ON` and `WITH_OTLP_RETRY_PREVIEW=OFF`. Each run reports its own counts rather than going through CTest, because a case an `#ifdef` compiles out still gets a CTest entry and passes as zero tests run. Nothing skipped in any of them, the three sanitizers reported nothing, and the two configurations whose counts move are the ones with gated cases: compression adds two and retry-off drops three. Each of them now carries a CTest timeout of 120 seconds, read back from `ctest --show-only=json-v1`, because what these cases catch fails by hanging: without a bound a regression stops the job rather than reporting, which happened three times while this was being written. The whole binary takes 23.5 seconds, and the two slowest cases are the ones that wait out a request timeout on purpose, at 5007 ms each.

One number in here is pinned as it is rather than as it should be. Cancelling once the transfer belongs to the IO thread delivers two terminal events, one from `Cleanup` and one from the completion callback, and `CancelFromConnectingCompletes` requires exactly two so that a change to it is visible. One is the goal, and that is #4360, which this change does not close. A cancel from `Created` is reported as a failed create for the same reason: the event is dispatched from the constructor, before the `Session` holds the operation, so the cancel never reaches it. Moving the first events out of the constructor is the startup ordering work, not something to add here, and both are pinned rather than smoothed over. Under ThreadSanitizer the whole binary reports zero races with the change, and the new case reports the state store race five times out of five without it. Clean under `OTELCPP_MAINTAINER_MODE=ON` and `clang-format` 18.1.8, clang-tidy one warning below `main` under the CI filters, since deleting the accessor class took one with it and the scope guard's own two were fixed rather than accepted, and IWYU reports correct includes for both changed translation units on all three of its presets.

## Related

#4392 fixed the cross thread write to the easy handle on the same file and has landed, and this is rebased on it: the resolution keeps its CURLOPT_NOPROGRESS placement and its atomic AsyncData::session alongside the ordering here. #4402 is closed here. That is a correction to what this description said before: I had scoped the claim to the terminal event this client generates itself, but the mechanism is not specific to it. A callback scope wraps every dispatch and `Finish()` consults it, so a handler is covered on the events the IO thread delivers during a transfer as well, which is exactly the case the issue describes. `FinishSessionFromAnEventTheIoThreadDelivers` is that case, taken from the issue and made deterministic by waiting on a condition rather than sleeping; take the check out of `Finish()` and it hangs. The alternative was to narrow the mechanism back to the unscheduled path, and that only puts the deadlock back for everyone else, so I would rather claim what the code does. `FinishFromAnotherThreadWaitsForTheTerminalEvent` holds the other direction, that an outside `Finish()` waits for the terminal event and the completion callback rather than for the promise alone.

#4405 changes teardown after `Cleanup()` has handed the easy resource back to the client, so the two are independent, but whichever lands second needs a real rebase rather than a metadata one because both touch the same removal path. I stacked all of #4405 on top of this branch to find out what that costs rather than leave you to work it out. Four of its eight commits conflict. `doAddSessions` is one thing wanted from two ends, since this branch finishes a session `curl_multi_add_handle` rejected and #4405 keeps that same handle off its attachment ledger, and it resolves as the union. `PerformCurlMessage` is another, where this branch adds a function immediately above it and #4405 changes what it returns. The rest is the test file, and that part wants judgement rather than a merge tool: both branches add cases at the same anchor whose first four lines are identical, so a mechanical union folds them into one function with two `TEST_F` headers stacked and a variable declared twice, and it does not compile. Taken apart again and resolved, the combined tree passes the curl suite 43 of 43 under AddressSanitizer with `detect_leaks=1`, the same with `WITH_OTLP_RETRY_PREVIEW=ON`, and under bazel.

## About the lines Codecov marks

Codecov reports 5 lines of this change without cover. Two are the guards in
`doAddSessions()` for a session whose operation or easy handle has gone before
the queue drains; they were evaluated 39 times each and never taken, and
reaching them needs a cancel to land inside that window.

The other three are the wait in `~HttpOperation()` for a transfer that is still
in flight. Those are worth naming rather than covering. The only route to them
is a second `SendRequest()` on a session whose first is still running, since the
client holds a strong reference to every registered session and nothing else
destroys an operation in `Connecting`, `Connected` or `Sending`. #4431 refuses
that second call, so a case written here would pin behaviour that PR removes.
Once it lands those three lines become unreachable and should go, rather than
gain a test now.

Measured rather than read: `527` is hit twice and `528` not at all, so the
condition short circuits before the callback scope test is even evaluated.

### DIFF
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -20,7 +20,6 @@
 #include <random>
 #include <sstream>
 #include <string>
-#include <thread>
 #include <utility>
 #include <vector>
 
@@ -97,28 +96,6 @@ namespace client
 {
 namespace curl
 {
-
-class HttpOperationAccessor
-{
-public:
-  OPENTELEMETRY_SANITIZER_NO_THREAD static std::thread::id GetThreadId(
-      const HttpOperation::AsyncData &async_data)
-  {
-#if !(defined(OPENTELEMETRY_HAVE_THREAD_SANITIZER) && OPENTELEMETRY_HAVE_THREAD_SANITIZER)
-    std::atomic_thread_fence(std::memory_order_acquire);
-#endif
-    return async_data.callback_thread;
-  }
-
-  OPENTELEMETRY_SANITIZER_NO_THREAD static void SetThreadId(HttpOperation::AsyncData &async_data,
-                                                            std::thread::id thread_id)
-  {
-    async_data.callback_thread = thread_id;
-#if !(defined(OPENTELEMETRY_HAVE_THREAD_SANITIZER) && OPENTELEMETRY_HAVE_THREAD_SANITIZER)
-    std::atomic_thread_fence(std::memory_order_release);
-#endif
-  }
-};
 
 size_t HttpOperation::WriteMemoryCallback(void *contents, size_t size, size_t nmemb, void *userp)
 {
@@ -542,13 +519,14 @@ HttpOperation::~HttpOperation()
     case opentelemetry::ext::http::client::SessionState::Connecting:
     case opentelemetry::ext::http::client::SessionState::Connected:
     case opentelemetry::ext::http::client::SessionState::Sending: {
-      if (async_data_ && async_data_->result_future.valid())
+      // Not while inside a callback this operation dispatched: a handler that destroys the
+      // operation it is being told about would be waiting for a completion that only the thread
+      // running the handler can publish.
+      if (async_data_ && async_data_->result_future.valid() &&
+          !CallbackScope::InsideCallbackFor(this))
       {
-        if (HttpOperationAccessor::GetThreadId(*async_data_) != std::this_thread::get_id())
-        {
-          async_data_->result_future.wait();
-          last_curl_result_ = async_data_->result_future.get();
-        }
+        async_data_->result_future.wait();
+        last_curl_result_ = async_data_->result_future.get();
       }
       break;
     }
@@ -576,12 +554,8 @@ void HttpOperation::Finish()
 
   if (async_data_ && async_data_->result_future.valid())
   {
-    // We should not wait in callback from Cleanup()
-    if (HttpOperationAccessor::GetThreadId(*async_data_) != std::this_thread::get_id())
-    {
-      async_data_->result_future.wait();
-      last_curl_result_ = async_data_->result_future.get();
-    }
+    async_data_->result_future.wait();
+    last_curl_result_ = async_data_->result_future.get();
   }
 }
 
@@ -628,9 +602,7 @@ void HttpOperation::Cleanup()
     if (callback)
     {
       const CallbackScope scope{this};
-      HttpOperationAccessor::SetThreadId(*async_data_, std::this_thread::get_id());
       callback(*this);
-      HttpOperationAccessor::SetThreadId(*async_data_, std::thread::id());
     }
 
     // Set value to promise to continue Finish()
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -413,6 +413,39 @@ static_assert(
         alignof(opentelemetry::ext::http::client::SessionState),
     "std::atomic<SessionState> is more aligned, which changes the layout of HttpOperation");
 
+namespace
+{
+// Which operations the calling thread is currently inside a callback for. A handler is allowed to
+// call FinishSession() on the request it is being told about, and that call must not wait for a
+// completion only the thread it is running on can publish.
+thread_local std::vector<const HttpOperation *> callbacks_in_progress;
+
+class CallbackScope
+{
+public:
+  explicit CallbackScope(const HttpOperation *operation) noexcept
+  {
+    callbacks_in_progress.push_back(operation);
+  }
+  ~CallbackScope() { callbacks_in_progress.pop_back(); }
+
+  CallbackScope(const CallbackScope &)            = delete;
+  CallbackScope &operator=(const CallbackScope &) = delete;
+
+  static bool InsideCallbackFor(const HttpOperation *operation) noexcept
+  {
+    for (const auto *entry : callbacks_in_progress)
+    {
+      if (entry == operation)
+      {
+        return true;
+      }
+    }
+    return false;
+  }
+};
+}  // namespace
+
 void HttpOperation::DispatchEvent(opentelemetry::ext::http::client::SessionState type,
                                   const std::string &reason)
 {
@@ -422,6 +455,7 @@ void HttpOperation::DispatchEvent(opentelemetry::ext::http::client::SessionState
 
   if (event_handle_ != nullptr)
   {
+    const CallbackScope scope{this};
     event_handle_->OnEvent(type, reason);
   }
 }
@@ -514,6 +548,14 @@ HttpOperation::~HttpOperation()
 
 void HttpOperation::Finish()
 {
+  // Called from inside a callback this operation dispatched, so the completion being waited for is
+  // the one this thread has not published yet. Returning before the flag below leaves the wait
+  // available to a caller that is not inside a callback.
+  if (CallbackScope::InsideCallbackFor(this))
+  {
+    return;
+  }
+
   if (is_finished_.exchange(true, std::memory_order_acq_rel))
   {
     return;
@@ -572,6 +614,7 @@ void HttpOperation::Cleanup()
     callback.swap(async_data_->callback);
     if (callback)
     {
+      const CallbackScope scope{this};
       HttpOperationAccessor::SetThreadId(*async_data_, std::this_thread::get_id());
       callback(*this);
       HttpOperationAccessor::SetThreadId(*async_data_, std::thread::id());
@@ -1472,26 +1515,18 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   is_finished_.store(false, std::memory_order_release);
   is_aborted_.store(false, std::memory_order_release);
   is_cleaned_.store(false, std::memory_order_release);
-  async_data_->session.store(session, std::memory_order_release);
-  async_data_->callback = std::move(callback);
-
-  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
-
-  // The future alone stays unpublished until after the event, so a handler calling
-  // FinishSession() from it returns instead of waiting on an unscheduled transfer.
+  async_data_->callback       = std::move(callback);
   async_data_->result_promise = std::promise<CURLcode>();
   async_data_->result_future  = async_data_->result_promise.get_future();
   async_data_->is_promise_running.store(true, std::memory_order_release);
 
-  // A cancel from the event may have torn the operation down before the future existed.
-  if (is_cleaned_.load(std::memory_order_acquire))
-  {
-    if (async_data_->is_promise_running.exchange(false, std::memory_order_acq_rel))
-    {
-      async_data_->result_promise.set_value(last_curl_result_);
-    }
-    return CURLE_OK;
-  }
+  // Last, and the only thing that makes this operation reachable from the IO thread: an abort is
+  // queued through this route, so nothing can bring the operation there before the callback and
+  // the completion above exist. Cleanup is therefore the only thing that ever fulfils the promise,
+  // at its tail, once the terminal event and the completion callback have run.
+  async_data_->session.store(session, std::memory_order_release);
+
+  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
 
   if (WasAborted())
   {
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -418,32 +418,45 @@ namespace
 // Which operations the calling thread is currently inside a callback for. A handler is allowed to
 // call FinishSession() on the request it is being told about, and that call must not wait for a
 // completion only the thread it is running on can publish.
-thread_local std::vector<const HttpOperation *> callbacks_in_progress;
-
+// A stack of scopes linked through the scopes themselves, so entering one allocates nothing. Each
+// lives on the stack frame that dispatches the callback, which is exactly as long as the entry
+// needs to be there. A container here would put a heap allocation on every event, and would put
+// it inside a noexcept constructor, where running out of memory calls std::terminate rather than
+// reaching whoever asked for the request.
 class CallbackScope
 {
 public:
   explicit CallbackScope(const HttpOperation *operation) noexcept
+      : operation_{operation}, previous_{current_}
   {
-    callbacks_in_progress.push_back(operation);
+    current_ = this;
   }
-  ~CallbackScope() { callbacks_in_progress.pop_back(); }
+
+  ~CallbackScope() { current_ = previous_; }
 
   CallbackScope(const CallbackScope &)            = delete;
   CallbackScope &operator=(const CallbackScope &) = delete;
 
   static bool InsideCallbackFor(const HttpOperation *operation) noexcept
   {
-    for (const auto *entry : callbacks_in_progress)
+    for (const CallbackScope *scope = current_; nullptr != scope; scope = scope->previous_)
     {
-      if (entry == operation)
+      if (scope->operation_ == operation)
       {
         return true;
       }
     }
     return false;
   }
+
+private:
+  const HttpOperation *operation_;
+  const CallbackScope *previous_;
+
+  static thread_local const CallbackScope *current_;
 };
+
+thread_local const CallbackScope *CallbackScope::current_ = nullptr;
 }  // namespace
 
 void HttpOperation::DispatchEvent(opentelemetry::ext::http::client::SessionState type,
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -404,9 +404,8 @@ int HttpOperation::OnProgressCallback(void *clientp,
 void HttpOperation::DispatchEvent(opentelemetry::ext::http::client::SessionState type,
                                   const std::string &reason)
 {
-  // Published before the handler runs. A handler can cancel from here, which lets the background
-  // thread finish the operation and publish a state of its own, and a store left until after the
-  // handler returned would overwrite that one.
+  // Store before dispatching: a handler may cancel, and a later store would overwrite the state
+  // the background thread publishes.
   session_state_.store(type, std::memory_order_release);
 
   if (event_handle_ != nullptr)
@@ -1455,11 +1454,9 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   // Abort() with nothing to do but raise the flag.
   curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_NOPROGRESS, 0L);
 
-  // A handler can cancel from this event, so everything Cleanup() needs has to be published
-  // before it runs. The callback most of all: Cleanup() takes is_cleaned_ at its start and
-  // swaps the callback much later, so a cleanup racing an assignment placed after the event
-  // would swap an empty one, never run the completion, and leave the session marked active
-  // for good.
+  // Publish everything Cleanup() needs before dispatching, the callback most of all: Cleanup()
+  // reads is_cleaned_ first and swaps the callback last, so one assigned after the event would be
+  // swapped out empty and the completion would never run.
   is_finished_.store(false, std::memory_order_release);
   is_aborted_.store(false, std::memory_order_release);
   is_cleaned_.store(false, std::memory_order_release);
@@ -1468,15 +1465,13 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
 
   DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
 
-  // The future is the one thing that stays unpublished until after the event, so a handler that
-  // calls FinishSession() from it returns instead of waiting on a transfer that has not been
-  // scheduled yet.
+  // The future alone stays unpublished until after the event, so a handler calling
+  // FinishSession() from it returns instead of waiting on an unscheduled transfer.
   async_data_->result_promise = std::promise<CURLcode>();
   async_data_->result_future  = async_data_->result_promise.get_future();
   async_data_->is_promise_running.store(true, std::memory_order_release);
 
-  // Cancelling from the event hands the session to the IO thread, which may already have torn
-  // the operation down and found no future to complete. Settle it here if so.
+  // A cancel from the event may have torn the operation down before the future existed.
   if (is_cleaned_.load(std::memory_order_acquire))
   {
     if (async_data_->is_promise_running.exchange(false, std::memory_order_acq_rel))
@@ -1488,9 +1483,8 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
 
   if (WasAborted() || !session->GetHttpClient().ScheduleAddSession(session->GetSessionId()))
   {
-    // Nothing is going to run this operation. It was cancelled while the event was dispatched,
-    // or its session was never registered because the URL did not parse. Finish it here rather
-    // than leave a future nobody can complete.
+    // Nothing will run this operation, so finish it here rather than leave an unfulfillable
+    // future.
     Cleanup();
   }
 
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -401,6 +401,18 @@ int HttpOperation::OnProgressCallback(void *clientp,
 }
 #endif
 
+// The atomic member is only free of layout cost while it stays the size and alignment of the
+// enum it replaced. The standard does not promise that, so it is checked here rather than
+// asserted in prose: a toolchain where it does not hold changes an installed type and should say
+// so at build time.
+static_assert(sizeof(std::atomic<opentelemetry::ext::http::client::SessionState>) ==
+                  sizeof(opentelemetry::ext::http::client::SessionState),
+              "std::atomic<SessionState> grew, which changes the layout of HttpOperation");
+static_assert(
+    alignof(std::atomic<opentelemetry::ext::http::client::SessionState>) ==
+        alignof(opentelemetry::ext::http::client::SessionState),
+    "std::atomic<SessionState> is more aligned, which changes the layout of HttpOperation");
+
 void HttpOperation::DispatchEvent(opentelemetry::ext::http::client::SessionState type,
                                   const std::string &reason)
 {
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1599,18 +1599,21 @@ void HttpOperation::Abort()
 
 void HttpOperation::FinishUnscheduled(const char *reason)
 {
-  // Ahead of the cleanup: Cleanup() reports a manual cancel for any state short of a terminal
-  // one, with a reason read from a curl result that is still CURLE_OK, and nobody cancelled
-  // this. The enum documents Cancelled as manually cancelled and both exporters print that word.
-  session_state_.store(opentelemetry::ext::http::client::SessionState::CreateFailed,
-                       std::memory_order_release);
-
-  // Ahead of the event, so a handler calling FinishSession() from it is not waiting on the
-  // promise that this cleanup is the only thing able to fulfil.
-  Cleanup();
-
+  // The event first, because the operation holds the handler as a bare pointer and the only
+  // strong reference to it is the one the completion callback captured. Cleanup() takes that
+  // callback and lets it go, so an event dispatched afterwards can be talking to a handler
+  // nothing owns any more. A handler calling FinishSession() from here is inside a callback for
+  // this operation, so Finish() returns rather than waiting on a promise this thread has not
+  // published yet, which is what used to make the other order necessary.
+  //
+  // DispatchEvent stores the state before it calls the handler, so the cleanup below sees a
+  // terminal state and does not report a manual cancel for something nobody cancelled.
   DispatchEvent(opentelemetry::ext::http::client::SessionState::CreateFailed,
                 nullptr != reason ? reason : "");
+
+  // Last, so that a Finish() on another thread waits for the terminal event and the completion
+  // callback rather than for the promise alone.
+  Cleanup();
 }
 
 void HttpOperation::PerformCurlMessage(CURLcode code)
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -403,8 +403,7 @@ namespace
 class CallbackScope
 {
 public:
-  explicit CallbackScope(const HttpOperation *operation) noexcept
-      : operation_{operation}, previous_{current_}
+  explicit CallbackScope(const HttpOperation *operation) noexcept : operation_{operation}
   {
     current_ = this;
   }
@@ -412,7 +411,9 @@ public:
   ~CallbackScope() { current_ = previous_; }
 
   CallbackScope(const CallbackScope &)            = delete;
+  CallbackScope(CallbackScope &&)                 = delete;
   CallbackScope &operator=(const CallbackScope &) = delete;
+  CallbackScope &operator=(CallbackScope &&)      = delete;
 
   static bool InsideCallbackFor(const HttpOperation *operation) noexcept
   {
@@ -427,10 +428,11 @@ public:
   }
 
 private:
-  const HttpOperation *operation_;
-  const CallbackScope *previous_;
-
   static thread_local const CallbackScope *current_;
+
+  const HttpOperation *operation_;
+  // Read before the constructor body replaces it, which is what makes the stack a stack.
+  const CallbackScope *previous_ = current_;
 };
 
 thread_local const CallbackScope *CallbackScope::current_ = nullptr;
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1452,20 +1452,45 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   // Abort() with nothing to do but raise the flag.
   curl_easy_setopt(curl_resource_.easy_handle, CURLOPT_NOPROGRESS, 0L);
 
-  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
+  // A handler can cancel from this event, so everything Cleanup() needs has to be published
+  // before it runs. The callback most of all: Cleanup() takes is_cleaned_ at its start and
+  // swaps the callback much later, so a cleanup racing an assignment placed after the event
+  // would swap an empty one, never run the completion, and leave the session marked active
+  // for good.
   is_finished_.store(false, std::memory_order_release);
   is_aborted_.store(false, std::memory_order_release);
   is_cleaned_.store(false, std::memory_order_release);
-
   async_data_->session.store(session, std::memory_order_release);
-  if (false == async_data_->is_promise_running.exchange(true, std::memory_order_acq_rel))
-  {
-    async_data_->result_promise = std::promise<CURLcode>();
-    async_data_->result_future  = async_data_->result_promise.get_future();
-  }
   async_data_->callback = std::move(callback);
 
-  session->GetHttpClient().ScheduleAddSession(session->GetSessionId());
+  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
+
+  // The future is the one thing that stays unpublished until after the event, so a handler that
+  // calls FinishSession() from it returns instead of waiting on a transfer that has not been
+  // scheduled yet.
+  async_data_->result_promise = std::promise<CURLcode>();
+  async_data_->result_future  = async_data_->result_promise.get_future();
+  async_data_->is_promise_running.store(true, std::memory_order_release);
+
+  // Cancelling from the event hands the session to the IO thread, which may already have torn
+  // the operation down and found no future to complete. Settle it here if so.
+  if (is_cleaned_.load(std::memory_order_acquire))
+  {
+    if (async_data_->is_promise_running.exchange(false, std::memory_order_acq_rel))
+    {
+      async_data_->result_promise.set_value(last_curl_result_);
+    }
+    return CURLE_OK;
+  }
+
+  if (WasAborted() || !session->GetHttpClient().ScheduleAddSession(session->GetSessionId()))
+  {
+    // Nothing is going to run this operation. It was cancelled while the event was dispatched,
+    // or its session was never registered because the URL did not parse. Finish it here rather
+    // than leave a future nobody can complete.
+    Cleanup();
+  }
+
   return CURLE_OK;
 }
 
--- a/ext/src/http/client/curl/http_operation_curl.cc
+++ b/ext/src/http/client/curl/http_operation_curl.cc
@@ -1536,16 +1536,8 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
   }
   else if (!session->GetHttpClient().ScheduleAddSession(session->GetSessionId()))
   {
-    // The same, except nobody cancelled anything. The terminal state goes in first so Cleanup()
-    // does not report a manual cancel, with a reason read from a curl result that is still
-    // CURLE_OK, and the operation is finished before the handler hears about it: a handler that
-    // calls FinishSession() from this event would otherwise wait on the promise that the
-    // Cleanup() below it is the only thing able to fulfil.
-    session_state_.store(opentelemetry::ext::http::client::SessionState::CreateFailed,
-                         std::memory_order_release);
-    Cleanup();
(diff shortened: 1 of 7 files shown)


thc1006:
One thing on this page that a reviewer would otherwise have to dig for.

## The two red jobs are both #4265

`CMake gcc 14 (maintainer mode, abiv2)` and `CMake clang 18 (maintainer mode, sync)` each fail on one test and pass everything else, and it is the same test with the same number in both:

```text
ext/test/http/curl_http_test.cc:881: Failure
Value of: cost < std::chrono::milliseconds{20}
  Actual: false      cost ms: 247
```

That is #4265, which @dbarker opened on 17 July and @yogarajalakshmi-s is on. `main` hit it on 7 August as well.

This change does touch session teardown, so I did not want to leave it at "known flake". Same case, same build configuration, 25 runs on `main` and 25 on this branch:

| | n | median | max | over 20 ms |
| --- | ---: | ---: | ---: | ---: |
| `main` | 25 | 0 ms | 0 ms | 0 |
| this branch | 25 | 0 ms | 0 ms | 0 |

The assertion is a 20 ms wall clock budget, so an idle box gives 0 and a loaded runner gives 247. Nothing here moves it.

**Edited 16 Aug.** This used to carry a second section on the eleven uncovered patch lines. It is gone rather than kept, because the comment below supersedes its numbers and one of its claims stopped being true: it said a test could not make libcurl reject a handle without more machinery than the branch was worth, and the branch now has that seam, so a case names the code it wants back instead of arranging an unusable multi handle.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1451,20 +1451,41 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
...
   async_data_->session = session;
-  if (false == async_data_->is_promise_running.exchange(true, std::memory_order_acq_rel))
+
+  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
+
+  // The future stays unpublished until after the event, so a handler that calls FinishSession()
+  // from it returns instead of waiting on a transfer that has not been scheduled yet.
+  async_data_->result_promise = std::promise<CURLcode>();
+  async_data_->result_future  = async_data_->result_promise.get_future();
+  async_data_->callback       = std::move(callback);
```
lalitb: Could the IO thread be inside `Cleanup()` while this line stores the callback?

If a handler cancels from the `Connecting` event, it schedules an abort and wakes the IO thread. When a background thread is already running, it can enter `Cleanup()` before `SendAsync()` returns from the event and reaches this line.

The `is_cleaned_` check below does not fully protect this. `Cleanup()` sets `is_cleaned_` at its beginning, before it swaps `async_data_->callback`. Seeing `is_cleaned_ == true` therefore tells us that cleanup has started, not that it has finished accessing the callback.

If `Cleanup()` swaps the callback before this assignment, it sees an empty callback and calls nothing. The real callback is then stored here after its only caller has passed. That callback is also what sets `is_session_active_` to false, so the session can remain marked active even though the handler already received `Cancelled`. If the assignment and swap overlap, they may also race on the `std::function`.

Could we synchronize callback publication with cleanup, or publish the callback before dispatching `Connecting`? I understand that publishing the future before the event would make `FinishSession()` risk waiting for a request that has not been scheduled, so that case would need to remain non-blocking.

thc1006: You are right, and the consequence is worse than the callback simply going missing.

I checked the ordering rather than take it at face value. `Cleanup()` takes `is_cleaned_` at `:524` and swaps `async_data_->callback` at `:563`, thirty nine lines apart, so the recheck below only tells me cleanup has begun. If the IO thread reaches the swap first it finds an empty function, and the completion that never runs is the lambda in `Session::SendRequest` that sets `is_session_active_` to false. So the session stays active forever, and for the OTLP client that means `ReleaseSession` is never called, `finished_session_counter_` never moves, and a `ForceFlush` waits on a request that already finished. Plus the `swap` and the assignment on the `std::function` itself.

Worth saying that this window is on `main` too, since the callback is stored after the event there as well. This change makes it much easier to reach, because `Abort()` can now actually schedule the abort and wake the IO thread, where before the cancel was dropped on the floor. So it is mine to close either way.

Published the callback with the flags and the session route, before the event, exactly as you suggested. The future stays behind the event for the reason you named, so `FinishSession()` from a handler still returns rather than waiting on a transfer that has not been scheduled.

```cpp
  is_finished_.store(false, std::memory_order_release);
  is_aborted_.store(false, std::memory_order_release);
  is_cleaned_.store(false, std::memory_order_release);
  async_data_->session  = session;
  async_data_->callback = std::move(callback);

  DispatchEvent(...SessionState::Connecting);

  async_data_->result_promise = std::promise<CURLcode>();
  async_data_->result_future  = async_data_->result_promise.get_future();
  async_data_->is_promise_running.store(true, std::memory_order_release);
```

That also gives the publication a happens-before it did not have: the callback is written before the handler can cancel, the cancel goes through `ScheduleAbortSession` under `sessions_m_`, and the IO thread takes the same lock before it can get to `Cleanup()`.

26 of 26 locally and 26 of 26 under ThreadSanitizer with no warnings, and the three cases this change adds run fifteen times without a failure and five times under TSAN without one.

**Update, 16 Aug.** The ordering shown above changed in `237ddba6`, which moved the event below the whole publication block, so the future no longer sits behind it. What keeps `FinishSession()` from a handler returning is now the callback scope `Finish()` consults, and that is also why this PR closes #4402. Leaving the rest as written, since it is the reasoning the change at the time was based on.

ext/src/http/client/curl/http_operation_curl.cc
```diff
@@ -1452,20 +1452,45 @@ CURLcode HttpOperation::SendAsync(Session *session, std::function<void(HttpOpera
...
+  // swaps the callback much later, so a cleanup racing an assignment placed after the event
+  // would swap an empty one, never run the completion, and leave the session marked active
+  // for good.
   is_finished_.store(false, std::memory_order_release);
   is_aborted_.store(false, std::memory_order_release);
   is_cleaned_.store(false, std::memory_order_release);
+  async_data_->session  = session;
+  async_data_->callback = std::move(callback);
+
+  DispatchEvent(opentelemetry::ext::http::client::SessionState::Connecting);
```
lalitb: Thanks, the callback publication reads correct now.

One more concern on this line, from the same window: if the client already has an IO thread, a handler can call `CancelSession()` from `Connecting`, and `Cleanup()` can dispatch `Cancelled` on the IO thread before this `Connecting` callback returns. That also makes both `DispatchEvent()` calls write `session_state_` concurrently.

The current test uses a new client, so the background thread starts only after `SendAsync()` returns and does not cover this interleaving. Could we serialize cleanup with the initial event dispatch and add a test that starts the IO thread before cancelling from `Connecting`?

thc1006: Measured it, and both halves hold.

**The interleaving happens.** On a client that has already completed one request, a handler cancelling from `Connecting` is still inside `OnEvent` when the IO thread runs `Cleanup()` and dispatches its own `Cancelled`. I held the handler open and counted threads inside it: two, five runs out of five. Drop the warming request and it is one, which is what every case in the file does today.

**The stores race.** ThreadSanitizer on that scenario, on this branch before the change, five runs out of five:

```text
  Write of size 1 by main thread:
    #0 HttpOperation::DispatchEvent(...)  http_operation_curl.cc:415
    #1 HttpOperation::SendAsync(...)      http_operation_curl.cc:1466
    #2 Session::SendRequest(...)          http_client_curl.cc:209

  Previous write of size 1 by thread T2:
    #0 HttpOperation::DispatchEvent(...)  http_operation_curl.cc:415
    #1 HttpOperation::Cleanup()           http_operation_curl.cc:536
    #2 Session::FinishOperation()         http_client_curl.cc:269
    #3 HttpClient::doAbortSessions()      http_client_curl.cc:808
```

Past the undefined behavior the later store wins, so `GetSessionState()` finished at `Connecting` for an operation that had been cancelled.

**What changed.** `DispatchEvent` stores before it calls the handler, and `session_state_` is now a `std::atomic`. Storing first is what fixes the value: the store is then ordered ahead of the cancel, which goes through `sessions_m_` and `session_ids_m_`, and the IO thread takes `session_ids_m_` before it can reach `Cleanup()`. The atomic covers the paths with no such handoff, where a third thread cancels rather than the handler. `SessionState` is a `std::uint8_t` enum and `std::atomic` of it is one byte with alignment one and lock free, so the layout of the installed type does not move.

I went through the other readers before moving the store. Nothing reads `GetSessionState()` from inside its own event, and every internal reader, `~HttpOperation`, `Cleanup`, the progress and debug callbacks, and `PerformCurlMessage`, reads it after a dispatch has already returned, where both orders give the same value. The two synchronous `Get` and `Post` paths run on one thread with no handler at all.

**The test.** `CancelFromConnectingWhilePollingCompletes` warms the client, cancels from `Connecting`, and stays in the handler until the IO thread has entered it too, so the overlap is deterministic rather than lucky. Without the change it reports the race five runs out of five, with it none, and the whole file is clean under ThreadSanitizer. It is also the only thing in this file that reaches the `is_cleaned_` recheck at the end of `SendAsync`, which closes the coverage gap I flagged on the Codecov comment. I instrumented both tail branches and ran the whole binary: the recheck was entered exactly once, by this scenario, and the three cases already on the branch all took the other branch instead.

One detail worth knowing for anyone who touches it. My first version counted the overlap with `acq_rel` atomics and ThreadSanitizer went quiet. The IO thread's release and the caller's later read-modify-write on the same counter synchronize with each other, which orders the two state stores and hides exactly what the case exists to catch. The counters are `relaxed` now. I confirmed that with a deliberate unsynchronized counter on the same line as a control: the `acq_rel` build reported neither it nor the real race, the `relaxed` build reported both.

I had opened #4408 for this before deciding to fix it here, since it reproduces on `main` independently of this change. This branch closes it, and I have corrected the claim there that the member could not be made atomic without moving the layout.

thc1006:
Update on the eleven uncovered patch lines I wrote about earlier, now that the polling client case is on the branch. Measured with the same `all-options-abiv2-preview` configuration `ci/do_ci.sh code.coverage` uses, intersecting the uncovered lines with the lines this branch adds.

| | before | now |
| --- | ---: | ---: |
| `http_operation_curl.cc` | 3 | **0** |
| `http_client_curl.cc` | 8 | 8 |

The three were the `is_cleaned_` recheck at the end of `SendAsync`, which I had said nothing in the suite reached. `CancelFromConnectingWhilePollingCompletes` reaches it, and it is still the only thing that does: instrumenting both tail branches and running the whole binary shows the recheck entered exactly once, by that case.

Of the eight left, three are the `continue` guards at the top of `doAddSessions`, unchanged except for the indentation of wrapping that loop in a scope, so the diff counts them as added. They were not executed before this change either.

The other five are the `curl_multi_add_handle` rejection branch, and I owe a better answer than the one I gave. I said a test cannot make libcurl reject a handle. What I have since measured is more specific than that. A client whose multi handle is null does reject, but never in `doAddSessions`: the IO loop calls `curl_multi_perform` first, that rejects the null handle, the `mc != CURLM_OK` branch resets it, and `doAddSessions` then runs against a working handle. I confirmed that on a client built with `curl_multi_init` forced to fail, with the internal log captured in the same run to prove the handle really was null: the request completed normally, three runs out of three.

So reaching that branch needs the handle to be taken away and `doAddSessions` driven directly, which the `HttpClientTestPeer` friend added by #4394 can do in about fifteen lines. The reason I have not sent it is that the peer would have to win a race against the IO loop's reset, and a case that covers the branch only when it wins is a flaky case rather than a test. If you would rather have that branch covered than have the suite stay deterministic, say so and I will add it with the peer holding `multi_handle_m_` across the call, which makes the reset wait rather than race.

ext/src/http/client/curl/http_operation_curl.cc
thc1006: One correction to what I wrote above, because I stated it more broadly than it holds.

I said every internal reader takes `GetSessionState()` after a dispatch has already returned. There is one that does not have to. `~HttpOperation` switches on it to decide whether to wait on the future, and a handler that destroys its own operation from inside an event, which is what `Session::SendRequest` does when a session is reused and is the shape of #4396, runs that switch while the dispatch is still in the handler. Its case list is `Connecting`, `Connected`, `Sending`, and it does not include `Created`, so storing first can change which side of it that path lands on.

I measured it rather than reason about it. Instrumenting the destructor to record every state it observes and running the same set of cases both ways:

```text
store after the handler:   Response 28  ConnectFailed 11  SendFailed 2  Created 2  Connecting 1
store before the handler:  Response 29  ConnectFailed 11  SendFailed 1  Created 2  Connecting 1
```

The one that moves between `SendFailed` and `Response` is run to run variation in a suite that talks to a socket, not the ordering. So nothing the suite exercises changes, and the only path where the two could differ is a handler destroying its own operation, which is a use after free today with or without this.

